PR #2, opened 2021-03-03
Support http requests

njgheorghita:
- added [clap library](https://docs.rs/clap/2.33.3/clap/index.html) for cli arg parsing
- added some tests for cli args
- added support for http requests


src/main.rs
```diff
@@ -1,13 +1,186 @@
...
-        Ok(val) => cli::launch_trin(val),
-        Err(_) => println!(
-            "Must supply Infura key as environment variable, like:\n\
-            TRIN_INFURA_PROJECT_ID=\"your-key-here\" trin"
-        ),
+    let trin_config = TrinConfig::new();
+    cli::launch_trin(trin_config);
+}
+
+#[cfg(test)]
```
njgheorghita, line 101: From what I've read - to do proper unit testing we'll need to pull out the majority of the logic from the binary (`main.rs`) into the library (`lib.rs`) and just replace `main.rs` with a thin wrapper around the library logic. I've just included the tests here for now, but can do the re-organizing in this or a separate pr.

src/main.rs
```diff
@@ -1,13 +1,186 @@
...
+        Self::new_from(std::env::args_os().into_iter()).unwrap_or_else(|e| e.exit())
+    }
+
+    fn new_from<I, T>(args: I) -> Result<Self, clap::Error>
+    where
+        I: Iterator<Item = T>,
+        T: Into<OsString> + Clone,
+    {
+        // TODO: things to configure:
+        //  - infura project id (not just env var?)
```
njgheorghita, line 20: Any further thoughts on this? At the moment, using just an env var makes testing slightly tricky. Some other options include a config file or a cli arg (i'm not a fan of the cli arg option, it would get rather tedious imo)

njgheorghita:
First rust pr! So I'm sure there's plenty of room for improvement here, but it's ready for a first pass :eyes:

njgheorghita:
ping @carver for :eyes: ... not sure why, but you're not showing up in the `Reviewers` section at the top right of the pr. But maybe that's because this repo is still so fresh

carver:
Thanks!

I assume you ran `rustfmt` or something. (I may not have) Just wondering why the IPC code is showing up as changed.

carver:
> First rust pr! So I'm sure there's plenty of room for improvement here, but it's ready for a first pass

Yeah, sort of the blind leading the blind here...

carver:
> I assume you ran `rustfmt` or something. (I may not have) Just wondering why the IPC code is showing up as changed.

Yeah, just confirmed that I left out the formatting last time. My bad, sorry for the extra work. Would you mind rebasing now? It will make the diff easier for me to read.

src/main.rs
carver: Yeah, I think all of the `impl TrinConfig` can go into the `cli.rs` and be tested there, leaving only the `main()` here.

src/cli.rs
```diff
@@ -28,25 +38,25 @@ pub fn launch_trin(infura_project_id: String) {
...
             let mut rx = stream.try_clone().unwrap();
             let mut tx = stream;
-            serve_client(&mut rx, &mut tx, &infura_url);
+            serve_ipc_client(&mut rx, &mut tx, &infura_url);
         });
     }
 }
 
-fn serve_client(rx: &mut impl Read, tx: &mut impl Write, infura_url: &String) {
+fn serve_ipc_client(rx: &mut impl Read, tx: &mut impl Write, infura_url: &String) {
```
carver, line 59: This, `serve_http_client`, and `launch_*_client` all probably make more sense in a new `src/jsonrpc.rs` file, but let's save that for the next PR. (`TrinConfig` though, makes a lot of sense in `cli.rs`, to me)

src/cli.rs
```diff
@@ -3,20 +3,30 @@ use reqwest::blocking as reqwest;
...
+}
 
-pub fn launch_trin(infura_project_id: String) {
-    println!("Launching with infura key: '{}'", infura_project_id);
+pub fn launch_trin(trin_config: TrinConfig) {
+    let pool = ThreadPool::new(trin_config.pool_size as usize);
 
-    let pool = ThreadPool::new(2);
+    match &trin_config.protocol[..] {
+        "ipc" => launch_ipc_client(pool, trin_config),
```
carver, line 23: Maybe pass in `trin_config.infura_project_id` instead, so the downstream methods don't need to depend on `TrinConfig`.

src/cli.rs
```diff
@@ -56,31 +66,115 @@ fn serve_client(rx: &mut impl Read, tx: &mut impl Write, infura_url: &String) {
...
+fn launch_http_client(pool: ThreadPool, trin_config: TrinConfig) {
+    let uri = format!("127.0.0.1:{}", trin_config.endpoint);
+    let listener = TcpListener::bind(uri).unwrap();
+    for stream in listener.incoming() {
+        match stream {
+            Ok(stream) => {
+                let infura_project_id = trin_config.infura_project_id.clone();
+                pool.execute(move || {
+                    let infura_url =
+                        format!("https://mainnet.infura.io:443/v3/{}", infura_project_id);
```
carver, line 103: Now that this magic URL string is duplicated, seems like it should be pulled into a new `fn get_infura_url(project_id)`

src/main.rs
carver: I'm okay leaving it as env var only for now. I think we'll want support for that at some point. (One plausible use case is as a colocated endpoint for a web server, where env var would be the nicest way to configure IMO)

So I guess we'll just have to figure out a way to test where we can mock out the env var.

carver:
A natural follow-up will be to figure out a way to do integration tests on the http and ipc servers. Maybe we can hit only the `clientVersion` endpoint for now, though we'll obviously want to test Infura endpoints eventually. Also, we'll probably want arbitrary backing URLs. That's important soon, so we don't hit Infura for real during tests, and later so we can support any type of endpoint (like [alchemyapi.io](https://www.alchemyapi.io/)).

carver:
Oh, also a reminder to update the README

src/main.rs
njgheorghita: Yeah, that seems reasonable. I think once we get CI setup, it'll help illuminate the best way forward. Also, from the [docs](https://doc.rust-lang.org/book/ch11-02-running-tests.html)...

> Because the tests are running at the same time, make sure your tests don’t depend on each other or on any shared state, including a shared environment, such as the current working directory or environment variables.

Which seems to me rules out any tests that check for the presence of an env variable. From my tinkering around any test that manually cleared and reset the env var (to check for presence) would cause every other test to simultaneously fail.

src/cli.rs
njgheorghita: 👍 made the change for `launch_ipc_client` but left the args for `launch_http_client` as is since that function also needs access to `trin_config.endpoint`.

njgheorghita:
@carver Ready for another :eyes:

src/main.rs
carver: Yeah, I think if we want to keep the env var long term, we would read the value in `main()` and pass it into the inner function call. Then, all layers beneath it are just tested with supplied arguments.

carver (approved):
Looks good!

Cool that building the raw HTTP wasn't too bad.

For some of the comments, I mention follow-up work. Would you mind opening short issues for each of them that you don't end up fixing in this PR? (Including ones I said I'd do :P )

README.md
```diff
@@ -31,6 +31,22 @@ cd trin
...
+```sh
+USAGE:
+    trin [OPTIONS]
+
+FLAGS:
+    -h, --help       Prints help information
+    -V, --version    Prints version information
+
+OPTIONS:
+    -e, --endpoint <endpoint>      http port [default: 7878]
```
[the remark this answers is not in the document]
njgheorghita: Yeah, that makes sense, it seems unlikely that somebody would want to run both at the same time

src/cli.rs
```diff
@@ -1,22 +1,120 @@
...
+                    .short("p")
+                    .long("protocol")
+                    .help("select transport protocol")
+                    .takes_value(true)
+                    .default_value("http"),
+            )
+            .arg(
+                Arg::with_name("endpoint")
+                    .short("e")
+                    .long("endpoint")
```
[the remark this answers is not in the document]
njgheorghita: I like `"http_port"` - just `"hp"` won't be possible as the short flag. [short](https://docs.rs/clap/2.33.3/clap/struct.Arg.html#method.short) automatically strips whatever the argument is down to just the first level. This made it kinda hard to come up with neat intuitive short flags. Though it seems to me like supporting a confusing short flag is better than no short flags? Also check out that the short flag for `"pool_size"` is currently `"s"` (since `"protocol"` gets `"p"`). But I'm happy to switch to whatever naming convention you feel makes sense.

src/cli.rs
```diff
@@ -1,22 +1,120 @@
...
+                    .takes_value(true)
+                    .default_value("2"),
+            )
+            .get_matches_from_safe(args)
+            .unwrap_or_else(|e| panic!("Unable to parse args: {}", e));
+
+        println!("Launching Trin...");
+        let protocol = matches.value_of("protocol").unwrap();
+        let endpoint = matches.value_of("endpoint").unwrap();
+        let endpoint = match endpoint.parse::<u32>() {
```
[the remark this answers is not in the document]
njgheorghita: If I understand correctly, then in the case that ipc protocol is selected - we don't care to check a `http_port` argument might also be incorrectly provided? It makes sense - after all the name `http_port` is pretty indicative that it's incompatible with ipc protocol.

src/cli.rs
```diff
@@ -1,22 +1,120 @@
...
+            pool_size: pool_size,
+            protocol: protocol.to_string(),
+        })
+    }
+}
+
+pub fn launch_trin(trin_config: TrinConfig) {
+    let pool = ThreadPool::new(trin_config.pool_size as usize);
+
+    match &trin_config.protocol[..] {
```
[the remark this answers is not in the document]
njgheorghita: The deep mechanics - I'm not sure I can entirely explain 100%. `[..]` is syntax to splice the entire string - so my intuition was that it's necessary that the expression returns a value to `match` against.  It was inspired from [stack-overflow](https://stackoverflow.com/questions/25383488/how-to-match-a-string-against-string-literals) - but now I notice that there's a `as_str()` method which is cleaner imo and performs the same function.

src/cli.rs
```diff
@@ -81,6 +179,91 @@ fn serve_client(rx: &mut impl Read, tx: &mut impl Write, infura_url: &String) {
...
+
+    let deser = serde_json::Deserializer::from_str(&json_request);
+    for obj in deser.into_iter::<serde_json::Value>() {
+        let obj = obj.unwrap();
+        assert!(obj.is_object());
+        assert_eq!(obj["jsonrpc"], "2.0");
+        let request_id = obj.get("id").unwrap();
+        let method = obj.get("method").unwrap();
+
+        let response = match method.as_str().unwrap() {
```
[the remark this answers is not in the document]
njgheorghita: 👍 I've added this to be tackled in the refactor issue #6

src/cli.rs
```diff
@@ -1,22 +1,120 @@
...
+            .arg(
+                Arg::with_name("pool_size")
+                    .short("s")
+                    .long("pool-size")
+                    .help("max size of threadpool")
+                    .takes_value(true)
+                    .default_value("2"),
+            )
+            .get_matches_from_safe(args)
+            .unwrap_or_else(|e| panic!("Unable to parse args: {}", e));
```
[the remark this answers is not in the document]
njgheorghita: It turns out `get_matches_from_safe` has some [peculiarities](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from_safe) in terms of handling `--help` and `--version` and will panic if those are called. It felt weird to me to panic on help and version input, so I added some handling to exit gracefully on those calls and panic for any other errors.

src/cli.rs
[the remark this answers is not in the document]
njgheorghita: I found a clap macro (`value_t`) that simplifies the code parsing/casting the cli args. How strongly do you feel about this? I see the benefit of a constant.. But, I've also tried a couple variants and haven't found a reasonably simple solution (but, my rust type-fu is weak so who knows what obvious solutions i'm missing). 

- since `TrinConfig` has defined `http_port: u32`, we eventually have to cast a `const http_port: String = "8545"` into an int when we use the default constant (in both the http & ipc cases)
- the `default_arg()` param for the clap parser only accepts strings, so we have to do some casting if `const http_port: u32 = 8545`
- A performance benefit is somewhat irrelevant if a user selects `ipc` and a custom `http_port` since we error out. In the case that the user just selects `ipc`, we will still have to cast a `const http_port: String = "8545"` into a string to feed into `TrinConfig`

I'm happy to keep tinkering with this, just reporting my initial findings

njgheorghita:
@carver Sorry, was a bit sporadic in wrapping this up. Just a couple of last comments if you have any thoughts, and then I'll get this merged

src/cli.rs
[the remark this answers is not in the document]
carver: Oh weird. Well I think if we're not going to do anything clever on a failure, maybe we should just use:
https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from ?

njgheorghita:
@carver Ahh, I don't have write access to this repo - so for now I'll have to leave merging up to you if it checks out on your end

PR #8, opened 2021-03-29
Support custom ipc paths

### LINKED ISSUE #5: Allow specifying the IPC location
Right now, it's hard-coded to `/tmp/trin-jsonrpc.ipc`

### LINKED ISSUE #7: Extract client logic from cli module into new jsonrpc module
> This, serve_http_client, and launch_*_client all probably make more sense in a new src/jsonrpc.rs file, but let's save that for the next PR. (TrinConfig though, makes a lot of sense in cli.rs, to me)

https://github.com/carver/trin/pull/2/files#r587927213

njgheorghita:
Ready for review after #2 is merged

- Fixes #7 (extract jsonrpc logic into separate module)
- Fixes #5 (support custom ipc paths)
- Updated `DEFAULT_HTTP_PORT` and `DEFAULT_IPC_PATH` to be constants

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,5 +1,4 @@
 use clap::{value_t, App, Arg};
-use reqwest::blocking as reqwest;
 use std::env;
 use std::ffi::OsString;
 
@@ -67,6 +66,7 @@ impl TrinConfig {
         let pool_size = value_t!(matches.value_of("pool_size"), u32)?;
         let http_port = value_t!(matches.value_of("http_port"), u32)?;
         let protocol = value_t!(matches.value_of("protocol"), String)?;
+        let ipc_path = value_t!(matches.value_of("ipc_path"), String)?;
 
         match protocol.as_str() {
             "http" => match &ipc_path[..] {
@@ -74,7 +74,7 @@ impl TrinConfig {
                 _ => panic!("Must not supply an ipc path when using http protocol"),
             },
             "ipc" => match &http_port.to_string()[..] {
-                DEFAULT_HTTP_PORT => println!("Protocol: {}", protocol),
+                DEFAULT_HTTP_PORT => println!("Protocol: {}\nIPC path: {}", protocol, ipc_path),
                 _ => panic!("Must not supply an http port when using ipc protocol"),
             },
             val => panic!("Unsupported protocol: {}", val),
@@ -84,6 +84,7 @@ impl TrinConfig {
 
         Ok(TrinConfig {
             http_port,
+            ipc_path,
             pool_size,
             protocol,
         })
@@ -93,6 +94,7 @@ impl TrinConfig {
 #[cfg(test)]
 mod test {
     use super::*;
+    use std::env;
 
     fn env_is_set() -> bool {
         match env::var("TRIN_INFURA_PROJECT_ID") {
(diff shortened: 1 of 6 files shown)


src/cli.rs
```diff
@@ -63,15 +67,18 @@ impl TrinConfig {
...
-                println!("Protocol: {}\nHTTP port: {}", protocol, http_port)
-            }
-            "ipc" => match http_port {
-                8545 => println!("Protocol: {}", protocol),
+            "http" => match &ipc_path[..] {
+                DEFAULT_IPC_PATH => println!("Protocol: {}\nHTTP port: {}", protocol, http_port),
+                _ => panic!("Must not supply an ipc path when using http protocol"),
+            },
+            "ipc" => match &http_port.to_string()[..] {
+                DEFAULT_HTTP_PORT => println!("Protocol: {}", protocol),
```
carver, line 81: Would be nice to print the IPC path here.

src/jsonrpc.rs
```diff
@@ -0,0 +1,199 @@
...
+fn launch_ipc_client(pool: ThreadPool, infura_project_id: String, ipc_path: &String) {
+    let listener_result = unix::net::UnixListener::bind(ipc_path);
+    let listener = match listener_result {
+        Ok(listener) => listener,
+        Err(err) if err.kind() == io::ErrorKind::AddrInUse => {
+            // TODO something smarter than just dropping the existing file and/or
+            // make sure file gets cleaned up on shutdown.
+            match fs::remove_file(ipc_path) {
+                Err(_) => panic!("Could not serve IPC from existing path '{}'", ipc_path),
+                Ok(()) => unix::net::UnixListener::bind(ipc_path).unwrap(),
```
carver, line 29: So this approach no longer seems acceptable with a configurable IPC path (just dropping whatever file is found there and replacing it).

The simple answer is that we should just panic instead of trying to remove the file and reattempting. But then to solve the original problem: I think that we were leaking the file on shutdown sometimes. (Maybe anytime there's a panic?)

So the next step is to make sure that the IPC file gets removed when trin shuts down, even after a panic. (Although, it's probably worth verifying the theory. All I really know is that sometimes it used to happen that I had a pre-existing file on startup, and it was annoying. Not sure where it came from).

carver:
I took the liberty of rebasing

carver:
Reminder to rebase on master to get CI tests

src/jsonrpc.rs
njgheorghita: I updated the code to add some handlers that will clean up the ipc file upon `ctrl+c` shutdown and all panics. It works as expected locally, but could use some integration tests to make sure the cleanup happens as expected. I'm leaning towards saving those for a larger integration test pr coming soon<sup>tm</sup>.

njgheorghita:
@carver ping for another :eyes:. I updated the settings in circle ci to `"Build forked pull requests"` and `"Pass secrets to builds from forked pull requests"` - both of which are necessary to get ci working here.  FYI re: passing secrets -> `"passing secrets to builds from forked pull requests can cause security vulnerabilities"`- which seems fine since we're using a dummy infura key for now. Even when we use a real infura key for integration testing, imo it doesn't warrant a more secure setup

PR #12, opened 2021-04-13
[WIP] Initial integration testing

### LINKED ISSUE #4: Initial Integration tests for HTTP / IPC servers
From https://github.com/carver/trin/pull/2#issuecomment-791053427

> A natural follow-up will be to figure out a way to do integration tests on the http and ipc servers. Maybe we can hit only the clientVersion endpoint for now, though we'll obviously want to test Infura endpoints eventually. Also, we'll probably want arbitrary backing URLs. That's important soon, so we don't hit Infura for real during tests, and later so we can support any type of endpoint (like alchemyapi.io).

njgheorghita:
Fixes #4 .
Integration tests for the `web3_clientVersion` and `eth_blockNumber` endpoints - for both http and ipc protocols.

- tests are run synchronously using `serial` package. this slows down tests significantly, however the overhead of configuring each test to not interfere with the others as they are run in parallel justifies the speed tradeoff imo. 
- I've created an infura project id and updated the environment variable in circleci

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,6 +67,12 @@ checksum = "e3c69b077ad434294d3ce9f1f6143a2a4b89a8a2d54ef813d85003a4fd1137fd"
 
 [[package]]
 name = "cfg-if"
+version = "0.1.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


njgheorghita:
closing this for now to clean up the pr list - and since the code base has moved along significantly since. there's still an issue tracking this for future reference. when i free up more time from nethermind plugin work, i'll pick this up if the issue's still open.

PR #13, opened 2021-05-06
Rename CLI arguments for disambiguation

carver:
There will be ports used on both the p2p endpoint and the json-rpc
endpoint. It's too ambiguous to have a CLI argument of simply "http
port".

Also, single-letter flags are too short to handle the argument
complexity that's coming. Not to mention that we should not overwrite
the -h (help) arg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,12 @@ In a python shell:
 ```
 
 ### Connect over HTTP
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
👍 LGTM!

PR #17, opened 2021-05-14
Try to ping the bootnodes on the overlay network, at launch time

carver:
This makes for an easier cross-client ping test. Once the handshake is working, the intention would be to immediately start populating the buckets for the overlay network.

Also, fix bug where ENR sequence was only 32 bit.

It should be 64-bit according to discv5 and the spec: 
https://github.com/ethereum/devp2p/blob/6eddaf50298d551a83bcc242e7ce7024c6cc8590/enr.md

### DIFF
--- a/src/alexandria/protocol.rs
+++ b/src/alexandria/protocol.rs
@@ -29,7 +29,7 @@ impl Default for PortalConfig {
     }
 }
 
-pub const PROTOCOL: &str = "state-network";
+pub const PROTOCOL: &str = "portal";
 
 pub struct AlexandriaProtocol {
     discovery: Discovery,
--- a/src/alexandria/protocol.rs
+++ b/src/alexandria/protocol.rs
@@ -143,7 +143,7 @@ impl AlexandriaProtocol {
             Request::Ping(Ping { .. }) => {
                 let enr_seq = self.discovery.local_enr().seq();
                 Response::Pong(Pong {
-                    enr_seq: enr_seq as u32,
+                    enr_seq: enr_seq,
                     data_radius: self.data_radius,
                 })
             }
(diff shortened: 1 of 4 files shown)


src/alexandria/protocol.rs
```diff
@@ -163,4 +161,16 @@ impl AlexandriaProtocol {
 
         Ok(response)
     }
+
+    /// Convenience call for testing, quick way to ping bootnodes
+    pub async fn ping_bootnodes(&mut self) -> Result<(), String> {
+        // trigger bonding with bootnodes
+        self.discovery.discover_nodes().await?;
```
njgheorghita, line 168: My understanding of `discover_nodes()` is that it pings random node ids to build the table, not any special hardcoded / cli-input "bootnode". Am I misunderstading? Maybe it's just semantic technicality and in practice any randomly pinged node id is in effect acting as a bootnode.

njgheorghita (approved):
Looks good!

src/alexandria/protocol.rs
carver: Yeah, this definitely would be better with an explanation. Discovery pings the bootnodes, because the bootnodes are the only ENRs available to ask. That's the only way it can try to discover the random node ID that becomes the target. Importantly, it doesn't bond with all bootnodes, just (at least) one of them.

I kind of don't expect this method to be around for long (famous last words), but it's still worth adding a comment for sure.

carver: Actually, the even simpler answer is that it's totally unnecessary, because if we do the portal ping over `TALKREQ`, that will automatically trigger a session on the base layer.

PR #18, opened 2021-05-14
Rename Alexandria to Portalnet

### LINKED ISSUE #16: Rename module alexandria to statenet
> Ah, I think we should probably update the python one actually. "alexandria" should be the chain history overlay network. Which is also a reminder to rename this module from alexandria to something like `statenet`.

_Originally posted by @carver in https://github.com/carver/trin/pull/14#discussion_r631476900_

carver:
Fixes #16 

Alexandria is a great name, but makes way more sense to use for chain history instead of the active state network. (Chain history was the first usage of the name Alexandria, as I understand it.)

Bonus change: alphabetize Cargo.toml


carver:
First two commits are waiting in #17 -- I'll rebase after that's merged before pinging for review.

src/main.rs
```diff
@@ -10,8 +10,8 @@ use log::info;
 use std::env;
 use std::time::Duration;
 
-mod alexandria;
-use alexandria::protocol::{AlexandriaProtocol, PortalConfig};
+mod portalnet;
+use portalnet::protocol::{PortalConfig, PortalnetProtocol};
```
njgheorghita, line 14: Is it worth updating `PortalConfig` to `PortalnetConfig` to maintain consistency? It's not obvious to me that one way is better than the other - or that consistency in this case greatly improves clarity, since the intent is fairly obvious.

njgheorghita (approved):
:+1:

PR #20, opened 2021-05-20
Configure p2p port and bootnode ENRs via command line

carver:
Bonus: switch the http port to `uint16`, which is sufficient for network ports.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -6,7 +6,7 @@ use std::ffi::OsString;
 pub struct TrinConfig {
     pub web3_transport: String,
     pub web3_ipc_path: String,
-    pub web3_http_port: u32,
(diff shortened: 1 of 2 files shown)


src/cli.rs
```diff
@@ -6,12 +6,15 @@ use std::ffi::OsString;
 pub struct TrinConfig {
     pub web3_transport: String,
     pub web3_ipc_path: String,
-    pub web3_http_port: u32,
+    pub web3_http_port: u16,
```
njgheorghita, line 9: Yeah, this is basically like a hangover coming from python land when I'm used to treating all integers as the same type (`u32`) - need to drink some coffee and sober up here in rust world.

njgheorghita (approved):
Looks good!

PR #21, opened 2021-05-25
Log discv5 to stderr

carver:
The `tracing` package doesn't use `env_logger` rules by default. We need to initialize it.

Includes two bonus fixes:
- when `--bootnodes` is empty, or includes empty entries, don't crash
- run until ctrl-c properly, instead of just a very long sleep


njgheorghita (approved):
:+1:

PR #26, opened 2021-05-27
STUN support

### LINKED ISSUE #23: Use STUN server to speed up the external socket address discovery
Add a client like `stunclient` to trin, and run `rustun` server with something like:
```sh
cargo run --release --example binding_srv
```

On launch, we can check our external address to create an ideal ENR much faster than waiting for a bunch of independent PONGs to come in. (It also simplifies bootstrapping, where we launch new peers and connect them to each other)

carver:
Closes #23 

- Bind against 0.0.0.0 instead of loopback: Binding to 127.0.0.1 prevents messages from properly sending to a non-loopback address
- Use external STUN server to populate IP & port in the initial ENR
- If STUN check fails, pick a local IP addr for the ENR. That way, you can use the encoded ENR immediately from another (local network) peer

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ env_logger = "0.8.2"
 eth2_ssz = "0.1.2"
 eth2_ssz_derive = "0.1.0"
 futures = "0.3.13"
+get_if_addrs = "0.5.3"
 lazy_static = "1.4.0"
 log = "0.4.14"
 reqwest = { version = "0.11.0", features = ["blocking"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ reqwest = { version = "0.11.0", features = ["blocking"] }
 rlp = "0.5.0"
 serde = {version = "1.0.125", features = ["derive"] }
 serde_json = "1.0.59"
+stunclient = "0.1.2"
 threadpool = "1.8.1"
 tracing = "0.1.26"
 tracing-subscriber = "0.2.18"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ env_logger = "0.8.2"
 eth2_ssz = "0.1.2"
 eth2_ssz_derive = "0.1.0"
 futures = "0.3.13"
-get_if_addrs = "0.5.3"
+interfaces = "0.0.7"
 lazy_static = "1.4.0"
 log = "0.4.14"
 reqwest = { version = "0.11.0", features = ["blocking"] }
--- a/src/portalnet/protocol.rs
+++ b/src/portalnet/protocol.rs
@@ -115,16 +115,32 @@ impl TalkReqHandler for ProtocolHandler {
     }
 }
 
+/// Ping a STUN server on the public network. This does two things:
+/// - Creates an externally-addressable UDP port, if you are behind a NAT
+/// - Returns the public IP and port that corresponds to your local port
+fn punch_nat(local_socket_addr: &SocketAddr) -> SocketAddr {
+    let socket = std::net::UdpSocket::bind(local_socket_addr).unwrap();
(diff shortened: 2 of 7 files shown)


src/portalnet/socket.rs
```diff
@@ -0,0 +1,56 @@
...
+        external_addr,
+    );
+    external_addr.ok()
+}
+
+pub fn default_local_address(port: u16) -> SocketAddr {
+    let ip = match first_interface() {
+        Some(interface) => interface.addr.ip(),
+        None => "127.0.0.1".parse().unwrap(),
+    };
```
g-r-a-n-t, line 26: [`map_or`](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or) or `map_or_else` might be neater

src/portalnet/protocol.rs
```diff
@@ -52,22 +54,30 @@ impl TalkReqHandler for ProtocolHandler {
 
 impl PortalnetProtocol {
     pub async fn new(portal_config: PortalnetConfig) -> Result<Self, String> {
+        let listen_all_ips = SocketAddr::new("0.0.0.0".parse().unwrap(), portal_config.listen_port);
+
+        let external_addr = match socket::stun_for_external(&listen_all_ips) {
+            Some(addr) => addr,
+            None => socket::default_local_address(portal_config.listen_port),
+        };
```
g-r-a-n-t, line 62: [`unwrap_or_default`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_default)

g-r-a-n-t (commented):
just linked a few helpful functions

src/portalnet/protocol.rs
carver: Hm, this method might be better called `discover_local_address`. If I understand `unwrap_or_default` correctly, I wouldn't be able to pass in any variable (like the port to listen on) to the default definition. (and would rust even let me define a Default for a `SocketAddr`, since it's in an external module?)

carver:
@SamWilsn would you mind taking a look at this too?

src/portalnet/socket.rs
```diff
@@ -0,0 +1,58 @@
...
+    debug!(
+        "STUN claims that public network endpoint is: {:?}",
+        external_addr,
+    );
+    external_addr.ok()
+}
+
+pub fn default_local_address(port: u16) -> SocketAddr {
+    let ip = first_interface().map_or_else(
+        || "127.0.0.1".parse().unwrap(),
```
SamWilsn, line 26: [`Ipv4Addr::LOCALHOST`](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#associatedconstant.LOCALHOST)

Cargo.toml
```diff
@@ -14,12 +14,14 @@ env_logger = "0.8.2"
 eth2_ssz = "0.1.2"
 eth2_ssz_derive = "0.1.0"
 futures = "0.3.13"
+get_if_addrs = "0.5.3"
```
SamWilsn, line 17: I'm worried that this hasn't been updated in some time. Some alternatives might be [`heim`](https://crates.io/crates/heim) or [`interfaces`](https://crates.io/crates/interfaces). I think `heim` is more mature.

src/portalnet/protocol.rs
g-r-a-n-t: ah yes, of course

Cargo.toml
carver: I got stuck on getting async working with `heim`. If you feel like pairing with me to fix it, that would be cool. Otherwise, I've got an implementation with `interfaces` ready to go in c2a564c.

SamWilsn: I'm fine with `interfaces` if you are, but I'm okay with pairing too. I think you'll want to do something like:

```rust
use futures_util::stream::TryStreamExt;

let ifaces: Result<Vec<_>, _> = heim::net::nic().try_collect().await;
```

I'd actually have to sit down and give it a go though.

Issue #19, opened 2021-05-18
Handle an inbound Portal WHOAREYOU


carver:
## Current state

We try to ping a portal node now, but when we get a `whoareyou`, we... maybe try to send the same message back?

```rust
impl TalkReqHandler for ProtocolHandler {
    fn talkreq_response(&self, request: TalkRequest) {
        // Send the same request back to the sender!?
        if self.protocol_sender.send(request).is_err() {
            warn!("sending on disconnected channel");
        }
    }
}
```

## Desired state

Send back a handshake packet: https://github.com/ethereum/devp2p/blob/6eddaf50298d551a83bcc242e7ce7024c6cc8590/discv5/discv5-wire.md#handshake-message-packet-flag--2

Save the session to a simple local list of sessions, for later decryption of messages.

Next up will be to resend the original message, encrypted with the new handshake parameters.

We can skip all the kbucket tracking for now, just a simple list of active sessions is sufficient for this issue.

devp2p should have plenty of reference methods for how to implement this, since it's essentially the same as the base-layer protocol. Maybe it's possible to even use some of the library methods directly?

carver:
@SamWilsn This is next, after #18 is merged. Are you up for taking a look again (now that the base layer session init is working)?

SamWilsn:
Sure, yeah!

carver:
Sweet, #18 is merged :+1:

carver:
@SamWilsn you probably want to pull from master. Now that #21 is merged, you'll see useful logs from the discv5 library with the `RUST_LOG=debug` environment var.

SamWilsn:
Sorry it's taken me a while to come around to this, but we don't actually send the same message back. The `protocol_sender`/`protocol_receiver` channel is to transition between the synchronous world of discv5 to the asynchronous world of trin.

I added a call to `process_requests` in [67f62e564...](https://github.com/quilt/trin/commit/67f62e5642d6beaff06875b4c485bde08c352196), and will take a stab at sending back a handshake message soon.

njgheorghita:
Closing since this is implemented

PR #30, opened 2021-06-02
Spawn a new task for processing events; fix ssz for Nodes/FindNodes

SamWilsn:
I'm not particularly familiar with SSZ, so I hacked together the implementations for the `Nodes` and `FindNodes` methods, but it does seem compatible with ddht now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2018"
 [dependencies]
 clap = "2.33.3"
 ctrlc = "3.1.8"
-discv5 = "0.1.0-beta.4"
+discv5 = "=0.1.0-beta.4" # Temporarily lock the version until the patch below is resolved.
 env_logger = "0.8.2"
 eth2_ssz = "0.1.2"
 eth2_ssz_derive = "0.1.0"
--- a/src/portalnet/protocol.rs
+++ b/src/portalnet/protocol.rs
@@ -1,6 +1,7 @@
 #![allow(dead_code)]
 
 use std::net::SocketAddr;
+use std::sync::Arc;
 
 use super::{
     discovery::{Config as DiscoveryConfig, Discovery},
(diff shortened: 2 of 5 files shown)


src/portalnet/protocol.rs
```diff
@@ -32,11 +33,74 @@ impl Default for PortalnetConfig {
...
+                debug!("Got overlay ping request {:?}", request);
+                let enr_seq = self.discovery.local_enr().seq();
+                Response::Pong(Pong {
+                    enr_seq: enr_seq,
+                    data_radius: self.data_radius,
+                })
+            }
+
+            Request::FindNodes(FindNodes { distances }) => {
+                let enrs = self.discovery.find_nodes_response(distances);
```
carver, line 88: I think we can't simply reuse the discv5 find_nodes machinery here, because that will be populated by *all* nodes on the network (not necessarily only portal nodes). We only want to return nodes that are definitely in the portal network. I think we'll have to have a whole other copy of the buckets that are populated only by the portal nodes.

That seems like a big enough thing to go into its own PR, though.

src/portalnet/discovery.rs
```diff
@@ -112,8 +112,9 @@ impl Discovery {
     }
 
     /// Returns closest nodes according to given distances.
-    pub fn find_nodes_response(&self, distances: Vec<u64>) -> Vec<Enr> {
-        self.discv5.nodes_by_distance(distances)
+    pub fn find_nodes_response(&self, distances: Vec<u16>) -> Vec<Enr> {
+        let distances64: Vec<u64> = distances.iter().map(|x| (*x).into()).collect();
+        self.discv5.nodes_by_distance(distances64)
```
carver, line 117: If we aren't reusing this for the portal layer (which I think we don't, from the previous comment), then I think this can stay 64 bit.

carver (approved):
Yeah, unfortunately, I don't know SSZ very well either. I don't see anything superficially wrong. :man_shrugging: 

This seems good enough to ship, mostly commentary/suggestions for next PR. (Unless you want to remove `data_radius` from `PortalnetProtocol` now?)

carver:
Fixes #19

Cargo.toml
```diff
@@ -9,7 +9,7 @@ edition = "2018"
 [dependencies]
 clap = "2.33.3"
 ctrlc = "3.1.8"
-discv5 = "0.1.0-beta.4"
+discv5 = "=0.1.0-beta.4" # Temporarily lock the version until the patch below is resolved.
```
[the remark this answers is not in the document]
SamWilsn: Sure thing!

src/portalnet/protocol.rs
SamWilsn: Yeah, I think that's the right approach. I was just copying what was there before my PR.

src/portalnet/protocol.rs
```diff
@@ -32,11 +33,74 @@ impl Default for PortalnetConfig {
...
 
+impl PortalnetEvents {
+    /// Receives a request from the talkreq handler and sends a response back
+    pub async fn process_requests(mut self) {
+        while let Some(request) = self.protocol_receiver.recv().await {
+            debug!("Got talkreq message {:?}", request);
+            let reply = match self.process_one_request(&request).await {
+                Ok(r) => Message::Response(r).to_bytes(),
+                Err(e) => {
+                    error!("failed to process portal event: {}", e);
```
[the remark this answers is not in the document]
SamWilsn: The implementation currently sends the error text over the wire, so I guess we can fix this at the same time as that :rofl:

src/portalnet/protocol.rs
```diff
@@ -32,11 +33,75 @@ impl Default for PortalnetConfig {
 pub const PROTOCOL: &str = "portal";
 
 pub struct PortalnetProtocol {
-    discovery: Discovery,
+    discovery: Arc<Discovery>,
     data_radius: U256,
+}
+
+pub struct PortalnetEvents {
+    data_radius: U256,
```
carver, line 41: I still think `data_radius` can be removed from `PortalnetProtocol` (hm, looking back, the comment might be lost because it was made on a now-deleted commit). Fine to put that cleanup in the next PR...

carver (approved):

PR #35, opened 2021-06-30
Add private key cli arg

njgheorghita:
Add a cli arg for custom private keys.


src/portalnet/discovery.rs
```diff
@@ -35,7 +37,12 @@ pub struct Discovery {
 
 impl Discovery {
     pub fn new(config: Config) -> Result<Self, String> {
-        let enr_key = CombinedKey::generate_secp256k1();
+        let enr_key = match config.private_key {
+            Some(val) => {
+                CombinedKey::secp256k1_from_bytes(val.clone().as_mut_slice()).unwrap()
```
njgheorghita, line 42: This pr took way longer than it should have, but most of that time was wasted b/c I was using `b'0' * 32` as my test private key, but for whatever reason this library rejects the 0 bytestring as an invalid private key. Is this just a security thing? Is this desired behavior? `b'0' * 31 + b'1'` is accepted, along with all other private keys i've used

src/cli.rs
```diff
@@ -78,9 +79,15 @@ impl TrinConfig {
...
+                    .long("external-address")
+                    .help("The public IP address and port under which this node is accessible")
+                    .takes_value(true),
+            )
+            .arg(
+                Arg::with_name("private_key")
+                    .long("private-key")
+                    .help("Hex encoded 32 byte private key")
+                    .takes_value(true),
             )
```
SamWilsn, line 91: Private keys should not be taken as command line arguments, since they end up in terminal history and in the `/proc` file system.

Probably should pass a path, and read from there?

SamWilsn (changes requested):

src/cli.rs
njgheorghita: Yup, this makes sense and I def agree. Any objections to me just creating an issue to track the update for now? And consider the update "critical" before any kind of initial release. This pr is mostly to help with simplicity of spinning up a testnet and all the various nodes - which becomes more complex if we require keyfiles. All of the private keys I'm using are not even real (`00.01`, `00.02`, etc).

SamWilsn: Maybe rename the argument to `unsafe_private_key`?

In any case, my review is nonbinding, I'm only a contributor :rofl:

njgheorghita: Hahaha #allcontributorsmatter. I updated the `long` param to `unsafe-private-key`. left the name `private_key` as is, since that's internal to the code. Tracking the update here: https://github.com/carver/trin/issues/36

src/portalnet/discovery.rs
carver: Yup, I don't remember the details of why, but 0 is an invalid key. Rejecting that key is the correct behavior AIUI.

README.md
```diff
@@ -83,11 +83,16 @@ FLAGS:
...
-        --web3-http-port <web3_http_port>    port to accept json-rpc http connections [default: 8545]
-        --web3-ipc-path <web3_ipc_path>      path to json-rpc endpoint over IPC [default: /tmp/trin-jsonrpc.ipc]
-        --web3-transport <web3_transport>    select transport protocol to serve json-rpc endpoint [default: ipc]
-                                             [possible values: http, ipc]
+        --bootnodes <bootnodes>               One or more comma-delimited base64-encoded ENR's or multiaddr strings of
+                                              peers to initially add to the local routing table [default: ]
+        --discovery-port <discovery_port>     The UDP port to listen on. [default: 9000]
+        --external-address <external_addr>    The public IP address and port under which this node is accessible
+        --pool-size <pool_size>               max size of threadpool [default: 2]
+        --private-key <private_key>           Hex encoded 32 byte private key
```
carver, line 91: New name now?

src/cli.rs
```diff
@@ -78,9 +79,15 @@ impl TrinConfig {
...
-                .help("The public IP address and port under which this node is accessible")
-                .takes_value(true),
+                    .long("external-address")
+                    .help("The public IP address and port under which this node is accessible")
+                    .takes_value(true),
+            )
+            .arg(
+                Arg::with_name("private_key")
+                    .long("unsafe-private-key")
+                    .help("Hex encoded 32 byte private key (unsafe)")
```
carver, line 89: Probably explain the "unsafe" bit in the description for the arg.

src/cli.rs
```diff
@@ -96,6 +103,19 @@ impl TrinConfig {
         } else {
             None
         };
+        let private_key = if matches.is_present("private_key") {
+            let hex_private_key = value_t!(matches.value_of("private_key"), String)?;
+            match hex_private_key.len() {
+                64 => (),
+                val => panic!(
```
carver, line 110: Maybe could return error instead of panic?

carver (approved):

src/cli.rs
njgheorghita: From what I [read](https://doc.rust-lang.org/book/ch09-00-error-handling.html) - `panic!` is for unrecoverable errors and `Err` is for recoverable errors. My assumption was that if a user enters an invalid key - we want to terminate and report. Similarly, we panic with other invalid cli args. 

If I understand your suggestion correctly, you're suggesting that we want to use a recoverable error? So that trin will continue to run with an invalid key? Seems like this could be problematic, since trin'll generate a random key if the user-provided key is invalid, and then the node has a different identity from what the user is expecting.

In terms of ux, imo it seems like the `panic!` output is slightly less muddled than `Err`

`panic!`
```
thread 'main' panicked at 'Invalid private key length: 6, expected 32 byte hexstring', src/cli.rs:110:24
```

`Err`
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: "Invalid private key length: 6, expected 32 byte hexstring"', src/cli.rs:121:13
```

SamWilsn (approved):

src/cli.rs
njgheorghita: Merging for now, but i'll circle back to this in a follow up pr if we want to improve the cli-arg validation strategy

PR #40, opened 2021-07-12
Support encode/decode for FoundContent messages

njgheorghita:
Implemented encoding / decoding for `FoundContent` message. Tested locally against ddht node. Also added a couple tests for sanity.


src/portalnet/protocol.rs
```diff
@@ -148,10 +168,24 @@ impl PortalnetProtocol {
...
         };
 
+        let db_path = match env::var(TRIN_DB_ENV_VAR) {
+            Ok(val) => val,
+            Err(_) => panic!(
+                "Must supply target trin db path as environment variable, like:\n\
+                {}=\"/path\"",
+                TRIN_DB_ENV_VAR,
+            ),
+        };
```
njgheorghita, line 178: This is basically the bare minimum to get things running. Seems like this strategy should be improved - I'm open to any suggestions. If I remember correctly, the plan is to stick with rocksdb until it becomes obvious/painful enough that we need to move to a relational database.

src/portalnet/discovery.rs
```diff
@@ -112,6 +114,22 @@ impl Discovery {
         self.discv5.nodes_by_distance(distances)
     }
 
+    /// Returns list of nodes (max 32) closer to content than self.
+    pub fn find_nodes_close_to_content(&self, content_key: Vec<u8>) -> Vec<SszEnr> {
+        let self_node_id = self.local_enr().node_id();
+        let self_distance = xor_two_values(&content_key, &self_node_id.raw().to_vec());
+        // TODO: sort these by distance to return the closest 32 nodes
```
carver, line 121: A "good first issue"?

src/portalnet/protocol.rs
carver: I mean, we're already putting stuff in `~/.trin`, last I remember. Maybe just keep it there as a default for a while, until we think of something better.

carver (approved):

PR #41, opened 2021-07-14
Add message channel to serve portal jsonrpc requests

njgheorghita:
Added a [message channel](https://tokio.rs/tokio/tutorial/channels) to serve portal protocol requests over jsonrpc. My first time implementing something like this in rust, so hopefully it's not too clunky.

Added support for a couple basic admin endpoints `discv5_nodeInfo` and `discv5_routingTableInfo`

### DIFF
--- a/src/jsonrpc.rs
+++ b/src/jsonrpc.rs
@@ -1,8 +1,8 @@
 use crate::cli::TrinConfig;
-use crate::portalnet::protocol::PortalEndpoint;
+use crate::portalnet::protocol::{PortalEndpoint, PortalEndpointKind};
 use reqwest::blocking as reqwest;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde_json::{json, Value};
 use std::fs;
 use std::io::{self, Read, Write};
 use std::net::{TcpListener, TcpStream};
@@ -10,8 +10,8 @@ use std::os::unix;
 use std::sync::Mutex;
 use std::{panic, process};
 use threadpool::ThreadPool;
-use tokio::sync::mpsc::Sender;
-use tokio::sync::oneshot;
+use tokio::sync::mpsc;
+use tokio::sync::mpsc::UnboundedSender;
 use validator::{Validate, ValidationError};
 
 #[derive(Debug, Deserialize, Serialize, Validate)]
@@ -36,7 +36,7 @@ lazy_static! {
 pub fn launch_trin(
     trin_config: TrinConfig,
     infura_project_id: String,
-    portal_tx: Sender<PortalEndpoint>,
+    portal_tx: UnboundedSender<PortalEndpoint>,
 ) {
     let pool = ThreadPool::new(trin_config.pool_size as usize);
 
@@ -76,7 +76,7 @@ fn launch_ipc_client(
     pool: ThreadPool,
     infura_project_id: String,
     ipc_path: &str,
-    portal_tx: Sender<PortalEndpoint>,
+    portal_tx: UnboundedSender<PortalEndpoint>,
 ) {
     let listener_result = unix::net::UnixListener::bind(ipc_path);
     let listener = match listener_result {
@@ -107,7 +107,7 @@ fn launch_http_client(
     pool: ThreadPool,
     infura_project_id: String,
     trin_config: TrinConfig,
-    portal_tx: Sender<PortalEndpoint>,
+    portal_tx: UnboundedSender<PortalEndpoint>,
 ) {
     let uri = format!("127.0.0.1:{}", trin_config.web3_http_port);
     let listener = TcpListener::bind(uri).unwrap();
@@ -132,7 +132,7 @@ fn serve_ipc_client(
     rx: &mut impl Read,
     tx: &mut impl Write,
     infura_url: &str,
-    portal_tx: Sender<PortalEndpoint>,
+    portal_tx: UnboundedSender<PortalEndpoint>,
 ) {
     let deser = serde_json::Deserializer::from_reader(rx);
     for obj in deser.into_iter::<JsonRequest>() {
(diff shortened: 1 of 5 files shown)


njgheorghita:
ping @SamWilsn for :eyes: if you have any input

src/jsonrpc.rs
```diff
@@ -156,26 +190,69 @@ fn process_http_request(obj: JsonRequest, infura_url: &str) -> Vec<u8> {
...
+            "web3_clientVersion" => Ok(format!(
+                r#"{{"jsonrpc":"2.0","id":{},"result":"trin 0.0.1-alpha"}}"#,
+                obj.id,
+            )),
+            _ => dispatch_infura_request(obj, infura_url),
+        }
+    }
+}
+
+fn dispatch_infura_request(obj: JsonRequest, infura_url: &str) -> Result<String, String> {
```
carver, line 212: 👍🏻 for pulling this into a separate method.

src/portalnet/protocol.rs
```diff
@@ -11,10 +11,19 @@ use super::{
...
 
 use super::socket;
 
+type Responder<T, E> = oneshot::Sender<Result<T, E>>;
+
+#[derive(Debug)]
+pub enum PortalEndpoint {
+    NodeInfo { resp: Responder<Value, String> },
+    RoutingTableInfo { resp: Responder<Value, String> },
+}
```
carver, line 25: Something about this enum smells funny to me, but I can't put my finger on it. 🤷🏻‍♂️  carry on

carver (approved):
LGTM

src/jsonrpc.rs
```diff
@@ -156,26 +190,69 @@ fn process_http_request(obj: JsonRequest, infura_url: &str) -> Vec<u8> {
...
+        Err(err) => Err(format!(
+            r#"{{"jsonrpc":"2.0","id":"{}","error":"Infura failure: {}"}}"#,
+            obj.id,
+            err.to_string(),
         )),
+    }
+}
+
+#[tokio::main]
+async fn dispatch_portal_request(
```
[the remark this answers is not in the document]
SamWilsn: You can put `tokio::main` at any boundary that starts an async context, but yeah, I think it would probably be better to hoist this as far up as possible.

src/jsonrpc.rs
```diff
@@ -156,26 +190,69 @@ fn process_http_request(obj: JsonRequest, infura_url: &str) -> Vec<u8> {
...
-                )),
-            }
+            return Err(format!(
+                r#"{{"jsonrpc":"2.0","id":"{}","error":"Unsupported discv5 endpoint: {}"}}"#,
+                obj.id, method,
+            ))
         }
-    }
+    };
+    tokio::task::spawn(async move { portal_tx.send(message).await });
```
SamWilsn, line 242: Alternatively, you could use an [`UnboundedSender`](https://docs.rs/tokio/1.8.1/tokio/sync/mpsc/struct.UnboundedSender.html), which doesn't require an `await` to send.

src/jsonrpc.rs
```diff
@@ -156,26 +190,69 @@ fn process_http_request(obj: JsonRequest, infura_url: &str) -> Vec<u8> {
...
+    let res = match resp_rx.await.unwrap() {
+        Ok(val) => val,
+        Err(msg) => {
+            return Err(format!(
+                r#"{{"jsonrpc":"2.0","id":"{}","error":"Error while processing {}: {}"}}"#,
```
SamWilsn, line 247: Since we have `serde_json`, would the [`json!`](https://docs.rs/serde_json/1.0.64/serde_json/macro.json.html) macro be more elegant?

src/portalnet/protocol.rs
SamWilsn: It could be factored a bit:

```rust
pub enum PortalEndpointKind {
    NodeInfo,
    RoutingTableInfo,
}

pub struct PortalEndpoint {
    kind: PortalEndpointKind,
    resp: Responder<Value, String>,
}
```

src/portalnet/protocol.rs
```diff
@@ -47,9 +57,39 @@ pub struct PortalnetEvents {
     protocol_receiver: mpsc::Receiver<Discv5Event>,
 }
 
+pub struct JsonRPCHandler {
```
SamWilsn, line 60:
```suggestion
pub struct JsonRpcHandler {
```

src/jsonrpc.rs
njgheorghita: Definitely! Wasn't aware of that - updated the other responses also

src/jsonrpc.rs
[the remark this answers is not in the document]
njgheorghita: > ... but yeah, I think it would probably be better to hoist this as far up as possible.

Just to confirm my understanding, you're suggesting making all the functions that are called b/w `main()` and here `async`? Any suggestion on which approach is better b/w this or using an unbounded channel?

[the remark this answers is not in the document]
SamWilsn: To quote from the [tokio docs](https://docs.rs/tokio/1.8.1/tokio/attr.main.html):

> [...] starting a new runtime each time it is called. If the function is called often, it is preferable to create the runtime using the runtime builder so the runtime can be reused across calls.

Using the `tokio::main` attribute fires up all the machinery for tokio every time the annotated function is called, so I think we probably should move it up to the top `fn main()`. On the other hand, we shouldn't just convert every non-async function all the way up to the entry point into an async one, especially if they contain I/O calls.

I think, for now, the best approach would be to use an `UnboundedSender` to bridge the sync/async worlds.

Thoughts @carver?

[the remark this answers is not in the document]
njgheorghita: Sweet! Thanks for the clarification - I implemented the `UnboundedSender` and got rid of the excessive `tokio::main`. I'd always welcome a final :eyes: from you or @carver to double check, but everything is functioning as expected locally so i'll go ahead with the merge unless i hear otherwise by eod.

SamWilsn (approved):

PR #43, opened 2021-07-17
Sort closest nodes by distance

ogenev:
Hello, I'm part of the apprenticeship program and new to Rust. Happy to learn and get feedback from you guys.

This PR sorts nodes by distance.

Fixes https://github.com/carver/trin/issues/42.

### DIFF
--- a/src/portalnet/discovery.rs
+++ b/src/portalnet/discovery.rs
@@ -119,20 +119,33 @@ impl Discovery {
         self.discv5.nodes_by_distance(distances)
     }
 
-    /// Returns list of nodes (max 32) closer to content than self.
+    /// Returns list of nodes (max 32) closer to content than self, sorted by distance.
     pub fn find_nodes_close_to_content(&self, content_key: Vec<u8>) -> Vec<SszEnr> {
         let self_node_id = self.local_enr().node_id();
         let self_distance = xor_two_values(&content_key, &self_node_id.raw().to_vec());
-        // TODO: sort these by distance to return the closest 32 nodes
-        self.discv5
+
+        let mut nodes_with_distance: Vec<(Vec<u8>, Enr)> = self
+            .discv5
             .table_entries_enr()
             .into_iter()
-            .filter(|enr| {
-                xor_two_values(&content_key, &enr.node_id().raw().to_vec()) < self_distance
+            .map(|enr| {
+                (
+                    xor_two_values(&content_key, &enr.node_id().raw().to_vec()),
+                    enr,
+                )
             })
+            .collect();
+
+        nodes_with_distance.sort_by(|a, b| a.0.cmp(&b.0));
+
+        let closest_nodes = nodes_with_distance
+            .into_iter()
             .take(32)
-            .map(|enr| SszEnr::new(enr))
-            .collect()
+            .filter(|node_record| &node_record.0 < &self_distance)
+            .map(|node_record| SszEnr::new(node_record.1))
+            .collect();
+
+        closest_nodes
     }
 
     pub async fn send_talkreq(


src/portalnet/discovery.rs
```diff
@@ -119,17 +119,27 @@ impl Discovery {
...
             })
+            .collect();
+
+        closest_nodes.sort_by(|a, b| {
+            xor_two_values(&content_key, &a.node_id().raw().to_vec())
+                .cmp(&xor_two_values(&content_key, &b.node_id().raw().to_vec()))
+        });
+
+        closest_nodes
+            .into_iter()
```
njgheorghita, line 142: Hey @ogenev ! I'm also pretty new with rust, and unfamiliar with the "best" way to do things. However, in this case it seems to me like we're running the `xor_two_values()` computation too much. It might not be that expensive, but I'd suggest trying to rework this so that we only run `xor_two_values()` once. Maybe in the initial `closest_nodes` collection, we don't do any filtering and simply store each `Enr` along with its xor distance from the content key in some kind of data structure (like a `HashMap`? maybe there's a better option?). Then in the next step we can simply sort by the already-calculated distance, take 32, and cut any distances that are `< self_distance`. 

It seems this might be cheaper than the current setup, but lmk if this makes sense or if I'm missing something

ogenev: Hi, thanks for the feedback, this makes perfect sense to me. I updated the code as suggested. 
I had some troubles storing `Enr` as key in the hashmap and a workaround was to convert it to base_64 string.

Let me know if it makes more sense now.

njgheorghita: Thanks for the update! Makes sense about the enr / hashmap conflict. It seems like the enr type doesn't implement hash, and implementation isn't completely obvious to me. Just two notes... 

1. Do you mind taking another stab at this, but with a different datatype other than `HashMap`? The enr conversion / conversion back are both steps we'd want to avoid if possible - since it's just more calculation / code complexity. Maybe try `HashMap<Vec<u8>, Vec<Enr>>`? If i understand hashmap correctly, it's only the key we worry about hashing. So we could avoid this by using the distance as the key, and have the distance point towards a vec of enrs that share that distance (since it seems possible though unlikely that there'll be more than one enr that shares the exact same distance? but i could be wrong on this.). Another datatype you could try is just a [tuple of tuples](https://doc.rust-lang.org/rust-by-example/primitives/tuples.html) with each interior tuple containing `(ENR, distance)` - and then just iterate through that...
2. on line 142/143 - I'm pretty sure we want to filter by distance before we take 32... But I guess, since we're now sorting by distance on line 138, then it guarantees that no record exists outside of the first 32 that will be closer to the target than any of the first 32 records. So, actually I think this is just fine as is

SamWilsn: Before going too much further with this one, I feel like this particular comment needs to be benchmarked to actually see what the most performant implementation is.

Using a map adds a non-trivial amount of overhead, and I can totally imagine that modern processors xor linear memory at blazing speeds. It may actually be the case that repeating the xor at every step is faster than using a map, and it might be true the other way around.

If this is a performance critical bit of code we should absolutely measure, and if the performance here doesn't matter that much, we should go with the most readable implementation.

carver: > If this is a performance critical bit of code we should absolutely measure, and if the performance here doesn't matter that much, we should go with the most readable implementation.

I'd say: this isn't an architectural decision that is hard to reverse later. If it turns out to be a hot spot, it's a few lines of code we can optimize. So 👍🏻 for the most readable implementation, without having to do any benchmarking.

On readability, I found `HashMap` confusing, because the code never uses the lookup property, and it makes me wonder if I've missed something. So yeah, a vector of tuples of `(distance, ENR)` would make plenty of sense to me.

ogenev: Hey, thanks guys, I agree that vector of tuples is more readable and that we don't need to do any benchmarks for now.

_"Premature optimization is the root of all evil"_ :)

PR #49, opened 2021-07-23
Add some of the RPC methods to the readme

acolytec3:
Lists several of the RPC methods exposed by trin (especially the `discv5_nodeInfo` one that gives the all important ENR!!!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,10 +104,11 @@ OPTIONS:
                                               [possible values: http, ipc]
 ```
 
-## RPC Methods
+## Custom RPC Methods
 - `discv5_nodeInfo`     Returns the ENR of the client
-- `web3_clientVersion`  Returns the current version of Trin being run
-- `eth_blockNumber `    Returns the current block number from the tip fo the chain (as provided by Infura at present)
+- `discv5_routingTableInfo`     Returns the list of discovery peers that have recently been available
+
+See https://eth.wiki/json-rpc/API#json-rpc-methods for other standard methods that are implemented. Most of the currently proxy to Infura.
 
 ## Gotchas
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +104,11 @@ OPTIONS:
                                               [possible values: http, ipc]
 ```
 
+## RPC Methods
+- `discv5_nodeInfo`     Returns the ENR of the client
+- `web3_clientVersion`  Returns the current version of Trin being run
+- `eth_blockNumber `    Returns the current block number from the tip fo the chain (as provided by Infura at present)
```
carver, line 110: Hm, this is going to turn into a very long list, if it becomes comprehensive. It's maybe fine to just have a short selection of these, but maybe it's worth also linking to the json-rpc wiki, and only listing things that *aren't* in the wiki.

```suggestion
## Custom RPC Methods
- `discv5_nodeInfo`     Returns the ENR of the client
- `discv5_routingTableInfo`     Returns the list of discovery peers that have recently been available

See https://eth.wiki/json-rpc/API#json-rpc-methods for other standard methods that are implemented. Most of the currently proxy to Infura.
```

carver (approved):
Yay for docs! Let me know if you want to make any other changes. Otherwise, I could just apply my suggestion and merge.

carver:
Doh, free plan isn't going to cut it I guess. I'm sure this didn't break the windows build, though...

Oh, also, you would probably like to know that `RUST_LOG=debug cargo ...` will show you a lot more stuff that's going on. That might be a good thing to include in the readme too. :)

acolytec3:
That looks good. I just needed a starting point to find the enr and there
wasn't anything. And deciphering rust took me a minute. 😅

PR #52, opened 2021-07-27
Add default data directory

KolbyML:
This PR just adds default data directories that are used if ``TRIN_DATA_PATH`` (the replacement for ``TRIN_DB_PATH``) isn't set.

```
// Windows: C:\Users\Username\AppData\Roaming\Trin
// macOS: ~/Library/Application Support/Trin
// Unix-like: ~/.trin
```
These are the default data directories for each respected operating system. Just the standard locations with each operating system and normal naming scheme (caps or no caps etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ cd trin
 TRIN_INFURA_PROJECT_ID="YoUr-Id-HeRe" cargo run
 ```
 
+**Optional:** Custom data directory
+```shell
+TRIN_DATA_PATH="/path"
+```
+
 ### Connect over IPC
 In a python shell:
 ```py
--- a/src/portalnet/protocol.rs
+++ b/src/portalnet/protocol.rs
@@ -1,12 +1,12 @@
 #![allow(dead_code)]
 
-use std::env;
 use std::net::SocketAddr;
 use std::sync::Arc;
 
 use super::{
     discovery::{Config as DiscoveryConfig, Discovery},
     types::{FindContent, FindNodes, FoundContent, Nodes, Ping, Pong, Request, Response, SszEnr},
+    utils::get_data_dir,
     U256,
 };
 use super::{types::Message, Enr};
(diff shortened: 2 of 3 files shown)


carver (approved):
Looks good to me. :)

Could you also add an issue to use the following spec for the default directory, when applicable? 
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
(or if you'd rather just update this PR, obviously, that's cool too)

SamWilsn:
> Looks good to me. :)
> 
> Could you also add an issue to use the following spec for the default directory, when applicable?
> https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
> (or if you'd rather just ~update this~ _add a quick follow up_ PR, obviously, that's cool too)

https://crates.io/crates/directories is probably what you want!

KolbyML:
> Looks good to me. :)
> 
> Could you also add an issue to use the following spec for the default directory, when applicable? 
> https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
> (or if you'd rather just ~~update this~~ *add a quick follow up* PR, obviously, that's cool too)

To my knowledge every Unix-Like operating system supports $HOME, but only some support $XDG* environment variables. A notable one that doesn't support XDG is Ubuntu. Also I know we are following the same file path as thunderbird, Skype, xchat, bitcoin, just to list a few. .config is normally used if it is just a configuration file to my knowledge.

So I think it is fine :), if someone else wants to make a PR for this be my guest, XDG is just not as univeral as HOME.

SamWilsn:
Screenshot from 2021-07-27 14-23-26

Ubuntu most certainly supports some subset of the XDG spec!

The `directories` crate will follow XDG (including fallback) on linux, and the recommended paths on windows and mac, so it mostly takes care of everything.

KolbyML:
> Screenshot from 2021-07-27 14-23-26
> 
> Ubuntu most certainly supports some subset of the XDG spec!
> 
> The `directories` crate will follow XDG (including fallback) on linux, and the recommended paths on windows and mac, so it mostly takes care of everything.

Hmm interesting I guess they are just not defined by default or something, I wonder if like kde or some other gui defined them for you

PR #53, opened 2021-07-27
Update get_data_dir() to use directories

KolbyML:
I updated get_data_dir() to use directories instead of env with a ton of insight from Sam


src/portalnet/utils.rs
```diff
@@ -41,26 +42,47 @@ mod test {
...
     match env::var(TRIN_DATA_ENV_VAR) {
-        Ok(data_path) => data_path,
-        Err(_) => get_default_data_dir(),
+        Ok(data_path) => {
+            path = data_path;
+        }
+        Err(_) => {
+            path = get_default_data_dir();
+        }
+    }
```
SamWilsn, line 53: It would be more rusty (rustic?) to do something like:

```
let path = env::var(TRIN_DATA_ENV_VAR).unwrap_or_else(get_default_data_dir);
```

src/portalnet/utils.rs
```diff
@@ -41,26 +42,47 @@ mod test {
...
+
+    if !Path::new(&path).exists() {
+        match fs::create_dir_all(&path) {
+            Ok(_) => {}
+            Err(_) => {
+                panic!("unable to create directory, possibly from lack of permissions");
+            }
+        }
```
SamWilsn, line 61: I don't think `create_dir_all` fails if the directories already exist, and using `expect` will preserve the error information:

```suggestion
    fs::create_dir_all(&path).expect("unable to create data directory")
```

src/portalnet/utils.rs
KolbyML: Damn I didn't even realise I could do that xD

src/portalnet/utils.rs
KolbyML: My bad, I think I got confused when reading the documentation for it.

carver:
Argh, I hate merging something "red", but I also don't really feel like dealing with devops to get you all access after this repo is moved to the ethereum.org github group.

carver:
Fixes #55

PR #54, opened 2021-07-27
Create workspace for multiple networks

jacobkaufmann:
- Create individual packages for the state and history networks.
- Move the main() function from main.rs to lib.rs for the state network.
- Remove main.rs for the state nework.
- Add a basic main() function in the history network lib.rs.
- Create a 'trin' package to host a single entry point for the program.


carver:
Hey, moving from side-channel to the PR:

> I could not remember whether we wanted to maintain the ability to run the binaries individually or always spin them all up at once. The latter is what is implemented.

It's fine to launch them combined. We can add separate launches later, if we want.

Though it does become extra important for the version of chain-history on master to not crash on launch, so it doesn't halt development on state network. (that seems like a good invariant to target anyway)

> I was also not sure how command line args are handled with this approach, once you CLI args for history

Yeah, similar to full clients, there will probably just be a ton of arguments that need to be named in a disambiguated way. The more we can just use good defaults, the better.

---

Feel fee to mention me when it's ready for final review.

SamWilsn:
> I was also not sure how command line args are handled with this approach, once you CLI args for history

If you use `structopt` (which I think we are now), you can define the argument structs in each library crate and merge them together in the binary crate, which might be a neat way to do it.

trin/src/main.rs
```diff
@@ -0,0 +1,11 @@
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    tokio::select! {
+        history = trin_history::main() => {
+            history
+        },
+        state = trin_state::main() => {
+            state
+        },
+    }
```
jacobkaufmann, line 10: Is there a cleaner way to do this? Do we want any logging here, or should we leave that to the individual network `main()` functions?

jacobkaufmann:
It looks like the `win-build` CI test has been failing ever since the [merge](https://app.circleci.com/pipelines/github/carver/trin/170/workflows/2d9347dd-c960-4b57-ab8b-26faf432d0f6) of PR #49. Perhaps we want to create an issue to figure out why.

A number of additional items we could handle in this PR or as separate tasks:

* Reorganize `structopt` to differentiate arguments for state vs history
  * The history network has no CLI arguments as of now, but we could go ahead and create a structure in the binary crate to wrap the state network's arguments.
* Move shareable `jsonrpc.rs` and `portalnet` code to a separate library crate to be shared among state and history.

KolbyML:
> It looks like the `win-build` CI test has been failing ever since the [merge](https://app.circleci.com/pipelines/github/carver/trin/170/workflows/2d9347dd-c960-4b57-ab8b-26faf432d0f6) of PR #49. Perhaps we want to create an issue to figure out why.

@jacobkaufmann I think it is because I used a ton of credits when I was originally making the CI test for windows then after I realized that I started using the credits on my circleci account, so it should be fine whenever it resets.

carver:
> A number of additional items we could handle in this PR or as separate tasks:
> 
>    * Reorganize `structopt` to differentiate arguments for state vs history
>       
>       * The history network has no CLI arguments as of now, but we could go ahead and create a structure in the binary crate to wrap the state network's arguments.
> 
>    * Move shareable `jsonrpc.rs` and `portalnet` code to a separate library crate to be shared among state and history.

I'm all for adding issues to tackle these both after #54 is merged. I would only consider letting them delay merging this PR if it either was necessary to unblock work on chain history.

README.md
```diff
@@ -12,6 +12,16 @@ This should sound similar to a light client. It is, but with a peer-to-peer
 philosophy rather than the LES client/server model, which has introduced
 challenges in an altruistic environment.
 
+## Repository Structure
+
+Trin is a collection of networks.
```
carver, line 17: Maybe: 
```suggestion
The portal protocol is a collection of networks. Trin needs to connect to all of them.
```

trin/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "trin"
+version = "0.1.0"
+authors = ["Jacob Kaufmann <jacobkaufmann18@gmail.com>"]
```
carver, line 4: Maybe add both of us for the joint one? At some point, this will probably turn into a joint member list, like `trin@ethereum.org`. But for now, we can just list individual maintainers of the individual crates.

trin/src/main.rs
carver: Yeah, fine to leave it bland for now. Seems okay to me.

At some point we can play with whether we want them in separate threads/runtimes, but for now a 👍🏻 from me.

carver (approved):
Make sure the Cargo.toml change in #53 makes it in here, then should be good to go.

PR #56, opened 2021-08-01
Allows trin packages to be run individually

mrferris:
This update adds binaries to the trin-state and trin-history packages so that we can run each individually (in case one is breaking, etc.) by running `cargo run -p trin-state` or `cargo run -p trin-history`. 

Note that the behavior when we run just `cargo run` is untouched: it still defaults to just running the `trin` binary.


ogenev (approved):
LGTM 👍

Cargo.toml
```diff
@@ -4,3 +4,5 @@ members = [
     "trin-state",
     "trin-history",
 ]
+
+default-members = ["trin"]
```
jacobkaufmann, line 8: Does this change the behavior of any Cargo commands?

trin-history/src/main.rs
```diff
@@ -0,0 +1,6 @@
+use trin_history;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    return trin_history::main().await;
```
jacobkaufmann, line 5: Nit: Maybe the following is more idiomatic? Similarly for state.

```suggestion
    trin_history::main().await
```

jacobkaufmann (commented):
Left some minor comments. Once addressed I have no problem approving. Code suggestion not necessary for approval.

Cargo.toml
mrferris: If we don't have this, running `cargo run` won't default to running `trin/src/main.rs`, it will instead say: 
```error: cargo run could not determine which binary to run. Use the --bin option to specify a binary, or the default-run manifest key. available binaries: trin, trin-history, trin-state```.

I don't think that running `trin` by itself will be very useful for a while, so maybe it wouldn't be bad to not have this and force a choice for now. 

Also, this makes it so that we have to run `cargo run -p trin-state` instead of `cargo run --bin trin-state` the latter of which seems way more common in Rust.

So maybe we should get rid of this. I'll keep it in for now and we'll see if we hate it.

trin-history/src/main.rs
mrferris: thanks for the nit, added a commit to make those changes

PR #57, opened 2021-08-02
Set trin main package as root package

ogenev:
Trin main package seems to be the entry point for different portal network packages and depends on them. I think it makes more sense to set the main package as a root package.  I'm not sure if there are any benefits to use virtual manifest in root `cargo.toml` and duplicate `trin` folder in the root path. This is typically useful when there isn't a "primary" package.

We are also renaming package folders from `trin-state` and `trin-history` to just `state` and `history` because the main crate is `trin` and there is some redundancy in folder names.

Let me know what do you think, I'm using this PR as a proposition, instead of opening a separate issue.


ogenev:
Because I'm starting to work on implementing CLI parsing support for both networks, I'm thinking of what is the best way to structure the common code.

As we are moving forward to workspace architecture and because circular dependencies in workspaces are not possible (we can't import root module in the current workspaces, because the root depends on them), If we want more than one workspace to use the same module, it seems natural to just have a third `common` workspace, and both `state` and `history` to depend on it.

Let me know if this makes sense and we can add the third workspace to this PR.

mrferris:
I agree with both points:

- trin should be a root package
- the directory naming is redundant

Could you elaborate on the problem with cyclical module imports that you mentioned this solving too?

ogenev:
> 
> Could you elaborate on the problem with cyclical module imports that you mentioned on zoom this solving too?

This PR does not solve cyclical dependencies, it is just moving `trin` package as a root package.

As for the cyclical dependencies, If you want to import a module from the root package in `trin-state`, you have to set a dependency in `trin-state` as:

```rust
[dependencies]
trin = { path = ".." }
```

But the root package depends on `trin-state`:

```rust
[dependencies]
trin-history = { path = "history" }
trin-state = { path = "state" }
```
And you get a cyclic package dependency:error:

```rust
error: cyclic package dependency: package `trin v0.1.0 (/home/ogi/dev/eth/trin)` depends on itself. Cycle:
package `trin v0.1.0 (/home/ogi/dev/eth/trin)`
    ... which is depended on by `trin-state v0.1.0 (/home/ogi/dev/eth/trin/state)`
    ... which is depended on by `trin v0.1.0 (/home/ogi/dev/eth/trin)`
```

SamWilsn:
Repeating the name of the root package (ie. `trin` containing `trin-state`) is a fairly standard practice, for example [tokio](https://github.com/tokio-rs/tokio) does it. I think it's because normally the directory name matches the crate name, and the crate name would be `trin-history`.

ogenev:
> Repeating the name of the root package (ie. `trin` containing `trin-state`) is a fairly standard practice, for example [tokio](https://github.com/tokio-rs/tokio) does it. I think it's because normally the directory name matches the crate name, and the crate name would be `trin-history`.

Yeah, most of the rust libraries seems to use that pattern, there are some popular libraries, for example as [syn](https://github.com/dtolnay/syn) which use custom folder names. I'm happy with both ways.

carver:
> Repeating the name of the root package (ie. `trin` containing `trin-state`) is a fairly standard practice, for example [tokio](https://github.com/tokio-rs/tokio) does it. I think it's because normally the directory name matches the crate name, and the crate name would be `trin-history`.

+1 for keeping the trin- prefix

ogenev:
I'm closing this because the changes are included in https://github.com/carver/trin/pull/60.

PR #60, opened 2021-08-03
Setup trin-core package

ogenev:
Experimental PR implementing core package for common modules.

Fixes https://github.com/carver/trin/issues/58


carver (approved):
LGTM. Except it looks like the conversation on #57 is leaning toward leaving the trin- prefix in place.

PR #61, opened 2021-08-04
Bootstrap trin-history network on separate ports

ogenev:
Bootstrap `trin-history` network by changing the default ports to not match `trin-state`. It also adds method to display the config values.

This is a hack until we make StructOpt handle CLI simultaneously for both networks.

### DIFF
--- a/trin-core/src/cli.rs
+++ b/trin-core/src/cli.rs
@@ -1,4 +1,5 @@
 use crate::portalnet::types::HexData;
+use log::info;
 
 use std::env;
 use std::ffi::OsString;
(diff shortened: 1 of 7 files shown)


trin-state/src/lib.rs
```diff
@@ -8,9 +8,10 @@ use trin_core::jsonrpc::launch_trin;
 use trin_core::portalnet::protocol::{PortalEndpoint, PortalnetConfig, PortalnetProtocol};
 
 pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    tracing_subscriber::fmt::init();
```
mrferris, line 11: what's the reasoning behind removing this from lib.rs and only having it in main.rs?

mrferris:
I approve 👍

trin-state/src/lib.rs
ogenev: I guess it is usually added in an executable as explained in https://docs.rs/tracing-subscriber/0.2.19/tracing_subscriber/fmt/index.html.
If we leave it in `lib.rs` for `trin-state` and `trin-history`,  it is raising an error if we try to run both networks in parallel from the root crate, because we are trying to initialize `tracing_subscriber` more than once.

trin-core/src/cli.rs
```diff
@@ -90,37 +90,44 @@ impl TrinConfig {
...
+        Ok(config)
+    }
+
+    pub fn display_config(&self) {
+        match self.web3_transport.as_str() {
+            "http" => {
+                println!(
+                    "Protocol: {}\nWEB3 HTTP port: {}",
+                    self.web3_transport, self.web3_http_port
+                )
```
carver, line 114: I'd also support if you wanted to upgrade these to info logs. We probably don't want any `println`s by the time we get to a stable release.

carver (approved):
LGTM

PR #66, opened 2021-08-09
raw implementation of advertise & requestProofs

victor-wei126:
This is a draft implementation of advertise & requestProofs, probably not ready to be merged. There are some issues, mainly about whether the Advertise & RequestProofs structs should derive traits.

Also, the flow might not be correct. As I have it implemented now, when an `Advertise` request comes in, a `RequestProofs` (response) is immediately sent in response to request the proofs of the advertised key. This doesn't seem to align with the spec, where I believe nodes only send `RequestProofs` when they want some certain proofs, not necessarily immediately in response to `Advertise`.


ogenev:
Hey, during the last implementation call, Piper suggested OFFER, ACCEPT and STORE to supersede Advertise and RequestProofs on state network (So we will have the same messages as trin-history, for full state network client functionality please check [this doc](https://notes.ethereum.org/8LbisMwtR4yK64SBosb_2g?view)).

As this PR now overlaps with https://github.com/carver/trin/issues/64, do you want to assign this to you?

victor-wei126:
Sure, I can take over that new PR and work on implementing OFFER, ACCEPT, and STORE. So OFFER now becomes the new Advertise, and ACCEPT becomes the new RequestProofs? If both of these messages are on trin-history and trin-state, should the implementation be in trin-core.

PR #67, opened 2021-08-11
add default data path for readme for unix

victor-wei126:
was trying to connect 2 clients for testing, but if a new data path is not set, it will fallback to default, and that causes issues since 2 nodes cannot use same data path. so would be useful to have the default data path in readme for development purposes


ogenev:
Could you fix please the git commits, we expect to see only one commit in this PR?

victor-wei126:
I'm actually not sure how to get rid of those commits, they just appeared while I was merging carver/trin master branch into my fork. Not sure if this is just a git thing, but they show no changed files either, so I'm not sure what the point of them is.

victor-wei126:
searched it up and found the solution! opened a new PR so closing this now

PR #68, opened 2021-08-12
add default data path for readme for unix

victor-wei126:
fixed commit issues from previous PR

### DIFF
--- a/trin-history/src/lib.rs
+++ b/trin-history/src/lib.rs
@@ -60,10 +60,8 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .await
             .unwrap();
 
-        tokio::join!(
-            events.process_discv5_requests(),
-            rpc_handler.process_jsonrpc_requests()
-        );
+        tokio::spawn(events.process_discv5_requests());
+        tokio::spawn(rpc_handler.process_jsonrpc_requests());
 
         // hacky test: make sure we establish a session with the boot node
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -54,8 +54,12 @@ cargo run -p trin-state|trin-history
 
 **Optional:** Custom data directory
 ```shell
-TRIN_DATA_PATH="/path"
+TRIN_DATA_PATH="/your_path"
 ```
+*Note, default data paths are:*\
+Linux/Unix - `$HOME/.local/share/trin` OR `~/.trin`
```
mrferris, line 60: what does the OR refer to? When does it use the former path and when the latter?

victor-wei126: I believe it just depends on the system. I'm running ubuntu, and my data was not stored in the first dir. @KolbyML had the second dir on their previously. The docs are here: https://docs.rs/directories/3.0.2/directories/struct.ProjectDirs.html#method.data_local_dir

KolbyML: I don't think it will store files in ``~/.trin``. That is the directory I wanted to use, but other members wanted follow the XRC spec or something.

carver (approved):
So it looks like this PR is waiting on removing the `~/.trin` from both comments (since the `directories` project almost certainly doesn't put data files there).

victor-wei126:
Okay I removed the `~/.trin` data paths from both locations. I don't know how to get rid of the stupid "merge branch into" commits but if that's an issue I'll look into how to get those removed.

PR #69, opened 2021-08-12
[WIP] Storage Class for abstracting Trin storage logic

mrferris:
This is a draft PR of a work in progress of a PortalStorage struct that handles the logic of determining our node's radius at any given time, and determining if/how to store a given piece of data given our node's current radius and data capacity.

The main functions that it aims to provide to the rest of the client code are:

- `new(config: PortalStorageConfig) -> PortalStorage`: The constructor, PortalStorageConfig allows user to specify the client's max capacity in kb
- `should_store(key: String) -> bool`: used for determining whether to respond with an ACCEPT to an OFFER request
- `store(key: String, value: String) -> Result` for storing a new piece of content from the network. This function automatically takes into account the capacity and, if necessary, drops the farthest piece of data, finds the new farthest and updates the radius.
- `get(key: String) -> Result`: Gets the value for the key if it exists in our DB.
- `get_current_radius() -> u64`: Gets the current radius (currently u64, should probably be a u256? unsure).

Thoughts on this interface welcome.

Notes:

- This wraps RocksDB as the main keystore, but is also using SQLite in order to calculate the xor-farthest key on the fly (once we drop the old farthest key). In the code I refer to this SQLite DB as the "meta_db". I'm using SQLite purely based off of a tip from Piper, and I haven't fully gotten the SQLite query to do an xor with the node id, sort, and take the top one yet. And now I'm reading that SQLite doesn't support xor, but also there's [this SO post that might work](https://stackoverflow.com/questions/2294137/sqlite-exclusive-or-how-do-i-clear-a-flag) but might not. So if anyone has any other ideas on a better way to do that in the meantime that would be cool. I don't think that RocksDB can do it, hence trying another dedicated DB. 

- This also includes the `generate-data` bin, which can be run as so: `cargo run --bin generate-data -- --kb 120`, which will populate the RocksDB instance with 120kb of random key value pairs. Use `cargo run --bin generate-data -- -h` to see all options. I plan to use this to help while testing that the intricacies of the store function are working properly. I also plan to add get and delete functionality, so it will probably end up being called `trin-db-util` instead of `generate-data`.

Things to be done:

- [ ] Finish the SQLite call to search all of the stored keys for the farthest one
- [ ] Make it gracefully handle re-starts
- [ ] Finalize all `store` logic, not quite there yet
- [ ] More consistent Result / Option / Error handling
- [ ] Clean up and organize, break out db calls into smaller functions especially in store
- [ ] Make tests actually test results, instead of just making sure things don't blow up
- [ ] Fix tab formatting issues from VSCode


trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,385 @@
...
+        }
+
+        Ok(size)
+
+    }
+
+    pub fn distance_to_key(&self, key: &String) -> u64 {
+
+        let byte_vector = xor_two_values(
+            key.as_bytes(), &self.node_id.raw().to_vec()
```
victor-wei126, line 218: I'm not sure what `as_bytes()` does but, doesn't `key` need to be hashed to get the `content-id` here? Since I thought that the hash of the key was to be used in the distance fn.

mrferris: Yea this code doesn't yet account for the `key`-hash->`id` (where `id` is the actual key into the DB) thing. `key` anywhere in this currently refers to what we know as the `id`.  

Off the top of my head I think that this PortalStorage code should do that conversion itself, so that the code that uses it never has to know/care about content IDs. Just pass the key into `should_store(key)`, `store(key)`, and `get(key)`. 

`as_bytes()` is just turning the string into the `Vec<u8>` that the `xor_two_values` function expects.

victor-wei126: Are we using the `content_id` as the key into the db? It seemed to me that we are using the `content_key` as the actual index, since in protocol.rs in the main branch, that is what is being used as the index for DB queries (in the `FindContent`)

victor-wei126:
So this is just some of my thoughts on a couple of implementation details that involves this PR and PR #84.

### Where to put code for "content-id derivation function"
In the history network, to derive `content-id` you simply take the sha256 hash of the `content_key` but in the state network there is a somewhat more complicated scheme depending on the exact content type. Since the `content_id` is used in the distance function, we'll need to implement all these "derivation functions" somewhere. My two thoughts were:
1) Implement at top-level, ie, in #84, in the OFFER msg handler.
2) Implement in the `should_store(key)` method of this PR

I am thinking the second method is a cleaner way to do it, and like Mike mentioned above, it abstracts away `content_id` so the handlers of protocol messages never have to know about `content_id` and simply pass in the keys. Perhaps a new method like `compute_content_id(key)`, which takes in keys and parses them to figure out content type, and then applies the appropriate function according to the spec.

### Use `content_id` or `content_key` as the primary key index into Trin's database
I'm not sure which one Trin currently uses, but reading over the code, it looks to me as if `content_key` is used as the key to index into the database. Would `content_id` be any more efficient since it is just 256 bit integer, whereas `content_key` is more complex?

mrferris:
> I'm not sure which one Trin currently uses, but reading over the code, it looks to me as if content_key is used as the key to index into the database. Would content_id be any more efficient since it is just 256 bit integer, whereas content_key is more complex?

`content_id` will be used to index into the database. Anywhere in this code where it says `key` it actually means `id` because I haven't yet updated to account for the `key`-`hash`>`id` thing. Adding that to the todo list above.

And yea, a `compute_content_id` function will be necessary. I think it should be in `utils.rs`.

victor-wei126:
In `protocol.rs` of the main branch, this [line](https://github.com/ethereum/trin/blob/master/trin-core/src/portalnet/protocol.rs#L163) of code, it seems to be using `content_key` as the index though?

trin-core/src/portalnet/storage.rs
mrferris: i don't think that that code yet takes into account the fact the `content_id` is what's actually used as the key with which we index into the database.

I do think that the key/id nomenclature is a bit confusing because I think of a key as what's used to index a hash table, but in Portal Network the `id` is. Perhaps that's my own misunderstanding.

trin-core/src/bin/generate-data.rs
```diff
@@ -0,0 +1,138 @@
...
+    rand::thread_rng()
+        .sample_iter(&Alphanumeric)
+        .take(number_of_bytes as usize)
+        .map(char::from)
+        .collect()
+
+}
+
+// Placeholder content key -> content id conversion function
+fn sha256(key: &String) -> [u8; 32] {
```
victor-wei126, line 83: I was wondering why use an array of 32 bytes instead of U256 data type?

mrferris: Good question, I moved away from U256s internally because SQLite won't take them so I had to convert them anyway, but this is external facing so I'll make it a U256.

Are you returning U256 from the derivation functions?

victor-wei126: Gotcha. I'm not 100% sure yet what the derivation functions will return. Currently, I'm using the sha3 crate that you're using, and my function returns whatever their `finalize()` function returns

mrferris: ok, i think that's returning a GenericArray. if we do decide to return a [u8; 32], you can see how to convert from that to a [u8; 32] in the above sha256 function. 

Also, just to put this here: I'm not using a U256 because I can't figure out how to turn one back into bytes. Primitive Rust integer types have `.to_bytes()` functions, U256 (built with parity's uint library) does not seem to have anything like that. It does have the `::from()` function which allows going in the other direction (byte array -> U256) just fine. Any ideas welcome.

mrferris: @victor-wei126 I can convert U256 to what I'm working with, so let's continue using U256.

njgheorghita: @mrferris Sorry this slipped through the cracks somehow, but yeah I'd support sticking with `U256`

carver:
>     * `get_current_radius() -> u64`: Gets the current radius (currently u64, should probably be a u256 because I'm under the assumption that a radius of infinity is going to be represented over the wire by u256::max).

- Will probably want the ability to modify the radius on the fly, but that can be saved for another PR.
- One option for how to represent the radius is as the high 64 bits of a 256-bit value. I think it's okay to not support a radius smaller than `2^(256-64)`. We probably want a custom type that stores the 64-bit value internally but has easy methods to compare with U256 numbers.

I say all this without looking at the code yet. I'll do a quick scan now. Feel free to ping me again after a rebase and the tests pass.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+            [content_id.to_vec()],
+        ).unwrap();
+        
+    }
+
+    pub fn store(&mut self, key: &String, value: &String) {
+
+        // Check whether we should store this data
+        if !self.should_store(&key) {
+            println!("Not storing: {}", key);
```
carver, line 168: Log instead, probably `debug!`.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+            
+            println!("Capacity reached, deleting farthest: {}", hex::encode(&id_to_remove.clone().unwrap()));
+
+            self.db.delete(&id_to_remove.clone().unwrap()).expect("Failed to delete key.");
+            self.meta_db_remove(&id_to_remove.clone().unwrap());
+
+            self.db.flush().expect("Failed to flush db.");
+            
+            match self.find_farthest_content_id() {
+                None => { error!("Failed to find farthest!") },
```
carver, line 203:
```suggestion
                None => {
                    error!("Database is over-capacity, but could not find find entry to delete!");
                    self.farthest_content_id = None;
                    // stop attempting to delete, to avoid risk of infinite loop
                    break;
                },
```

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+    }
+
+    pub fn capacity_reached(&self) -> bool {
+
+        let storage_usage = self.get_total_storage_usage_in_bytes_from_network().unwrap();
+        storage_usage > (self.storage_capacity_kb * 1000)
+
+    }
+
+    pub fn get_total_storage_usage_in_bytes_on_disk(&self) -> u64 {
```
carver, line 235: Is this used anywhere? Would like to minimize dead code.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+
+        self.get_total_size_of_directory_in_bytes(get_data_dir()).unwrap()
+
+    }
+
+    pub fn get_total_storage_usage_in_bytes_from_network(&self) -> Result<u64, String> {
```
carver, line 241: Since this is returning a `Result` anyway, we might as well bubble up the errors instead of `unwrap` to panic on them.  (I'm thinking of the `?` operator here)

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+                panic!("State distance function is not implemented yet.")
+            }
+
+        };
+
+        Some(result)
+
+    }
+
+    fn get_total_size_of_directory_in_bytes(&self, path: String) -> std::io::Result<u64> {
```
carver, line 304: Part of the same dead code, I think?

carver (commented):
There's a *lot* of unhandled errors (`unwrap`s) happening here. Would be best to either 1) handle them, 2) bubble them up, or 3) include in-code comments about why it's okay for trin to crash in all the scenarios where each `unwrap()` might fail.

mrferris:
I don't feel like a `git rebase master` was supposed to leave each of my commits on here twice, along with every commit from master. Any thoughts on that are welcome.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+    db: rocksdb::DB,
+    meta_db: rusqlite::Connection,
+    distance_function: DistanceFunction,
+    content_key_to_id_function: Box<ContentKeyToIdDerivationFunction>,
+
+}
+
+impl PortalStorage {
+
+    pub fn new(config: &PortalStorageConfig, convert_function: impl Fn(&String) -> U256 + 'static) -> Result<Self, String> {
```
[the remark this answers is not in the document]
mrferris: Yea I was going to bubble up DB errors but hadn't yet. Changed it to not return a Result, and have DBs passed in.

pipermerriam:
You can try a `git rebase -i master` which does interactive mode, allowing you to see what commits are about to be rebased.  Similarly `git reflog` is your ultimate "undo" button letting you rewind to previous moments if something goes awry.

mrferris:
@pipermerriam I should have mentioned that the actual `rebase` itself went fine (didn't use the `-i`, but went just as fine as when I have in the past). `git log` looks completely normal on my machine. The only weirdness is on this pull request page, Not sure if it's a github issue.

trin-core/src/portalnet/storage.rs
mrferris: This is not currently used inside PortalStorage, but I was thinking that it would be a useful method to be exposed in order to keep track of actual usage on disk.

Now that I think again, I think that it should be non-public and called & result logged on each call to `store(..)`. I'll ad that.

mrferris:
> There's a _lot_ of unhandled errors (`unwrap`s) happening here. 

Yea, I think that the "Error handling is pretty close to non-existent." in the Trin README went to my head. 👍
I'm cleaning that up now. 

> Will probably want the ability to modify the radius on the fly, but that can be saved for another PR.

When do you envision setting the radius on the fly from outside of this struct?

carver:
> Yea, I think that the "Error handling is pretty close to non-existent" in the Trin README went to my head. grin
> I'm cleaning that up now.

Hah thanks. 😆 That readme hasn't been updated in a while, it's probably time to take another pass at it. The situation has improved a bit since then. :)


> When do you envision setting the radius on the fly from outside of this struct? I'm guessing for testing but wonder if there are scenarios you have in mind during actual operations.

Oh yeah I think I wrote that comment before the review, so didn't understand the approach yet. What I mean is the ability to change `storage_capacity_kb` on the fly.

---

Can you get all the formatting passing so we can see the tests pass?

mrferris:
I edited this PR's initial description's list of method signatures to reflect the cleaned up Error handling.

I experimented with a few different crates for custom errors and settled on `thiserror` because it doesn't require any boilerplate `From<> for` in order to use the `?` operator, and it's well-maintained compared to the competition.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,605 @@
...
+            content_key_to_id_function: Box::new(convert_function)
+        };
+
+        // Check whether we already have data, and if so 
+        // use it to set the farthest_key and data_radius fields
+        match storage.find_farthest_content_id() {
+            Some(content_id) => {
+                storage.farthest_content_id = Some(content_id.clone());
+                if storage.capacity_reached() {
+                    storage.data_radius = storage.distance_to_content_id(&content_id);
```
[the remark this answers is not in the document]
mrferris: created issue #124

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,707 @@
...
+
+    /// Internal method that returns the distance between our node ID and a given content ID
+    fn distance_to_content_id(&self, content_id: &[u8; 32]) -> u64 {
+        let byte_vector = xor_two_values(content_id, &self.node_id.raw().to_vec());
+
+        PortalStorage::byte_vector_to_u64(byte_vector)
+    }
+
+    // Takes the most significant 8 bytes of a vector and casts them into a u64.
+    // The equivalent of a conversion from nanometers to meters.
```
carver, line 375:
```suggestion
    // With a 32-byte big-endian unsigned integer input, this is equivalent to dividing by 2^(256 - 32)
```

carver (approved):
Cool, looks like the big things left are removing/replacing the `println`s and squashing the commits, after addressing a few suggestions/suggestions below.

trin-core/src/bin/generate-data.rs
```diff
@@ -0,0 +1,120 @@
...
+    let storage_config = PortalStorageConfig {
+        // Arbitrarily set capacity at a quarter of what we're storing.
+        // todo: make this ratio configurable
+        storage_capacity_kb: (num_kilobytes / 4) as u64,
+        node_id,
+        distance_function: DistanceFunction::Xor,
+        db: Arc::new(db),
+        meta_db: Arc::new(meta_db),
+    };
+    let mut storage = PortalStorage::new(storage_config, |key| sha256(key))?;
```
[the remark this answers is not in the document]
mrferris: I don't have strong feelings either way, though I'm curious what your there reasoning is

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,707 @@
...
+
+        let mut array: [u8; 4] = [0, 0, 0, 0];
+        for (index, byte) in vec.iter().take(4).enumerate() {
+            array[index] = byte.clone();
+        }
+
+        u32::from_be_bytes(array)
+    }
+
+    /// Helper function for opening a SQLite connection.
```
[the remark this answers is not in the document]
mrferris: Reasoning for this is that I'm also using these functions in `bin/generate-data.rs`. Do you know the syntax for importing functions from the module `test` from within `bin/generate-data.rs`? I tried something like `pub mod test` in this file and then `use trin_core::portalnet::storage::test:{setup_sqlite, setup_rocksdb}` in `bin/generate-data.rs` but was getting `could not find test`, so I settled on this approach.

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,707 @@
...
+#[derive(Copy, Clone)]
+pub enum DistanceFunction {
+    Xor,
+    State,
+}
+
+/// Signature of the function that must be passed into the call to new
+type ContentKeyToIdDerivationFunction = dyn Fn(&String) -> U256;
+
+/// Struct for spe
```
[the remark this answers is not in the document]
mrferris: good catch 😅

trin-core/Cargo.toml
```diff
@@ -15,19 +15,23 @@ futures = "0.3.13"
 hex = "0.4.3"
 lazy_static = "1.4.0"
 log = "0.4.14"
+parking_lot = "0.11.1"
```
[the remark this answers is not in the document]
mrferris: i'm very surprised that clippy didn't yell at me for that

trin-core/src/portalnet/storage.rs
```diff
@@ -0,0 +1,707 @@
...
+        };
+        let mut storage = PortalStorage::new(storage_config, |key| sha256(key))?;
+
+        let key: String = "YlHPPvteGytjbPHbrMOVlK3Z90IcO4UR".to_string();
+        let value: String = "OGFWs179fWnqmjvHQFGHszXloc3Wzdb4".to_string();
+        storage.store(&key, &value)?;
+
+        let bytes = storage.get_total_storage_usage_in_bytes_from_network()?;
+
+        println!("{}", bytes);
```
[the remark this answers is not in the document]
mrferris: Good catch, that's a leftover from a previous approach where `get_total_storage` measured the actual space on disk, which was fairly unpredictable.

Updated to use assert, now that the approach is simpler & predictable.

trin-core/Cargo.toml
```diff
@@ -15,19 +15,23 @@ futures = "0.3.13"
...
 reqwest = { version = "0.11.0", features = ["blocking"] }
 rlp = "0.5.0"
 rocksdb = "0.16.0"
 serde = {version = "1.0.125", features = ["derive"] }
 serde_json = "1.0.59"
-structopt = "0.3"
 stunclient = "0.1.2"
+structopt = "0.3"
+sha3 = "0.9.1"
+
```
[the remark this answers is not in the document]
mrferris: surprised that makes it through clippy & fmt

trin-core/src/bin/generate-data.rs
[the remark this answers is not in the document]
ogenev: My intuition is that the configuration is for things that wer would like to tweak often, like different kind of parameters and values. The DB sounds more like a parameter to the PortalStorage instead of a configuration param, because probably we will not want to swap the db often.  Also when writing integration tests I would like to stub the db and directly pass it to the new function,  but it is personal preference I guess...

trin-core/src/portalnet/storage.rs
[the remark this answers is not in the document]
ogenev: I see, so I guess the comment is kinda misleading then. If it was me, I would create new db folder inside `trin-core` and put everything related to storage inside. This way we can also have utils.rs file with methods shared between the different db modules.

Issue #46, opened 2021-07-22
Track State Network peers independently from discv5 base layer


carver:
As I imagine it, that means: replicating (reusing?) the Kademlia bucket code used in the discv5 crate, but with state-net clients only. Presumably, we also want to do the ping/pong to ensure liveness, and the network crawl to try to fill the buckets.

So that would break down into:
- [ ] A new set of Kademlia buckets for tracking
- [ ] Use ping to ensure peer liveness
- [ ] Respond to the overlay "find nodes" using peers from the new buckets
- [ ] Crawl the peers in the overlay network to fill buckets
- [ ] When "found content" has an empty payload, use the new buckets to fill it (ie~ in `find_nodes_close_to_content()`)

This is a fairly large issue, bigger than one PR presumably. So if whoever works on this wants to open separate issues for each task I'd 👍🏻 . Otherwise, I'm happy to split this into subtasks after someone submits the first PR (maybe just the first bullet above?). I think that splitting, and identifying next tasks, will be easier once there's a bit more structure.

carver:
@SamWilsn want to take this on, or at least the first steps of it?

carver:
@jacobkaufmann seems to already be looking at how to extract and make some of the Kademlia logic easily available, so maybe look for an initial PR from him first, @SamWilsn .

jacobkaufmann:
It looks like Piper's proposal for an alternative, non-XOR distance metric would likely rule out the option to use the `sigp/discv5` library directly for the state network bucket table. The XOR distance metric method is baked in to a lot of core logic, and I'm not sure supporting a custom distance metric is something that makes sense for `sigp/discv5`.

And even if we did not require support for a non-XOR distance metric, the Kademlia structures are kept private in the `sigp/discv5` library.

For the `libp2p/rust-libp2p` library, which is what the `sigp/discv5` Kademlia implementation is based on, the distance situation appears the same, but the Kademlia structures are made public.

I'm not sure we can get away from extracting and modifying the Kademlia logic, at least for the state network. And if we need to do so for the state network, then my thought is that we should use the same implementation for both the state and history networks. In other words, if we need to extract our own implementation for at least one overlay network, we should re-use that implementation across all overlay networks.

jacobkaufmann:
Opened a [PR](https://github.com/sigp/discv5/pull/82) on the `sigp/discv5` repository to make the Kademlia items public.

PR #70, opened 2021-08-12
Implement distance function

njgheorghita:
Implement new distance function & tests as defined at 
https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout

Uses the [num::bigint](https://docs.rs/num/0.4.0/num/bigint/struct.BigInt.html) library - which might not be the most efficient way to do things? it seemed like the [`ethereum-types` - `U256`](https://docs.rs/ethereum-types/0.12.0/ethereum_types/struct.U256.html) wasn't ideal for the calculations needed here. Either way, I'm open to changing the argument types here to whatever people think is the most useful. 

This is also probably not the most performant way to handle this calculation - and as it's likely to be used a lot, further optimizations down the road would probably become very useful.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -187,6 +187,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "63396b8a4b9de3f4fdfb320ab6080762242f66a8ef174c49d8e19b674db4cdbe"
 
 [[package]]
+name = "byte-slice-cast"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "65c1bf4a04a88c54f589125563643d773f3254b5c38571395e2b591c693bbc81"
+
+[[package]]
 name = "bytecodec"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


njgheorghita:
Not quite sure who to tag for reviews - but if any contributors have input, I'd love to hear it. @jacobkaufmann maybe if I'm remembering correctly, you were working with `ethereum-types` `U256` recently?

Also @carver it looks like Circle is complaining with: "This job has been blocked because no credits are available on your plan. Please upgrade to continue building.", which is why the `win-build` CI run is not passing. Not sure what the best way to proceed is - either load some credits, or maybe it's a good opportunity to migrate this repo to the ethereum org?

jacobkaufmann:
> Not quite sure who to tag for reviews - but if any contributors have input, I'd love to hear it. @jacobkaufmann maybe if I'm remembering correctly, you were working with `ethereum-types` `U256` recently?

@njgheorghita I think we may wish to hold off on this implementation until we are able to define a `Metric` trait for the `KBucketsTable` so that we can define the distance function with the trait implementation. I'm working on a PR for #46, which I think would be a prerequisite for this distance function.

After the PR for the separation of the discv5 and overlay routing tables, I was going to start on a PR for the `Metric` trait.

Also, AFAIK everything for `U256` uses the `uint` crate. Is there a reason to bring in the dependency on `ethereum-types`?

njgheorghita:
@jacobkaufmann Sounds good, I'll close this for now and just leave it up as a branch on my fork in case it's useful down the road. 

> Is there a reason to bring in the dependency on ethereum-types?
I actually used the `num` [crate](https://docs.rs/num/0.4.0/num/bigint/struct.BigInt.html) and basically because it was the easiest way to handle large negative integers (ie. `- 2**255`). I'm sure there's a more performant way to do the math in the distance function, but I wasn't too worried about optimization at the moment.

PR #72, opened 2021-08-12
Reduce clippy to only run for trin

### LINKED ISSUE #71: Explore why `Run Clippy` job in CI is taking so long
Right now the `Run Clippy` job on my current [ci build](https://app.circleci.com/pipelines/github/carver/trin/213/workflows/791e4a28-5deb-4692-aa44-e0035a979a6b/jobs/310) took 8 1/2 minutes. Seems way too long, it would be nice if somebody could explore our options for cutting this down. It seems as if clippy is checking every single dependency, which imo is completely unnecessary, and a waste of a time sitting around waiting for the ci check to complete.

njgheorghita:
Clippy now only checks the trin crates - reduces job time from ~8mins to ~8secs

Fixes #71 

From circle output
```
    Checking trin-core v0.1.0 (/home/circleci/project/trin-core)
    Checking trin-state v0.1.0 (/home/circleci/project/trin-state)
    Checking trin-history v0.1.0 (/home/circleci/project/trin-history)
    Checking trin v0.1.0 (/home/circleci/project)
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,14 +29,12 @@ jobs:
             - run:
                 name: Run rustfmt
                 command: cargo fmt -- --check
-            # Run the sub-commands of the orb jobs separately, since the combined steps for Clippy can be long enough to trigger a timeout (>10min). Building the source is the main contributor.
             - run:
                 name: Install Clippy
                 command: rustup component add clippy
             - run:
                 name: Run Clippy
                 command: cargo clippy --package trin -- --deny warnings
-                no_output_timeout: 20m
             - rust/build:
                 release: false
                 with_cache: false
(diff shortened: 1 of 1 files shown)


KolbyML (changes requested):
I haven't tested the change yet, but if we were going to add this ``--package trin --`` should also be added to Windows CI Clippy run

njgheorghita:
@KolbyML Good catch - added it to the windows build. Can't imagine it's going to cause a problem on the windows side of things - but I've been wrong before. I'm happy to just leave this as is for now until we get the `win-build` up and running again on circleci to verify it's working.

KolbyML (approved):

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
                 command: rustup component add clippy
             - run:
                 name: Run Clippy
-                command: cargo clippy -- --deny warnings
+                command: cargo clippy --package trin -- --deny warnings
                 no_output_timeout: 20m
```
KolbyML, line 39: If this is going to shorten clippies time you can probably also just remove ``no_output_timeout: 20m`` in the same PR because we shouldn't need the extra time anymore.

Also maybe the comment on why I did it, but I guess that is optional.

KolbyML (changes requested):

PR #73, opened 2021-08-13
Add overlay node state to manage overlay routing table

jacobkaufmann:
- Add Overlay struct to manage node state for overlay network.
- Add APIs to Overlay to query node state.

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -1,0 +1,169 @@
+use crate::portalnet::types::SszEnr;
+use crate::portalnet::{Enr, U256};
+use crate::utils::xor_two_values;
+
+use discv5::enr::NodeId;
+use discv5::kbucket::{Filter, KBucketsTable};
+use parking_lot::RwLock;
+use std::sync::Arc;
+use std::time::Duration;
+
+/// Maximum number of ENRs in response to FindNodes.
+const FIND_NODES_MAX_NODES: usize = 32;
+/// Maximum numer of ENRs in response to FindContent.
+const FIND_CONTENT_MAX_NODES: usize = 32;
+
+#[derive(Clone)]
+pub struct Node {
+    enr: Enr,
+    data_radius: U256,
+}
+
+impl Node {
+    pub fn enr(&self) -> Enr {
+        self.enr.clone()
+    }
+
+    pub fn data_radius(&self) -> U256 {
+        self.data_radius.clone()
+    }
+}
+
+impl std::cmp::Eq for Node {}
+
+impl PartialEq for Node {
+    fn eq(&self, other: &Self) -> bool {
+        self.enr == other.enr
+    }
+}
+
+/// Configuration parameters for the overlay network.
+#[derive(Clone)]
+pub struct Config {
+    bucket_pending_timeout: Duration,
+    max_incoming_per_bucket: usize,
+    table_filter: Option<Box<dyn Filter<Node>>>,
+    bucket_filter: Option<Box<dyn Filter<Node>>>,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            bucket_pending_timeout: Duration::from_secs(60),
+            max_incoming_per_bucket: 16,
+            table_filter: None,
+            bucket_filter: None,
+        }
+    }
+}
+
+/// The node state for a node in an overlay network on top of Discovery v5.
+pub struct Overlay {
+    // The ENR of the local node.
+    local_enr: Arc<RwLock<Enr>>,
+    // The data radius of the local node.
+    data_radius: Arc<RwLock<U256>>,
+    // The routing table of the local node.
+    kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
+}
+
+impl Overlay {
+    pub fn new(local_enr: Enr, data_radius: U256, config: Config) -> Self {
+        let kbuckets = Arc::new(RwLock::new(KBucketsTable::new(
+            local_enr.node_id().into(),
+            config.bucket_pending_timeout,
+            config.max_incoming_per_bucket,
+            config.table_filter,
+            config.bucket_filter,
+        )));
+        let local_enr = Arc::new(RwLock::new(local_enr));
+        let data_radius = Arc::new(RwLock::new(data_radius));
+
+        Self {
+            local_enr,
+            data_radius,
+            kbuckets,
+        }
+    }
+
+    /// Returns the local ENR of the node.
+    pub fn local_enr(&self) -> Enr {
+        self.local_enr.read().clone()
+    }
+
+    // Returns the data radius of the node.
+    pub fn data_radius(&self) -> U256 {
+        self.data_radius.read().clone()
+    }
+
+    /// Returns a vector of the ENRs of the closest nodes by the given distances.
+    pub fn nodes_by_distance(&self, mut distances: Vec<u64>) -> Vec<Enr> {
+        let mut nodes_to_send = Vec::new();
+        distances.sort_unstable();
+        distances.dedup();
+
+        if let Some(0) = distances.first() {
+            // if the distance is 0 send our local ENR
+            nodes_to_send.push(self.local_enr.read().clone());
+            distances.remove(0);
+        }
+
+        if !distances.is_empty() {
+            let mut kbuckets = self.kbuckets.write();
+            for node in kbuckets
+                .nodes_by_distances(distances, FIND_NODES_MAX_NODES)
+                .into_iter()
+                .map(|entry| entry.node.value.clone())
+            {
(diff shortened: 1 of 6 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+            .take(FIND_CONTENT_MAX_NODES)
+            .filter(|node_record| &node_record.0 < &self_distance)
+            .map(|node_record| SszEnr::new(node_record.1))
+            .collect();
+
+        closest_nodes
+    }
+
+    /// Returns an iterator over all ENR node IDs of nodes currently contained in the routing table.
+    pub fn table_entries_id(&self) -> Vec<NodeId> {
```
ogenev, line 153: Not sure, but I think it makes sense to combine this and the next method `tables_entries_enr` into one and return a vector of tuples.

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+            .into_iter()
+            .take(FIND_CONTENT_MAX_NODES)
+            .filter(|node_record| &node_record.0 < &self_distance)
+            .map(|node_record| SszEnr::new(node_record.1))
+            .collect();
+
+        closest_nodes
+    }
+
+    /// Returns an iterator over all ENR node IDs of nodes currently contained in the routing table.
```
ogenev, line 152:
```suggestion
    /// Returns a vector of all ENR node IDs of nodes currently contained in the routing table.
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn table_entries_id(&self) -> Vec<NodeId> {
+        self.kbuckets
+            .write()
+            .iter()
+            .map(|entry| *entry.node.key.preimage())
+            .collect()
+    }
+
+    /// Returns an iterator over all the ENR's of nodes currently contained in the routing table.
```
ogenev, line 161:
```suggestion
    /// Returns a vector of all the ENR's of nodes currently contained in the routing table.
```

ogenev (approved):
Left some minor comments, LGTM 🚀

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+
+impl PartialEq for Node {
+    fn eq(&self, other: &Self) -> bool {
+        self.enr == other.enr
+    }
+}
+
+/// Configuration parameters for the overlay network.
+#[derive(Clone)]
+pub struct Config {
```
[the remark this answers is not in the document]
jacobkaufmann: Here I was following convention from `discovery.rs`, where the struct is named `Config` but is imported as `DiscoveryConfig`. Similarly, when we need to use this struct, we can import as `OverlayConfig`.

carver:
Ah @jacobkaufmann, I was expecting that your overlay work was going to on the chain history crate, where Sam was working on the same thing state network crate.

Maybe it's no issue, but I'd like to see @SamWilsn take a look at it.

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+}
+
+/// The node state for a node in an overlay network on top of Discovery v5.
+pub struct Overlay {
+    // The ENR of the local node.
+    local_enr: Arc<RwLock<Enr>>,
+    // The data radius of the local node.
+    data_radius: Arc<RwLock<U256>>,
```
SamWilsn, line 65: Does anything need to happen if either of these update? Like sending an announcement? Could we worth looking at the broadcast or watch channels in tokio.

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Returns a vector of the ENRs of the closest nodes by the given distances.
+    pub fn nodes_by_distance(&self, mut distances: Vec<u64>) -> Vec<Enr> {
+        let mut nodes_to_send = Vec::new();
+        distances.sort_unstable();
+        distances.dedup();
+
+        if let Some(0) = distances.first() {
+            // if the distance is 0 send our local ENR
+            nodes_to_send.push(self.local_enr.read().clone());
+            distances.remove(0);
```
SamWilsn, line 108: Extremely minor nit, but removing the zeroth element would require shifting the whole `Vec`.

You could use a mutable binding:

```rust
let mut distances = distances.as_slice();
if let Some(0) = distances.first() {
    // TODO
    distances = &mut distances[1..];
}

File: trin-core/src/portalnet/protocol.rs [hunk 7]
```diff
@@ -215,11 +215,17 @@ impl PortalnetProtocol {
             .await
             .map_err(|e| e.to_string())?;
 
+        let overlay = Arc::new(Overlay::new(
```
SamWilsn, line 218: Everything in `Overlay` is in an `Arc` already, so you could probably `derive(Clone)` on it, and save a level of indirection.

jacobkaufmann:
> Ah @jacobkaufmann, I was expecting that your overlay work was going to on the chain history crate, where Sam was working on the same thing state network crate.

Apologies here for the misunderstanding. I should have communicated this more clearly. If preferred, I can move this work into the chain history crate. It definitely falls in the realm of "core code", but maybe we don't want to introduce that common code yet. If `Overlay` also manages the local content store or holds the object that does, then we may not be able to develop the two networks as independently as we would like to. Alternatively we could define traits in the core crate and keep implementations separate.

SamWilsn:
Do _not_ worry about stepping on my toes. When/if I get the chance to switch back to actively coding on trin, I'll ping you to get an idea of where I can contribute, but it doesn't look like that'll happen for a bit.

trin-core/src/portalnet/overlay.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn enr(&self) -> Enr {
+        self.enr.clone()
+    }
+
+    pub fn data_radius(&self) -> U256 {
+        self.data_radius.clone()
+    }
+}
+
+impl std::cmp::Eq for Node {}
```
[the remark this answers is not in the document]
jacobkaufmann: The `Eq` trait has no methods, and it requires that we implement the `PartialEq` trait. There is more info available in Rust docs. To derive the `Eq` trait, then all fields must be `Eq`, and I'm not sure that is the desired behavior. Should a node's data radius be part of its equivalence to another node, or should we only consider a node's ENR?

trin-core/src/portalnet/overlay.rs
jacobkaufmann: Doing this would require us to call `to_vec` on `distances` below when we call `nodes_by_distances` on `kbuckets`. If the memory overhead of the in-place shift versus the copy to a new vector are similar, then I think the existing code is easier to understand.

trin-core/src/portalnet/protocol.rs
jacobkaufmann: To clarify, you are suggesting that we `derive(Clone)` for `Overlay`? If so, wouldn't we still require `Arc<Overlay>` to share it between threads? I think there's some piece that I'm missing here.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: For `data_radius`, I don't think we would require a channel because `Overlay` would encapsulate all logic around `data_radius`.

For `local_enr`, I think those updates would originate from `Discv5`, which has public methods to update the ENR.

trin-core/src/portalnet/overlay.rs
SamWilsn: Ugh, that [should be changed in `discv5` to a slice](https://github.com/sigp/discv5/blob/e3c7e308ef906dab57fb5088c0cf3c0a970475f4/src/kbucket.rs#L563). I can make a PR for that, unless you want to.

trin-core/src/portalnet/protocol.rs
SamWilsn: Yep, that's exactly what I'm suggesting.

Instead of passing an `Arc<Overlay>` between threads, you can just clone a single `Overlay` and share it between threads. The `Arc`s inside will still point to the same data.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: Opened a PR [here](https://github.com/sigp/discv5/pull/86).

carver (approved):
I don't have any more to add here, but I assume you're blocking on the `discv5` PR? I'd recommend just taking care of the other suggested changes, and add a new issue to make this change: https://github.com/carver/trin/pull/73#discussion_r689803016

That way this PR can get merged without waiting on the `discv5` folks.

PR #75, opened 2021-08-15
Added protocol param to send_talkreq()

KolbyML:
This is something I added to one of my branches, but I thought I might as well make a PR for it because I think it will be quite useful for upcoming protocols also my uTP code will use this so I might as well get it in just so it is already there because I might not be PR'ing that for a while longer.

### DIFF
--- a/trin-core/src/portalnet/discovery.rs
+++ b/trin-core/src/portalnet/discovery.rs
@@ -1,7 +1,7 @@
 #![allow(dead_code)]
 
 use super::types::{HexData, SszEnr};
-use super::{protocol::PROTOCOL, Enr};
+use super::Enr;
 use crate::utils::xor_two_values;
 use discv5::enr::{CombinedKey, EnrBuilder, NodeId};
 use discv5::{Discv5, Discv5Config};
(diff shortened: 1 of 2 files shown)


carver (approved):
I'm not seeing how it will be used by uTP yet, but I'm sure it will come clear. No objections.

trin-core/src/portalnet/discovery.rs
```diff
@@ -151,11 +151,12 @@ impl Discovery {
     pub async fn send_talkreq(
         &self,
         enr: Enr,
+        protocol: &str,
```
SamWilsn, line 154: Minor nit: if you're converting to a `Vec` later anyways, you should take a `String`, which _can_ save a copy.

KolbyML:
> I'm not seeing how it will be used by uTP yet, but I'm sure it will come clear. No objections.

Well uTP uses talk requests subbing in uTP instead of portal in the uTP spec

image

image

so overlay messages would use Overlay for example here as well.

Then for example 
image

This is just an idea, but yeah

carver:
> > I'm not seeing how it will be used by uTP yet, but I'm sure it will come clear. No objections.
> 
> Well uTP uses talk requests subbing in uTP instead of portal in the uTP spec
> 
> ...

Yup, totally makes sense! I just haven't looked at the our utp spec in a while :)

Next time, feel free to merge after I've given a 👍🏻  and you've resolved any suggestions. I'll merge it now.

PR #78, opened 2021-08-19
Portal network cross-client testing framework

ogenev:
Following the discussion on https://github.com/carver/trin/issues/50, this is an experimental crate with a dummy node that sends request messages to a target node ( a node under test). A dummy peer is a minimal bare-bones node that is capable of executing request to a remote node and receive a response.

This crate may be useful for testing how a portal node responds to different network calls.

Current implementation is very primitive, it is just sending two ping requests to state and history network. This can be expanded by adding JSON-RPC API that will execute different test suite scenarios on the target node.

_JSON-RPC commands are not supported for now. I think we should first refactor the json-rpc code in `trin-core`, so we can make those parts reusable._

### DIFF
--- a/ethportal-peertest/Cargo.toml
+++ b/ethportal-peertest/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "ethportal-peertest"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+clap = "2.33.3"
+log = "0.4.14"
+rocksdb = "0.16.0"
+structopt = "0.3"
+tokio = {version = "1.8.0", features = ["full"]}
+tracing = "0.1.26"
+tracing-subscriber = "0.2.18"
+trin-core = { path = "../trin-core" }
+
+[dependencies.discv5]
+version = "0.1.0-beta.5"
+git = "https://github.com/sigp/discv5"
--- a/ethportal-peertest/README.md
+++ b/ethportal-peertest/README.md
@@ -1,0 +1,10 @@
+# ethportal-peertest
+
+Run a portal network node that you want to test and pass node's Enr as a target node argument.
+
+```sh
+cd ethportal-peertest
+
+RUST_LOG=debug cargo run -p ethportal-peertest -- --target_node enr:-IS4QBDHCSMoYoC5UziAwKSyTmMPrhMaEpaE52L8DDAkipqvZQe9fgLy2wVuuEJwO9l1KsYrRoFGCsNjylbd0CDNw60BgmlkgnY0gmlwhMCoXUSJc2VjcDI1NmsxoQJPAZUFErHK1DZYRTLjk3SCNgye9sS-MxoQI-gLiUdwc4N1ZHCCIyk
+
+```
--- a/ethportal-peertest/src/cli.rs
+++ b/ethportal-peertest/src/cli.rs
@@ -1,0 +1,58 @@
+use std::env;
+use std::ffi::OsString;
+use structopt::StructOpt;
+
+const DEFAULT_LISTEN_PORT: &str = "9876";
+const DEFAULT_WEB3_IPC_PATH: &str = "/tmp/json-rpc-peertest.ipc";
+
+#[derive(StructOpt, Debug, PartialEq)]
+#[structopt(
+    name = "ethportal-peertest",
+    version = "0.0.1",
+    about = "Testing framework for portal network peer-to-peer network calls"
+)]
+pub struct PeertestConfig {
+    #[structopt(
+        default_value(DEFAULT_LISTEN_PORT),
+        short = "p",
+        long = "listen_port",
+        help = "The UDP port to listen on."
+    )]
+    pub listen_port: u16,
+
+    #[structopt(
+        default_value(DEFAULT_WEB3_IPC_PATH),
+        long = "web3_ipc_path",
+        help = "path to json-rpc socket address over IPC"
+    )]
+    pub web3_ipc_path: String,
+
+    #[structopt(
+        short,
+        long = "target_node",
+        help = "Base64-encoded ENR's of the nodes under test"
+    )]
+    pub target_node: String,
+}
+
+impl PeertestConfig {
+    pub fn new() -> Self {
+        Self::new_from(env::args_os()).expect("Could not parse trin arguments")
+    }
+
+    pub fn new_from<I, T>(args: I) -> Result<Self, String>
+    where
+        I: Iterator<Item = T>,
+        T: Into<OsString> + Clone,
+    {
+        let config = Self::from_iter(args);
+
+        Ok(config)
+    }
+}
+
+impl Default for PeertestConfig {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/ethportal-peertest/src/events.rs
+++ b/ethportal-peertest/src/events.rs
@@ -1,0 +1,109 @@
+use discv5::{Discv5Event, TalkRequest};
+use log::{debug, error, warn};
+use std::convert::TryInto;
+use std::sync::Arc;
+use tokio::sync::mpsc;
+use trin_core::portalnet::overlay::OverlayProtocol;
+use trin_core::portalnet::types::Message;
+use trin_core::portalnet::utp::UtpListener;
+
+pub const HISTORY_NETWORK: &str = "history";
+pub const STATE_NETWORK: &str = "state";
+pub const UTP_PROTOCOL: &str = "utp";
+
+pub struct PortalnetEvents {
+    pub overlay: Arc<OverlayProtocol>,
+    pub protocol_receiver: mpsc::Receiver<Discv5Event>,
+    pub utp_listener: UtpListener,
+}
+
+impl PortalnetEvents {
+    pub async fn new(overlay: Arc<OverlayProtocol>, utp_listener: UtpListener) -> Self {
+        let protocol_receiver = overlay
+            .discovery
+            .write()
+            .await
+            .discv5
+            .event_stream()
+            .await
+            .map_err(|e| e.to_string())
+            .unwrap();
+
+        Self {
+            overlay: Arc::clone(&overlay),
+            protocol_receiver,
+            utp_listener,
+        }
+    }
+
+    /// Receives a request from the talkreq handler and sends a response back
+    pub async fn process_discv5_requests(mut self) {
+        while let Some(event) = self.protocol_receiver.recv().await {
+            debug!("Got discv5 event {:?}", event);
+
+            let request = match event {
+                Discv5Event::TalkRequest(r) => r,
+                _ => continue,
+            };
+
+            match std::str::from_utf8(request.protocol()) {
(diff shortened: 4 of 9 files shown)


ogenev:
Updated this PR by addressing some of the comments in https://github.com/carver/trin/pull/79. (based on this branch).

I think we should decide how exactly we want to proceed with cross-client testing, the current dummy implementation is just sending a batch of request messages to the target node and it is printing the results. some possible options that come to my mind:

- interactive testing
We set a communication channel from the command line to the dummy node for interaction and it is left to the user to execute command with the desired request behaviors. This is something similar to the geth devp2p command [here](https://github.com/ethereum/go-ethereum/tree/master/cmd/devp2p).

- predefined testing
Create different request scenarios and expecting responses from the target node/nodes. We can panicking or keep track of failures to display them at the end of the test.

We can of course combine the best of the both worlds and execute test suites from the command line on top of single request to the target node, similar to [discv5 tests ](https://github.com/ethereum/go-ethereum/blob/master/cmd/devp2p/internal/v5test/discv5tests.go)in geth.

ogenev:
@carver, It looks like circleci build is failing unexpectedly (it builds fine on my machine). After some investigation, I'm suspecting that the circleci container is running out of memory and it is randomly killing the trin package builds.

carver:
> @carver, It looks like circleci build is failing unexpectedly (it builds fine on my machine). After some investigation, I'm suspecting that the circleci container is running out of memory and it is randomly killing the trin package builds.

Yup, looks right to me, too. One reasonable approach is to reduce the number of parallel builds in the circle job to maybe 4 or 2. It's currently defaulting to the number of CPUs. You could do that by swapping this command in where the `rust/build` is currently in the `.circleci/config.yml`:
```sh
cargo build --jobs 4
```
~ source https://stackoverflow.com/a/68031406/8412986

I wouldn't drop it down to `--jobs 1` without testing the higher parallelism, because I'm guessing that will severely slow the build.

carver:
> We set a communication channel from the command line to the dummy node for interaction and it is left to the user to execute command with the desired request behaviors. This is something similar to the geth devp2p command here.
>
> We can of course combine the best of the both worlds and execute test suites from the command line on top of single request to the target node, similar to [discv5 tests ](https://github.com/ethereum/go-ethereum/blob/master/cmd/devp2p/internal/v5test/discv5tests.go)in geth.

This seems right. The CLI is already big, and I'm not keen to make it bigger, though. I think it is worth exploring if we can drive p2p commands through new json-rpc endpoints, in a reasonable way. Then the test suite driver would launch the dummy node and connect to its json-rpc in order to drive it, I imagine.

carver:
> You could do that by swapping this command in where the `rust/build` is currently in the `.circleci/config.yml`:

If you're feeling ambitious, you could submit a PR with a new parameter to the build job of the orb, like `parallel_jobs`. In this file: https://github.com/CircleCI-Public/rust-orb/blob/11318667c1d8cd99d8376b2d7f00b1d93ec09236/src/commands/build.yml

njgheorghita:
> This seems right. The CLI is already big, and I'm not keen to make it bigger, though. I think it is worth exploring if we can drive p2p commands through new json-rpc endpoints, in a reasonable way. Then the test suite driver would launch the dummy node and connect to its json-rpc in order to drive it, I imagine.

This seems like the approach Piper took in Alexandria, which I found super useful when I was using it. [Link to the API](https://ddht.readthedocs.io/en/latest/jsonrpc.html#common) - I'm not sure if this API needs to be replicated 1-for-1, but it seems like the kinda thing we want. I've already implemented `discv5_nodeInfo` and `discv5_routingTableInfo` [here](https://github.com/carver/trin/blob/master/trin-core/src/jsonrpc.rs#L248-L255)

ogenev:
> If you're feeling ambitious, you could submit a PR with a new parameter to the build job of the orb, like `parallel_jobs`. In this file: https://github.com/CircleCI-Public/rust-orb/blob/11318667c1d8cd99d8376b2d7f00b1d93ec09236/src/commands/build.yml

It looks like running 2 jobs in parallel did the trick (4 failed). I'm not sure if we should add `parallel_jobs`  parameter in rust circleci orbs, because the default value depends on the CPUs count and probably we don't want to explicitly set this parameter every time.

> This seems right. The CLI is already big, and I'm not keen to make it bigger, though. I think it is worth exploring if we can drive p2p commands through new json-rpc endpoints, in a reasonable way. Then the test suite driver would launch the dummy node and connect to its json-rpc in order to drive it, I imagine.

I was talking about the testing framework cli which may look something like:

`cargo run -p ethportal-peertest -- --target_node enr:ISQ...KJJc`

which will run the dummy node in the background and opens a communication channel in the terminal, so you can type commands to interact with the node, for example:

`send_ping`
`send_find_nodes`
`test_suite`

which is similar to what json-rpc is doing but will not require web3. Maybe the question is: _"Should we use json-rpc to communicate with the dummy node  and what are the alternatives?_"

ogenev:
@carver, @njgheorghita are there any specific reasons  why we are not using [this](https://github.com/paritytech/jsonrpc) json-rpc library in Trin?

carver:
> @carver, @njgheorghita are there any specific reasons why we are not using for example [this](https://github.com/paritytech/jsonrpc) json-rpc library in Trin?

No, it's definitely worth trying out to see if it meets our needs! Right now, we make heavy use of a wildcard matching, and I'm not clear if that lib supports it. But the feature isn't critical IMO.

trin-core/src/jsonrpc.rs
```diff
@@ -220,7 +220,15 @@ fn handle_request(
         })
         .to_string()),
         _ if obj.method.as_str().starts_with("discv5") => dispatch_portal_request(obj, portal_tx),
-        _ => dispatch_infura_request(obj, infura_url),
+        _ => match infura_url {
+            Some(url) => dispatch_infura_request(obj, &url),
+            None => Err(json!({
+                "jsonrpc": "2.0",
+                "id": obj.id,
+                "error": "Infura project id not provided!".to_string(),
```
carver, line 228: Could be nice to include the environment variable name in the error message so folks know how to resolve the problem.

ogenev:
> Yup, looks right to me, too. One reasonable approach is to reduce the number of parallel builds in the circle job to maybe 4 or 2. It's currently defaulting to the number of CPUs. You could do that by swapping this command in where the `rust/build` is currently in the `.circleci/config.yml`:
> 
> ```
> cargo build --jobs 4
> ```
> 
> ~ source https://stackoverflow.com/a/68031406/8412986
> 
> I wouldn't drop it down to `--jobs 1` without testing the higher parallelism, because I'm guessing that will severely slow the build.

This is not needed anymore as we moved to Ethereum organisation.

ethportal-peertest/src/events.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    async fn get_talk_req_reply(&self, request: &TalkRequest) -> Vec<u8> {
+        let reply = match self.overlay.process_one_request(request).await {
+            Ok(r) => Message::Response(r).to_bytes(),
+            Err(e) => {
+                error!("failed to process portal event: {}", e);
+                e.into_bytes()
+            }
+        };
+        reply
```
carver, line 88: If you like it:
```suggestion
        match self.overlay.process_one_request(request).await {
            Ok(r) => Message::Response(r).to_bytes(),
            Err(e) => {
                error!("failed to process portal event: {}", e);
                e.into_bytes()
            }
        }
```

carver (approved):
Were you able to see CircleCI complain about any clippy errors in `ethportal-peertest`? Just to really make sure it's included in linting.

ethportal-peertest/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "ethportal-peertest"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+trin-core = { path = "../trin-core" }
```
njgheorghita, line 9: nitpick - can you alphabetize these

ethportal-peertest/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "ethportal-peertest"
```
njgheorghita, line 2: nitpick - I'm not sure I love this name - but this is a minor subjective nitpick. I'm assuming with the `ethportal` prefix - you're implying that it's an implementation agnostic testing framework? And the `peertest` suffix is somewhat confusing for me. Some ideas off the top of my head (which tbh I'm not sure are any better) `portal-test-harness` /  `test-framework` / `test-harness` / `testing-node` ... again, totally subjective nitpick here, not a blocker for me at all.

ethportal-peertest/README.md
```diff
@@ -0,0 +1,10 @@
+# ethportal-peertest
+
+Run a portal network node that you want to test and pass node's Enr as a target node argument.
```
njgheorghita, line 3: I would like to see some expansion / more detail on how to use this. But, I totally think that will come naturally as the test harness develops.

njgheorghita (approved):

ethportal-peertest/src/cli.rs
```diff
@@ -0,0 +1,59 @@
...
+    #[structopt(
+        default_value(DEFAULT_WEB3_IPC_PATH),
+        short = "w3",
+        long = "web3_ipc_path",
+        help = "path to json-rpc socket address over IPC"
+    )]
+    pub web3_ipc_path: String,
+
+    #[structopt(
+        short = "tn",
```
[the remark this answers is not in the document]
ogenev: you are right,  the short arguments are always one letter, fixed this

ethportal-peertest/Cargo.toml
ogenev: The crate name was a product of [this](https://github.com/ethereum/trin/pull/79#discussion_r694191099) conversation. Happy to change it any time.

ethportal-peertest/src/cli.rs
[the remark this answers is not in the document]
ogenev: > Were you able to see CircleCI complain about any clippy errors in `ethportal-peertest`? Just to really make sure it's included in linting.

Good catch, changed the clippy command in CIrcleCI to include all workspace crates

ogenev:
> Were you able to see CircleCI complain about any clippy errors in `ethportal-peertest`? Just to really make sure it's included in linting.

Good catch, changed the clippy command in CIrcleCI to include all workspace crates.

carver:
> > Were you able to see CircleCI complain about any clippy errors in `ethportal-peertest`? Just to really make sure it's included in linting.
> 
> Good catch, changed the clippy command in CIrcleCI to include all workspace crates.

Ah, just noticed that build & test have the same problem. Maybe we should just add a new circle job that targets `ethportal-peertest`.

carver:
> > Ah, just noticed that build & test have the same problem. Maybe we should just add a new circle job that targets `ethportal-peertest`.
> 
> I think testing is working, I will address the build issue in #120. I think for now `cargo build --workspace` will do the trick.

Yup, I see now. 👍🏻 I saw it missing in the build, and must have just not looked carefully enough at the test list.

PR #79, opened 2021-08-19
add a bootstrap script that starts a dummy node and test client

victor-wei126:
I added a script that starts the test client, grabs its enr, then uses that to start testing with the dummy node. Here, I'm just assuming the test client is just an existing trin client from the codebase we have (specifically trin-state) since that is all we have to test against so far. Let me know if the test client is something different though.

### DIFF
--- a/devp2p/README.md
+++ b/devp2p/README.md
@@ -2,9 +2,15 @@
 
 Run a portal network node that you want to test and pass node's Enr as a target node.
 
+## Running tests
+To test a target client against a dummy peer, do:
 ```sh
 cd devp2p
 
-TRIN_DATA_PATH="/tmp/devp2p" RUST_LOG=debug cargo run -p devp2p -- --target_node enr:-IS4QBDHCSMoYoC5UziAwKSyTmMPrhMaEpaE52L8DDAkipqvZQe9fgLy2wVuuEJwO9l1KsYrRoFGCsNjylbd0CDNw60BgmlkgnY0gmlwhMCoXUSJc2VjcDI1NmsxoQJPAZUFErHK1DZYRTLjk3SCNgye9sS-MxoQI-gLiUdwc4N1ZHCCIyk
-
+./boot.sh
 ```
+This runs a shell script which:
+1. Starts up the test client (currently, it just runs: `cargo run -p trin-state`)
+2. Grabs the test client's ENR
+3. Runs the portal network testing suite, which starts up a dummy node 
+and passes the test client ENR to connect the nodes.
--- a/devp2p/README.md
+++ b/devp2p/README.md
@@ -1,0 +1,10 @@
+# devp2p
+
+Run a portal network node that you want to test and pass node's Enr as a target node.
+
+```sh
+cd devp2p
+
+TRIN_DATA_PATH="/tmp/devp2p" RUST_LOG=debug cargo run -p devp2p -- --target_node enr:-IS4QBDHCSMoYoC5UziAwKSyTmMPrhMaEpaE52L8DDAkipqvZQe9fgLy2wVuuEJwO9l1KsYrRoFGCsNjylbd0CDNw60BgmlkgnY0gmlwhMCoXUSJc2VjcDI1NmsxoQJPAZUFErHK1DZYRTLjk3SCNgye9sS-MxoQI-gLiUdwc4N1ZHCCIyk
+
+```
--- a/devp2p/src/cli.rs
+++ b/devp2p/src/cli.rs
@@ -1,0 +1,41 @@
+use std::env;
+use std::ffi::OsString;
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug, PartialEq)]
+#[structopt(
+    name = "trin-devp2p2",
+    version = "0.0.1",
+    about = "Testing framework for portal network peer-to-peer network calls"
+)]
+pub struct DummyConfig {
+    #[structopt(
+        use_delimiter = true,
+        short = "tn",
+        long = "target_node",
+        help = "Base64-encoded ENR's of the nodes under test"
+    )]
+    pub target_nodes: Vec<String>,
+}
+
+impl DummyConfig {
+    pub fn new() -> Self {
+        Self::new_from(env::args_os()).expect("Could not parse trin arguments")
+    }
+
+    pub fn new_from<I, T>(args: I) -> Result<Self, String>
+    where
+        I: Iterator<Item = T>,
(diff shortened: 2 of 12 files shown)


ogenev:
Hey @victor-wei126 , I think this script will work only if the target client is a `trin` node running on `ipc` for web3 transport, for example it will not work if the client under test runs with `http` web3 transport.

devp2p/README.md
```diff
@@ -2,9 +2,15 @@
...
+## Running tests
+To test a target client against a dummy peer, do:
 ```sh
 cd devp2p
 
-TRIN_DATA_PATH="/tmp/devp2p" RUST_LOG=debug cargo run -p devp2p -- --target_node enr:-IS4QBDHCSMoYoC5UziAwKSyTmMPrhMaEpaE52L8DDAkipqvZQe9fgLy2wVuuEJwO9l1KsYrRoFGCsNjylbd0CDNw60BgmlkgnY0gmlwhMCoXUSJc2VjcDI1NmsxoQJPAZUFErHK1DZYRTLjk3SCNgye9sS-MxoQI-gLiUdwc4N1ZHCCIyk
-
-```
\ No newline at end of file
+./boot.sh
```
ogenev, line 10: I think it will be better if we can support cross-client testing (not only `trin` nodes), which makes the changes you propose in the README not viable.

devp2p/src/dummy_protocol.rs
```diff
@@ -0,0 +1,95 @@
...
+
+const LISTEN_PORT: u16 = 9876;
+
+#[derive(Clone)]
+pub struct DummyProtocol {
+    pub discovery: Arc<Discovery>,
+    data_radius: U256,
+}
+
+impl DummyProtocol {
```
carver, line 21: I wonder if there's some way to refactor the current implementation so that we don't have all this duplicate dummy code. I'm not deep enough in it to see the best solution here, so this doesn't need to be a blocker if you don't come up with something quickly. I'm okay waiting to see how much of a pain maintaining it will be, and dealing with it then.

devp2p/src/main.rs
```diff
@@ -0,0 +1,56 @@
...
+            .collect();
+
+        // Test Ping target nodes
+        for enr in &target_enrs {
+            info!("Pinging {} on portal network", enr);
+            let ping_result = p2p
+                .send_ping(U256::from(u64::MAX), enr.clone())
+                .await
+                .unwrap();
+            info!("Portal network Ping result: {:?}", ping_result);
```
carver, line 31: Presumably the idea is to either panicking here or keep track of failures to display them at the end of the run. Is that your plan for a follow-up PR?

carver (commented):
The only thing I that seems important to agree on is the crate name. The rest are suggestions (some stronger than others).

devp2p/src/dummy_protocol.rs
ogenev: Yeah, this was just experimenting and prototyping some code, when https://github.com/carver/trin/pull/80 is merged, I will update it to use the default PortalProtocol struct

devp2p/src/main.rs
ogenev: I'm currently exploring different possibilities, this code is just to make sure it is working in the first place :)

devp2p/src/cli.rs
```diff
@@ -0,0 +1,41 @@
+use std::env;
+use std::ffi::OsString;
+use structopt::StructOpt;
+
+#[derive(StructOpt, Debug, PartialEq)]
+#[structopt(
+    name = "trin-devp2p2",
+    version = "0.0.1",
+    about = "Testing framework for portal network peer-to-peer network calls"
```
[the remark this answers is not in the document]
ogenev: I'm happy with `trin-peertest`, I'm assuming we are following the same crate/folder naming convention and set the folder name same as the crate name? It is funny because it looks strange to me to have dozen of folders starting with `trin-` in the repo but it is probably just me :smile:

devp2p/README.md
victor-wei126: I was thinking that we could start by making a testing framework just for trin, and then expand to cross-client, since that would probably be faster. But I'm not sure--what exactly would a cross-client testing framework entail (vs. a non cross client)?

victor-wei126:
>  I think this script will work only if the target client is a trin node running on ipc for web3 transport, for example it will not work if the client under test runs with http web3 transport.

Yeah, the script definitely needs a lot of tweaking to make it more robust. I didn't mean for it to be a final, explicit piece of the testing framework, just a quick hacky script that might help boost our developer workflow, probably could've just shared it with you. It seemed to me like a convenient way to test the testing framework, just getting everything up and connected in one fell swoop.

Come to think of it, it might not be best to even have it in the repo depending on where we go with the framework. What do you guys think?

devp2p/src/cli.rs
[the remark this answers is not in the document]
victor-wei126: `trin-peertest` sounds good to me as well. Minor thing, but I second @ogenev on this one, having `trin` at the start of every crate seems repetitive. Also, if we're designing the framework with interoperability in mind, I think it would be best to take out `trin` to avoid confusion.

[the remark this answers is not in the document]
carver: > `trin-peertest` sounds good to me as well. Minor thing, but I second @ogenev on this one, having `trin` at the start of every crate seems repetitive. Also, if we're designing the framework with interoperability in mind, I think it would be best to take out `trin` to avoid confusion.

Right, I'm new enough to rust that I don't know the conventions here. My thought is that maybe it's okay to call the folder just `state` but to name the crate `trin-state` (assuming that's the name you see in crates.io). With `peertest`, and the idea being to test multiple clients eventually, maybe it should be named `portal-peertest` or `ethportal-peertest`, or something like that.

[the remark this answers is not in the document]
ogenev: > My thought is that maybe it's okay to call the folder just `state` but to name the crate `trin-state` (assuming that's the name you see in crates.io). 
I'm thinking the same way, what seems important to me is the crate name in `Cargo.toml` (because of crates.io and collusion with other crate names). Folder naming is just personal preference, although most of the popular rust crates are naming folders same as the crates. 

> With `peertest`, and the idea being to test multiple clients eventually, maybe it should be named `portal-peertest` or `ethportal-peertest`, or something like that. Since it's a global name, it should stand well on its own, among _all_ rust crates.

Good point to have a global name for this crate, it is possible in the future even this to be in its own repo.

PR #80, opened 2021-08-21
Remove jsonrpc dependency from PortalnetProtocol struct and change default folder name

ogenev:
Sometimes, we may want to initialize a portal net protocol without jsonrpc support (for example when running tests or creating "dummy" node).

It will be also good if we can run multiple portal protocol instances simultaneously (for testing req/resp network calls). The current default db folder structure prevent this, because there is a db collusion between the different instances.

This PR removes jsonrpc dependency from PortalnetProtocol struct and adds last 8 chars from local base64 encoded Enr as a suffix to the default data application folder.

### DIFF
--- a/trin-core/src/portalnet/protocol.rs
+++ b/trin-core/src/portalnet/protocol.rs
@@ -73,8 +73,8 @@ pub struct PortalnetEvents {
 }
 
 pub struct JsonRpcHandler {
-    discovery: Arc<Discovery>,
-    jsonrpc_rx: mpsc::UnboundedReceiver<PortalEndpoint>,
+    pub discovery: Arc<Discovery>,
+    pub jsonrpc_rx: mpsc::UnboundedReceiver<PortalEndpoint>,
 }
 
 impl JsonRpcHandler {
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/protocol.rs
```diff
@@ -241,12 +232,12 @@ impl PortalnetProtocol {
             db,
         };
 
-        let rpc_handler = JsonRpcHandler {
-            discovery,
-            jsonrpc_rx,
+        let proto = Self {
+            discovery: discovery.clone(),
+            overlay: overlay.clone(),
```
carver, line 237: Was there anything important about the move of `let proto`? It's helpful to me as a reviewer when a PR avoids moving code if the move has no effect, to make the diff smaller and quicker to review.

trin-core/src/utils.rs
```diff
@@ -42,19 +43,25 @@ mod test {
...
+pub fn get_default_data_dir(local_enr: Enr) -> String {
     // Windows: C:\Users\Username\AppData\Roaming\Trin\data
     // macOS: ~/Library/Application Support/Trin
     // Unix-like: $HOME/.local/share/trin
 
-    match ProjectDirs::from("", "", "Trin") {
+    // Append last 8 enr base64 encoded chars to application dir name
+    let mut application_string = "Trin_".to_owned();
+    let len = &local_enr.to_base64().len();
+    let suffix = &local_enr.to_base64()[len - 8..];
```
carver, line 61: If I were new to trin, and found a directory named like this, I would expect it to be the first few characters of the as-displayed node-id. I understand if that feels like a tangent to you right now, and if so you want to skip it. In that case, maybe you could add a "good first issue" with that suggestion? Seems like a nice bite-size chunk for a newbie who wants to get started.

carver (approved):
LGTM!

trin-core/src/portalnet/protocol.rs
ogenev: That's right, I will refrain from such moves in the future.

PR #84, opened 2021-08-24
[WIP] implementation of offer, accept, store

victor-wei126:
The main issue i am having is deciding whether accept should be a request type or response type. I am leaning towards request, since that makes more sense to me. If it's a request, then node A sends `offer`, node B receives the request and responds, not with a `response` message, but calls the `send_accept()` method, which then sends the accept to A. Node A can then respond to this `request` msg by sending over the `store` response. The problem with this is we have to deal with the return type of the `process_one_request` method.

For the other strategy, we make accept a response type. I haven't explored this as much but at least I know we would have to add a new arm to the match statement to match an incoming response. I guess this seems kinda weird linguistically cuz we are matching requests, but now also responses. It would be helpful to have your guys insights on this.


trin-core/src/portalnet/protocol.rs
```diff
@@ -136,6 +136,7 @@ impl PortalnetEvents {
         }
 
         let request = match Message::from_bytes(talk_request.body()) {
+            // we would need to add an Ok(Message::Response(Response::Accept(r))) in order to accept Accept responses
```
carver, line 139: I think there's a whole conversation that needs to happen here. We don't really do anything with responses at all anywhere. So this comment can be removed and created as an issue to: design an API for decoding responses. The suggestion in this comment can be added to the body. I think there are some significant alternatives to consider.

trin-core/src/portalnet/protocol.rs
```diff
@@ -178,6 +179,16 @@ impl PortalnetEvents {
                 }
                 Err(e) => panic!("Unable to respond to FindContent: {}", e),
             },
+            // ACCEPT - in response to an offer msg, requests the data for given content-key
+            // this demonstrates the recipients willingness to store the data
+            Request::Offer(Offer { content_keys }) => {
+                // by default, we accept to store data
+                let connection_id = vec![9,9];
```
carver, line 186: Probably worth commenting that this `vec![9,9]` is just a placeholder until uTP is implemented and used to generate and send a connection ID.

carver (commented):
I'm realizing you're in the middle of updating this for the bit vector solution. Please ping me when you are ready for another review.

trin-core/src/portalnet/protocol.rs
victor-wei126: I originally had this comment in for thinking about how nodes would then respond to the `ACCEPT` message, but since we are dropping `STORE`, I don't need it. But I guess it is still an option to handle responses if there is no where else that does it.

> We don't really do anything with responses at all anywhere.

Yeah I was racking my head thinking about where the code to handle response messages like `FindContent` and `FindNodes` are. Has this not been implemented in Trin yet or am I blatantly missing that section?

victor-wei126:
> Under this logic, the STORE message is actually not necessary. The ACCEPT message can contain a connection_id that the offer-er can use for the uTP stream.

Since I am implementing these messages in the `trin-core` crate, this means that both the state and history networks (which is what we have in trin as of now) will be using this implementation. The STORE message is also being dropped for history, right? I noticed STORE is still in the history network spec, but I am guessing that it hasn't been updated yet.

jacobkaufmann:
I talked with @pipermerriam some about whether we should keep the `Store` RPC. In the existing data model, the only data that would fit within the UDP packet size are block headers. (If we decide that we want to support requests for individual transactions, then this could change.) For each block that becomes a part of chain history, the block header only makes up a small fraction of the total data for that block. However, I'm not sure whether block header data will be disproportionately requested and transmitted relative to block body data. If this is the case, then it could make sense to have a `Store` message for fast/lightweight transmission of header data.

The case against `Store`:

* Currently only supports transmission of header data
* Simplification of protocol messages
* Identical protocol messages to state network
  * Allows abstraction of overlay storage network

The case for `Store`:

* Support the fast(er) transmission of header data (and possibly other more granular data)

trin-core/src/portalnet/types.rs
```diff
@@ -219,6 +234,17 @@ pub struct FoundContent {
...
 
+#[derive(Debug, PartialEq, Clone, Encode, Decode)]
+pub struct Offer {
+    pub content_keys: Vec<u8>,
+}
+
+#[derive(Debug, PartialEq, Clone, Encode, Decode)]
+pub struct Accept {
+    pub connection_id: Vec<u8>,
+    pub content_keys: Vec<u8>
```
jacobkaufmann, line 245: For change to SSZ Bitlist

```suggestion
    pub content_keys: Vec<bool>
```

trin-core/src/portalnet/protocol.rs
carver: > Yeah I was racking my head thinking about where the code to handle response messages like `FindContent` and `FindNodes` are. Has this not been implemented in Trin yet or am I blatantly missing that section?

Nope, not really implemented yet. Discv5 returns the talkreq response as a slice of bytes, and right now we're not doing anything beyond returning that (we're not parsing the result at all). For example:

https://github.com/ethereum/trin/blob/5c5a1b3015451e402781fd349a7f3e296cf2639b/trin-core/src/portalnet/protocol.rs#L269-L282

victor-wei126:
With the utp implementation now complete, once an OFFER message is received, and an ACCEPT message is sent, 2 things happen.

1. On the side sending ACCEPT, it now waits and listens for an incoming uTP stream on (connection_id, node_id) pair, where `connection_id` is a randomly generated u16, and `node_id` is the NodeId of the node sending the OFFER msg.
2. On the side receiving ACCEPT, the node must initiate a uTP connection on the given `connection_id`, sending back the appropriate data as specified by `content_keys` bitlist.

For 1, a uTP syn packet will arrive back at the node, and `process_utp_request()` will be called. Since this is a new utp stream, line 435 of `utp.rs` will be executed. I believe we need to check here that the (connection_id, node_id) matches what is expected, which I don't think is currently handled so that's an issue that needs to be addressed.

For 2, the ACCEPT response must be handled by the node. As a sketch, I think the handler would call `connect()` to initiate uTP (on given connection_id) and sends a syn packet, waits for a response, then calls `write()` to write a message to the channel. In this case, the msg would be constructed based off of the `content_keys` bitlist. 

The question is where to write this code. Currently trin doesn't have an API to decode & handle response, as detailed in #94 and there are some possible architectures discussed in #102. This is my understanding of the problem.

KolbyML:
If anyone is wondering, I will be collaborating with Victor on this PR

trin-core/src/portalnet/mod.rs
```diff
@@ -63,6 +63,12 @@ impl ssz::Decode for U256 {
     }
 }
 
+// impl From<String> for U256 {
+//     fn from(string: String) -> Self {
+//         pub struct U256()
+//     }
+// }
+
```
jacobkaufmann, line 71: Looks like this was left by mistake. If unnecessary, let's remove.

jacobkaufmann:
I started to review this because review was requested, but it looks like there are still a lot of gaps and pseudocode in the comments, so I would like to know whether this is in a state ready for review, and if so, what aspects of the code are you looking for review on?

KolbyML:
Ok, I opened the PR for review. So I just wanted to start with a few things. Handing responses still is a problem that has to be handled. Victor started off with a stub function I just built the Accept offer off of. If you guys have any suggestions for a better way to handle this I am open, saying that passing in arguments not used by other messages isn't the cleanest solution.

Now that I am thinking about it maybe instead of a channel to tell if the node is connected or not we can do that within uTP itself. I can look into that, but it would probably be a follow-up PR, I just had that idea now so it is newly brewing in my mind.

I am now going to make another small PR because I realized that tasks might be waiting when messages are waiting to be sent..

trin-core/src/portalnet/utp.rs
```diff
@@ -414,56 +416,65 @@ impl UtpListener {
...
-                        } else if let Some(conn) =
-                            self.utp_connections.get_mut(&key_fn(2)).filter(f)
-                        {
-                            conn.state = ConnectionState::Disconnected;
-                        } else if let Some(conn) =
-                            self.utp_connections.get_mut(&key_fn(0)).filter(f)
-                        {
-                            conn.state = ConnectionState::Disconnected;
+                // Only handle packets if they are on our watchlist and have been negotiated
+                // I believe this is what is meant in the specs when they say after getting this message
```
jacobkaufmann, line 420: If we are going to reference a spec, then we should include a link.

trin-core/src/portalnet/overlay.rs
```diff
@@ -343,14 +494,63 @@ impl OverlayProtocol {
...
+                    .unwrap();
+                Ok(result)
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    // offer is sent in order to store content to k nodes with radii that contain content-id
+    // offer is also sent to nodes after FindContent (POKE)
+    pub async fn send_offer(
```
jacobkaufmann, line 520: Would be good to add `///` doc comment here. I know we don't do it for the similar `send_*` methods, but I think we should get in the habit of doing it for all public methods.

trin-core/src/portalnet/overlay.rs
```diff
@@ -138,9 +143,111 @@ impl OverlayProtocol {
...
+                }
+
+                let content_message = UtpAccept {
+                    message: content_vec,
+                };
+
+                let utp_listener = self.utp_listener.clone();
+                tokio::spawn(async move {
+                    while let Some(state) = rx.recv().await {
+                        if state == "connected".to_string() {
```
jacobkaufmann, line 213: We ought to use an enum here.

trin-core/src/portalnet/overlay.rs
```diff
@@ -138,9 +143,111 @@ impl OverlayProtocol {
...
+            }
+            // might not be a good idea to make a separate function because of what is needed
+            Response::Accept(Accept {
+                connection_id,
+                content_keys,
+            }) => {
+                let (tx, mut rx) = mpsc::unbounded_channel::<String>();
+
+                // Add connection_id we will be receiving to the listening list and the
+                // response number. The connection_id is + 1 because that is their id
```
jacobkaufmann, line 180: Even with the comment, I don't understand why we add 1 here.

trin-core/src/portalnet/overlay.rs
```diff
@@ -138,9 +143,111 @@ impl OverlayProtocol {
...
+            utp_listener,
         }
     }
 
+    pub async fn process_response(
+        &self,
+        response_vec: Vec<u8>,
+        enr: Enr,
+        content_keys_vec: Vec<Vec<u8>>,
+    ) -> Result<(), String> {
```
jacobkaufmann, line 155: We ought to return a non-`String` error type here. Instead, we can use `OverlayRequestError`.

trin-core/src/portalnet/events.rs
```diff
@@ -114,20 +112,93 @@ impl PortalnetEvents {
...
+    }
+
+    pub fn encode(&mut self) -> Vec<u8> {
+        let mut buf = vec![];
+        buf.extend_from_slice(&self.length.to_be_bytes());
+        buf.append(&mut self.payload);
+        buf
+    }
+
+    pub fn decode(bytes: &[u8]) -> UtpMessage {
```
jacobkaufmann, line 171: It seems that it would be more idiomatic to combine `decode` and `is_valid` so that `decode` returns a `Result`. Also would like to see some unit tests for `decode`.

trin-core/src/portalnet/overlay.rs
```diff
@@ -138,9 +143,111 @@ impl OverlayProtocol {
...
+                    .listening
+                    .insert(connection_id.clone() + 1, 8);
+
+                self.utp_listener
+                    .write()
+                    .await
+                    .connect(connection_id.clone(), enr.node_id(), tx)
+                    .await;
+
+                let mut content_vec: Vec<(Vec<u8>, Vec<u8>)> = vec![];
```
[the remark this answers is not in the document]
KolbyML: I was trying to think of a good name for it ty 😁

jacobkaufmann:
I took an initial pass and left some comments on some things that stood out to me. I will need to take a second pass after another look at the existing uTP code. It looks like a good start, but I'm just not confident enough yet in my understanding of that to give further review. 

Further general comments:

* Additional comments for this code would go a long way for contributors unfamiliar with uTP.
* Prefer the new `read_with_warn` and `write_with_warn` methods.

trin-core/src/portalnet/overlay.rs
KolbyML: We aren't going to actually use this till we are done testing, I have talked with some of the Fluffy team. I also have to double-check if this is the one that should be plus one.

Anyways we will probably add both the conn_recv_id and the conn_send_id. The conn_recv_id is used on the first packet for the handshake. After the first packet in the handshake, they both use conn_send_id or their conn_send_id.

One will be receiving packets with + 1 as each node has a different conn_send_id with a 1 increment difference

trin-core/src/portalnet/events.rs
```diff
@@ -114,20 +112,93 @@ impl PortalnetEvents {
...
+                match self
+                    .utp_listener
+                    .read()
+                    .await
+                    .listening
+                    .get(&conn.conn_id_recv)
+                {
+                    Some(message_type) => match message_type {
+                        // 8 is the Offer/Accept uTP stream handler
+                        8 => {
```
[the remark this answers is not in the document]
KolbyML: Sure we can this is an implementation-specific internal tracker I guess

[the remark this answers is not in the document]
KolbyML: But it is always good to have it as a constant

trin-core/src/portalnet/events.rs
```diff
@@ -114,20 +112,93 @@ impl PortalnetEvents {
...
+
+                // message was successfully handled, remove it from the buffer
+                conn.recv_data_stream = received_stream[(message.len() as usize)..].to_owned();
             }
+        }
+    }
+}
 
-            let message_len = u32::from_be_bytes(received_stream[0..4].try_into().unwrap());
+pub struct UtpMessage {
```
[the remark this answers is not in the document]
KolbyML: We can do that, I wonder what the file should be called since it shouldn't go into utp.rs and it could be confused for something uTP implementation wise. I will give it some thought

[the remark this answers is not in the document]
ogenev: I think it is probably worth having a separate folder or why not a crate for files related to the utp implementation.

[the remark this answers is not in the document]
KolbyML: @ogenev UtpMessage has nothing to do with the uTP implementation. I mean we can move utp.rs to a crate. UtpMessage is just for handling uTP messages on Trin's side.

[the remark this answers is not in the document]
KolbyML: Anyways I will make a PR for that I guess because Sam also made that suggestion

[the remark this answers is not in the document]
KolbyML: Giving it some thought I will just make a folder. The uTP implementation is heavily tied to Discv5. Also this UtpMessage stuff isn't uTP implementation it is just a impl for cleaning up the handling of streams for Trin so it would probably make the most sense to just have it in a folder. I still need to think about what to call the new file uTP-handling I honestly have no idea.

trin-core/src/portalnet/overlay.rs
victor-wei126: This confused me as well until I read the uTP spec where they describe connection_id:

> Each socket has one connection ID for sending packets and a different connection ID for receiving packets. The endpoint initiating the connection decides which ID to use, and the return path has the same ID + 1.

I think what Kolby is trying to say is that node A (the offerer) receives the `connection_id` from node B (the acceptor) and uses this as its `conn_send_id`. For receiving uTP packets, it will listen on a socket with `id+1`, ie it's `conn_recv_id`.

victor-wei126: Looking over the code again, I think there is some difference between original uTP spec and current implementation. I think there is some misalignment in what we are using as id for send/recv. ie, a node can have recv id as just `id` and send as `id+1`, or it can have the reverse, it doesn't really matter, but we just have to make sure it is consistent.

trin-core/src/utp/mod.rs
```diff
@@ -0,0 +1,3 @@
+#[allow(clippy::module_inception)]
+pub mod utp;
```
carver, line 2: Certainly, we want to rethink any choices that cause clippy warnings. `utp` in utp does seem like a surprising naming choice. Can we do better so that we don't need this clippy silence?

KolbyML: Ok I refactored utp so this is no longer an issue. I want to rework the utp implementation so it doesn't rely on and async as much where it doesn't need to.

trin-core/src/utp/trin_helpers.rs
```diff
@@ -0,0 +1,107 @@
+// The code here isn't affiliated with stream and isn't part of the uTP implementation
+// this code is used in handling the uTP responses handled in events.rs.
+// just a clarification. It is just for handling the uTP stream higher up what stream releases.
```
carver, line 3: This line is a little hard to follow. Probably don't need "just a clarification", and I'm not sure what the phrase "higher up what stream releases" is getting at.

trin-core/src/utp/trin_helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn encode(&mut self) -> Vec<u8> {
+        let mut buf = vec![];
+        buf.extend_from_slice(&self.length.to_be_bytes());
+        buf.append(&mut self.payload);
+        buf
+    }
+
+    pub fn decode(bytes: &[u8]) -> Result<UtpMessage, String> {
+        if bytes.len() >= 4 {
+            let len = u32::from_be_bytes([bytes[0], bytes[1], bytes[2], bytes[3]]);
```
carver, line 30: nitpick: Seems like there ought to be some way to pass in some equivalent of `bytes[0..4]` instead of manually entering every item into an array.

trin-core/src/utp/trin_helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+            if len + 4 <= bytes.len() as u32 {
+                let mut payload_vec: Vec<u8> = vec![];
+                payload_vec.extend_from_slice(&bytes[4..(len as usize + 4)]);
```
carver, line 33: It's valid to just have ignored garbage at the end of the payload, past the stated length?

trin-core/src/utp/trin_helpers.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn test_too_short_message() {
+        let buf = [0x00, 0x00, 0x00, 0x03, 0x09, 0x0B];
+
+        let message = UtpMessage::decode(&buf[..]);
+        assert!(message.is_err());
+    }
+
+    #[test]
+    fn test_message_decode() {
+        // Added an extra byte since we will be parsing a vector
```
carver, line 83: What does this mean? I still don't understand when/why a payload is longer than the stated length.

trin-core/src/utp/stream.rs
```diff
@@ -395,11 +84,19 @@ impl UtpListener {
             Ok(packet) => {
                 let connection_id = packet.connection_id();
 
+                // Only handle packets if they are on our watchlist and have been negotiated
+                // I believe this is what is meant in the specs when they say after getting this message
+                // Listen for a response
+                // [discv5-utp]: https://github.com/ethereum/portal-network-specs/blob/master/discv5-utp.md?plain=1#L26
+
+                // todo: uncomment this at a future date, as this makes it harder to test with other
+                // clients for the time being
```
carver, line 93: It would be helpful to include the trigger here, instead of just "later". Does this mean: the other clients have implemented utp differently, so we can switch when the change the implementation somehow? What needs to change in those clients? How can we confirm/test that that has happened? Otherwise, this will never change.

Really, my intuition here is to not put code in comments, and instead open a second WIP PR with these changes, and a description that answers the above questions.

trin-core/src/portalnet/events.rs
```diff
@@ -110,20 +112,49 @@ impl PortalnetEvents {
...
+                        _ => warn!("uTP listening HashMap doesn't have uTP stream message type"),
+                    }
 
-                // Removed the handled data from the buffer
-                conn.recv_data_stream = received_stream[((message_len + 4) as usize)..].to_owned();
+                    // message was successfully handled, remove it from the buffer
+                    conn.recv_data_stream = received_stream[(message.len() as usize)..].to_owned();
+                }
+                // Ignore messages which are too short
+                _ => {}
```
carver, line 151: Maybe a `debug!`, while this code is new? We'll probably want to remove it eventually, once we see that it's not popping up constantly.

trin-core/src/portalnet/events.rs
```diff
@@ -110,20 +112,49 @@ impl PortalnetEvents {
...
+                        .utp_listener
+                        .read_with_warn()
+                        .await
+                        .listening
+                        .get(&conn.conn_id_recv)
+                    {
+                        Some(message_type) => match message_type {
+                            UtpMessageId::OfferAcceptStream => {
+                                let payload = UtpAccept::from_ssz_bytes(&message.payload[..])
+                                    .unwrap()
```
carver, line 136: Don't want to panic on malformed input from peers. Should handle & log it.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1385,3 +1447,7 @@ mod tests {
         };
     }
 }
+
+fn should_store(_key: &Vec<u8>) -> bool {
+    return true;
+}
```
carver, line 1453: If there's no github issue to implement this, could you create one?

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+    }
+}
+
+#[derive(Debug)]
+pub struct PacketHeader {
+    type_ver: u8,
+    extension: u8,
+    connection_id: u16,
+    // This time is in microseconds
+    timestamp: u32,
```
carver, line 110: A 32-bit timestamp in microseconds can only count up to ~4200 seconds, so this could really benefit from an explanation about why rollover isn't a concern.

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+        PacketHeader::decode(&self.0[0..20])
+    }
+
+    pub fn type_(&self) -> PacketType {
+        self.get_header().type_()
+    }
+
+    pub fn version(&self) -> u8 {
+        self.get_header().version()
+    }
```
carver, line 233: Decoding the header every time we want a header field seems like it might be wasteful, but I'm okay kicking that can down the road. It seems like this layout gives us plenty of flexibility to resolve that later, easily.

carver (approved):
This is too big for me to feel like I deeply understand all of it, or carefully reviewed all of it, but I took a pass at it.

Feel free to make the call about whether it's ready to merge 👍🏻

trin-core/src/utp/packets.rs
KolbyML: The actual time doesn't matter this is just for congestion control. Which is generated from get_time()

trin-core/src/portalnet/overlay_service.rs
KolbyML: This is implemented in storage.rs, but storage.rs isn't implemented into the codebase yet.

trin-core/src/utp/trin_helpers.rs
KolbyML: Well, we will get the messages as it comes in, we might not end up using this since it isn't agreed upon how we will handle the stream yet, but the stream could be endless if they keep writing data to us. I like this approach because it means we could send multiple chunks of data.

I think we should have a length prefix, but we might just end up trying to decode whatever we have on close, it really isn't decided yet.

trin-core/src/utp/trin_helpers.rs
KolbyML: I wish 😁. That was actually the first thing I tried. We can either do this or I forget what other command, but I believe this ends up being shorter, I would have to check again.

carver: How about `try_into()`? Something like:

```suggestion
            let len = u32::from_be_bytes(bytes[0..4].try_into().unwrap()]);
```

trin-core/src/utp/packets.rs
carver: Ah yes, I see what's happening now in `get_time()`. Ok, well I will create a separate issue on this discussion, since this PR is just a refactor and doesn't really touch the logic about timestamps. Key issues:
- Calling that thing a timestamp is very confusing. It's a reference counter that rolls over every ~hour, which just feels very different. So maybe we should name it `usec_counter` or something like that.
- Calculating the time difference after a rollover looks wrong to me. Current time (rolled over) minus previous time (before rollover) is not equal to `previous_time - current_time`. Instead it's `U32_MAX - previous_time + current_time`.

KolbyML: That is what it is called in the spec. Also to be fair it still is technically a timestamp

KolbyML: Also, I  moved the refactor to #226 with some other changes because this PR is large enough and it isn't really related to this PR.

carver: > Also to be fair it still is technically a timestamp

I disagree, but #227 seems like the right place to hash that out. :)

trin-core/src/utp/trin_helpers.rs
KolbyML: Ok I changed it to that ty

ogenev (approved):
Haven't reviewed it in detail but I think it is good enough to get merged and iterate :+1:

Issue #96, opened 2021-09-07
Support JSON-RPC params member


ogenev:
Current json-rpc implementation does not support passing parameter values to the request object.
To provide an adequate json-rpc API, we want to support passing parameters which will used during the invocation of the methods. 

Full `JSON-RPC 2.0` specification is provided [here](https://www.jsonrpc.org/specification).

PR #97, opened 2021-09-07
Add params member to JSON-RPC request

ogenev:
Adds support for parsing parameters for the JSON-RPC call.

Fixes https://github.com/ethereum/trin/issues/96

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,7 +138,7 @@ dependencies = [
  "lazy_static 1.4.0",
  "lazycell",
  "peeking_take_while",
- "proc-macro2 1.0.26",
+ "proc-macro2 1.0.29",
  "quote 1.0.9",
  "regex 1.5.4",
  "rustc-hash",
@@ -831,7 +831,7 @@ checksum = "c54913bae956fb8df7f4dc6fc90362aa72e69148e3f39041fbe8742d21e0ac57"
 dependencies = [
  "autocfg 1.0.1",
  "proc-macro-hack",
- "proc-macro2 1.0.26",
+ "proc-macro2 1.0.29",
  "quote 1.0.9",
  "syn 1.0.72",
 ]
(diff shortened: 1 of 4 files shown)


njgheorghita (approved):
Looks good! I love the test - but it seems like a test that could benefit from [parameterization](https://docs.rs/rstest/0.11.0/rstest/#creating-parametrized-tests). I'm just bringing it up now since I've run into a couple similar situations on my local branches. From what I can tell, [rstest](https://docs.rs/rstest/0.11.0/rstest/) is probably the best crate to use for various testing utils. Even if it's not necessary for this test, it seems like a crate we'll want to use in other places as the project grows.

ogenev:
> Looks good! I love the test - but it seems like a test that could benefit from [parameterization](https://docs.rs/rstest/0.11.0/rstest/#creating-parametrized-tests). I'm just bringing it up now since I've run into a couple similar situations on my local branches. From what I can tell, [rstest](https://docs.rs/rstest/0.11.0/rstest/) is probably the best crate to use for various testing utils. Even if it's not necessary for this test, it seems like a crate we'll want to use in other places as the project grows.

Let me tell you a secret, I stole this test from one json-rpc crate :smile:  Good to know about `rstest`, I tried it, but in this case, it requires to add more boilerplate code which I think will look confusing.

njgheorghita:
> I tried it, but in this case, it requires to add more boilerplate code which I think will look confusing.

Fair enough, but I have a feeling like we'll need it sooner or later. Also, it's often worthwhile to add some code complexity if it means improved / more granular tests. Obviously there's a limit somewhere. I don't think it's a blocker for merging this, but you might think about leaving a comment that these tests should be expanded at some point

ogenev:
> Fair enough, but I have a feeling like we'll need it sooner or later. Also, it's often worthwhile to add some code complexity if it means improved / more granular tests. Obviously there's a limit somewhere. I don't think it's a blocker for merging this, but you might think about leaving a comment that these tests should be expanded at some point

Probably you are right, updated the test with parametrization.

njgheorghita:
@ogenev Looks great! Thanks for updating those tests. lgtm!

PR #99, opened 2021-09-08
Fix failing tests on master

ogenev:
By removing `default-members` section from the root Cargo.toml, `cargo test` in CircleCI is not able to detect all the tests in the crates.

This PR adds explicit `cargo test --workspace` command which will run tests on all workspace members in CIrcleCI pipeline. It also fixes some failing tests on master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,8 +38,9 @@ jobs:
             - rust/build:
                 release: false
                 with_cache: false
-            - rust/test:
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
Great catch, thanks! lgtm

njgheorghita:
Going ahead and merging this for now, since it's pretty useful to develop against locally

PR #100, opened 2021-09-08
State network routing table distance function

njgheorghita:
Implementation of custom distance function for state network routing table.


trin-state/src/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+// 2 ** 255
+const MID: [u8; 77] = [
+    53, 55, 56, 57, 54, 48, 52, 52, 54, 49, 56, 54, 53, 56, 48, 57, 55, 55, 49, 49, 55, 56, 53, 52,
+    57, 50, 53, 48, 52, 51, 52, 51, 57, 53, 51, 57, 50, 54, 54, 51, 52, 57, 57, 50, 51, 51, 50, 56,
+    50, 48, 50, 56, 50, 48, 49, 57, 55, 50, 56, 55, 57, 50, 48, 48, 51, 57, 53, 54, 53, 54, 52, 56,
+    49, 57, 57, 54, 56,
+];
+
+// distance function as defined at...
+// https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout
```
ogenev, line 21: @pipermerriam Shall we change the distance function in this document to use the modulus instead of the reminder, like it is implemented below?

ogenev (approved):
LGTM :rocket:

trin-state/src/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+// todo: measure performance & optimize if necessary
+pub fn distance(node_id: BigInt, content_id: BigInt) -> BigInt {
```
jacobkaufmann, line 23: Given that the rest of the codebase uses `UInt`, can we implement `Into<BigInt>` for a wrapped `UInt` (assuming it does not already exist) and then return a `UInt` from this function?

Also, to use a custom distance function in the underlying routing table, we need to make a change to the `discv5` library because it strictly uses [XOR](https://github.com/sigp/discv5/blob/master/src/kbucket/key.rs#L81). This is something I was going to take on, but it was de-prioritized. I'm happy to take that up, but probably not immediately. I'm not sure whether there is something else blocked on this.

There is an existing distance function test in the `discv5` repo for unidirectionality, which IIRC this custom distance function does not satisfy. I'm not sure whether this is significant, but I thought it was worth mentioning.

pipermerriam: I also *think* that it would be *ok* for us to duplicate the base routing table implementation if we want to stay hands-off on the base protocol.  The python implementation comes in at 374 SLOC which isn't that large.  I do get some unease at *polluting* the simple base APIs used to power the base protocol with the things we need for our overlay protocols.

trin-core/src/utils.rs
```diff
@@ -16,6 +16,30 @@ pub fn xor_two_values(first: &[u8], second: &[u8]) -> Vec<u8> {
...
+    let len = &local_enr.to_base64().len();
+    let suffix = &local_enr.to_base64()[len - 8..];
+    application_string.push_str(suffix);
+
+    match ProjectDirs::from("", "", &application_string) {
+        Some(proj_dirs) => proj_dirs.data_local_dir().to_str().unwrap().to_string(),
+        None => panic!("Unable to find data directory"),
+    }
+}
+
```
carver, line 42: Why was this moved? Makes it tougher to see what was changed.

trin-state/src/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+    let modulo = BigInt::parse_bytes(&MODULO, 10).unwrap();
+    let mid = BigInt::parse_bytes(&MID, 10).unwrap();
```
carver, line 25: There's probably some low-hanging-fruit to do this once, statically. But I'm not worried about that at the moment.

carver (approved):
Just gave a quick once-over. Nothing jumped out at me.

trin-core/src/utils.rs
njgheorghita: Imo - tests belong at the bottom of the file. Not sure if that's convention or not in Rust? Happy to switch it back tho

trin-state/src/utils.rs
njgheorghita: AFAIK the ["canonical" spec](https://github.com/ethereum/stateless-ethereum-specs/pull/70/files#diff-5b6bb1c3a7fc540f6bc7fc30ba3c1d6debce52506d16fe5364edd464a42537ecR118) python code is technically correct. It's just how modulus is implemented in various programming languages that requires the extra handling in Rust. https://stackoverflow.com/a/31210691

trin-state/src/utils.rs
njgheorghita: > Given that the rest of the codebase uses UInt, can we implement Into<BigInt> for a wrapped UInt (assuming it does not already exist) and then return a UInt from this function?

I could be missing something, but my reason for using `BigInt` in the function signature here was b/c we will pretty much always expect to have `node_id`s (256 bits) and `content_id`s (256 bits) and `distance`s greater than 128 bits. Since `u128` is the largest integer rust supports, I'm not sure it makes sense to return `uint`. One option I considered was accepting and returning `Vec[u8; 32]` types. And handling the conversion inside the `distance()` function. I'm not sure which is the better option tbh, but I went with `BigInt` since it seemed more appropriate to return an int from a function where we're calculating the distance. 

I guess a compromise might be something like `distance(node_id: Vec[u8;32], content_id: Vec[u8;32]) -> BigInt`? Which actually makes sense to me now that I think about it...

jacobkaufmann: Apologies for the miscommunication here. I meant `U256` and not `UInt`. Distance is always a positive (unsigned) value. The `node_id` and `content_id` are unsigned 256-bit values. This is why I think it makes sense to take `U256` for input and return `U256` for output.

trin-core/src/utils.rs
carver: Ah yes, I missed that the tests were basically what moved. 👍🏻 I don't know the standards that well either, but tests at the bottom is the only way I've seen it, too.

PR #101, opened 2021-09-09
Decouple Discovery struct from PortalProtocol

ogenev:
Initialize Discovery protocol outside of PortalProtocol struct and pass it as parameter.


njgheorghita (approved):
Cool, yeah as I've been getting into the overlay network implementation, I can see the benefit of splitting this out.

PR #103, opened 2021-09-10
Overlay network refactoring

### LINKED ISSUE #93: Subnetwork overlay protocols implementation
- How are the subnetwork overlay discv5 protocols instantiated?
- What responsibilities does the orchestrator have, and what is needed in each subnetwork initialization to kick off their overlay networks?
- If a user runs a subnetwork as a solo binary, will it still work?

ogenev:
This introduces some new abstractions for the overlay networks:

```rust
// Base Discovery layer
pub struct Discovery {
    /// discv5 protocol
    pub discv5: Discv5,
    /// Indicates if the discv5 service has been started
    pub started: bool,
   /// listen port for the discovery protocol
    pub listen_socket: SocketAddr,
}

// Overlay layer, should be used by every network
pub struct OverlayProtocol {
    // Contains discovery layer
    pub discovery: Arc<RwLock<Discovery>>,
    // The data radius of the local node
    pub data_radius: Arc<RwLock<U256>>,
    // The overlay routing table of the local node.
    pub kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
}

// Network specific layers, currently contain only the overlay protocol
pub struct HistoryProtocol {
    pub overlay: Arc<OverlayProtocol>,
}

pub struct StateProtocol {
    pub overlay: Arc<OverlayProtocol>,
}
```
**Other changes:**

- Creates one Discovery struct as `Arc<RwLock<Discovery>>`: one writer, many readers and passes it as argument to every network.
- Creates one database instance and passes it as argument to every network.
- Spawns separate thread for the JSON-RPC handler

Both networks runs concurrently right now in the main package. I just commented out the event handler for the history network, we need to refactor this to handle the events from both networks concurrently.

**Not included in this PR:**
- Refactor events handlers to work with both networks concurrently.

Fixes #93

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,9 +6,9 @@ use tokio::sync::mpsc;
 use trin_core::cli::TrinConfig;
 use trin_core::jsonrpc::launch_jsonrpc_server;
 use trin_core::portalnet::discovery::Discovery;
+use trin_core::portalnet::overlay::StateProtocol;
 use trin_core::portalnet::protocol::{
-    HistoryNetworkEndpoint, JsonRpcHandler, PortalEndpoint, PortalnetConfig, PortalnetProtocol,
-    StateNetworkEndpoint,
+    HistoryNetworkEndpoint, JsonRpcHandler, PortalEndpoint, PortalnetConfig, StateNetworkEndpoint,
 };
 
 #[tokio::main]
@@ -81,12 +81,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             portalnet_config.bootnode_enrs
         );
 
-        let (mut p2p, events) = PortalnetProtocol::new(discovery, portalnet_config)
+        let (mut p2p, events) = StateProtocol::new(discovery, portalnet_config)
             .await
             .unwrap();
 
         let rpc_handler = JsonRpcHandler {
-            discovery: p2p.discovery.clone(),
+            discovery: p2p.overlay.discovery.clone(),
             jsonrpc_rx,
             state_tx,
             history_tx,
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,15 +1,24 @@
 use std::env;
 
-use log::info;
+use log::{debug, error, info};
 use tokio::sync::mpsc;
+use tokio::sync::RwLock;
 
-use trin_core::cli::TrinConfig;
-use trin_core::jsonrpc::launch_jsonrpc_server;
-use trin_core::portalnet::discovery::Discovery;
-use trin_core::portalnet::overlay::StateProtocol;
-use trin_core::portalnet::protocol::{
-    HistoryNetworkEndpoint, JsonRpcHandler, PortalEndpoint, PortalnetConfig, StateNetworkEndpoint,
+use std::sync::Arc;
+use trin_core::{
+    cli::{TrinConfig, HISTORY_NETWORK, STATE_NETWORK},
+    jsonrpc::launch_jsonrpc_server,
+    portalnet::{
+        discovery::Discovery,
+        protocol::{
+            HistoryNetworkEndpoint, JsonRpcHandler, PortalEndpoint, PortalnetConfig,
+            StateNetworkEndpoint,
+        },
+    },
+    utils::setup_overlay_db,
 };
+use trin_history::network::HistoryNetwork;
+use trin_state::network::StateNetwork;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -42,11 +51,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         ..Default::default()
     };
 
-    let mut discovery = Discovery::new(portalnet_config.clone()).unwrap();
-    discovery.start().await.unwrap();
+    // Initialize base discovery protocol
+    let discovery = Arc::new(RwLock::new(
+        Discovery::new(portalnet_config.clone()).unwrap(),
+    ));
+    discovery.write().await.start().await.unwrap();
 
-    // Initialize state sub-network, if selected
-    let (state_tx, state_handler) = if trin_config.networks.iter().any(|val| val == "state") {
+    // Initialize state sub-network jsonrpc handler, if selected
+    let (state_tx, state_handler) = if trin_config.networks.iter().any(|val| val == STATE_NETWORK) {
         let (state_tx, state_rx) = mpsc::unbounded_channel::<StateNetworkEndpoint>();
         let state_handler = match trin_state::initialize(state_rx) {
             Ok(val) => val,
@@ -57,8 +69,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         (None, None)
     };
 
-    // Initialize chain history sub-network, if selected
-    let (history_tx, history_handler) = if trin_config.networks.iter().any(|val| val == "history") {
+    // Initialize chain history sub-network jsonrpc handler, if selected
+    let (history_tx, history_handler) = if trin_config
+        .networks
+        .iter()
+        .any(|val| val == HISTORY_NETWORK)
+    {
         let (history_tx, history_rx) = mpsc::unbounded_channel::<HistoryNetworkEndpoint>();
         let history_handler = match trin_history::initialize(history_rx) {
             Ok(val) => val,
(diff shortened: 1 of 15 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -152,20 +142,227 @@ impl Overlay {
...
+        let overlay = OverlayProtocol {
+            discovery: discovery.clone(),
+            data_radius,
+            kbuckets,
+        };
+
+        let overlay = Arc::new(overlay);
+
+        let utp_listener = UtpListener {
+            discovery: discovery.clone(),
```
njgheorghita, line 256: With this setup, we're cloning a lot of arcs - which seems fine, but I'm just curious how you feel about being more explicit in our arc variable naming? ie. `discovery_arc` / `overlay_arc`. I don't love it, but it's hard to keep track of all the arc's floating about, and we want to avoid accidentally cloning the base datatype. This isn't really a suggestion, just thinking out loud. It's reasonable that the compiler helps us out with this so we don't have to worry about being so explicit

njgheorghita (changes requested):
Great work! I was taking a much more complicated approach in my implementation, creating message channels between base discovery & overlay networks. I do prefer your implementation b/c it avoids a lot of the complexity. I don't have a great intuition for Rust of when to go for channels or r/w lock. But it seems like the lock will handle this situation just fine. Just a couple of a suggestions to address.

trin-core/src/portalnet/overlay.rs
```diff
@@ -59,59 +67,40 @@ impl Default for Config {
...
-    local_enr: Arc<RwLock<Enr>>,
+pub struct OverlayProtocol {
+    pub discovery: Arc<RwLock<Discovery>>,
     // The data radius of the local node.
-    data_radius: Arc<RwLock<U256>>,
-    // The routing table of the local node.
-    kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
+    pub data_radius: Arc<RwLock<U256>>,
+    // The overlay routing table of the local node.
+    pub kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
```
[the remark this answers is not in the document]
ogenev: We have access to the liveness/maintenance methods (see [here](https://github.com/sigp/discv5/blob/10247bbd299227fef20233f2f5a8de9780de09ac/src/kbucket/bucket.rs#L238)) but we need our own logic to manage them.

trin-core/src/portalnet/protocol.rs
```diff
@@ -309,103 +305,3 @@ impl PortalnetEvents {
         }
     }
 }
-
-impl PortalnetProtocol {
```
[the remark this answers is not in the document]
ogenev: Yes, the idea for the overlay protocol is that it will handle all requests from the overlay networks (state, history) and dispatch them to the discv5 protocol TalkReq.
So the implication is that every network should implement the overlay protocol and the overlay protocol is where we can encapsulate the logic for handling common network requests/responses. 

If there is something specific to a network that we don't want to implement it in the overlay protocol, we can encapsulate it on the network level aka `struct StateProtocol`, `struct HistoryProtocol`. Probably it makes sense to differentiate between 'protocols' and 'networks' and rename those struct to `StateNetwork`, `HistoryNetwork` for example...

trin-core/src/utils.rs
```diff
@@ -40,6 +41,13 @@ pub fn get_default_data_dir(local_enr: Enr) -> String {
     }
 }
 
+pub fn setup_overlay_db(local_enr: Enr) -> DB {
+    let data_path = get_data_dir(local_enr);
+    let mut db_opts = Options::default();
+    db_opts.create_if_missing(true);
+    DB::open(&db_opts, data_path).unwrap()
```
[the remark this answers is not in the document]
ogenev: I'm trying to resist to error handle the database stuff for now, because there is a PR in progress for the DB layer and we will have to refactor this anyway really soon..

trin-core/src/portalnet/overlay.rs
ogenev: I'm not sure what are the Rust's best practices here..., but I think the type checking does a great job and it will not allow us to pass base datatype if it is not expected...

src/main.rs
```diff
@@ -42,10 +50,13 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    discovery.start().await.unwrap();
+    // Initialize base discovery protocol
+    let discovery = Arc::new(RwLock::new(
+        Discovery::new(portalnet_config.clone()).unwrap(),
+    ));
+    discovery.write().await.start().await.unwrap();
 
-    // Initialize state sub-network, if selected
+    // Initialize state sub-network jsonrpc handler, if selected
     let (state_tx, state_handler) = if trin_config.networks.iter().any(|val| val == "state") {
```
njgheorghita, line 60: I'd recommend using the new constants for `"state"` here and `"history"` just below

src/main.rs
```diff
@@ -71,46 +82,102 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            .expect("failed to pause until ctrl-c");
-    })
-    .await
-    .unwrap();
+    debug!("Selected networks to spawn: {:?}", trin_config.networks);
+
+    for network in trin_config.networks.iter() {
+        match network.as_str() {
+            STATE_NETWORK => {
+                let discovery_state = discovery.clone();
```
njgheorghita, line 118: Sorry for being so nitpicky - but while I do think these are an improvement over `discovery_1`. It's still pretty confusing imo. I understand the implication of this naming ("**discovery** for the **state** network") - but if I was brand new reading this, I'd be pretty confused as I'd interpret it more like this is the "current state of discovery", or  this is the database state for `db_state`,  and below is the "history of discovery". IDK, maybe I'm being a little paranoid, but I always err on the side of readability vs. small names. Something like `state_subnet_discovery` is quite explicit.  I do prefer explicit and obvious names, especially when we're working on a project with many different contributors jumping in and out of the codebase.

I'm definitely open to hear pushback though.

trin-core/src/portalnet/protocol.rs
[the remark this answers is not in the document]
njgheorghita: That sounds like a great idea, if you don't mind tackling it. I'd also recommend a quick comment above the struct definitions giving a brief explanation of the data types and their responsibilities

njgheorghita (approved):
Looks great! Just a couple of suggestions. Also, adding a "Fixes #103" to the pr description will automatically close that issue after this gets merged, but of course you can always do that manually if you prefer.

src/main.rs
ogenev: Thinking about this now, we can also make the arc cloning more readable and explicit, for example, instead of `discovery.clone()`, we can say `Arc::clone(&discovery)`, how do you feel about this? Reference: [link](https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references)

For the naming, probably `state_network_discovery` makes more sense?

njgheorghita: :+1: to both of those!

trin-core/src/portalnet/overlay.rs
```diff
@@ -152,20 +145,229 @@ impl Overlay {
...
+    pub overlay: Arc<OverlayProtocol>,
+}
+
+/// State network layer on top of the overlay protocol. Encapsulates state network specific data and logic.
+#[derive(Clone)]
+pub struct StateNetwork {
+    pub overlay: Arc<OverlayProtocol>,
+}
+
+impl StateNetwork {
```
carver, line 227: I was still expecting these definitions to live in the respective crates, like `trin-state`. It feels backwards if we have to define every possible overlay definition in the core package, instead of only defining the shared logic in core, and the distinct logic in each crate. What are the challenges to keeping it split out?

trin-core/src/portalnet/overlay.rs
```diff
@@ -152,20 +145,229 @@ impl Overlay {
...
+        // a session on the (overlay) portal network.
+        for enr in self
+            .overlay
+            .discovery
+            .read()
+            .await
+            .discv5
+            .table_entries_enr()
+        {
+            debug!("Pinging {} on portal network", enr);
```
carver, line 293: I guess these logs can be more specific now, like:
```suggestion
            debug!("Pinging {} on portal state network", enr);
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -152,20 +145,229 @@ impl Overlay {
...
```
carver, line 367: and
```suggestion
            debug!("Pinging {} on portal history network", enr);
```

carver (approved):

trin-core/src/portalnet/overlay.rs
ogenev: Yeah, good catch, I totally forgot to move those in their own crates :)

KolbyML:
Wait can you even recv from protocol_receiver in two different spots? Maybe event_stream() does some trickery to allow you to have multiple receivers or maybe it just sends it multiple times. I thought you could only have one recv, but honestly I could have just read something wrong.

ogenev:
> Wait can you even recv from protocol_receiver in two different spots? Maybe event_stream() does some trickery to allow you to have multiple receivers or maybe it just sends it multiple times. I thought you could only have one recv, but honestly I could have just read something wrong.

Hey, this PR does not include support for handling events from multiple networks, The issue is addressed in https://github.com/ethereum/trin/pull/110.

PR #104, opened 2021-09-14
Add initial bridge trait and Infura implementation

jacobkaufmann:


jacobkaufmann:
The `Bridge` trait defines the possible ways to interact with a bridge node.

The trait defines the following interactions:

* Request some piece of Ethereum data (e.g. block by hash)
* Subscribe to a stream of Ethereum data (e.g. new block headers)
* Unsubscribe from an existing subscription

This would allow a portal node to subscribe to various streams of new Ethereum data as it becomes available. A portal node who participates in the chain history subnetwork may subscribe to new block headers. A portal node who participates in the state subnetwork may subscribe to new state changes (e.g. modified accounts or storage slots). A portal node that participates in multiple subnetworks can create multiple subscriptions.

The ability to subscribe to events does not make any assumptions about the transport protocol used to communicate with and acquire data from the bridge node.

Although the bridge node specification is incomplete and subject to change, this general and minimal interface should remain compatible with the specification as it evolves.

The finalization of the bridge node specification would allow us to create a trait for a bridge node client and more effectively use Rust's type system. This may not be the final form of `Bridge`, but I expect the usage patterns to remain mostly similar.

jacobkaufmann:
Several things to fix here, and in the process of fixing those things, I'm favoring splitting the following sub-changes into separate PRs:

* Add additional structures to `jsonrpc.rs` to handle JSON-RPC responses and notifications
* Move block header structure from `trin-history` to `trin-core` and create a block structure to parse JSON-RPC messages

As I mentioned on the portal network implementers call, I don't think is necessarily the best structure from a Rust type system standpoint. When we have a specification for the bridge node, I think something like the following might be better:

```rust
trait BridgeClient {
  // ... bridge API functions ...
}

struct Bridge<T> {
  inner: T
  // ... snip ...
}

impl<T: BridgeClient> Bridge<T> {
  // request function
  // subscribe function
  // unsubscribe function
}
```

@njgheorghita any thoughts? Like I said previously, I'm mainly concerned with the bridge interface at this point, given that the Infura implementation is more of a temporary proxy. There are implementation details whose deficiencies I'm aware of, and that I have fixed or plan to fix before moving out of draft, but I would like to get some approval on this structure before proceeding. Thanks.

njgheorghita (commented):
> I don't think is necessarily the best structure from a Rust type system standpoint. 

This is something for which I don't have a great instinct - I tend to fall in the camp of get it working then worry about optimization (which is not always the best strategy). 

> I would like to get some approval on this structure before proceeding. 

I'm assuming you're referring to the structure in the pr - not the "ideal" structure you prototyped in your comment. Just some initial thoughts...

- This structure seems sound to me - though, I'm curious as to what the deficiencies you mentioned are?
- We do have this [wip custom bridge jsonrpc api spec](https://github.com/ethereum/stateless-ethereum-specs/blob/master/portal-bridge-nodes.md). My current take is that, if we can get the state bridge data from the established jsonrpc api into the network without building a custom plugin, we should. Whether or not this is possible is tbd.
- It seems inevitable that at least for some intermediate period a trin bridge node should support multiple types of bridge data sources (infura for history, geth for state (& history?)). But, I guess I don't see a case in where a bridge node would use both at the same time. If they have access to a geth node, they'll use that for history & state. If not, they'll use infura for history.
- Is the Infura bridge node data source only temporary?
  - If a user doesn't have access to a geth node, but they have an infura account, and want to run their trin client in bridge mode... Should we allow the ability to use infura to source history network bridge data?

These are more just general thoughts / open questions about how we approach the bridge node. Imo this pr looks like a solid start - though I would feel more comfortable if you added some tests.

jacobkaufmann:
To start, yes I am referring to the existing `Bridge` trait structure.

And tests are certainly part of the next revision. I just wanted to get the structure in front of the eyes of the contributors before writing tests for logic that may not withstand the initial review.

On deficiencies, I was mainly referring to incomplete data structures, error handling, etc.

On using the Ethereum JSON-RPC API if possible and multiple data sources:

* What APIs does `geth` expose that are non-standard (not supported by Infura)? 

It may be better to define a trait for the standard Ethereum JSON-RPC API and a subtrait for the "extended" `geth` API.

```rust
trait EthJsonRpcClient {
  // ... standard APIs ...
}

trait ExtEthJsonRpcClient: EthJsonRpcClient {
  // ... geth-specific APIs ...
}
```

Then the bridge clients could look something like the following:

```rust
struct HistoryBridge<T> {
  T: inner
}

impl<T: EthJsonRpcClient> HistoryBridge<T> {
  // ... bridge APIs ...
}

struct StateBridge<T> {
  T: inner
}

impl<T: ExtEthJsonRpcClient> StateBridge<T> {
  // ... bridge APIs ...
}
```

I think there's a preference to not use Infura as a data source as opposed to something like `geth` (from a decentralization perspective), but it seems reasonable to allow a user to use Infura if that would suffice for the type of portal node the user wants to run.

njgheorghita:
NIce, yeah I think those look great.

> What APIs does geth expose that are non-standard (not supported by Infura)?

One we've been exploring includes [`debug_dumpBlock`](https://geth.ethereum.org/docs/rpc/ns-debug#debug_dumpblock) another is `eth_getProof` - which actually is available for infura but only if you pay for access to an archive node. But, I'm not convinced that we should initially worry about supporting the archive infura api.

> I think there's a preference to not use Infura as a data source as opposed to something like geth (from a decentralization perspective)

Yeah, for sure. And it's possible that I'm missing some reason that we actually don't want to allow infura in the final version - I just haven't heard a definitive reason yet.

jacobkaufmann:
Closing in favor of #119

Issue #98, opened 2021-09-07
Content-id derivation function


victor-wei126:
So in order to get the content-id from the content-key, there is some function that maps the key to the id. On the history network, this is just simply taking the sha256 hash of the entire key. On the state network, it is more complicated since depending on the type of content, we may use different hash functions, which may hash different portions of the content-key.

So I am thinking this method will be defined in `utils.rs` and it will take in a key as parameter, and return the id after performing appropriate computation. To do the right computation, it needs to determine 1) if this key is for content from history or state and 2) if from state, what type of content exactly is it (account trie node, contract trie node). 

For 2), the function can just extract the first byte from the key, since that is the `content_type` and be done with it there. For 1, I am not so sure how to do this. If the only parameter is the key, how can we differentiate between the networks?

njgheorghita:
If I understand correctly, I don't see why we couldn't have separate derivation utils within their respective package (`trin-state`, `trin-history`). So, deriving content-ids will be encapsulated within a subnetworks logic (i.e. `derive_content_id()` within `trin-state/utils.rs` will only have to worry about deriving state network ids, etc.).

victor-wei126:
I think that makes a lot of sense. Although, I thought about this some more and I believe @mrferris implementation of should_store() method will call these derivation functions, and so depending on whether the key came from a state network or history network message will still have to be determined either way. Unless, this is just purely an issue of keeping things encapsulated, then that makes sense.

mrferris:
Right now I'm working on making the PortalStorage's content-id function parameter-izable, so (if we end up going the route of one PortalStorage instance per network) it will look something like:
```
fn history_network_content_key_to_id(&String key) -> String {
...
}
history_storage_config.content_key_to_id = &network_specific_content_key_to_id;
```

And then should_store() and all of the other functions that also need to to do the conversion just call that function internally.

I'm thinking of doing it this way because I don't think that we should be importing from any of the individual network packages into `trin-core` (where `storage.rs` is), and should instead only import `trin-core` into the individual network packages (and `trin` itself). Let me know if anybody understands differently.

If we end up having one instance of PortalStorage handle the data for all networks, then yes @victor-wei126 I was thinking the same: we'll need another way to determine which network the data is for each time we call should_store etc.

victor-wei126:
[...]

PR #107, opened 2021-09-15
[WIP] Content-key to content-id derivation functions

victor-wei126:
This is following issue #98. If anyone can direct my where to look for decoding the payload? I know I need to use SSZ `Container` sedes, and I've tried looking at the specs and `from_ssz_bytes` function in `types.rs`. 

And plz ignore the extraneous commits and code that is "new" but not really. I've been having issues with merging and need to figure that out. New code is everything after line 99.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -138,7 +138,7 @@ dependencies = [
  "lazy_static 1.4.0",
  "lazycell",
  "peeking_take_while",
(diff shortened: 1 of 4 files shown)


pipermerriam:
Lookup the "git rebase" workflow.  This is what you want to cleanup your history.

PR #108, opened 2021-09-16
Content-Keys Type Definitions & key->id derivation functions

victor-wei126:
This PR does a few things:

- Implements the type definitions for the state and history network `content-keys`. The associated methods for SSZ encoding and decoding are defined as well
- Implements the key->id derivation functions for both state and history networks. Currently, these are associated functions of the corresponding type def. There is another wrapper function which determines if a key is for state/history and chooses the appropriate derivation function to call
- Separate `types.rs` in `trin-core` into separate files since it is getting large.

After this PR is merged, I plan to:
- Move the u256 and u512 from mod.rs to `types/utils.rs`
- Implement cross-client test vectors


victor-wei126:
@mrferris I've gotten a derivation function for state. I used SSZ encode & decode traits and basically it matches the decoded key to perform the appropriate function. Let me know if this is where we're trying to head

trin-state/src/utils.rs
```diff
@@ -18,6 +23,69 @@ const MID: [u8; 77] = [
     49, 57, 57, 54, 56,
 ];
 
+pub fn state_network_derive_content_id(key: &String) -> U256 {
+    // SSZ decode ContentKey
+    let decoded_key = match StateContentKey::from_bytes(&key.as_bytes()) {
+        Ok(k) => k,
+        Err(_) => panic!("Cannot decode key")
```
carver, line 30: Looks like the derive method should return a `Result` instead. Crashing the program on an invalid content key sounds bad.

trin-state/src/utils.rs
```diff
@@ -18,6 +23,69 @@ const MID: [u8; 77] = [
...
+    };
+
+    match decoded_key {
+        StateContentKey::AccountTrieNode(AccountTrieNode { path, node_hash, .. }) => {
+            let mut input = String::from_utf8(path).unwrap();
```
carver, line 35: Along the same lines: shouldn't crash on a bad utf-8.

trin-state/src/utils.rs
```diff
@@ -18,6 +23,69 @@ const MID: [u8; 77] = [
...
+
+fn sha256(key: &String) -> U256 {
+    let mut hasher = Sha3_256::new();
+    hasher.update(key);
+    let result = hasher.finalize().to_vec();
+    U256::from(vec_to_array(result))
+}
+
+fn keccak(key: &String) -> U256 {
+    let mut hasher = Keccak256::new();
```
carver, line 83: Ah, interesting, I may have missed this convo. Are we really switching to sha3_256 proper? There's a lot of room for confusion, for it to live alongside keccak256. Wondering if the spec was just a little loose in saying sha, when it meant keccak.

trin-history/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+    v.try_into()
+      .unwrap_or_else(|v: Vec<T>| panic!("Expected a Vec of length {} but it was {}", N, v.len()))
+}
+
+fn sha256(key: &String) -> U256 {
+    let mut hasher = Sha3_256::new();
+    hasher.update(key);
+    let result = hasher.finalize().to_vec();
+    U256::from(vec_to_array(result))
+}
```
carver, line 41: nit: it's nice to end the file with a newline

trin-history/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+        }
+    }
+}
+
+fn vec_to_array<T, const N: usize>(v: Vec<T>) -> [T; N] {
+    v.try_into()
+      .unwrap_or_else(|v: Vec<T>| panic!("Expected a Vec of length {} but it was {}", N, v.len()))
+}
+
+fn sha256(key: &String) -> U256 {
```
carver, line 36: Feels silly to redefine this in every crate. Seems like this and keccak could go into the `trin-core` utils.

trin-history/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+    // decode key
+    let decoded_key = match HistoryContentKey::from_bytes(&key.as_bytes()) {
+        Ok(k) => k,
+        Err(_) => panic!("Could not decode key"),
+    };
+
+    // calculate content-id
+    match decoded_key {
+        HistoryContentKey::HeaderKey(HeaderKey { chain_id, content_type, block_hash }) => {
+            let input = U256::from(chain_id) + U256::from(content_type) + block_hash;
```
carver, line 17: It's not immediately obvious to me that using `+` on a U256 will concatenate the bytes instead of numerically add the results. In fact, I would have guessed the opposite, but I guess that's just more of a complaint with the U256 lib. If there's another way to be more explicit that we are concatenating bytes, just for readability in trin, I think that would be preferable.

trin-core/src/portalnet/types.rs
```diff
@@ -345,6 +345,137 @@ impl FromStr for HexData {
     }
 }
 
+pub enum ContentKey {
+    StateContentKey(StateContentKey),
+    HistoryContentKey(HistoryContentKey),
+}
+
+pub enum StateContentKey {
+  AccountTrieNode(AccountTrieNode),
```
carver, line 354: I worry that this duplicated name can be a bit confusing. Maybe something like this?
```suggestion
  AccountTrieNodeKey(AccountTrieNode),
```

trin-core/src/portalnet/types.rs
```diff
@@ -345,6 +345,137 @@ impl FromStr for HexData {
...
+}
+
+impl StateContentKey {
+  // decodes from bytes to ssz container
+  pub fn from_bytes(bytes: &[u8]) -> Result<Self, String> {
+      if let Some(content_type) = bytes.first() {
+          match content_type {
+              0x00 => Ok(StateContentKey::AccountTrieNode(
+                  AccountTrieNode::from_ssz_bytes(&bytes[1..])
+                    .map_err(|e| format!("Failed to decode ssz: {:?}", e))?
```
carver, line 368: Have you looked at the errors resulting from this? It seems like it might be useful to explicitly mention that the code was attempting to decode with `AccountTrieNode`, maybe maybe that's already clear enough in `e`? Also, if `e` doesn't include the byte source, I'd like to see it added here, hex-encoded. It will be so much easier to track down bugs later from the logs, if we have the source bytes and know which thing we were trying to decode as.

trin-core/src/portalnet/types.rs
```diff
@@ -345,6 +345,137 @@ impl FromStr for HexData {
...
+              )),
+              0x04 => Ok(StateContentKey::ContractBytecode(
+                  ContractBytecode::from_ssz_bytes(&bytes[1..])
+                    .map_err(|e| format!("Failed to decode ssz: {:?}", e))?
+              )),
+              _ => Err("No matching content type".to_string()),
+          }
+      } else {
+          Err("No content type".to_string())
```
carver, line 389: This message be the only text you see in a spew of logs, when trying to debug a problem. Without more context, "No content type" wouldn't help me find the problem easily. Maybe something like this would help a bit more?
```suggestion
          Err("Cannot decode StateContentKey from empty bytes".to_string())
```

trin-core/src/portalnet/types.rs
```diff
@@ -345,6 +345,137 @@ impl FromStr for HexData {
...
+              )),
+              0x03 => Ok(StateContentKey::ContractStorageTrieProof(
+                  ContractStorageTrieProof::from_ssz_bytes(&bytes[1..])
+                    .map_err(|e| format!("Failed to decode ssz: {:?}", e))?
+              )),
+              0x04 => Ok(StateContentKey::ContractBytecode(
+                  ContractBytecode::from_ssz_bytes(&bytes[1..])
+                    .map_err(|e| format!("Failed to decode ssz: {:?}", e))?
+              )),
+              _ => Err("No matching content type".to_string()),
```
carver, line 386: Similarly, would recommend including the `content_type` byte that has no match in this message (and include that you're looking in the `StateContentKey`).

carver (commented):
I think the code would really benefit from some tests of decoding content keys, and converting to IDs, for both successful cases and error cases.

Feel free to ping me for another review when tests are added (and existing tests are passing).

trin-state/src/utils.rs
victor-wei126: The crate I am using has both sha3_256 and Keccak256. It notes that Keccak256 is the NIST submission without padding changes. So I'm guessing sha3_256 is the actual "controversial" standard? The specs for state and history specify both "sha256" and just "keccak" so my guess is they mean the SHA-2 family for "sha256" and "keccak" as the "unproper". @pipermerriam

trin-history/src/utils.rs
victor-wei126: I was under the impression that the `+` was adding the U256s, which is what I thought we wanted? I'm not sure since the spec is not clear on this point. I also add two U256s in calculating the ContractStorageTrieProof key, specified by the spec [here](https://github.com/ethereum/portal-network-specs/blob/master/state-network.md#contract-storage-trie-proof). The only thing I have to support that `+` is actually addition and not concatenation is that in the state-network spec `|` is used for concatenation, but `+` symbol is used, so I assumed its addition. History spec is unclear.

trin-core/src/portalnet/types.rs
victor-wei126: idk why I added key to everything else except for state keys, probably cuz i thought the names were getting too long lol. will add

trin-history/src/utils.rs
carver: Oh, interesting! I just hadn't noticed that it was actually addition in some places. Will U256 automatically roll-over as the spec defines? More things to test. I'm also not clear why we are sometimes using addition and sometimes concatenation. There are fewer question marks with concat (like not having to worry about rollover), so I would recommend pushing back on the spec a little, if there's no obviously good reason to use addition.

carver:
Can you fix the formatting errors so we can see the tests passing?

trin-history/src/utils.rs
```diff
@@ -0,0 +1,93 @@
...
+  portalnet::{
+    U256,
+    types::{HistoryContentKey, HeaderKey, BodyKey, ReceiptsKey},
+  },
+  utils::content_key::{keccak}
+};
+use hex;
+
+// content-id derivation function for history network
+pub fn history_network_derive_content_id(key: Vec<u8>) -> Result<U256, String> {
```
mrferris, line 11: I'm thinking that these should be methods of the `ContentKey` enums defined in `types.rs` instead of external `util` functions, and then I'll update`PortalStorage` methods take a `ContentKey` and just call `derive_content_id` on it internally instead of having to pass in the function like we're doing now. That will be simpler and better organized.

Any reason you can see that we shouldn't do that?

trin-core/src/utils/content_key.rs
```diff
@@ -0,0 +1,15 @@
...
+    let mut hasher = Keccak256::new();
+    hasher.update(key);
+    let result = hasher.finalize().to_vec();
+    Ok(U256::from(vec_to_array(result)?))
+}
+
+pub fn vec_to_array<T, const N: usize>(v: Vec<T>) -> Result<[T; N], String> {
+    v.try_into()
+      .map_err(|v: Vec<T>| format!("Expected a Vec of length {} but it was {}", N, v.len()))
+}
```
mrferris, line 15: nit: needs a newline at end of file

src/main.rs
```diff
@@ -12,7 +12,7 @@ use trin_core::{
     cli::{TrinConfig, HISTORY_NETWORK, STATE_NETWORK},
     jsonrpc::service::launch_jsonrpc_server,
     portalnet::{discovery::Discovery, types::PortalnetConfig},
-    utils::setup_overlay_db,
+    utils::utils::setup_overlay_db,
```
mrferris, line 15: not for this PR, but I don't love `utils::utils`. Once you merge this I'm going to make a `utils/db.rs` that will include this function, and then we can think of a better name for a file containing the other things.

mrferris (commented):
Looks pretty good, I left a few thoughts, some of which are related to each other.

Looks like you need a `cargo fmt` too.

trin-core/src/portalnet/types.rs
```diff
@@ -345,11 +345,405 @@ impl FromStr for HexData {
...
+#[derive(PartialEq, Clone, Debug, Encode, Decode)]
+pub struct AccountTrieNode {
+    pub path: Vec<u8>,
+    pub node_hash: U256,
+    pub state_root: [u8; 32],
+}
+
+#[derive(Debug, Clone, PartialEq, Encode, Decode)]
+pub struct ContractStorageTrieNode {
+    pub address: Vec<u8>, // should be bytes20
```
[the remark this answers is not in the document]
victor-wei126: I think you mean `FixedVector`? Since this needs 20 bytes and 20 bytes only. Also, since we can now start using `ssz_types` this raises the larger question of if we want to replace all types with `ssz_types` implemented in the crate? For example, for a bytes32, we can use a bare rust type of `[u8; 32]` or use `ssz_types` of `FixedVector<u8, typenum::U32`. Having the bare type makes things simpler but using the ssz_type gives us additional flexibility.

trin-history/src/utils.rs
victor-wei126: This means you can get rid of the closure function in PortalStorage right? Yeah 100% I think this has multiple benefits. The only thing is, since it is now a method, we have to call it like `key.derive_content_id()`, where `key` is of type `ContentKey`, as I understand. So we'd need to get the type first, which means instead of passing a byte array to `derive_content_id` and then decoding in that method, we would have to decode the byte array first somewhere outside of the method, which gives us the type, and then we can call the derivation function. But regardless, I think this way of doing things is much cleaner and organized.

mrferris: 👍 I'd think you can just do a `match` on the first byte inside the function but I might not be seeing what you mean fully.

Either way, let's go that route. Once you merge this in I'll refactor PortalStorage to take the `ContentKey` type you define in this PR.

trin-core/src/cli.rs
```diff
@@ -1,4 +1,4 @@
-use crate::portalnet::types::HexData;
+use crate::portalnet::types::messages::HexData;
```
pipermerriam, line 1: A good practice is to keep strong separation between organizational changes and functionality changes.  A good way to deal with this is to extract just the part that moved `HexData` to this new location and then open a pull request with just that change.  Then you can rebase this branch on top of that other PR which will allow a reviewer of this PR to not have to sort through more changed lines than absolutely necessary.

victor-wei126:
> I would prefer to see the content key derivation fns split out into trin-state/ and trin-history/ in this pr.

@njgheorghita Just to confirm I'm understanding you correctly, do you mean move just the derivation functions out of trin-core, or also move the entire `StateContentKey` to `trin-state/` and same for history? If anything, I think if we were to move things out of trin-core, then we should move the entire enum and all it's methods, not just the derivation function.

But I would like to push back a bit and say that by moving these type definitions out of trin-core and into their respective crates, we then have to make trin-state and trin-history a dependency of trin-core (unless there is some way around this that I'm unaware of), since `PortalStorage` relies on this functionality, and that is in trin-core. So my opinion as of now is to just keep it everything in trin-core, and if the specific subnetworks need the functionality we can move it out.

trin-core/src/portalnet/mod.rs
```diff
@@ -15,6 +17,49 @@ mod overlay_service;
 
 pub type Enr = discv5::enr::Enr<CombinedKey>;
 
+construct_uint! {
```
njgheorghita, line 20: So, this was the home for `U256` - but it doesn't feel like the appropriate place for these uint definitions/tests. Personally, I don't think we should have any code inside our `mod.rs` files - it should be strictly for `pub mod ..` declarations. How do you feel about moving `U256` and `U512` definitions and tests over to `trin-core/src/types/uint.rs` or something similar?

trin-core/src/portalnet/mod.rs
```diff
@@ -68,8 +113,47 @@ impl ssz::Decode for U256 {
...
+    const U256_MAX: U256 = U256([u64::MAX, u64::MAX, u64::MAX, u64::MAX]);
+    const U512_MAX: U512 = U512([u64::MAX, u64::MAX, u64::MAX, u64::MAX, 0, 0, 0, 0]);
+
+    #[rstest]
+    #[case(U256([1,2,3,4]), U512([1,2,3,4,0,0,0,0]))]
+    #[case(U256([492,4039,129,9984]), U512([492,4039,129,9984,0,0,0,0]))]
+    #[case(U256_MAX, U512_MAX)]
+    #[case(U256([u64::MAX,u64::MAX,u64::MAX-1,u64::MAX]), U512([u64::MAX,u64::MAX,u64::MAX-1,u64::MAX,0,0,0,0]))]
+    #[case(U256([0,0,0,0]), U512([0,0,0,0,0,0,0,0]))]
+    fn u256_to_u512_test(#[case] input: U256, #[case] expected: U512) {
```
njgheorghita, line 128: *nitpick* this and the following test name should begin with `test_*`

trin-core/src/portalnet/mod.rs
```diff
@@ -68,8 +113,47 @@ impl ssz::Decode for U256 {
...
+        let result = U512::from(input);
+        assert_eq!(expected, result);
+    }
+
+    #[rstest]
+    #[case(U512([1,2,3,4,0,0,0,0]), U256([1,2,3,4]))]
+    #[case(U512([420,69,420,69,0,0,0,0]), U256([420,69,420,69]))]
```
njgheorghita, line 135: Sorry for going full boomer, and tbf I do this kinda thing constantly when I'm developing locally, but I'd rather not push any meme code to master if you can replace this test with something more boring.

trin-core/src/portalnet/mod.rs
```diff
@@ -68,8 +113,47 @@ impl ssz::Decode for U256 {
...
+        let result = input.try_into().unwrap();
+        assert_eq!(expected, result);
+    }
+
+    #[rstest]
+    #[case(U512([1,2,3,4,1,0,0,0]), U256([1,2,3,4]))]
+    #[case(U512([1,2,3,4,0,34,0,0]), U256([1,2,3,4]))]
+    #[case(U512([1,2,3,4,0,0,234,0]), U256([1,2,3,4]))]
+    #[case(U512([1,2,3,4,0,0,0,349]), U256([1,2,3,4]))]
+    #[case(U512([u64::MAX,u64::MAX,u64::MAX,u64::MAX,1,0,0,0]), U256([1,2,3,4]))]
```
njgheorghita, line 149: It doesn't seem to me like this is the correct expected case (and for the test below)? 
- eg. shouldn't the expected case (even though it'll always panic) be `U256([U64::MAX;4])`?

It's not obvious to me what's throwing the panic here - is it line 153 or 154? If it's 154, it would be better to not panic and `assert_eq` explicitly against the `U512toU256Error::Overflow` that's being returned for overflows.

trin-core/src/portalnet/types.rs
```diff
@@ -555,10 +559,327 @@ impl FromStr for HexData {
     }
 }
 
+// ===========================================
+// CONTENT KEY FUNCTIONALITY (will be moved to separate file)
+// ====================================
```
njgheorghita, line 564: I'd rather see this move happen in this pr. I can't think of a great reason to delay the move for a later pr (but i'm all ears). But yes, in regards to your other comment, it makes sense to keep them inside `trin-core` if `trin-core` depends on these types.

trin-core/src/portalnet/types.rs
```diff
@@ -705,4 +1026,412 @@ mod test {
...
+        };
+        let test_key =
+            ContentKey::StateContentKey(StateContentKey::AccountTrieNodeKey(container.clone()));
+
+        let ssz_bytes = test_key.to_bytes();
+        println!("{:?}", ssz_bytes);
+        let decoded = StateContentKey::from_bytes(&ssz_bytes).unwrap();
+        println!("{:?}", decoded);
+
+        if let StateContentKey::AccountTrieNodeKey(k) = decoded {
```
njgheorghita, line 1055: I don't think this is ideal for our tests. It seems possible that these might not run at all if the `let` statement doesn't match. At the very least you should add an `else { panic!(..) }` statement to the end. Or you could do something like [here](https://stackoverflow.com/questions/34953711/unwrap-inner-type-when-enum-variant-is-known) where you implement functions that'll return the inner value

njgheorghita (commented):
Great! It feels like this is almost there - just a couple nitpicks / re-org suggestions. It makes sense to me to handle the re-org in this pr and not save it for a later pr. 

I'd like to go through this one more time though after the next round of fixes, just to go through the test cases in a little more detail - though everything checks out at first glance. If you could also squash / rebase the majority of those commits, we don't want those ending up in master.

Also, it seems like some cross-client test vectors for the encoding / decoding would be extremely useful (like https://github.com/ethereum/portal-network-specs/blob/dbbb4c8774/portal-wire-test-vectors.md). Is that something you could pick up after this pr? The other test vectors were great for pointing out some blind spots in our code.

trin-core/src/portalnet/mod.rs
victor-wei126: Yeah that makes sense to me since those are types. Should I address that in this PR?

trin-core/src/portalnet/mod.rs
victor-wei126: haha yes i will change that

trin-core/src/portalnet/mod.rs
njgheorghita: Probably not this pr since there's a lot going on already, but i'd just take a not to address it in a follow-up pr after this gets merged

trin-core/src/portalnet/mod.rs
victor-wei126: Well in my impl if there is an overflow the conversion just fails and doesn't default to the max u256 value, so I changed this test case to have an expected value of `Ut12toU256Error::Overflow` as that should be the actual expected value. Let me know if this is what you were thinking.

trin-core/src/portalnet/types.rs
victor-wei126: A general testing question: we're expecting this test to succeed, since I made sure to have the proper input, but is including a panic just for safekeeping, in case the input was wrong somehow?

victor-wei126: Also, I just reviewed the code again and I think adding an `else { panic!(..) }` at the end of the `if` would be redundant because the `unwrap()` above will cause the test to panic in the case of decoding failure, which is what the else would be doing. The `if let` statement handles the case of successful decoding.

njgheorghita: Yeah including the panic is just for safekeeping. In this example, let's say the deserializing code changes somehow that instead of returning `AccountTrieNodeKey` for `decoded` - it returns `ContractStorageTrieNodeKey`. The `unwrap()` wouldn't catch this, b/c it was successfully deserialized. Then the if statement never gets run and the test passes successfully. Generally, I think it's smart to avoid `if` statements or other logic in your tests unless it's strictly necessary. In this scenario, adding a little `else {panic!()}` clause guarantees that either the tested assumptions inside the `if()` statement are run - or it panics.

njgheorghita (approved):
Looks great! Thanks for keeping with this and handling all the suggestions. Sorry for the late response here, it fell through the cracks a bit on my end. In the future, if you're waiting on a review from me for more than a couple days - feel free to re-ping me here or on discord.

Just one comment to address re: testing & panics. Also curious if you've put any time towards creating test vectors for these messages (like [these](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-test-vectors.md)? If not, creating a small issue to track that would be neat.

Issue #81, opened 2021-08-24
Implement --ephemeral flag


ogenev:
When testing internal p2p network calls, we would like to be able to run multiple node instances simultaneously.
To avoid db folder collusion, we want to set unique db folder name for every node id.

As a default db folder, current implementation is using last 8 characters of base64 encoded `Enr` appended to the application name:
https://github.com/carver/trin/blob/d99232fd43c8dd0ec055150f382546cbba747bef/trin-core/src/utils.rs#L58

It will be better, if we change this to append the first few characters (instead of last 8) of the as-displayed node-id to the application name.

mrferris:
What's the reasoning for using the first 8 instead of last 8?

ogenev:
> What's the reasoning for using the first few instead of last 8?
> 
> Now that I ask I'm guessing it's because the beginning part is more easily comparable by eye, like with git commit hashes.

Yeah, it is related to Jason's comment [here](https://github.com/ethereum/trin/pull/80#discussion_r694203573).

hmrtn:
PR #277 is ready for review 👍

PR #116, opened 2021-09-23
Modifies get_default_data_dir to take node id instead of enr

mrferris:
Implements @ogenev's initial thoughts from #81.

Uses the first 8 characters of the Node ID's base 10 representation.
Any thoughts on whether we may want hex instead, or less than 8 characters?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -54,7 +54,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     discovery.write().await.start().await.unwrap();
 
     // Setup Overlay database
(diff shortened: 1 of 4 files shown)


njgheorghita:
Personally, I'd prefer to see hex used here. Not a strong opinion, but we'd use hex to display the node id to terminal output, so it seems more consistent to also use it here.

jacobkaufmann:
LGTM. I think @pipermerriam's comments about the `--ephemeral` flag in #81 can be addressed separately if desired.

jacobkaufmann (approved):

PR #119, opened 2021-09-27
Add Ethereum and Go-Ethereum JSON-RPC bridge clients

jacobkaufmann:
This is a continuation of #104. The changes for this PR were done on a separate local branch. It seems that it's not possible to change the branch in the original PR.

### DIFF
--- a/trin-core/src/ethereum/bridge.rs
+++ b/trin-core/src/ethereum/bridge.rs
@@ -1,0 +1,236 @@
+use async_trait::async_trait;
+use ethers_core::types::{
+    Address, Block, Bytes, EIP1186ProofResponse, Transaction, TransactionReceipt, TxHash, H256,
+    U256, U64,
+};
+
+#[cfg(unix)]
+use ethers_providers::Ipc;
+use ethers_providers::{
+    JsonRpcClient, Middleware, Provider, ProviderError, PubsubClient, SubscriptionStream, Ws,
+};
+
+use log::error;
+use url::Url;
+
+use std::error::Error;
+use std::fmt;
+
+#[cfg(unix)]
+use std::path::Path;
+
+// TODO: Add more comprehensive errors with more concrete types.
+/// An error that can occur in communications with a bridge node.
+#[derive(Debug)]
+pub enum BridgeError {
+    EthClientError(Box<dyn std::error::Error + Send + Sync>),
+    SerializationError(Box<dyn std::error::Error + Send + Sync>),
+    ConnectionFailure,
+    Other(Box<dyn std::error::Error + Send + Sync>),
+}
+
+impl fmt::Display for BridgeError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::EthClientError(error) => write!(f, "bridge ETH client error: {}", error),
+            Self::SerializationError(error) => write!(f, "bridge serialization error: {}", error),
+            Self::ConnectionFailure => write!(f, "bridge connection failure"),
+            Self::Other(error) => write!(f, "bridge error: {}", error),
+        }
+    }
+}
+
+// TODO: Implement source.
+impl Error for BridgeError {}
+
+impl From<serde_json::Error> for BridgeError {
+    fn from(error: serde_json::Error) -> Self {
+        BridgeError::SerializationError(error.into())
+    }
+}
+
+impl From<ProviderError> for BridgeError {
+    fn from(error: ProviderError) -> Self {
+        match error {
+            ProviderError::JsonRpcClientError(error) => BridgeError::EthClientError(error),
+            ProviderError::SerdeJson(error) => BridgeError::SerializationError(error.into()),
+            ProviderError::HexError(error) => BridgeError::SerializationError(error.into()),
+            error => BridgeError::Other(error.into()),
+        }
+    }
+}
+
+/// A type alias for a `Result` whose `Err` variant holds a `BridgeError`.
+type Result<T> = std::result::Result<T, BridgeError>;
+
+/// An Ethereum bridge client.
+#[async_trait]
+pub trait EthBridge {
+    type Client: PubsubClient;
+
+    /// Requests the balance of an account at a particular block number.
+    async fn get_balance(&self, address: Address, block_number: U64) -> Result<U256>;
+    /// Requests a block (with full transaction objects) by hash.
+    async fn get_block(&self, hash: H256) -> Result<Option<Block<Transaction>>>;
+    /// Requests the code of an account at a particular block number.
+    async fn get_code(&self, address: Address, block_number: U64) -> Result<Bytes>;
+    /// Requests the value of a storage slot at a particular block number.
+    async fn get_storage(
+        &self,
+        address: Address,
+        location: H256,
+        block_number: U64,
+    ) -> Result<H256>;
+    // TODO: Replace parameters with block hash and transaction index.
+    /// Requests the transaction for a given transaction hash.
+    async fn get_transaction(&self, hash: H256) -> Result<Option<Transaction>>;
+    /// Requests the number of transactions sent (nonce) from an account at a particular block number.
+    async fn get_transaction_count(&self, address: Address, block_number: U64) -> Result<U256>;
+    /// Requests the transaction receipt for a given transaction hash.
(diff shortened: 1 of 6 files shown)


jacobkaufmann:
After several false starts and feedback from other contributors, I figured that this was the simplest initial approach to building an initial bridge. Here, we leverage the `ethers-rs` package for the initial implementations of `EthBridge` and `GethBridge`.

One for sure to-do is to change the types exposed by the API to types that are RLP-serializable. I think this can be taken up in a subsequent PR.

Aside from that, I think these would be the APIs that we would like to expose from the bridge clients. However, I'm not sure that we need the level of granularity to support requests for individual transactions or transaction receipts. Instead, I think the pattern might be to request all transactions and the corresponding receipts for an entire block. Therefore, I think the API for `EthBridge` could be further simplified to something like the following:

```rust
trait EthBridge {
  async fn get_account(address: Address, block_number: U64) -> Account;
  async fn get_block(hash: H256) -> Block;
  async fn get_receipts_for_block(hash: H256) -> Vec<TransactionReceipt>;
  async fn get_storage(address: Address, location: H256, block_number: U64) -> StorageSlot;
  async fn subscribe_new_blocks() -> Subscription<Block>;
}
```

Unfortunately, there are no convenient JSON-RPC APIs to retrieve the storage root or code hash for some account. You can only retrieve individual storage slots and/or the entire code for some account. AFAIK, to support a `get_account` from a connection to a general node would require multiple requests and some logic to reconstruct the hashes.

Using the `eth_getProof` API offered by `geth`, you can retrieve all the values for some account along with the proofs.

jacobkaufmann:
It's also worth mentioning that right now, the `EthBridge` and `GethBridge` traits require a `PubsubClient`, when they should only require a `JsonRpcClient`. Even without pub-sub support, you can implement subscriptions with some kind of polling mechanism under the hood. I was running into issues with Rust lifetimes + pinning and decided that this would be okay for now. What I think is important now is the definition of a stable bridge API, and we can tune the implementation in the future.

pipermerriam:
Feel free to be opinionated and only support a single JSON-RPC transport if that makes your life easier.  For example, it would be totally ok for this functionality to be built out such that it is geth specific and that it only works over websockets.

trin-core/src/ethereum/bridge.rs
```diff
@@ -0,0 +1,227 @@
...
+            Self::ConnectionFailure => write!(f, "bridge connection failure"),
+        }
+    }
+}
+
+impl Error for BridgeError {}
+
+impl From<serde_json::Error> for BridgeError {
+    fn from(_: serde_json::Error) -> Self {
+        BridgeError::SerializationError
```
lithp, line 39: I'm still new to rust so feel free to ignore this if it's misguided, but it looks like this (and the following method) throws away the original error. I bet that most times someone sees this error they'll be unsure what happened and they'll want to see the original error. I think implementing `source()` on `BridgeError` will make future problems easier to debug.

jacobkaufmann: Totally agree here. There is a `TODO` above the `BridgeError` definition to add more information to both the serialization and client error variants. We could either wrap the underlying errors or implement `source`. I'm not sure what is more idiomatic. I can go ahead and take this up and remove the `TODO` if we feel it necessary now.

trin-core/src/ethereum/bridge.rs
```diff
@@ -0,0 +1,227 @@
+use async_trait::async_trait;
```
njgheorghita, line 1: Just curious as to your reasoning for creating a new directory for this? Why is it named `ethereum`? What else do you envision living inside this namespace?

trin-core/src/ethereum/bridge.rs
njgheorghita: I'm not sure how high priority this is, so I would support implementing `source()` in this pr or create an issue to track it and handle in a subsequent pr.

trin-core/src/ethereum/bridge.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+}
+
+/// A type alias for a `Result` whose `Err` variant holds a `BridgeError`.
+type Result<T> = std::result::Result<T, BridgeError>;
+
+/// An Ethereum bridge client.
+#[async_trait]
+pub trait EthBridge {
```
njgheorghita, line 60: I feel like I've seen you explain the reasoning somewhere else, but I can't remember where... But, why did you choose the `EthBridge` / `GethBridge` names? Something like `HistoryBridge` and `StateBridge` seem much more intuitive to me, especially to help delineate the various responsibilities of each type, as this code gets used elsewhere in the codebase.

trin-core/src/ethereum/bridge.rs
```diff
@@ -0,0 +1,227 @@
...
+    async fn get_transaction_receipt(&self, hash: H256) -> Result<Option<TransactionReceipt>>;
+    /// Subscribes to a stream of new block headers as new blocks are appended to the chain.
+    /// In the case of chain reorganizations, all new block headers will be sent.
+    async fn subscribe_new_block_headers(
+        &self,
+    ) -> Result<SubscriptionStream<'_, Self::Client, Block<TxHash>>>;
+}
+
+/// A Go-Ethereum (Geth) bridge client. Geth clients support a superset of the standard Ethereum
+/// JSON-RPC APIs.
```
njgheorghita, line 91: I'm a big fan of all these comments, super useful 🥳

trin-core/src/ethereum/bridge.rs
```diff
@@ -0,0 +1,227 @@
...
+        &self,
+        address: Address,
+        storage_locations: &[H256],
+        block_number: U64,
+    ) -> Result<EIP1186ProofResponse>;
+}
+
+/// An ethers-rs Ethereum bridge client.
+pub struct EthersBridge<T: JsonRpcClient> {
+    inner: Provider<T>,
```
njgheorghita, line 108: `inner`? Something like `provider` would make more sense to me

njgheorghita (commented):
Just some general questions to help me better understand the direction of this..

trin-core/src/ethereum/bridge.rs
jacobkaufmann: Right, so the idea is that this would also host our own type definitions for Ethereum data structures. One example for why these would be necessary is that the types defined in `ethers-core` do not implement the `rlp` serialization traits. We would need to create our own types for this. I'm open to other ways of organizing this, but I thought that the bridge would live alongside the types of data that it collects.

trin-core/src/ethereum/bridge.rs
jacobkaufmann: Yeah, this convention came from `ethers-rs`, but I agree that it's not very meaningful. I can rename this to `provider` or `client` along with a doc comment. Not sure if there is a preference for one over the other.

trin-core/src/ethereum/bridge.rs
jacobkaufmann: Yeah, so I felt that `EthBridge` and `GethBridge` more accurately captured why there are two separate structures in the first place, denoting the additional set of capabilities of `geth`. I guess I saw that these names being more general was an advantage instead of a disadvantage. I cannot find a mention of "bridge nodes" in the docs on the header/transaction gossip network, but if those networks would also require a bridge, then there is not going to be some naming confusion around state versus history versus gossip. Instead, these bridges are more general components.

njgheorghita: Cool, ok hmmm I guess one question is that with this architecture, am I correct in assuming that you expect there to be a single bridge node process regardless of how many overlay networks are activated?

jacobkaufmann: That's correct. One thing to note with this implementation is that if there are multiple overlay networks that wish to subscribe to new block headers, then the bridge will initiate and manage multiple subscriptions (one for each overlay) instead of a single subscription with a fan-out pattern. This is definitely not ideal, but I'm not sure that the performance improvement would be worth the additional complexity at this time.

My thought was that this is more of a "minimum viable bridge" and that there is room for this optimization along with others when we want to build a more optimized implementation.

jacobkaufmann:
Oh, yay. Lots of old commits. Let me fix that.

jacobkaufmann:
Closing in favor of more recent work on the bridge node.

Issue #113, opened 2021-09-21
JSON-RPC handler does not support the Overlay protocol


ogenev:
Currently, the main JSON-RPC handler is initialized with discovery protocol, before the initialization of the overlay networks.

The problem of this implementation is that the handler has access only to the underlying discv5 protocol which prevents us from serving network specific endpoints. Examples of such endpoints could be requesting the overlay data radius value or modifying records in the overlay routing table.

One possible way to refactor this is to first initialize the overlay networks and then pass them to the network specific json-rpc handlers.

PR #120, opened 2021-09-29
Add JSON-RPC support for overlay networks

ogenev:
This PR includes:

- JSON-RPC support for overlay networks.

- Refactoring of the main json-rpc handler and service.

~~`HistoryEndpointKind`, `HistoryNetworkEndpoint`, `StateNetworkEndpoint`, `StateEndpointKind`, `PortalEndpointKind`, `PortalEndpoint` are removed in favor of more generic `PortalJsonRequest`, `HistoryJsonRequest` and `StateJsonRequest` which are simple wrappers around `JsonRequest` and a responder. I think this helps us moving the responsibility for handling the requests to the handlers only.~~

In simple terms, the json-rpc server is sending `PortalJsonRpcRequest` to the main handler. Then, based on enum matching in the request endpoint (`discv5_`, `historyNetwork_`, `stateNetwork_`), the main handler dispatches this request to the network specific handlers.  Any request with prefix `discv5_` is handled in the main handler for now.

Every network handler then is responsible for matching the received enum and returns a response.

Additional json-rpc endpoints and infura support will be included in a following PR.

_I'm new to Rust but I was trying to find a way to use a common data structure (trait) for the matching logic in the network handlers (because I assume most of the endpoints will be the same for most of the networks), but without success. 
It seems that, it is not trivial in Rust to use default methods in traits which need to access self data. Some proposed workarounds are to use getter methods in the trait but I wasn't able to make this work, because we have different network instances and the compiler does not know what data they contains when compiling the trait, and we want that trait to be generic. Another problem is that we don't want to import types from the network crates to `trin-core`, because the core crate should not depends on the network crates._

Fixes https://github.com/ethereum/trin/issues/113

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,9 +35,9 @@ jobs:
             - run:
                 name: Run Clippy
                 command: cargo clippy --all -- --deny warnings
-            - rust/build:
-                release: false
-                with_cache: false
+            - run:
+                name: Build Trin workspace
+                command: cargo build --workspace --jobs 2
             - run:
                 name: Test Trin workspace
                 command: cargo test --workspace


njgheorghita (changes requested):
Cool, you made some nice steps here towards coupling the overlay networks and the overlay handlers. But, I have some doubts about getting rid of the enum endpoint types for the overlay networks.
- I think it's a much better mechanism for pattern matching. 
- It makes the code easier to understand. For example, all I have to do is look that the type to immediately see all the supported endpoints, instead of reading through the code to see all the strings that are matched. I believe the usefulness of this increase as we add more and more endpoints (ok to be fair, good documentation is another option). 
- It seems to me like sooner or later, we're going to need a consistent, readable, API between the main jsonrpc handler & the overlay networks, special endpoint types helps enforce this 
- Keep in mind that we also want to support `eth_*` endpoints not just `discv5_*` / `portalHistory_*` / `portalState_*` endpoints. `eth_*` endpoints will be more complex to support, as they'll need to request data from many different overlay networks. And what does the api look like for those requests? 
  - Using `RwLock`s and adding methods to the handlers (eg. `handler.get_account_trie_proof(..)`
  - Using message channels and adding endpoints to the overlay network rpc api (eg. the `EthHandler` might request data via `portalHistory_getAccountTrieProof`)

So, sorry for the rambling, a lot of that is tbd / out of scope for this pr. But in terms of this pr - I'd be interested to hear any pushback you might have on the enum vs string pattern matching.

ogenev:
> * I think it's a much better mechanism for pattern matching.
> * It makes the code easier to understand. For example, all I have to do is look that the type to immediately see all the supported endpoints, instead of reading through the code to see all the strings that are matched. I believe the usefulness of this increase as we add more and more endpoints (ok to be fair, good documentation is another option).
> * It seems to me like sooner or later, we're going to need a consistent, readable, API between the main jsonrpc handler & the overlay networks, special endpoint types helps enforce this

I agree that it will be more clear if we store the json-rpc methods in enums. At the end of the day, every  json-rpc request method should be parsed from string to enum. Probably then the question is where to do this, which object will be responsible for converting from strings to enums, should this be the main json-rpc handler or the network specific handlers?

njgheorghita:
> Probably then the question is where to do this, which object will be responsible for converting from strings to enums with pattern matching? Should this be the main json-rpc handler or the network specific handlers?

It seems to me like this should be handled as soon as possible. The `JsonRpcHandler` should have knowledge of all supported endpoints. If it's supported - pass it off to the respective request handler. If it's not supported - return an "Unsupported endpoint" error. 

Now the best/cleanest way to accomplish this... It's not obvious to me without giving the implementation a shot, but I don't see why it wouldn't be possible

ogenev:
> It seems to me like this should be handled as soon as possible. The `JsonRpcHandler` should have knowledge of all supported endpoints. If it's supported - pass it off to the respective request handler. If it's not supported - return an "Unsupported endpoint" error.

Hm, so probably it makes sense this to be handled in the json-rpc server? And the main handler will expect only known enums?

njgheorghita:
Yup - and while it's probably outside of the scope of this pr, I'd love to see an enum for all the supported Infura endpoints - b/c I think currently, if somebody sends in `trash_endpoint` - it will currently end up going to infura, and returning infura's `"endpoint not supported"` error message. Essentially, the main handler should return an error immediately if a user submits any unsupported endpoint (infura, history, state, discv, eth, ...)

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -2,71 +2,31 @@
...
 }
 
 impl JsonRpcHandler {
     pub async fn process_jsonrpc_requests(mut self) {
         while let Some(cmd) = self.jsonrpc_rx.recv().await {
-            use PortalEndpointKind::*;
-
-            match cmd.kind {
-                NodeInfo => {
+            match cmd.request.method.as_str() {
```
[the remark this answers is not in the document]
jacobkaufmann: I guess we were previously pattern matching strings in `dispatch_portal_request` and doing the enum conversion there. So it's not like we weren't doing this before. We've just removed the intermediate conversion step. I would prefer that we keep a form of enum-based pattern matching, and I think we can keep the code tidy by creating an enum for all possible JSON-RPC endpoints and implementing the [`FromStr`](https://doc.rust-lang.org/std/str/trait.FromStr.html) trait on that enum, and whose error type would be something like `UnsupportedPortalJsonRpcEndpoint`.

When we get a request, we map the method string to the enum with a call to `from_str` and can exit early if it's not an endpoint that's part of the enum.

trin-core/src/jsonrpc/types.rs
```diff
@@ -34,6 +37,24 @@ pub struct JsonRequest {
...
 }
 
+#[derive(Debug, Clone)]
+pub struct PortalJsonRequest {
+    pub request: JsonRequest,
+    pub resp: Responder<Value, String>,
+}
+
+#[derive(Debug)]
+pub struct HistoryJsonRequest {
```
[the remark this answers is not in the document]
jacobkaufmann: Is there any reason to have separate structs for these various requests? All types are equivalent. I guess there is a semantic argument there, but you could just use a type alias if that's the motivation. If the `Responder` type could hold different errors for each network, then that could be addressed with a generic.

trin-history/src/jsonrpc.rs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            self.match_request(&cmd, method).await
+        }
+    }
+
+    async fn match_request(&self, cmd: &HistoryJsonRequest, method: Vec<&str>) {
+        let suffix = method[1];
+
+        match suffix {
+            "dataRadius" => {
```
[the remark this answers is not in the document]
jacobkaufmann: I would suggest same as I commented above.

src/main.rs
```diff
@@ -74,26 +75,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            initialize_history_network(&discovery, portalnet_config.clone(), Arc::clone(&db)).await
         } else {
             (None, None, None, None)
         };
 
     // Initialize service server
-    let (jsonrpc_tx, jsonrpc_rx) = mpsc::unbounded_channel::<PortalEndpoint>();
+    let (portal_jsonrpc_tx, portal_jsonrpc_rx) = mpsc::unbounded_channel::<PortalJsonRpcRequest>();
     let jsonrpc_trin_config = trin_config.clone();
     let web3_server_task = tokio::task::spawn_blocking(|| {
```
njgheorghita, line 86: While you're in here refactoring - can you update this to `jsonrpc_server_task`

src/main.rs
```diff
@@ -74,26 +75,26 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .iter()
             .any(|val| val == HISTORY_NETWORK)
         {
-            initialize_history_network(&discovery, portalnet_config.clone(), &db)
+            initialize_history_network(&discovery, portalnet_config.clone(), Arc::clone(&db)).await
         } else {
             (None, None, None, None)
         };
 
     // Initialize service server
```
njgheorghita, line 83: Also, this comment is confusing.... `// Initialize jsonrpc server` maybe?

trin-core/src/jsonrpc/service.rs
```diff
@@ -219,9 +221,13 @@ fn handle_request(
             "result": "trin 0.0.1-alpha",
         })
         .to_string()),
-        "test_historyNetwork" => dispatch_portal_request(obj, portal_tx),
-        "test_stateNetwork" => dispatch_portal_request(obj, portal_tx),
         _ if obj.method.as_str().starts_with("discv5") => dispatch_portal_request(obj, portal_tx),
```
njgheorghita, line 224: Well, now after all this effort to use enums and improve our parsing - I gotta say it's a bit weird to still have this semi-string parsing step. Do you see any obvious way we could consolidate them?

trin-core/src/jsonrpc/service.rs
```diff
@@ -242,44 +248,29 @@ fn dispatch_infura_request(obj: JsonRequest, infura_url: &str) -> Result<String,
...
-        },
-        "test_stateNetwork" => PortalEndpoint {
-            kind: PortalEndpointKind::DummyStateNetworkData,
-            params,
-            resp: resp_tx,
-        },
-        _ => {
+    let endpoint = PortalEndpointKind::from_str(method);
+
+    match endpoint {
```
njgheorghita, line 257: :heart:

trin-core/src/jsonrpc/types.rs
```diff
@@ -34,6 +38,27 @@ pub struct JsonRequest {
...
+/// History network JSON-RPC request
+#[derive(Debug)]
+pub struct HistoryJsonRpcRequest {
+    pub endpoint: HistoryEndpointKind,
+    pub resp: Responder<Value, String>,
+}
+
+/// State network JSON-RPC request
+#[derive(Debug)]
+pub struct StateJsonRpcRequest {
```
njgheorghita, line 57: Personally, I think these distinct types are fine for now. @jacobkaufmann mentioned some possible consolidation - which I also think is a good idea - similarly I'm sure we could consolidate the `StateEndpointKind::DataRadius`/ `HistoryEndpointKind::DataRadius` in some way - but I really think these are ok, and maybe even preferable in terms of readability.

njgheorghita (approved):
Great! Just a couple tweaks to address. I also think we have some thinking to do around the general `Handler` -> `Network` api - but I'll try and get an issue up soon where we can discuss that

trin-core/src/jsonrpc/service.rs
ogenev: Everything is now send to `dispatch_portal_request`, where the parsing is done. If it is `InfuraEndpointKind`, it is then send to `dispatch_infura_request` for processing. When we start to deal with more `eth_` endpoints, we may refactor this more.

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -6,113 +6,78 @@ use serde_json::Value;
...
-                                tx,
-                                HistoryEndpointKind::GetHistoryNetworkData,
-                            )
+        while let Some(request) = self.portal_jsonrpc_rx.recv().await {
+            match request.endpoint {
+                PortalEndpointKind::Discv5EndpointKind(endpoint) => match endpoint {
+                    Discv5EndpointKind::NodeInfo => {
+                        let node_id = self.discovery.read().await.local_enr().to_base64();
+                        let _ = request.resp.send(Ok(Value::String(node_id)));
+                    }
```
carver, line 33: Hm, the list of `discv5_` things to do could get very long. My instinct is to minimize as much as possible by having each match branch be a one-liner. (Though maybe ) That could look something like:
```rs
            let response = match request.endpoint {
                PortalEndpointKind::Discv5EndpointKind(endpoint) => match endpoint {
                    Discv5EndpointKind::NodeInfo => self.discv5.node_info(),
                    # ...
                }
            }
```

Where `discv5` is a class responsible for implementing all the `discv5_*` handlers, like `node_info()`, which does this:
```rs
let node_id = self.discovery.read().await.local_enr().to_base64();
Ok(Value::String(node_id))
```

I think that refactor translates nicely enough for the other calls that get proxied through to other handlers, too.

carver (approved):
Yup, looks good to me! Just some cleanup suggestions.

jacobkaufmann (approved):
Looks good to me. I don't anything to add on top of the existing comments.

trin-core/src/jsonrpc/handlers.rs
ogenev: No sure if we can move the response outside the `match` block, because It looks like some responses require different `serde_json::Value` wrappers (String, Array, Number). The handler has access to discv5 class (self.discovery), what we can do is to add helper methods in `self.discovery` for things like `node_id`.

PR #121, opened 2021-10-02
doc - update --external-address helpstring

lithp:
I wanted to change the port my node was listening to and after skimming
through the output of --help I tried to do so using the
--external-address flag. It turns out this is not the correct flag to
use!

While the previous help string hinted at the behavior of this flag the
new one should help to prevent future confusion.


ogenev (approved):
LGTM :+1:

PR #123, opened 2021-10-02
refactor - use RwLock::read() when possible

lithp:
@ogenev I *think* this change should leave behavior unchanged, and all the tests pass, but I don't know how comprehensive the tests are. I'm hoping you have time to run this and make sure it still works for you, tagging you because it seems like you wrote most of this code.


ogenev (approved):
Hey, good catch. I thought clippy should catch such unnecessary mutable borrows... 

LGTM :100: , I tested it with the peertest node. Probably this is a sign we need some req/resp end-to-end integration tests for the overlay protocol.

PR #127, opened 2021-10-08
Update localhost address for use in docker container

njgheorghita:
In order to communicate to our http server from outside of a docker container, we need to bind to `0.0.0.0` rather than `127.0.0.1`. Though, I'm not sure if this opens up some security vulnerabilities as the new default? And instead should be controlled via cli flag / env var?


lithp (approved):
The long-term solution is probably to turn `web3_http_port` into `web3_http_addr`, and default to localhost. Right now this is technically insecure because it defaults to turning any trin vulnerability into a vulnerability on a public open port, but as trin is pre-alpha I don't think that's worth slowing yourself down to fix.

PR #132, opened 2021-10-10
Add custom payload element to PING/PONG message calls

ogenev:
This PR:

- Updates current discv5 json-rpc endpoints to match the draft specs in https://github.com/ethereum/portal-network-specs/pull/88
- Bumps `eth2_ssz` and `eth2_ssz_derive` crates versions, so we can introduce `eth2_ssz_types` crate for SSZ types..
- Adds custom payload element to PING/PONG message calls.

  As proposed in https://github.com/ethereum/portal-network-specs/pull/87, the custom payload composes of `data_radius` and `payload` elements. The payload element is SSZ type `List[uint8, max_length=2048]` and it is an optional value, because we may not want to use it when we send underlying discv5 messages (used in `discv5_*` json-rpc endpoints).


mrferris (approved):
Reviewed the 3 changes, LGTM 👍

victor-wei126 (approved):
LGTM as well

PR #134, opened 2021-10-11
Bugfix ipc/http servers

njgheorghita:
It looks like the ipc/http servers were broken recently. I've got them back up and functional locally. My next pr is going to be integrating tests for these into the test harness. Some other things I've addressed in this pr....
- Replace `reqwest` library with `httparse` & `ureq` (`reqwest` is a heavy library and really murders compile time when building docker images)
- Updated the names of the endpoints. IMO `TrinEndpoint` makes sense as the superset of all endpoints that the trin client supports - including those we proxy to infura. `PortalEndpoint` makes sense as the superset of all endpoints that rely on any data actually fetched from the portal network. 
- Better parsing of requests to our http server. This is functional when I test locally, but might be somewhat flaky. I'm going to focus on hammering out edge cases as I add tests to the test harness.

### DIFF
--- a/trin-core/src/jsonrpc/service.rs
+++ b/trin-core/src/jsonrpc/service.rs
@@ -9,7 +9,7 @@ use std::{fs, panic, process};
 #[cfg(unix)]
 use std::os::unix;
 
-use crate::jsonrpc::endpoints::TrinEndpointKind;
+use crate::jsonrpc::endpoints::TrinEndpoint;
 use crate::jsonrpc::types::PortalJsonRpcRequest;
 use httparse;
 use log::{debug, info};
@@ -235,7 +235,7 @@ fn handle_request(
     portal_tx: UnboundedSender<PortalJsonRpcRequest>,
 ) -> Result<String, String> {
     let method = obj.method.as_str();
-    match TrinEndpointKind::from_str(method) {
+    match TrinEndpoint::from_str(method) {
         Ok(val) => dispatch_trin_request(obj, val, infura_url, portal_tx),
         Err(_) => Err(json!({
             "jsonrpc": "2.0",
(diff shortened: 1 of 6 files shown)


ogenev (approved):
LGTM 👍

trin-core/src/jsonrpc/service.rs
```diff
@@ -171,11 +171,13 @@ fn serve_http_client(
     infura_url: &str,
     portal_tx: UnboundedSender<PortalJsonRpcRequest>,
 ) {
-    let mut buffer = Vec::new();
-    stream.read_to_end(&mut buffer).unwrap();
+    let mut reader = io::BufReader::new(&mut stream);
+    let received: Vec<u8> = reader.fill_buf().unwrap().to_vec();
+    // Mark the bytes read as consumed so the buffer will not return them in a subsequent read
+    reader.consume(received.len());
```
carver, line 177: Hm, I don't think this is sufficient. `fill_buf()` doesn't seem to make any guarantees about getting a complete message. It could return in the middle. Maybe `read_to_end()` still makes sense here?

trin-core/src/jsonrpc/service.rs
```diff
@@ -171,11 +171,13 @@ fn serve_http_client(
...
 
-    let json_request = String::from_utf8_lossy(&buffer);
-    let deser = serde_json::Deserializer::from_str(&json_request);
+    let http_body = parse_http_body(received).unwrap();
```
carver, line 179: We definitely shouldn't crash if an invalid http message is sent. Maybe a `warn!` at most.

trin-core/src/jsonrpc/service.rs
```diff
@@ -209,28 +228,49 @@ fn process_http_request(
...
+        TrinEndpoint::StateEndpoint(_) => dispatch_portal_request(obj, endpoint, portal_tx),
     }
 }
 
+// Handle all requests served by infura
 fn dispatch_infura_request(obj: JsonRequest, infura_url: &str) -> Result<String, String> {
     //Re-encode json to proxy to Infura
-    let request = serde_json::to_string(&obj).unwrap();
-    match proxy_to_url(request, infura_url) {
+    match proxy_to_url(&obj, infura_url) {
```
carver, line 273: looks like the comment above this is stale now

trin-core/src/jsonrpc/service.rs
njgheorghita: The problem I was running into with `read_to_end()` is that it hangs until it receives an EOF. Which is not the desired behavior if I'm understanding things correctly. In fact, I'm surprised it was ever functional in the first place.... I do understand the concern with incomplete messages. Do you think the parsing/handling invalid messages in `parse_http_request()` is sufficient? I don't think that we'll expect very large requests that often

carver: Hm, I thought that the deal was that a connection is established, then an http request is sent, then the connection exits. That's why I thought reading to EOF would work. That would not for persistent connections, of course, but I didn't think we needed any support for that. I'll admit plenty of ignorance about the details, though.

Issue #124, opened 2021-10-03
PortalStorage should delete items in a background "tending" thread


mrferris:
PortalStorage currently will delete items if launched with a smaller capacity, it just won't start doing that until the first `store(...)` is called. 

It should instead do that upon initialization so that the first `store(..)` doesn't take a long time due to large deletions.

pipermerriam:
I would suggest for this to be a separate background "tending" process so that neither initialization, or any call to `store` will hang unnecessarily.

PR #136, opened 2021-10-12
Moves PortalStorage capacity management to background thread

mrferris:
This PR changes `PortalStorage`'s `DB` and `rusqlite::Connection` objects to be wrapped in an `Arc<RwLock<>>` instead of just an `Arc<>`, so that the capacity management that `PortalStorage` already does can be done from a separate thread so as to never block on intialization or storage (implements #124)

I'm not 100% on the approach taken thus far. My intuition says that I should keep the storage interface synchronous, so I will likely add wrappers so that `.await` doesn't need to be added to everything.

Further commits will:

- [ ] Implement the capacity management background thread
- [ ] add testing

### DIFF
--- a/trin-core/src/bin/generate-data.rs
+++ b/trin-core/src/bin/generate-data.rs
@@ -5,6 +5,7 @@ use std::convert::TryInto;
 use std::process::Command;
 use std::sync::Arc;
 use structopt::StructOpt;
+use tokio::sync::RwLock;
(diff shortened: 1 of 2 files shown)


mrferris:
Closing, because this approach is likely not the approach we want to go. Instead we likely want to have one thread manage all DB interactions, and each PortalStorage has a channel to send and request data to/from the DB.

PR #137, opened 2021-10-13
ux - do not crash when bootnode bonding times out

lithp:
- begin the slow process of replacing Result<_, String> with meaningful
  error types,
- when one bootnode times out don't crash, move on to the next bootnode

---

I'm not convinced that this is the correct way to slice our errors, very open to suggestions! For example it might be better for there to be a `DiscoveryError` which Overlay passes through to its callers. However, I believe this change is a step in the right direction, just about anything is better than using `String` everywhere!

Before:

```
Oct 12 17:27:48.411 DEBUG discv5::service: Sending RPC Request: id: 28fd4d25334fc35a: TALK: protocol: 7374617465, request: 0101000000000000000c00000001000000ffffffffffffffff0000000000000000000000000000000000000000000000002400000000000000 to node: Node: 0x76f9..6f28, addr: Some(71.202.127.37:4567)
Oct 12 17:27:49.413 DEBUG discv5::service: RPC Request timed out. id: 28fd4d25334fc35a
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: "TalkReq query failed: Timeout"', trin-state/src/lib.rs:120:54
```

After:

```
Oct 12 18:46:12.485 DEBUG discv5::service: Sending RPC Request: id: 8c58437377e21182: TALK: protocol: 7374617465, request: 0101000000000000000c00000001000000ffffffffffffffff0000000000000000000000000000000000000000000000002400000000000000 to node: Node: 0x76f9..6f28, addr: Some(71.202.127.37:4567)
Oct 12 18:46:13.489 DEBUG discv5::service: RPC Request timed out. id: 8c58437377e21182
Oct 12 18:46:13.489 DEBUG trin_state::network:  Timed out while bonding with ENR: NodeId: 0x76f9..6f28, Socket: Some(71.202.127.37:4567)    
```

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -12,12 +12,13 @@ use crate::portalnet::types::CustomPayload;
 use discv5::{
     enr::NodeId,
     kbucket::{Filter, KBucketsTable},
-    TalkRequest,
+    RequestError, TalkRequest,
 };
 use log::debug;
 use rocksdb::DB;
 use std::sync::Arc;
 use std::time::Duration;
+use thiserror::Error;
 use tokio::sync::RwLock;
 
 /// Maximum number of ENRs in response to FindNodes.
@@ -65,6 +66,24 @@ impl Default for OverlayConfig {
             max_incoming_per_bucket: 16,
             table_filter: None,
             bucket_filter: None,
+        }
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum SendPingError {
+    #[error("The request timed out")]
+    Timeout,
+
+    #[error("Internal error: ")]
+    Other(discv5::RequestError),
+}
+
+impl From<discv5::RequestError> for SendPingError {
+    fn from(err: discv5::RequestError) -> Self {
+        match err {
+            discv5::RequestError::Timeout => Self::Timeout,
+            err => Self::Other(err),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -69,6 +70,24 @@ impl Default for OverlayConfig {
...
 }
 
+#[derive(Error, Debug)]
+pub enum SendPingError {
+    #[error("The request timed out")]
+    Timeout,
+
+    #[error("Internal error: ")]
+    Other(discv5::RequestError),
+}
```
jacobkaufmann, line 80: I'm not sure why we would need a specific error type for a particular request type (e.g. ping). I think we can generalize this error to something like the `discv5::RequestError`. I don't see why we need our own timeout error when the `discv5` enum already defines that variant, and when we do not have any of our own timeout logic. I think it would be better to start with the existing `discv5::RequestError` or define a similar general request error enum that wraps the `discv5::RequestError`. As we encounter overlay-specific error types, we can add them to the enum.

njgheorghita (approved):
lgtm! Agreed about phasing out `Result<_, String>` - adding a note to the contributing guidelines.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: I appreciate you taking the time to write out the thorough justification, and I should have taken more time to consider implementation implications. Taking another look at the `discv5::RequestError` enum, I overestimated the number of variants that would be meaningful to consumers of the "send" APIs. And this matching does get quite cumbersome with wrapped errors. While I think there may be other variants that are meaningful to the API consumer (e.g. `InvalidRemoteEnr`), this looks like a good start.

lithp: No worries :) writing it all out was a good exercise, I probably should have added this rationale to the original PR!

PR #139, opened 2021-10-15
Add HashDelayQueue utility

jacobkaufmann:
Add a `HashDelayQueue` struct that can act as a generator that yields "expired" items. This struct is intended for the implementation of routing table management. With a `HashDelayQueue`, we can insert keys with a given timeout or delay, and when polled, the delay queue will yield an expired item (if one exists).

* Create a directory for the `trin_core::utils` module.
* Create a `HashDelayQueue` utility in the `trin_core::utils` module.

### DIFF
--- a/trin-core/src/utils/hash_delay_queue.rs
+++ b/trin-core/src/utils/hash_delay_queue.rs
@@ -1,0 +1,416 @@
+//! A `DelayQueue` with keyed entries implemented with a `HashMap`.
+//!
+//! A `HashDelayQueue` implements `Stream` which removes expired items from the map.
+
+/// The default delay for entries, in seconds. This is only used when `insert()` is used to add
+/// entries.
+const DEFAULT_DELAY: Duration = Duration::from_secs(30);
+
+use futures::prelude::*;
+use std::{
+    collections::HashMap,
+    pin::Pin,
+    task::{Context, Poll},
+    time::Duration,
+};
+use tokio_util::time::delay_queue::{self, DelayQueue};
+
+pub struct HashDelayQueue<K>
+where
+    K: std::cmp::Eq + std::hash::Hash + std::clone::Clone + Unpin,
+{
+    /// The given entries.
+    entries: HashMap<K, delay_queue::Key>,
+    /// A queue holding the timeouts of each entry.
+    expirations: DelayQueue<K>,
+    /// The default expiration timeout of an entry.
+    default_expiration_timeout: Duration,
+}
+
+impl<K> Default for HashDelayQueue<K>
+where
+    K: std::cmp::Eq + std::hash::Hash + std::clone::Clone + Unpin,
+{
+    fn default() -> Self {
+        HashDelayQueue::new(DEFAULT_DELAY)
+    }
+}
+
+impl<K> HashDelayQueue<K>
+where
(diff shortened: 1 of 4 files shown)


trin-core/src/utils/hash_delay_queue.rs
```diff
@@ -0,0 +1,395 @@
...
+        time::pause();
+
+        let mut queue = task::spawn(HashDelayQueue::<String>::new(DELAY));
+        let key = "foo".to_string();
+
+        queue.insert_with_timeout(key.clone(), Duration::ZERO);
+
+        sleep(Duration::from_millis(1)).await;
+
+        let entry = assert_ready_ok_some!(poll!(queue));
```
njgheorghita, line 186: It feels a little strange to me that an `assert_*` statement returns some kind of value... but I see that's also the behavior of some of these tokio `assert_*`s so it's consistent

trin-core/src/utils/hash_delay_queue.rs
```diff
@@ -0,0 +1,395 @@
...
+    time::Duration,
+};
+use tokio_util::time::delay_queue::{self, DelayQueue};
+
+pub struct HashDelayQueue<K>
+where
+    K: std::cmp::Eq + std::hash::Hash + std::clone::Clone + Unpin,
+{
+    /// The given entries.
+    entries: HashMap<K, delay_queue::Key>,
```
njgheorghita, line 23: Maybe expand the comment just a bit here to explain the value

trin-core/src/utils/hash_delay_queue.rs
```diff
@@ -0,0 +1,395 @@
...
+        } else {
+            let delay_key = self.expirations.insert(key.clone(), timeout);
+            self.entries.insert(key, delay_key);
+        }
+    }
+
+    /// Resets the expiration timeout of an entry for `key`. Returns `true` if the key existed,
+    /// `false` otherwise.
+    ///
+    /// # Panics
```
njgheorghita, line 79: Why panic (here and `insert_with_timeout()`  instead of returning a `Result` of some kind?

trin-core/src/utils/hash_delay_queue.rs
```diff
@@ -0,0 +1,395 @@
...
+        let key = "foo".to_string();
+
+        let removed = queue.remove(&key);
+        assert!(!removed);
+    }
+
+    #[tokio::test]
+    async fn clear() {
+        time::pause();
+
```
njgheorghita, line 365: nitpick - but it seems trivial to go ahead and bump up the # of entries here and in `len()` to something like 3. I wouldn't call it necessary, you just never know what bugs are hiding above the `0-1` range

trin-core/src/utils/hash_delay_queue.rs
jacobkaufmann: The panic here is because that is how the [`tokio-util`](https://docs.rs/tokio-util/0.6.8/tokio_util/time/delay_queue/struct.DelayQueue.html#method.insert) library is handling the exception. It looks like the [maximum duration is ~2 years](https://github.com/tokio-rs/tokio/pull/3126), so I think a panic is acceptable here. We should never insert anything with such a long delay.

PR #141, opened 2021-10-15
Emit warnings when locks are held for too long

lithp:
Quick note: I'm not sure who to give this to so I gave it to you @njgheorghita, but feel free to pass it off to somebody else if you think that's appropriate!

This PR is not yet finished, there's some cleanup to do (such as removing code duplication, and logging when `rw_write!()` calls hold onto the lock for too long, but opening a PR now to get feedback on the general approach.

---

- Add rw_read and rw_write macros which wrap access to locks
  - They emit warnings if lock acqusition takes over 100ms
  - They emit warnings if locks are held for over 100ms

- Use rw_read and rw_write macros in most places RwLock's are accessed

Using these macros I was able to immediately identify the cause of a
deadlock during boot.

---

Example output:

```
$ RUST_LOG=debug TRIN_INFURA_PROJECT_ID=XXX cargo run -- --networks state --bootnodes $ENR
Launching trin
[more output]
Oct 15 14:25:25.497 DEBUG trin_state::network: Attempting bond with bootnode ENR: NodeId: 0x76f9..6f28, Socket: Some(71.202.127.37:4567)    
Oct 15 14:25:25.598  WARN trin_core::portalnet::events: [trin-core/src/portalnet/events.rs:31] took more than 100 ms to acquire lock    
Oct 15 14:25:25.598  WARN trin_core: [trin-state/src/network.rs:38] lock held for over 100ms, not yet released    
Oct 15 14:25:25.699  WARN trin_core::portalnet::overlay: [trin-core/src/portalnet/overlay.rs:279] waiting more than 100ms to acquire lock, still waiting    
```

This makes it trivial to figure out where the deadlock is!

### DIFF
--- a/trin-core/src/lib.rs
+++ b/trin-core/src/lib.rs
@@ -3,6 +3,7 @@ extern crate lazy_static;
 
 pub mod cli;
 pub mod jsonrpc;
+pub mod locks;
 pub mod portalnet;
 pub mod socket;
 pub mod utils;
--- a/ethportal-peertest/src/main.rs
+++ b/ethportal-peertest/src/main.rs
@@ -9,6 +9,7 @@ use ethportal_peertest::events::PortalnetEvents;
 use ethportal_peertest::jsonrpc::{
     test_jsonrpc_endpoints_over_http, test_jsonrpc_endpoints_over_ipc,
 };
+use trin_core::locks::RwLoggingExt;
 use trin_core::portalnet::{
     discovery::Discovery,
     overlay::{OverlayConfig, OverlayProtocol},
@@ -30,10 +31,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         };
 
         let discovery = Arc::new(RwLock::new(Discovery::new(portal_config).unwrap()));
-        discovery.write().await.start().await.unwrap();
+        discovery.write_with_warn().await.start().await.unwrap();
 
         let db = Arc::new(setup_overlay_db(
-            discovery.read().await.local_enr().node_id(),
+            discovery.read_with_warn().await.local_enr().node_id(),
         ));
 
         let overlay = Arc::new(
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,6 +7,7 @@ use tokio::sync::RwLock;
 
 use trin_core::jsonrpc::handlers::JsonRpcHandler;
 use trin_core::jsonrpc::types::PortalJsonRpcRequest;
+use trin_core::locks::RwLoggingExt;
 use trin_core::portalnet::events::PortalnetEvents;
 use trin_core::{
     cli::{TrinConfig, HISTORY_NETWORK, STATE_NETWORK},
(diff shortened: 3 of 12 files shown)


trin-core/src/lib.rs
```diff
@@ -6,3 +6,137 @@ pub mod jsonrpc;
...
+                self.acquisition_file,
+                self.acquisition_line,
+                held_for,
+            )
+        }
+    }
+}
+
+#[macro_export]
+macro_rules! rw_read {
```
carver, line 78: If we're going to leave this lock macro in place everywhere (which I'm cool with), then we should probably only run this tooling overhead on non-release builds. Something like branching on `debug_assertions` (I found that trick in [this post](https://stackoverflow.com/a/39205417/8412986)).

carver (approved):
Neat. I didn't do a deep dive on the macro, but generally LGTM 👍🏻

trin-core/src/lib.rs
```diff
@@ -6,3 +6,137 @@ pub mod jsonrpc;
...
+    ($lock_name:expr) => {{
+        let acquire_timeout = std::time::Duration::from_millis($crate::ACQUIRE_TIMEOUT_MS);
+        let sleep = tokio::time::sleep(acquire_timeout);
+        tokio::pin!(sleep);
+
+        let mut did_log: bool = false;
+        let now = std::time::Instant::now();
+
+        loop {
+            tokio::select! {
```
SamWilsn, line 88: Since there's already a dependency on `futures`, might be more understandable to write this using [`futures::future::select`](https://docs.rs/futures/0.3.17/futures/future/fn.select.html)?

lithp: I'll give it a try

SamWilsn: I find the function based approach easier to read, though I don't really have that strong of an opinion on the subject.

trin-core/src/lib.rs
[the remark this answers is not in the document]
lithp: I gave it a try but ended up failing. I don't know enough Rust to feel productive at figuring out what's going on so I think I'll leave this as a macro for now.

When I apply `#[track_caller]` to the method everything builds cleanly but the caller is not tracked. I'm using cargo 1.55.0, and support for `#[track_caller]` in extensions is from 1.46.0, so this *should* work. My best guess is that the trait is somehow interacting badly with the `#[async_trait]` macro I needed to use, I expect that when this trait rewrites the method signature it doesn't faithfully copy over attributes.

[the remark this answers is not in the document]
lithp: Oh, scratch that! I manually expanded the call to `#[async_trait]` and now it works! The code is a bit... circumlocutious though, I think I'll open a second PR for comparison.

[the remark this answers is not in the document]
lithp: The problem seems to be that `#[track_caller]` does not work for async functions. https://github.com/rust-lang/rust/issues/78840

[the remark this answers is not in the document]
SamWilsn: Would something like [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=16c271460b9cbed80ba3e91ce6c669e1) work?

lithp: I experimented with using `debug_assertions`: https://github.com/ethereum/trin/pull/141/commits/0cb24458a48077882f49bc69352a2494516b3c8a

I think I'd prefer to drop this commit though. It makes the code a fair amount less readable, doesn't seem like it drops very much overhead, and prevents us from being able to diagnose deadlocks if someone running a release build tries to send us their logs.

What do you think?

trin-core/src/lib.rs
lithp: I wasn't able to find much of a difference between the two `select!`s but refactored this code a bit by unrolling the loop which should make this code a lot more understandable for first time readers.

PR #143, opened 2021-10-17
Process base layer discv5 RPC messages

ogenev:
This is the beginning of implementing the json-rpc endpoints with highest priority for the test network.

**What is included in this PR?**

- Better message SSZ decode error handling.
- We are now passing the payload of the json-rpc request to the handlers, so we can extract the params.
- Method to process base layer discv5 RPC messages.
- Expose `discv5_findNodes` endpoint.

**What is next?**

We need to validate the parameters of the json-rpc requests. In a following PR, I'm thinking of implementing rpc handler for every endpoint and add common validation methods for the parameters.

### DIFF
--- a/trin-core/src/jsonrpc/handlers.rs
+++ b/trin-core/src/jsonrpc/handlers.rs
@@ -7,7 +7,9 @@ use tokio::sync::mpsc;
 use tokio::sync::RwLock;
 
 use crate::jsonrpc::endpoints::{Discv5Endpoint, HistoryEndpoint, StateEndpoint, TrinEndpoint};
-use crate::jsonrpc::types::{HistoryJsonRpcRequest, PortalJsonRpcRequest, StateJsonRpcRequest};
+use crate::jsonrpc::types::{
+    HistoryJsonRpcRequest, JsonRequest, PortalJsonRpcRequest, StateJsonRpcRequest,
+};
 use crate::portalnet::discovery::Discovery;
 
 type Responder<T, E> = mpsc::UnboundedSender<Result<T, E>>;
@@ -22,8 +24,8 @@ pub struct JsonRpcHandler {
 
 impl JsonRpcHandler {
     pub async fn process_jsonrpc_requests(mut self) {
-        while let Some(request) = self.portal_jsonrpc_rx.recv().await {
-            let response: Value = match request.endpoint {
+        while let Some(portal_jsonrpc_req) = self.portal_jsonrpc_rx.recv().await {
+            let response: Value = match portal_jsonrpc_req.endpoint {
                 TrinEndpoint::Discv5Endpoint(endpoint) => match endpoint {
                     Discv5Endpoint::NodeInfo => self.discovery.read().await.node_info(),
                     Discv5Endpoint::RoutingTableInfo => {
@@ -32,34 +34,42 @@ impl JsonRpcHandler {
                 },
                 TrinEndpoint::HistoryEndpoint(endpoint) => {
                     let response = match self.history_jsonrpc_tx.as_ref() {
-                        Some(tx) => proxy_query_to_history_subnet(tx, endpoint).await,
+                        Some(tx) => {
+                            proxy_query_to_history_subnet(portal_jsonrpc_req.request, tx, endpoint)
+                                .await
+                        }
                         None => Err("Chain history subnetwork unavailable.".to_string()),
                     };
                     response.unwrap_or_else(Value::String)
                 }
                 TrinEndpoint::StateEndpoint(endpoint) => {
                     let response = match self.state_jsonrpc_tx.as_ref() {
-                        Some(tx) => proxy_query_to_state_subnet(tx, endpoint).await,
+                        Some(tx) => {
+                            proxy_query_to_state_subnet(portal_jsonrpc_req.request, tx, endpoint)
+                                .await
+                        }
                         None => Err("State subnetwork unavailable.".to_string()),
                     };
                     response.unwrap_or_else(Value::String)
                 }
                 _ => Value::String(format!(
                     "Can't process portal network endpoint {:?}",
-                    request.endpoint
+                    portal_jsonrpc_req.endpoint
                 )),
             };
-            let _ = request.resp.send(Ok(response));
+            let _ = portal_jsonrpc_req.resp.send(Ok(response));
         }
     }
 }
 
 async fn proxy_query_to_history_subnet(
+    request: JsonRequest,
     subnet_tx: &mpsc::UnboundedSender<HistoryJsonRpcRequest>,
     endpoint: HistoryEndpoint,
 ) -> Result<Value, String> {
     let (resp_tx, mut resp_rx) = mpsc::unbounded_channel::<Result<Value, String>>();
     let message = HistoryJsonRpcRequest {
+        request,
         endpoint,
         resp: resp_tx,
     };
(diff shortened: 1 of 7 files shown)


trin-core/src/portalnet/discovery.rs
```diff
@@ -139,6 +144,18 @@ impl Discovery {
...
+        let msg = FindNodes { distances };
+
+        Ok(self
+            .send_talkreq(
+                enr,
+                DISCV5_PROTOCOL.to_owned(),
+                Message::Request(Request::FindNodes(msg)).to_bytes(),
+            )
+            .await?)
+    }
```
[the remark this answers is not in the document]
njgheorghita: I'd agree, also it appears that the base `find_node` method returns a `Vec<Enr>` which is preferable to `Vec<u8>`

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -22,44 +24,56 @@ pub struct JsonRpcHandler {
 
 impl JsonRpcHandler {
     pub async fn process_jsonrpc_requests(mut self) {
-        while let Some(request) = self.portal_jsonrpc_rx.recv().await {
-            let response: Value = match request.endpoint {
+        while let Some(portal_jsonrpc_req) = self.portal_jsonrpc_rx.recv().await {
+            let response: Value = match portal_jsonrpc_req.endpoint {
                 TrinEndpoint::Discv5Endpoint(endpoint) => match endpoint {
```
njgheorghita, line 29: Assuming that we plan to support more `discv5_*` endpoints, it's probably about time to break out this match into its own function that handles the fetching of data for `Discv5Endpoint`. Maybe adding something like a `Discv5RequestHandler` to maintain consistency with the other subdomains?

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -22,44 +24,56 @@ pub struct JsonRpcHandler {
...
                     Discv5Endpoint::NodeInfo => self.discovery.read().await.node_info(),
                     Discv5Endpoint::RoutingTableInfo => {
                         self.discovery.write().await.routing_table_info()
                     }
+                    Discv5Endpoint::FindNodes => {
+                        // TODO: Add RPC handler for every endpoint and validate parameters
+                        Value::String("Disscv5 FindNodes endpoint not implemented yet!".to_owned())
```
njgheorghita, line 36: Couldn't we just call `discovery.find_node(..)` here? Also minor typo `Disscv5`. Also, unless I'm missing something... it does seem like this endpoint is implemented, why are we returning this message?

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -22,44 +24,56 @@ pub struct JsonRpcHandler {
...
-            let response: Value = match request.endpoint {
+        while let Some(portal_jsonrpc_req) = self.portal_jsonrpc_rx.recv().await {
+            let response: Value = match portal_jsonrpc_req.endpoint {
                 TrinEndpoint::Discv5Endpoint(endpoint) => match endpoint {
                     Discv5Endpoint::NodeInfo => self.discovery.read().await.node_info(),
                     Discv5Endpoint::RoutingTableInfo => {
                         self.discovery.write().await.routing_table_info()
                     }
+                    Discv5Endpoint::FindNodes => {
+                        // TODO: Add RPC handler for every endpoint and validate parameters
```
njgheorghita, line 35: I do agree, parameter validation for all incoming requests is important and needs to be done. I'm a bit confused what you mean by add RPC handler for every endpoint? As I understand, in this repo the term "handler" is meant for intermediary objects that receive, match and fulfill rpc requests to their domain `portalState_*`, `portalHistory_*` etc.

trin-core/src/portalnet/types.rs
```diff
@@ -9,11 +9,39 @@ use ssz;
 use ssz::{Decode, DecodeError, Encode, SszDecoderBuilder, SszEncoder};
 use ssz_derive::{Decode, Encode};
 use ssz_types::{typenum, VariableList};
+use thiserror::Error;
 
 use super::{Enr, U256};
 
 type ByteList = VariableList<u8, typenum::U2048>;
 
+#[derive(Error, Debug)]
```
njgheorghita, line 18: Nice errors 🥳

njgheorghita (commented):
Just a couple of thoughts/questions

trin-core/src/portalnet/discovery.rs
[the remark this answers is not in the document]
ogenev: To my understanding, discv5's [find_node ](https://github.com/sigp/discv5/blob/master/src/discv5.rs#L484) public method returns nodes closest to the given NodeId, which in the background is calling `nodes_by_distances` (for reference check [here](https://github.com/sigp/discv5/blob/f11a60a678c285428d47dd2b24db57633f0b3346/src/service.rs#L524) and [here](https://github.com/sigp/discv5/blob/f11a60a678c285428d47dd2b24db57633f0b3346/src/service.rs#L904)). It is not possible to pass a list of distances as parameter to this method.

On the other hand, `discv5_foundNodes` [RPC](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ogenev/portal-network-specs/jsonrpc-assembled-specs/jsonrpc-specs/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) requires from us to send a request to a target ENR with a list of a distances and is expecting to receive all nodes in the given distances from the routing table of the target node, which lead us to use the public discv5 [nodes_by_distances](https://docs.rs/discv5/0.1.0-beta.11/src/discv5/discv5.rs.html#223-245) method.

Because I think we can't use directly a discv5 `FIndNode` message, we are sending `FindNodes` request via TALKREQ and  use `discv5` as protocol identifier, this way, the main handler can distinguish between overlay and base layer discv5 `FindNodes`.

Does this makes sense? Please correct me if i'm wrong, It is possible that I'm misunderstanding something. It is kinda confusing why we use `FindNodes` for overlay, and the base discv5 message is `FindNode` (without 's).

trin-core/src/jsonrpc/handlers.rs
ogenev: By handlers for every endpoint, I mean a data structures responsible for doing the endpoint validations similar to those in [ddht](https://github.com/ethereum/ddht/blob/master/ddht/v5_1/rpc_handlers.py).

trin-core/src/jsonrpc/handlers.rs
ogenev: I left this as a placeholder because we still don't have validations for the parameters, but probably we can still call discovery.find_nodes anyways.

trin-core/src/portalnet/discovery.rs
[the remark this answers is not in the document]
ogenev: I think the problem we are facing here is that the public [ find_node](https://docs.rs/discv5/0.1.0-beta.11/discv5/struct.Discv5.html#method.find_node) method exposed by discv5 implementation in rust is not the actual FindNode RPC from the [specs](https://docs.rs/discv5/0.1.0-beta.11/discv5/struct.Discv5.html#method.find_node).
 The public `find_node` method is returning nodes closer to a given NodeId from the local routing table. I think the FindNode RPC message from the specs is not exposed in the public API [here](https://docs.rs/discv5/0.1.0-beta.11/discv5/struct.Discv5.html#method.find_node) (same for the PING message).

PR #144, opened 2021-10-18
reorganize utils/mod.rs into separate utils/db.rs and utils/distance.rs files

oslfmt:
This PR separates `utils.rs` in `trin-core` into separate files, the reason is because at this point `utils.rs` is becoming a dump place for any miscellaneous functions, and as the code base grows it will be helpful to logically separate this out for better organization. So what I've done is create a new `utils` crate within trin-core, and this crate contains:

- db.rs (contains helper fnc related to storage functionality)
- misc.rs (this just contains the xor function, maybe a better name?)

In a coming PR, I will add another file into this crate `content_keys.rs` which contains some helper functions relating to the content-key SSZ definitions and derivation functions


mrferris (changes requested):
Hey, can you rebase with master and then build on top of [Jacob's recently merged PR](https://github.com/ethereum/trin/commit/dbf430f90b435b15d53543a9b50b3180ff9f9814), which also adds a `utils/` folder? 

I think that the `db.rs` file is great, though I would change the `misc.rs` to `distance.rs`.

Issue #147, opened 2021-10-19
Overlay network pings from test harness are not pinging


njgheorghita:
Both pings to each overlay network from the test harness seem to be broken. Hopefully #141 will help with debugging the issue.

PR #148, opened 2021-10-19
Add http/ipc jsonrpc testing to harness

### LINKED ISSUE #4: Initial Integration tests for HTTP / IPC servers

njgheorghita:
First steps towards a more robust test harness. For some reason, the overlay network pings seem to be broken. I dug into it a bit, but couldn't quite figure out the solution. It also seems like either https://github.com/ethereum/trin/pull/141 might help the situation, or at least help with debugging the situation. Tracking that in this issue https://github.com/ethereum/trin/issues/147

Fixes #4

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -1,0 +1,131 @@
+use std::io::prelude::*;
+use std::os::unix::net::UnixStream;
+use std::slice::Iter;
+
+use hyper::{self, Body, Client, Method, Request};
+use log::info;
+use serde_json::{self, Value};
+
+use trin_core::portalnet::U256;
+
+#[derive(Copy, Clone)]
+pub struct JsonRpcEndpoint {
+    pub method: &'static str,
+    pub id: &'static u8,
+}
+
+const ALL_ENDPOINTS: [JsonRpcEndpoint; 6] = [
+    JsonRpcEndpoint {
+        method: "web3_clientVersion",
+        id: &0,
+    },
+    JsonRpcEndpoint {
+        method: "discv5_nodeInfo",
+        id: &1,
+    },
+    JsonRpcEndpoint {
+        method: "discv5_routingTableInfo",
+        id: &2,
+    },
+    JsonRpcEndpoint {
+        method: "eth_blockNumber",
+        id: &3,
+    },
+    JsonRpcEndpoint {
+        method: "portalHistory_dataRadius",
+        id: &4,
+    },
+    JsonRpcEndpoint {
+        method: "portalState_dataRadius",
+        id: &5,
+    },
+];
+
+fn validate_endpoint_response(method: &str, result: &Value) {
+    match method {
+        "web3_clientVersion" => {
+            assert_eq!(result.as_str().unwrap(), "trin 0.0.1-alpha");
+        }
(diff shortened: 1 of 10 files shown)


ethportal-peertest/src/jsonrpc.rs
```diff
@@ -0,0 +1,131 @@
...
+use hyper::{self, Body, Client, Method, Request};
+use log::info;
+use serde_json::{self, Value};
+
+use trin_core::portalnet::U256;
+
+#[derive(Copy, Clone)]
+pub struct JsonRpcEndpoint {
+    pub method: &'static str,
+    pub id: &'static u8,
```
jacobkaufmann, line 14: Is the `id` just a convenience field so that we do not need to generate a unique `id` for each JSON-RPC request that we test?

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+pub async fn test_jsonrpc_endpoints_over_http() {
+    let client = Client::new();
+    for endpoint in JsonRpcEndpoint::all_endpoints() {
+        info!("Testing over HTTP: {:?}", endpoint.method);
+        let json_string = format!(
+            r#"{{"jsonrpc":"2.0","id":0,"method":"{}","params":[]}}"#,
+            endpoint.method
+        );
```
jacobkaufmann, line 118: Why not use `endpoint.to_jsonrpc()` here?

ethportal-peertest/src/jsonrpc.rs
njgheorghita: Yup, exactly. Not strictly necessary to do here, but since `params` will also soon be added to the struct, seemed harmless to stick these in there as well

jacobkaufmann:
@njgheorghita Gotcha. Maybe we can do something similar to the `validate` attribute that we use in `trin-core/jsonrpc`. I'll do some exploring. I think it's okay to track separately for now. I do think we should add it before or along with the next set of endpoints that we introduce to the test harness, before the set of endpoints gets much larger.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -0,0 +1,158 @@
...
+            }}"#,
+            self.id, self.method
+        )
+    }
+}
+
+pub async fn test_jsonrpc_endpoints_over_ipc() {
+    for endpoint in JsonRpcEndpoint::all_endpoints() {
+        info!("Testing over IPC: {:?}", endpoint.method);
+        let mut stream = UnixStream::connect("/tmp/trin-jsonrpc.ipc").unwrap();
```
ogenev, line 98: If we want to keep the direction to support cross-client testing in this crate, probably it is worth including the socket path as a parameter to CLI. I guess we can keep the default value as `/tmp/trin-jsonrpc.ipc`.

PR #149, opened 2021-10-20
Programmatically track version number for web3_clientVersion

njgheorghita:
Automatically update trin version in `web3_clientVersion` response


jacobkaufmann (approved):
Looks like a straightforward change. I guess this will use the version from `trin-core` as opposed to the top-level `trin`? I don't think that's a problem, but one question that came up is whether we expect all of these packages to upgrade to new versions together?

PR #161, opened 2021-10-27
Use newly-released eth-trie

carver:
Include the new eth-trie, and add a `TrieDB` that implements the trie's DB trait for rocksDB.

### DIFF
--- a/trin-state/src/trie.rs
+++ b/trin-state/src/trie.rs
@@ -1,0 +1,31 @@
+use std::sync::Arc;
+
+pub struct TrieDB {
+    db: Arc<rocksdb::DB>,
+}
+
+impl TrieDB {
+    pub fn new(db: Arc<rocksdb::DB>) -> TrieDB {
+        TrieDB { db }
+    }
+}
+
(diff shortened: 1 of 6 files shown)


trin-state/src/trie.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn insert(&self, key: &[u8], value: Vec<u8>) -> Result<(), Self::Error> {
+        self.db.put(key, value)
+    }
+
+    fn remove(&self, key: &[u8]) -> Result<(), Self::Error> {
+        self.db.delete(key)
+    }
+
+    fn flush(&self) -> Result<(), Self::Error> {
+        Ok(())
```
lithp, line 29: woah! does this mean rocksdb does a commit / fsyncs after every operation?

lithp (approved):
looks good to me!

trin-state/src/trie.rs
carver: Yeah, there is some way to do a batch write operation in their API, but I haven't used it yet.

PR #171, opened 2021-11-01
Update web3-http-port to web3-http-address

### LINKED ISSUE #129: web3_http_address should replace web3_http_port in TrinConfig
From #127:

@njgheorghita:
> In order to communicate to our http server from outside of a docker container, we need to bind to 0.0.0.0 rather than 127.0.0.1. Though, I'm not sure if this opens up some security vulnerabilities as the new default? And instead should be controlled via cli flag / env var?

@lithp:
> The long-term solution is probably to turn web3_http_port into web3_http_addr, and default to localhost. Right now this [hardcoding `0.0.0.0` instead of allowing the entire ip + port be configurable] is technically insecure because it defaults to turning any trin vulnerability into a vulnerability on a public open port.

njgheorghita:
Updates cli arg `web3-http-port` to `web3-http-address` - useful for use with docker container to allow using `0.0.0.0` as the ip address. 

Fixes #129

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ OPTIONS:
         --pool-size <pool_size>               max size of threadpool [default: 2]
         --unsafe-private-key <private_key>    Hex encoded 32 byte private key (considered unsafe to pass in pk as cli
                                               arg, as it's stored in terminal history - keyfile support coming soon)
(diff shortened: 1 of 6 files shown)


jacobkaufmann (approved):
LGTM 👍

PR #174, opened 2021-11-03
Fixed two issues found in uTP impl

KolbyML:
@KonradStaniec notified me of issues he found during his testing of the interop between fluffy and trin.

Here is a PR of the changes he requested and a few other spots I found the same issues he pointed out.

### DIFF
--- a/trin-core/src/portalnet/utp.rs
+++ b/trin-core/src/portalnet/utp.rs
@@ -677,7 +677,7 @@ impl UtpStream {
     async fn send_finalize(&self) {
         let mut response = PacketHeader::new();
         response.set_type(Type::StReset);
-        response.set_connection_id(self.conn_id_recv);
+        response.set_connection_id(self.conn_id_send);
         response.set_timestamp(get_time());
         response.set_timestamp_difference(0);
         response.set_seq_nr(self.seq_nr + 1);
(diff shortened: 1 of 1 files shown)


pipermerriam (commented):
Is there a way to write a test for this?  It is often good practice when addressing a bug to first write a test that demonstrates the failure and then to fix it.  Otherwise, it's not always clear that the issue has actually been resolved correctly.

KolbyML:
@pipermerriam I believe the only way to write tests for this would be to add integration tests. I remember I wanted to write some when I was making the PR, but there was some issue like it would be better to write an integration testing framework first or something.

ogenev:
> @pipermerriam I believe the only way to write tests for this would be to add integration tests. I remember I wanted to write some when I was making the PR, but there was some issue like it would be better to write an integration testing framework first or something.

We have now the test harness (ethportal-peertest), I think we can use it for those kinds of cross-clients integration tests. We still need to decide on the implementation details but the proposition is to have for example cross-client utp test suite which we can run in CicrleCI or localy. What do you think @njgheorghita?

njgheorghita:
It seems like we want two different sets of integration tests, one for trin and one for cross-client. There's a ton of overlap between the two, so it makes sense to build them in the same place. But, i'd recommend keeping the cross-client tests under a separate flag at least during the early stages of development

KolbyML:
@ogenev @njgheorghita to be clear I would want integration tests for Trin so a Trin client would connect to a Trin client. So Nick seemed to pick up on that. For early testing just having an integration testing framework (where we could test Trin to itself to clarify) would be great because once we have the deadlock problem solve issues like that can be avoided, also we could easily test whether things like uTP work properly.

Anyways I think this discussion should probably become an issue and if there are no issues with the PR itself it can be merged.

njgheorghita:
@KolbyML Sorry, I missed this. Yeah, let's get this merged. I added a task for utp-specific tests in #185 so we don't drop that thread

PR #175, opened 2021-11-03
Add dockerfile

njgheorghita:
Add dockerfile. There's still some optimization that can be done here, particular with caching the layer for compiling dependencies. However, I wasn't able to figure out how to get this working with multiple workspaces, and after enough time sitting through docker build steps, I figured it's not a huge priority. A possible solution seems to be [cargo-chef](https://www.lpalmieri.com/posts/fast-rust-docker-builds/), but it's maybe overkill for the time being.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+# select build image
+FROM rust as builder
+
+# create a new empty shell project
+RUN USER=root cargo new --bin trin
+WORKDIR /trin
+
+RUN apt-get update && apt-get install clang -y
+RUN rustup component add rustfmt
+
+# copy over manifests and source to build image
+COPY ./Cargo.lock ./Cargo.lock
+COPY ./Cargo.toml ./Cargo.toml
+COPY ./src ./src 
+COPY ./trin-core ./trin-core 
+COPY ./trin-history ./trin-history 
+COPY ./trin-state ./trin-state 
+COPY ./ethportal-peertest ./ethportal-peertest 
+
+# build for release
+RUN cargo build --release
+
+# final base
+FROM rust
+
+# copy build artifact from build stage
+COPY --from=builder /trin/target/release/trin .
+
+ENV RUST_LOG=debug
+
+ENTRYPOINT ["./trin"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,11 @@
 # select build image
-FROM rust as builder
+FROM rust:1.56.1 AS builder
 
 # create a new empty shell project
 RUN USER=root cargo new --bin trin
 WORKDIR /trin
 
 RUN apt-get update && apt-get install clang -y
-RUN rustup component add rustfmt
 
 # copy over manifests and source to build image
 COPY ./Cargo.lock ./Cargo.lock
@@ -21,7 +20,7 @@ COPY ./ethportal-peertest ./ethportal-peertest
 RUN cargo build --release
 
 # final base
-FROM rust
+FROM rust:1.56.1
 
 # copy build artifact from build stage
 COPY --from=builder /trin/target/release/trin .


Dockerfile
```diff
@@ -0,0 +1,31 @@
+# select build image
+FROM rust as builder
```
carver, line 2: This is neat, I hadn't seen multi-stage docker builds before (granted I haven't really used docker for maybe ~6 years).

Based on [this link](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages), and my intuition, the `as` would be clearest to be capitalized to highlight it as a Docker keyword.
```suggestion
FROM rust AS builder
```

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# create a new empty shell project
+RUN USER=root cargo new --bin trin
+WORKDIR /trin
+
+RUN apt-get update && apt-get install clang -y
+RUN rustup component add rustfmt
```
carver, line 9: Just curious: why do we need rustfmt in this docker builder?

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# copy over manifests and source to build image
+COPY ./Cargo.lock ./Cargo.lock
+COPY ./Cargo.toml ./Cargo.toml
+COPY ./src ./src 
+COPY ./trin-core ./trin-core 
+COPY ./trin-history ./trin-history 
+COPY ./trin-state ./trin-state 
+COPY ./ethportal-peertest ./ethportal-peertest
```
carver, line 18: This is going to bite us roughly every time a new crate is added... I'm wondering if there's any better way, like just copying the whole directory minus `.git`. This isn't a hard blocker, just something to consider.

carver (approved):

Dockerfile
```diff
@@ -0,0 +1,31 @@
+# select build image
+FROM rust as builder
+
+# create a new empty shell project
+RUN USER=root cargo new --bin trin
```
ogenev, line 5: Is there any reason not to use `COPY . .` here instead of creating an empty project and then copying all folders one by one? We can use also `.dockerignore` [file](https://docs.docker.com/engine/reference/builder/#dockerignore-file) if we want to exclude something.

ogenev (approved):

njgheorghita:
@ogenev @carver re: copying over the source folders individually. There are two reasons for the current setup...

- It's a bit of a hangover from a more advanced setup I was trying for caching dependencies, where you have to copy over the manifests from each workspace individually. To be fair, we're not using that setup so it's no longer a consideration (though, I could see us using it in the future to optimize docker builds)
- I have a lot of extra files in my trin folder, python scripts, rocks databases, md files, etc. This is the easiest way to explicitly copy over only the necessary files. But, I also understand that this isn't a great justification for the complexity

So, those are the reasons. Neither of which are convincing. But, also part of me likes the explicit copy. It's somewhat wordy, but avoids the scenario of accidentally building the image with a `.env` file or some other file type that you forgot wasn't accounted for in `.dockerignore`. The most convincing argument for `.dockerignore` is what Carver mentioned, but docker will yell at you immediately during the build while you're compiling that it's missing those files, and then it's simply adding another line. 

But, if you agree / disagree, lmk and I can do the `.dockerignore` no problem

Dockerfile
njgheorghita: Good catch, this was necessary in an earlier, fancier setup I was attempting when I was running `cargo build` in default mode (which required rustfmt to compile) - but it's no longer necessary so I've removed it

carver:
Sure, it's a little messy, but I'm okay leaving the PR as-is.

It's compelling that the most likely failure mode would be: immediate build failure with somewhat clear message about missing a crate.

That's much better than the alternative: probably not notice that a bunch of extra stuff gets included in the build, and a bunch of unnecessary info gets pushed out to all the testnet nodes.

PR #176, opened 2021-11-04
Update protocol id match in test harness handler

ogenev:
Update test harness handler with protocolid introduced in https://github.com/ethereum/trin/pull/162 and bump all crates rust version to 1.56.1 (related to https://github.com/ethereum/trin/pull/173).

### DIFF
--- a/ethportal-peertest/src/events.rs
+++ b/ethportal-peertest/src/events.rs
@@ -1,15 +1,13 @@
 use discv5::{Discv5Event, TalkRequest};
+use hex;
 use log::{debug, error, warn};
 use std::convert::TryInto;
+use std::str::FromStr;
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use trin_core::portalnet::overlay::OverlayProtocol;
-use trin_core::portalnet::types::Message;
+use trin_core::portalnet::types::{Message, ProtocolId};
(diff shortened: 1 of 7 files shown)


jacobkaufmann (approved):

ethportal-peertest/src/events.rs
```diff
@@ -1,10 +1,12 @@
...
 use std::convert::TryInto;
+use std::str::FromStr;
 use std::sync::Arc;
 use tokio::sync::mpsc;
 use trin_core::portalnet::overlay::OverlayProtocol;
-use trin_core::portalnet::types::Message;
+use trin_core::portalnet::types::{Message, ProtocolId};
 use trin_core::portalnet::utp::UtpListener;
 
 pub const HISTORY_NETWORK: &str = "history";
```
KolbyML, line 12: If we are now using ProtocolId these constants for history, state, and utp should be removed since uTP is being used instead now.

KolbyML (changes requested):

ethportal-peertest/src/events.rs
ogenev: Right, good catch! I'm curious why clippy doesn't yell at me.

njgheorghita (approved):
:ship:

KolbyML (approved):

PR #178, opened 2021-11-05
Run Discovery and StateNetwork without locks

carver:
So I've been trying to get trin to talk to itself. I was running into something that looks like a deadlock. Locks, especially spread across a codebase, are a huge PITA to debug. Rather than do the first of many painful lock debug sessions, I wanted to see what it would look like to get rid of them.

It may be doable, but let's chat about it!

---

This requires some API updates to sigp/discv5. The good news is that the change is just to re-frame a bunch of methods as borrow instead of mutable-borrow. Then sharing the discovery instance will be trivial.

This seems to be a trivial change to discv5, because most of the internal variables are in a `RwLock` anyway.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -497,7 +497,7 @@ dependencies = [
 [[package]]
 name = "discv5"
 version = "0.1.0-beta.10"
-source = "git+https://github.com/sigp/discv5#10247bbd299227fef20233f2f5a8de9780de09ac"
+source = "git+https://github.com/carver/discv5?branch=fewer-mutable-borrows#7cd3f424f1b9e2e30e84c019ba8f3a2469df3ead"
 dependencies = [
  "aes",
  "aes-gcm",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,8 @@ trin-state = { path = "trin-state" }
 trin-cli = { path = "trin-cli" }
 
 [dependencies.discv5]
-version = "0.1.0-beta.10"
-git = "https://github.com/sigp/discv5"
+branch = "fewer-mutable-borrows"
+git = "https://github.com/carver/discv5"
 
 [workspace]
 members = [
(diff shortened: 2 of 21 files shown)


lithp:
Want to look at it a little more closely but on first pass this seems great! I didn't even think to check whether we were overborrowing like this!

One concern I have is that at some point we might genuinely want to call a `&mut self` method on the `Discv5`, and then we'll need to add all these locks right back in. How likely do you think that is to happen?

ogenev:
Looks good to me, I hadn't thought to check why we need to pass a mutable reference but it seems most of those `DIscv5` methods do not mutate the struct. Curious the see what the library maintainers think about this.

njgheorghita:
I've also run into a significant amount of lockups running trin & the test harness, and would love to see us remove any number of them.

carver:
> One concern I have is that at some point we might genuinely want to call a `&mut self` method on the `Discv5`, and then we'll need to add all these locks right back in. How likely do you think that is to happen?

Right, I don't have a strong intuition about this. My simple logic is that they seem to have already built `Discv5` to be threadsafe by using `RwLock` internally on any inner variable mutations. So they (and we) probably want to keep it that way for now.

> Curious the see what the library maintainers think about this.

Me too! I'm curious about @SamWilsn 's thoughts on these supporting changes to `Discv5`:
https://github.com/sigp/discv5/pull/92

- Is there some better way to remove the discovery locks in `trin`?
- Is the change to discv5 a reasonable thing to do?
- Is there some better way to accomplish this effect in discv5?
- Is it reasonable to expect that we can keep (most) calls to `Discv5` as "non-mutable," as a way to convey that they are threadsafe operations?

carver:
Ok, since this is fixing basic functionality, I want to get it in quickly. If anyone has any more feedback, I'll put it in a new PR asap.

PR #179, opened 2021-11-08
Support for general overlay network jsonrpc endpoints

njgheorghita:
Added support for `overlay_*` jsonrpc endpoints to the jsonrpc handler

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -13,7 +13,7 @@ use trin_core::portalnet::U256;
 pub struct JsonRpcEndpoint {
     pub method: &'static str,
     pub id: &'static u8,
-    pub params: Params
+    pub params: Params,
 }
 
 fn validate_endpoint_response(id: &u8, result: &Value) {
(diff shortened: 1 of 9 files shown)


ethportal-peertest/src/jsonrpc.rs
```diff
@@ -1,85 +1,90 @@
...
     };
-    info!("{:?} returned a valid response.", method);
+    info!("RPC endpoint: id #{:?} returned a valid response.", id);
 }
 
 impl JsonRpcEndpoint {
-    pub fn all_endpoints() -> Iter<'static, Self> {
-        ALL_ENDPOINTS.iter()
+    pub fn all_endpoints() -> Vec<JsonRpcEndpoint> {
+        vec![
```
njgheorghita, line 53: Moved these endpoints out of the `CONST` in anticipation that we will soon need dynamic parameters (i.e. enrs for pinging)

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -33,61 +36,97 @@ impl JsonRpcHandler {
...
+    }
+
+    async fn process_overlay_request(
+        &self,
+        endpoint: OverlayEndpoint,
+        params: Params,
+    ) -> Result<Value, JsonRpcError> {
+        match params.clone() {
+            Params::Array(val) => match &val[0].as_str() {
+                Some("history") => match self.history_jsonrpc_tx.as_ref() {
```
ogenev, line 62: Is it worth adding an enum for all possible network identifiers ("state", 'history", .. ) and then matching it here instead of using strings?

trin-core/src/jsonrpc/handlers.rs
```diff
@@ -33,61 +36,97 @@ impl JsonRpcHandler {
...
+                    Some(tx) => proxy_query_to_history_subnet(tx, endpoint, params.clone()).await,
+                    None => Err(JsonRpcError::UnavailableNetwork("history".to_string())),
+                },
+                Some("state") => match self.state_jsonrpc_tx.as_ref() {
+                    Some(tx) => proxy_query_to_state_subnet(tx, endpoint, params).await,
```
ogenev, line 67: Do you think it is a good idea instead of passing the same `Params` enum (with the network identifier) to try to exclude the first value (the network identifier) before sending it to the overlay handlers? I think this may introduce some redundancy later in the data flow when we will be interested only in the actual *params.

I'm thinking about two ways to accomplish this:

- Add additional `OverlayParams` enum and a `try_from` trait for converting from `Params`. Then pass `OverlayParams` to the overlay handlers.
- Use the same `Params` enum but implement an additional method ("to_overlay_params" or something similar) which will remove the first network identifier element.

ogenev (approved):

carver (changes requested):
Seems like we're heading in a different direction for the method names, based on chatter in Discord.

PR #169, opened 2021-10-28
WIP - Add little command for easy access to JSON-RPC interface

lithp:
Very WIP, current interface:

```
brian@beat:~/ef/trin$ cargo run -p trin-cli -- rpc discv5_routingTableInfo
RoutingTableInfoResponse { buckets: [RoutingTableNodeInfo { node_id: "0xd40a..2cb0", enr: "enr:-IS4QGAwk1EybH8deeq9xrlH6HF_ZnWAu4vMODJbgZ0i78_IUnN0vdQgBpJ0p4ZXWD6FNq_k-kP1-9ZpHCqvIQphtoMBgmlkgnY0gmlwhH8AAAGJc2VjcDI1NmsxoQL0kZO-svDs-h0vbTo-QsWhJ7XeRmaQH0g8jqej35K0t4N1ZHCCEdg", status: "Connected" }], local_node_id: "0x302c..1cd1" }
```

Would appreciate suggestions:
- `trin-cli rpc discv5_routingTableInfo` is honest but... still not a very good interface. How could it be improved?
  - what should this command be called?
  - should commands be given short names like `routingTable`?

There's still work to do:
- [ ] Refactor, move a lot of this logic into `trin_core::jsonrpc`
- [ ] Refactor `peertest` (clientside) to use these structs
  - [ ] This might require adding support for connecting over HTTP
- [ ] Refactor the handlers (serverside) to use these nice serializable structs instead of manually generating responses
- [ ] Come up with a clean way to map the user-supplied and dynamically typed argv into method calls
- [ ] Add support for the rest of the JSON-RPC endpoints (I'll probably defer that work for future PRs)
- [ ] Implement a `trin-cli endpoints` which lists available endpoints (maybe this is just `trin-cli help` and I let structopt do most of the work)

PR #180, opened 2021-11-09
Add little command for easy access to JSON-RPC interface

lithp:
This is a much less ambitious version of #169 and replaces it.

Example output: 

```
$ cargo run -p trin-cli -- discv5_routingTableInfo
Attempting RPC. endpoint=discv5_routingTableInfo file=/tmp/trin-jsonrpc.ipc
{
  "id": 0,
  "jsonrpc": "2.0",
  "result": {
    "buckets": [
      [
        "0xa133..79e2",
        "enr:-IS4QBD940ObAwu3B4K1ddW15ohedlJW96z04vmJ3W9nKwsOOsQVsN_OZUTLH6zYJrq7GknkHQQ1r24zlSBr6aT6MngBgmlkgnY0gmlwhH8AAAGJc2VjcDI1NmsxoQNPYMlwKY62uH1WcO38VPTHHHD1CSHPJblkwxvbiwKXW4N1ZHCCEdg",
        "Connected"
      ]
    ],
    "localKey": "0x1c74..82db"
  }
}
```

(the first line is emitted to stderr, so the output of this command can be piped into jq)

### DIFF
--- a/trin-cli/src/main.rs
+++ b/trin-cli/src/main.rs
@@ -1,5 +1,5 @@
 use std::os::unix::net::UnixStream;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use structopt::StructOpt;
 use thiserror::Error;
 
@@ -24,7 +24,7 @@ struct Config {
     endpoint: String,
 }
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let Config { ipc, endpoint } = Config::from_args();
 
     eprintln!(
@@ -32,31 +32,18 @@ fn main() {
         endpoint,
         ipc.to_string_lossy()
     );
-    let mut client = match TrinClient::from_ipc(&ipc) {
-        Ok(client) => client,
-        Err(err) => {
-            eprintln!("Could not connect. err={:?}", err);
-            return;
-        }
-    };
+    let mut client = TrinClient::from_ipc(&ipc)?;
 
     let req = client.build_request(endpoint.as_str());
-    let resp = client.make_request(req);
+    let resp = client.make_request(req)?;
 
-    match resp {
-        Err(error) => {
-            eprintln!("error: {}", error);
-        }
-        Ok(value) => {
-            // unwrap: this value is safe to serialize, it was just deserialized!
-            println!("{}", serde_json::to_string_pretty(&value).unwrap());
-        }
-    }
+    println!("{}", serde_json::to_string_pretty(&resp).unwrap());
+    Ok(())
 }
 
-fn build_request<'a>(method: &'a str, request_id: u64) -> jsonrpc::Request<'a> {
+fn build_request(method: &str, request_id: u64) -> jsonrpc::Request {
     jsonrpc::Request {
-        method: method,
+        method,
         params: &[],
         id: serde_json::json!(request_id),
         jsonrpc: Some("2.0"),
@@ -84,7 +71,7 @@ where
 }
 
 impl TrinClient<UnixStream> {
-    fn from_ipc(path: &PathBuf) -> std::io::Result<Self> {
+    fn from_ipc(path: &Path) -> std::io::Result<Self> {
         // TODO: a nice error if this file does not exist
         Ok(Self {
             stream: UnixStream::connect(path)?,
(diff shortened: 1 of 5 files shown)


trin-cli/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    match resp {
+        Err(error) => {
+            eprintln!("error: {}", error);
+        }
+        Ok(value) => {
+            // unwrap: this value is safe to serialize, it was just deserialized!
+            println!("{}", serde_json::to_string_pretty(&value).unwrap());
+        }
+    }
```
carver, line 54: If you like it, it looks like this could shrink a bit by [returning a result from main](https://doc.rust-lang.org/rust-by-example/error/result.html#using-result-in-main) (with a bit of error tweaking):

```suggestion
    let mut client = TrinClient::from_ipc(&ipc)?;

    let req = client.build_request(endpoint.as_str());
    let resp = client.make_request(req);

    let value = resp?;
    // unwrap: this value is safe to serialize, it was just deserialized!
    println!("{}", serde_json::to_string_pretty(&value).unwrap());
```

Or, in this kind of one-shot CLI tool, it doesn't bother me to `.unwrap()` either (as is already used elsewhere in the module).

trin-cli/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+fn build_request<'a>(method: &'a str, request_id: u64) -> jsonrpc::Request<'a> {
+    jsonrpc::Request {
+        method: method,
```
carver, line 59: clippy said: 
```suggestion
        method,
```

trin-cli/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    #[error("Received empty response")]
+    Empty,
+}
+
+// TryClone is used because JSON-RPC responses are not followed by EOF. We must read bytes
+// from the stream until a complete object is detected, and the simplest way of doing that
+// with available APIs is to give ownership of a Read to a serde_json::Deserializer. If we
+// gave it exclusive ownership that would require us to open a new connection for every
+// command we wanted to send! By making a clone (or, by trying to) we can have our cake
```
carver, line 109: I guess that's only important with an interactive REPL-type tool, rather than this CLI one-shot, right?

carver (approved):
Yup, LGTM after tests pass!

trin-cli/src/main.rs
lithp: Yeah, definitely. I can drop it if you would prefer. This was originally added because I had grand plans of turning this into a client we could use in other parts of the codebase which want to make JSON-RPC calls, such as in the peer tester. You're right that it's unnecessary for current usage but I still think it's useful to have that long-term goal for `TrinClient`.

trin-cli/src/main.rs
lithp: Great point, just pushed a much shorter version

njgheorghita (approved):
Looks cool! It'd be nice if you could add a `README.md` with a couple basic usage examples. On a broader note, I typically use a python script / ipython console to hit the rpc endpoints. The ipython repl is pretty solid for just hitting endpoints ad nauseam. Just curious, how do you see this crate evolving? To play devils advocate just a bit, would distributing a python script be simpler?

trin-cli/src/main.rs
carver: Yeah, I'm happy to keep it. Maybe just add that little bit of context to the comment.

lithp:
> Looks cool! It'd be nice if you could add a `README.md` with a couple basic usage examples. On a broader note, I typically use a python script / ipython console to hit the rpc endpoints. The ipython repl is pretty solid for just hitting endpoints ad nauseam. Just curious, how do you see this crate evolving? To play devils advocate just a bit, would distributing a python script be simpler?

I really wish that distributing a python script was easier, but the second that script uses a library from outside the standard library  then it becomes very difficult for anyone who is not a python developer to use, by bundling simple tooling with trin we can make it easy for anyone who can install trin to install the associated tooling.

Unfortunately I will not be around to see it out, but my plan had a couple different prongs:
1. This command is already useful, I can hit ctrl-r from any shell and inspect my trin node's routing table. My next step was going to be adding some JSON-RPC endpoints which allow manually inserting data into the state network, or manually triggering specific `TALKREQ` commands, having all of these in a little cli seems like it provides some nice debugging UX!
2. Most of the code in this crate is not specific to `trin-cli`, it's useful anywhere we want to be a JSON-RPC client. This can be pulled out and moved into `trin-core` and used in other parts of the codebase.

PR #181, opened 2021-11-12
Update message types to match wire protocol test vectors

njgheorghita:
Running our message types against the test vectors [here](https://github.com/ethereum/portal-network-specs/pull/106/files) revealed some problems with our encoding/decoding. 

We we using a struct with optional fields to represent a union, which results in an invalid encoding. Imo, there were two solutions.. 1. Implement a Union type for the `ssz` library. 2. Create somewhat hacky types to emulate the desired behaviour. In this pr I went for # 2. Mostly because it seemed the quickest way forward since from my brief exploration, implementing a union type with encoding/decoding is not trivial (and working with/extending types from external libraries is not so trivial in rust in my experience).

Mostly, we had problems with `Ping`, `Pong`, `FindContent`, `FoundContent` (`FindNodes` and `Nodes` worked fine, since they didn't contain any unions). Those are fixed in this pr, as well as tests with all the vectors round-tripped.


jacobkaufmann:
I might be missing something here, but it looks like [the SSZ union type was implemented](https://github.com/sigp/lighthouse/pull/2579).

njgheorghita:
@jacobkaufmann Nice find! Looks to me like it's not available yet on a stable release? Three possible ways forward as I see it...

1. Move forward with this pr and then update the types once the next round of libraries are released
2. Pause these changes until the next round of libraries are released (compromises cross-client compatibility, but trin can still speak with itself)
3. Implement the changes now and build off the unstable branches as dependencies

Personally, I'm leaning towards 1, but curious to hear what others suggest

ogenev:
I would vote for **3)**. 

I think the `stable/unstable` tags are more related to Lighthouse releases than the ssz libraries. The ssz union type for Rust enums is also implemented in our current versions but it is using 4 bytes selectors instead of 1 byte, so what @jacobkaufmann referenced is an update on the union type to match the updated SSZ specs.

For more refs: https://github.com/ethereum/consensus-specs/issues/2270

jacobkaufmann:
@njgheorghita I agree with @ogenev here and favor (3). It looks like that change is on the `stable` branch of `lighthouse`. I'm not sure why the new version was not pushed to crates.io, but I would prefer to go ahead on build on this given that it is what we want in the long-term. If there are concerns, then I think we should reach out to the folks at sigma prime.

njgheorghita:
Quick update. I've had no luck building off the lighthouse stable branch (even if I could get it working, it feels like massive overkill, even as a temporary solution). I do agree that building off the proper union types from `eth_ssz` is the best way forward. Just heard from the lighthouse guys they had some issues with ci publish workflows, but they'll try and get those new versions published next week. I'm off next week, but will pick this up when I get back and build off the new crates.

njgheorghita:
@ogenev @jacobkaufmann @carver Ping, ready for another :eyes:, I've updated this pr to use the native union type from the latest `eth2_ssz` crate

trin-core/src/portalnet/types.rs
```diff
@@ -220,113 +219,61 @@ impl Request {
...
-                    data_radius,
-                    payload: Some(message),
-                }
-            }
-            None => Self {
-                data_radius,
-                payload: None,
-            },
-        }
+    pub fn new(payload: Vec<u8>) -> ByteList {
```
jacobkaufmann, line 240: nit: Not sure why we need the `CustomPayload` type if it is a unit struct. Perhaps instead we can implement `From<Vec<u8>>` for `ByteList`?

trin-core/src/portalnet/types.rs
```diff
@@ -579,46 +511,22 @@ mod test {
...
-    }
-
-    #[test]
-    fn test_found_content_encodes_payload() {
-        let empty_enrs: Vec<SszEnr> = vec![];
-        let msg = FoundContent {
-            enrs: empty_enrs,
-            payload: vec![1; 32],
-        };
+    fn test_found_content_encodes_content() {
```
jacobkaufmann, line 514: nit: Given the name change from `FoundContent` to `Content`, it would be more clear to make that consistent in the names of the tests.

trin-core/src/portalnet/types.rs
```diff
@@ -705,4 +609,115 @@ mod test {
...
+        }));
+        assert_eq!(hex::encode(message.to_bytes()), expected);
+        let decoded = Message::from_bytes(message.to_bytes().as_slice()).unwrap();
+        assert_eq!(decoded, message);
+    }
+
+    #[test]
+    fn test_vector_find_nodes() {
+        let distances = vec![256, 255];
+        let expected = "03040000000001ff00";
```
jacobkaufmann, line 648: Because we have not updated the [union selector to start at 0](https://github.com/ethereum/portal-network-specs/pull/108), the expected value does not match the linked test vector. This affects other test vectors as well. If the intent of this PR is to match the wire protocol test vectors, then we should ensure a match before we merge.

jacobkaufmann (commented):
Looks good! Along with a couple of minor nits, I did leave a comment with concern for test vector values. This is due to a mismatch in the union selector values for message types. We can either lump that in here, or merge a small PR to make that fix before we merge this to rectify the test vectors.

trin-core/src/portalnet/types.rs
njgheorghita: Nice, it was a hangover from the previous hacky union implementation, but turns out it's already implemented

trin-core/src/portalnet/types.rs
njgheorghita: Ahh good catch, made the update here

PR #184, opened 2021-11-16
Add IPC timeout to ethportal-peertest

carver:
There was some situation where trin was not responding to IPC calls. In that case, ethportal-peertest was hanging indefinitely instead of failing. So, add a timeout when reading from the socket.

Also, print some useful info to stderr.
Also, remove a test that hits Infura, before we start using it in CI.

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -1,6 +1,7 @@
 use std::io::prelude::*;
 use std::os::unix::net::UnixStream;
 use std::slice::Iter;
+use std::time::Duration;
 
 use hyper::{self, Body, Client, Method, Request};
 use log::info;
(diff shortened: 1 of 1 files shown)


ethportal-peertest/src/jsonrpc.rs
```diff
@@ -28,10 +29,6 @@ const ALL_ENDPOINTS: [JsonRpcEndpoint; 6] = [
         method: "discv5_routingTableInfo",
         id: &2,
     },
-    JsonRpcEndpoint {
-        method: "eth_blockNumber",
-        id: &3,
-    },
     JsonRpcEndpoint {
         method: "portalHistory_dataRadius",
         id: &4,
```
njgheorghita, line 34: Technically these ids are no longer in order, but it doesn't really make a difference.

njgheorghita (approved):
Looks good! Just curious why we're removing the infura test? Rate limits?

ethportal-peertest/src/jsonrpc.rs
carver: Yeah, these ids seem arbitrary, and maybe not around for a long time, so I won't tinker too much.

carver:
> Looks good! Just curious why we're removing the infura test? Rate limits?

Yeah, several reasons. I'm not keen to run a test that depends on a 3rd-party service being up and running. It's going to suck when they're unavailable. Also, removing the test makes local testing that much easier. Also, yes, rate limits. Even if Infura itself doesn't go down, we could hit some kind of cap on usage. Then we would have the same painful situation of CI going down for everyone until the key problem is resolved.

Issue #185, opened 2021-11-16
Testing Structure


njgheorghita:
# Proposed direction of trin's testing structure

## Unit tests
- If it can be written as a unit test, it should be.

## `tests` module
- Integration tests for trin-specific functionality 
  - Runs `ethportal-peertest` against trin (or a group of trin nodes)
  - Contains additional integration tests that don't fit nicely into `ethportal-peertest` (routing table management, etc.)

## `ethportal-peertest`
- A tool to test any client in a cross-client compatible manner
- Contains a set of json fixtures that other clients can pull in to test against

## Todos
- [ ] Common test library to share logic for trin-specific and client-agnostic integration tests
- [ ] Tests directory to hold integration tests
- [ ] Use `match` / something else to validate that every desired endpoint is consistently tested re: https://github.com/ethereum/trin/pull/188/files#r751917711
- [ ] Use ethportal-peertest as a library in tests directory
- [x] Invoke tests directory tests in Circle CI
- [ ] Cross-Client utp tests

njgheorghita:
Closing since test structure is already implemented

Issue #186, opened 2021-11-16
Fix build warning


carver:
This line appears during cargo build and cargo test:
```
warning: trin v0.1.0 (/home/jcarver/code/trin) ignoring invalid dependency `trin-cli` which is missing a lib target
```

- I think this just means that trin-cli needs a src/lib.rs
- Can we add a flag that would cause this warning to trigger a failure in CI?

ogenev:
fixed

Issue #151, opened 2021-10-21
Add validation to test harness that every endpoint is being tested


njgheorghita:
In one way or another, we want the test harness run to fail if it hasn't run against every currently implemented `TrinEndpoint`

re: https://github.com/ethereum/trin/pull/148#issuecomment-947908793

njgheorghita:
Closing since this is really hard to do ince all of the endpoints are stored as a large enum, which you cannot iterate through easily in rust. Further complications come from the fact that we might want one test for some endpoints and multiple for others. General guideline is that as you add a jsonrpc endpoint, you should also add tests to peertest

PR #188, opened 2021-11-17
Add overlay ping test to peertest

njgheorghita:
In this pr...
- Update `portalState_dataRadius` -> `portal_stateRadius`, `portalHistory_dataRadius` -> `portal_historyRadius`
- Add support for `portal_statePing` / `portal_historyPing`
- Adds `Params` field to jsonrpc overlay message types.
- Updates `ethportal-peertest` to behave more like an actual trin node w/ its own jsonrpc server for testing. This may or may not be relevant anymore due to recent discussion re: #185 - but it seems fine to leave as is until we reorg the tests
- Adds ipc file cleanup handler to `ethportal-peertest`
- Add `trin-cli` to `Dockerfile`


trin-history/src/jsonrpc.rs
```diff
@@ -19,6 +24,23 @@ impl HistoryRequestHandler {
                         self.network.read().await.overlay.data_radius.to_string(),
                     )));
                 }
+                HistoryEndpoint::Ping => {
```
njgheorghita, line 27: I'm curious if anybody has some insight into why the History network requires a `RwLock` here? I noticed, the state network doesn't. I also tried removing this `RwLock`, but after a fair amount of digging, wasn't able to resolve it. I'm happy to pick that thread back up in another pr, it seems like we've adopted the strategy of the fewer locks the better. But curious to hear any other thoughts first

carver: The state network had one until 048c1eca5f16597b5b0502c420ee10166ba72925 in #178 -- that commit might help. Though it's probably just big enough that it's worth being in its own PR.

carver:
I think the most important part of #186 is making sure that build warnings don't go unnoticed for a long time again. I'd like to see a CI test that fails while the build warning is still there.

trin-history/src/jsonrpc.rs
njgheorghita: :+1: tracking it in an issue for now

njgheorghita:
@carver Sounds good, I've removed `trin-cli/src/lib.rs` from this pr and the fixes comment in the description.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -113,7 +157,7 @@ pub async fn test_jsonrpc_endpoints_over_ipc(target_ipc_path: String) {
                 ),
             };
             match get_response_result(response_obj) {
-                Ok(result) => validate_endpoint_response(endpoint.method, &result),
+                Ok(result) => validate_endpoint_response(endpoint.method.as_str(), &result),
```
carver, line 160: I don't have a strong sense of when's the right time to do use `as_str()` vs `&`, but I've tended to prefer `&` when possible.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -144,22 +189,22 @@ fn get_response_result(response: Value) -> Result<Value, JsonRpcResponseError> {
...
         info!("Testing over HTTP: {:?}", endpoint.method);
-        let json_string = endpoint.to_jsonrpc();
+        let json_string = &endpoint.to_jsonrpc();
         let req = Request::builder()
             .method(Method::POST)
-            .uri(format!("http://{}", target_http_address))
+            .uri(format!("http://{}", peertest_config.target_http_address))
             .header("content-type", "application/json")
-            .body(Body::from(json_string))
+            .body(Body::from(json_string.clone()))
```
carver, line 201: Why the borrow in line 196, followed by a clone here? I don't see why either line needs to change at all.

ethportal-peertest/src/main.rs
```diff
@@ -78,27 +134,14 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        match ping_result {
-            Ok(val) => info!("Successful ping to State network: {:?}", val),
-            Err(msg) => panic!("Invalid ping to State network: {:?}", msg),
-        }
-
+        tokio::spawn(async { history_network_task.unwrap().await });
+        tokio::spawn(async { state_network_task.unwrap().await });
+        tokio::spawn(async { jsonrpc_server_task.await.unwrap() });
+        // short sleep so server task gets running
+        thread::sleep(time::Duration::from_millis(1000));
```
carver, line 141: Sleeping to wait for something to happen will eventually cause heartache: it adds unnecessary slowdown time on launch at times, or even worse causes some funny race condition when the server takes longer than 1s to launch.

Ideally we want something that we can await, which returns when the server is launched. Maybe that means a one-shot channel that sends a signal when the server is up?

My approach would be to just try it out with a short timebox. See if you can bang it out in an hour, without too much code. If not, then push it into an issue (now explaining what you learned from a quick attempt).

trin-state/src/jsonrpc.rs
```diff
@@ -18,6 +23,16 @@ impl StateRequestHandler {
                     let radius = &self.network.overlay.data_radius;
                     let _ = request.resp.send(Ok(Value::String(radius.to_string())));
                 }
+                StateEndpoint::Ping => {
+                    let response = match PingParams::try_from(request.params) {
+                        Ok(val) => match self.network.overlay.send_ping(val.enr, None).await {
+                            Ok(pong) => Ok(Value::String(format!("{:?}", pong.payload.unwrap()))),
```
carver, line 29: Looks like an invalid pong payload would crash trin here, so we need a little more error handling.

carver (approved):
Some things to take care of, but looks like it should be fine to ship after those. Ping me if you want another look after your changes!

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -56,49 +36,113 @@ fn validate_endpoint_response(method: &str, result: &Value) {
...
+        }
         _ => panic!("Unsupported endpoint"),
     };
     info!("{:?} returned a valid response.", method);
 }
 
 impl JsonRpcEndpoint {
-    pub fn all_endpoints() -> Iter<'static, Self> {
-        ALL_ENDPOINTS.iter()
+    pub fn all_endpoints(target_node: String) -> Vec<JsonRpcEndpoint> {
```
jacobkaufmann, line 61: Don't think we need to address #151 here, but it would be nice to enforce some exhaustive list, perhaps with some clever use of a `match` statement on `TrinEndpoint`. A nice/simple way to do that is not clear to me, so I haven't suggested it here, but it's probably something we should address before we add too many more endpoints to the test harness.

trin-core/src/jsonrpc/types.rs
```diff
@@ -70,6 +77,27 @@ fn validate_jsonrpc_version(jsonrpc: &str) -> Result<(), ValidationError> {
...
+    pub enr: Enr,
+}
+
+impl TryFrom<Params> for PingParams {
+    type Error = ValidationError;
+
+    fn try_from(params: Params) -> Result<Self, Self::Error> {
+        match params {
+            Params::Array(val) => match val.len() {
+                1 => match Enr::from_str(val[0].as_str().unwrap()) {
```
jacobkaufmann, line 90: Similar to @carver's comment below, we would want to add another `match` on the call to [`as_str`](https://docs.serde.rs/serde_json/value/enum.Value.html#method.as_str) to check for a `None` value, which would indicate that we received some non-string value.

jacobkaufmann (approved):
Looks good to go after addressing the `unwrap` that I noted along with the comments from prior review.

ethportal-peertest/src/jsonrpc.rs
njgheorghita: Yup, definitely agree. Though, it needs to be done in a way that we can also exclude certain tests (eg. jsonrpc endpoints that call to infura which we don't always want to run). Tracking this in #185

trin-core/src/jsonrpc/service.rs
```diff
@@ -30,6 +30,7 @@ pub fn launch_jsonrpc_server(
     trin_config: TrinConfig,
     infura_project_id: String,
     portal_tx: UnboundedSender<PortalJsonRpcRequest>,
+    live_server_tx: tokio::sync::mpsc::Sender<bool>,
```
njgheorghita, line 33: I also considered making the `live_server_tx` an `Option`. In the main boot process, I've never run into this issue, and can't imagine a scenario where you need the jsonrpc server immediately (as we do in the tests) and don't end up waiting around however many ms it takes for the server to boot.

njgheorghita:
@carver If you wouldn't mind giving this another :eyes: that'd be great. I spent probably too much time trying to use a oneshot channel. But since that only works as a channel between two async processes, I had to convert the whole `launch_jsonrpc_server()` chain to async. This was tricky and I couldn't exactly figure out a good way to handle it, so in the end I went with an `mpsc` channel which can do communication between an async and sync process

carver:
> @carver Sounds good, I've removed `trin-cli/src/lib.rs` from this pr and the fixes comment in the description.

Looks like the whole commit cc8672b42c994cc80a53e784fb952f52938ff79d should be dropped?

trin-core/src/jsonrpc/types.rs
```diff
@@ -70,6 +77,38 @@ fn validate_jsonrpc_version(jsonrpc: &str) -> Result<(), ValidationError> {
...
+    pub enr: Enr,
+}
+
+impl TryFrom<Params> for PingParams {
+    type Error = ValidationError;
+
+    fn try_from(params: Params) -> Result<Self, Self::Error> {
+        match params {
+            Params::Array(val) => match val.len() {
+                1 => PingParams::try_from(&val[0]),
```
carver, line 90: Yeah, this seems like a solid solution to not wanting to nest too many `match`es.

carver (approved):
Yup, I can see why the mpsc feels just a little messy, but I don't have a better suggestion. It's definitely better than a sleep!

PR #191, opened 2021-11-19
Remove unnecessary rwlock from history network

### LINKED ISSUE #190: Remove redundant locks from history network
re: https://github.com/ethereum/trin/pull/188/files#r751380124

History network still uses `RwLock`s which should be removed now that they're no longer strictly required.

njgheorghita:
With the latest `discv5` library, we no longer need to use locks on overlay networks. fixes #190


carver (approved):
LGTM

PR #192, opened 2021-12-01
Add trin cli binary to dockerfile

njgheorghita:
A couple minor bugfixes for using `trin-cli` inside a docker container
- Make `--params` arg optional 
- Add copy `trin-cli` binary to dockerfile

### DIFF
--- a/trin-cli/src/main.rs
+++ b/trin-cli/src/main.rs
@@ -26,7 +26,7 @@ struct Config {
     endpoint: String,
 
     #[structopt(long, help = "first parameter")]
-    params: String,
+    params: Option<String>,
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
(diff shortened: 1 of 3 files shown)


trin-cli/src/main.rs
```diff
@@ -54,14 +54,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 fn build_request<'a>(
     method: &'a str,
-    raw_params: &'a [Box<RawValue>],
+    raw_params: &'a Option<[Box<RawValue>; 1]>,
```
jacobkaufmann, line 57: Seems like a strange/ugly type, but I'm assuming there's a non-trivial reason we use it here?

jacobkaufmann (approved):
LGTM. Minor curiosity about a type choice, but given its limited scope, I don't think it's much of a concern. Assuming that `trin-cli` command was tested for Docker image.

trin-cli/src/main.rs
njgheorghita: Yeah it's not pretty, we're required to use it because of the [jsonrpc library](https://docs.rs/jsonrpc/0.12.0/jsonrpc/struct.Request.html#structfield.params) we're using. I tried a couple tweaks to clean it up, but couldn't find a suitable solution

Issue #198, opened 2021-12-10
Correctly respond to json-rpc requests: portal_*Ping


carver:
Right now, we're returning the ssz-encoded message response from the peer, when asking over json-rpc. That's wrong.

Build out the json structure returned here: trin-state/src/jsonrpc.rs (`pong.into()` is where the pong object is converted to ssz bytes right now).

PR #199, opened 2021-12-10
Re-enable ping peertests

### LINKED ISSUE #196: Fix portal_statePing and portal_historyPing tests in ethportal-peertest
In order to expedite #187 I disabled those tests.

TODO:

- [x] merge #187
- [x] re-enable the test
- [x] launch two instances of trin, to be the target of the ping?
- [x] pass in the ENR of the 2nd trin for the trin target

carver:
Fix #196 

Includes a refactor to reduce duplicate code when making IPC requests from peertest. Also, was crashing when ctrl-c was registering twice (on the 2nd trin run).

Note that we're not actually returning the correct values for portal_*Ping json requests, but that belongs in a separate issue (#198). This PR is just about building the structure necessary to run two copies of trin and trigger messages between them.

### DIFF
--- a/trin-core/src/jsonrpc/service.rs
+++ b/trin-core/src/jsonrpc/service.rs
@@ -84,13 +84,17 @@ pub fn launch_jsonrpc_server(
 fn set_ipc_cleanup_handlers(ipc_path: &str) {
     {
         let ipc_path = ipc_path.to_string();
-        ctrlc::set_handler(move || {
+        if let Err(err) = ctrlc::set_handler(move || {
             if let Err(err) = fs::remove_file(&ipc_path) {
                 debug!("Ctrl-C: Skipped removing {} because: {}", ipc_path, err);
             };
             std::process::exit(1);
-        })
-        .expect("Error setting Ctrl-C handler.");
+        }) {
+            warn!(
+                "Could not set the Ctrl-C handler for removing the IPC socket: {}",
+                err
+            );
+        }
     }
 
     {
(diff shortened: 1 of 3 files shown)


trin-core/src/jsonrpc/service.rs
```diff
@@ -84,13 +84,17 @@ pub fn launch_jsonrpc_server(
...
-        ctrlc::set_handler(move || {
+        if let Err(err) = ctrlc::set_handler(move || {
             if let Err(err) = fs::remove_file(&ipc_path) {
                 debug!("Ctrl-C: Skipped removing {} because: {}", ipc_path, err);
             };
             std::process::exit(1);
-        })
-        .expect("Error setting Ctrl-C handler.");
+        }) {
+            warn!(
```
njgheorghita, line 93: Have you experienced dangling ipc sockets when running the peertest? They're not the end of the world, it's a pretty obvious debug when trying to restart trin with an unavailable socket. But, it does seem like the warning could pass by pretty quickly on the output, especially if you're connecting with bootnodes immediately.

njgheorghita (approved):
Looks good!

trin-core/src/jsonrpc/service.rs
carver: I haven't yet. For better or worse, we seem to have lots of failsafes (when catching the panic, or at the end of the serving loop).

carver: Maybe the other thing to note is that this should be quite unusual to run multiple instances of trin in the same parent process.

PR #200, opened 2021-12-11
move u256/u512 defs into types/uint.rs

victor-wei126:
Moved the uint type defs into `portalnet/types/uint.rs` in order to keep the main mod file clean


njgheorghita (approved):
Looks great! Just need to run `rust fmt` one more time to clear up that ci error

PR #202, opened 2021-12-13
Overlay content

### LINKED ISSUE #201: Trin should not panic on random `content_key` value for `FindContent`
When receiving `FindContent` messages with random `content_key` data, Trin panics because of:

https://github.com/ethereum/trin/blob/1bc231a4b0ef3884532371fd670aa3a21b4a64f9/trin-core/src/utils/distance.rs#L3

We need to handle this error instead of panic.

njgheorghita:
- updates `trin-cli` to accept multiple parameters for a jsonrpc request
- adds support for `portal_stateFindContent(enr, content_key)` & `portal_historyFindContent(enr, content_key)` jsonrpc endpoints
- fixes #201

### DIFF
--- a/trin-cli/src/main.rs
+++ b/trin-cli/src/main.rs
@@ -25,8 +25,8 @@ struct Config {
     #[structopt(help = "e.g. discv5_routingTableInfo", required = true)]
     endpoint: String,
 
-    #[structopt(long, help = "first parameter")]
-    params: Option<String>,
+    #[structopt(long, help = "parameters", use_delimiter = true)]
+    params: Option<Vec<String>>,
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -36,6 +36,10 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         params,
     } = Config::from_args();
 
+    let params: Option<Vec<Box<RawValue>>> = match params {
+        Some(val) => Some(val.into_iter().map(|val| jsonrpc::arg(val)).collect()),
+        None => None,
+    };
     eprintln!(
         "Attempting RPC. endpoint={} params={:?} file={}",
         endpoint,
@@ -44,7 +48,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     );
     let mut client = TrinClient::from_ipc(&ipc)?;
 
-    let params = params.map(|val| [jsonrpc::arg(val)]);
     let req = client.build_request(endpoint.as_str(), &params);
     let resp = client.make_request(req)?;
 
@@ -54,16 +57,18 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 fn build_request<'a>(
     method: &'a str,
-    raw_params: &'a Option<[Box<RawValue>; 1]>,
+    raw_params: &'a Option<Vec<Box<RawValue>>>,
     request_id: u64,
 ) -> jsonrpc::Request<'a> {
     match raw_params {
-        Some(val) => jsonrpc::Request {
-            method,
-            params: val,
-            id: serde_json::json!(request_id),
-            jsonrpc: Some("2.0"),
-        },
+        Some(val) => {
+            jsonrpc::Request {
+                method,
+                params: val,
+                id: serde_json::json!(request_id),
+                jsonrpc: Some("2.0"),
+            }
+        }
         None => jsonrpc::Request {
             method,
             params: &[],
@@ -131,7 +136,7 @@ where
     fn build_request(
         &mut self,
         method: &'a str,
-        params: &'a Option<[Box<RawValue>; 1]>,
+        params: &'a Option<Vec<Box<RawValue>>>,
     ) -> jsonrpc::Request<'a> {
         let result = build_request(method, params, self.request_id);
         self.request_id += 1;
--- a/trin-cli/src/main.rs
+++ b/trin-cli/src/main.rs
@@ -36,10 +36,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         params,
     } = Config::from_args();
 
-    let params: Option<Vec<Box<RawValue>>> = match params {
-        Some(val) => Some(val.into_iter().map(|val| jsonrpc::arg(val)).collect()),
-        None => None,
-    };
+    let params: Option<Vec<Box<RawValue>>> =
+        params.map(|param| param.into_iter().map(jsonrpc::arg).collect());
+
     eprintln!(
         "Attempting RPC. endpoint={} params={:?} file={}",
         endpoint,
@@ -61,14 +60,12 @@ fn build_request<'a>(
     request_id: u64,
 ) -> jsonrpc::Request<'a> {
     match raw_params {
-        Some(val) => {
-            jsonrpc::Request {
-                method,
-                params: val,
-                id: serde_json::json!(request_id),
-                jsonrpc: Some("2.0"),
-            }
-        }
+        Some(val) => jsonrpc::Request {
+            method,
+            params: val,
+            id: serde_json::json!(request_id),
+            jsonrpc: Some("2.0"),
+        },
         None => jsonrpc::Request {
             method,
             params: &[],
--- a/trin-core/src/jsonrpc/types.rs
+++ b/trin-core/src/jsonrpc/types.rs
@@ -1,12 +1,15 @@
 use std::str::FromStr;
 
+use hex::decode;
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
+use ssz_types::VariableList;
 use tokio::sync::mpsc;
 use validator::{Validate, ValidationError};
 
 use crate::jsonrpc::endpoints::{HistoryEndpoint, StateEndpoint, TrinEndpoint};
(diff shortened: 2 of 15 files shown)


trin-core/src/jsonrpc/types.rs
```diff
@@ -101,14 +104,56 @@ impl TryFrom<&Value> for PingParams {
     fn try_from(param: &Value) -> Result<Self, Self::Error> {
         match param.as_str() {
             Some(val) => match Enr::from_str(val) {
-                Ok(val) => Ok(PingParams { enr: val }),
+                Ok(enr) => Ok(Self { enr }),
```
carver, line 107: nice cleanup

trin-core/src/jsonrpc/types.rs
```diff
@@ -101,14 +104,56 @@ impl TryFrom<&Value> for PingParams {
...
+    pub content_key: ByteList,
+}
+
+impl TryFrom<Params> for FindContentParams {
+    type Error = ValidationError;
+
+    fn try_from(params: Params) -> Result<Self, Self::Error> {
+        match params {
+            Params::Array(val) => match val.len() {
+                2 => FindContentParams::try_from([&val[0], &val[1]]),
```
carver, line 126: Feels like there ought to be a cleaner way to do this. I believe [this post](https://stackoverflow.com/a/29570662/8412986) is suggesting that you could do: 
```suggestion
                2 => FindContentParams::try_from(val.try_into().unwrap()),
```
Hm, maybe that's not any better...

trin-core/src/portalnet/overlay.rs
```diff
@@ -125,7 +125,11 @@ impl OverlayProtocol {
     ) -> Result<Response, OverlayRequestError> {
         let request = match Message::from_bytes(talk_request.body()) {
             Ok(Message::Request(request)) => request,
-            Ok(_) => return Err(OverlayRequestError::InvalidRequest),
+            Ok(_) => {
+                return Err(OverlayRequestError::InvalidRequest(
+                    "Message not a recognized request type".to_string(),
```
carver, line 130: Maybe we'd want to include the request bytes in this message?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -357,17 +366,24 @@ impl OverlayService {
...
-    async fn find_nodes_close_to_content(&self, content_key: Vec<u8>) -> Vec<SszEnr> {
+    async fn find_nodes_close_to_content(
+        &self,
+        content_key: Vec<u8>,
+    ) -> Result<Vec<SszEnr>, OverlayRequestError> {
         let self_node_id = self.local_enr().await.node_id();
-        let self_distance = xor_two_values(&content_key, &self_node_id.raw().to_vec());
+        let self_distance = match xor_two_values(&content_key, &self_node_id.raw().to_vec()) {
+            Ok(val) => val,
+            Err(msg) => return Err(OverlayRequestError::InvalidRequest(msg.to_string())),
```
carver, line 376: Maybe a bit more context here? Like `format!("Could not find distance from node, because content key is malformed: {}", msg.to_string())`

carver (approved):

trin-cli/src/main.rs
```diff
@@ -54,7 +56,7 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
 fn build_request<'a>(
     method: &'a str,
-    raw_params: &'a Option<[Box<RawValue>; 1]>,
+    raw_params: &'a Option<Vec<Box<RawValue>>>,
```
jacobkaufmann, line 59: I'm not sure what functionality we have for this CLI, or how extensively it was tested initially, but did we test both `Ping` and `FindContent` after this change?

trin-core/src/jsonrpc/types.rs
```diff
@@ -101,14 +104,56 @@ impl TryFrom<&Value> for PingParams {
...
+            .ok_or_else(|| ValidationError::new("Empty enr param"))?;
+        let content_key = params[1]
+            .as_str()
+            .ok_or_else(|| ValidationError::new("Empty content key param"))?;
+
+        let enr = match Enr::from_str(enr) {
+            Ok(enr) => enr,
+            Err(_msg) => return Err(ValidationError::new("Invalid enr param")),
+        };
+        let content_key = match decode(content_key) {
```
jacobkaufmann, line 149: minor nit: I think it's idiomatic in Rust to preserve the namespace for methods, so that we would have `hex::decode` here. This makes it clear that we are decoding hex, and not some other encoding.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -290,9 +294,14 @@ impl OverlayService {
...
-                Ok(Content::Enrs(enrs))
+                match enrs {
+                    Ok(val) => Ok(Content::Enrs(val)),
+                    Err(msg) => Err(OverlayRequestError::InvalidRequest(msg.to_string())),
+                }
             }
-            Err(error) => panic!("Unable to respond to FindContent: {}", error),
+            Err(_) => Err(OverlayRequestError::Failure(
+                "Unable to respond to FindContent".to_string(),
+            )),
```
jacobkaufmann, line 304: I agree that we should not panic here, but I think it would be good to log the error from the `db`.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -244,7 +244,7 @@ impl fmt::Display for Ping {
             f,
             "Ping(enr_seq={}, radius={})",
             self.enr_seq,
-            hex::encode(self.custom_payload.as_ssz_bytes())
+            encode(self.custom_payload.as_ssz_bytes())
```
jacobkaufmann, line 247: minor nit: Similar to above on preserving `hex` namespace in the call.

jacobkaufmann (approved):
Looks good. Some minor comments, but nothing that would block merge.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -417,6 +417,22 @@ impl Content {
...
+impl TryInto<Value> for Content {
+    type Error = MessageDecodeError;
+
+    fn try_into(self) -> Result<Value, Self::Error> {
+        if let Content::ConnectionId(val) = self {
+            Ok(Value::String(val.to_string()))
+        } else if let Content::Content(val) = self {
+            Ok(Value::String(encode(val.to_vec())))
+        } else if let Content::Enrs(val) = self {
+            Ok(Value::String(val[0].to_string()))
```
ogenev, line 429: Is there any specific reason we are returning only the first value of the `Vec<SzzEnr>`?

When the response is empty `Content::Enrs`, this is failing with:

```bash
thread 'tokio-runtime-worker' panicked at 'index out of bounds: the len is 0 but the index is 0', trin-core/src/portalnet/types/messages.rs:429:30
```
We need to handle this in a way to allow empty `Content::Enrs` value.

ogenev (changes requested):
I tested this manually with Fluffy and Trin panics when receiving a CONTENT message with an empty `enrs` array.

Other than this, looks good to me :+1:

trin-cli/src/main.rs
njgheorghita: Yeah, tests would require serving some kind of static content from a database. We're sort of in a limbo state between two databases models right now - so it felt out of scope to address that in this pr. This was just tested locally, I snuck some code in that manually inserted random values into the db and then sent it over the wire to another client (and I also tested `Ping` to make sure that didn't break). Not ideal, but enough to get some confidence that the basic functionality is there. Also, imo tests for `FindContent`/`Content` are so glaringly obvious that I'm not worried about them falling through the cracks as we get the db sorted / peertest infrastructure up.

trin-core/src/portalnet/types/messages.rs
njgheorghita: Ahh good catch! This is just leftover sloppy code from the initial implementation phase. It's no longer failing on empty content. Also updated the jsonrpc result to show what kind of data was being returned 
```json
{
  ...
  "result": {
    "enrs": "[]"
  }
}```
instead of 
```json
{
  ...
  "result": "[]"
}```

File: trin-state/src/jsonrpc.rs [hunk 10]
```diff
@@ -23,6 +23,24 @@ impl StateRequestHandler {
                     let radius = &self.network.overlay.data_radius;
                     let _ = request.resp.send(Ok(Value::String(radius.to_string())));
                 }
+                StateEndpoint::FindContent => {
```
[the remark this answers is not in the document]
njgheorghita: Well there definitely is plenty of duplicated logic here. In fact, all of the logic for the 3 currently supported endpoints seem to be identical. But, I'm not convinced the solution is to extract it into common methods. If we're doing that, then why are we splitting the logic out into separate modules in the first place? My intuition is that they are identical currently, but as the needs and nuances of the respective networks develop - then they'll require differentiation. So it seems premature to extract the data. But, this could also just be me justifying my laziness. What do you think?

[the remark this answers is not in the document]
ogenev: Maybe you are right, I'm happy to leave this as it is. We can address it when we start to add other networks in the stack :)

ogenev (approved):

PR #206, opened 2021-12-16
Peertest CLI Redesign

carver:
By default, connect to nodes over the json-rpc IPC, instead of via enode.


carver:
Hey @ogenev, sorry I am guessing this conflicts a bit with the POC you are working on. If this aligns enough, then I guess let's just merge (after normal review). If you're thinking a really different direction, then I'm happy to pause this PR until you can show me your POC.

ethportal-peertest/src/lib.rs
```diff
@@ -2,6 +2,32 @@ pub mod cli;
...
-
 pub use cli::PeertestConfig;
+pub use jsonrpc::{get_enode, test_jsonrpc_endpoints_over_http, test_jsonrpc_endpoints_over_ipc};
+
+use std::sync::Arc;
+
+use trin_core::cli::TrinConfig;
+use trin_core::jsonrpc::service::JsonRpcExiter;
+
+pub async fn launch_buddy_node() -> (String, Arc<JsonRpcExiter>) {
```
ogenev, line 13: I think we can make this method more general and add `discovery_port` and `web3_ipc_path` as method arguments. This way, we can use it to spin multiple nodes.

tests/self_peertest.rs
```diff
@@ -8,43 +8,24 @@ mod test {
...
-        };
+        let (buddy_enode, buddy_client_exiter) = peertest::launch_buddy_node().await;
 
-        let peertest_config =
-            peertest::PeertestConfig::new_from([".", "--target-node", &buddy_enode].iter())
-                .unwrap();
+        let peertest_config = peertest::PeertestConfig::new_from(["."].iter()).unwrap();
 
         // Run a client, to be tested
         let trin_config = TrinConfig::new_from(
```
ogenev, line 16: If we refactor `launch_buddy_node` as suggested above, we can use it here too

ethportal-peertest/src/main.rs
```diff
@@ -1,161 +1,23 @@
...
-use std::env;
-use std::sync::Arc;
-
-use log::{debug, info};
-use structopt::StructOpt;
-use tokio::sync::mpsc;
+use log::info;
 
 use ethportal_peertest::cli::PeertestConfig;
-use ethportal_peertest::events::PortalnetEvents;
```
ogenev, line 10: I guess we can get rid now of the `events.rs` file :scissors:.

ogenev (approved):
Yes, I'm thinking in the same direction :100:. LGTM :+1:

tests/self_peertest.rs
carver: Hm, I'm not really sold on this generalization of `launch_buddy_node`. I think `run_trin()` is the general one-liner function, already. For one example: I want the buddy peer to have a consistent node id during testing, but don't want that same node id for the node under test. So I guess we'd have to make that an argument too. At some point, we're just recreating the whole run_trin() interface, passing in the config in a slightly different way.

PR #208, opened 2021-12-16
Overlay nodes

njgheorghita:
- Add support for `portal_stateFindNodes(enr, distances)`/`portal_historyFindNodes(enr, distances)` jsonrpc endpoint
- Implement some helper traits on `SszEnr` for cleaner param conversion
- Add validation methods to `send_find_nodes()` according to the [spec](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-protocol.md#find-nodes-0x02)

### DIFF
--- a/trin-core/src/jsonrpc/types.rs
+++ b/trin-core/src/jsonrpc/types.rs
@@ -1,5 +1,3 @@
-use std::str::FromStr;
-
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
 use ssz_types::VariableList;
@@ -7,8 +5,7 @@ use tokio::sync::mpsc;
 use validator::{Validate, ValidationError};
 
 use crate::jsonrpc::endpoints::{HistoryEndpoint, StateEndpoint, TrinEndpoint};
-use crate::portalnet::types::messages::ByteList;
-use crate::portalnet::Enr;
+use crate::portalnet::types::messages::{ByteList, SszEnr};
 
 type Responder<T, E> = mpsc::UnboundedSender<Result<T, E>>;
 
@@ -80,7 +77,7 @@ fn validate_jsonrpc_version(jsonrpc: &str) -> Result<(), ValidationError> {
 }
 
 pub struct PingParams {
-    pub enr: Enr,
+    pub enr: SszEnr,
 }
 
 impl TryFrom<Params> for PingParams {
@@ -101,18 +98,16 @@ impl TryFrom<&Value> for PingParams {
     type Error = ValidationError;
 
     fn try_from(param: &Value) -> Result<Self, Self::Error> {
-        match param.as_str() {
-            Some(val) => match Enr::from_str(val) {
-                Ok(enr) => Ok(Self { enr }),
-                Err(_) => Err(ValidationError::new("Invalid enr param")),
-            },
-            None => Err(ValidationError::new("Missing or empty enr param")),
-        }
+        let enr: SszEnr = match param.try_into() {
+            Ok(val) => val,
+            Err(msg) => return Err(msg),
+        };
+        Ok(Self { enr })
     }
 }
 
 pub struct FindNodesParams {
-    pub enr: Enr,
+    pub enr: SszEnr,
     pub distances: Vec<u16>,
 }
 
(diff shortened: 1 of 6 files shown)


trin-core/src/jsonrpc/types.rs
```diff
@@ -101,18 +98,55 @@ impl TryFrom<&Value> for PingParams {
...
 
     fn try_from(param: &Value) -> Result<Self, Self::Error> {
-        match param.as_str() {
-            Some(val) => match Enr::from_str(val) {
-                Ok(enr) => Ok(Self { enr }),
-                Err(_) => Err(ValidationError::new("Invalid enr param")),
+        let enr: SszEnr = match param.try_into() {
+            Ok(val) => val,
+            Err(msg) => return Err(msg),
+        };
```
carver, line 104: Looks equivalent to:
```suggestion
        let enr: SszEnr = param.try_into()?;
```

trin-core/src/jsonrpc/types.rs
```diff
@@ -101,18 +98,55 @@ impl TryFrom<&Value> for PingParams {
...
 }
 
+impl TryFrom<[&Value; 2]> for FindNodesParams {
+    type Error = ValidationError;
+
+    fn try_from(params: [&Value; 2]) -> Result<Self, Self::Error> {
+        let enr: SszEnr = match params[0].try_into() {
+            Ok(val) => val,
+            Err(msg) => return Err(msg),
+        };
```
carver, line 135: Looks like another ? opportunity.

trin-core/src/jsonrpc/types.rs
```diff
@@ -134,20 +168,17 @@ impl TryFrom<[&Value; 2]> for FindContentParams {
     type Error = ValidationError;
 
     fn try_from(params: [&Value; 2]) -> Result<Self, Self::Error> {
-        let enr = params[0]
-            .as_str()
-            .ok_or_else(|| ValidationError::new("Empty enr param"))?;
+        let enr: SszEnr = match params[0].try_into() {
```
carver, line 171: here too

trin-core/src/portalnet/overlay.rs
```diff
@@ -194,6 +195,10 @@ impl OverlayProtocol {
         distances: Vec<u16>,
     ) -> Result<Nodes, OverlayRequestError> {
         // Construct the request.
+        match validate_find_nodes_distances(&distances) {
+            Ok(_) => (),
+            Err(msg) => return Err(msg),
+        }
```
carver, line 201:
```suggestion
        validate_find_nodes_distances(&distances)?;
```
I guess it works here, too?

trin-core/src/portalnet/overlay.rs
```diff
@@ -256,3 +261,61 @@ impl OverlayProtocol {
...
+        let result = validate_find_nodes_distances(&input);
+        match result {
+            Ok(val) => assert_eq!(val, ()),
+            Err(_) => panic!("Valid test case fails"),
+        }
+    }
+
+    #[rstest]
+    #[case(vec![], "Empty list")]
+    #[case(vec![0u16; 257], "More than 256")]
```
carver, line 311: Technically this would also fail the duplicate test (as would anything array over length 256, I suppose). I'm fine leaving it as-is. Maybe a comment that it's reasonable to assume we test length first, since that's a faster check than duplicates.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -442,6 +449,26 @@ impl SszEnr {
     }
 }
 
+impl Into<Enr> for SszEnr {
+    fn into(self) -> Enr {
+        Enr::from(self.0.clone())
```
carver, line 454: Since this isn't a borrow of self, I don't think we need the clone here.

carver (approved):
Yup, seems good to me!

ogenev (approved):
LGTM :+1:

trin-core/src/portalnet/overlay.rs
```diff
@@ -256,3 +261,61 @@ impl OverlayProtocol {
...
+    if invalid_distances.len() > 0 {
+        return Err(OverlayRequestError::InvalidRequest(format!(
+            "Invalid distances: Distances greater than 256 are not allowed. Found: {:?}",
+            invalid_distances
+        )));
+    }
+    let unique: HashSet<u16> = HashSet::from_iter(distances.iter().cloned());
+    if unique.len() != distances.len() {
+        return Err(OverlayRequestError::InvalidRequest(
+            "Invalid distances: Duplicate elements detected".to_string(),
```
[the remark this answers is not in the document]
njgheorghita: The spec is pretty explicit about [duplicated distances being invalid](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-protocol.md#find-nodes-0x02). We could do something like simply removing all duplicated elements, but I'm not sure I'm comfortable with implementing that kind of "magic" in the background.

Issue #203, opened 2021-12-14
Add `custom_payload` support for overlay `Ping` json-rpc params.


ogenev:
Current overlay jsonrpc `Ping` implementation does not support passing a `custom_payload` element to the request.
https://github.com/ethereum/trin/blob/62314a48935b3fcfbf9eb1772008e787ddcc006b/trin-core/src/jsonrpc/types.rs#L98

We need to add such support according to specs [here](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-protocol.md#ping-0x00).

PR #209, opened 2021-12-20
Implement `custom_payload` element of overlay Ping/Pong JSON-RPC params

ogenev:
What does this PR do:

- Add new `CustomPayload` struct
- Add support for custom payload param of `historyPing` and `statePing` JSON-RPC
- Implement `CustomPayload` struct in overlay Ping/Pong messages. This required some custom SSZ decode/encode implementation, so we can match the test vectors.

Fixes https://github.com/ethereum/trin/issues/203

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -87,12 +87,18 @@ impl JsonRpcEndpoint {
             JsonRpcEndpoint {
                 method: "portal_statePing".to_string(),
                 id: 6,
-                params: Params::Array(vec![Value::String(buddy_node.clone())]),
+                params: Params::Array(vec![
+                    Value::String(buddy_node.clone()),
+                    Value::String("010100".to_string()),
+                ]),
             },
             JsonRpcEndpoint {
                 method: "portal_historyPing".to_string(),
                 id: 7,
-                params: Params::Array(vec![Value::String(buddy_node)]),
+                params: Params::Array(vec![
+                    Value::String(buddy_node),
+                    Value::String("010100".to_string()),
+                ]),
             },
         ]
     }
--- a/trin-core/src/portalnet/types/messages.rs
+++ b/trin-core/src/portalnet/types/messages.rs
@@ -19,7 +19,11 @@ use crate::portalnet::{types::uint::U256, Enr};
 
 pub type ByteList = VariableList<u8, typenum::U2048>;
 
-pub struct CustomPayload(ByteList);
+/// Custom payload element of Ping and Pong overlay messages
+#[derive(Debug, PartialEq, Clone)]
+pub struct CustomPayload {
+    payload: ByteList,
+}
 
 impl TryFrom<&Value> for CustomPayload {
     type Error = ValidationError;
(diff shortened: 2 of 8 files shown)


trin-core/src/portalnet/types/messages.rs
```diff
@@ -19,6 +19,66 @@ use crate::portalnet::{types::uint::U256, Enr};
...
+    payload: ByteList,
+}
+
+impl TryFrom<&Value> for CustomPayload {
+    type Error = ValidationError;
+
+    fn try_from(value: &Value) -> Result<Self, Self::Error> {
+        let value = value
+            .as_str()
+            .ok_or_else(|| ValidationError::new("Empty custom payload value"))?;
```
njgheorghita, line 34: I might be confused here, but shouldn't this error message be more along the lines of "this value is not a string"? `Value` is an enum, and I guess we'd want to display if deserialization returned variants other than `String` (which would also catch any `Null` cases)

njgheorghita (approved):
Looks good!

trin-core/src/portalnet/types/messages.rs
ogenev: You are right, we can use a better message. This was copy-paste from https://github.com/ethereum/trin/blob/73f38b26c1177aa7326ebc91b6503a2257e64e3d/trin-core/src/portalnet/types/messages.rs#L464 :)

njgheorghita: Hahah, yeah looks like it was my crappy error message - mind updating that one as well here?

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -87,12 +87,18 @@ impl JsonRpcEndpoint {
             JsonRpcEndpoint {
                 method: "portal_statePing".to_string(),
                 id: 6,
-                params: Params::Array(vec![Value::String(buddy_node.clone())]),
+                params: Params::Array(vec![
+                    Value::String(buddy_node.clone()),
+                    Value::String("010100".to_string()),
```
jacobkaufmann, line 92: My understanding is that we do not implement `validate_endpoint_response` here because that will be taken up in #211. Is that correct?

ogenev: Yes, correct, it is just a placeholder in this PR.

jacobkaufmann (approved):
LGTM. Only comment would be addition of basic unit tests, but I think this is covered partially through our test vectors.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -19,6 +19,66 @@ use crate::portalnet::{types::uint::U256, Enr};
 
 pub type ByteList = VariableList<u8, typenum::U2048>;
 
+/// Custom payload element of Ping and Pong overlay messages
+#[derive(Debug, PartialEq, Clone)]
+pub struct CustomPayload {
+    payload: ByteList,
+}
```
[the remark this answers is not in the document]
ogenev: Not sure either. Using it as a newtype and implementing `Deref` is considered an anti-pattern for some of the Rust community: 
https://stackoverflow.com/questions/45086595/is-it-considered-a-bad-practice-to-implement-deref-for-newtypes

So I prefer to leave it as it is for now and use `self.payload` instead of `self.0`...

PR #211, opened 2021-12-21
Make Ping JSON-RPC result human readable

ogenev:
Return JSON Value from Pong ssz bytes on json-rpc portal_*Ping requests and update tests.

Fixes https://github.com/ethereum/trin/issues/198

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -11,7 +11,6 @@ use thiserror::Error;
 
 use crate::cli::PeertestConfig;
 use trin_core::jsonrpc::types::Params;
-use trin_core::portalnet::types::uint::U256;
 
 #[derive(Clone)]
 pub struct JsonRpcEndpoint {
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
Looks good! Just curious why you went with camelCase here?

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -38,18 +40,30 @@ fn validate_endpoint_response(method: &str, result: &Value) {
...
         "portal_historyPing" => {
-            assert!(result.is_string());
-            // TODO test for enrSeq, recipient{IP,Port}
+            assert_eq!(
+                result.get("dataRadius").unwrap().as_str().unwrap(),
+                DATA_RADIUS
+            );
+            assert_eq!(
+                result.get("enrSeq").unwrap().as_str().unwrap(),
+                1.to_string()
```
jacobkaufmann, line 55: We could use a const for one here to be consistent with data radius.

jacobkaufmann (approved):
Looks good! Looks like camelCase here is according to the [JSON-RPC spec](https://github.com/ethereum/portal-network-specs/pull/88).

ogenev:
> Looks like camelCase here is according to the [JSON-RPC spec](https://github.com/ethereum/portal-network-specs/pull/88)

Yeah, we decided to go with this route mainly because of this guide: https://google.github.io/styleguide/jsoncstyleguide.xml.

PR #212, opened 2022-01-03
Update Discv5 version

### LINKED ISSUE #182: Update discv5 to v0.1.0-beta.12
The #178 PR required a change that has now been merged upstream, but not released yet, as of v0.1.0-beta.11. When v0.1.0-beta.12 is released, update to use that instead of the `fewer-mutable-borrows` branch in carver/trin.

njgheorghita:
Fixes #182 
Maybe fixes # 186

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
             name: rust/default
             tag: 1.56.1
         environment:
-            RUSTFLAGS: '-D warnings'
+            RUSTFLAGS: '-Dwarnings'
         steps:
             - checkout
             - restore_cache:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ jobs:
             name: rust/default
             tag: 1.56.1
         environment:
-            RUSTFLAGS: '-Dwarnings'
+            RUSTFLAGS: '-D warnings'
         steps:
             - checkout
             - restore_cache:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,36 +13,30 @@ dependencies = [
 
 [[package]]
 name = "aes"
-version = "0.7.4"
+version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "495ee669413bfbe9e8cace80f4d3d78e6d8c8d99579f97fb93bde351b185f2d4"
+checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
 dependencies = [
  "cfg-if 1.0.0",
  "cipher",
- "cpufeatures",
- "ctr 0.7.0",
+ "cpufeatures 0.2.1",
+ "ctr",
  "opaque-debug",
 ]
 
 [[package]]
 name = "aes-gcm"
-version = "0.9.3"
+version = "0.9.4"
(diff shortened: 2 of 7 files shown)


njgheorghita:
@carver Ping for thoughts on #186 - I'm a bit baffled as to what's going on with this one. IMO it comes down to the kind of error that's being emitted (`ignoring invalid dependency `trin-cli` which is missing a lib target`).  

From what I can tell, we're well set up to raise this warning as an error. We have `RUSTFLAGS="-D warnings"` set in in our ci environment. We even specifically run `cargo clippy` with `--deny warnings`. So it seems to me like this warning, which appears to be a rustc warning and not a regular linting warning (which are all raised as errors in our workflow via clippy) should be flagged as an error by cargo.

I've tried a couple other approaches to catch this warning
- Adding a `.cargo/config.toml` to pass in `RUSTFLAGS`
- Using `#![deny(warnings)]` inside crate roots
- Using `"-Dwarnings"` syntax for `RUSTFLAGS` which is syntax I've seen used fairly frequently

IMO it comes down to us using multiple crates in a workspace. It seems to be a fairly problematic setup for various config / passing rustc settings to cargo ([here](https://users.rust-lang.org/t/how-to-fail-a-build-with-warnings/2687/10) and [here](https://github.com/rust-lang/cargo/issues/8424))

Any thoughts as to whether you want me to fix the warning in this pr (by adding a `lib.rs`) and leave the issue open/closed?

ogenev:
> Any thoughts as to whether you want me to fix the warning in this pr (by adding a `lib.rs`) and leave the issue open/closed?

I addressed this issue in https://github.com/ethereum/trin/pull/209. (sorry for not including it in the PR description). I think removing `trin-cli` from workspace dependencies makes sense, because the crate is not a library (no lib.rs file) and `trin` can't depend on a binary crate.

njgheorghita:
@ogenev Oh nice, yeah that makes sense to me. I've removed the redundant `lib.rs` from this pr. Still waiting final confirmation re: open/closing #186 from @carver since the motivation behind the issue was more about generally catching warnings rather than specifically fixing that warning (re [discussion here](https://github.com/ethereum/trin/pull/188#issuecomment-971868924))

.circleci/config.yml
```diff
@@ -10,7 +10,7 @@ jobs:
             name: rust/default
             tag: 1.56.1
         environment:
-            RUSTFLAGS: '-D warnings'
+            RUSTFLAGS: '-Dwarnings'
```
ogenev, line 13: Is this a typo, shouldn't be `-D warnings`?

njgheorghita: I don't believe so. In many of the various blogs/forums I read about this I saw both `-D warnings` and `-Dwarnings` used interchangeably as the `RUSTFLAGS` env variable. But, I've changed it back to its original `-D warnings` since yeah I'd agree it's more obviously not a typo

ogenev (approved):

carver:
> the motivation behind the issue was more about generally catching warnings rather than specifically fixing that warning (re [discussion here](https://github.com/ethereum/trin/pull/188#issuecomment-971868924))

Yeah, this is definitely the reason for #186, to prevent new warnings to show up. The existing warning was just a convenient way to test that we can configure CI to catch that special type of warning.

So in summary:
1) Don't close #186, since we can't confirm that we can catch warnings in CI
2) Let's remove any attempts at flagging warnings, since we can't confirm that they work

Issue #210, opened 2021-12-21
Implement overlay bucket refresh procedure


jacobkaufmann:
From the Kademlia paper section 2.3: "Buckets are generally kept fresh by the traffic of requests traveling through nodes. To handle pathological cases in which there are no lookups for a particular ID range, each node refreshes any bucket to which it has not performed a node lookup in the past hour. Refreshing means picking a random ID in the bucket's range and performing a node search for that ID."

This refresh procedure is also used when a node joins a node network to bootstrap the local routing table.

PR #213, opened 2022-01-03
Overlay bucket refresh lookup

ogenev:
Implement bucket refresh lookup algorithm based on @pipermerriam's suggestion:

_A: Look at the routing table and select the first N buckets which are not full.  Any value of N < 10 is probably fine here.
B: Randomly pick one of these buckets.  Weighting this random selection to prefer "larger" buckets can be done here to prioritize finding the easier to find nodes first.
C: Randomly generate a NodeID that falls within this bucket.
D: Do the random lookup on this node-id._

#### 1. Random bucket index selection

We decided to pick a random bucket with even weight, except we want to skip the closest buckets that will all be empty which is probably only the first 17 buckets.

There is a discussion about this in the comments.

#### 2. Distance metric and the bucket index

From disv5 implementation, we know that the distance metric falls in a bucket index based on the leading zeroes in the 256-bit representation of the distance, where `distance_leading_zeroes = 255 - target_bucket_index`. 

So if we want a distance to fall in a bucket index 5, we need to provide a 256-bit distance with 250 leading zeroes.
To accomplish this, we are generating a random 32-byte array with a set number of leading bit zeros.

#### 3. Generate random target NodeId

We know that `local_node_id XOR target_node_id = distance` so if we have the generated random distance and the local nodeid, we can XOR those values to get the random target node id because the inverse function to XOR is XOR itself.
This way we can make sure the random NodeID will fall in the target bucket index.

#### 4. Upstream discv5 changes

To do all of the above, we need at least access to all buckets in the routing table (not only the nodes inside them) and mainly for testing - a method to get a bucket index based on a `discv5::Key`. 

This PR is using a patched disv5 version with the following changes:
https://github.com/ogenev/discv5/commit/963536043c9f7bea86cab04e2ed751c9b708e438

I tried to be as minimalistic as possible adding public methods to the discv5 library but If we don't find another better way, we may need to try merging this branch upstream.

#### 5. Testing

I'm not sure how we can test the main bucket refresh lookup logic, because of the randomness introduced with the bucket selection. I added tests for the binary string generation and checked if a random node id falls in the target bucket.

This PR also introduces the `anyhow` crate for error handling ( https://github.com/ethereum/trin/issues/214).

Fixes https://github.com/ethereum/trin/issues/210


trin-core/src/utils/binary_string.rs
```diff
@@ -0,0 +1,66 @@
...
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, Clone)]
+pub enum BinaryStringError {
+    #[error("Unable to parse binary string!")]
+    ParseError(#[from] ParseIntError),
+}
+
+/// Generate random 256-bit binary string with leading zeroes
+pub fn generate_random(leading_zeroes: u8) -> Result<String, BinaryStringError> {
```
jacobkaufmann, line 12: When would this ever return a `Result::Err`?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -365,10 +389,64 @@ impl OverlayService {
...
         }
     }
 
+    /// Generate random NodeId based on bucket index target.
+    /// First we generate a random distance metric with leading zeroes based on the target bucket.
+    /// Then we XOR the result distance with the local NodeId to get the random target NodeId
+    fn generate_random_node_id(&self, target_bucket_idx: u8) -> Result<NodeId, RandomNodeIdError> {
+        let distance_leading_zeroes = 255 - target_bucket_idx;
+        let random_distance = binary_string::generate_random(distance_leading_zeroes)?;
+        let random_distance = binary_string::to_byte_array(&random_distance)?;
```
jacobkaufmann, line 406: Is there a reason for the intermediate generation of a string? It seems that we should be able to generate a random byte array directly, and this function should never fail.

trin-core/src/utils/binary_string.rs
ogenev: Ah, this is a leftover from some previous versions of the method. It is fixed now.

trin-core/src/portalnet/overlay_service.rs
ogenev: We are able to generate a random byte array directly but we want to have control over the leading zeroes in the 256-bit value. This means we will need additional logic to decide what byte value range we want for every 8 bits. For example, if we want to generate a random 32 size byte array with N leading bit zeroes, we need to do something like:

```python
first_zero_bytes = N // 8
leading_zeroes_in_next_byte = N % 8

If leading_zeroes_in_next_byte == 7
   next_byte_value = 1
elif leading_zeroes_in_next_byte == 6
   next_byte_value = random.randrange(3)
elif leading_zeroes_in_next_byte == 5
   next_byte_value = random.randrange(5)
elif leading_zeroes_in_next_byte == 4
   next_byte_value = random.randrange(9)
 .......
 ```
 
 Then we have to prefix the final result with the first zero bytes, append the next non-zero byte, and finally generate random bytes for the rest of the array.
 
Not sure if this or the current implementation using a binary string is more clear/performant. Happy to change it if we agree on it.

Comment (jacobkaufmann) [reply trin-core/src/portalnet/overlay_service.rs]: I'm less concerned with performance and more concerned with unnecessary error handling.

I was thinking we could something like the following.

```rust
fn random_32_byte_array(leading_zeros: usize) -> Vec<u8> {
    let first_zero_bytes = leading_zeros / 8;
    let first_nonzero_byte_leading_zeros = leading_zeros % 8;

    let mut bytes: Vec<u8> = vec![];
    for i in 0..first_zero_bytes {
        vec.push(0);
    }

    let mut random_bytes = [0u8, 32 - first_zero_bytes];
    rand::thread_rng().fill_bytes(&mut random_bytes);
    random_bytes[0] &= (u8::MAX >> first_nonzero_byte_leading_zeros);

    bytes.append(random_bytes.into());

    bytes
}
```

Not sure that the code above works as written, but essentially use some bitwise operations to do this.

carver:
> Our approach now is to set the weights based on the number of nodes in a bucket + 1.

I have concerns about this approach. It seems like it would get stuck in a feedback loop of preferring to keep the (nearly) full buckets full, and empty-ish buckets empty. Also, when all of the buckets are empty, it doesn't seem to prefer the large buckets correctly. My read of Piper's suggestion was to prefer the choice by size of bucket, rather than number of peers in bucket.

One approach would be to pick a random number for the node ID from the full space. You get the node ID and the bucket implicitly in the same step. That would naturally prefer the bigger buckets. It would take more work to exclude the full buckets, of course. It might make sense to split it back into two steps, but get the same distribution (ie~ every bucket that has a 2x bigger space is 2x more likely to get chosen, if it's not full)

Maybe we want to prefer close nodes a *bit* more than that, but I don't have much intuition for a "natural" distribution. I guess half-way in between is an option. In one construction, every bucket is the same weight. In another, each bucket is 2x more weighted than the last. So in the middle would be 1.414x more weight from one bucket to the next (root 2, the geometric mean).

carver:
> Implementation details

I love the way you wrote all this up. It was really helpful introduction to the PR.


> I tried to be as minimalistic as possible adding public methods to discv5 library but If we don't find another better way, we may need to try merging this branch upstream.

Yeah, it looks like we'll definitely want to make *some* change upstream. I haven't thought too hard about this, but what do you think about making the key and the buckets be public variables? That minimizes the changes even more, and we can do the rest of the work we need to do in trin.

ogenev:
> I have concerns about this approach. It seems like it would get stuck in a feedback loop of preferring to keep the (nearly) full buckets full, and empty-ish buckets empty. Also, when all of the buckets are empty, it doesn't seem to prefer the large buckets correctly. My read of Piper's suggestion was to prefer the choice by size of bucket, rather than number of peers in bucket.

Apparently, I didn't get it right. Using the number of nodes in a bucket doesn't make sense.

Thinking about it now, I guess "larger" buckets are buckets with a large bucket index, not size (every bucket is of size 16). 
So then, weighting buckets with large indexes means prioritizing finding nodes that are closer to the local node id.

One concern I have here is that the probability to find a node with a closer distance is very low, probably any bucket with an index higher than 20 is very unlikely to have a node ever, so then what is the point to try to refresh those larger buckets closer to us? If my math is right, the probability to find a node in let's say bucket index 50 is something like 1 in 2^50, for index 250 - 1 in 2^250...

ogenev:
> Yeah, it looks like we'll definitely want to make some change upstream. I haven't thought too hard about this, but what do you think about making the key and the buckets be public variables? That minimizes the changes even more, and we can do the rest of the work we need to do in trin.

I thought about this, we can make the `buckets` field in `KBucketsTable` public but I'm with the impression that the library maintainers prefer exposing public methods accessing those private fields and returning an iterator rather than making the fields public. But maybe this is only my assumption and it is worth seeing what have to say.

carver:
> One concern I have here is that the probability to find a node with a closer distance is very low, probably any bucket with an index higher than 20 is very unlikely to have a node ever, so then what is the point to try to refresh those larger buckets closer to us?

Right, this is exactly why I read the suggestion the other way around. So the size of the bucket is the number of node IDs in the space. The bucket with one bit distance away only has at most 1 other peer in it. 2 bits has at most 3 other peers, 3 bits has at most 7 other peers, etc. So yes, we actually want to prefer the buckets that are larger (ie~ farther), because it's more likely that there will be some on the network.

We have a *lot* of flexibility here, and it should be straightforward to update the strategy over time. So it probably makes sense to do the easiest possible thing for now that doesn't feel broken, and move on.

Picking a random bucket with even weight is probably fine. Except we want to skip the closest buckets that will all be impty. Assuming about ~100k nodes (more than 10x more than ethereum mainnet), my back-of-the-envelope math says we only need the 17 bits furthest from our own node ID. In other words, the closest 239 bits of buckets are going to be empty-ish.

trin-core/src/portalnet/overlay_service.rs
ogenev: It turns out the logic should be a little more complex but I updated the PR to generate a random 32-byte array without intermediate binary string generation. Let me know what you think!

njgheorghita (approved):
lgtm! The new error handling library looks great as well

Cargo.toml
```diff
@@ -28,3 +28,7 @@ members = [
     "trin-cli",
     "ethportal-peertest"
 ]
+
+[patch.crates-io]
+# Use temporaly discv5 version
+discv5 = { git = "https://github.com/ogenev/discv5.git", branch = "public_kbucket" }
```
carver, line 34: Why use this instead of something like:
```suggestion
[dependencies.discv5]
branch = "public_kbucket"
git = "https://github.com/ogenev/discv5"
```

It seems like `patch.crates-io` requires also installing the patch crate and running `cargo patch`, based on: https://crates.io/crates/cargo-patch

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -917,12 +955,30 @@ impl OverlayService {
...
+    /// First we generate a random distance metric with leading zeroes based on the target bucket.
+    /// Then we XOR the result distance with the local NodeId to get the random target NodeId
+    fn generate_random_node_id(&self, target_bucket_idx: u8) -> anyhow::Result<NodeId> {
+        let distance_leading_zeroes = 255 - target_bucket_idx;
+        let random_distance = bytes::random_32byte_array(distance_leading_zeroes);
+        let raw_node_id = xor_two_values(&self.local_enr().node_id().raw(), &random_distance)?;
+        match NodeId::parse(raw_node_id.as_slice()) {
+            Ok(node_id) => Ok(node_id),
+            Err(msg) => Err(anyhow::Error::msg(msg)),
+        }
```
carver, line 973: I think the `anyhow` docs suggest that this has the same effect:
```suggestion
        let parsed_id = NodeId::parse(raw_node_id.as_slice())?;
        Ok(parsed_id)
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -365,7 +369,41 @@ impl OverlayService {
...
+    }
+
+    /// Main bucket refresh lookup logic
+    fn bucket_refresh_lookup(&self) {
+        // Look at local routing table and select the largest 17 buckets.
+        // We only need the 17 bits furthest from our own node ID, because the closest 239 bits of
+        // buckets are going to be empty-ish.
+        let buckets = self.kbuckets.read();
+        let buckets = buckets.buckets_iter().enumerate().collect::<Vec<_>>();
+        let buckets = &buckets[239..];
```
carver, line 387: We probably want to move this to a constant somewhere, like `256-EXPECTED_NON_EMPTY_BUCKETS`, and a comment by there can explain how we got to 17 (estimating 100k nodes, which is more than 10x the ethereum mainnet node count). 2**17 is enough to put each node into a unique bucket by the 17th bucket index, with some hand-waving of even distribution assumptions.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -365,7 +369,41 @@ impl OverlayService {
                     }
                 }
                 _ = OverlayService::bucket_maintenance_poll(self.protocol.clone(), &self.kbuckets) => {}
+                _ = bucket_refresh_interval.tick() => {
+                    debug!("Bucket refresh lookup");
```
carver, line 373: Probably won't want this in long-term, but I'm fine leaving it in while the code is fresh.

carver (approved):

trin-core/src/portalnet/overlay_service.rs
ogenev: Yeah, true but the weird thing is that `NodeId::parse` returns `Result<Self, &'static str> ` instead of `Result<Self, std::error::Error>`, and this force us to handle the string manually...

carver: Ah I see, thanks for clarifying. I guess we can add it to the pile of nice-to-have features for discv5.

Cargo.toml
carver: Ah nice, thanks for the link

PR #215, opened 2022-01-10
Add Konrad's test vectors for uTP

KolbyML:
I added the test vectors Konrad's detailed in his PR https://github.com/ethereum/portal-network-specs/pull/127

### DIFF
--- a/trin-core/src/portalnet/utp.rs
+++ b/trin-core/src/portalnet/utp.rs
@@ -1013,4 +1013,123 @@ mod tests {
         assert_eq!(reply.get_extensions()[0].bitmask[2], 0b0101_1010);
         assert_eq!(reply.get_extensions()[0].bitmask[3], 0b0000_0001);
     }
+
+    // https://github.com/ethereum/portal-network-specs/pull/127
+    // konrad uTP test vectors
+
+    #[test]
+    fn test_syn_packet() {
+        let mut response = PacketHeader::new();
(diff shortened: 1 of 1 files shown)


trin-core/src/portalnet/utp.rs
```diff
@@ -1013,4 +1013,123 @@ mod tests {
...
+        response.set_seq_nr(16807);
+        response.set_ack_nr(11885);
+
+        let mut incoming_buffer: BTreeMap<u16, Packet> = Default::default();
+        incoming_buffer.insert(12050, Packet::new(PacketHeader::new()));
+        incoming_buffer.insert(12081, Packet::new(PacketHeader::new()));
+
+        let mut reply = Packet::new(response);
+
+        if reply.seq_nr().wrapping_sub(12048) > 1 {
```
njgheorghita, line 1073: I'm not closely familiar with the utp spec, but it feels a bit weird that we'd have this conditional in the test as it appears that the test should be deterministic - what's the need for the `if` statement?

njgheorghita (approved):
Sweet! Glad to see all the cases are passing

trin-core/src/portalnet/utp.rs
KolbyML: I refactored set_selective_ack and removed that cause you are correct this isn't needed here, but it is needed in the actual code

PR #217, opened 2022-01-11
Fix local closest nodes lookup to use content ID

### LINKED ISSUE #216: Use `content_id` inside `OverlayService.find_nodes_close_to_content()`
If I'm not misunderstanding things, we should be using `content_id` instead of `content_key` to [calculate distances](https://github.com/ethereum/trin/blob/d5413c58ec55216f266dd68db0d528bb76eb5e39/trin-core/src/portalnet/overlay_service.rs#L882) in https://github.com/ethereum/trin/blob/d5413c58ec55216f266dd68db0d528bb76eb5e39/trin-core/src/portalnet/overlay_service.rs#L882. Right now this throws an error since we cannot `xor` two vectors of different lengths (`node_id` and `content_key`)

I played around with a couple solutions, but didn't find anything suitably simple. If anybody has some ideas, lmk or feel free to take this on. Otherwise I'll pick this back up asap

jacobkaufmann:
Fix #216

Use content ID instead of content key to perform local lookup for nodes closest to some content.

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -8,6 +8,7 @@ use crate::{
     portalnet::{
         discovery::Discovery,
         types::{
+            content_keys::{ContentKeyError, HistoryContentKey, StateContentKey},
             messages::{
                 ByteList, Content, CustomPayload, FindContent, FindNodes, Message, Nodes, Ping,
                 Pong, ProtocolId, Request, Response, SszEnr,
@@ -82,6 +83,12 @@ impl From<discv5::RequestError> for OverlayRequestError {
             discv5::RequestError::Timeout => Self::Timeout,
             err => Self::Discv5Error(err),
         }
+    }
+}
+
+impl From<ContentKeyError> for OverlayRequestError {
+    fn from(err: ContentKeyError) -> Self {
+        Self::InvalidRequest(format!("Invalid content key: {}", err))
     }
 }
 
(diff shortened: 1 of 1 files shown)


jacobkaufmann:
I went with the shortest solution here. Although, I think we should modify `ContentKey` to leverage the `From` and `Into` traits. I plan on taking this up as an enhancement.

njgheorghita (commented):
Ahh, cool. I missed that there was the `OverlayService.protocol` attribute available to match on, so I was working towards something more complex. So this solution is definitely great for the time being - it's a working solution, I have no problem merging it for now. But I'm a bit nervous as to how sustainable this will be. With every overlay protocol we add, the clutter here will increase. 

I kinda see the optimal solution for this being similar to the solution for #217 - and by that I mean the logic should be moved into the respective crates (`trin-state` / `trin-history`). We have those crates split out for a reason, let's take advantage of that. 

I'm still not certain I've set upon the ideal solution. I was working towards a function that combined content id derivation and calculating the distance. This function (or maybe it's better off as 2 functions (`derive_content_id_from_key()` and `calculate_distance()`) lives in the overlay protocol crate (`trin-history`) and then gets passed into `OverlayProtocol` when it's created. 

Any thoughts?

jacobkaufmann:
I agree that what I've implemented here is not the ideal long-term solution, and that the content-id derivation logic for each protocol should live in its respective crate. I just went for the simplest solution that satisfies correctness.

I do think the long-term solution looks something like what you've described, where each crate implements a trait or a closure to derive a content ID from a content key byte array. I think that the solution for distance looks similar, so that each protocol crate specifies how to derive content IDs from content keys and how to compute distances between identifiers.

njgheorghita (approved):
Cool cool, maybe just leave a comment in the code to track.

jacobkaufmann:
Sorry, got sidetracked yesterday. What about a new issue instead of a `TODO` in the code? I think that issues generally have better visibility.

njgheorghita:
For sure, that sounds good to me

jacobkaufmann:
Created #219. Merging.

PR #220, opened 2022-01-12
Add processing for nodes discovered from Nodes and Content

### LINKED ISSUE #205: Add logic to update routing table according to Nodes and Content responses
For now, #195 only performs updates to the routing table based on `Ping` and `Pong` messages. As outlined in the predecessor #135, we need to update the routing table based on `Nodes` and `Content` responses, where those messages contain [discovered nodes](https://github.com/ethereum/trin/pull/135/files#diff-9973a70f6c6f9e96c941fddd94f03613e9495f036d8f96b15c83738881f4077dR456). (See `process_discovered_nodes` function in the PR diff.)

jacobkaufmann:
Closes #205.

Add logic to process ENRs contained `Nodes` and `Content` responses. We attempt to insert newly discovered nodes into the routing table. If successful, we insert those nodes into the ping queue to establish an outgoing connection. For ENRs that we already have in the routing table, we attempt to update the local routing table entry if the ENR sequence number in the response is higher than what we have locally in the corresponding entry.

Add integration test in `trin-core` to test core overlay routing table logic according to various messages exchanged between a set of nodes. If desired, this piece of the PR can be separated out into its own PR.

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -808,7 +808,6 @@ impl OverlayService {
             "[{:?}] Processing Pong response from node. Node: {}",
             self.protocol, node_id
         );
-
         // If the ENR sequence number in pong is less than the ENR sequence number for the routing
         // table entry, then request the node.
         //
@@ -849,8 +848,19 @@ impl OverlayService {
             self.protocol,
             source.node_id()
         );
-        if let Content::Enrs(enrs) = content {
-            self.process_discovered_enrs(enrs.into_iter().map(|ssz_enr| ssz_enr.into()).collect());
+        match content {
+            Content::ConnectionId(id) => debug!(
+                "[{:?}] Skipping processing for content connection ID {}",
+                self.protocol, id
+            ),
+            Content::Content(_) => {
+                debug!(
+                    "[{:?}] Skipping processing for content bytes",
+                    self.protocol
+                )
+            }
+            Content::Enrs(enrs) => self
+                .process_discovered_enrs(enrs.into_iter().map(|ssz_enr| ssz_enr.into()).collect()),
         }
     }
 
@@ -872,7 +882,7 @@ impl OverlayService {
             }
 
             let key = kbucket::Key::from(node_id);
-            let node = match kbuckets.entry(&key) {
+            let optional_node = match kbuckets.entry(&key) {
                 kbucket::Entry::Present(entry, _) => Some(entry.value().clone()),
                 kbucket::Entry::Pending(ref mut entry, _) => Some(entry.value().clone()),
                 _ => None,
@@ -882,11 +892,11 @@ impl OverlayService {
             // If the node is not in the routing table, then add the node in a disconnected state.
             // A subsequent ping will establish connectivity with the node. If the insertion succeeds,
             // then add the node to the ping queue. Ignore insertion failures.
-            if let Some(present_node) = node {
-                if present_node.enr().seq() < enr.seq() {
+            if let Some(node) = optional_node {
+                if node.enr().seq() < enr.seq() {
                     let updated_node = Node {
                         enr,
-                        data_radius: present_node.data_radius(),
+                        data_radius: node.data_radius(),
                     };
 
                     // The update removed the node because it would violate the incoming peers condition
@@ -1154,7 +1164,10 @@ mod tests {
         utils::db,
     };
 
-    use discv5::enr::{CombinedKey, EnrBuilder};
+    use discv5::{
+        enr::{CombinedKey, EnrBuilder},
+        kbucket::Entry,
+    };
     use rand::Rng;
     use tokio_test::{assert_pending, assert_ready, task};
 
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
:fire: Love the comments, they were super helpful for understanding the code. The tests are also great. Sooner or later I'm expecting that we should see somewhat complementary tests in `peertest`? I'm not sure exactly what form it'll take but this seems like an important function that we'd want to test in a simulation manner

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -827,23 +827,116 @@ impl OverlayService {
...
+    fn process_content(&mut self, content: Content, source: Enr) {
         debug!(
             "[{:?}] Processing Content response from node. Node: {}",
             self.protocol,
             source.node_id()
         );
-        // TODO: Implement processing logic.
+        if let Content::Enrs(enrs) = content {
+            self.process_discovered_enrs(enrs.into_iter().map(|ssz_enr| ssz_enr.into()).collect());
+        }
```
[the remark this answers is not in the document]
jacobkaufmann: It doesn't look like we are explicitly tracking that anywhere. I suppose these are sort of implicit in our goals to add support for content transmission, but I can either add some `debug` logs or create an issue or both.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -827,23 +827,116 @@ impl OverlayService {
...
+
+            // Ignore ourself.
+            if node_id == local_node_id {
+                continue;
+            }
+
+            let key = kbucket::Key::from(node_id);
+            let node = match kbuckets.entry(&key) {
+                kbucket::Entry::Present(entry, _) => Some(entry.value().clone()),
+                kbucket::Entry::Pending(ref mut entry, _) => Some(entry.value().clone()),
```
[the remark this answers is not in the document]
jacobkaufmann: It's been some time since I originally looked at this, but taking another look at the [bucket update function](https://github.com/sigp/discv5/blob/9b21ec1e274138661fd2fa35bef54143099f2ef2/src/kbucket/bucket.rs#L482), it looks like it is defined to handle both present and pending entries.

I do think that the `present_node` variable name below is a bit confusing, given that the node may be present or pending, so I can change that to something more neutral.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
njgheorghita: Yeah, it seems like we can't possibly let this slip through the cracks. Idk, maybe just a debug log for now will be helpful

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -827,23 +827,116 @@ impl OverlayService {
...
+                        debug!(
+                            "Failed to update discovered node. Node: {}, Reason: {:?}",
+                            node_id, reason
+                        );
+                    }
+                }
+            } else {
+                let node = Node {
+                    enr,
+                    data_radius: U256::from(u64::MAX),
```
ogenev, line 907: I think this is for another PR but it is probably worth adding a data structure for the radius, something like `DataRadius`  and implementing Default trait, so this line will look something like:

```rust
data_radius: DataRadius::default()
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1327,11 +1420,129 @@ mod tests {
...
+        // Check routing table for second ENR.
+        // Node should be present with ENR sequence number equal to 1.
+        match service.kbuckets.write().entry(&key2) {
+            kbucket::Entry::Present(entry, _status) => {
+                assert_eq!(1, entry.value().enr.seq());
+            }
+            _ => panic!(),
+        };
+    }
+
```
ogenev, line 1540: Maybe it is worth also checking the use case when we process our own local `Enr` and ignoring it.

trin-core/tests/overlay.rs
```diff
@@ -0,0 +1,223 @@
...
+        listen_port: 8001,
+        internal_ip: true,
+        ..PortalnetConfig::default()
+    };
+    let mut discovery_one = Discovery::new(portal_config_one).unwrap();
+    let _ = discovery_one.start().await.unwrap();
+    let discovery_one = Arc::new(discovery_one);
+    let overlay_one = Arc::new(init_overlay(Arc::clone(&discovery_one), protocol.clone()).await);
+    spawn_overlay(Arc::clone(&discovery_one), Arc::clone(&overlay_one)).await;
+    time::sleep(Duration::from_millis(5)).await;
```
ogenev, line 107: nitpick: It may be useful to extract the milliseconds' value in a constant above the tests.

ogenev (approved):
LGTM :+1:. Love the tests and comments too :100:.

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann: I was thinking a constant, but like you said, I think this is best left for a later PR.

PR #221, opened 2022-01-13
Fix content lookup

njgheorghita:
This is a fairly minor bugfix, basically the fix that was needed in #216 is also required for direct db fetches. Data is stored under its `content_id` - looking it up directly via its `content_key` will return nothing.

But this brings up another question... Right now, we're in a bit of a limbo between two db paradigms (a simple rocksdb / `PortalStorage`). The script (`trin-core/src/bin/generate-data.rs`) I'm using to seed the database uses `PortalStorage`. `trin-core` currently uses a simple rocksdb strategy. 

The weird thing about this bugfix is that it's currently needed, but once `PortalStorage` is integrated, it's no longer needed (since `PortalStorage` handles the content id derivation).

I've talked with @mrferris about this recently. The decision to hold off on `PortalStorage` integration was made since it's not the highest priority. But, based off this issue, I'm starting to lean towards bumping that priority higher, and I'm happy to pick up that task up myself. It also seems, like we're on track for our q1 goals (but please correct me if this sentiment isn't shared). Another option is to change the seed script I'm using to use the simple rocks db strategy, but there are a lot of advantages to the `PortalStorage` implementation, and since we'll need it eventually, I figure we might as well start leveraging it.

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -488,13 +488,38 @@ impl OverlayService {
 
     /// Attempts to build a `Content` response for a `FindContent` request.
     fn handle_find_content(&self, request: FindContent) -> Result<Content, OverlayRequestError> {
-        match self.db.get(&request.content_key) {
+        // Attempt to derive the content ID for the content key.
+        let content_id = match self.protocol {
+            ProtocolId::State => {
+                let state_content_key =
+                    StateContentKey::from_bytes(request.content_key.as_slice())?;
+                state_content_key.derive_content_id()?
+            }
+            ProtocolId::History => {
+                let history_content_key =
+                    HistoryContentKey::from_bytes(request.content_key.as_slice())?;
+                history_content_key.derive_content_id()?
+            }
+            _ => {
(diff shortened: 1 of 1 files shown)


jacobkaufmann:
I generally agree that `PortalStorage` integration is something that we could start working toward, but I'm not familiar enough with it to estimate that effort. That said, I don't think it should take priority over the Ping/Pong/FindNodes/Nodes functionality on the test network. It seems like we are on track for our Q1 goals, but interoperability testing may reveal otherwise. One other ripe area of investment is in some basic monitoring capabilities so that we can assess whether we meet the functionality described in our goals.

jacobkaufmann (approved):
Good catch! LGTM.

njgheorghita:
>  One other ripe area of investment is in some basic monitoring capabilities so that we can assess whether we meet the functionality described in our goals.

@jacobkaufmann Yeah, definitely sounds useful. Seems like it could be combined nicely with the testnet monitoring, so I might turn my attention in this direction shortly

@carver Yeah, that would be great. I'd connect with @mrferris to gauge the overall effort that integration will require.

PR #225, opened 2022-01-15
Modify XOR distance function to operate on length 32 byte arrays

jacobkaufmann:
Modify XOR distance function to operate on length 32 byte arrays.

All distance computations should operate on 32-byte (256-bit) values. With this modification, we can remove unnecessary error handling and type conversions. I went with big-endian in this implementation, but if desired, I could add an additional parameter to specify endian-ness.

I think this is the general form that a `Distance` trait should take. I chose not to define that trait here, and defer it to a later PR.

### DIFF
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -1,5 +1,5 @@
 use super::types::uint::U256;
-use crate::utils::{db::get_data_dir, distance::xor_two_values};
+use crate::utils::{db::get_data_dir, distance::xor};
 use discv5::enr::NodeId;
 use hex;
 use log;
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/storage.rs
```diff
@@ -362,28 +362,11 @@ impl PortalStorage {
...
-        }
-
-        let mut array: [u8; 8] = [0, 0, 0, 0, 0, 0, 0, 0];
-        for (index, byte) in vec.iter().take(8).enumerate() {
-            array[index] = byte.clone();
-        }
-
-        u64::from_be_bytes(array)
+        let distance = xor(content_id, &self.node_id.raw());
+        distance.0[3]
```
njgheorghita, line 369: I'm probably confused here, my byte math/knowledge is fairly shallow, I'm sure I'm misunderstanding something. Doesn't big endian mean the most significant bytes are stored first (AIU in the left-most slots => `distance.0[0]`?) - but we are returning the least significant (furthest to the right => `distance.0[3]`) 64 bits (or the 4th of 4x8 byte arrays in the `U256`)?

njgheorghita (approved):
Looks great! Love the cleanup. Just a question for my understanding.

njgheorghita:
Also, just a reminder to add the snippet about big-endianness to the specs.

trin-core/src/portalnet/storage.rs
jacobkaufmann: Yes, but the [U256](https://docs.rs/ethereum-types/0.12.1/ethereum_types/struct.U256.html#) stores the data in little endian internally.

njgheorghita: Ahh gotcha, good to know

PR #226, opened 2022-01-15
Refactor uTP and clean up async/await

KolbyML:
This PR refactors uTP into its own folder, which has been requested numerous times.

It also cleans up a lot of await calls and runs the send talkreq in a tokio::spawn

It also corrects the finalization packet response type

### DIFF
--- a/trin-core/src/utp/packets.rs
+++ b/trin-core/src/utp/packets.rs
@@ -1,0 +1,322 @@
+pub const HEADER_SIZE: usize = 20;
+pub const VERSION: u8 = 1;
+
+#[derive(PartialEq, Debug)]
+pub enum PacketType {
+    Data,
+    Fin,
+    State,
+    Reset,
+    Syn,
+}
+
+impl TryFrom<u8> for PacketType {
+    type Error = &'static str;
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(PacketType::Data),
+            1 => Ok(PacketType::Fin),
+            2 => Ok(PacketType::State),
+            3 => Ok(PacketType::Reset),
+            4 => Ok(PacketType::Syn),
+            _ => Err("invalid packet type"),
+        }
+    }
+}
+
+impl From<PacketType> for u8 {
+    fn from(value: PacketType) -> u8 {
+        match value {
+            PacketType::Data => 0,
+            PacketType::Fin => 1,
+            PacketType::State => 2,
+            PacketType::Reset => 3,
+            PacketType::Syn => 4,
+        }
+    }
+}
+
+impl<'a> TryFrom<&'a [u8]> for PacketHeader {
+    type Error = &'static str;
+
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        if value.len() < HEADER_SIZE {
+            return Err("invalid header size");
+        }
+
+        if value[0] & 0xf != VERSION {
+            return Err("invalid packet version");
+        }
+
+        if let Err(e) = PacketType::try_from(value[0] >> 4) {
+            return Err(e);
+        }
+
+        Ok(PacketHeader::decode(value))
+    }
+}
+
+fn check_extensions(data: &[u8]) -> Result<(), &'static str> {
+    if data.len() < HEADER_SIZE {
+        return Err("invalid header size");
+    }
+
+    let mut extension = data[1];
+    let mut index = HEADER_SIZE;
+
+    // must be at least 4 bytes, and in multiples of 4
+    while extension != 0 {
+        if data.len() < index + 2 {
+            return Err("Invalid packet length");
+        }
+        extension = data[index];
+        let len = data[index + 1] as usize;
+
+        if len == 0 || len % 4 != 0 || len + index + 2 > data.len() {
+            return Err("Invalid Extension Length");
+        }
+
+        index += len + 2;
+    }
+    Ok(())
+}
+
+impl<'a> TryFrom<&'a [u8]> for Packet {
+    type Error = &'static str;
+
+    // todo: Added support to check validity of extensions,
+    // we don't need to worry about checking payloads
+    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
+        match PacketHeader::try_from(value).and(check_extensions(value)) {
+            Ok(_) => Ok(Packet(value.to_owned())),
+            Err(e) => Err(e),
+        }
+    }
+}
+
+impl Default for PacketHeader {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[derive(Debug)]
+pub struct PacketHeader {
+    type_ver: u8,
+    extension: u8,
+    connection_id: u16,
+    // This time is in microseconds
+    timestamp: u32,
+    timestamp_difference: u32,
+    wnd_size: u32,
+    pub seq_nr: u16,
+    ack_nr: u16,
+}
+
+impl PacketHeader {
+    pub fn new() -> Self {
+        PacketHeader {
+            type_ver: u8::from(PacketType::Data) << 4 | VERSION,
+            extension: 0,
+            connection_id: 0,
+            timestamp: 0,
+            timestamp_difference: 0,
+            wnd_size: 0xf000,
+            seq_nr: 0,
+            ack_nr: 0,
+        }
+    }
+
+    pub fn encode(&mut self) -> Vec<u8> {
+        let mut buf = vec![];
+        buf.extend_from_slice(&[self.type_ver]);
+        buf.extend_from_slice(&[self.extension]);
+        buf.extend_from_slice(&self.connection_id.to_be_bytes());
+        buf.extend_from_slice(&self.timestamp.to_be_bytes());
+        buf.extend_from_slice(&self.timestamp_difference.to_be_bytes());
+        buf.extend_from_slice(&self.wnd_size.to_be_bytes());
+        buf.extend_from_slice(&self.seq_nr.to_be_bytes());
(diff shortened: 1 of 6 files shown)


ogenev:
Is this PR part of https://github.com/ethereum/trin/pull/84 ?

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+
+        if let Err(e) = PacketType::try_from(value[0] >> 4) {
+            return Err(e);
+        }
+
+        Ok(PacketHeader::decode(value))
+    }
+}
+
+fn check_extensions(data: &[u8]) -> Result<(), &'static str> {
```
ogenev, line 60: Same as above, let's change those `&'static str` with custom errors.

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+            timestamp_difference: u32::from_be_bytes([bytes[8], bytes[9], bytes[10], bytes[11]]),
+            wnd_size: u32::from_be_bytes([bytes[12], bytes[13], bytes[14], bytes[15]]),
+            seq_nr: u16::from_be_bytes([bytes[16], bytes[17]]),
+            ack_nr: u16::from_be_bytes([bytes[18], bytes[19]]),
+        }
+    }
+
+    pub fn version(&self) -> u8 {
+        self.type_ver & 0xf
+    }
```
ogenev, line 159: Nitpick: I think line spacing between methods helps with readability

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+    pub fn set_version(&mut self, int: u8) {
+        self.type_ver = (self.type_ver & 0xf0) | (int & 0xf)
+    }
+    pub fn type_(&self) -> PacketType {
+        PacketType::try_from(self.type_ver >> 4).unwrap()
```
ogenev, line 164: Should we handle this error instead of unwrapping it?

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+
+#[derive(Debug)]
+pub struct Extension {
+    pub extension_type: u8,
+    pub len: u8,
+    pub bitmask: Vec<u8>,
+}
+
+#[derive(Clone, Debug)]
+pub struct Packet(pub Vec<u8>);
```
ogenev, line 203: To get rid of the `get_header()` repetitions, we can add `header` and a `payload` fields in this struct and then update those on initialization with `new` or `with_payload` methods.

KolbyML:
> Is this PR part of #84 ? For someone who is new to uTP, I would like to see more documented source code as described [here](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/documentation.html).

The only part of the PR what was in #84 was the refactor, but I thought I would take it out cause it is easier to review and would probably get in quicker.

Plus there is two other changes I wanted to make

trin-core/src/utp/packets.rs
KolbyML: Because of how data is handled after the header, it would just be cleaner to store the header with the vector as a tuple then anything that would be stored after the header would be in the Vec extension payload etc.

I like this solution because it is very readable though

trin-core/src/utp/packets.rs
KolbyML: This will never report an error, it is deserialized and checked if there is an issue when we originally decode the incoming packet, we have to do try_from just because of Rust unless we want to use unsafe to do this.

ogenev: What do you mean this can't ever fail? The `TryFrom` trait always return a `Result<T, E>` and we implement it here:

```rust
impl TryFrom<u8> for PacketType {
    type Error = &'static str;

    fn try_from(value: u8) -> Result<Self, Self::Error> {
        match value {
            0 => Ok(PacketType::Data),
            1 => Ok(PacketType::Fin),
            2 => Ok(PacketType::State),
            3 => Ok(PacketType::Reset),
            4 => Ok(PacketType::Syn),
            _ => Err("invalid packet type"),
        }
    }
}
```

So if we pass a `value = 5` it will fail. There is a `From` trait that doesn't return an error. I understand that in the current implementation we may have additional assertions to never pass invalid value but in a future refactoring, for example, this may not be the case and we want our code to be foolproof especially when we are building a utp library.

KolbyML: image

We check if it is invalid when we first receive the packet from a 3rd party client before we even handle it. As seen in the picture. Internally we can't set the value to 5, because it can only be set from the enum PacketType.

carver:
👍🏻 👍🏻 for shrinking the PR size of #84 by splitting this out

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    pub fn from_slice(bytes: &[u8]) -> Self {
+        Self(bytes.to_vec())
+    }
+
+    pub fn get_header(&self) -> PacketHeader {
+        PacketHeader::decode(&self.0[0..20])
+    }
+
+    pub fn type_(&self) -> PacketType {
```
njgheorghita, line 227: Is there a reason/convention for the trailing `_`? It's something I haven't seen used in Rust before

KolbyML: type is a keyword in Rust. To be able use the word type we need an _ there are other ways we can do this, but I thought this was the cleanest

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+            seq_nr: u16::from_be_bytes([bytes[16], bytes[17]]),
+            ack_nr: u16::from_be_bytes([bytes[18], bytes[19]]),
+        }
+    }
+
+    pub fn version(&self) -> u8 {
+        self.type_ver & 0xf
+    }
+    pub fn set_version(&mut self, int: u8) {
+        self.type_ver = (self.type_ver & 0xf0) | (int & 0xf)
```
ogenev, line 161: This is really confusing for me, in this method we are setting a private field `self.type_var` and below we have another method that is also setting the same field:

```rust
pub fn set_type(&mut self, t: PacketType) {
        self.type_ver = (self.type_ver & 0xf) | (u8::from(t) << 4)
    }
```
So we have two setters for the same field, what is the case here when to use one over another?

trin-core/src/utp/packets.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    pub fn get_header(&self) -> PacketHeader {
+        PacketHeader::decode(&self.0[0..20])
+    }
+
+    pub fn type_(&self) -> PacketType {
+        self.get_header().type_()
+    }
+
+    pub fn version(&self) -> u8 {
```
ogenev, line 231: What is the reason to have all of those helper methods in the `Packet`? If `Packet` is composed of `PacketHeader` and `PacketHeader` is composed of `type`, `ver`, `extension`, etc, then we can just call `packet.header.version` etc..

KolbyML: It is cleaner to look at in code if we have to call packet.header.version 10 times. packet.version() is a lot nicer to read.

trin-core/src/utp/packets.rs
KolbyML: Please refer to the spec
image

So the smallest int literal is a byte or u8, but it would be wasteful over the wire to have 2 separate bytes for such small values so we split the u8 in half 0000_0000 so the first half will store the type and the second half will store the version of the packet.

If you look at set_type.  there is a 4-bit shift when we right to it and when we use set_version there isn't.

KolbyML: Do you think this would be an acceptable comment for that?

image

ogenev: > So the smallest int literal is a byte or u8, but it would be wasteful over the wire to have 2 separate bytes for such small values so we split the u8 in half 0000_0000 so the first half will store the type and the second half will store the version of the packet.

I see, it would be helpful to add a comment in the code about this.

trin-core/src/utp/packets.rs
ogenev: > It is cleaner to look at in code if we have to call packet.header.version 10 times. packet.version() is a lot nicer to read.

If this is the case, I think we can implement the [Deref](https://doc.rust-lang.org/std/ops/trait.Deref.html) trait and call `packet.version()`.

trin-core/src/utp/packets.rs
ogenev: > Do you think this would be an acceptable comment for that?

I like this more :)

```rust
/// It would be wasteful over the wire to have 2 separate bytes for type and ver, so we split the u8 in half 0000_0000 
/// so the first half will store the type and the second half will store the version of the packet
```
PS: When commenting with `///` instead of `//` this serves as [doc comments](https://doc.rust-lang.org/rust-by-example/meta/doc.html) which is used right above data structures, methods, and fields.

PS 1: I think maybe it is better if you need to post a code in github, to use a [code block](https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting) as ```rust instead of uploading a code screenshot. :thinking:

ogenev (approved):
I'm happy to get this merged ASAP, no blocker from me to address all of the comments. I understand this is meant to be just a slight refactor PR, we can refactor it more in the next iterations.

KolbyML:
> I'm happy to get this merged ASAP, no blocker from me to address all of the comments. I understand this is meant to be just a slight refactor PR, we can refactor it more in the next iterations.

With that, I will get it merged later today, ty for taking the time to review it. Sounds like a plan.

PR #229, opened 2022-01-20
Modify content-key tests to conform to content keys test vectors spec

victor-wei126:
This modifies the tests to conform to the test vectors here: https://github.com/ethereum/portal-network-specs/blob/master/content-keys-test-vectors.md.

Minor changes were made to the hash functions to conform to the updated spec as well.

Also it combines the test for encoding/decoding of key x and converting key x to the respective id into a single test. This reduces lines of code by a lot and also possibility for error.

### DIFF
--- a/trin-core/src/portalnet/types/content_keys.rs
+++ b/trin-core/src/portalnet/types/content_keys.rs
@@ -164,7 +164,7 @@ impl StateContentKey {
                 let mut input = vec![];
                 input.append(&mut path.to_vec());
                 input.append(&mut node_hash.to_vec());
-                Ok(keccak(&hex::encode(input))?)
+                Ok(sha256(&input))
             }
             StateContentKey::ContractStorageTrieNodeKey(ContractStorageTrieNode {
                 address,
@@ -176,18 +176,20 @@ impl StateContentKey {
                 input.append(&mut address.to_vec());
                 input.append(&mut path.to_vec());
                 input.append(&mut node_hash.to_vec());
-                Ok(keccak(&hex::encode(input))?)
+                Ok(sha256(&input))
             }
             StateContentKey::AccountTrieProofKey(AccountTrieProof { address, .. }) => {
-                Ok(keccak(&hex::encode(address.to_vec()))?)
+                Ok(keccak256(address))
             }
             StateContentKey::ContractStorageTrieProofKey(ContractStorageTrieProof {
                 address,
                 slot,
                 ..
             }) => {
-                let address_hash = keccak(&hex::encode(address.to_vec()))?;
-                let slot_hash = keccak(&slot.to_string())?;
+                let address_hash = keccak256(address);
+                let mut slot_as_be_bytes: [u8; 32] = [0; 32];
+                slot.to_big_endian(&mut slot_as_be_bytes);
+                let slot_hash = keccak256(&slot_as_be_bytes);
 
                 let final_hash = U512::from(address_hash) + U512::from(slot_hash);
                 let modulus: U512 = U512::from(2).pow(U512::from(256));
@@ -198,7 +200,7 @@ impl StateContentKey {
                 let mut input = vec![];
                 input.append(&mut address.to_vec());
                 input.append(&mut code_hash.to_vec());
-                Ok(keccak(&hex::encode(input))?)
+                Ok(sha256(&input))
             }
         }
     }
@@ -216,9 +218,9 @@ impl HistoryContentKey {
     /// Returns the content-type of a HistoryContentKey
     pub fn content_type(&self) -> u8 {
         match self {
-            HistoryContentKey::HeaderKey(_) => 0x01,
-            HistoryContentKey::BodyKey(_) => 0x02,
-            HistoryContentKey::ReceiptsKey(_) => 0x03,
+            HistoryContentKey::HeaderKey(_) => 0x00,
+            HistoryContentKey::BodyKey(_) => 0x01,
+            HistoryContentKey::ReceiptsKey(_) => 0x02,
         }
     }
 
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/types/content_keys.rs
```diff
@@ -176,18 +176,20 @@ impl StateContentKey {
                 input.append(&mut address.to_vec());
                 input.append(&mut path.to_vec());
                 input.append(&mut node_hash.to_vec());
-                Ok(keccak(&hex::encode(input))?)
+                Ok(sha256(&input))
             }
             StateContentKey::AccountTrieProofKey(AccountTrieProof { address, .. }) => {
-                Ok(keccak(&hex::encode(address.to_vec()))?)
+                Ok(keccak256(address))
```
njgheorghita, line 182: Is there a reason this hash function was left as `keccak256` while the others were changed to `sha256`?

trin-core/src/portalnet/types/content_keys.rs
```diff
@@ -249,23 +251,27 @@ impl HistoryContentKey {
 
     /// Derives the proper content-id for a given HistoryContentKey
     pub fn derive_content_id(&self) -> Result<U256, ContentKeyError> {
-        let (chain_id, block_hash) = match self {
+        let (content_type, chain_id, block_hash) = match self {
             HistoryContentKey::HeaderKey(HeaderKey {
                 chain_id,
                 block_hash,
-            }) => (chain_id, block_hash),
+            }) => (0x00 as u8, chain_id, block_hash),
```
njgheorghita, line 258: Super nitpick, but `0u8`/`1u8`/`2u8` should be fine here I believe

trin-core/src/portalnet/types/content_keys.rs
```diff
@@ -216,25 +218,25 @@ impl HistoryContentKey {
     /// Returns the content-type of a HistoryContentKey
     pub fn content_type(&self) -> u8 {
         match self {
-            HistoryContentKey::HeaderKey(_) => 0x01,
-            HistoryContentKey::BodyKey(_) => 0x02,
-            HistoryContentKey::ReceiptsKey(_) => 0x03,
+            HistoryContentKey::HeaderKey(_) => 0x00,
+            HistoryContentKey::BodyKey(_) => 0x01,
+            HistoryContentKey::ReceiptsKey(_) => 0x02,
```
njgheorghita, line 223: Hmmm, a bit concerning that we didn't catch this earlier, but glad we're getting it right now. Just for my understanding, the reason we weren't catching it earlier seems to be because we were just checking the round trip encoding / decoding, and not checking that the intermediate value was compliant with the test vectors?

trin-core/src/utils/content_key.rs
```diff
@@ -21,12 +22,16 @@ impl VecToArrayError {
...
     let mut hasher = Keccak256::new();
     hasher.update(key);
-    let result = hasher.finalize().to_vec();
-    // if the underlying hash library is secure, then a 32 len vector will always be returned, so no error should ever occur
-    Ok(U256::from(vec_to_array::<u8, 32>(result)?))
+    let result = &hasher.finalize()[..];
+    U256::from(result)
+}
+
+pub fn sha256(key: &[u8]) -> U256 {
```
njgheorghita, line 32: Would it be possible to switch this out for something more similar to https://github.com/ethereum/trin/blob/master/trin-core/src/bin/generate-data.rs#L77 ? The fewer dependencies the better, so if we can avoid a redundant import in this case (and remove the `hmac_sha256` dependency entirely) that seems preferable

trin-core/src/utils/content_key.rs
```diff
@@ -39,10 +44,20 @@ mod test {
...
-        let test_input = "302938402a9f";
-        let result = keccak(test_input).unwrap();
-        let inter = hex::decode("3fea4dacec93ff6fd6a7462d1a2cc656f7a13361bb36abf0dde7c27a3ca7606d")
+    fn test_keccak256() {
+        let test_input = hex::decode("302938402a9f").unwrap();
+        let result = keccak256(&test_input);
+        let bytes = hex::decode("924fd25eb4fdce477517e48f6e4d25d2fe53139c7ebcd7a1ac102577a465e4ca")
+            .unwrap();
+        let actual = U256::from(vec_to_array(bytes).unwrap());
+        assert_eq!(actual, result);
```
njgheorghita, line 53: While you didn't change it in this pr, I'm just slightly confused by these names. Typically I prefer using `expected` and `actual`.... between `actual` and `result` it's hard for me to tell which value we're testing against. If you don't mind updating those var names in these two tests

trin-core/src/portalnet/types/content_keys.rs
```diff
@@ -435,10 +472,17 @@ mod test {
         } else {
             panic!("Assertions not tested. Check if proper decoding occurred")
         }
+
+        // TESTING CONTENT-ID DERIVATION
+        let actual_id = test_key.to_content_id().unwrap().to_string();
+        let expected_id = String::from(
```
njgheorghita, line 478: *nitpick* - In these conversions throughout the tests, I'd go with `"804....".to_string()` rather than `String::from("804...")` just to be similar with the previous step - but, that's just me, I could be easily convinced if you prefer the way it is

njgheorghita (approved):
Nice! Glad we're getting these cleaned up. Just a couple comment/nitpicks to address

trin-core/src/portalnet/types/content_keys.rs
victor-wei126: Keccak is used for AccountTrieProof and ContractStorageTrieProof because apparently it is necessary to preserve the ordering of the trie data.

trin-core/src/portalnet/types/content_keys.rs
victor-wei126: The spec was changed since last time in order to be able to use a union more easily, starting from the 0th selector (instead of the 1st). But yeah, previously I was only testing if the encoding and then decoding returned the original value, was not doing any intermediate checking.

trin-core/src/utils/content_key.rs
victor-wei126: I would actually prefer it if the `sha3` dependency was removed and we keep the one I have. The hash function that you linked uses sha3_256, from the `sha3` family whereas the one I'm using is from the `sha2` family. I remember talking about this in some detail, that we were specifically going to use the one from the `sha2` family. The one from `sha3` was modified by NIST.

PR #228, opened 2022-01-19
Update generate data script to seed mainnet data

njgheorghita:
- Renamed `generate-data.rs` to `seed-database.rs` and moved it to top-level `/src/bin/` since that helps simplify Ansible playbook configuration
- Updated `seed-database.rs` script to iterate through the `mainnetMM` data dump and seed a trin database with content.
- Adds `portal_stateLocalContent` & `portal_historyLocalContent` json rpc endpoints to help with debugging testnets & their local data

PR #230, opened 2022-01-21
PortalStorage db integration

njgheorghita:
This pr integrates the `PortalStorage` implementation with the rest of the code base. So that we can automatically benefit from its distance functionality. A couple things happened en route to integration....

- Our `rusqlite` library is not threadsafe, so I built an `SqlHandler` type to manage r/w requests to the metadb from the various threads
- Updated some of the function args in `OverlayService` to use our custom `ContentKey` type
- Now that `PortalStorage` uses message channels to communicate with the `SqlHandler` it required some async-ification
- Updated some of our tests to run serially and to use a temporary directory 

This pr got slightly larger than I hoped for, but I'm happy to hop on zoom to discuss parts.


KolbyML:
I think unsafe code is great if it is beneficial. Rust can be very restrictive and this can in some cases cause very unideal code. If unsafe code has a clear benefit (performance, readability, usability, etc) and we know it is actually safe I would go for it.

njgheorghita:
> and we know it is actually safe I would go for it.
@KolbyML the important part, haha. Point taken, but in this case I wasn't quite sure that it was actually safe and seem to have done something odd to my computer

trin-core/src/portalnet/storage.rs
```diff
@@ -127,20 +155,28 @@ impl PortalStorage {
...
-            debug!("Not storing: {}", key);
-            return Ok(());
+            debug!("Not storing: {:02X?}", key);
+            return Err(PortalStorageError::OutsideDistanceError);
         }
 
         // Store the data.
         self.db_insert(&content_id, value)?;
-        self.meta_db_insert(&content_id, &key, value)?;
+        self.meta_db_insert(&content_id, &key.to_bytes(), value)
```
jacobkaufmann, line 178: It seems that there is nothing that would rollback the `db_insert` if the `meta_db_insert` fails here, which would allow the two to move out of sync. Not sure what the best way to enforce this synchronization is, but we need to have error handling for such cases. Similarly for remove.

jacobkaufmann (changes requested):
Took an initial pass and left some comments, but I have some additional general questions.

Did we rule out the possibility of wrapping the `rustqlite::Connection` in a `RwLock` or some other synchronization primitive? My understanding of Rust concurrency is limited, but this seems possible.

Does `PortalStorage` need to be aware of content keys? I see that we store them, but I can't remember why. I don't think we need to address that in this PR, but it seems like an opportunity to simplify the code.

trin-core/src/portalnet/storage.rs
njgheorghita: Hmmmm, I guess we could alway just manually perform the opposite operation on `rocksdb` if `metadb` fails? That seems like the desired behavior right?
Something like...
```rust
self.db_insert(content_id, value)?;
match self.meta_db_insert(....).unwrap() {
    Ok(_) => (),
    Err(_) => self.db_remove(content_id)?,
}
```

trin-core/src/bin/generate-data.rs
```diff
@@ -44,16 +46,17 @@ pub fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-    let mut storage = PortalStorage::new(storage_config, |key| sha256(key))?;
+    let mut storage = PortalStorage::new(storage_config).await?;
 
     for _ in 0..num_of_entries {
         let value = generate_random_value(size_of_values);
         let key = generate_random_value(SIZE_OF_KEYS);
 
-        println!("{} -> {}", &key, &value);
-        storage.store(&key, &value)?;
+        let content_key = ContentKey::StateContentKey(StateContentKey::from_bytes(&key).unwrap());
```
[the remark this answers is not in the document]
njgheorghita: Hahah yeah good catch, random bytes should fail pretty much every time. I sort of glossed over this script since it's mostly obsolete given #228 but I've added a proper error handling for now just in case anybody uses it in the meantime. Though it's also pretty useless if it'll only write random bytes that happen to start with the few possible selectors... but I'm quite confident nobody's actually making use of this script right now.... but if you disagree or know anybody who is lmk - and I'll change it so that it's a bit more useful

njgheorghita:
@jacobkaufmann Thanks for the thoughts!

> Did we rule out the possibility of wrapping the rustqlite::Connection in a RwLock or some other synchronization primitive? My understanding of Rust concurrency is limited, but this seems possible.

My understanding is also quite limited tbh, so I might have made the wrong decision here...
- Wrapping it in a `RwLock` definitely doesn't work. This is what I was using up to the [Only instantiate db handlers once](https://github.com/ethereum/trin/pull/230/commits/96c47f06e9d2688345a7914482cc30147469d874) commit. `rusqlite` doesn't implement `Sync` and therefor `Arc<PortalStorage>` would not be `Send` - and the compiler doesn't let that fly. This is also where I tried using some unsafe code and borked my computer.
- After some reading, mostly [here](https://www.reddit.com/r/rust/comments/6z7gs2/using_rusqlite_from_multiple_threads/) and [here](https://github.com/rusqlite/rusqlite/issues/697), my impression is that 
    1. sqlite itself is not great with async operations
    2. as a result `rusqlite` doesn't easily support async / crossthread operations
- I did tinker around with [sqlx](https://docs.rs/sqlx/latest/sqlx/index.html) as was suggested in one of the threads, but I ran into what I understood as a hard-blocker b/c of how `sqlx` manages the db config path.
- I spoke with @mrferris about this, and he can confirm/correct me if I'm missing something
- Personally I ruled out trying to implement sql-esque queries using just the rocksdb, b/c it was something I was not confident is even possible or I could implement

After this, I went with this `SqlHandler` setup (somewhat inspired by [this comment](https://github.com/rusqlite/rusqlite/issues/697#issuecomment-770261931)). Yeah for sure it's not a trivial amount of added complexity. But to me, it was the solution I was most confident I could implement. And also the complexity seems fairly well contained.

I'm not against trying to go a different path, for sure if you have any thoughts/suggestions lmk and let's explore them.

> Does PortalStorage need to be aware of content keys? I see that we store them, but I can't remember why. I don't think we need to address that in this PR, but it seems like an opportunity to simplify the code.

This is a good point. In this pr, I just followed the existing api. But looking a bit more closely, it really doesn't appear that we make use of the content key at all. We simply store it in the sql db. But we never retrieve it when looking up a value. It seems to me like we could remove the concept of a content-key entirely from `PortalStorage` / `SqlHandler` - which would simplify the code. But this is for sure something @mrferris would need to confirm. Maybe there's some purpose for storing the content keys in the metadb down the road? Either way, yeah let's save that for a future pull request.

trin-core/src/portalnet/storage.rs
njgheorghita: Cool, the solution I implemented is fairly simple, and based mostly under the assumption that if the `rocksdb` operation succeeded before we reach the `metadb` operation, then the corrective/reverting `rocksdb` operation **should** succeed. Tbf this is not a guarantee, but should be good for now. At this point, I do think the greater issue of enforcing synchronicity between the two databases is not over (including some better testing), but is beyond the scope of this pr and can be tracked in an issue. Lmk what you think

ogenev:
> After this, I went with this `SqlHandler` setup (somewhat inspired by [this comment](https://github.com/rusqlite/rusqlite/issues/697#issuecomment-770261931)). Yeah for sure it's not a trivial amount of added complexity. But to me, it was the solution I was most confident I could implement. And also the complexity seems fairly well contained.

Did we check the possibility to use connection pooling like [r2d2](https://github.com/sfackler/r2d2) ? Here is the [sqlite](https://docs.rs/r2d2_sqlite/latest/r2d2_sqlite/) version.

njgheorghita:
@ogenev Great call! Thanks for pointing this out, hadn't yet explored that option, and that definitely seems like the best option to me. I've updated this pr to use a connection pool and was able to remove the `SqlHandler` entirely and a lot of the async code I introduced previously

ogenev (approved):
LGTM :rocket:. Haven't thought deeply about some eventual corner cases tests.

On a side note, I think it will be useful if we refactor the `ContentId` struct in following PR and encapsulate more logic into this type because right now we are passing around mostly `[u8; 32]` which I think doesn't help with code readability.

mrferris (changes requested):
Look great! Exactly what I had in mind as far as integration of `PortalStorage` goes.

I think the only thing that needs to be done (other than fix the most recent merge conflict) is to merge #233 and then account for those changes here. small nit: There are a few `unwrap`s that didn't appear safe in here, my understanding is that there won't be once the changes introduced by #233 are integrated.

I think the changes once #233 should just be to change `PortalStorage`'s `store` and `get` method's signatures to take a `key: &impl OverlayContentKey' parameter, which is nice.

The discussion on behavior around database failures brings up a great point, and is yet another excellent reason for us to move to only using one database. In the meantime, I think that the revert-if-meta-db-fails logic is sufficient to have this merged, because even if the RocksDB ops fail, we at least having the error messages to know that that happened.

PR #233, opened 2022-01-27
Refactor overlay content key

jacobkaufmann:
Define content key trait in `trin-core` and move content key implementations into respective crates.

Define `OverlayContentKey` trait, with a single method to derive a content ID. At the most fundamental level, all that we require from a content key is to derive a content ID from it.

Add `Into<Vec<u8>>` and `TryFrom<Vec<u8>>` as super traits. We do not make any assumptions about serialization format, and we leave that up to the implementation. This is why we do not derive the SSZ encode/decode traits here. Deserialization may fail. Serialization cannot fail.

Make `Overlay` and `OverlayService` generic over the content key type. We must add a `PhantomData` here because neither struct holds any data of that type.

Test based on content key test vectors defined in the [portal network specifications](https://github.com/ethereum/portal-network-specs).


njgheorghita:
Looks pretty cool, so far. Just gave it a quick browse but will do a more thorough review tomorrow. A more general question, how do you foresee the distance function fitting into this setup? Will it rely on a similar `OverlayDistance` trait?

jacobkaufmann:
@njgheorghita The implementation I'm fiddling with is very similar, but it does not define a new trait and instead takes a `Fn(&[u8; 32], &[u8; 32]) -> U256` implementer.

I may opt to use a trait object like `Box<dyn Fn>` instead, but I'm still investigating those trade-offs.

trin-history/src/content_key.rs
```diff
@@ -0,0 +1,148 @@
...
+
+#[allow(clippy::from_over_into)]
+impl Into<Vec<u8>> for HistoryContentKey {
+    fn into(self) -> Vec<u8> {
+        self.as_ssz_bytes()
+    }
+}
+
+impl TryFrom<Vec<u8>> for HistoryContentKey {
+    type Error = &'static str;
```
njgheorghita, line 53: I feel like we've been trying to move towards custom error types in this codebase. Using `String` above for `MockContentKey` in the tests seems perfectly reasonable, but is there a reason you didn't go for a custom error here?

trin-state/src/content_key.rs
```diff
@@ -0,0 +1,314 @@
...
+
+#[allow(clippy::from_over_into)]
+impl Into<Vec<u8>> for StateContentKey {
+    fn into(self) -> Vec<u8> {
+        self.as_ssz_bytes()
+    }
+}
+
+impl TryFrom<Vec<u8>> for StateContentKey {
+    type Error = &'static str;
```
njgheorghita, line 92: Same comment about error types here.

njgheorghita (approved):
Nice, looks good to me. Seems like a definite improvement and cleaner way forward than the matching by `protocol` we were doing. I have a feeling that we will also benefit from implementing the `OverlayContentKey` on `PortalStorage` - but I'd need to fiddle around with it for a bit.

trin-history/src/content_key.rs
jacobkaufmann: So in the trait definition of `OverlayContentKey`, I didn't specify a bound on the `TryFrom::Error` associated type for flexibility, but looking at things again, that's not so useful, and we might want some kind of enum like `ContentKeyDecodeError` to provide more useful information. That, or we should at least pass through the `ssz::DecodeError`.

carver (approved):
Any PR that brings improvement to the code and reduces line count is ⭐ -- I also like how `content_keys.rs` got split up into more context-relevant locations. 👍🏻 

👍🏻 for more tests

I raised some issues, but don't need to re-review them after you address them, unless you want me to.

trin-history/src/content_key.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// Hash of the block.
+    block_hash: [u8; 32],
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<Vec<u8>> for HistoryContentKey {
+    fn into(self) -> Vec<u8> {
+        self.as_ssz_bytes()
+    }
+}
```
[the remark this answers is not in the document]
jacobkaufmann: In order to implement what you've suggested, we would have to use the [newtype pattern](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-the-newtype-pattern-to-implement-external-traits-on-external-types) to bypass the orphan rule. In this case I prefer the local clippy warning suppression over that indirection.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -545,30 +542,15 @@ impl OverlayService {
...
-            _ => {
-                warn!(
-                    "Attempt to find undefined content for {:?} protocol",
-                    self.protocol
-                );
-                return Err(OverlayRequestError::InvalidRequest(format!(
-                    "Content undefined for {:?} protocol",
-                    self.protocol,
-                )));
+        let content_id = match (K::try_from)(request.content_key) {
```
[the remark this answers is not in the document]
mrferris: sigp's discv5 implementation uses the standard of a descriptive name with a `T` (for type) in front of it, e.g.: `TContentKey`. I think that this would improve the readability here.

mrferris (approved):
This looks great. 

A thought I had on the discussion about distance function is: I wonder if we want to add a `distance_to(node: NodeId)` method to the `OverlayContentKey` trait. It might be nice to have both necessary functions abstracted behind this trait. I may also not be taking into account certain use cases, though.

PR #235, opened 2022-01-31
Rename refactoring and add comments to uTP

ogenev:
This is the first PR of a series of PRs to refactor the uTP implementation and fix some of the uTP bugs.

This PR includes a simple renaming of some of the variables, methods and adding comments to improve code readability and debugging. This PR doesn't change any functionality.

It is using the uTP [library](https://github.com/meqif/rust-utp) implemented in Rust for reference.

### DIFF
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -16,6 +16,9 @@ use crate::portalnet::types::messages::ProtocolId;
 use crate::utp::packets::{Packet, PacketHeader, PacketType, HEADER_SIZE};
 use crate::utp::trin_helpers::{UtpMessageId, UtpStreamState};
 
+// Maximum time (in microseconds) to wait for incoming packets when the send window is full
+const PRE_SEND_TIMEOUT: u32 = 500_000;
+
 pub const MAX_DISCV5_PACKET_SIZE: usize = 1280;
 pub const MIN_PACKET_SIZE: usize = 150;
 const MIN_DISCV5_PACKET_SIZE: usize = 63;
@@ -24,7 +27,8 @@ const CCONTROL_TARGET: usize = 100 * 1000;
 const MAX_CWND_INCREASE_BYTES_PER_RTT: usize = 3000;
 const MIN_WINDOW_SIZE: usize = 10;
 
-fn get_time() -> u32 {
+/// Return current time in microseconds since the UNIX epoch.
+fn now_microseconds() -> u32 {
     (SystemTime::now()
         .duration_since(UNIX_EPOCH)
         .unwrap()
@@ -33,7 +37,7 @@ fn get_time() -> u32 {
 }
 
 fn get_time_diff(response_time: u32) -> u32 {
-    let current_time = get_time();
+    let current_time = now_microseconds();
 
     if current_time > response_time {
         current_time - response_time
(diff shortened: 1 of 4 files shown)


trin-core/src/utp/stream.rs
```diff
@@ -159,145 +164,227 @@ impl UtpListener {
...
             MAX_DISCV5_PACKET_SIZE,
             min(self.max_window as usize, self.remote_wnd_size as usize),
         );
-        let now = get_time();
+        let now = now_microseconds();
 
-        // Wait for rate control, but don't wait forever
-        while self.cur_window + packet.0.len() as u32 > max_send as u32 && get_time() - now < 500000
+        // Wait until enough in-flight packets are acknowledged for rate control purposes, but don't
+        // wait more than 500 ms (PRE_SEND_TIMEOUT) before sending the packetr
```
njgheorghita, line 336: `packetr` -> `packet`

njgheorghita (approved):
Looks good to me! Didn't do an in-depth review of the functionality since you said it's mostly aesthetic changes here. `now_microseconds` does sound slightly awkward to me, but it does seem to make sense in the context of how it's used.

carver:
You mentioned in chat that you found a number of issues by transferring data over uTP from trin to trin. That sounds like an important test to add to CI to prevent regressions. Doesn't necessarily have to be in included in this PR, but I'd love to hear your plans for testing.

KolbyML (approved):
Looks good 👍

ogenev:
> You mentioned in chat that you found a number of issues by transferring data over uTP from trin to trin. That sounds like an important test to add to CI to prevent regressions. Doesn't necessarily have to be in included in this PR, but I'd love to hear your plans for testing.

I'm doing the test manually right now using a different branch which is kinda hacky. I'm checking the packet transfer between two nodes by sending a uTP payload and manually looking for inconsistencies (the code can be found [here](https://github.com/ethereum/trin/blob/3db8afe0a1e3e0d6d24ac9f64773a87b613dc9eb/utp-testing/src/main.rs)).

The general idea to test uTP is to use a uTP testing framework. The testing framework will look something similar to what Fluffy is doing [here](https://github.com/status-im/nimbus-eth1/pull/931/files). It is spinning two uTP implementations (test_apps) in docker containers and a network simulator ([ns-3](https://www.nsnam.org/) bridges the connection. 
This approach comes from a QUIC testing framework, more info about it can be found [here ](https://github.com/marten-seemann/quic-network-simulator )and [here](https://github.com/marten-seemann/quic-interop-runner). This way, we can test the uTP transfer on different network conditions including delays, drop rate, corrupted packets and etc.

To make this possible, we need first to do the following:

1. Fix all obvious issues in the current implementation and add more unit tests (this requires refactoring).
2. Add receive(), close(), flush() methods to our uTP implementation and make it more "socket like".
3. Encapsulate all uTP logic so we can test it by running only a uTP socket without running a whole trin node as a black box.
4. Integrate uTP with the testing framework mentioned above and add it to CI

So the first stage will include a few PRs (this is the first), mainly refactoring and adding additional unit tests.

carver:
> The general idea to test uTP is to use a uTP testing framework. The testing framework will look something similar to what Fluffy is doing [here](https://github.com/status-im/nimbus-eth1/pull/931/files). It is spinning two uTP implementations (test_apps) in docker containers and a network simulator ([ns-3](https://www.nsnam.org/)) bridges the connection. This approach comes from a QUIC testing framework, more info about it can be found [here ](https://github.com/marten-seemann/quic-network-simulator)and [here](https://github.com/marten-seemann/quic-interop-runner). 
> 
> So the first stage will include a few PRs (this is the first), mainly refactoring and adding additional unit tests.

Awesome, thanks for the summary. This looks great!

PR #236, opened 2022-02-02
Refactor uTP packet header

ogenev:
The goal of this PR is to refactor the way we are accessing the packet header. This accomplishes a few things:

- When setting the header fields, there is no need to pass the header around, we can set them on the packet, for example as `packet.set_seq_nr()`

- This refactoring allows us to set the packet header timestamp fields right before sending a packet without adding boilerplate code and it is fixing the issue where a packet timestamp was not sent when we resend a packet.

- Improves code readability

Uses [rust-utp](https://github.com/meqif/rust-utp) as referance.

### DIFF
--- a/trin-core/src/utp/packets.rs
+++ b/trin-core/src/utp/packets.rs
@@ -3,13 +3,37 @@ use std::fmt;
 pub const HEADER_SIZE: usize = 20;
 pub const VERSION: u8 = 1;
 
+macro_rules! u8_to_unsigned_be {
+    ($src:ident, $start:expr, $end:expr, $t:ty) => ({
+        (0 .. $end - $start + 1).rev().fold(0, |acc, i| acc | $src[$start+i] as $t << (i * 8))
+    })
+}
+
+macro_rules! make_getter {
+    ($name:ident, $t:ty, $m:ident) => {
+        pub fn $name(&self) -> $t {
+            let header = unsafe { &*(self.0.as_ptr() as *const PacketHeader) };
+            $m::from_be(header.$name)
+        }
+    };
+}
+
+macro_rules! make_setter {
+    ($fn_name:ident, $field:ident, $t: ty) => {
+        pub fn $fn_name(&mut self, new: $t) {
+            let mut header = unsafe { &mut *(self.0.as_mut_ptr() as *mut PacketHeader) };
+            header.$field = new.to_be();
+        }
+    };
+}
+
 #[derive(PartialEq, Debug)]
 pub enum PacketType {
-    Data,
-    Fin,
-    State,
-    Reset,
-    Syn,
+    Data,  // packet carries a data payload
+    Fin,   // signals the end of a connection
+    State, // signals acknowledgment of a packet
+    Reset, // forcibly terminates a connection
+    Syn,   // initiates a new connection with a peer
 }
 
 impl TryFrom<u8> for PacketType {
@@ -39,26 +63,7 @@ impl From<PacketType> for u8 {
     }
 }
 
-impl<'a> TryFrom<&'a [u8]> for PacketHeader {
-    type Error = &'static str;
-
-    fn try_from(value: &[u8]) -> Result<Self, Self::Error> {
-        if value.len() < HEADER_SIZE {
-            return Err("invalid header size");
-        }
-
-        if value[0] & 0xf != VERSION {
-            return Err("invalid packet version");
-        }
-
-        if let Err(e) = PacketType::try_from(value[0] >> 4) {
-            return Err(e);
-        }
-
-        Ok(PacketHeader::decode(value))
-    }
-}
-
+/// Validate correctness of packet extensions, if any, in byte slice
 fn check_extensions(data: &[u8]) -> Result<(), &'static str> {
     if data.len() < HEADER_SIZE {
         return Err("invalid header size");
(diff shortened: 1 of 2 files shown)


trin-core/src/utp/packets.rs
```diff
@@ -3,13 +3,37 @@ use std::fmt;
...
+macro_rules! u8_to_unsigned_be {
+    ($src:ident, $start:expr, $end:expr, $t:ty) => ({
+        (0 .. $end - $start + 1).rev().fold(0, |acc, i| acc | $src[$start+i] as $t << (i * 8))
+    })
+}
+
+macro_rules! make_getter {
+    ($name:ident, $t:ty, $m:ident) => {
+        pub fn $name(&self) -> $t {
+            let header = unsafe { &*(self.0.as_ptr() as *const PacketHeader) };
```
carver, line 15: I would love to see comments every time we use `unsafe`: what alternatives were considered, why the alternatives were unusable or how much less efficient they would be, etc. Otherwise, it becomes a time-sink, when code readers might wonder if there's a subtle bug lurking in there.

trin-core/src/utp/packets.rs
```diff
@@ -39,26 +63,7 @@ impl From<PacketType> for u8 {
...
-
-        if let Err(e) = PacketType::try_from(value[0] >> 4) {
-            return Err(e);
-        }
-
-        Ok(PacketHeader::decode(value))
-    }
-}
-
+/// Validate correctness of packet extensions, if any, in byte slice
```
carver, line 66: 👍🏻 for docstrings

trin-core/src/utp/packets.rs
```diff
@@ -113,90 +99,82 @@ pub struct PacketHeader {
...
-    }
+impl<'a> TryFrom<&'a [u8]> for PacketHeader {
+    // TODO: Refactor this to use anyhow crate
+    type Error = &'static str;
+    /// Reads a byte buffer and returns the corresponding packet header.
+    /// It assumes the fields are in network (big-endian) byte order,
+    /// preserving it.
+    fn try_from(buf: &[u8]) -> Result<Self, Self::Error> {
+        // Check length
+        if buf.len() < HEADER_SIZE {
```
carver, line 132: I don't have any *problem* with the simple comments like this one, but I don't think it adds much when the code is this easy to read.

trin-core/src/utp/packets.rs
```diff
@@ -113,90 +99,82 @@ pub struct PacketHeader {
...
-
-    pub fn set_connection_id(&mut self, connection_id: u16) {
-        self.connection_id = connection_id;
-    }
+impl<'a> TryFrom<&'a [u8]> for PacketHeader {
+    // TODO: Refactor this to use anyhow crate
+    type Error = &'static str;
+    /// Reads a byte buffer and returns the corresponding packet header.
+    /// It assumes the fields are in network (big-endian) byte order,
+    /// preserving it.
```
carver, line 129: Yay for docstrings 👍🏻

trin-core/src/utp/packets.rs
```diff
@@ -113,90 +99,82 @@ pub struct PacketHeader {
...
+        Ok(PacketHeader {
+            type_ver: buf[0],
+            extension: buf[1],
+            connection_id: u8_to_unsigned_be!(buf, 2, 3, u16),
+            timestamp_microseconds: u8_to_unsigned_be!(buf, 4, 7, u32),
+            timestamp_difference_microseconds: u8_to_unsigned_be!(buf, 8, 11, u32),
+            wnd_size: u8_to_unsigned_be!(buf, 12, 15, u32),
+            seq_nr: u8_to_unsigned_be!(buf, 16, 17, u16),
+            ack_nr: u8_to_unsigned_be!(buf, 18, 19, u16),
+        })
```
carver, line 155: Huh, interesting, why remove the `decode()` in favor of in-lining here? I don't really have a preference, but I my first instinct was that having a `PacketHeader::decode` seemed like a reasonable choice.

trin-core/src/utp/packets.rs
```diff
@@ -215,67 +193,87 @@ impl AsRef<[u8]> for Packet {
...
+    }
+}
+
 impl Packet {
-    pub fn new(mut packet_header: PacketHeader) -> Packet {
-        let mut vec = Vec::with_capacity(HEADER_SIZE);
-        vec.append(&mut packet_header.encode());
-        Packet(vec)
+    pub fn new() -> Packet {
+        Packet(PacketHeader::default().as_ref().to_owned())
```
carver, line 204: This is a surprising pattern that maybe I just haven't run into before. Why build a new object, take a reference, and then convert it back to owned (presumably incurring the cost of copying along the way)?

trin-core/src/utp/packets.rs
```diff
@@ -84,26 +89,7 @@ fn check_extensions(data: &[u8]) -> Result<(), &'static str> {
...
-}
-
-impl Default for PacketHeader {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-#[derive(Debug)]
+#[repr(C)]
```
carver, line 92: Maybe a one-liner explaining why we want/need this? I'm seeing articles about it being useful for FFI, but I don't know  about plans to use that.

carver (approved):
Cool, looking good! I definitely have some questions, but I'll chalk it up to just using parts of rust I'm not familiar with (and maybe some requests for comments here and there).

I suspect a lot of it derives from the `repr(C)` which I've never used before. But yeah, I'd love it if you could help me understand, either with an explanation in-code or in the PR, whichever seems appropriate.

trin-core/src/utp/packets.rs
ogenev: I think we are creating a simple `PacketHeader` object here and I'm not sure if using `decode()` makes sense anymore, because we are not `decoding` anything anymore, we are creating a new type instance.

trin-core/src/utp/packets.rs
ogenev: In Rust, `unsafe` serves mainly for two purposes: calling unsafe functions and dereferencing raw pointers. 
In this case `*const` and `self.0.as_ptr()` are our raw pointers which are analogous to &T, except that it doesn’t have a lifetime and is not subject to the same validity rule as its & counterpart.

Pointer cast seems a classic Rust pattern and is useful when we want to interpret a sequence
of bytes as #[repr(C)] structs of these. We need this because we don't know the byte length of the packet (self.0) because of the unknown payload, we only know that the first X bytes are the header.

trin-core/src/utp/packets.rs
ogenev: We have `AsRef` trait implemented for `PacketHeader` here:

```rust
impl AsRef<[u8]> for PacketHeader {
    /// Returns the packet header as a slice of bytes.
    fn as_ref(&self) -> &[u8] {
        unsafe { &*(self as *const PacketHeader as *const [u8; HEADER_SIZE]) }
    }
}
```

which is returning a slice of bytes and one way to convert from a slice of bytes to `Vec<u8>` is `to_owned()`. It is similar to converting from `&str` to `String`.

trin-core/src/utp/packets.rs
ogenev: Here is a more detailed explanation of [repr(C)](https://doc.rust-lang.org/reference/type-layout.html#the-c-representation). 
Default Rust representation doesn't guarantee the data layout and because we want to cast raw pointers to `PacketHeader`, we need consistent data layout representation of the struct.

Issue #232, opened 2022-01-25
Testing uTP protocol


victor-wei126:
See this: https://github.com/ethereum/portal-network-specs/issues/122

ogenev:
I'm currently working on this issue and I'm researching various alternatives.

KonradStaniec:
one minor comment, I would be carful with this https://github.com/meqif/rust-utp/, as I saw that it do some weird things like for example: https://github.com/meqif/rust-utp/blob/main/src/socket.rs#L478. 

Imo it pretty unusual for transport level protocol to do something like sending 3 consecutive duplicated acks and treat it like some kinda request, not to mention in amplifies network traffic unnecessary.

I am not saying this library in necessary all bad, but it probably requires thorough review.

ogenev:
> Imo it pretty unusual for transport level protocol to do something like sending 3 consecutive duplicated acks and treat it like some kinda request, not to mention in amplifies network traffic unnecessary.

Isn't this part of the "Fast retransmit" modification in [TCP](https://www.isi.edu/nsnam/DIRECTED_RESEARCH/DR_HYUNAH/D-Research/fast-retransmit.html)?

KonradStaniec:
yep fast retransmit is pretty standard thing, but the case is it is not triggered intentionally by sending 3 duplicated acks, but it is triggered when 3 duplicated acks are received due to network conditions like packet loss or delays.

ogenev:
> yep fast retransmit is pretty standard thing, but the case is it is not triggered intentionally by sending 3 duplicated acks, but it is triggered when 3 duplicated acks are received due to network conditions like packet loss or delays.

Hm, you are right, I think this is some kind of a hack to deal with packet loss. This is from the library readme:

_"The current implementation is somewhat incomplete, lacking a complete implementation of congestion control. However, it does support packet loss detection (except by timeout)..."_

So I guess we should implement the packet loss detection by timeout by ourselves, does Fluffy have this implemented already?

KonradStaniec:
Yep, we follow reference implementation in that regards. When we send first packet we start a timer based on current calculated timeout, if the timer runs out (packet wont be acket in rto time) we resend oldest packet.  The whole logic starts here https://github.com/bittorrent/libutp/blob/master/utp_internal.cpp#L1147.

PR #238, opened 2022-02-08
Refactor uTP timestamps and congestion control

ogenev:
This PR is part of phase 1 defined in https://github.com/ethereum/trin/issues/232#issuecomment-1033862510.

 It tries to accomplish the following:

- add Timestamp and Delay structs to wrap timestamp and timestamp difference values. 
- improve LEDBAT congestion algorithm (the core of uTP) to match [rust-utp](https://github.com/meqif/rust-utp/blob/aad6257176896313cb2f8ec40fa6c80e9cf9b3ea/src/socket.rs#L971) implementation 1:1

Those changes bring us one step closer to having 1:1 core uTP implementation with `rust-utp`.

Unfortunately, there is still no way to write adequate unit tests for the core uTP logic, because critical parts of the implementation are missing, but after a few more refactoring PRs, this will be possible.


trin-core/src/utp/util.rs
```diff
@@ -0,0 +1,70 @@
...
+/// factor `alpha` between 0 and 1. A higher `alpha` discounts older observations faster.
+pub fn ewma<'a, T, I>(mut samples: I, alpha: f64) -> f64
+where
+    T: ToPrimitive + 'a,
+    I: Iterator<Item = &'a T>,
+{
+    let first = samples.next().map_or(0.0, |v| v.to_f64().unwrap());
+    samples
+        .map(|v| v.to_f64().unwrap())
+        .fold(first, |avg, sample| alpha * sample + (1.0 - alpha) * avg)
```
njgheorghita, line 14: Neat!

Issue #214, opened 2022-01-07
Error handling: Consider switching from `thiserror` to `anyhow` crate


ogenev:
Currently, we use [thiserror](https://docs.rs/thiserror/latest/thiserror/) crate to create different types of errors and return them inside our methods. It seems this way of handling errors adds unnecessary overhead (more boilerplate code) and makes sense when building libraries.

It looks like the [anyhow](https://docs.rs/anyhow/1.0.52/anyhow/) crate provides an easy and idiomatic way of handling errors in rust applications.

This blog post describes well library vs application error handling in Rust:
https://nick.groenen.me/posts/rust-error-handling/#libraries-versus-applications

ogenev:
> 'm not remembering exactly, but @ogenev did you once mention that there is some reason we can't entirely switch all our errors over from `thiserror` to `anyhow`?

I had in mind mainly `OverlayRequestError` in `overlay_service` because we are passing some errors between handlers/channels and I think refactoring them will be a more involving process.

Other than this, I guess we can easily switch to `anyhow` in other parts of the code where the error handling is more encapsulated.

njgheorghita:
Personally, I'd agree with the switch over to `anyhow` but only in the case that it can fully replace `thiserror`. Using two different error handling libraries seems fuzzy and confusing where one should use which one. Though, I guess we're currently already using both of them...

njgheorghita:
Closing - `CONTRIBUTING.md` has been updated with new error guidelines

PR #239, opened 2022-02-09
Experimental anyhow integration

njgheorghita:
This is an experimental pr to see what using `anyhow` would look like. We've been talking about this in #214, but I'm not able to truly wrap my head around it without giving it a go. To be fair, this is my first time using the library, and It's entirely possible that my integration is not the cleanest / idiomatic way to go about using this library. If so, please let me know what I'm missing b/c that for sure influences my opinion...

The pros...
- It's cleaner and neater and shorter
- I'm sure it would help boost my speed in rust, not having to worry about custom error types all the time
- The `context` method seems pretty useful for adding detail to messages

The cons...
- It makes me feel nervous... all these untyped errors. smh what has rust done to me
- It seems far too easy to write insufficient error messages. At minimum, using `thiserror` enforces a well thought out message and heirarchy of error types.

Ultimately....
[This blog](https://nick.groenen.me/posts/rust-error-handling/#libraries-versus-applications) inspired the issue. The key advice being `"use thiserror for libraries and anyhow for applications"`. After playing with `anyhow` I'm starting to come to the belief that `trin` isn't really an application. It's closer to a collection of libraries acting as a single application. A well-thought out, consistent, typed error system is something we would benefit from. Especially dealing with bugs on various networks. 

Maybe we use `thiserror` in certain parts of the codebase and `anyhow` in others? Imo, one error library is better than two.

I'm certainly open to having my mind changed. Maybe I've missed out some of the more useful features of `anyhow`. Since this decision touches the entire codebase, I'm curious to hear what everyone thinks....

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -15,6 +15,7 @@ use crate::portalnet::types::messages::{
 
 use crate::utp::stream::{ConnectionKey, UtpListener};
 use crate::utp::trin_helpers::{UtpAccept, UtpMessage, UtpMessageId, UtpStreamState};
+use anyhow::anyhow;
 use discv5::{
     enr::NodeId,
     kbucket::{Filter, KBucketsTable},
@@ -28,8 +29,6 @@ use tokio::sync::mpsc;
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::RwLock as RwLockT;
 use tracing::{debug, warn};
-
-pub use super::overlay_service::OverlayRequestError;
 
 /// Configuration parameters for the overlay network.
 #[derive(Clone)]
@@ -137,16 +136,22 @@ impl OverlayProtocol {
     pub async fn process_one_request(
         &self,
         talk_request: &TalkRequest,
-    ) -> Result<Response, OverlayRequestError> {
+    ) -> anyhow::Result<Response> {
         let request = match Message::from_bytes(talk_request.body()) {
             Ok(Message::Request(request)) => request,
             Ok(_) => {
-                return Err(OverlayRequestError::InvalidRequest(format!(
+                return Err(anyhow!(
                     "Message not recognized request type: {:?}",
                     talk_request.body()
-                )))
+                ))
             }
-            Err(_) => return Err(OverlayRequestError::DecodeError),
+            Err(msg) => {
+                return Err(anyhow!(
+                    "Error decoding request on {:?} overlay network: {:?}",
+                    self.protocol,
+                    msg
+                ))
+            }
         };
         let direction = RequestDirection::Incoming {
             id: talk_request.id().clone(),
@@ -176,7 +181,7 @@ impl OverlayProtocol {
     }
 
     /// Sends a `Ping` request to `enr`.
-    pub async fn send_ping(&self, enr: Enr) -> Result<Pong, OverlayRequestError> {
+    pub async fn send_ping(&self, enr: Enr) -> anyhow::Result<Pong> {
         // Construct the request.
         let enr_seq = self.discovery.local_enr().seq();
         let data_radius = self.data_radius();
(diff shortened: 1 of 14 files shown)


ogenev:
I'm less or more biased towards `anyhow` (because I proposed this :rofl: ) but let me think about this again.

At this stage of the project, I feel iteration speed is more valuable than a well-defined error type system because there are still a lot of unknown implementation details.

It is true that we can consider most of the workspace creates libraries, but as long as we are the only ones that have ownership over those libraries, I think it is pretty straightforward to manage our `anyhow` error handling in a way to helps us with debugging and tracing.
If there is a library that is used by third parties, a good error type structure with `thiserror` makes sense because the application code is responsible for handling those errors and it will need error details.

I'm not concerned about using two error handling libraries, as long as there is only one in any workspace crate. But maybe we are still in the process to define what sub-crates we want to have, for example probably `discovery`, `overlay`, `utp`, and  `json-rpc` should be in separate crates...

trin-core/src/portalnet/overlay.rs
```diff
@@ -137,16 +136,22 @@ impl OverlayProtocol {
     pub async fn process_one_request(
         &self,
         talk_request: &TalkRequest,
-    ) -> Result<Response, OverlayRequestError> {
+    ) -> anyhow::Result<Response> {
```
mrferris, line 139: What do you think of putting at the top of the file:`type Result = anyhow::Result` so that all of these lines can just be: 
```suggestion
    ) -> Result<Response> {
```

Might improve readability a bit.

mrferris (approved):
This is great! Yea, seeing it used definitely shows how much boilerplate/time it saves when all we'll do with the errors is log them out anyway. Great suggestion @ogenev.

I think that using both is fine, I think for example that using `thiserror` in `PortalStorage` is still preferable, because with some of the errors returned we might want to re-try the operation instead of only logging it. I like the idea of having storage, utp, etc broken down into crates with defined error models, while the higher level stuff just emits strings for logs via anywhow.

njgheorghita:
Thanks for the thoughts. I'm shifting towards being ok with using both `thiserror` and `anyhow` together. Prs like #242 really help prove the point that sometimes fast > precise when it comes to error handling. But, there are still certain situations where typed errors are fairly crucial (aka in `PortalStorage` like @mrferris mentioned or in the [overlay networks](https://github.com/ethereum/trin/blob/master/trin-history/src/network.rs#L58). 

I guess my rule of thumb would be something like: Use `anyhow` unless there is a concrete benefit to using typed errors. (aka we should never downcast an `anyhow` error to a string and match on that string). 

The next open question to me seems like: how do we want to handle the migration? Should we merge this pr? This pr only touches part of the codebase... Should I extend it to the entire codebase where the rule of thumb applies? Should we close this pr and just gradually migrate over to `anyhow` where applicable in the following prs?

trin-core/src/portalnet/overlay.rs
njgheorghita: I'm not so keen on this. Especially if we settle on two different error libraries. Yeah, it's a definite improvement in readability, but...
1 - it might confuse people new to codebase, familiar with Rust but not `anyhow`
2 - keeping track of which error handling library is used where in the codebase, I'd get frustrated with having to scroll to the top of each file to see if this import is there or not.

trin-core/src/portalnet/types/uint.rs
```diff
@@ -139,17 +132,18 @@ mod test {
...
+    #[case(U512([1,2,3,4,0,34,0,0]))]
+    #[case(U512([1,2,3,4,0,0,234,0]))]
+    #[case(U512([1,2,3,4,0,0,0,349]))]
+    #[case(U512([u64::MAX,u64::MAX,u64::MAX,u64::MAX,1,0,0,0]))]
+    #[case(U512([0,0,0,0,1,0,0,0]))]
+    fn test_u512_to_u256_failure(#[case] input: U512) {
+        let result: anyhow::Result<U256> = input.try_into();
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            "Failed to convert U512 to U256 due to overflow"
```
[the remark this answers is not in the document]
njgheorghita: I'm a big fan of matching error messages in tests, that way you're guaranteed that the test failed for the exact reason it's was designed to. If somebody wants to change the error message (which shouldn't happen too frequently) then the test will fail with a fairly explicit reason and the fix will be obvious.

ogenev (approved):
> The next open question to me seems like: how do we want to handle the migration? Should we merge this pr? 

I'm ok with merging this and gradually migrating our errors to anyhow where it is possible in later PRs. I think the early stage of development is the best time to experiment with such things :)

trin-state/src/network.rs
```diff
@@ -64,41 +64,13 @@ impl StateNetwork {
                     debug!("Successfully bonded with {}", enr);
                     continue;
                 }
-                Err(OverlayRequestError::ChannelFailure(error)) => {
-                    debug!("Channel failure sending ping: {}", error);
+                // Tbh I'm a bit stumped on how to handle this area, so this is just a temporary
+                // solution to get the compiler passing. Will revisit if we decide to continue with
+                // anyhow
+                Err(msg) => {
```
ogenev, line 70: Yeah, I was stumped also here when trying prototyping it and that's why I assumed that getting rid of `OverlayRequestError` may not be so trivial.

njgheorghita:
Sounds good. I reverted all of the `OverlayRequestError` -> `anyhow` conversions here, since we couldn't make the transformation 100%, picking and choosing some arbitrary error border (not to mention somewhat clunky conversion b/w error types) for where to use one crate or the other felt messy. I also updated the contributing guidelines to contain both error libraries.

PR #241, opened 2022-02-10
Setup initial release process

njgheorghita:
- Add `newsfragments` to help aggregate and compile release notes
- Add `Makefile` with some useful commands
- Add pull request template
- Move docs to `docs/` folder, plucked some of the docs out of the `README` into `Getting Started`
- Add release steps to docs

This pr contains some of the utils / docs for a formal release process. I plan on cutting an `0.2.0-alpha` release once this gets merged. So far, the dry runs using `cargo publish` have worked out alright, we'll see how that continues when the publish command is actually executed after this gets merged. 

In terms of version release numbers, it seems appropriate to me to continue with a successive run of alpha releases. i.e.
- `0.2.0-alpha`
- `0.3.0-alpha`
- `0.4.0-alpha`

Until we hit our first stable minor release, and then simply drop the alpha tag. I'm definitely open to other ideas on how to structure this.


ogenev (approved):
Looks good, I'm happy with the proposed versioning. Once we release the first stable 1.0.0 version, I guess we can switch to [semver](https://semver.org/) then.

mrferris (approved):
Looks great to me. Only small thing is that I agree that the release notes categories suggested by @ogenev are more in-line with what I've seen before.

PR #243, opened 2022-02-15
Refactor uTP packet handler and add tests

ogenev:
What does this PR do:

- refactors the uTP packet handler by adding two more socket states: `FinSent` and `ResetReceived`. This allows us to cover more corner cases when receiving uTP packets depending on the current socket state.
- adds unit tests for the core uTP logic :tada: 

This is the last PR of Phase 1 in https://github.com/ethereum/trin/issues/232#issuecomment-1033862510. :heavy_check_mark:

### DIFF
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -528,6 +528,7 @@ impl UtpSocket {
         debug!("self.their_delay: {}", self.their_delay);
 
         match (self.state, packet.get_type()) {
+            // New connection, when we receive SYN packet, respond with STATE packet
             (SocketState::Uninitialized, PacketType::Syn) => {
                 self.connected_to = src;
                 self.ack_nr = packet.seq_nr();
@@ -539,7 +540,10 @@ impl UtpSocket {
 
                 Ok(Some(self.prepare_reply(packet, PacketType::State)))
             }
+            // When connection is already initialised and we receive SYN packet,
+            // we want to forcibly terminate the connection
             (_, PacketType::Syn) => Ok(Some(self.prepare_reply(packet, PacketType::Reset))),
+            // When SYN is send and we receive STATE, do not reply
             (SocketState::SynSent, PacketType::State) => {
                 self.connected_to = src;
                 self.ack_nr = packet.seq_nr();
@@ -549,18 +553,23 @@ impl UtpSocket {
                 self.last_acked_timestamp = now_microseconds();
                 Ok(None)
             }
+            // Only STATE packet is expected when SYN is sent
             (SocketState::SynSent, _) => Err(anyhow!("The remote peer sent an invalid reply")),
+            // Handle data packet if socket state is `Connected` or `FinSent` and packet type is DATA
             (SocketState::Connected, PacketType::Data)
             | (SocketState::FinSent, PacketType::Data) => Ok(self.handle_data_packet(packet)),
+            // Handle state packet if socket state is `Connected` and packet type is STATE
             (SocketState::Connected, PacketType::State) => {
                 self.handle_state_packet(packet);
                 Ok(None)
             }
+            // Handle FIN packet. Check if all send packets are acknowledged.
             (SocketState::Connected, PacketType::Fin) | (SocketState::FinSent, PacketType::Fin) => {
                 if packet.ack_nr() < self.seq_nr {
                     debug!("FIN received but there are missing acknowledgements for sent packets");
                 }
                 let mut reply = self.prepare_reply(packet, PacketType::State);
+
                 if packet.seq_nr().wrapping_sub(self.ack_nr) > 1 {
                     debug!(
                         "current ack_nr ({}) is behind received packet seq_nr ({})",
@@ -580,6 +589,7 @@ impl UtpSocket {
                 self.state = SocketState::Closed;
                 Ok(Some(reply))
             }
+            // Confirm with STATE packet when socket state is `Closed` and we receive FIN packet
             (SocketState::Closed, PacketType::Fin) => {
                 Ok(Some(self.prepare_reply(packet, PacketType::State)))
             }
@@ -591,6 +601,7 @@ impl UtpSocket {
                 }
                 Ok(None)
             }
+            // Reset connection when receiving RESET packet
             (_, PacketType::Reset) => {
                 self.state = SocketState::ResetReceived;
                 Err(anyhow!("Connection reset by remote peer"))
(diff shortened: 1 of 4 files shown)


trin-core/src/utp/stream.rs
```diff
@@ -488,179 +496,249 @@ impl UtpSocket {
...
+    }
 
-        let mut packet_reply = Packet::new();
-        packet_reply.set_type(PacketType::State);
-        packet_reply.set_connection_id(self.sender_connection_id);
-        packet_reply.set_seq_nr(self.seq_nr);
-        packet_reply.set_ack_nr(self.ack_nr);
+    fn handle_data_packet(&mut self, packet: &Packet) -> Option<Packet> {
+        // If a FIN was previously sent, reply with a FIN packet acknowledging the received packet.
+        let packet_type = if self.state == SocketState::FinSent {
```
njgheorghita, line 623: Maybe a match here would be slightly neater? 
```rust
let packet_type = match self.state {
    SocketState::FinSent => PacketType::Fin,
    _ => PacketType::State
}
```

trin-core/src/utp/stream.rs
```diff
@@ -488,179 +496,249 @@ impl UtpSocket {
...
-                && (self.ack_nr == packet_seq || self.ack_nr + 1 == packet_seq)
-            {
-                self.incoming_buffer.remove(&packet_seq);
-                self.ack_nr = packet_seq;
+            // Set SACK extension payload if the packet is not in order
+            let sack_bitfield = self.build_selective_ack();
 
-                self.recv_data_stream
-                    .append(&mut Vec::from(packet_from_buffer.payload()));
+            if !sack_bitfield.is_empty() {
```
njgheorghita, line 640: *nitpick* but I tend to prefer `is_some()` over `!...is_empty()`

trin-core/src/utp/stream.rs
```diff
@@ -488,179 +496,249 @@ impl UtpSocket {
...
+            self.resend_lost_packet(packet.ack_nr() + 1);
+        }
 
-            // Success, advance send window
-            self.advance_send_window();
+        // Packet lost, halve the congestion window
+        if packet_loss_detected {
+            debug!("packet loss detected, halving congestion window");
+            self.cwnd = max(self.cwnd / 2, MIN_CWND * MAX_DISCV5_PACKET_SIZE);
+            debug!("congestion window: {}", self.cwnd);
```
njgheorghita, line 737: I'm just curious what your thoughts are on these debug statements. My impression is that once trin is up and running in debug mode, the utp debug comments are going to dominate the majority of the output. Do you agree? Is it worth considering adding another log level just for utp debug output?

trin-core/src/utp/stream.rs
```diff
@@ -766,35 +844,337 @@ impl UtpSocket {
...
         }
     }
 
-    fn handle_syn_packet(&mut self, packet: Packet) {
-        self.receiver_connection_id = packet.connection_id() + 1;
-        self.sender_connection_id = packet.connection_id();
-        self.seq_nr = rand();
-        self.ack_nr = packet.seq_nr();
-        self.state = SocketState::SynRecv;
+    fn generate_random_enr() -> Enr {
```
njgheorghita, line 876: Is it worth pulling this out as a util fn and putting it into `trin-core/src/utils/node_id.rs`? Could you see it being useful in other places in the codebase?

trin-core/src/utp/stream.rs
```diff
@@ -766,35 +844,337 @@ impl UtpSocket {
...
+        assert_eq!(response.get_type(), PacketType::State);
 
-        self.send_packet(&mut packet_reply);
+        // Same connection id on both ends during connection establishment
+        assert_eq!(response.connection_id(), packet.connection_id());
+
+        // Response acknowledges SYN
+        assert_eq!(response.ack_nr(), packet.seq_nr());
+
+        // No payload?
```
njgheorghita, line 925: Is this a question? Is there supposed to be a payload? Something like `// Expect no payload` might be better

trin-core/src/utp/stream.rs
```diff
@@ -766,35 +844,337 @@ impl UtpSocket {
...
+        assert_eq!(response.get_type(), PacketType::Reset);
+        assert_eq!(response.ack_nr(), packet.seq_nr());
+
+        // Mark socket as closed
+        socket.state = SocketState::Closed;
+    }
+
+    #[tokio::test]
+    async fn test_unordered_packets() {
+        // Boilerplate test setup
```
njgheorghita, line 1088: Any reason not to pull out this setup into a single helper function? It seems to be used across all the tests here

njgheorghita (approved):
Lookin' good!

trin-core/src/utp/stream.rs
ogenev: In Rust, `is_some()` is a method implemented for `Option<T>`, no such a method for `Vec<T>`.

trin-core/src/utp/stream.rs
ogenev: I put it in `node_id.rs` for now with  `#[cfg(test)]` flag, it is already used in `overlay_service`.

trin-core/src/utp/stream.rs
```diff
@@ -488,179 +496,249 @@ impl UtpSocket {
...
         self.their_delay = abs_diff(now, packet.timestamp());
         debug!("self.their_delay: {}", self.their_delay);
 
-        match packet.get_type() {
-            PacketType::Data => self.handle_data_packet(packet),
-            PacketType::Fin => self.handle_finalize_packet(),
-            PacketType::State => self.handle_state_packet(packet),
-            PacketType::Reset => unreachable!("Reset should never make it here"),
-            PacketType::Syn => self.handle_syn_packet(packet),
+        match (self.state, packet.get_type()) {
```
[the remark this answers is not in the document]
ogenev: Not sure how we can improve code readability here because we have a variety of cases. For now, I added some more comments, hope this helps. Imo is not a huge PITA because you get used to it fast  :smile: , maybe we can look more into it in later PR.

trin-core/src/utp/stream.rs
ogenev: Probably we need to revisit logging and levels, not sure what is the idiomatic way to do this with the `tracing` library.

Currently, log levels are controlled on a per-module basis, so for example, if we want to run it only for the `portalnet` module we can do it with:

```sh
RUST_LOG=trin_core::portalnet=debug
```
I know this is not the best solution but I need to do more research on rust logging and tracing and come up with a strategy to support multiple log levels, or we can add it as an issue in the repo.

PR #244, opened 2022-02-16
Move uTP packet tests to packets.rs file

ogenev:
SImple PR, it moves all uTP packet tests to where they belong - packets.rs file.


njgheorghita (approved):
lgtm!

PR #248, opened 2022-02-23
Implement uTP socket recv() method and init uTP testing framework

ogenev:
### What was wrong?

To be able to write meaningful unit tests and build the uTP testing framework, we want to encapsulate the uTP module and consume the uTP socket packets.

### How was it fixed?

- Implemented uTP socket `recv()` and `close()` methods that allows us to handle a uTP packet from the discv5 socket, consume the result and gracefully close the uTP connection, so it  works now as a standard socket call, for example, we can say:

```rust
client.send_to(..);
server.recv(..);
client.close()
```

- added a simple uTP testing code just to make sure we can transfer data between two uTP implementations without any known issues.

- added test for triple ack response

### To-Do

- [x] Clean up commit history

This PR is completing Phase 2 in https://github.com/ethereum/trin/issues/232#issuecomment-1033862510.

### DIFF
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -2,21 +2,25 @@
 
 use crate::portalnet::discovery::Discovery;
 use anyhow::anyhow;
+use async_recursion::async_recursion;
 use discv5::enr::NodeId;
-use discv5::Enr;
+use discv5::{Enr, TalkRequest};
 use log::debug;
 use rand::Rng;
 use std::cmp::{max, min};
-use std::collections::{BTreeMap, HashMap, VecDeque};
+use std::collections::{HashMap, VecDeque};
 use std::convert::TryFrom;
 use std::sync::Arc;
 use tokio::sync::mpsc;
+use tokio::sync::RwLock;
+use tokio::time::timeout;
 
 use crate::portalnet::types::messages::ProtocolId;
 use crate::utp::packets::{ExtensionType, Packet, PacketType, HEADER_SIZE};
 use crate::utp::time::{now_microseconds, Delay, Timestamp};
 use crate::utp::trin_helpers::{UtpMessageId, UtpStreamState};
-use crate::utp::util::{abs_diff, ewma};
+use crate::utp::util::{abs_diff, ewma, generate_sequential_identifiers};
+use std::time::Duration;
 
 // For simplicity's sake, let us assume no packet will ever exceed the
 // Ethernet maximum transfer unit of 1500 bytes.
@@ -27,6 +31,8 @@ const MIN_CWND: u32 = 2; // minimum congestion window size
 const INIT_CWND: u32 = 2; // init congestion window size
 const MIN_CONGESTION_TIMEOUT: u64 = 500; // 500 ms
 const MAX_CONGESTION_TIMEOUT: u64 = 60_000; // one minute
+const MAX_RETRANSMISSION_RETRIES: u32 = 5; // maximum retransmission retries
+const WINDOW_SIZE: u32 = 1024 * 1024; // local receive window size
 
 // Maximum time (in microseconds) to wait for incoming packets when the send window is full
 const PRE_SEND_TIMEOUT: u32 = 500_000;
@@ -54,6 +60,7 @@ pub enum SocketState {
     ResetReceived,
 }
 
+#[derive(Clone)]
 struct DelayDifferenceSample {
     received_at: Timestamp,
     difference: Delay,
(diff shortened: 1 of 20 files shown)


trin-core/src/utp/stream.rs
```diff
@@ -111,30 +141,59 @@ impl UtpListener {
                     PacketType::Syn => {
                         if let Some(enr) = self.discovery.discv5.find_enr(node_id) {
                             // If neither of those cases happened handle this is a new request
-                            let mut conn = UtpSocket::new(Arc::clone(&self.discovery), enr.clone());
-                            let _ = conn.handle_packet(&packet, enr);
+                            let mut conn = UtpSocket::new(
+                                Arc::clone(&self.discovery),
+                                enr.clone(),
+                                self.rx.clone(),
```
njgheorghita, line 147: I was under the impression that `clone`s in Rust is not ideal, and should be used only when dealing with static data. Doesn't seem like a receive channel is a great candidate for cloning

trin-core/src/utp/stream.rs
```diff
@@ -111,30 +141,59 @@ impl UtpListener {
...
                                 },
-                                conn,
+                                conn.clone(),
                             );
                         } else {
                             debug!("Query requested an unknown ENR");
                         }
                     }
-                    _ => {
+                    // Receve DATA and FIN packets
```
njgheorghita, line 168: *nitpick* `Receive`

trin-core/src/utp/stream.rs
```diff
@@ -855,16 +984,238 @@ impl UtpSocket {
...
+        Ok(read)
+    }
+
+    /// Discards sequential, ordered packets in incoming buffer, starting from
+    /// the most recently acknowledged to the most recent, as long as there are
+    /// no missing packets. The discarded packets' payload is written to the
+    /// slice `buf`, starting in position `start`.
+    /// Returns the last written index.
+    fn flush_incoming_buffer(&mut self, buf: &mut [u8]) -> usize {
+        fn unsafe_copy(src: &[u8], dst: &mut [u8]) -> usize {
```
njgheorghita, line 1049: Maybe a note here on why unsafe code is necessary would be helpful

trin-core/src/utp/stream.rs
```diff
@@ -855,16 +984,238 @@ impl UtpSocket {
...
+
+        // Send FIN
+        self.socket
+            .send_talk_req(
+                self.connected_to.clone(),
+                ProtocolId::Utp,
+                Vec::from(packet.as_ref()),
+            )
+            .await
+            .unwrap();
```
njgheorghita, line 1169: Seems like `?` might be better here than the naked unwrap?

trin-core/src/utp/stream.rs
```diff
@@ -855,16 +984,238 @@ impl UtpSocket {
...
     use crate::utp::stream::{SocketState, UtpSocket, BUF_SIZE};
+    use discv5::Discv5Event;
+    use std::convert::TryFrom;
+    use std::str::FromStr;
     use std::sync::Arc;
+    use tokio::sync::mpsc;
+    use tokio::sync::mpsc::UnboundedSender;
+    use tokio::sync::RwLock;
+
+    macro_rules! iotry {
```
njgheorghita, line 1211: I'd suggest at least a comment here to explain what this means

trin-core/src/utp/stream.rs
```diff
@@ -1160,4 +1568,102 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_response_to_triple_ack() {
+        let mut buf = [0; BUF_SIZE];
+        let mut server = server_setup().await;
+
+        // Fits in a packet
+        const LEN: usize = 50;
+        let data = (0..LEN).map(|idx| idx as u8).collect::<Vec<u8>>();
+        let d = data.clone();
```
njgheorghita, line 1580: *nitpick* single letter variable

trin-core/src/utp/util.rs
```diff
@@ -23,6 +24,21 @@ pub fn abs_diff<T: PartialOrd + Sub<Output = U>, U>(a: T, b: T) -> U {
     }
 }
 
+/// Safely generates two sequential connection identifiers.
+///
+/// This avoids an overflow when the generated receiver identifier is the largest
+/// representable value in u16 and it is incremented to yield the corresponding sender
+/// identifier.
+pub fn generate_sequential_identifiers() -> (u16, u16) {
```
njgheorghita, line 32: Seems like a unit test with a couple cases would be worthwhile for this fn

utp-testing/src/main.rs
```diff
@@ -0,0 +1,147 @@
...
+            .listening
+            .insert(connection_id + 1, UtpMessageId::OfferAcceptStream);
+    }
+}
+
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::fmt::init();
+
+    let client_port = 9002;
```
njgheorghita, line 97: Might not be a big issue for now - but this will probably fail if those ports are not available.

utp-testing/README.md
```diff
@@ -0,0 +1,7 @@
+# utp-testing
+
+Utp testing framework (WIP)
```
njgheorghita, line 3: Was there a reason for putting this in a new crate rather than `ethportal-peertest`?

njgheorghita (approved):
Looks great! New api is cleaner. I would say the pr itself is fairly large, I'd aim for more modular prs down the road. Again, I'm mostly relying on testing to ensure correctness of the implementation - most of my comments are just nitpicks. Curious to hear what the plans are for the utp-testing crate, but I'm sure that'll follow in subsequent prs. One note is that it seems like a lot of the debug messages are somewhat minimal - I'm not sure if it's on purpose - or if more information would be helpful. But that's something we'll find out firsthand as we start running this code

utp-testing/src/main.rs
ogenev: Yeah, this is just a init way to test manually simple uTP to uTP connection. This framework will evolve a lot and will use docker containers in the future.

utp-testing/README.md
ogenev: Hmm, the plan is to build uTP testing framework in the future with clent/server docker instances and network simulator between them as described [here.](https://github.com/marten-seemann/quic-network-simulator). Not sure how this is going to evolve exactly and if it makes sense to be part of `ethportal-peertest` at this point. Maybe we leave it as it is now in a separate crate and see?

trin-core/src/utp/stream.rs
ogenev: This should be `Arc::clone(&self.rx)`.

> Doesn't seem like a receive channel is a great candidate for cloning

I'm not sure what do you exactly mean, could you elaborate? I think it is not unusual to clone a receiver channel as long as it is inside a lock, for example: `Arc<RwLock<T>>`.

trin-core/src/utp/stream.rs
ogenev: I'm removing this, I think we don't need it, it seems a fancy way to get rid of `unwrap` in the test code.

PR #249, opened 2022-02-23
Remove empty OverlayService unit tests

jacobkaufmann:
### What was wrong?

There were leftover empty (unimplemented) unit tests for `OverlayService`.

### How was it fixed?

Remove empty unit tests.

If the tests are deemed necessary, then those tests can be added back later. We should not have dead code.

### To-Do

N/A


njgheorghita (approved):
👍

PR #250, opened 2022-02-24
Serialize buggy tests

njgheorghita:
### What was wrong?
Reports of flaky tests...
- `trin-history/src/content_key::{test_should_store, test_distance_to_key}`
- `trin-core/src/portalnet/overlay_service::{process_request_failure, test_generate_random_node_id}`

### How was it fixed?
Serialized buggy `content_key` tests to eliminate contested file lock between test runs.

### To-Do

- [x] Clean up commit history


njgheorghita:
This pr fixes (afaik after approx 10 successive runs) the buggy tests in `trin-history/` - but I wasn't able to reproduce the `trin-core` flaky tests reported by @jacobkaufmann - I'm happy to create an issue, but maybe we should drop them for now (since this bugfix might actually have addressed the file deadlock issue in the other tests) and Jacob if you see those tests act up again in the future, let's pop it in an issue and I can take a closer look.

jacobkaufmann (approved):
LGTM. And I agree on no issue for this unless we see it resurface.

PR #251, opened 2022-02-24
Misc improvements

njgheorghita:
### What was wrong?
These are just a couple of bugfixes/improvements I'd like to get into `master`. They popped up while trying to cut a first release. First release status is potentially on hold for now. There's something about patching in @ogenev 's `discv5` branch that `cargo-release` doesn't pick up on when packaging each crate individually - which results in a compile error since it reverts to `discv5 = 0.13.0-beta`. There definitely might be a workaround, but I wasn't able to find one. Ultimately, it seems to me like patches are intended for quick-fixes, and not a sustainable dependency. It also seems like fairly clean code habit to not cut any releases that rely on a patch. Tbh, a technical first release is not immediately urgent, but it does seem worthwhile to get those patched fixes in to `discv5`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,8 +14,9 @@ notes:
 		git commit -m "Compile release notes"
 
 release:
+		# First run `make notes version=<version>` before this
 		./newsfragments/validate_files.py is-empty
-		cargo release $(version) --all
+		cargo release $(version) --all --execute
 
 create-docker-image:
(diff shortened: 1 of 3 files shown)


njgheorghita:
Going ahead with the merge on this since it's only a couple minor bugfixes/debug statements

PR #253, opened 2022-02-28
Replace MockContentKey with IdentityContentKey

jacobkaufmann:
### What was wrong?

The `content_id` method for `MockContentKey` always returns a zeroed byte array. In other words, the `content_id` method is the zero function. This is not so useful for testing.

### How was it fixed?

Replace `MockContentKey` with `IdentityContentKey`, where the `content_id` method is the identity function with respect to the value within the content key. This allows us to create content keys with arbitrary content IDs.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -488,16 +488,17 @@ struct DataSizeSum {
 pub mod test {
 
     use super::*;
-    use crate::portalnet::types::content_key::MockContentKey;
+    use crate::portalnet::types::content_key::IdentityContentKey;
+
+    use rand::RngCore;
     use serial_test::serial;
     use std::env;
     use tempdir::TempDir;
 
-    fn generate_content_key(block_hash: &str) -> MockContentKey {
-        let block_hash = hex::decode(block_hash).unwrap();
+    fn generate_random_content_key() -> IdentityContentKey {
         let mut key = [0u8; 32];
-        key.copy_from_slice(block_hash.as_slice());
-        MockContentKey::try_from(key.to_vec()).unwrap()
+        rand::thread_rng().fill_bytes(&mut key);
+        IdentityContentKey::new(key)
     }
 
     fn setup_temp_dir() -> TempDir {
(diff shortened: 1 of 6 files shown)


trin-core/src/portalnet/storage.rs
```diff
@@ -659,39 +646,17 @@ pub mod test {
 
         let value: Vec<u8> = "value".into();
 
-        // block 14115690
-        // content id as u256:              5701445789546971387853890390228320669946681132619292758904237535384791812776
-        // distance from node id as u256:   29607079854947394638644290140513652007972538914554032181524285051455066058182
-        let block_hash = "05c7941834c39a98cbcec5a4890cc6dfcde245ba9fd885980b1544dca2373ff7";
-        let content_key_a = generate_content_key(block_hash);
+        let content_key_a = IdentityContentKey::new([0; 32]);
```
njgheorghita, line 649: This is a definite improvement on reducing the clutter. IIUC - this key has a `0` distance to the `NodeId`? This test just feels a tad underwhelming to me - it doesn't inspire confidence that the correct assumptions are being tested. Idk, that could be just me. I'd suggest at least using a different value for `content_key_a` and also adding a `content_key_c` after line `659` that's yet a further distance away - and then validating that the correct furthest content id property updates accordingly.

njgheorghita (approved):
Nice!

trin-core/src/portalnet/storage.rs
jacobkaufmann: I agree that this is not a great test. Furthermore, I think that the existing tests are sort of brittle. We do not test for the case where the database is empty, where there is no farthest. I think the solution is to add that test case, and for cases where the database is non-empty, we need to add fuzzing, where we use the `distance_to_content_id` method to perform assertions.

```rust
let expected_farthest = if distance_to_content_id(x) > distance_to_content_id(y) {
    x
} else {
    y
};
```

I thought those enhancements were out of scope here and should go in a subsequent PR, but I should have made that intent clear somewhere in these changes. I can add the enhancement you suggested along with an issue or `TODO` to improve the testing there.

jacobkaufmann: I agree that this is not a great test. Furthermore, I think that the existing tests are sort of brittle. We do not test for the case where the database is empty, where there is no farthest. I think the solution is to add that test case, and for cases where the database is non-empty, we need to add fuzzing, where we use the `distance_to_content_id` method to perform assertions.

```rust
let expected_farthest = if distance_to_content_id(x) > distance_to_content_id(y) {
    x
} else {
    y
};
```

I thought those enhancements were out of scope here and should go in a subsequent PR, but I should have made that intent clear somewhere in these changes. I can add the enhancement you suggested along with an issue or `TODO` to improve the testing there.

njgheorghita: Cool, sounds good .... let's do an issue - it's sometimes too easy for `todo`'s to get dropped.... or maybe one week we should have a "spring cleaning" and resolve/issue-ify all the `todo`s that are scattered around the codebase.

PR #255, opened 2022-03-01
[WIP] Implement `--ephemeral` flag (issue #81)

hmrtn:
### What was wrong?
This addresses issue #81, Implement `--ephemeral` flag and is a WIP.

### How was it fixed?
A potential implementation of the ephemeral flag is described [here](https://github.com/ethereum/trin/issues/81#issuecomment-906847076) with a few listed requirements: 
- Implement --ephemeral mode that uses a temporary directory within XDG_CACHE_HOME, using the node-id prefix solution outlined here for determining the directory
-  Leave some sort of marker in that directory to signal when the directory was last used
-  some form of automation that will cleanup this directory, or alternatively, make the --ephemeral flag automatically purge the oldest directories once there are too many.

Changes:  
- [x] Added: the `--ephemeral/-e` boolean flag to `cli.rs`
- [x] Modified: the directory structure to`$XDG_DATA_HOME/Trin/Trin_nodeId` instead of `$XDG_DATA_HOME/Trin_nodeId`
- [x] Added: check if `--ephemeral` is flagged, and if so, delete the oldest `$XDG_DATA_HOME/Trin/Trin_nodeId` directory, based on the last-modified UNIX timestamp
- [x] Renamed: `application_string` to `application_path` for consistency. 
- [x] Revise usage of `println!` to `debug!`
- [x] Replace ambiguous single letter variables
- [x] Use [Path](https://doc.rust-lang.org/stable/std/path/struct.Path.html)

### To-Do

- [ ] Documentation of the `--ephemeral` flag
- [ ] Review usage of `unwrap` 
- [ ] Validate the modified directory structure with unit tests
- [ ] Review usage with the `trin-cli`, e.g. `cargo run -p trin-cli -- discv5_routingTableInfo`
- [ ] Clean up commit history

### DIFF
--- a/newsfragments/270.added.md
+++ b/newsfragments/270.added.md
@@ -1,0 +1,1 @@
+Implement FindContent/Content uTP stream
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -15,7 +15,7 @@ use crate::portalnet::types::messages::{
     ProtocolId, Request, Response,
 };
 
-use crate::utp::stream::{ConnectionKey, UtpListener};
+use crate::utp::stream::{ConnectionKey, UtpListener, BUF_SIZE};
 use crate::utp::trin_helpers::{UtpAccept, UtpMessage, UtpMessageId};
 use discv5::{
     enr::NodeId,
@@ -242,14 +242,60 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
     ) -> Result<Content, OverlayRequestError> {
         // Construct the request.
         let request = FindContent { content_key };
-        let direction = RequestDirection::Outgoing { destination: enr };
+        let direction = RequestDirection::Outgoing {
+            destination: enr.clone(),
+        };
 
         // Send the request and wait on the response.
         match self
             .send_overlay_request(Request::FindContent(request), direction)
             .await
         {
-            Ok(Response::Content(found_content)) => Ok(found_content),
+            Ok(Response::Content(found_content)) => {
+                match found_content {
+                    Content::Content(_) => Ok(found_content),
+                    Content::Enrs(_) => Ok(found_content),
+                    // Init uTP stream if `connection_id` is received
+                    Content::ConnectionId(conn_id) => {
+                        let conn_id = u16::from_be(conn_id);
+                        self.utp_listener
+                            .write()
+                            .await
+                            .listening
+                            .insert(conn_id, UtpMessageId::FindContentStream);
+
+                        // initiate the connection to the acceptor
+                        let conn = self
+                            .utp_listener
+                            .write()
+                            .await
+                            .connect(conn_id, enr.node_id())
+                            .await;
+
+                        match conn {
+                            Ok(mut conn) => {
+                                // TODO: Implement this with recv_from
+                                // Handle STATE packet for SYN
+                                let mut buf = [0; BUF_SIZE];
+                                match conn.recv(&mut buf).await {
+                                    Ok(_) => {
+                                        // TODO: Wait for uTP transfer and return the data
+                                        Ok(found_content)
+                                    }
+                                    Err(msg) => {
+                                        warn!("Unable to handle SYN-ACK packet: {msg}");
+                                        Err(OverlayRequestError::UtpError(msg.to_string()))
+                                    }
+                                }
+                            }
+                            Err(msg) => {
+                                warn!("Unable to initiate uTP stream with remote node {msg}");
+                                Err(OverlayRequestError::UtpError(msg.to_string()))
+                            }
+                        }
+                    }
+                }
+            }
             Ok(_) => Err(OverlayRequestError::InvalidResponse),
             Err(error) => Err(error),
         }
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -96,6 +96,10 @@ pub enum OverlayRequestError {
     /// Error types resulting from building ACCEPT message
     #[error("Error while sending offer message")]
     OfferError(String),
+
+    /// uTP request error
+    #[error("uTP request error")]
+    UtpError(String),
 }
 
 impl From<discv5::RequestError> for OverlayRequestError {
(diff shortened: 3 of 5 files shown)


trin-core/src/utils/db.rs
```diff
@@ -1,30 +1,64 @@
...
-    // macOS: ~/Library/Application Support/Trin
-    // Unix-like: $HOME/.local/share/trin
+    fs::create_dir_all(&application_path).expect("Unable to create data directory folder");
 
     // Append first 8 characters of Node ID
-    let mut application_string = "Trin_".to_owned();
     let node_id_string = hex::encode(node_id.raw());
     let suffix = &node_id_string[..8];
-    application_string.push_str(suffix);
+    let path = format!("{}/Trin_{}", &application_path, &suffix);
```
ogenev, line 21: I think we should use here [Path](https://doc.rust-lang.org/stable/std/path/struct.Path.html) and make this OS agnostic.

ogenev:
I think this is a great start :+1:.
If I understand the issue correctly, It seems the `--ephemeral` flag should create a **temporary** directory which should be deleted right after Trin exit.

In the current implementation, we delete only the oldest used dirs when creating a new one, which I think should be the default behavior in a mixed scenario with the `--ephemeral` flag described in the issue.

I would look at how we can use the flag to delete the current directory right after the Trin process exit.

trin-core/src/utils/db.rs
```diff
@@ -1,30 +1,64 @@
...
+            n_paths += 1;
+            if let Ok(p) = p {
+                // Get oldest last-modified Trin_nodeId directory timestamp and path
+                if let Ok(metadata) = p.metadata() {
+                    if metadata.modified().unwrap() < min_timestamp {
+                        min_timestamp = metadata.modified().unwrap();
+                        min_path = p.path();
+                    }
+                } else {
+                    println!("Couldn't get metadata for {:?}", p.path());
```
ogenev, line 40:
```suggestion
                    debug!("Couldn't get metadata for {:?}", p.path());
```

trin-core/src/utils/db.rs
```diff
@@ -1,30 +1,64 @@
...
+    let path = format!("{}/Trin_{}", &application_path, &suffix);
+
+    if trin_config.ephemeral {
+        fs::create_dir_all(&path).expect("Unable to create ephemeral data directory");
+        let paths = fs::read_dir(&application_path).unwrap();
+        // NOTE: SystemTime is machine dependent
+        let mut min_timestamp = SystemTime::now();
+        let mut min_path = PathBuf::new();
+        let mut n_paths = 0;
+        for p in paths {
```
njgheorghita, line 30: *minor nitpick* - we try to stay away from single-letter variables - I'd suggest going with `path` for `p` (and updating the prior `path` on line 21 to something else) - as well as `path_count` / `num_paths` for `n_paths`. In general avoiding single letter vars helps a ton with readability which is a priority over minimizing code length, especially in open-source when you can expect a fair amount of different people to be reading this code.

trin-core/src/utils/db.rs
```diff
@@ -1,30 +1,64 @@
...
     // Append first 8 characters of Node ID
-    let mut application_string = "Trin_".to_owned();
     let node_id_string = hex::encode(node_id.raw());
     let suffix = &node_id_string[..8];
-    application_string.push_str(suffix);
+    let path = format!("{}/Trin_{}", &application_path, &suffix);
+
+    if trin_config.ephemeral {
+        fs::create_dir_all(&path).expect("Unable to create ephemeral data directory");
+        let paths = fs::read_dir(&application_path).unwrap();
```
njgheorghita, line 25: I'm seeing a couple naked unwraps here and there, [which we try to avoid](https://github.com/ethereum/trin/blob/master/docs/contributing.md#error-handling).

trin-core/src/utils/db.rs
```diff
@@ -1,30 +1,64 @@
...
+    fs::create_dir_all(&application_path).expect("Unable to create data directory folder");
 
     // Append first 8 characters of Node ID
-    let mut application_string = "Trin_".to_owned();
     let node_id_string = hex::encode(node_id.raw());
     let suffix = &node_id_string[..8];
-    application_string.push_str(suffix);
+    let path = format!("{}/Trin_{}", &application_path, &suffix);
+
+    if trin_config.ephemeral {
```
njgheorghita, line 23: *nitpick* I'd consider moving the functionality within these brackets out into its own function

njgheorghita (commented):
Great start! Seems like you're on the right track - most of my comments are just nitpicks. Seems like this functionality could also be isolated and unit-tested - I'd recommend looking into using a tempdir in the tests - which we use [here](https://github.com/ethereum/trin/blob/053bcd70668621d1d66432681fd3debae5913ea0/trin-history/src/content_key.rs#L175)

trin-core/src/utils/db.rs
hmrtn: > It seems the --ephemeral flag should create a temporary directory which should be deleted right after Trin exit.

I think that definitely makes more sense 👍 

[This](https://rust-cli.github.io/book/in-depth/signals.html) looks like a good resource on rust signal handling. 

Thanks for the feedback!

hmrtn:
I appreciate the feedback and I updated the Tasks based on the suggestions here!

hmrtn:
> I think this is a great start 👍. If I understand the issue correctly, It seems the `--ephemeral` flag should create a **temporary** directory which should be deleted right after Trin exit.

> I would look at how we can use the flag to delete the current directory right after the Trin process exit.

That makes way more sense 👍 . What is the current exit process in Trin? Maybe we can use `std::process:exit` or something outlined in [this](https://rust-cli.github.io/book/in-depth/signals.html)?

hmrtn:
Is it reasonable to rely on the OS to cleanup its temporary directory?

Doing so would simplify implementing the `--ephemeral` flag. If it is flagged, a temporary directory can be created with [`tempdir::TempDir::new("Trin")`](https://docs.rs/tempdir/0.3.5/tempdir/struct.TempDir.html), in the OS's temporary directory location, similar to what @njgheorghita mentioned with some of the tests. And instead of explicitly deleting the temp directory on `SIGINT` with `tokio::signal::crtl_c()`, we leave it up to the OS to handle. 

If we can rely on the OS for cleanup, then I believe this would be an appropriate and simple solution. I've tried a few other approaches, but I think many require too many changes. Let me know if think of any alternative approaches!

I've updated the commit to include the changes described. Looking for thoughts!

ogenev:
> Is it reasonable to rely on the OS to cleanup its temporary directory?

If I understand the [docs](https://docs.rs/tempdir/latest/tempdir/struct.TempDir.html) correctly, `TempDir` relies on Rust ownership, not directly on OS, to decide when to delete the folder. If `TempDir` goes out of scope, the folder is deleted.

I think this is not a reasonable solution in our case because `Trin` is a long-living process and most of the time will be terminated with `SIGINT` signal and according to tempdir docs:

_"Note that if the program exits before the TempDir destructor is run, such as via [std::process::exit](http://doc.rust-lang.org/std/process/fn.exit.html), by segfaulting, or by receiving a signal like SIGINT, then the temporary directory will not be deleted."_

It will also require extra caution when exactly the `TempDir `value is going out of scope, that's why it is probably more suitable for short-running processes like test execution for example.

PR #257, opened 2022-03-02
Refactor Message

jacobkaufmann:
### What was wrong?

1. `Message` did not implement SSZ `Encode` and `Decode`.
2.  `Message` implemented custom `to_bytes` and `from_bytes` methods.

### How was it fixed?

1. Unite all Portal message variants in `Message` and derive SSZ `Encode` and `Decode`.
2. Implement `Into<Vec<u8>>` and `TryFrom<Vec<u8>>` for `Message`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -144,14 +144,11 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
         &self,
         talk_request: &TalkRequest,
     ) -> Result<Response, OverlayRequestError> {
-        let request = match Message::from_bytes(talk_request.body()) {
-            Ok(Message::Request(request)) => request,
-            Ok(_) => {
-                return Err(OverlayRequestError::InvalidRequest(format!(
-                    "Message not recognized request type: {:?}",
-                    talk_request.body()
-                )))
-            }
+        let request = match Message::try_from(Vec::<u8>::from(talk_request.body())) {
+            Ok(message) => match Request::try_from(message) {
+                Ok(request) => request,
+                Err(err) => return Err(OverlayRequestError::InvalidRequest(err.to_string())),
+            },
             Err(_) => return Err(OverlayRequestError::DecodeError),
         };
         let direction = RequestDirection::Incoming {
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -620,12 +620,14 @@ impl<TContentKey: OverlayContentKey + Send> OverlayService<TContentKey> {
         // which will be received in the main loop.
         tokio::spawn(async move {
             let response = match discovery
-                .send_talk_req(destination, protocol, Message::Request(request).to_bytes())
+                .send_talk_req(destination, protocol, Message::from(request).into())
                 .await
             {
-                Ok(talk_resp) => match Message::from_bytes(&talk_resp) {
-                    Ok(Message::Response(response)) => Ok(response),
-                    Ok(_) => Err(OverlayRequestError::InvalidResponse),
+                Ok(talk_resp) => match Message::try_from(talk_resp) {
+                    Ok(message) => match Response::try_from(message) {
+                        Ok(response) => Ok(response),
+                        Err(_) => Err(OverlayRequestError::InvalidResponse),
+                    },
                     Err(_) => Err(OverlayRequestError::DecodeError),
                 },
                 Err(error) => Err(error.into()),
--- a/trin-core/src/portalnet/types/messages.rs
+++ b/trin-core/src/portalnet/types/messages.rs
@@ -14,7 +14,6 @@ use ssz_types::{typenum, BitList, VariableList};
 use thiserror::Error;
 use validator::ValidationError;
 
-use crate::portalnet::overlay_service::OverlayRequestError;
 use crate::portalnet::{types::uint::U256, Enr};
 
 pub type ByteList = VariableList<u8, typenum::U2048>;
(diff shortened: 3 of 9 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -144,14 +144,11 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
...
     ) -> Result<Response, OverlayRequestError> {
-        let request = match Message::from_bytes(talk_request.body()) {
-            Ok(Message::Request(request)) => request,
-            Ok(_) => {
-                return Err(OverlayRequestError::InvalidRequest(format!(
-                    "Message not recognized request type: {:?}",
-                    talk_request.body()
-                )))
-            }
+        let request = match Message::try_from(Vec::<u8>::from(talk_request.body())) {
```
njgheorghita, line 147: Would it make sense to implement `TryFrom<TalkRequest>` as well as `TryFrom<Vec<u8>>` on `Message` .... or maybe even `impl TryFrom<TalkRequest> for Request`?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -620,12 +620,14 @@ impl<TContentKey: OverlayContentKey + Send> OverlayService<TContentKey> {
...
             let response = match discovery
-                .send_talk_req(destination, protocol, Message::Request(request).to_bytes())
+                .send_talk_req(destination, protocol, Message::from(request).into())
                 .await
             {
-                Ok(talk_resp) => match Message::from_bytes(&talk_resp) {
-                    Ok(Message::Response(response)) => Ok(response),
-                    Ok(_) => Err(OverlayRequestError::InvalidResponse),
+                Ok(talk_resp) => match Message::try_from(talk_resp) {
+                    Ok(message) => match Response::try_from(message) {
```
njgheorghita, line 627: I guess in this case something like `impl TryFrom<Vec<u8>> for Response`? It seems cleaner to abstract away as much type-matching code as we can behind an `impl` since it'll reduce clutter/improve readability in the more "core" code - but there's probably also a point where it can be taken too far

trin-core/src/portalnet/types/messages.rs
```diff
@@ -619,272 +570,205 @@ impl FromStr for HexData {
...
+        let expected_encoded = "05017468652063616b652069732061206c6965";
+        assert_eq!(encoded, expected_encoded);
+
+        let decoded = Message::try_from(hex::decode(expected_encoded).unwrap()).unwrap();
+        assert_eq!(decoded, content);
     }
 
     #[test]
-    fn test_vector_content_with_enrs() {
+    fn message_encoding_content_enrs() {
```
njgheorghita, line 725: It might be a hangover from python - but I prefix every test with `test_*` - seems like it's not really a thing in Rust. I guess given the `#[test]` decorator, it's obviously a test. But on the other hand I do find it useful to use the `test_` prefix when grepping through a file. Whatever we decide on should probably be added to the `CONTRIBUTING.md` doc so the codebase stays consistent.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -619,272 +570,205 @@ impl FromStr for HexData {
...
-    #[should_panic]
-    fn invalid_protocol_id() {
-        let hex_string = "500F";
-        ProtocolId::from_str(hex_string).unwrap();
+        let decoded = Message::try_from(hex::decode(expected_encoded).unwrap()).unwrap();
+        assert_eq!(decoded, nodes);
     }
 
-    // test vectors sourced from
-    // https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-test-vectors.md
```
njgheorghita, line 743: I think it's a good idea to leave this comment in for reference

trin-core/src/portalnet/types/messages.rs
```diff
@@ -619,272 +570,205 @@ impl FromStr for HexData {
...
 
-    #[test]
-    fn test_vector_find_nodes() {
-        let distances = vec![256, 255];
-        let expected = "02040000000001ff00";
-        let message = Message::Request(Request::FindNodes(FindNodes { distances }));
-        assert_eq!(hex::encode(message.to_bytes()), expected);
-        let decoded = Message::from_bytes(message.to_bytes().as_slice()).unwrap();
-        assert_eq!(decoded, message);
+        let decoded = Message::try_from(hex::decode(expected_encoded).unwrap()).unwrap();
```
njgheorghita, line 689: *nitpick* on one hand you just validated that `expected_encoded == encoded` on the other hand it makes more sense to me to use the tested variable for the round trip test (aka `hex::decode(encoded)`)

njgheorghita (approved):
Lookin' good!

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann: The limitation here is that only `Message` is aware of the SSZ encoding for some arbitrary byte array, because the derived SSZ traits on the enum recognize the union type selector. This is not true for the `Request` and `Response` types.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: For a similar reason mentioned below, we cannot implement `TryFrom<TalkRequest>` for `Request`. The `Request` type does not derive the SSZ encoding traits, so it cannot handle an arbitrary byte array.

We could implement `TryFrom<TalkRequest>` for `Message`, but I don't see much additional utility here.

trin-core/src/portalnet/types/messages.rs
jacobkaufmann: I see what you mean. My thinking was that we should decode from "ground truth" encoding, but like you said we just validated its equality to `encoded`. And I do think it reads more nicely the way you've proposed.

trin-core/src/portalnet/types/messages.rs
jacobkaufmann: Yeah, it's just my personal preference to omit the `test` prefix due to redundancy. It's not a strong preference, but in the interest of consistency, I can add a note to `CONTRIBUTING.md` and perhaps a follow-up PR to refactor any existing tests. If you have further thoughts, please let me know. If not, I'll assume that we will go with this convention.

PR #259, opened 2022-03-07
Adds modified versions of QueryPool, Query types from sigp discv5

mrferris:
### What does this add?

This PR adds relatively generic utility types used for iterative FindNodes/FindContent implementations, mostly taken from https://github.com/sigp/discv5/blob/master/src/query_pool.rs with some customizations/re-organization done. 

Customizations include:

- Stripping out of `PredicateQuery` types. These allowed one to specify an arbitrary filter to run over ENRs discovered. Removing this type is the reason for the compiler currently requiring `PhantomData` fields in `query_pool.rs`, because the `PredicateQuery` definition used the `TResult` field while the `FindNodeQuery` definition does not.
- Modify `QueryInfo` to use `Trin` types when generating requests instead of `sigp/discv5' types.

The main types defined are the `QueryPool` type, the `Query` type (both defined in `find/query_pool.rs`), and the `FindNodeQuery`  type (defined in `find/iterators/findnodes.rs`). 

The `QueryPool` type holds a set of `Query` entries, each of which has an instance of the `QueryPeerIter` enum, which points to a `FindNodeQuery` (or a `FindContentQuery` once implemented), whose `on_success`/`on_failure` methods define the rules by which the query is advanced each time a response is received from a peer. 

### How will it be used?

The subsequent PR that uses these types:

-  adds a `QueryPool` member to the `OverlayService` struct.
-  adds a function that polls the `QueryPool` for `QueryPoolState` objects, based off of which the query will either finish or a`FINDNODES` message will be sent to the next peer (whose info is contained in the `QueryPoolState` object) via a `TALK_REQ`.
- adds a function that calls `on_success`/`on_failure` when a `NODES` message is received from a node that is part of a query.
- calls the `QueryPool`'s `add_findnode_query` method in the OverlayService to begin a FindNodes query.

Another subsequent PR will need to define `FindContentQuery` and its respective `on_success`/`on_failure` methods, at which point this approach will work for `FindContent` as well. I envision the `OverlayService` having two `QueryPool` members, one for each type of query. Each `QueryPool` will be instantiated with different types corresponding to the types of data that will be returned (content or ENRs).

### DIFF
--- a/trin-core/src/portalnet/find/iterators/findnodes.rs
+++ b/trin-core/src/portalnet/find/iterators/findnodes.rs
@@ -40,6 +40,7 @@ pub struct FindNodeQuery<TNodeId> {
     progress: QueryProgress,
 
     /// The closest peers to the target, ordered by increasing distance.
+    /// Assumes that no two peers are equidistant.
     closest_peers: BTreeMap<Distance, QueryPeer<TNodeId>>,
 
     /// The number of peers for which the query is currently waiting for results.
@@ -152,19 +153,17 @@ where
             Entry::Occupied(mut entry) => match entry.get().state {
                 QueryPeerState::Waiting(..) => {
                     if self.num_waiting < 0 {
-                       error!("In findenodes.rs on_success num_waiting is negative. Aborting.")
+                       error!("In findenodes.rs on_success num_waiting is negative.")
                        return
                     }
                     self.num_waiting -= 1;
                     let peer = e.get_mut();
                     peer.peers_returned += closer_peers.len();
-                    // mark the peer as succeeded
                     peer.state = QueryPeerState::Succeeded;
                 }
                 QueryPeerState::Unresponsive => {
                     let peer = e.get_mut();
                     peer.peers_returned += closer_peers.len();
-                    // mark the peer as succeeded
                     peer.state = QueryPeerState::Succeeded;
                 }
                 QueryPeerState::NotContacted
--- a/trin-core/src/portalnet/find/iterators/findnodes.rs
+++ b/trin-core/src/portalnet/find/iterators/findnodes.rs
@@ -1,0 +1,701 @@
+// Copyright 2018 Parity Technologies (UK) Ltd.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a
+// copy of this software and associated documentation files (the "Software"),
+// to deal in the Software without restriction, including without limitation
+// the rights to use, copy, modify, merge, publish, distribute, sublicense,
+// and/or sell copies of the Software, and to permit persons to whom the
+// Software is furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+// OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+// FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+// DEALINGS IN THE SOFTWARE.
+
+// This basis of this file has been taken from the rust-libp2p codebase:
+// https://github.com/libp2p/rust-libp2p
+//
+use super::super::query_pool::{QueryState};
+use discv5::{
+    kbucket::{Distance, Key},
+};
+use std::{
+    collections::btree_map::{BTreeMap, Entry},
+    time::{Duration, Instant},
+};
+
+
+#[derive(Debug, Clone)]
+pub struct FindNodeQuery<TNodeId> {
+    /// The target key we are looking for
+    target_key: Key<TNodeId>,
+
+    /// The current state of progress of the query.
+    progress: QueryProgress,
+
+    /// The closest peers to the target, ordered by increasing distance.
+    closest_peers: BTreeMap<Distance, QueryPeer<TNodeId>>,
+
+    /// The number of peers for which the query is currently waiting for results.
+    num_waiting: usize,
+
+    /// The configuration of the query.
+    config: FindNodeQueryConfig,
+}
+
+/// Configuration for a `Query`.
+#[derive(Debug, Clone)]
+pub struct FindNodeQueryConfig {
+    /// Allowed level of parallelism.
+    ///
+    /// The `α` parameter in the Kademlia paper. The maximum number of peers that a query
+    /// is allowed to wait for in parallel while iterating towards the closest
+    /// nodes to a target. Defaults to `3`.
+    pub parallelism: usize,
+
+    /// Number of results to produce.
+    ///
+    /// The number of closest peers that a query must obtain successful results
+    /// for before it terminates. Defaults to the maximum number of entries in a
+    /// single k-bucket, i.e. the `k` parameter in the Kademlia paper.
+    pub num_results: usize,
+
+    /// The timeout for a single peer.
+    ///
+    /// If a successful result is not reported for a peer within this timeout
+    /// window, the iterator considers the peer unresponsive and will not wait for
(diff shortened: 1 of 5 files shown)


trin-core/src/portalnet/find/query_pool.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+/// The result of a `Query`.
+pub struct QueryResult<TTarget, TClosest> {
+    /// The target of the query.
+    pub target: TTarget,
+    /// The closest peers to the target found by the query.
+    pub closest_peers: TClosest,
+}
```
jacobkaufmann, line 272: Have you considered how we would extend this to results for `FindContent` queries, where the result may not contain a list of closest peers?

jacobkaufmann, line 272: Have you considered how we would extend this to results for `FindContent` queries, where the result may not contain a list of closest peers?

trin-core/src/portalnet/find/query_pool.rs
```diff
@@ -0,0 +1,295 @@
...
+enum QueryPeerIter<TNodeId> {
+    FindNode(FindNodeQuery<TNodeId>)
+}
+
+impl<TTarget, TNodeId, TResult> Query<TTarget, TNodeId, TResult>
+where
+    TTarget: TargetKey<TNodeId>,
+    TNodeId: Into<Key<TNodeId>> + Eq + Clone,
+    TResult: Into<TNodeId> + Clone,
+{
```
jacobkaufmann, line 199: Related to comment below. This implementation seems to imply that the result type will be some key for a node ID. This is also evident in the `on_success` method with the type of the `new_peers` argument.

trin-core/src/portalnet/find/query_pool.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn key(&self) -> Key<TNodeId>;
+}
+
+/// A `QueryPool` provides an aggregate state machine for driving `Query`s to completion.
+///
+/// Internally, a `Query` is in turn driven by an underlying `QueryPeerIter`
+/// that determines the peer selection strategy, i.e. the order in which the
+/// peers involved in the query should be contacted.
+pub struct QueryPool<TTarget, TNodeId, TResult> {
+    next_id: usize,
```
jacobkaufmann, line 41: Why not use `QueryId` type here?

trin-core/src/portalnet/find/query_info.rs
```diff
@@ -0,0 +1,135 @@
...
+///
+/// As an example, if the target has a distance of 12 from the remote peer, the sequence of distances that are sent for increasing iterations would be [12, 13, 11, 14, 10, .. ].
+fn findnode_log2distance(target: NodeId, peer: NodeId, size: usize) -> Option<Vec<u16>> {
+    if size > 127 {
+        // invoke and endless loop - coding error
+        panic!("Iterations cannot be greater than 127");
+    }
+
+    let dst_key: Key<NodeId> = peer.into();
+    let distance_u64 = dst_key.log2_distance(&target.into())?;
```
jacobkaufmann, line 67: The `log2_distance` method relies on [XOR distance](https://github.com/sigp/discv5/blob/master/src/kbucket/key.rs#L91) defined on `Key` in `sigp/discv5`. This is not a problem for the history network, but this may require some rework for the state network.

trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -0,0 +1,701 @@
...
+}
+
+/// Representation of a peer in the context of a query.
+#[derive(Debug, Clone)]
+struct QueryPeer<TNodeId> {
+    /// The `KBucket` key used to identify the peer.
+    key: Key<TNodeId>,
+
+    /// The current rpc request iteration that has been made on this peer.
+    iteration: usize,
```
jacobkaufmann, line 388: Why do we have `iteration` field? It looks like this is not a field for `QueryPeer` on the master branch of `sigp/discv5`. Also, we seem to never use this field.

jacobkaufmann (commented):
A change like this is hard to evaluate on its own so most of my comments focus on extensibility to other queries like `FindContent`.

Also, it looks like there was a minor discrepancy between the implementation of the `FindNodeQuery` here and the one in `sigp/discv5`, so I would like to confirm that we based this PR off the latest from the master branch there.

I don't have any changes to request, but I would like to wait until the comments are addressed before approval.

trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -0,0 +1,701 @@
...
+                        // The query is still waiting for a result from a peer and the
+                        // `result_counter` did not yet reach `num_results`. Therefore
+                        // the query is not yet done, regardless of already successful
+                        // queries to peers farther from the target.
+                        result_counter = None;
+                    }
+                }
+
+                QueryPeerState::Succeeded => {
+                    if let Some(ref mut cnt) = result_counter {
```
njgheorghita, line 284: *nitpick* `count` > `cnt`

trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -0,0 +1,701 @@
...
+                    QueryPeerState::NotContacted | QueryPeerState::Waiting { .. }
+                )
+            });
+
+            let target_key = query.target_key.clone();
+            let num_results = query.config.num_results;
+            let result = query.into_result();
+            let closest = result.into_iter().map(Key::from).collect::<Vec<_>>();
+
+            // assert_eq!(result.target, target);
```
njgheorghita, line 590: And this

trin-core/src/portalnet/find/query_info.rs
```diff
@@ -0,0 +1,135 @@
...
+                    .ok_or("Requested a node find itself")?;
+                Request::FindNodes(FindNodes { distances })
+            }
+        };
+
+        Ok(request)
+    }
+}
+
+impl TargetKey<NodeId> for QueryInfo {
```
njgheorghita, line 44: Why not `TNodeId` to maintain consistency with the `findnodes.rs` code?

njgheorghita (approved):
Lookin' good! The comments are super helpful. Mostly nitpicks from me, a bit hard to judge the api in isolation, but i'm looking forward to the following prs & findcontent support

trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -0,0 +1,701 @@
...
+        if let QueryProgress::Finished = self.progress {
+            return;
+        }
+
+        let key: Key<TNodeId> = node_id.clone().into();
+        let distance = key.distance(&self.target_key);
+
+        // Mark the peer's progress, the total nodes it has returned and it's current iteration.
+        // If the node returned peers, mark it as succeeded.
+        match self.closest_peers.entry(distance) {
```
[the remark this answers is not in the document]
jacobkaufmann: IIUC, this is true for the XOR distance metric, but not for the state distance function.

trin-core/src/portalnet/find/query_pool.rs
mrferris: I'm thinking that we could add a `retrieved_content: Option<Vec<u8>>` field to this struct.

trin-core/src/portalnet/find/query_pool.rs
mrferris: Similar to above, I picture us adding an optional content field to `on_success`.

As for `TResult`, that's a good point, but in my mind I don't see us needing to use `TResult` to return the content if we pass in a channel to send the content back over. In this case, we might be able to get rid of `TResult` altogether. It was mostly used for the predicate queries, which have been removed.

mrferris: Similar to above, I picture us adding an optional content field to on_success.

As for TResult, that's a good point, but in my mind I don't see us needing to use TResult to return the content if we pass in a channel to send the content back over. In this case, we might be able to get rid of TResult altogether. It was mostly used for the predicate queries, which have been removed.

trin-core/src/portalnet/find/query_info.rs
mrferris: good point, i'll file an issue after i merge this

trin-core/src/portalnet/find/iterators/findnodes.rs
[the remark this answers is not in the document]
mrferris: Good point @njgheorghita. Added a note. After thinking about it, I think it's extremely unlikely that it does happen (with XOR) but even if it does, it doesn't look like the results would be very severe if presented with results from a node that it thinks it already got results from. It would just return, so basically one of the nodes would be chosen and the other ignored.

Good point @jacobkaufmann. I'll add that to the issue I file about making queries distance function configurable.

PR #260, opened 2022-03-11
Add Metric trait for generic DHT distance function

jacobkaufmann:
### What was wrong?

The `OverlayProtocol` and `OverlayService` structs are not generic over the distance function (i.e. metric) The distance function computes distances between both node-node pairs and node-content pairs. Without this change, all distance computations use the XOR metric.

### How was it fixed?

- Define a `Metric` trait with a single associated method `distance`.
- Add a generic type parameter to `OverlayProtocol` and `OverlayService` with a trait bound on `Metric`.
- Define `XorMetric` as a unit struct in `trin-core` and implement `Metric` for this type.

### Additional notes

Given that the XOR metric is the common metric used in Kademlia-based DHTs, I think it makes sense to define this metric in `trin-core`. Given the current specs, multiple networks intend to use the XOR metric, and a single definition in `trin-core` prevents code duplication and possible inconsistencies. It also provides a default metric and a metric for testing, where we do not care about the specific metric.

The `PortalStorage` struct is not generic over the distance function, but I chose not to handle that here. For `PortalStorage`, the distance function also dictates what the proper "find farthest" query is for the database. It's not obvious to me how to cleanly bake that query into the trait definition, or how to match up a generic type parameter to that query.

### To-Do

- [x] Clean up commit history


trin-core/src/portalnet/overlay.rs
```diff
@@ -81,9 +81,13 @@ pub struct OverlayProtocol<TContentKey> {
     /// Use a phantom, because we don't store any keys in this struct.
     /// For example, this type is used when decoding a content key received over the network.
     phantom_content_key: PhantomData<TContentKey>,
+    /// Associate a metric with the overlay network.
+    phanton_metric: PhantomData<TMetric>,
```
njgheorghita, line 85: *typo* `phantom_metric`

trin-core/src/portalnet/overlay.rs
```diff
@@ -124,6 +128,7 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
             request_tx,
             utp_listener,
             phantom_content_key: PhantomData,
+            phanton_metric: PhantomData,
```
njgheorghita, line 131: same here

njgheorghita (approved):
Looks good! I'd also create an issue to track the question about how we can use a similar trait in `PortalStorage`.

PR #265, opened 2022-03-15
Integrates FindNodes query into OverlayService

mrferris:
### What was wrong?

Trin didn't implement FindNode queries. (#315)

### How was it fixed?

- adds a QueryPool member `query_pool` to the OverlayService struct.

- adds a `query_event_poll` function that polls the QueryPool for QueryPoolState objects, based off of which (in the main `start` loop) the query will either finish or a FINDNODES message will be sent to the next peer (whose info is contained in the QueryPoolState object) via a TALK_REQ.

- calls on_success/on_failure when a NODES message is received from a node that is part of a query.

- calls the QueryPool's `add_query` method from within `init_findnode_query` to begin bucket refresh queries. 

- adds an `initialize_routing_table` method that runs a node lookup on the local node ID in order to find nearby nodes as soon as Trin boots.

- adds unit tests for new `OverlayService` methods: `init_find_nodes`, `find_enr`, and `advance_query`

- adds fields for configuring queries to the `OverlayConfig` that is used to configure the `OverlayService`

### To-Do

- [x] Add initialization lookup for local node ID 
- [x] Add unit tests
- [x] Clean up commit history

Subsequent PRs will:
- Add an async version of `init_find_nodes` that will return ENRs via the callback field of the `QueryInfo`, and include high level integration tests that use this async method. 
- Add `portal_historyRoutingTableInfo` and `portal_stateRoutingTableInfo` jsonrpc endpoints for viewing populated routing tables.
- After self-lookup, we refresh every bucket farther than the closest peer in ascending order, starting with the closest bucket. We currently effectively do this in random order via the bucket refreshes that begin after the initial self-lookup.
- Removing nodes that become disconnected mid-query from active queries to end with fewer than k results if necessary, rather than wait until time-out.


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -560,7 +668,16 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+
+        // Currently limiting to 3 ENRs until multiple messages is supported.
+        let enrs = {
+            let enrs = self.nodes_by_distance(distances64);
+            if enrs.len() > 3 {
+                enrs[0..3].to_vec()
+            } else {
+                enrs
+            }
+        };
```
jacobkaufmann, line 680: While we don't yet support multiple `Nodes` responses, we should not limit the number of ENRs to return to 3. Based on [this spec issue](https://github.com/ethereum/portal-network-specs/issues/136), which will inform the longer term strategy, we should be able to return 8 or 9.

We can use the [`size` method of `Enr`](https://docs.rs/enr/0.5.1/enr/struct.Enr.html#method.size) to determine the maximum number of ENRs that we can fit in a single message. This is better than relying on a heuristic.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -848,13 +965,43 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            .map(|ssz_enr| ssz_enr.into())
+            .collect();
+
+        let enrs_clone = enrs.clone();
+        self.process_discovered_enrs(enrs);
+
+        let node_id = source.node_id();
+
+        // if this is part of a query, update the query
+        if let Some(query_id) = self.active_query_requests.write().remove(&node_id) {
```
jacobkaufmann, line 981: This logic would be common to both `Nodes` and possibly `Content` responses, so we could put this logic into `process_discovered_enrs`.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -848,13 +965,43 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            debug!(
+                "[{:?}] Nodes response is part of query {}",
+                self.protocol, query_id
+            );
```
jacobkaufmann, line 985: I'm not sure whether this log will provide much useful information.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1152,15 +1301,55 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         Ok(closest_nodes)
     }
 
-    fn send_recursive_findnode(&self, _target: &NodeId) {
-        // TODO: Implement Recursive(iterative) FINDNODE. This is a stub.
+    fn send_iterative_findnode(&mut self, _target: &NodeId) {
+        const DISTANCES_TO_REQUEST_PER_PEER: usize = 1;
```
jacobkaufmann, line 1305: This belongs in a config struct or a `const` at the module level, not in the method.

How did we decide on the value 1? This value is 3 in `sigp/discv5`.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1152,15 +1301,55 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            for closest in kbuckets.closest_values(&target_key) {
+                // Add the known ENR's to the untrusted list
+                target.untrusted_enrs.push(closest.value.enr);
+                // Add the key to the list for the query
+                known_closest_peers.push(closest.key);
+            }
+        }
+
+        if known_closest_peers.is_empty() {
+            warn!("No known_closest_peers found. Return empty result without sending query.");
```
jacobkaufmann, line 1325: This log message only makes sense if the reader knows what the `known_closest_peers` variable represents. We can make this message more clear to the reader.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1152,15 +1301,55 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        } else {
+            let query_config = FindNodeQueryConfig {
```
jacobkaufmann, line 1327: We ought to derive this from some higher-level config.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1152,15 +1301,55 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                parallelism: 3,
+                num_results: 25,
```
jacobkaufmann, line 1329: How did we decide on 25? `sigp/discv5` uses the maximum number of nodes per k-bucket, i.e. `k`.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -410,6 +428,54 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                                    request,
+                                    RequestDirection::Outgoing {
+                                        destination: enr.clone()
+                                    },
+                                    None,
+                                );
+
+                                self.active_query_requests.write().insert(node_id, query_id);
+
+                                debug!("[{:?}] Sending FINDNODES on behalf of query {}", self.protocol, query_id);
```
jacobkaufmann, line 449: I don't think we need this log.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -410,6 +428,54 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                                let _ = self.request_tx.send(request);
+
+                            } else {
+                                error!("[{:?}] Query Pool wants to send a FINDNODES to {},
+                                        but ENR is not known.", self.protocol, node_id);
```
jacobkaufmann, line 454: I think we can make this more concise: "Unable to send query {query_id} to unknown ENR". Something like that.

jacobkaufmann (changes requested):
Took an initial pass. Happy to see this logic! That said, a few things we should address:

- Certain query constants/parameters belong in a configuration struct.
- Adjust some query parameters to more sensible defaults.
- Use ENR sizes to determine the number of ENRs to put in the response.
- Move query update logic to `process_discovered_enrs`.

The rest of the comments represent questions or suggestions on things like structure or logging. Those might require further discussion in their own threads though.

Another thing that we need to address the `TODO` in `OverlayService::spawn` to populate the local routing table and insert the local node into the routing tables of other nodes. I'm not sure whether that was deemed out-of-scope here or whether it was just overlooked.

And finally, how did we validate that this logic functions as desired? I understand that it's difficult to test, but there are zero changes to unit tests. There is also no testing methodology described in the current PR description.

trin-core/Cargo.toml
```diff
@@ -30,6 +31,7 @@ num = "0.4.0"
 parking_lot = "0.11.2"
 prometheus_exporter = "0.8.4"
 rand = "0.8.4"
+rand_07 = { package = "rand", version = "0.7" }
```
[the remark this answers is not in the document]
mrferris: Very much agreed. Both of these are due to inheriting tests from `sigp/discv5` that I'll update to use latest once this is merged.

trin-core/src/portalnet/find/query_pool.rs
```diff
@@ -38,10 +38,10 @@ pub trait TargetKey<TNodeId> {
 /// that determines the peer selection strategy, i.e. the order in which the
 /// peers involved in the query should be contacted.
 pub struct QueryPool<TTarget, TNodeId, TResult> {
-    next_id: QueryId,
+    next_id: usize,
```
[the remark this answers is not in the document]
mrferris: When I merged in the previous `QueryPool` PR I didn't realize that, since the code wasn't used elsewhere in the codebase, rustc/clippy wouldn't throw warnings/errors when things were amiss. 

Once this code became used, the compiler correctly alerted me to the fact that `QueryId` is a struct containing a `usize`, not an alias for `usize`. It makes more sense to me to have it instead be an aliased type for `usize`, and I plan to do that refactor, during which I might realize why `sigp/discv5` declared it this way.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -54,6 +59,8 @@ pub const FIND_CONTENT_MAX_NODES: usize = 32;
 /// With even distribution assumptions, 2**17 is enough to put each node (estimating 100k nodes,
 /// which is more than 10x the ethereum mainnet node count) into a unique bucket by the 17th bucket index.
 const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
+/// Timeout for FINDNODES queries in seconds.
+pub const QUERY_TIMEOUT: u64 = 5;
```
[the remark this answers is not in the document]
mrferris: Good catch & good point. I had this low for testing purposes and must have accidentally reverted my reversion of it.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -250,6 +257,12 @@ pub struct OverlayService<TContentKey, TMetric> {
     request_tx: UnboundedSender<OverlayRequest>,
     /// A map of active outgoing requests.
     active_outgoing_requests: Arc<RwLock<HashMap<OverlayRequestId, ActiveOutgoingRequest>>>,
+    /// Keeps track of which queries a given node has been requested on behalf of.
+    active_query_requests: Arc<RwLock<HashMap<NodeId, QueryId>>>,
+    /// All of the iterative queries currently being performed.
+    queries: QueryPool<QueryInfo, NodeId, Enr>,
+    /// A map of Node IDs to the most recently seen ENR for that Node ID. Todo: make this persistent.
+    node_enrs: RwLock<HashMap<NodeId, Enr>>,
```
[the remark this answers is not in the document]
mrferris: In order to get the ENR for a given Node ID from the routing table, I was originally doing the following :
```
let optional_node = match self.kbuckets.write().entry(&key) {
            kbucket::Entry::Present(ref mut entry, _) => Some(entry.value().clone()),
            kbucket::Entry::Pending(ref mut entry, _) => Some(entry.value().clone()),
            _ => None,
};
```

But I found that sometimes ENRs passed into `process_discovered_enrs` were not yet present in the routing table (the above approach would return `None`) by the time the `query_pool` needed a request to be sent to the node a few moments later. 

I determined (perhaps incorrectly) that the `kbucket` has a delay period before the entry could be accessed, and figured that because we intend to have a flat, persistent store for mapping node IDs to ENRs outside of the kbuckets (incase they are full) anyway, this would be a good time to introduce/use that mechanism (in a temporarily not-persistent fashion).

Let me know if you know of a reason for not finding the nodes in the routing table close-to-immediately after calling `process_discovered_enrs`  that I might have overlooked.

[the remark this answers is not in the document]
jacobkaufmann: I could be missing something here, but I think we resolve that issue if we update the query state at the end of `process_discovered_enrs` after all ENR updates/insertions.

If we wait until all ENR changes are made to the routing table before we update the query state, then by the time we call `on_success` for that query with the new ENRs, those ENRs are guaranteed to be in the routing table (assuming there is enough space).

Like you said, it might happen that not all of the ENRs from the query fit in the routing table. I'm not so sure that this is a problem though, at least for now. If the corresponding bucket is full, then that would mean that we already know `k` nodes who are just as close to the target as the left out ENR. Presumably, those known ENRs would be among the known closest nodes when we initiate the query. Any ENRs that don't fit in the routing table could be filtered from the query update.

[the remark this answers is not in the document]
jacobkaufmann: So looking into this a bit more, we could adopt [this method](https://github.com/sigp/discv5/blob/506c9afe6a037dc3ba920f691068dac29f6cba7c/src/service.rs#L499) of `sigp/discv5`. This checks both the routing table (for present entries) and all outstanding queries for a given `NodeId`. It checks the `untrusted_enrs` field of the `QueryInfo`, which we update when we perform the query update.

This seems far better than what I suggested above, which would limit the reach of the lookup procedure. With this solution, I don't see the need for `node_enrs`.

mrferris:
Thanks! I've:
- Addressed your individual comments in code, or in the case of a few, addressed below.
- Added endpoint for verifying population of routing table. 
- Added the self-lookup part of the routing table initialization `todo` that you pointed out. This is in the `initialize_routing_table` function. The other part of that `todo` is addressed below.

Here's are a few things I plan to defer beyond this PR for the sake of scope, and make into issues for continued work going forward:

- Automated test coverage of high level queries, [like so](https://github.com/sigp/discv5/blob/4b4688b85c0d4c6ae1f8135e5a6e27d49fc4c909/src/discv5/test.rs#L225). Requires a bit of a refactor to make this kind of test feasible. 
- After self-lookup, we refresh every bucket farther than the closest peer in ascending order, starting with the closest bucket. We currently effectively do this in random order via the bucket refreshes that begin after the initial self-lookup.
- Inheriting FindNodesQueryConfig from the high level trin configs passed in via CLI. 
- Removing nodes that become disconnected mid-query from active queries to end with fewer than k results if necessary, rather than wait until time-out.
- ENR overhead seems to be about double what I’d expect based off doing the header calculations, going to figure out why and document it in-line. Probably SSZ overhead, but double seems like a lot.
- Refactor `findnodes.rs` tests to not require pinned versions of `quickcheck` and `rand`.

jacobkaufmann (changes requested):
Took a pass and noted down what stood out. I may have missed some things, because the force push made it so that I cannot see what changed since my last review. This required me to read over all the logic. In the future, we should not force push once review has begun.

trin-core/src/portalnet/overlay.rs
```diff
@@ -187,6 +188,44 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                    *entry.node.key.preimage(),
+                    entry.node.value.enr().clone(),
+                    entry.status,
+                )
+            })
+            .collect()
+    }
+
+    /// Returns vector of all ENR node IDs of nodes currently contained in the routing table mapped to JSON Value.
+    pub fn routing_table_info(&self) -> Value {
```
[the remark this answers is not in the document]
mrferris: Good point

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -315,59 +333,86 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
 
-            info!("[{:?}] Starting overlay service", overlay_protocol);
-            service.start().await;
-        });
+    /// Begins initial FINDNODES query to populate the routing table.
+    fn initialize_routing_table(&mut self) {
+        // Perform FindNode lookup for local node's own node ID.
+        //
+        // A bug(?) in kbuckets.closest_values(key) returns 0 values if the key is our local key.
+        // Get the raw key, increment it by a few, then turn it back into node ID and start the query.
```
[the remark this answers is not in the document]
mrferris: I'm not seeing why it's not a bug, but good point below that our only nodes at initialization are the bootnodes anyway.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1812,8 +2042,71 @@ mod tests {
...
+        55 + // Header
+        1 + // Discv5 Message Type
+        3 + // RLP Encoding of outer list
+        9 + // Request ID, max 8 bytes + 1 for RLP encoding
+        3 + // RLP Encoding of inner response
+        16; // RLP HMAC
+
+    // ENR SSZ overhead empirically observed to be double.
+    // Todo: determine why this is. It seems too high.
+    const MAX_NODES_SIZE: usize = (MAX_DISCV5_PACKET_SIZE - TALK_REQ_PACKET_OVERHEAD) / 2;
```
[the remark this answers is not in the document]
jacobkaufmann: Disregard the question about max ENR size. I see that now in [the EIP](https://eips.ethereum.org/EIPS/eip-778).

trin-core/src/jsonrpc/endpoints.rs
```diff
@@ -65,6 +66,9 @@ impl FromStr for TrinEndpoint {
             "discv5_routingTableInfo" => Ok(TrinEndpoint::Discv5Endpoint(
                 Discv5Endpoint::RoutingTableInfo,
             )),
+            "portal_historyRoutingTableInfo" => Ok(TrinEndpoint::HistoryEndpoint(
+                HistoryEndpoint::RoutingTableInfo,
+            )),
```
njgheorghita, line 71: I'd go for isolating this endpoint & its logic into a separate pr and add a test for it to `ethportal-peertest`. As well as support in the state network. More of a nitpick for now, since we'll get tests for all the endpoints sooner or later. And it's probably a bit too late for this pr, but down the road easily extractable chunks of logic like this are better off in their own pr to reduce clutter.

trin-core/src/portalnet/overlay.rs
```diff
@@ -187,6 +188,44 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        let buckets: Vec<(String, String, String)> = self
+            .table_entries()
+            .iter()
+            .map(|(node_id, enr, node_status)| {
+                (
+                    node_id.to_string(),
```
njgheorghita, line 213: I've been working on a branch doing some jsonrpc cleanup. `node_id.to_string()` returns something like `0xabcd..1234`. No doubt, this is nice and readable, but I've been updating it to `format!("0x{}", hex::encode(self.discv5.local_enr().node_id().raw()))` b/c I feel like we should be returning the actual node id rather than the abbrev.

trin-core/src/portalnet/overlay.rs
```diff
@@ -187,6 +188,44 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                    enr.to_base64(),
+                    format!("{:?}", node_status.state),
+                )
+            })
+            .collect();
+
+        json!(
```
njgheorghita, line 220: Agreed w/ ^, I feel like the json conversion should be handled in `trin-history/src/jsonrpc.rs`

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -54,6 +59,12 @@ pub const FIND_CONTENT_MAX_NODES: usize = 32;
 /// With even distribution assumptions, 2**17 is enough to put each node (estimating 100k nodes,
 /// which is more than 10x the ethereum mainnet node count) into a unique bucket by the 17th bucket index.
 const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
+/// Timeout for FINDNODES queries in seconds.
+pub const QUERY_TIMEOUT: u64 = 60;
```
njgheorghita, line 63: If these two constants are specific to `FINDNODES` - I'd prefix the const name with `FIND_NODES_*`

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1812,8 +2042,71 @@ mod tests {
...
+            17 => 8,
+            25 => 8,
+            _ => 0,
+        };
+
+        assert_eq!(enrs_limited.len(), correct_limited_size);
+    }
 }
 
 fn should_store(_key: &Vec<u8>) -> bool {
```
njgheorghita, line 2075: *nitpick* - imo tests should be the final piece of code in a file, so these two functions should be moved above the tests section

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1812,8 +2042,71 @@ mod tests {
...
+        let mut enrs: Vec<SszEnr> = Vec::new();
+        for _ in 0..original_nodes_size {
+            // Generates an ENR of size 63 bytes.
+            let (_, enr) = generate_random_remote_enr();
+            enrs.push(SszEnr::new(enr));
+        }
+
+        let enrs_limited = limit_nodes_response_size(enrs);
+
+        let correct_limited_size = match original_nodes_size {
```
njgheorghita, line 2062: I think you can just add a second argument to the parameterization... [like here](https://github.com/ethereum/trin/blob/37c1931a2f2eb9e4832d980574070e8ff1a5311e/trin-core/src/utils/bootnodes.rs#L54)

njgheorghita (commented):
Just left some general thoughts for now. Would like to give this another, more in-depth review once you've addressed some of the comments. IDK how you feel about moving `portal_historyRoutingTableInfo` into another pr. It would be helpful, but I don't think it's critical tbh.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -583,9 +708,11 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             .as_ref()
             .and_then(|m| Some(m.report_inbound_find_nodes()));
         let distances64: Vec<u64> = request.distances.iter().map(|x| (*x).into()).collect();
-        let enrs = self.nodes_by_distance(distances64);
-        // `total` represents the total number of `Nodes` response messages being sent.
-        // TODO: support returning multiple messages.
+        let all_enrs = self.nodes_by_distance(distances64);
+
+        // Limit the ENRs so that their summed sizes do not surpass the max TALKREQ packet size.
+        let enrs = limit_nodes_response_size(all_enrs);
```
jacobkaufmann, line 714: nit: Shadow variable `enrs` instead of using a separate variable name.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1017,6 +1147,28 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         }
     }
 
+    /// Advances a query (if one is active for the node) using the received ENRs.
+    /// Does nothing if called with a node_id that does not have a corresponding active query request.
+    fn advance_query(&mut self, source: Enr, enrs: Vec<Enr>, query_id: QueryId) {
```
jacobkaufmann, line 1152: What do you think about changing the function signature to have a `Query` parameter instead of `QueryId`? With this change, it would not be possible to call for a query unless that query exists in the map.

jacobkaufmann (changes requested):
For me, getting close but not all the way there.

Aside from some minor things noted in comments, I would like to see the following:

- Unit tests for new methods in `OverlayService`
- Config struct to hold parameters for `FindNodeQuery`

I saw that we marked the config struct out of scope, but I don't understand why.

I'm also not sure that we should include the `portal_historyRoutingTableInfo` JSON-RPC endpoint here. It may be useful in testing that the query returns some peers and puts those peers in the routing table, but it's not sufficient to verify correct behavior. This will require some integration tests as noted above. That said, I think the change to add the endpoint belongs in a separate PR.

jacobkaufmann:
Forgot to add that we need to resolve the conflicts.

trin-core/src/portalnet/overlay_service.rs
mrferris: I tried this. Taking the query directly in the function signature required calling out the specific types of the query:
`fn advance_query(&mut self, source: Enr, enrs: Vec<Enr>, query: Query<QueryInfo, NodeID, Enr>) ` whereas the current approach allows everything to stay generic, which I find cleaner. 

I also prefer checking whether the `QueryId` maps to a valid `Query` within the query-specific function, rather than having that logic bloat the `process_nodes` function.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1924,4 +2204,190 @@ mod tests {
...
+            node_id_2
+        } else {
+            node_id_1
+        };
+        match event {
+            FindNodeQueryEvent::Waiting(_, node_id, _) => {
+                assert_eq!(node_id, second_node_id);
+            }
+            _ => panic!(),
+        };
```
jacobkaufmann, line 2345: We do not test that we get the expected result from a finished or timed out query here.

jacobkaufmann (changes requested):
Requesting changes here because I do not understand why we have removed `QueryInfo` from the finished and timeout query pool states. It seems like the lack of `QueryInfo` in those states makes certain parts of the logic incorrect. See comments for further details.

mrferris:
@jacobkaufmann You're right, removing `QueryInfo` there was an oversight on my part. Mistakenly removed those last week and forgot to add them back. Addressing now.

njgheorghita (approved):
Looks good! Mostly style nits, but I think I'll need to get my hands dirty interacting with the query pool to have a more informed idea around it's api

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1312,15 +1473,127 @@ where
...
+    3 + // RLP Encoding of inner response
+    16; // RLP HMAC
+        // ENR SSZ overhead empirically observed to be double.
+        // Todo: determine why this is. It seems too high.
+const MAX_NODES_SIZE: usize = MAX_DISCV5_PACKET_SIZE - TALK_REQ_PACKET_OVERHEAD;
+
+/// Limits a to a maximum packet size, including the discv5 header overhead.
+fn limit_enr_list_to_max_bytes(enrs: Vec<SszEnr>, max_size: usize) -> Vec<SszEnr> {
+    // If all ENRs would fit at max size, don't check individual sizes.
+    if enrs.len() < (MAX_NODES_SIZE / MAX_ENR_SIZE) {
```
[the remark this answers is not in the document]
mrferris: `MAX_PACKET_SIZE` wouldn't really be accurate here though

I recall deciding on the this mathematical expression being clearer than any english phrase I could come up with

`MAX_NUMBER_OF_MAXIMUM_SIZED_ENRS_THAT_WOULD_FIT_IN_A_MAXIMUM_SIZED_DISCV5_PACKET` 😃

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -637,8 +760,10 @@ where
             .and_then(|m| Some(m.report_inbound_find_nodes()));
         let distances64: Vec<u64> = request.distances.iter().map(|x| (*x).into()).collect();
         let enrs = self.nodes_by_distance(distances64);
-        // `total` represents the total number of `Nodes` response messages being sent.
-        // TODO: support returning multiple messages.
+
+        // Limit the ENRs so that their summed sizes do not surpass the max TALKREQ packet size.
+        let enrs = limit_enr_list_to_max_bytes(enrs, MAX_NODES_SIZE);
+
```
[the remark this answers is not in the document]
mrferris: Based off discussion in [this issue](https://github.com/ethereum/portal-network-specs/issues/136) it doesn't look like there's specs in place to do anything but return a single NODES message. Do you disagree?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -431,10 +478,53 @@ where
...
+                            for node_id in result.into_iter() {
+                                if let Some(enr) = self.find_enr(&node_id) {
+                                    found_enrs.push(enr);
+                                } else if let Some(enr) = self.discovery.discv5.find_enr(&node_id) {
+                                    found_enrs.push(enr);
+                                }
+                                else {
+                                    warn!("ENR not present in queries results.");
+                                }
+                            }
```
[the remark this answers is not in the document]
mrferris: Yea, I removed query_info in an apparent brainfart end of last week.

I think that this was still working though, because `find_enr` could still get the ENRs from the routing table.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -469,30 +559,37 @@ where
...
-            Some(bucket) => {
-                let target_bucket_idx = u8::try_from(bucket.0);
-                if let Ok(idx) = target_bucket_idx {
-                    // Randomly generate a node ID that falls within the target bucket.
-                    let target_node_id = self.generate_random_node_id(idx);
-                    // Do the random lookup on this node ID.
-                    self.send_recursive_findnode(&target_node_id);
-                } else {
-                    error!("Unable to downcast bucket index.")
+        let target_node_id = {
```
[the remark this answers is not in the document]
mrferris: I tried breaking this out into a separate function and it ended up being up a bit much because I could no longer just `return` if anything went wrong, I needed to return an error, which this function then had to handle. That seemed like a lot for a piece of functionality only used in one place.

The reason I use the nested setup in the first place is to avoid conflict with the borrow checker: I have `self` borrowed immutably, and then `init_find_nodes_query` has to borrow it immutably. So I need the immutable `self` references to go out of scope before I call `init_find_nodes_query`.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -530,14 +627,39 @@ where
...
+            QueryPoolState::Waiting(Some((query_id, query_info, query, return_peer))) => {
+                let node_id = return_peer;
+
+                let request_body = match query_info.rpc_request(return_peer) {
+                    Ok(request_body) => request_body,
+                    Err(_) => {
+                        query.on_failure(&node_id);
+                        return Poll::Pending;
+                    }
+                };
```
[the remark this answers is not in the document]
mrferris: Hmm, I thought about it and I think it's fine as-is. What do you think it should do in that case instead?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1924,4 +2204,190 @@ mod tests {
...
+    #[case(17, 8)]
+    #[case(25, 8)]
+    fn test_limit_nodes_response_size(
+        #[case] original_nodes_size: usize,
+        #[case] correct_limited_size: usize,
+    ) {
+        let mut enrs: Vec<SszEnr> = Vec::new();
+        for _ in 0..original_nodes_size {
+            // Generates an ENR of size 63 bytes.
+            let (_, enr) = generate_random_remote_enr();
```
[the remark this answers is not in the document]
mrferris: Yea, this is brittle. What do you think about asserting that the size is what we think it is? That's probably shitty too, because it breaks in the exact same case. I'll think more about this.

trin-core/src/portalnet/overlay_service.rs
mrferris: Good point, I'll add that.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
jacobkaufmann: My first thought would be to build random ENRs that contain some meaningless key-value with [`add_value`](https://docs.rs/enr/0.6.0/enr/struct.EnrBuilder.html#method.add_value), so that the ENR size is random and less than or equal to the maximum size. With that list, you would pass it to the function and check that the size of the output list is within the proper bounds. You would also check that if there were any remaining (non-included) ENRs in the input, then adding the next ENR would indeed cause the size of the output list to breach the threshold.

I feel like the logic is straightforward though, so if the above turns out to not be so simple, then we can add more robust testing later.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
jacobkaufmann: IIUC the failure should not adversely affect the query's progress. We should be able to prevent that match arm from occurring though. In `advance_query`, we do not filter out an ENR that is equal to that of the local node. We could add that to prevent this from occurring. This filtering does occur in `process_discovered_nodes`, and it may be preferable to move the `advance_query` logic into that function. That is, as long as we could apply the same function for ENRs returned in a `Content` message that may belong to a `FindContentQuery`.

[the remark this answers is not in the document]
mrferris: I added a filter in `advance_query` to make sure that this case doesn't happen. Though I kept the `Result` return type from `rpc_request` because we still need a way to denote the idea that log2 of 0 is undefined. I tried using an `Option<Request>` return instead and it wasn't any cleaner imo. 

And yea, I think that `advance_query` needs to be done outside of `process_discovered_enrs` because I think that a different function (`advance_findnode_query`/`advance_content_query`) will need to be called for advancing a `FindContent` query.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -499,32 +501,43 @@ where
...
                                 }
                                 else {
                                     warn!("ENR not present in queries results.");
                                 }
                             }
+                            if let Some(callback) = query_info.callback {
+                                match callback.send(found_enrs.clone()) {
+                                    Ok(_) => {}
+                                    Err(_) => error!("Failed to send FindNode query {} results to callback", query_id)
+                                }
```
jacobkaufmann, line 531:
```suggestion
                                if let Err(_) = callback.send(found_enrs.clone()) {
                                    error!("Failed to send FindNode query {} results to callback", query_id);
                                }
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1473,52 +1487,55 @@ where
...
     }
 
+    fn init_find_nodes_query_with_initial_enrs(&mut self, target: &NodeId, enrs: Vec<Enr>) {
+        let query_config = QueryConfig {
+            parallelism: self.query_parallelism,
+            num_results: self.query_num_results,
+            peer_timeout: self.query_peer_timeout,
+        };
+
+        self.init_find_nodes_query_with_initial_enrs_and_config(target, enrs, query_config);
```
jacobkaufmann, line 1527: It seems like we only added this latter method to use in a test, which I do not think is a good reason to add a function. I would be in favor of reverting to a single method here.

jacobkaufmann (approved):
I am sure that we will find ways to improve the query structure/logic (for both node and content lookups), but this is a solid start.

trin-core/src/portalnet/overlay_service.rs
mrferris: @jacobkaufmann Good point, I moved this function to the test module (at the cost of a bit of duplication) for now. In a subsequent PR I'm going to refactor the query config passing/defaults which will allow for having one simple method for initialization find nodes queries rather 3 separate methods for that build on each other for 3 separate cases.

trin-core/src/portalnet/overlay.rs
```diff
@@ -66,6 +71,11 @@ impl Default for OverlayConfig {
             bucket_filter: None,
             ping_queue_interval: None,
             enable_metrics: false,
+            query_parallelism: 3, // (recommended α from kademlia paper)
```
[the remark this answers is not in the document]
mrferris: agreed, i have a refactor planned for query config passing for a subsequent PR

PR #266, opened 2022-03-16
Fix cargo clippy warnings

ogenev:
### What was wrong?
`cargo clippy` shows the following warnings:

- warning: trin v0.1.0 (...) ignoring invalid dependency `cargo-release` which is missing a lib target
- warning: trin v0.1.0 (...) ignoring invalid dependency `trin-cli` which is missing a lib target

### How was it fixed?

Remove `trin-cli` and `cargo-release` package from `Cargo.toml` dependencies because those packages are not libraries.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -231,17 +231,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "771fe0050b883fcc3ea2359b1a96bcfbc090b7116eae7c3c512c7a083fdf23d3"
 
 [[package]]
-name = "bstr"
-version = "0.2.17"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
Nice! Do you have any insight into https://github.com/ethereum/trin/issues/186? Particularly `"Can we add a flag that would cause this warning to trigger a failure in CI?"`. I've played around with this trying to find something suitable - but never figured it out.

ogenev:
> Nice! Do you have any insight into #186? Particularly `"Can we add a flag that would cause this warning to trigger a failure in CI?"`. I've played around with this trying to find something suitable - but never figured it out.

I couldn't find a solution either. I posted a [question](https://users.rust-lang.org/t/deny-invalid-dependency-warnings/73093) in the Rust community forum.

PR #270, opened 2022-03-18
Implement FindContent <-> Content uTP stream

ogenev:
### What was wrong?
FindContent/Content uTP scenario was not implemented.

### How was it fixed?
- Implemented uTP stream for FindContent/Content.
- Some unnecessary uTP logs where deleted.

The current implementation is manually tested with Fluffy in both directions.

### To-Do
Those should be added in later PRs, as I would like to test Offer/Accept uTP next:

- write uTP bi-directional test (when we want to send DATA packet right after SYN-ACK).
- Return content data instead of connection-id  to json-rpc client when we send FIndContent request and uTP stream is initialized.

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -15,7 +15,7 @@ use crate::portalnet::types::messages::{
     ProtocolId, Request, Response,
 };
 
-use crate::utp::stream::{ConnectionKey, UtpListener};
+use crate::utp::stream::{ConnectionKey, UtpListener, BUF_SIZE};
 use crate::utp::trin_helpers::{UtpAccept, UtpMessage, UtpMessageId};
 use discv5::{
     enr::NodeId,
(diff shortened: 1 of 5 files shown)


trin-core/src/utp/stream.rs
```diff
@@ -430,7 +466,7 @@ impl UtpSocket {
 
     pub async fn raw_receive(&mut self) -> anyhow::Result<Option<Packet>> {
         // Listen on a channel for discovery utp packet
-        match timeout(Duration::from_millis(15), self.rx.write().await.recv()).await {
+        match timeout(Duration::from_millis(25), self.rx.write().await.recv()).await {
```
njgheorghita, line 469: This value could probably be pulled out into a constant

trin-core/src/utp/stream.rs
```diff
@@ -479,20 +514,16 @@ impl UtpSocket {
...
         }
 
-        debug!(
-            "out: now_microseconds() - now = {}",
-            now_microseconds() - now
-        );
+        // debug!(
+        //     "out: now_microseconds() - now = {}",
+        //     now_microseconds() - now
+        // );
```
njgheorghita, line 526: If this message is no longer needed, let's just delete it

njgheorghita (approved):
Super cool to see this coming together and the compatibility with Fluffy! Looks good to me, but agree that more testing would be great. I have a couple todo's in ethportal-peertest, but once those are done, let's chat about testing some utp functionality in there

trin-core/src/portalnet/overlay.rs
```diff
@@ -236,14 +236,54 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
...
+                            .await
+                            .connect(conn_id, enr.node_id())
+                            .await;
+
+                        match conn {
+                            Ok(mut conn) => {
+                                // TODO: Implement this with recv_from
+                                // Handle STATE packet for SYN
+                                let mut buf = [0; 1500];
+                                conn.recv(&mut buf).await.unwrap();
```
jacobkaufmann, line 274: Is this a safe unwrap?

trin-core/src/utp/stream.rs
```diff
@@ -1004,7 +1034,13 @@ impl UtpSocket {
                     }
                     break;
                 }
-                Err(msg) => debug!("{msg}"),
+                Err(_) => {
+                    recv_retries += 1;
+
+                    if recv_retries > 5 {
```
jacobkaufmann, line 1040: Should we use a const like `MAX_RETRANSMISSION_RETRIES` here?

jacobkaufmann (approved):
Looks good! Agree with Nick on testing. Hard to validate certain functionality without knowledge that things have been tested manually or without a solid understanding of the uTP logic. We can iterate and improve from here though. Left some minor comments.

trin-core/src/portalnet/overlay.rs
```diff
@@ -236,14 +236,54 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
...
+                            .write()
+                            .await
+                            .connect(conn_id, enr.node_id())
+                            .await;
+
+                        match conn {
+                            Ok(mut conn) => {
+                                // TODO: Implement this with recv_from
+                                // Handle STATE packet for SYN
+                                let mut buf = [0; 1500];
```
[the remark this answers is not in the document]
ogenev: Yeah, there is a `BUF_SIZE` constant, I forgot to use it :)

trin-core/src/portalnet/overlay.rs
```diff
@@ -236,14 +236,54 @@ impl<TContentKey: OverlayContentKey + Send> OverlayProtocol<TContentKey> {
...
+                                conn.recv(&mut buf).await.unwrap();
+
+                                // TODO: Wait for uTP transfer and return the data
+                                Ok(found_content)
+                            }
+                            Err(msg) => {
+                                warn!("Unable to initiate uTP stream with remote node {msg}");
```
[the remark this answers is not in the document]
ogenev: Is there any reason we don't want to use the newest string formating syntax?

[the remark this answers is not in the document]
jacobkaufmann: Ah I see now. Apologies! I will resolve all such comments.

PR #271, opened 2022-03-22
[WIP] Add portal_historyFindNodes test to ethportal-peertest

njgheorghita:
### To-Do

- [ ] Clean up commit history


ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,75 @@ pub mod cli;
...
-pub async fn launch_buddy_node() -> (String, Arc<JsonRpcExiter>) {
+pub struct PeertestNode {
+    pub enr: String,
+    pub exiter: Arc<JsonRpcExiter>,
+}
+
+pub struct AllPeertestNodes {
+    pub bootnode: PeertestNode,
+    pub nodes: Vec<PeertestNode>,
+}
```
jacobkaufmann, line 22: The name `AllPeertestNodes` sort of implies that all the nodes in the struct are homogeneous, and it does not convey that the boot node is distinct. I think it would be more clear if we used a different name, or if we removed the struct and handled the boot node separately from the other nodes.

ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,75 @@ pub mod cli;
...
+        self.bootnode.exiter.exit();
+        self.nodes.iter().for_each(|node| node.exiter.exit());
+    }
+}
+
+pub async fn launch_node(mut id: u8, bootnode: Option<&String>) -> anyhow::Result<PeertestNode> {
+    // The bootnode's id is 1, the other node ids are subsequently incremented by one
+    if bootnode.is_some() {
+        id += 1;
+    }
```
jacobkaufmann, line 35: I think we should enforce this ID logic outside of the function.

ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,75 @@ pub mod cli;
...
     // Run a client, as a buddy peer for ping tests, etc.
-    let trin_config = TrinConfig::new_from(
-        [
+    let discovery_port = format!("900{id}");
```
jacobkaufmann, line 37: I don't know that we would ever have 10 or more nodes, but that would create an invalid port. We ought to do something like

```rust
let discovery_port: u16 = 9000 + id;
```

trin-core/src/jsonrpc/types.rs
```diff
@@ -147,13 +147,13 @@ impl TryFrom<[&Value; 2]> for FindNodesParams {
...
+        let distances: &Vec<Value> = params[1]
+            .as_array()
             .ok_or_else(|| ValidationError::new("Empty distances param"))?;
-        let distances: Vec<u16> = match serde_json::from_str(distances) {
-            Ok(val) => val,
-            Err(_) => return Err(ValidationError::new("Unable to decode distances")),
-        };
+        let distances: Vec<u16> = distances
+            .iter()
+            .map(|val| val.as_u64().unwrap().try_into().unwrap())
```
jacobkaufmann, line 155: We should handle these unwraps so that we don't unnecessarily panic.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -67,13 +80,25 @@ fn validate_endpoint_response(method: &str, result: &Value) {
...
+                result.get("total").unwrap().as_u64().unwrap(),
+                peertest_nodes.nodes.len() as u64
+            );
+            assert!(result
+                .get("enrs")
+                .unwrap()
+                .as_str()
+                .unwrap()
+                .contains("enr:"));
+        }
```
jacobkaufmann, line 94: A better assertion would be to use the `FromStr` implementation of `Enr` to validate each ENR in the response.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -67,13 +80,25 @@ fn validate_endpoint_response(method: &str, result: &Value) {
                 ENR_SEQ.to_string()
             );
         }
+        "portal_stateFindNodes" => {
+            assert_eq!(
+                result.get("total").unwrap().as_u64().unwrap(),
+                peertest_nodes.nodes.len() as u64
+            );
```
jacobkaufmann, line 87: Wouldn't the total be equal to the number of `Nodes` messages, and not the number of nodes launched by the test?

jacobkaufmann (commented):
Happy to see these tests make their way into `ethportal-peertest`!

I am sort of skeptical as to whether this works with more than a single peer under test that is not a boot node. I see that we added functionality to launch more than a single non-boot-node peer, but we hardcode it so that we only ever launch a single non-boot-node peer, so the rationale for that is a little unclear.

I don't know that we need that functionality, and it would make validation of responses somewhat more complicated if interactions with the boot node from multiple peers make the responses from the boot node inconsistent based on ordering. I would like to know what you had in mind for that.

ethportal-peertest/src/jsonrpc.rs
njgheorghita: Ahh, good catch, I misinterpreted the spec there

ethportal-peertest/src/lib.rs
njgheorghita: Any suggestions for a different name? `AllTestnetNodes` (but that's pretty much the same name). I threw them together in a single struct, since it's easier to pass around a single struct rather than a bootnode and the other nodes separately. Imo, there really is not a big difference between a bootnode and a regular node, they are mostly homogeneous.

njgheorghita:
@jacobkaufmann Ready for another :eyes. It now launches 5 peertest nodes. With this setup, I was just trying to add a little flexibility to the setup. As the test suite grows, my assumption is that direct access to `AllPeertestNodes` (or whatever we end up calling it) will be useful inside `validate_endpoint_response()` (even though we're not directly using it right now). I'm not sure ordering of the enrs in the response is a huge concern right now... though as you can see, we only test that the `Nodes` response `"enrs"` params is `> 1` since I have been getting some flaky results on that locally. Imo, it's better than what we had and good enough for now, but should be revisited in the future after tests for all the endpoints are introduced - which imo is the priority right now. Open to any thoughts on this though

ethportal-peertest/src/lib.rs
jacobkaufmann: I think we should remove the struct altogether. It doesn't seem so useful.

The function `validate_endpoint_response` does not even use the argument.

The function `all_endpoints` only accesses the `bootnode` field of the argument.

If we need the two in a single type, like the return type for `launch_peertest_nodes`, then we can use a tuple struct.

jacobkaufmann (commented):
I'm sort of having second thoughts about more than a single peer in `ethportal-peertest`. I'm not sure how much utility there is in having multiple nodes to connect with the boot-node, especially if it means we have to weaken our tests (e.g. validate non-empty ENRs vs validate a single, specific ENR).

For now, multiple peers only complicates things and doesn't validate any additional functionality. It may also make it more difficult to add endpoints to `ethportal-peertest`.

I think we can revisit support for multiple peers and declare that functionality out of scope for this PR, given that the focus here is on support for `FindNodes` in `ethportal-peertest`.

What are your thoughts?

trin-core/src/jsonrpc/types.rs
```diff
@@ -150,14 +150,55 @@ impl TryFrom<[&Value; 2]> for FindNodesParams {
...
+            distances?.into_iter().map(|val| val.try_into()).collect();
+        match distances {
+            Ok(val) => Ok(Self {
+                enr,
+                distances: val,
+            }),
+            Err(_) => Err(Self::Error::new("Invalid distances param.")),
+        }
+    }
+}
```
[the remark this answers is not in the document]
njgheorghita: The previous decoding function was broken for an array of distances. Using `as_str()` decodes the param into a `Value::String`. So we first have to decode it into an array, then the only available option `serde_json` gives us for decoding integers is `as_u64()` and then we have to convert those into `u16`. Basically this was broken before and now it works, one of the nice things about having tests haha

ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,85 @@ pub mod cli;
...
+
+fn get_peertest_id_for_node(mut id: u8, bootnode_enr: Option<&String>) -> u16 {
+    // Peertest Id for bootnode is 1 (identified by bootnode_enr == None)
+    // For all other nodes (identified by bootnode_enr == Some()) Peertest ID needs to be
+    // incremented by one to account for the bootnode
+    if bootnode_enr.is_some() {
+        id += 1;
+    }
+    id as u16
+}
```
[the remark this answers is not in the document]
njgheorghita: > I think we should enforce this ID logic outside of the function.

Maybe I'm misunderstanding you? I thought you preferred the id generation to be pulled out of `launch_node` into it's own function?

ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,85 @@ pub mod cli;
...
+
+pub async fn launch_node(id: u8, bootnode_enr: Option<&String>) -> anyhow::Result<PeertestNode> {
+    let id = get_peertest_id_for_node(id, bootnode_enr);
+
     // Run a client, as a buddy peer for ping tests, etc.
-    let trin_config = TrinConfig::new_from(
-        [
+    let discovery_port: u16 = 9000 + id;
```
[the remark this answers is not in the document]
njgheorghita: So the `id` here serves two purposes. 1. As the suffix on a base `9000` port. 2. As the suffix for the ipc-path (eg. `/tmp/ethportal-peertest-buddy-{id}.ipc`). I guess we could use `AtomicUsize` to generate this id / offset... but I'm not sure it's necessary? Maybe, can you just explain why you think this will be better?

ethportal-peertest/src/lib.rs
njgheorghita: Fair enough, I was building with patterns in mind whose application were beyond this actual pr. So while they weren't immediately useful, I was building around the idea that they will be useful soon. Well, I've updated the general structure of our tests to exhibit why something like `AllTestnetNodes` is actually useful (imo).

The main problem with our testing architecture is that it was setup in such a way that you can only execute a single test for each jsonrpc endpoint. This is obviously not sustainable, as we will want to support multiple tests for a single endpoint. Case in point, we now test `portal_*FindNodes` with distances (`[0]` -> returns the peer nodes enr & `[256]` -> returns enrs from the peer nodes routing table). 

So another round of :eyes: on the new testing structure would be helpful, there might be a simpler way to set this up, but the blocker I ran into was that the compiler treats every closure as an independent type, so the only way to have them all pass the compiler check as similar types is to wrap the validation logic into a function. @carver If you have any thoughts on the new setup, I'd also be curious to hear what you think since you've spent some time in this part of the codebase.

Also, once we start including `FindContent`/`Content` tests, I believe this struct will also prove useful for storing test data. Now, maybe the problem is with the name? Something like `PeertestConfigData` might be more appropriate?

> If we need the two in a single type, like the return type for launch_peertest_nodes, then we can use a tuple struct.

I always prefer named structs over unnamed tuples. It's minimal overhead, it's easier to read, and you don't have to keep track of what type is in which index. For me, a rule of thumb, is avoid destructuring a tuple with indexes at all costs... 
```rust
let things = Things{ one_thing, another_thing};
...
do_something(things.one_thing);
```
is much better than
```rust
let things = (one_thing, another_thing);
...
do_something(things[0])
```

ethportal-peertest/src/lib.rs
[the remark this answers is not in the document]
jacobkaufmann: Outside of `launch_node` yes, but ID generation logic is essentially a for loop that does not merit its own function. I really think that we are unnecessarily complicating things here.

Current logic:

```rust
let bootnode = launch_node(1, None).await.unwrap();
let bootnode_enr = bootnode.enr.to_base64();
let nodes = future::try_join_all(
  (1..count)
     .into_iter()
     .map(|id| launch_node(id, Some(&bootnode_enr))),
)
.await
.unwrap();
```

Proposed logic (with removal of `get_peertest_id_for_node`):

```rust
let bootnode = launch_node(1, None).await.unwrap();
let bootnode_enr = bootnode.enr.to_base64();
let nodes = future::try_join_all(
  (2..count+1)
     .into_iter()
     .map(|id| launch_node(id, Some(&bootnode_enr))),
)
.await
.unwrap();
```

ethportal-peertest/src/lib.rs
jacobkaufmann: I didn't mean to imply that that sort of structure would never be useful, but a change to the testing structure belongs in a separate PR. The change in structure looks beneficial, but it's harder to evaluate when those changes are not isolated. The focus here is just on `FindNodes`.

I did not like the name, but that was a minor issue. The main issue was that the struct did not provide much utility. With this new structure, it seems like the struct would be useful, but like I said above, I think we should address that in a separate PR.

For the proposed use of the tuple struct, where we return a tuple from the function, we would not lose clarity in the way that your example describes. It would look like the following:

```rust
let (bootnode, other_nodes) = launch_peertest_nodes(2);
```

njgheorghita:
> but a change to the testing structure belongs in a separate PR. The change in structure looks beneficial, but it's harder to evaluate when those changes are not isolated. The focus here is just on FindNodes.

I'd agree, though I had to remove one of the FindNodes tests (the one with 0 distance) since without those changes we can't have multiple tests for a single endpoint. But i'll add that back in with the next pr that addresses the refactor. 

> The main issue was that the struct (AllPeertestNodes) did not provide much utility.

I slightly disagree, it was somewhat useful to group together the logic for exiting the bootnode and peertest nodes logic, which is used in multiple places. But sure, it's not the end of the world to just write that logic out where it's needed. Anyways, let's revisit this and see what it looks like in the upcoming refactor pr. 

Cool, I just rolled back all of the structural refactor changes, and addressed the feedback. I'll follow up with the refactor logic in a subsequent pr so it's easier to review. This pr is ready for a final :eyes:

trin-core/src/jsonrpc/types.rs
```diff
@@ -150,14 +150,54 @@ impl TryFrom<[&Value; 2]> for FindNodesParams {
...
+            .ok_or_else(|| ValidationError::new("Invalid total param"))? as u8;
+
+        let enrs: &Vec<Value> = value
+            .get("enrs")
+            .ok_or_else(|| ValidationError::new("Missing enrs param"))?
+            .as_array()
+            .ok_or_else(|| ValidationError::new("Empty enrs param"))?;
+        let enrs: Vec<SszEnr> = enrs
+            .iter()
+            .map(|val| SszEnr::try_from(val).unwrap())
```
jacobkaufmann, line 197: Let's handle this `unwrap` so that a bad request cannot make us panic.

jacobkaufmann (commented):
Appreciate you separating out the structural changes! And I did not mean to imply that the grouping was useless. I do think it will be easier to evaluate its utility in the context of a separate PR though.

Just one minor comment that we need to address and then I'm happy to approve!

njgheorghita:
@jacobkaufmann No problem, made the final change, lmk if I missed something else

jacobkaufmann (approved):

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -221,40 +258,26 @@ pub async fn test_jsonrpc_endpoints_over_ipc(peertest_config: PeertestConfig, bu
             // break out of loop here since EOF is not sent, and loop will hang
             break;
         }
+        // Short sleep to make sure all peertest nodes can connect
+        thread::sleep(time::Duration::from_secs(1));
```
carver, line 262: Sleeps stand out as a red flag in tests. One second is pretty long to set as a minimum for all of these tests. Also, it seems likely to be a source of flakiness in the future.

Ideal would be to have some kind of trigger that allows the test to continue after nodes connect. Otherwise, it can be reasonable to poll for results and retry up to a maximum test length. Much better to have a single test take 3 seconds before failing than to have every passing test take 1 second.

ethportal-peertest/src/lib.rs
```diff
@@ -2,31 +2,67 @@ pub mod cli;
...
+pub async fn launch_node(id: u16, bootnode_enr: Option<&String>) -> anyhow::Result<PeertestNode> {
+    let discovery_port: u16 = 9000 + id;
+    let discovery_port: String = discovery_port.to_string();
+    let web3_ipc_path = format!("/tmp/ethportal-peertest-buddy-{id}.ipc");
+    let trin_config_args: Vec<&str> = match bootnode_enr {
+        Some(enr) => vec![
             "trin",
             "--internal-ip",
-            "--unsafe-private-key",
-            "7261696e626f7773756e69636f726e737261696e626f7773756e69636f726e73",
```
carver, line 19: Ah, I see that this happened to support multiple concurrent peers. I thought we used the consistency of the enode in some test, so I'm just curious how that got resolved... Now I'm thinking that maybe there was a plan to test the enode of the test node and we never actually wrote that test.

PR #275, opened 2022-03-25
Refactor OFFER/ACCEPT uTP stream and add `portal_*Offer` json-rpc endpoint

ogenev:
### What was wrong?
-  Offer/Accept implementation was not tested.
- `portal_*Offer` json-rpc endpoint was not implemented.

### How was it fixed?
- Manually tested Offer/Accept uTP stream in both directions with Fluffy and fixed all known issues.
- Implemented `portal_*Offer` json-rpc endpoint.

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -15,7 +15,7 @@ use crate::portalnet::types::messages::{
     ProtocolId, Request, Response,
 };
 
-use crate::utp::stream::{ConnectionKey, UtpListener, BUF_SIZE};
+use crate::utp::stream::{UtpListener, BUF_SIZE};
 use crate::utp::trin_helpers::{UtpAccept, UtpMessage, UtpMessageId};
 use discv5::{
     enr::NodeId,
@@ -283,15 +283,15 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
                                         // TODO: Wait for uTP transfer and return the data
                                         Ok(found_content)
                                     }
-                                    Err(msg) => {
-                                        warn!("Unable to handle SYN-ACK packet: {msg}");
-                                        Err(OverlayRequestError::UtpError(msg.to_string()))
+                                    Err(err) => {
+                                        warn!("Unable to handle SYN-ACK packet: {err}");
+                                        Err(OverlayRequestError::UtpError(err.to_string()))
                                     }
                                 }
                             }
-                            Err(msg) => {
-                                warn!("Unable to initiate uTP stream with remote node {msg}");
-                                Err(OverlayRequestError::UtpError(msg.to_string()))
+                            Err(err) => {
+                                warn!("Unable to initiate uTP stream with remote node {err}");
+                                Err(OverlayRequestError::UtpError(err.to_string()))
                             }
                         }
                     }
@@ -324,7 +324,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             .collect();
         let content_keys_offered: Vec<TContentKey> = match content_keys_offered {
             Ok(val) => val,
-            Err(_msg) => return Err(OverlayRequestError::DecodeError),
+            Err(_) => return Err(OverlayRequestError::DecodeError),
         };
 
         // Send the request and wait on the response.
@@ -338,7 +338,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
                     .await
                 {
                     Ok(accept) => Ok(accept),
-                    Err(msg) => Err(OverlayRequestError::OfferError(msg.to_string())),
+                    Err(err) => Err(OverlayRequestError::OfferError(err.to_string())),
                 }
             }
             Ok(_) => Err(OverlayRequestError::InvalidResponse),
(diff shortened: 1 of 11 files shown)


trin-state/src/jsonrpc.rs
```diff
@@ -81,6 +82,26 @@ impl StateRequestHandler {
                     };
                     let _ = request.resp.send(response);
                 }
+                StateEndpoint::Offer => {
+                    let response = match OfferParams::try_from(request.params) {
+                        Ok(val) => {
+                            let content_keys =
+                                val.content_keys.iter().map(|x| x.to_vec()).collect();
```
njgheorghita, line 89: *nitpick* but i'd prefer using something like `key` rather than `x` here and in `trin-history`

njgheorghita (approved):
Lookin good!

trin-core/src/utp/stream.rs
```diff
@@ -274,20 +274,14 @@ impl UtpListener {
...
-                                        debug!("Store {key:?}, {content:?}");
-                                    }
-                                }
-                                Err(_) => debug!("Recv malformed data on handing UtpAccept"),
-                            }
+                        UtpMessageId::AcceptStream(content_keys) => {
+                            // TODO: Implement this with overlay store and decode receiver stream if multiple content values are send
+                            debug!("Store {content_keys:?}, {received_stream:?}");
+                        }
+                        UtpMessageId::FindContentData(_content) => {
```
jacobkaufmann, line 281: Wouldn't content data via uTP stream correspond to `UtpMessageId::FindContentStream`? Maybe I misunderstand the difference between the two variants.

trin-core/src/portalnet/overlay.rs
```diff
@@ -398,30 +403,18 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             message: content_items,
         };
 
-        let utp_listener = self.utp_listener.clone();
         tokio::spawn(async move {
-            if let Some(conn) = utp_listener
-                .write()
+            // send the content to the acceptor over a uTP stream
+            if let Err(msg) = conn
```
jacobkaufmann, line 408: nit: I think `err` is a more clear variable name here and in other `Err` matches.

trin-core/src/portalnet/types/messages.rs
```diff
@@ -491,6 +491,13 @@ pub struct Accept {
     pub content_keys: BitList<typenum::U8>,
 }
 
+impl Into<Value> for Accept {
+    fn into(self) -> Value {
+        // TODO: self.content_keys returns a number of set bits to the json-rpc client, probably we would like to return a list of 0 and 1 bits
+        serde_json::json!({ "connection_id": format!("{:?}", self.connection_id.to_be()) , "content_keys": self.content_keys.num_set_bits()})
```
jacobkaufmann, line 497: I think we can go ahead and return the hex-string for `content_keys`, which seems to be the type in the initial JSON-RPC spec.

trin-core/src/utp/stream.rs
```diff
@@ -717,7 +711,7 @@ impl UtpSocket {
             // When SYN is send and we receive STATE, do not reply
             (SocketState::SynSent, PacketType::State) => {
                 self.connected_to = src;
-                self.ack_nr = packet.seq_nr();
+                self.ack_nr = packet.seq_nr() - 1;
```
jacobkaufmann, line 714: It may be better to address this deviation from the BEP29 spec in a separate, tiny PR as it seems unrelated to the rest. I don't feel too strongly about this though.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -661,6 +661,9 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             .map_err(|e| OverlayRequestError::AcceptError(e))?;
         let connection_id: u16 = crate::utp::stream::rand();
 
+        // TODO: Pipe this with overlay DB and request only not available keys.
+        let accept_keys = request.content_keys.clone();
```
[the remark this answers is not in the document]
ogenev: I will create an issue for this because there is a `should_store` dummy method that is used a few lines below which needs to be addressed too.

trin-core/src/utp/stream.rs
ogenev: This is kinda tricky because when we respond with the `Content` message with `connection-id`,  we want to keep somewhere the actual content data that we want to stream when we receive `SYN` packet from the requester (the node which sent `FindContent` rpc). 

So right now we keep this data in memory as part of the `UtpMessageId` as `FindContentData(Content)` and we store it on the `send_send_id` channel only, we keep `FindContentStream` for `conn_recv_id`: https://github.com/ethereum/trin/blob/7f66b18221c1124a7e386c193feed5a8a132a0fd/trin-core/src/portalnet/overlay_service.rs#L626

trin-core/src/utp/stream.rs
ogenev: Sending OFFER messages and init uTP stream to Fluffy breaks if we don't address it here :) I'll create an issue to not forget to add uTP tests that catch those deviations.

PR #277, opened 2022-03-26
Implement `--ephemeral` CLI flag

### LINKED ISSUE #81: Implement --ephemeral flag

hmrtn:
### What was wrong?

This Fixes #81, implementing the `--ephemeral` CLI flag.
### How was it fixed?

This was fixed by creating a mutable singleton vector that stores the session's data path on startup and utilizes it in `utils::db::cleanup_data_dir`, which deletes the ephemeral directories on signal interruption (if the `--ephemeral` CLI flag is enabled).

Trin's signal interruption is already being handled in the `service.rs` via `setup_ipc_cleanup_handlers()`, and can only be called once. In order to minimize the amount of changes, a call to `cleanup_data_dir()` has been added here.  

### To-Do

- [x] Review thread safety and `lock()` usage with mutable global variables

### DIFF
--- a/trin-core/src/utils/db.rs
+++ b/trin-core/src/utils/db.rs
@@ -1,9 +1,15 @@
 use std::path::Path;
 use std::{env, fs};
 
+use crate::cli::TrinConfig;
 use directories::ProjectDirs;
 use discv5::enr::NodeId;
+use log::debug;
+use std::sync::Mutex;
 
+lazy_static! {
+    static ref TRIN_NODE_ID_DATA_DIRS: Mutex<Vec<String>> = Mutex::new(vec![]);
+}
 const TRIN_DATA_ENV_VAR: &str = "TRIN_DATA_PATH";
 
 pub fn get_data_dir(node_id: NodeId) -> String {
@@ -21,6 +27,10 @@ pub fn get_data_dir(node_id: NodeId) -> String {
         .to_str()
         .unwrap()
         .to_string();
+    TRIN_NODE_ID_DATA_DIRS
+        .lock()
+        .unwrap()
+        .push(node_id_data_dir.to_owned());
     fs::create_dir_all(&node_id_data_dir).expect("Unable to create data directory folder");
     node_id_data_dir
 }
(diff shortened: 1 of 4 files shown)


trin-core/src/utils/db.rs
```diff
@@ -34,3 +44,15 @@ pub fn get_default_data_dir() -> String {
         None => panic!("Unable to find data directory"),
     }
 }
+
+pub fn cleanup_data_dir() {
+    let trin_config = TrinConfig::from_cli();
+    if trin_config.ephemeral {
```
njgheorghita, line 50: It seems like this current design assumes that we'll end up with duplicate data dirs inside `TRIN_NODE_ID_DATA_DIRS`, since every time we call `get_data_dir` the path gets added. Which is going to happen more than once for a single node id. Though, we also want to support removing multiple, different data dirs since currently the state network [creates a random one for triedb to use](https://github.com/ethereum/trin/blob/4e0a885b2c7705bee00cd891d81904369303e149/trin-state/src/network.rs#L38). However, this isn't what we want longterm since eventually triedb should use the same db as `PortalStorage`. 

I'm not a fan of failing silently here, but I do agree that it's maybe the best solution for right now, until we merge `TrieDb` with `PortalStorage`'s db. I'd suggest adding a `todo` comment here explaining that once we only use a single data dir, that we should update `TRIN_NODE_ID_DATA_DIRS` to just be a single string and to actually panic if there's an error with removing that dir.

njgheorghita (approved):
Looks good!

trin-core/src/utils/db.rs
hmrtn: 👍

ogenev (changes requested):
Setting the cleanup handler inside json-rpc IPC exiter will not trigger if we use HTTP json-rpc transport. 
For me, it makes more sense to create a general tokio exit handler function and set the cleanup handler right after we receive ctrl-c signal in tokio as described in this example: https://docs.rs/tokio/latest/tokio/signal/fn.ctrl_c.html#examples.
```rust
async fn handle_tokio_exit() {
    tokio::signal::ctrl_c()
            .await

    cleanup_data_dir();
}
```
Then we can use this function in `src/main.rs`, `trin-history/src/main.rs` and `trin-state/src/main.rs` to wait for ctr-c and cleanup the directories.

ogenev:
Going to close this PR as it is superseded by https://github.com/ethereum/trin/pull/364 because we are quite in a rush to implement this feature. 

Thank you @hmrtn for the initial work!

PR #278, opened 2022-03-28
Bugfix data dir setup

njgheorghita:
### What was wrong?
@hmrtn pointed out that  `get_data_dir` was incorrectly namespaced under `/rocksdb` - when it should be returning simply the `/trin_xxxxx` directory.

### How was it fixed?
Moved the `/rocksdb` namespacing to its appropriate location.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/utils/db.rs
+++ b/trin-core/src/utils/db.rs
@@ -23,7 +23,6 @@ pub fn get_default_data_dir(node_id: NodeId) -> String {
     let node_id_string = hex::encode(node_id.raw());
     let suffix = &node_id_string[..8];
     application_string.push_str(suffix);
-    application_string.push_str("/rocksdb");
 
     match ProjectDirs::from("", "", &application_string) {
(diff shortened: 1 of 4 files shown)


trin-core/src/utils/db.rs
```diff
@@ -23,16 +23,18 @@ pub fn get_default_data_dir(node_id: NodeId) -> String {
...
-    application_string.push_str("/rocksdb");
 
     match ProjectDirs::from("", "", &application_string) {
         Some(proj_dirs) => proj_dirs.data_local_dir().to_str().unwrap().to_string(),
         None => panic!("Unable to find data directory"),
     }
 }
 
+/// Used to setup a database for TrieDB in State Network
 pub fn setup_overlay_db(node_id: NodeId) -> DB {
```
jacobkaufmann, line 34: I don't think we need this function at all. It's almost identical to `setup_rocksdb`. I'm okay if we defer clean up to a subsequent PR though.

jacobkaufmann (approved):
LGTM

PR #280, opened 2022-03-29
Improve unit tests for PortalStorage

### LINKED ISSUE #264: Improve unit tests for "find farthest" method in `PortalStorage`
The existing tests for the `find_farthest_content_id` method of `PortalStorage` are sort of brittle in a couple of ways:

- We do not test the case where the DB is empty.
- We only perform a single test.
- We do not test for consistency between `find_farthest_content_id` and `distance_to_content_id`.

We can address these concerns by:

- Implement the test case where the DB is empty.
- Add fuzzing with `quickcheck` or other to test multiple cases where assertions are performed against `distance_to_content_id`.

For example: 

```rust
let expected_farthest = if distance_to_content_id(x) > distance_to_content_id(y) {
    x
} else {
    y
};
```

Instead of a binary comparison, we would most likely want to perform a sort on a `Vec` according to this comparison.

jacobkaufmann:
### What was wrong?

#264

### How was it fixed?

Resolves #264.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -497,9 +497,11 @@ pub mod test {
     use super::*;
     use crate::portalnet::types::content_key::IdentityContentKey;
 
+    use std::env;
+
+    use quickcheck::{quickcheck, Arbitrary, Gen, TestResult};
     use rand::RngCore;
     use serial_test::serial;
-    use std::env;
     use tempdir::TempDir;
 
     fn generate_random_content_key() -> IdentityContentKey {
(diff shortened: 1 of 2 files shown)


trin-core/src/portalnet/storage.rs
```diff
@@ -649,28 +661,54 @@ pub mod test {
...
-        let result = storage.find_farthest_content_id()?;
-        assert_eq!(result.unwrap(), expected_content_id);
+            temp_dir.close().unwrap();
 
-        temp_dir.close()?;
-        Ok(())
+            TestResult::from_bool(farthest.unwrap().unwrap() == expected_farthest)
+        }
+
+        quickcheck(prop as fn(IdentityContentKey, IdentityContentKey) -> TestResult);
```
ogenev, line 712: Can we use the [quickcheck! ](https://docs.rs/quickcheck/latest/quickcheck/macro.quickcheck.html)macro and simplify this?

ogenev (approved):
LGTM :+1:

trin-core/src/portalnet/storage.rs
jacobkaufmann: With the macro I'm having some issues with the derivable traits above this test function, so I think we can leave it for now. I'm also not sure that it makes the test meaningfully simpler.

PR #282, opened 2022-03-30
Add support for portal_historyStore & portal_stateStore

njgheorghita:
### What was wrong?

Added support for inserting content keys & content directly into an overlay database.

### How was it fixed?

Implemented endpoints & added tests to `ethportal-peertest`
This requires wrapping `PortalStorage` in a `RwLock` to be able to write to `PortableStorage` from different threads. 

### To-Do

- [x] Clean up commit history


trin-state/src/jsonrpc.rs
```diff
@@ -45,6 +53,30 @@ impl StateRequestHandler {
...
+                            Ok(params) => {
+                                let content_key = params.content_key.clone();
+                                let content = params.content.clone();
+                                match self
+                                    .network
+                                    .overlay
+                                    .storage
+                                    .lock()
+                                    .unwrap()
+                                    .store(&content_key, &content)
```
jacobkaufmann, line 68: Related to my comment above. First, we do not check whether the content key corresponds to data that falls within the local node's radius. Second, we don't do any validation on the content that we are requested to store for a particular content key. This is a concern because some peer could request that we store junk data for some content key, and if some other peer later offers us the valid data, then we would reject the offer because we already store the junk data.

jacobkaufmann (commented):
So I went through and left some comments, but then I checked the [PR for the initial JSON-RPC spec](https://github.com/ethereum/portal-network-specs/pull/88/files), and I did not see that these "store" endpoints exist in the spec.

It's hard to evaluate the correctness of the implementation when there is no spec to reference.

Is the intent for this to be a permanent JSON-RPC endpoint, or a temporary endpoint to facilitate development and testing? It's not clear from the PR description.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -236,7 +236,7 @@ pub struct OverlayService<TContentKey, TMetric> {
     /// The underlying Discovery v5 protocol.
     discovery: Arc<Discovery>,
     /// The content database of the local node.
-    storage: Arc<PortalStorage>,
+    storage: Arc<Mutex<PortalStorage>>,
```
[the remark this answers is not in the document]
njgheorghita: Using the `RwLock` introduced a fair amount of `async` code, which bubbled up a fair amount leading to more changes than I thought was worth it, so I went with the `Mutex` to keep things sync. But, if you feel like the `RwLock` is the right mechanism to use here, I can update this pr so we can evaluate how significant those changes are.

njgheorghita:
> Is the intent for this to be a permanent JSON-RPC endpoint, or a temporary endpoint to facilitate development and testing? It's not clear from the PR description.

Yup, it's mainly a util for us (the devs) to use. Similar to the `portal_historyLocalContent`/`portal_stateLocalContent` endpoints we support, none of these are in the official spec. Imo, this is a useful endpoint for 1) seeding specific content into a node that we can use for testing `FindContent`/`Content` messages with. 2) We could update the `seed_database` script to use this endpoint (though, that's not urgent). I know @ogenev mentioned he would find this endpoint useful. I'm also aware that the fluffy team has implemented this endpoint, so maybe we coordinate with them tomorrow on the call, and add this endpoint to the official spec?

trin-state/src/jsonrpc.rs
njgheorghita: I'm not sure this is a concern? In order for a peer to request that we store junk data, they would need rpc access which isn't the case. I'm also not so keen on validation for proximity to the content, mostly because this is very much a dev / utility endpoint, and keeping this open and unrestricted seems like the call to me. 

I guess a concern could be if a dev adds some content to their node's database with mock data, and that in turn ends up blocking the storage of valid data from the network. Imo, I don't think this situation will occur enough to be a concern.

jacobkaufmann: I was just about to write out some further thoughts.

Not all content items defined in the portal spec contain all the information that we would need to validate them (e.g. block transaction receipts), so I'm not sure that's a viable option anyway.

I'm not sure what you mean when you say that a peer would not have RPC access.

It does seem like the radius check would limit the usefulness of this endpoint from a dev perspective.

My concern was more that invalid content may propagate beyond the local node, but like you said, it's not clear whether that would occur enough to impact the network.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
jacobkaufmann: Are you referring to using the `tokio::sync::RwLock` or the `parking_lot::RwLock`?

The `parking_lot::RwLock` has a synchronous API.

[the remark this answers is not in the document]
njgheorghita: Ooooh, good to know. I was talking about `tokio`- wasn't familiar with `parking_lot` - i'll switch to use that instead

trin-state/src/jsonrpc.rs
njgheorghita: > I'm not sure what you mean when you say that a peer would not have RPC access.

In order to execute either of the `portal_*StoreContent` messages, a malicious peer would need access to a trin node's json-rpc api - which means access to send messages via ipc (where they'd need to be able to connect with the ipc file) or http (where they'd need access to the port). It doesn't seem to me that this is a reasonable concern, but I could be wrong. These endpoints are only supported via jsonrpc, it's not some new kind of message type that a malicious peer who's connected on an overlay network could use to fill a db with spam.

> My concern was more that invalid content may propagate beyond the local node, but like you said, it's not clear whether that would occur enough to impact the network.

Hmmm, yeah I'd agree with you. Firstly, in terms of quantity, this endpoint will likely be used to seed only a couple values here and there. Very valuable for testing data availability, but to test data availability only requires using a couple test datapoints, and not filling up an entire db with mock data. 

This brings up a good point, that it's not a good idea to update the `seed_database` script to use this endpoint since the script does validate the proximity of data before inserting it into the db.

I'm not sure it's even a concern that mock data will propagate? I could be wrong here, but I thought that a node will only `Offer` data to it's neighbors after receiving it via a portal network message. I don't think nodes scan their database for data and offer that to neighbors?

ogenev (changes requested):
Updated json-rpc specs [here](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false).

I think `portal_*StoreContent` name is kinda redundant, Fluffy uses `portal_*Store` in their code, which probably makes more sense. Are we going to store something other than content?

I think we should also return boolean (`true` upon success). Let me know what do you think.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -152,6 +158,22 @@ impl JsonRpcEndpoint {
...
+                ]),
+            },
+            JsonRpcEndpoint {
+                method: "portal_stateStoreContent".to_string(),
+                id: 11,
+                params: Params::Array(vec![
+                    Value::String("0044000000b8be7903aee73b8f6a59cd44a1f52c62148e1f376c0dfa1f5f773a98666efc2bd1c390624d3bd4e409a61a858e5dcc5517729a9170d014a6c96530d64dd8621d01020001".to_string()),
+                    Value::String("02".to_string()),
+                ]),
+            },
```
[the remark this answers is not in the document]
njgheorghita: Agreed, it just gets pretty messy with this current setup. You pretty much end up with a nested monstrosity containing all the raw byte vectors needed to generate the content key. I've added this as a todo to be cleaned up post #295 for now if that works.

njgheorghita:
@ogenev @jacobkaufmann Ready for another :eyes: on this one.

trin-history/src/jsonrpc.rs
```diff
@@ -50,6 +51,27 @@ impl HistoryRequestHandler {
...
+                                .overlay
+                                .storage
+                                .write()
+                                .store(&content_key, &content)
+                            {
+                                Ok(_) => Ok(Value::Bool(true)),
+                                Err(_) => Ok(Value::Bool(false)),
+                            }
+                        }
+                        Err(_) => Ok(Value::Bool(false)),
```
jacobkaufmann, line 71: I'm not sure that we should return `false` on an error. It would be useful to the client to know why a particular call failed. For example, the target node's storage may be full.

jacobkaufmann: If we always propagate the error to the client, then we would never return `false`, so you could argue that a `Bool` result type does not make sense, but I think `Bool` is better than `String`. I can't think of a case where we would return `false` and the client would not be interested in the reason for failure.

jacobkaufmann (commented):
Left some thoughts. I think `Bool` is the right result type, but I do think there are at least some error cases that the client would be interested in. I'm not sure what the best ergonomics for the API are though.

ethportal-peertest/src/jsonrpc.rs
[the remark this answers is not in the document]
jacobkaufmann: I was thinking that it would be straightforward to define a few `const` for some block hash, address, etc. in the same way that we have `const` for data radius.

trin-history/src/jsonrpc.rs
jacobkaufmann: Given that this is a dev or debug endpoint, and that the API client would likely be a dev, I think it would be useful to have some information if some store request does not succeed. We could log the error to surface that information, but that is only useful if the client (dev) is running `trin` locally.

ogenev: > I can't think of a case where we would return `false` and the client would not be interested in the reason for failure.

This is a good point. I think it makes sense to return an error to the client if something fails and we are already doing it, for example, if we add more arguments than expected or if we are unable to decode the arguments.

Not sure what is better for `portal_*Store` specifications thought, we can just say something like: "You SHOULD return `false` on an error and you MAY return the error code/message too". I think it is easier to standardize this way and we leave the error message as "nice to have" implementation detail.

So in our case, I'm more in favor to return `boolean` plus an error message if available as `Value::Array`.  This way, we match the specs and provide an additional error message.

How does this sound?

ethportal-peertest/src/jsonrpc.rs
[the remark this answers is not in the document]
njgheorghita: Gotcha, the only side-effect is that I had to label the `AccountTrieNode` struct fields as `pub`, but that doesn't seem like a problem imo.

trin-history/src/jsonrpc.rs
njgheorghita: Tbh, since this is a "dev" endpoint, I don't feel strongly that we need to match the spec. I do agree that returning the error message is useful. 

> So in our case, I'm more in favor to return boolean plus an error message if available as Value::Array. This way, we match the specs and provide an additional error message.

I'd disagree here. This suggestion doesn't match the spec. The spec dictates that we return a single bool, true or false. But again, I don't feel strongly that we need to match the spec.

Imo, decision 1 is do we need to match the spec?
- Yes - Ok let's update the spec & collab with fluffy to flesh out the ideal return type
- No - Ok, what does our return type look like? 
  - I don't feel strongly here either, but I'd probably lean towards just returning a string. "true" for success, "{error_message}" for failure.

jacobkaufmann: I wasn't aware that the endpoint was in the spec. My main concern was that a dev would try to call this endpoint to store some data and it fails with no reason available to the dev.

ogenev (approved):
Returning "true" or an error message makes sense to me. :+1:

PR #283, opened 2022-03-31
Add uTP `recv_from` method and more uTP tests

ogenev:
### What was wrong?
- uTP socket `recv_from` method was missing

### How was it fixed?
- Adding `recv_from` method allow us to receive all streamed uTP data to a write buffer:

```rust
        let mut buf = [0; BUF_SIZE];
        socket.recv_from(&mut buf).await
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -275,7 +275,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
 
                         match conn {
                             Ok(mut conn) => {
-                                // TODO: Implement this with recv_from
                                 // Handle STATE packet for SYN
                                 let mut buf = [0; BUF_SIZE];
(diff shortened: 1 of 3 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -275,7 +275,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
 
                         match conn {
                             Ok(mut conn) => {
-                                // TODO: Implement this with recv_from
                                 // Handle STATE packet for SYN
                                 let mut buf = [0; BUF_SIZE];
                                 match conn.recv(&mut buf).await {
```
njgheorghita, line 280: Seems, like you didn't actually replace these `.recv()`s with `.recv_from`?

njgheorghita (approved):
lgtm!

PR #290, opened 2022-04-01
Remove Result wrap of return type in random NodeId gen

jacobkaufmann:
### What was wrong?

We unnecessarily wrap return type for random `NodeId` generation function in `Result`.

### How was it fixed?

Remove `Result` wrap from return type.

### To-Do

- [x] Clean up commit history


ogenev (approved):
LGTM :100:

PR #294, opened 2022-04-05
Fix data dir setup

njgheorghita:
### What was wrong?
@hmrtn pointed out that using `TRIN_DATA_PATH` environment variable was broken.

### How was it fixed?
- Use `Path` to handle env var paths with & without a trailing slash (`./path/to/db` & `./path/to/db/`)
- Updates default data dir to namespace under `/trin`. Previously, all of the `trin_xxxxx` directories were just stored under the default data dir.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/294.fixed.md
+++ b/newsfragments/294.fixed.md
@@ -1,0 +1,1 @@
+Update data dir setup that was broken when using an environment variable to set the path.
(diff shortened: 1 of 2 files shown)


ogenev (approved):
Looks good!

PR #295, opened 2022-04-05
Reconfigure ethportal test structure

njgheorghita:
### What was wrong?
`ethportal-peertest` had a big flaw, you could only run one test per jsonrpc endpoint. This is not ideal, since we'll want to run multiple tests for various functionality against a single endpoint.

### How was it fixed?
Reconfigured the test architecture to allow for multiple tests against a single endpoint. Furthermore, included some refactoring that will later allow us to introduce more complex tests (eg, a single test composed of multiple jsonrpc requests - which will be useful in testing content, iterative find nodes, and iterative find content...). It's not fully implemented here, but all the "validation functions" will need to perform these tests is whether to use `make_ipc_request` or `make_http_request` (aka, access to the `portalnet_config`) and then they can encapsulate any logic needed to perform and validate subsequent jsonrpc requests for the test.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -3,7 +3,6 @@ use std::io::prelude::*;
 use std::os::unix;
 use std::panic;
 use std::time::Duration;
-use std::{thread, time};
 
 use anyhow::anyhow;
 use hyper::{self, Body, Client, Method, Request};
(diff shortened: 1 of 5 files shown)


newsfragments/295.fixed.md
```diff
@@ -0,0 +1 @@
+Reconfigure ethportal-peertest testing structure.
```
carver, line 1:
```suggestion
Reconfigure ethportal-peertest testing structure to run multiple tests on a single spawned instance.
```

This is based on the PR description. Although I frankly have missed how this refactor makes it more possible than before. Could you explain a little more?.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -3,183 +3,266 @@ use std::io::prelude::*;
...
-                }}"#,
-                self.id,
-                self.method,
-                serde_json::to_string(&self.params).unwrap()
-            ),
-        }
-    }
+fn validate_web3_client_version(val: &Value, _peertest: &Peertest) {
+    assert_eq!(val.as_str().unwrap(), "trin v0.1.0");
+}
```
carver, line 198: So I know it was already like this, but since we're refactoring... I find it hard to quickly understand the test when the setup and validation are so separated. I am in favor of figuring out a way to combine them. Maybe even just inlining this function as an anonymous argument to the `Test`?

carver (approved):

newsfragments/295.fixed.md
njgheorghita: Ahh, not to run multiple tests on a single spawned instance, but multiple tests for a single jsonrpc endpoint (eg. `portal_stateFindContent`). Which was basically a result of how `validate_endpoint_response()` was built to match on a string representation of the endpoint itself.

ethportal-peertest/src/jsonrpc.rs
njgheorghita: Definitely agreed. I spent a fair amount of time trying to find a better solution, but I just kept on running into problems with the rust compiler. My approach was to inline the `validate_*` code as a closure on `Test`. The problem is that rust's compiler treats every closure as an individual type, regardless of what it accepts/returns. The only way around this that I could find was by wrapping the validation logic in a function with a defined signature. Is this similar to what you had in mind by "inlining this function as an anonymous argument"?

carver: Hm, it sounds like you maybe tried the thing that I was hoping would work, something like:
```rust
        Test::new(
            JsonRpcRequest {
                method: "portal_statePing".to_string(),
                id: 5,
                params: Params::Array(vec![
                    Value::String(peertest.bootnode.enr.to_base64()),
                    Value::String(DATA_RADIUS.to_owned()),
                ]),
            },
            |result: &Value, _peertest| {
                assert_eq!(
                    result.get("dataRadius").unwrap().as_str().unwrap(),
                    DATA_RADIUS
                );  
                assert_eq!(
                    result.get("enrSeq").unwrap().as_str().unwrap(),
                    ENR_SEQ.to_string()
                );  
            },
        ),
```

The main goal is to get the lines nearby each other that are defining the request made and the validation performed.

njgheorghita: Yup, exactly that. Unfortunately it's a nogo with the compiler. But I'll definitely keep this goal in mind as I add tests and hopefully figure out something that achieves this goal

Issue #298, opened 2022-04-07
Flaky tests are back on the menu


njgheorghita:
I've been experiencing a significant increase in flaky tests recently. I'm guessing this is due to the `FindNodes` test added to `peertest`. Would be curious to get confirmation from others. But, this has also raised an issue that when CI runs peertest, if there is a failure, then it doesn't display the error message, it simply just says that the test run failed ([example](https://app.circleci.com/pipelines/github/ethereum/trin/1267/workflows/11b900c3-b67e-40d7-9de7-9c516303c714/jobs/2473)). This makes it more difficult to diagnose the source of flakiness, so we should update how the tests are run to make sure that an error message / stack trace is displayed if it occurs.

jacobkaufmann:
I can confirm.

I added some output and I saw the following:

```sh
thread 'test::test_launches' panicked at 'assertion failed: nodes.enrs.contains(&peertest.nodes[0].enr)', ethportal-peertest/src/jsonrpc.rs:247:5
```

njgheorghita:
Nice, yeah there's a fix in #303 for logging output of the failed test and that's definitely the culprit. Having a tough time sussing out the source of the error though. Comparing logs for a successful vs. failed run doesn't yield much insight since the logs of all three nodes are output simultaneously. Do you have any ideas for a simple way to associate a particular logged line with its respective node while running peertest?

njgheorghita:
Nvm - 99% I figured this out, will get the fix in tomorrow

njgheorghita:
Nvm - 99% I figured this out, will get the fix in tomorrow

PR #299, opened 2022-04-07
Add utp-testing crate to Dockerfile

njgheorghita:
### What was wrong?
`utp-testing` crate wasn't added to Dockerfile, so compilation failed.

### How was it fixed?
Added `utp-testing` crate to Dockerfile

### To-Do

- [x] Clean up commit history


carver:
The win-build failed, but it didn't have much information about why, so I re-ran it. The change looks obviously reasonable, but maybe it uncovered some other problem?

njgheorghita:
@carver Yeah, that's most likely related to #298 - Which I'm digging into, I doubt it has anything to do with this change.

ogenev (approved):

Issue #302, opened 2022-04-08
Lock held for too long when processing uTP packet


ogenev:
Sometimes, there is a warning that a lock is held for too long:

```j
Apr 08 11:54:24.756  WARN trin_core::locks: [/home/ogenev/dev/eth/trin/trin-core/src/portalnet/events.rs:89] lock held for over 100ms, not yet released    
Apr 08 11:54:24.944  WARN trin_core::locks: [/home/ogenev/dev/eth/trin/trin-core/src/portalnet/events.rs:89] lock held for too long: 290ms 
```

This is because we are acquiring the uTP listener lock every time we process an uTP packet, and sometimes it takes some time to process it.

Refactor `process_utp_request` in a way to not acquire the lock by using a channel.

https://github.com/ethereum/trin/blob/e876d5b77b4d04c5bbd0a4c0006ae5cb61d1e371/trin-core/src/portalnet/events.rs#L91

PR #304, opened 2022-04-11
Add overlay layer <-> uTP listener communication channel instead of locks

ogenev:
### What was wrong?

The portal overlay layer communicates with uTP listener with locks which causes some of the locks to wait for too long.

FIxes https://github.com/ethereum/trin/issues/302

### How was it fixed?
- Implement  Overlay <-> uTP listener communication via a channel and remove the locks.
- `UtpListener` now runs as a self-contained service in the background and waits for requests from the overlay layer.

Those changes bring a performance boost and reduce the number of log warnings about locks held.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use std::collections::HashSet;
 use std::marker::PhantomData;
 use std::sync::Arc;
@@ -15,9 +16,8 @@ use crate::portalnet::types::messages::{
     ProtocolId, Request, Response,
 };
 
-use crate::locks::RwLoggingExt;
-use crate::utp::stream::{UtpListener, BUF_SIZE};
-use crate::utp::trin_helpers::{UtpAccept, UtpMessage, UtpMessageId};
+use crate::utp::stream::{UtpListenerRequest, UtpSocket, BUF_SIZE};
+use crate::utp::trin_helpers::{UtpAccept, UtpMessage};
 use discv5::{
     enr::NodeId,
     kbucket::{Filter, KBucketsTable},
@@ -29,7 +29,6 @@ use parking_lot::RwLock;
 use ssz::Encode;
 use ssz_types::VariableList;
 use tokio::sync::mpsc::UnboundedSender;
-use tokio::sync::RwLock as RwLockT;
 use tracing::{debug, warn};
 
 pub use super::overlay_service::OverlayRequestError;
@@ -79,7 +78,8 @@ pub struct OverlayProtocol<TContentKey, TMetric> {
     protocol: ProtocolId,
     /// A sender to send requests to the OverlayService.
     request_tx: UnboundedSender<OverlayRequest>,
-    utp_listener: Arc<RwLockT<UtpListener>>,
+    /// A sender to send request to UtpListener
+    utp_listener_tx: UnboundedSender<UtpListenerRequest>,
     /// Declare the allowed content key types for a given overlay network.
     /// Use a phantom, because we don't store any keys in this struct.
     /// For example, this type is used when decoding a content key received over the network.
@@ -94,7 +94,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
     pub async fn new(
         config: OverlayConfig,
         discovery: Arc<Discovery>,
-        utp_listener: Arc<RwLockT<UtpListener>>,
+        utp_listener_tx: UnboundedSender<UtpListenerRequest>,
         storage: Arc<RwLock<PortalStorage>>,
         data_radius: U256,
         protocol: ProtocolId,
@@ -116,7 +116,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             config.ping_queue_interval,
             Arc::clone(&data_radius),
             protocol.clone(),
-            utp_listener.clone(),
+            utp_listener_tx.clone(),
             config.enable_metrics,
         )
         .await
@@ -129,7 +129,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             storage,
             protocol,
             request_tx,
-            utp_listener,
+            utp_listener_tx,
             phantom_content_key: PhantomData,
             phantom_metric: PhantomData,
         }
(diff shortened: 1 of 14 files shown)


trin-core/tests/overlay.rs
```diff
@@ -35,12 +35,12 @@ async fn init_overlay(
     .unwrap();
     let db = Arc::new(RwLock::new(PortalStorage::new(storage_config).unwrap()));
     let overlay_config = OverlayConfig::default();
-    let utp_listener = Arc::new(RwLockT::new(UtpListener::new(Arc::clone(&discovery))));
+    let (utp_listener_tx, _) = unbounded_channel::<UtpListenerRequest>();
```
carver, line 38: This is just odd enough that it might be worth a little comment, like:
```suggestion
    // Ignore all UTP events
    let (utp_listener_tx, _) = unbounded_channel::<UtpListenerRequest>();
```

utp-testing/src/main.rs
```diff
@@ -43,20 +42,13 @@ impl TestApp {
...
-            .await
-            .unwrap();
+        let mut event_stream = self.discovery.discv5.event_stream().await.unwrap();
 
-        let listener = Arc::clone(&self.utp_listener);
+        let utp_sender = self.utp_event_tx.clone();
 
         tokio::spawn(async move {
             while let Some(event) = event_stream.recv().await {
+                debug!("event: {event:?}");
```
carver, line 51: Probably worth adding a bit more context into this log, like:
```suggestion
                debug!("utp-testing TestApp handling event: {event:?}");
```

utp-testing/src/main.rs
```diff
@@ -66,27 +58,23 @@ impl TestApp {
...
+            .utp_listener_tx
+            .send(UtpListenerRequest::OfferStream(conn_id));
 
         // also listen on conn_id + 1 because this is the actual receive path for acceptor
-        self.utp_listener
-            .write()
-            .await
-            .listening
-            .insert(connection_id + 1, UtpMessageId::OfferStream);
+        let conn_id_recv = conn_id.wrapping_add(conn_id);
```
carver, line 74: Typo in the +1?
```suggestion
        let conn_id_recv = conn_id.wrapping_add(1);
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -612,27 +610,29 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
-                        utp_listener.listening.insert(
-                            connection_id.clone(),
-                            // Temporaly storing content data, so we can send it right after we receive
-                            // SYN packet from the requester
-                            UtpMessageId::FindContentData(Content::Content(content)),
-                        );
+                    let conn_id: u16 = crate::utp::stream::rand();
+
+                    // listen for incoming connection request on conn_id, as part of utp handshake and
+                    // Temporally storing content data, so we can send it right after we receive
```
carver, line 616:
```suggestion
                    // temporarily store content data, so we can send it right after we receive
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -258,56 +258,72 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                    Err(err) => {
+                        warn!("Unable to initiate uTP stream with remote node {err}");
+                        Err(OverlayRequestError::UtpError(err.to_string()))
                     }
                 }
             }
-            Ok(_) => Err(OverlayRequestError::InvalidResponse),
-            Err(error) => Err(error),
+            Err(err) => {
+                warn!("Unable to initiate uTP stream with remote node {err}");
```
carver, line 324: If we start seeing this log, it would be nice to immediately be able to tell whether it's this line or 318, so maybe disambiguate the two somehow? Bonus points for further explaining the kind of failure it might be in the log.

trin-core/src/portalnet/overlay.rs
```diff
@@ -361,70 +377,75 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+
+                for (i, key) in response
+                    .content_keys
+                    .clone()
+                    .iter()
+                    .zip(content_keys_offered.iter())
+                {
+                    if i == true {
+                        match self.storage.read().get(&key.clone()) {
+                            Ok(content) => match content {
```
carver, line 416: Most of this logic didn't change, but got caught in the diff anyway because of an indent. Sometimes that's a clue to me that breaking down into smaller functions will make future reviews easier. So 👍🏻 👍🏻  if you feel up to doing that now.

trin-core/src/utp/stream.rs
```diff
@@ -367,10 +454,10 @@ pub struct UtpSocket {
     last_rollover: Timestamp,
 
     /// Maximum retransmission retries
-    pub max_retransmission_retries: u32,
+    max_retransmission_retries: u32,
 
     /// Receive channel for discv5 socket
-    rx: Arc<RwLock<mpsc::UnboundedReceiver<Packet>>>,
+    discv_rx: Arc<RwLock<UnboundedReceiver<Packet>>>,
```
carver, line 460: typo?
```suggestion
    discv5_rx: Arc<RwLock<UnboundedReceiver<Packet>>>,
```

Or just `disc_rx` seems reasonable too.

trin-core/src/utp/stream.rs
```diff
@@ -589,15 +676,7 @@ impl UtpSocket {
         });
     }
 
-    fn filter(current_delay: &[u32]) -> u32 {
```
carver, line 592: Was this just dead code?

carver (approved):
Yeah, looks solid to me! The `match` in `process_discv5_requests()` looks more consistent and cleaner now... I like the improvements in `trin-core/src/utp/stream` too!

trin-core/src/utp/stream.rs
ogenev: yep

utp-testing/src/main.rs
ogenev: Ah, good catch!

trin-core/src/utp/stream.rs
carver: `discv_rx` is still there.

ogenev: Should be fixed now. This slipped around because I renamed it only on line 469.

PR #305, opened 2022-04-12
Define Distance type for distance and radius values

jacobkaufmann:
### What was wrong?

Throughout the code, we use `u64::MAX` as the default radius. This value covers a small fraction of the DHT key space. The default radius should be the entire key space, which is `U256::MAX`.

### How was it fixed?

- Define new `Distance` type as a wrapper type around `U256`.
- Define associated constant `Distance::MAX` as `U256::MAX`.
- Use `Distance::MAX` in place of `u64::MAX`.

### To-Do

- [x] Clean up commit history


jacobkaufmann:
@carver this sort of goes in the opposite direction of your prior comment and replaces the `u64` in `PortalStorage` with the wrapped `U256` representation. AFAIK we are not doing any in-database computations at this time. If needed, we can add a method to `Distance` that returns the highest 64 bits as a `u64`.

trin-core/src/portalnet/storage.rs
```diff
@@ -131,7 +130,7 @@ impl PortalStorage {
         }
 
         // Don't store if it's outside our radius
-        if self.data_radius < u64::MAX {
+        if self.data_radius < Distance::MAX {
             Ok(self.distance_to_content_id(&content_id) < self.data_radius)
```
carver, line 140: It seems like the comment should go inside the block. I'm surprised the test is even there. Is it basically a performance optimization? Like if we've never shrunk the radius, then the radius will equal the max and we can skip the distance calculation? That seems worth a comment on the line before the `if`.

trin-core/src/portalnet/types/metric.rs
```diff
@@ -1,22 +1,73 @@
...
+
+impl Distance {
+    /// The maximum value.
+    pub const MAX: Self = Self(U256::MAX);
+
+    /// Returns the integer base-2 logarithm of `self`.
+    ///
+    /// Returns `None` is `self` is zero, because the logarithm of zero is undefined. Otherwise,
+    /// returns `Some(log2)` where `log2` is in the range [1, 256].
+    pub fn log2(&self) -> Option<u64> {
```
carver, line 24: I think a `Result` could make even more sense here. Like I would expect an `Err` if trying to divide by zero, more than a `None`.

Also, is `usize` preferable here? If it was a 32-bit system, it seems like that would be fine. I have no real opinion on this part.

trin-core/src/portalnet/types/metric.rs
```diff
@@ -45,6 +96,41 @@ mod test {
...
+    #[test]
+    fn distance_log2() {
+        fn prop(x: DhtPoint) -> TestResult {
+            let x = U256::from_big_endian(&x.0);
+            let distance = Distance(x);
+            let log2_distance = distance.log2();
+
+            match log2_distance {
+                Some(log2) => {
+                    let x_floor = U256::from(1u8) << log2 - 1;
```
carver, line 108: I'm a fan of being super explicit with order of operations. So as a reader I find it quicker to confidently comprehend:
```suggestion
                    let x_floor = U256::from(1u8) << (log2 - 1);
```

trin-core/src/portalnet/types/metric.rs
```diff
@@ -45,6 +96,41 @@ mod test {
...
+                    let x_ceil = if log2 == 256 {
+                        U256::MAX
+                    } else {
+                        U256::from(1u8) << log2
+                    };
+
+                    TestResult::from_bool(x >= x_floor && x <= x_ceil)
```
carver, line 115: I think it would be a correct tighter bound to say that x needs to be strictly beneath the ceiling. In order to handle the 256 case, it could follow this approach:
```suggestion
                    if log2 == 256 {
                        TestResult::from_bool(x >= x_floor && x <= U256::MAX)
                    } else {
                        let x_ceil = U256::from(1u8) << log2;
                        TestResult::from_bool(x >= x_floor && x < x_ceil)
                    }
```

The MAX check is actually a little redundant, since U256 cannot be higher, so it would also be fine to just use this in the 256 case IMO: `TestResult::from_bool(x >= x_floor)`

carver (approved):
I have heard Piper talk about the likely need for 64 bit comparisons in-database enough times that I don't think it falls in the YAGNI case.

But, I guess I don't have an objection to converting it at DB-insertion time, enough to make the pitch to you... :)

trin-core/src/portalnet/types/metric.rs
jacobkaufmann: I would push back in favor of `None`, because it captures the idea of "undefined". If we return a `Result`, that is another error that we have to create, or if we return the unit type `()` as the `Err` variant, then that is equivalent to `None`.

I would also note that the Rust primitive types return `None` for [`checked_div`](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_div) on divide by zero, so it seems `None` is more idiomatic.

Good suggestion on `usize`. `U256::leading_zeros` returns a `u32`, so there are no conversion problems.

trin-core/src/portalnet/storage.rs
jacobkaufmann: Yeah, this check was already present, and I didn't make any change there to minimize the diff size. I can update the comments though.

trin-core/src/portalnet/types/metric.rs
carver: > I would also note that the Rust primitive types return `None` for [`checked_div`](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_div) on divide by zero, so it seems `None` is more idiomatic.

An excellent reason to keep it as `Option`. 👍🏻

jacobkaufmann:
@carver just to clarify, you are okay with no method for `u64` representation for now? Even if we added it here, it would be dead code.

trin-core/src/portalnet/types/metric.rs
```diff
@@ -45,6 +96,41 @@ mod test {
...
+                        TestResult::from_bool(x >= x_floor)
+                    } else {
+                        let x_ceil = U256::from(1u8) << log2;
+                        TestResult::from_bool(x >= x_floor && x < x_ceil)
+                    }
+                }
+                None => TestResult::from_bool(distance.0.is_zero()),
+            }
+        }
+        quickcheck(prop as fn(DhtPoint) -> TestResult);
```
carver, line 120: In a quick read of `quickcheck` I couldn't find an obvious way to force it to test any particular examples. Normally I would want to see that a few points near the edge cases of exactly the log2 value as distance +/- 1 all pass.

If you know how to add that, that would be cool. Maybe another alternative is to manually run `prop()` with hard-coded arguments and match or unwrap the `TestResult`.

Not a big deal I think, but if it's quick, it would be a great addition.

carver (approved):
LGTM

carver:
> @carver just to clarify, you are okay with no method for `u64` representation for now? Even if we added it here, it would be dead code.

Yeah, agreed

trin-core/src/portalnet/types/metric.rs
```diff
@@ -1,22 +1,73 @@
+use std::fmt;
```
ogenev, line 1: The filename  `metric.rs` is a little confusing for me, by default, I'm assuming it is something related to Prometheus metrics, maybe change this filename to `distance.rs` or something more appropriate?

PR #306, opened 2022-04-14
Change discv5 crate source

ogenev:
### What was wrong?
Discv5 dependency used a patched source pointing to a forked branch of discv5 repo:

```rust
discv5 = { git = "https://github.com/ogenev/discv5.git", branch = "public_kbucket_stable" }
```

### How was it fixed?
Our patch was[ merged](https://github.com/sigp/discv5/pull/97) upstream a while ago. Switching to the main discv5 branch so we can downstream the latest changes. Bumps also `delay_map` crate version because of dependency issues.

### To-Do

- [x] Clean up commit history


jacobkaufmann (commented):
Should we go with latest version as opposed to master? The only concern is upstream breaking changes.

ogenev:
> Should we go with latest version as opposed to master? The only concern is upstream breaking changes.

I think the latest version is lagging a bit and I'm not too worried about breaking changes because at this stage we can quickly adjust. I think until we have a stable Trin release, it makes sense to always use the latest dicv5 updates.

jacobkaufmann (approved):
Sounds good to me.

carver (approved):
With some upcoming changes to discv5, this might get swapped again. But it's a step forward for now. I also 👍🏻  the argument about tracking master instead of a discv5 release, until trin is closer to a stable release.

PR #307, opened 2022-04-14
Add jsonrpc api docs.

njgheorghita:
### What was wrong?
It seemed to me like it's a good idea to have a resource for currently supported jsonrpc endpoints. We don't fully support the whole ETH spec - or the whole Portal spec. We also currently use a couple custom endpoints that aren't found in either spec. 

Updated return values....
- node ids to be full node ids rather than `0x1234..abcd`
- Hex values to have a `0x` prefix.

### How was it fixed?
Wrote docs

### To-Do

- [x] Clean up commit history


docs/jsonrpc_api.md
```diff
@@ -0,0 +1,499 @@
+# JSON-RPC API
+
+This is a document for all JSON-RPC API endpoints currently supported by Trin. Trin will eventually support the entire Portal Network JSON-RPC API, which can be found [here]().
```
jacobkaufmann, line 3: Looks like we forgot to add the URL in markdown link.

jacobkaufmann, line 3: Looks like we forgot to add the URL in the markdown link.

jacobkaufmann (commented):
This seems like we are duplicating some of the JSON-RPC specs for `eth` and `portal`, which I am not to keen to maintain. I think we can reduce this to (1) the list of endpoints that we support, (2) links to existing specs that contain those endpoints, and (3) descriptions for any non-standard endpoints.

Preferably, for (3) we would have some other viewable JSON-RPC spec similar to what [geth](https://geth.ethereum.org/docs/rpc/ns-eth) does for its custom endpoints. However, I don't think it's worth creating something like that at this stage.

PR #321, opened 2022-05-02
Fix missing `find` module in module tree

jacobkaufmann:
### What was wrong?

In #259, we did not add the new `find` module to the `mod.rs` for the `portalnet` module. This allowed numerous compile-time errors to slip through, because those new files were not in the module tree.

### How was it fixed?

Add the `find` module to the module tree and resolve the outstanding compile-time errors.

### To-Do

- [x] Clean up commit history


ogenev (approved):
LGTM :100:

PR #322, opened 2022-05-03
Merge imports from the same crate into a single use statement

ogenev:
### What was wrong?
Imports were messy and not structured in an intuitive way.

### How was it fixed?
 Used `cargo +nightly fmt` with `imports_granularity = "Crate"` [config](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity).

### To-Do

- [x] Clean up commit history


jacobkaufmann (approved):
🔥

Issue #312, opened 2022-04-22
Handle concurrently multiple uTP streams


ogenev:
We need to refactor the current internal uTP packet channel to send and receive packets based on the connection ids, so we can handle multiple uTP streams concurrently.

PR #309, opened 2022-04-21
Leave IP out of ENR by default

carver:
An alternative fix #286

This fits with the discv5 validation, which hangs up on clients
immediately if the ENR IP does not match the observed IP.

Also, had to update a couple tests to configure local IPs so they can
contact each other. (The internal IP feature randomly chose a local
IP, which caused the discv5 validation to reject the local connections)

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [x] Clean up commit history

PR #323, opened 2022-05-03
Handle concurrently multiple uTP streams

ogenev:
### What was wrong?
Fixes https://github.com/ethereum/trin/issues/312

### How was it fixed?

1. Create a uTP stream with an internal discv5 packet channel.
2. Every time when we handle a uTP packet, we check for an existing uTP connection matching the remote NodeId and received connection id, and then we send the packet to the corresponding internal uTP stream channel.
3. Refactor RESET packet uTP listener handler.
4. Fix broken local utp-testing execution caused by https://github.com/ethereum/trin/pull/309.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -66,6 +66,23 @@ pub fn rand() -> u16 {
     rand::thread_rng().gen()
 }
 
+/// Connection key for storing active uTP connections
+#[derive(Hash, Eq, PartialEq, Copy, Clone, Debug)]
+pub struct ConnectionKey {
+    node_id: NodeId,
+    conn_id_recv: ConnId,
+}
+
+impl ConnectionKey {
+    fn new(node_id: NodeId, conn_id_recv: ConnId) -> Self {
+        Self {
+            node_id,
+            conn_id_recv,
+        }
+    }
+}
+
+/// uTP socket connection state
 #[derive(PartialEq, Eq, Clone, Copy, Debug)]
 pub enum SocketState {
     Uninitialized,
@@ -103,7 +120,7 @@ pub struct UtpListener {
     /// Base discv5 layer
     discovery: Arc<Discovery>,
     /// Store all active connections
-    utp_connections: HashMap<NodeId, UtpSocket>,
+    utp_connections: HashMap<ConnectionKey, UtpSocket>,
     /// uTP connection ids to listen for
     listening: HashMap<ConnId, UtpMessageId>,
     /// Receiver for uTP events sent from the main portal event handler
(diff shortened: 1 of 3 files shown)


trin-core/src/utp/stream.rs
```diff
@@ -117,14 +103,9 @@ pub struct UtpListener {
     /// Base discv5 layer
     discovery: Arc<Discovery>,
     /// Store all active connections
-    utp_connections: HashMap<ConnectionKey, UtpSocket>,
+    utp_connections: HashMap<NodeId, UtpSocket>,
```
jacobkaufmann, line 106: It seems that this change would prevent more than one simultaneous uTP connection between the local node and some remote peer. This would be problematic if the local node and remote peer participate in multiple portal subnetworks.

I would need to investigate further to see whether this is possible, but another concern is the following scenario:

- The local node initiates a uTP connection with some remote peer on subnetwork `X`
- The local node and remote peer begin to exchange data
- The remote peer initiates a uTP connection with the local node on subnetwork `Y`
- The second connection disrupts the first connection by taking over its entry in the connection map

Either the uTP connection on subnetwork `Y` fails or it hijacks the existing uTP connection on subnetwork `X`.

Perhaps I'm missing something here though.

trin-core/src/utp/stream.rs
```diff
@@ -174,60 +152,43 @@ impl UtpListener {
...
+                        if let Some(conn) = self.utp_connections.get_mut(node_id) {
+                            if conn.discv5_tx.send(packet).is_ok() {
+                                let mut buf = [0; BUF_SIZE];
+                                if let Err(msg) = conn.recv(&mut buf).await {
+                                    error!("Unable to receive uTP RESET packet: {msg}")
+                                }
+                            } else {
+                                error!("Unable to send RESET packet to uTP stream handler")
+                            }
                         }
```
jacobkaufmann, line 173: Do we forget to set the socket state to `SocketState::Closed`?

jacobkaufmann (commented):
My main concern is the use of `NodeId` as the key in the connection map for the reason described below.

Also, it would be nice to have some rationale for the changes to `utp-testing` in the PR description.

trin-core/src/utp/stream.rs
ogenev: Good call, I brought back the `ConnectionKey` struct which is a combination of `NodeId` and uTP socket receive connection id. As we generate those connection ids randomly, it is a very low probability to have two active uTP connections with the same connection id and `NodeId`.

trin-core/src/utp/stream.rs
ogenev: No, the code here was leftover from some previous implementation and not up to date.
Now, we change the socket state implicitly when we receive and handle that packet (`conn.recv()` and in `handle_packet()`).

Issue #313, opened 2022-04-22
Propagate Gossip implementation


ogenev:
Handle accepted content (store) and propagate gossip with OFFER/ACCEPT..

Subtasks:

- [x] https://github.com/ethereum/trin/pull/324
- [x] https://github.com/ethereum/trin/pull/325
- [x] https://github.com/ethereum/trin/pull/337
- [ ] Implement propagate gossip algorithm when processing ACCEPT uTP payload:
https://github.com/ethereum/trin/pull/328

PR #324, opened 2022-05-09
Refactor handling `OFFER` req and update `should_store` db method 

ogenev:
### What was wrong?
We were accepting any content offered without checking if we are interested in the content.

### How was it fixed?
1. Update the `should_store` DB method to use `get_pinned` to avoid unnecessary memory copy and bump rocksdb crate version.
2. Remove the dummy `should_store` method when handling OFFER requests and accept only content we are interested in.

### To-Do

- [x] Clean up commit history


mrferris:
Interesting, if I'm understanding correctly this will result in a node sometimes not accepting data that it should accept.

Have you done any analysis on why & how often these false positives from `key_may_exist` will occur? I've glanced at the code but haven't come to a conclusion. I wonder whether it's worth the performance increase.

Also for what it's worth (if anything), Facebook doesn't use this function in production, at least in 2018:
https://github.com/facebook/rocksdb/issues/4149

ogenev:
> Also for what it's worth (if anything), Facebook doesn't use this function in production, at least in 2018: [facebook/rocksdb#4149](https://github.com/facebook/rocksdb/issues/4149)

Hmm, It seems it is not as reliable as I expected, I will revert back to the `get()` method. Thanks for pointing this out!

ogenev:
Actually, `get_pinned` seems more appropriate to avoid unnecessary memory copy.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -643,27 +643,42 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
         let accept_keys = request.content_keys.clone();
 
-        for (i, key) in request.content_keys.iter().enumerate() {
-            // should_store is currently a dummy function
-            // the actual function will take ContentKey type, so we'll  have to decode keys here
+        for (i, key) in request.content_keys.into_iter().enumerate() {
+            // TODO: This will not work when we receive state Offer message with state content keys
+            // we need to refactor it in a way to get access to the protocol id in this method
+            let key = HistoryContentKey::try_from(key)
+                .map_err(|err| OverlayRequestError::AcceptError(err.to_string()))?;
```
mrferris, line 665: I think that instead of having to check the protocol ID, we can use the generic `TContentKey` to do this:

```suggestion
           let key = (TContentKey::try_from)(key)
                .map_err(|err| OverlayRequestError::AcceptError("Invalid Content Key".to_string()))?;
```

This is also done [here](https://github.com/ethereum/trin/blob/26deb62b3a9e00af308ab0bf0f6c20b81f3d1153/trin-core/src/portalnet/overlay_service.rs#L589)

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -643,27 +643,42 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
             requested_keys
-                .set(i, should_store(key))
-                .map_err(|e| OverlayRequestError::AcceptError(e))?;
+                .set(
+                    i,
+                    self.storage.read().should_store(&key).map_err(|_| {
+                        OverlayRequestError::AcceptError(
+                            "Portal storage error: unable to check content availability".to_owned(),
+                        )
+                    })?,
```
mrferris, line 673: Seems like this will lose potentially useful context given by the `PortalStorageError` about why RocksDB failed:
```suggestion
                    self.storage.read().should_store(&key).map_err(|err| {
                        OverlayRequestError::AcceptError(
                           ("Failed to check whether data should be stored: " + err.to_string()).to_owned(),
                        )
                    })?,
```
I doubt that my string concatenation compiles 😁  but just expressing the idea

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -643,27 +643,42 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         }
     }
 
-    /// Attempts to build a `Accept` response for a `Offer` request.
+    /// Attempts to build a `Accept` response for an `Offer` request.
```
mrferris, line 646:
```suggestion
    /// Attempts to build an `Accept` response for an `Offer` request.
```
the nittiest nit-pick

mrferris (changes requested):
Great find on `get_pinned`, definitely looks like what we want to be using.

trin-core/src/portalnet/overlay_service.rs
ogenev: Yeah, `format!` will do the trick here, this is not python MIke :smile_cat: .

trin-core/src/portalnet/overlay_service.rs
ogenev: Nice, this is exactly what we need :100:!

mrferris (approved):
LGTM!

carver (approved):
LGTM!

PR #325, opened 2022-05-10
Emit global uTP payload event in uTP listener when an uTP stream is closed/reset

ogenev:
### What was wrong?

`UtpListener` runs as a self-contained service in the background and handles all active uTP streams.
We need a way to process all closed uTP streams and pass the payload to the overlay service, so we can verify, store and propagate gossip/POKE the received content via uTP.

### How was it fixed?

1. Rename `UtpSocket` to `UtpStream`.
2. Refactor the way we are storing the received payload (DATA packets) in the uTP socket.
3.  Add a new `AddActiveConnection` UtpListener request and move the initialization of an uTP stream inside UtpListener.
4.  Add `UtpStream` -> `UtpListener` event channel and emit event inside `UtpSocket` when stream state changes to `Closed` or `Reset`.
5. Upon every `UtpStream` event, `UtpListener` processes this event (doing some simple active streams maintenance by removing closed or reset streams) and emit a global `UtpListenerEvent`
6. Remove redundant and dead code.
7. Add a test case for uTP listener events.

Manually tested OFFER/ACCEPT scenarios with Fluffy.

### To-Do
- [x] 1. Add protocol id when creating uTP socket and move uTP socket creation outside of UtpListener.
- [x] 2. Add UtpSocket <-> UtpListener event channel and emit an event upon uTP socket closing.
- [x] 3. Emit global UtpListener event with uTP payload, stream id, and protocol id on closed uTP socket. 

The next two tasks will be addressed in the next PR (https://github.com/ethereum/trin/pull/337) as this is becoming quite large:

4. Add a uTP listener event handler to dispatch the global event to the specific overlay networks.
5. Listen and handle dispatched events in the overlay layer.

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -15,9 +15,12 @@ use crate::portalnet::{
     },
 };
 
-use crate::utp::{
-    stream::{UtpListenerRequest, UtpSocket, BUF_SIZE},
-    trin_helpers::{UtpAccept, UtpMessage},
+use crate::{
+    portalnet::types::content_key::RawContentKey,
+    utp::{
+        stream::{UtpListenerEvent, UtpListenerRequest, UtpSocket, BUF_SIZE},
+        trin_helpers::{UtpAccept, UtpMessage},
+    },
 };
 use discv5::{
     enr::NodeId,
@@ -29,7 +32,10 @@ use futures::channel::oneshot;
 use parking_lot::RwLock;
 use ssz::Encode;
 use ssz_types::VariableList;
-use tokio::sync::mpsc::UnboundedSender;
+use tokio::sync::{
+    mpsc::{UnboundedReceiver, UnboundedSender},
+    RwLock as TRwLock,
+};
 use tracing::{debug, warn};
 
 pub use super::overlay_service::{OverlayRequestError, RequestDirection};
@@ -95,6 +101,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         config: OverlayConfig,
         discovery: Arc<Discovery>,
         utp_listener_tx: UnboundedSender<UtpListenerRequest>,
+        utp_listener_rx: Arc<TRwLock<UnboundedReceiver<UtpListenerEvent>>>,
         storage: Arc<RwLock<PortalStorage>>,
         data_radius: U256,
         protocol: ProtocolId,
@@ -117,6 +124,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             Arc::clone(&data_radius),
             protocol.clone(),
             utp_listener_tx.clone(),
+            utp_listener_rx,
             config.enable_metrics,
         )
         .await
@@ -331,7 +339,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
     /// Offer is also sent to nodes after FindContent (POKE)
     pub async fn send_offer(
         &self,
-        content_keys: Vec<Vec<u8>>,
+        content_keys: Vec<RawContentKey>,
         enr: Enr,
     ) -> Result<Accept, OverlayRequestError> {
         // Construct the request.
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -18,7 +18,7 @@ use crate::portalnet::{
 use crate::{
     portalnet::types::content_key::RawContentKey,
     utp::{
-        stream::{UtpListenerEvent, UtpListenerRequest, UtpSocket, BUF_SIZE},
+        stream::{UtpListenerRequest, UtpSocket, BUF_SIZE},
         trin_helpers::{UtpAccept, UtpMessage, UtpStreamId},
     },
 };
@@ -32,10 +32,7 @@ use futures::channel::oneshot;
 use parking_lot::RwLock;
 use ssz::Encode;
 use ssz_types::VariableList;
-use tokio::sync::{
-    mpsc::{UnboundedReceiver, UnboundedSender},
-    RwLock as TRwLock,
-};
+use tokio::sync::mpsc::UnboundedSender;
 use tracing::{debug, warn};
 
 pub use super::overlay_service::{OverlayRequestError, RequestDirection};
@@ -101,7 +98,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         config: OverlayConfig,
         discovery: Arc<Discovery>,
         utp_listener_tx: UnboundedSender<UtpListenerRequest>,
-        utp_listener_rx: Arc<TRwLock<UnboundedReceiver<UtpListenerEvent>>>,
         storage: Arc<RwLock<PortalStorage>>,
         data_radius: U256,
         protocol: ProtocolId,
@@ -124,7 +120,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             Arc::clone(&data_radius),
             protocol.clone(),
             utp_listener_tx.clone(),
-            utp_listener_rx,
             config.enable_metrics,
         )
         .await
@@ -285,13 +280,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         enr: Enr,
         conn_id: u16,
     ) -> Result<Content, OverlayRequestError> {
-        let utp_request = UtpListenerRequest::FindContentStream(conn_id);
-        if let Err(err) = self.utp_listener_tx.send(utp_request) {
-            return Err(OverlayRequestError::UtpError(format!(
-                "Unable to send uTP FindContent stream request: {err}"
-            )));
-        }
-
         // initiate the connection to the acceptor
         let (tx, rx) = tokio::sync::oneshot::channel::<anyhow::Result<UtpSocket>>();
         self.utp_listener_tx
@@ -399,11 +387,6 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         // Do not initialize uTP stream if remote node doesn't have interest in the offered content keys
         if response.content_keys.is_zero() {
             return Ok(response);
-        }
-
-        let utp_request = UtpListenerRequest::OfferStream(conn_id);
-        if let Err(err) = self.utp_listener_tx.send(utp_request) {
-            return Err(anyhow!("Unable to send uTP Offer stream request: {err}"));
         }
 
         // initiate the connection to the acceptor
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -19,7 +19,7 @@ use crate::{
     portalnet::types::content_key::RawContentKey,
     utp::{
         stream::{UtpListenerEvent, UtpListenerRequest, UtpSocket, BUF_SIZE},
-        trin_helpers::{UtpAccept, UtpMessage},
+        trin_helpers::{UtpAccept, UtpMessage, UtpStreamId},
     },
 };
 use discv5::{
@@ -299,6 +299,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
                 conn_id,
                 enr.node_id(),
                 self.protocol.clone(),
+                UtpStreamId::FindContentStream,
                 tx,
             ))
             .map_err(|err| {
@@ -412,6 +413,7 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
             conn_id,
             enr.node_id(),
             self.protocol.clone(),
+            UtpStreamId::OfferStream,
             tx,
         )).map_err(|err| anyhow!("Unable to send Connect request to UtpListener when processing ACCEPT message: {err}"))?;
 
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -18,7 +18,7 @@ use crate::portalnet::{
 use crate::{
     portalnet::types::content_key::RawContentKey,
     utp::{
-        stream::{UtpListenerRequest, UtpSocket, BUF_SIZE},
+        stream::{UtpListenerRequest, UtpStream, BUF_SIZE},
         trin_helpers::{UtpAccept, UtpMessage, UtpStreamId},
     },
 };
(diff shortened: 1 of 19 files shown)


trin-core/src/portalnet/types/content_key.rs
```diff
@@ -6,6 +6,9 @@ use ssz::{self, Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use ssz_types::{typenum, FixedVector, VariableList};
 
+/// Raw overlay content key
+pub type RawContentKey = Vec<u8>;
```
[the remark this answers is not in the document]
ogenev: SSZ encoded Bytelist, maybe it is better to use `Bytelist` instead of this type alias?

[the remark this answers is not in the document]
njgheorghita: Hmmm, either way works for me. Maybe just update the description for the type alias

njgheorghita (approved):
👍

jacobkaufmann (commented):
I'm generally opposed to this sort of approach, where we send messages to `UtpListener` to "ask" for closed uTP streams. In my view, it adds both unnecessary complexity and unnecessary resource consumption.

The approach I would suggest is similar to the way `Discv5` emits `Discv5Event` to a channel that we can stream. Upon socket close, `UtpListener` can emit some message to a channel where the message contains the socket info and any data that the socket received. `OverlayService` can listen to this channel to handle those messages.

ogenev:
> The approach I would suggest is similar to the way `Discv5` emits `Discv5Event` to a channel that we can stream. Upon socket close, `UtpListener` can emit some message to a channel where the message contains the socket info and any data that the socket received. `OverlayService` can listen to this channel to handle those messages.

Yeah, this seems a more robust way to handle the closed streams.
However, this will require an event channel per network because if we emit a global overlay event, then only one of the overlay networks will consume it because we will run multiple overlay protocols. 

Also, the connection state is closed inside `UtpSocket` and I think at least for now, we will still need to process all active sockets in a set time interval in `UtpListener` to check which one is closed and then emit the event. Another alternative is to add event channels between `UtpSocket` and `UtpListener` but this will bring more complexity.

One major issue I realized today is that now `UtpListener` doesn't recognize which stream for which network is related and we can't emit specific events for a specific network.

I will try to refactor a bit and first store all active uTP connections in `UtpListener` in a separate `HashMap` per network and then try to pass state and history event receivers to the `OverlayService` struct and emit an event when a connection is closed (after `UtpListener` process all active connections to check which one is closed).

 As we are using the same `OverlayService` struct for every network, probably I should pass those event receivers as `Option<UnboundedReceiver<...>>`, because we may not always want to run all the networks, do you have any better idea?
 
 Let me know what you think and if this makes sense to you!

ogenev:
Good, I updated the PR description with some TODOs and will re-open it for review when all of them are done.

trin-core/src/utp/stream.rs
```diff
@@ -103,30 +111,51 @@ struct DelayDifferenceSample {
...
-    OfferStream(ConnId),
+    /// Request to initialize uTP stream with remote node
+    Connect(
+        ConnId,
+        NodeId,
+        ProtocolId,
+        UtpStreamId,
+        oneshot::Sender<anyhow::Result<UtpStream>>,
+    ),
+    /// Request to add uTP stream to the active connections
```
mrferris, line 122:
```suggestion
    /// Request to initiate and add uTP stream to the active connections
```

trin-core/src/utp/trin_helpers.rs
```diff
@@ -51,14 +51,17 @@ pub struct UtpAccept {
     pub message: Vec<(Vec<u8>, Vec<u8>)>,
 }
 
-// This is not in a spec, this is just for internally tracking for what portal message
-// negotiated the uTP stream
-#[derive(Debug, Clone)]
-pub enum UtpMessageId {
+/// Used to track which stream to which overlay request correspond
```
mrferris, line 54:
```suggestion
/// Used to track which stream an overlay request corresponds with
```

trin-core/src/utp/trin_helpers.rs
```diff
@@ -51,14 +51,17 @@ pub struct UtpAccept {
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum UtpStreamId {
```
mrferris, line 56: The name `UtpStreamId` is a bit confusing to me, I think because `ConnId` is already acting as an identifier for streams.
I wonder if `UtpStreamType` or `UtpStreamOrigin` would be a more descriptive name.

trin-core/src/utp/stream.rs
```diff
@@ -103,30 +111,51 @@ struct DelayDifferenceSample {
...
+    AddActiveConnection(Enr, ProtocolId, UtpStreamId, ConnId),
```
mrferris, line 123: I wonder if `InitiateConnection` would be a more descriptive name for this event.

The name `AddActiveConnection` is a little bit confusing to me because it seems like both `Connect` and `AddActiveConnection` events add an active connection to the `utp_connections`. In fact, `Connect` seems like it creates more of an "active" connection than `AddActiveConnection` does, because by the time a `Connect` event happens we necessarily have two participants in the stream, whereas an `AddActiveConnection` might only ever result in an inactive connection if the other peer never connects.

mrferris (commented):
👍 to the architecture decision & implementation that resulted from the above conversation.

I left some suggestions about comments and naming. Let me know if I've misunderstood anything.

trin-core/src/utp/stream.rs
```diff
@@ -210,73 +249,65 @@ impl UtpListener {
...
-                            warn!("Query requested an unknown ENR");
+                            warn!("Received SYN packet for an unknown active uTP stream");
                         }
                     }
                     // Receive DATA and FIN packets
                     PacketType::Data => {
                         if let Some(conn) = self
                             .utp_connections
-                            .get_mut(&ConnectionKey::new(*node_id, connection_id))
+                            .get_mut(&ConnectionKey::new(*node_id, connection_id - 1))
```
jacobkaufmann, line 284: Why the minus one here?

trin-core/src/utp/stream.rs
```diff
@@ -210,73 +249,65 @@ impl UtpListener {
...
                             }
+                        } else {
+                            warn!("Received DATA packet for an unknown active uTP stream")
                         }
                     }
                     PacketType::Fin => {
                         if let Some(conn) = self
                             .utp_connections
-                            .get_mut(&ConnectionKey::new(*node_id, connection_id))
+                            .get_mut(&ConnectionKey::new(*node_id, connection_id - 1))
```
jacobkaufmann, line 310: Same question as above regarding the minus one.

trin-core/src/utp/stream.rs
```diff
@@ -58,10 +58,18 @@ const BASE_HISTORY: usize = 10; // base delays history size
...
+// const PROCESS_UTP_STREAMS_INTERVAL: u64 = 20;
 
 /// uTP connection id
 type ConnId = u16;
 
+/// uTP payload data
+pub type UtpPayload = Vec<u8>;
+
+/// UtpListener unbounded receiver for emitted events
+pub type UtpListenerUnboundedReceiver = Arc<RwLock<UnboundedReceiver<UtpListenerEvent>>>;
```
jacobkaufmann, line 71: I sort of prefer no alias, so that it's clear that we are passing around a `Arc<RwLock<..>>`, and because we only use this once as far as I can tell. Just a personal preference though.

jacobkaufmann (commented):
Looking good. I feel that there are some comments that we ought to address before merge, but most of my comments are minor or questions about things I don't quite understand. Would like to get clarity on those questions before approval.

In the future, we should minimize refactor changes like naming that add a lot of non-logical line changes to the diff. If possible, deferring those changes can make the PR less noisy.

trin-core/src/utp/stream.rs
```diff
@@ -103,30 +111,51 @@ struct DelayDifferenceSample {
...
-    AcceptStream(ConnId, Vec<Vec<u8>>),
-    /// Request to initialize uTP streram with remote node
-    Connect(ConnId, NodeId, oneshot::Sender<anyhow::Result<UtpSocket>>),
-    /// Request to listen for FindCOntent stream and send content data
-    FindContentData(ConnId, ByteList),
-    /// Request to listen for FindContent stream
-    FindContentStream(ConnId),
-    /// Request to listen for Offer stream
-    OfferStream(ConnId),
+    /// Request to initialize uTP stream with remote node
```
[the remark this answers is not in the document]
ogenev: Changed this to `Request to create and connect to a uTP stream initiated by a remote node` as this is used to create an uTP stream AND to connect to the remote peer.

trin-core/src/utp/trin_helpers.rs
ogenev: `ConnId` is not really an identifier for streams in the portal network context, it is identifying all the packets that belong to the same connection. Each uTP socket has one connection ID for sending packets and a different connection ID for receiving packets.

Anyways, I'm happy to rename this if it will improve code readability, what is your opinion @jacobkaufmann?

trin-core/src/utp/stream.rs
ogenev: You are right, the idea of `AddActiveConnection` is to create an uTP stream and start listening for incoming SYN packets.
`Connect` is creating an uTP stream and is sending SYN packet for a handshake with the remote peer.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -417,6 +421,30 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
     }
 
+    /// Send request to UtpLister to add an uTP stream to the active connections
+    fn add_utp_connection(
+        &self,
+        source: &NodeId,
+        conn_id_recv: u16,
+        stream_id: UtpStreamId,
+    ) -> Result<(), OverlayRequestError> {
+        if let Some(enr) = self.discovery.discv5.find_enr(source) {
```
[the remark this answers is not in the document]
ogenev: Nice catch! We should check the overlay routing table here.  Added an overlay `find_enr` method similar to discv5's.

I think it is ok to do the check inside this method as we need the `Enr` only for sending uTP packets over discv5 as part of the uTP stream. We are calling it from `handle_find_content` and `handle_offer` which have knowledge only for the `NodeId` of the request.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -45,6 +46,8 @@ pub const FIND_CONTENT_MAX_NODES: usize = 32;
 /// With even distribution assumptions, 2**17 is enough to put each node (estimating 100k nodes,
 /// which is more than 10x the ethereum mainnet node count) into a unique bucket by the 17th bucket index.
 const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
+/// Bucket refresh lookup interval in seconds
+const BUCKET_REFRESH_INTERVAL: u64 = 60;
```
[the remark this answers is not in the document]
ogenev: Renamed. It is configurable with a constant for now :smile_cat:. Seriously, not sure if we need to add this to the CLI but I think it should not be part of this PR.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -523,7 +550,12 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         debug!("[{:?}] Initializing request", self.protocol);
         match request {
             Request::FindContent(find_content) => {
-                Ok(Response::Content(self.handle_find_content(find_content)?))
+                // TODO: Creating random NodeID here just to satisfy handle_find_content signature
+                // is a hack on top of the history RecursiveFindContent temporally hack.
+                let node_id = NodeId::random();
+                Ok(Response::Content(
+                    self.handle_find_content(find_content, &node_id)?,
+                ))
```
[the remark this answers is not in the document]
ogenev: It is just a nasty hack to make the code compile without refactoring the current hacky `RecursiveFindContent` json-rpc endpoint as it is going to be reimplemented soon. It is not expected currently to work recursively with uTP anyways, so we are not breaking anything that is working now :)

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -417,6 +421,30 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                stream_id,
+                conn_id_recv,
+            );
+            if let Err(err) = self.utp_listener_tx.send(utp_request) {
+                return Err(OverlayRequestError::UtpError(format!(
+                    "Unable to send uTP AddActiveConnection request: {err}"
+                )));
+            }
+        }
+        Ok(())
```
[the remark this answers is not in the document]
ogenev: FIxed.

trin-core/src/portalnet/overlay.rs
```diff
@@ -277,19 +280,21 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
-            .send(UtpListenerRequest::Connect(conn_id, enr.node_id(), tx));
+        let (tx, rx) = tokio::sync::oneshot::channel::<anyhow::Result<UtpStream>>();
+        self.utp_listener_tx
+            .send(UtpListenerRequest::Connect(
+                conn_id,
+                enr.node_id(),
+                self.protocol.clone(),
+                UtpStreamId::FindContentStream,
+                tx,
+            ))
```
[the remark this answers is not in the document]
ogenev: It seems `rustfmt` is formating it this way...

trin-core/src/utp/stream.rs
ogenev: This is because we are expecting to receive a SYN packet with the random connection id (conn_id) generated by the acceptor node (us) and this is the value we are using to create `ConnectionKey` to add the stream to the active connections hashmap (when handling OFFER message for example).

However, when we receive DATA or FIN packets, we are expecting a packet to have `conn_id + 1` value (according to uTP [specs](https://www.bittorrent.org/beps/bep_0029.html)). So if we want to check if the packet is part of an active uTP stream, when handling those packets, we have to generate a `ConnectionKey` using the` packet.connection_id - 1`.

The good point here is that we should handle the overflow and use `wrapped_sub(1)`

trin-core/src/utp/stream.rs
jacobkaufmann: Hmm, that's annoying. maybe we can use

```rust
pub type UnboundedUtpEventReceiver = UnboundedReceiver<UtpListenerEvent>;
```

and then use `Arc<RwLock<UnboundedUtpEventReceiver>>`.

If not, then this is okay.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
jacobkaufmann: We would probably also add the "ENR cache" to the search after we implement that.

trin-core/src/utp/stream.rs
jacobkaufmann: Hmm I see. Is there a test case that we can add for this logic?

trin-core/src/portalnet/overlay.rs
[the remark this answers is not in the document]
ogenev: Ah, I didn't know that this is what you want, fixed.

trin-core/src/utp/stream.rs
ogenev: Actually, we don't need to wrap this in Arc<RwLock<..>> anymore, so I removed the alias :partying_face:.

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
ogenev: Not sure if we have a use case to make it configurable right now but this line is in the diff because when changing the code in the previous version of the PR, I decided to extract it for better readability. I'll be more careful when renaming or making such unrelated changes in the future :innocent: .

This constant is about the overlay bucket refresh lookup logic which is out of the scope of this PR.

trin-core/src/utp/stream.rs
ogenev: Added an integration test for the uTP listener events that catches such types of changes.

trin-core/src/utp/trin_helpers.rs
mrferris: Yea sorry I wasn't making sense by conflating the terms stream and connection.

Passing something non-numeric called `Id` in the same calls as a numeric `Id` is what seemed off to me.

Though looking back, I think I'm splitting hairs. I'm good with leaving it as it is.

trin-core/src/utp/stream.rs
```diff
@@ -210,73 +248,65 @@ impl UtpListener {
...
-                                .insert(conn.sender_connection_id, UtpMessageId::FindContentStream);
-
-                            if let Some(UtpMessageId::FindContentData(Content(content_data))) =
-                                utp_message_id
+                            // Send content data if the stream is listening for FindContent SYN packet
+                            if let UtpStreamId::FindContentData(content_data) =
+                                conn.stream_id.clone()
                             {
                                 // We want to send uTP data only if the content is Content(ByteList)
-                                tokio::spawn(async move {
```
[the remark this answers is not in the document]
ogenev: No need for this at the moment, spawning a new task here requires additional refactoring and adds more complexity.  If it causes performance issues, we will change this in the future.

PR #327, opened 2022-05-16
Add FindContentQuery

jacobkaufmann:
### What was wrong?

`trin` lacked a definition for a content lookup query.

### How was it fixed?

- Define `FindContentQuery`, which manages query state for a content lookup.
- Rename `FindNodeQueryConfig` to `QueryConfig` to share between `FindNodeQuery` and `FindContentQuery`.
- Move `QueryConfig` to `query.rs`.
- Add Parity copyright to `query.rs`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/find/iterators/findcontent.rs
+++ b/trin-core/src/portalnet/find/iterators/findcontent.rs
@@ -29,7 +29,7 @@ use discv5::kbucket::{Distance, Key};
 use super::super::query_pool::QueryState;
 use super::query::{Query, QueryConfig, QueryPeer, QueryPeerState, QueryProgress};
 
-pub enum FindContentResponse<TNodeId> {
+pub enum FindContentQueryResponse<TNodeId> {
     ClosestNodes(Vec<TNodeId>),
     Content(Vec<u8>),
 }
@@ -74,7 +74,7 @@ pub struct FindContentQuery<TNodeId> {
     config: QueryConfig,
 }
 
-impl<TNodeId> Query<TNodeId, FindContentResponse<TNodeId>, FindContentQueryResult<TNodeId>>
+impl<TNodeId> Query<TNodeId, FindContentQueryResponse<TNodeId>, FindContentQueryResult<TNodeId>>
     for FindContentQuery<TNodeId>
 where
     TNodeId: Into<Key<TNodeId>> + Eq + Clone,
@@ -91,7 +91,7 @@ where
         self.started = Some(start);
     }
 
-    fn on_success(&mut self, peer: &TNodeId, peer_response: FindContentResponse<TNodeId>) {
+    fn on_success(&mut self, peer: &TNodeId, peer_response: FindContentQueryResponse<TNodeId>) {
         if let QueryProgress::Finished = self.progress {
             return;
         }
@@ -126,7 +126,7 @@ where
 
         // Incorporate the peer response into the query.
         match peer_response {
-            FindContentResponse::ClosestNodes(closer_peers) => {
+            FindContentQueryResponse::ClosestNodes(closer_peers) => {
                 // Incorporate the reported closer peers into the query.
                 let mut progress = false;
                 let num_closest = self.closest_peers.len();
@@ -163,7 +163,7 @@ where
                     QueryProgress::Finished => QueryProgress::Finished,
                 };
             }
-            FindContentResponse::Content(content) => {
+            FindContentQueryResponse::Content(content) => {
                 // Incorporate the found content into the query.
                 self.content = Some(ContentAndPeer {
                     content,
@@ -521,7 +521,7 @@ mod tests {
                         if rng.gen_bool(0.05) {
                             query.on_success(
                                 k.preimage(),
-                                FindContentResponse::Content(found_content.clone()),
+                                FindContentQueryResponse::Content(found_content.clone()),
                             );
                             content_peer = Some(k.clone());
                         } else {
@@ -530,7 +530,7 @@ mod tests {
                             remaining.extend(closer_peers.iter().map(|x| Key::from(*x)));
                             query.on_success(
                                 k.preimage(),
-                                FindContentResponse::ClosestNodes(closer_peers),
+                                FindContentQueryResponse::ClosestNodes(closer_peers),
                             );
                         }
                     } else {
@@ -617,16 +617,25 @@ mod tests {
                 panic!("No peer.");
             };
 
-            query.on_success(&peer1, FindContentResponse::ClosestNodes(closer.clone()));
+            query.on_success(
+                &peer1,
+                FindContentQueryResponse::ClosestNodes(closer.clone()),
+            );
 
             // Duplicate result from the same peer.
-            query.on_success(&peer1, FindContentResponse::ClosestNodes(closer.clone()));
+            query.on_success(
+                &peer1,
+                FindContentQueryResponse::ClosestNodes(closer.clone()),
+            );
 
             // If there is a second peer, let it also report the same "closer" peer.
             match query.poll(now) {
                 QueryState::Waiting(Some(p)) => {
                     let peer2 = p;
-                    query.on_success(&peer2, FindContentResponse::ClosestNodes(closer.clone()))
+                    query.on_success(
+                        &peer2,
+                        FindContentQueryResponse::ClosestNodes(closer.clone()),
+                    )
                 }
                 QueryState::Finished => {}
                 _ => panic!("Unexpectedly query state."),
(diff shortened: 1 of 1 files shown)


mrferris (approved):
Looks good to me!

`FindContentQuery` looks well thought-out to later be able to handle POKEs after a content query is complete. 

Tests are well-adapted to hit the possible cases. 

And combining the identical configs into `QueryConfig` is 👍

trin-core/src/portalnet/find/iterators/query.rs
```diff
@@ -1,9 +1,68 @@
...
+    /// result is delivered. Defaults to `10` seconds.
+    pub peer_timeout: Duration,
+}
+
+impl QueryConfig {
+    pub fn default() -> Self {
+        Self {
+            parallelism: 3,
+            num_results: 20,
+            peer_timeout: Duration::from_secs(10),
```
njgheorghita, line 61: Just curious why 10 was chosen for the default timeout?

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+
+use std::collections::btree_map::{BTreeMap, Entry};
+use std::time::Instant;
+
+use discv5::kbucket::{Distance, Key};
+
+use super::super::query_pool::QueryState;
+use super::query::{Query, QueryConfig, QueryPeer, QueryPeerState, QueryProgress};
+
+pub enum FindContentResponse<TNodeId> {
```
njgheorghita, line 32: *nit* Iiuc, this name is slightly confusing. Technically this represents a `Content` response to a `FindContent` query right? I guess `FindContentQueryResponse` is more explicit but wordy - but if this struct is limited to this module it makes more sense within this context. Do these structs (& `FindContentQueryResult` need to be `pub`)?

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+        if let QueryProgress::Finished = self.progress {
+            return;
+        }
+
+        let key: Key<TNodeId> = peer.clone().into();
+        let distance = key.distance(&self.target_key);
+
+        match self.closest_peers.entry(distance) {
+            Entry::Vacant(_) => {}
+            Entry::Occupied(mut e) => match e.get().state() {
```
njgheorghita, line 185: *nit* I'm not a fan of single letter variables - imo `entry` would be better here.

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+            // and the query is not waiting for any more results.
+            self.progress = QueryProgress::Finished;
+            QueryState::Finished
+        }
+    }
+
+    fn into_result(self) -> FindContentQueryResult<TNodeId> {
+        match self.content {
+            Some(content) => {
+                let closest_nodes = self
```
njgheorghita, line 292: This expression for getting `closest_nodes` and the one below in `None` seem to be almost identical - have you tried combining them into a shared method?

trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -232,11 +196,13 @@ where
                         return QueryState::WaitingAtCapacity;
                     }
                 }
-
                 QueryPeerState::Waiting(timeout) => {
                     if now >= *timeout {
                         // Peers that don't respond within timeout are set to `Failed`.
-                        debug_assert!(self.num_waiting > 0);
+                        assert!(
```
njgheorghita, line 202: Also noticed the usage of `assert!` in `FindContentQuery` - why the decision to panic if a query is unsuccessful / buggy instead of something like a failed query state that's handled?

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+
+use discv5::kbucket::{Distance, Key};
+
+use super::super::query_pool::QueryState;
+use super::query::{Query, QueryConfig, QueryPeer, QueryPeerState, QueryProgress};
+
+pub enum FindContentResponse<TNodeId> {
+    ClosestNodes(Vec<TNodeId>),
+    Content(Vec<u8>),
+}
```
njgheorghita, line 35: We don't seem to handle `connection_id` `Content` responses? Is there a plan for how to incorporate utp?

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+
+        let key: Key<TNodeId> = peer.clone().into();
+        let distance = key.distance(&self.target_key);
+
+        // Mark the peer's progress.
+        match self.closest_peers.entry(distance) {
+            Entry::Vacant(..) => return,
+            Entry::Occupied(mut entry) => match entry.get().state() {
+                QueryPeerState::Waiting(..) => {
+                    assert!(
```
njgheorghita, line 107: It looks like the 3 `assert!` statements in `FindContentQuery` are identical - seems like it could be pulled out into a shared method. But it also seems like a slightly more informative message could be useful. Just thinking through if we see a panic while running this live, messages that would help immediately differentiate between the 3 different fail cases could be nice to have.

njgheorghita (approved):
Lookin' good! Mostly style nits and general questions from me

trin-core/src/portalnet/find/iterators/findcontent.rs
jacobkaufmann: They are similar but not identical. If the content is `Some(..)`, then we make sure to filter out the peer who returned the content. I could make this a shared method with an argument for a (possibly empty) set of nodes to filter out, but I did not think there was a huge benefit doing that.

trin-core/src/portalnet/find/iterators/findcontent.rs
```diff
@@ -0,0 +1,715 @@
...
+        }
+    }
+}
+
+impl<TNodeId> FindContentQuery<TNodeId>
+where
+    TNodeId: Into<Key<TNodeId>> + Eq + Clone,
+{
+    /// Creates a new query with the given configuration.
+    pub fn with_config<I>(
```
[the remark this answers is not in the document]
jacobkaufmann: To me, `from_config` implies that the constructor only takes a single argument of type `QueryConfig`, like the trait `From<QueryConfig>`. Here, `with_config` sort of denotes that the `QueryConfig` is not the only argument.

I'm not a big fan of the name either, but I think it's okay for now. This naming is also consistent with the `FindNodeQuery` naming.

trin-core/src/portalnet/find/iterators/query.rs
jacobkaufmann: This was just carried over from the pre-existing `FindNodeQueryConfig`. We may want to bring this down. Looks like the [`sigp/discv5` query peer timeout](https://github.com/sigp/discv5/blob/6fa9bf62c98c0d4a34648429957fbf07dce5a2f8/src/config.rs#L120) is 2 seconds.

trin-core/src/portalnet/find/iterators/findcontent.rs
jacobkaufmann: Yes, this represents a `Content` response. Yes, these structs need to be `pub`, because those types will be exposed/used on calls to `into_result` and `on_success`.

I don't mind verbosity if it makes things easier to understand. The reason that we create a new type here is that `Content` also contains a `ConnectionId` variant, which we will not use here.

trin-core/src/portalnet/find/iterators/findcontent.rs
jacobkaufmann: No, because uTP is not relevant to the query state machine. The query state machine only needs to know whether it holds the desired content or not. Content transmitted via uTP for some query would result in a call to `on_success` with that content.

Please let me know if that does not make sense. I can elaborate further.

trin-core/src/portalnet/find/iterators/findnodes.rs
jacobkaufmann: This logic was carried over from `sigp/discv5`, albeit with `debug_assert`, which I think we can change back to. My understanding is that this type of error could represent a non-recoverable error, because if it can happen for one query, it can happen for many queries.

If we did not have the assertion, and we had `num_waiting == 0`, then we would panic on 'attempt to subtract with overflow'. So it seems that this assertion is a more useful panic message in the case that we run into such a situation.

I think we can assert that this is something that should never happen, because it represents a bug in our own logic. What do you think? If we keep the assertion, should go back to `debug_assert`? @njgheorghita

PR #328, opened 2022-05-17
[WIP} Propagate gossip accepted content

ogenev:
### To-Do

- [ ] Clean up commit history


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1031,6 +1148,58 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                .into_iter()
+                .filter(|content| {
+                    XorMetric::distance(&content.0.content_id(), &node.key.preimage().raw())
+                        < node.value.data_radius
+                })
+                .map(|content| content.0.into())
+                .collect();
+
+            if interested_content.is_empty() {
+                continue;
```
pipermerriam, line 1182: This seems like a good place for a logging message to be emitted:

> No nodes interested in neighborhood gossip:  content_key=XXXX  num_nodes_checked=NNNN

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            .buckets_iter()
+            .map(|kbucket| {
+                kbucket
+                    .iter()
+                    .collect::<Vec<&kbucket::Node<NodeId, Node>>>()
+            })
+            .flatten()
+            .collect();
+
+        // HashMap to temporally store all interested nodes and the content
```
mrferris, line 299:
```suggestion
        // HashMap to temporarily store all interested nodes and the content. Key is base64 string of node's ENR.
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                nodes_and_content
+                    .entry(enr.to_base64())
+                    .or_default()
+                    .push(content.clone().0.into());
+            }
+        }
+
+        // Create and send OFFER overlay request to the interested nodes
+        for (nodes, interested_content) in nodes_and_content.drain() {
+            let enr = match Enr::from_str(&nodes) {
```
mrferris, line 335:
```suggestion
        for (node, interested_content) in nodes_and_content.drain() {
            let enr = match Enr::from_str(&node) {
```
```suggestion
        for (nodes, interested_content) in nodes_and_content.drain() {
            let enr = match Enr::from_str(&nodes) {
```

mrferris, line 335:
```suggestion
        for (enr_string, interested_content) in nodes_and_content.drain() {
            let enr = match Enr::from_str(&enr_string) {
```

trin-core/src/portalnet/overlay.rs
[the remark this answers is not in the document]
mrferris: Can you not use the ENR itself as the key in the HashMap? I'm guessing there's a reason you're using the base64 string instead.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        match event {
+            UtpListenerEvent::ClosedStream(utp_payload, _, stream_id) => {
+                // Handle all closed uTP streams, currently we process only AcceptStream here.
+                // FindContent payload is processed explicitly when we send FindContent request.
+                if let AcceptStream(content_keys) = stream_id {
+                    self.process_accept_utp_payload(content_keys, utp_payload)?;
+                }
+            }
+            UtpListenerEvent::ResetStream(..) => {
+                // TODO: Handle reset streams
```
jacobkaufmann, line 203: Was this `TODO` left here purposefully? If so, why can't we handle this now?

mrferris (changes requested):
This is exciting functionality 🚀

In general looks great, but requesting small changes for where I think that the `TContentKey` should be used instead of the `match` in `process_accept_utp_payload`. Let me know if I've misunderstood the reason for that.

I'm also wondering about how this was tested, and your thoughts on how we might add unit/integration tests.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            ProtocolId::History => {
+                // TODO: Verify history content data with Oracle
+
+                // Temporally store content key/value pairs to propagate here
+                let mut content_keys_values: Vec<(HistoryContentKey, ByteList)> = Vec::new();
+
+                for (content_key, content_value) in content_keys.into_iter().zip(content_values) {
+                    let history_key = HistoryContentKey::try_from(content_key);
+
+                    match history_key {
```
jacobkaufmann, line 240: nit: we can combine these lines.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        // Accepted content keys len should match content value len
+        if content_keys.len() != content_values.len() {
+            return Err(anyhow!(
+                "Content keys len doesn't match content values len."
+            ));
+        }
+
+        // Try to store accepted content into the database and propagate gossip the content
+        match self.protocol {
+            ProtocolId::History => {
```
jacobkaufmann, line 231: This match is unnecessary. We can use `TContentKey::try_from`.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                // Propagate gossip accepted content
+                self.propagate_gossip(content_keys_values);
+            }
+            _ => return Err(anyhow!("Unsupported content key protocol.")),
+        };
+        Ok(())
+    }
+
+    /// Propagate gossip accepted content via OFFER/ACCEPT:
+    fn propagate_gossip(&self, accepted_content: Vec<(impl OverlayContentKey, ByteList)>) {
```
jacobkaufmann, line 286:
```suggestion
    fn propagate_gossip(&self, accepted_content: Vec<(TContentKey, ByteList)>) {
```

trin-core/src/portalnet/overlay.rs
jacobkaufmann, line 299:
```suggestion
        // HashMap to temporarily store all interested nodes and the content
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                .filter(|node| {
+                    XorMetric::distance(&content.0.content_id(), &node.key.preimage().raw())
+                        < node.value.data_radius()
+                })
+                .map(|node| node.value.enr())
+                .collect();
+
+            // Continue if no nodes are interested in the content
+            if interested_nodes.is_empty() {
+                debug!("No nodes interested in neighborhood gossip: content_key={} num_nodes_checked={}", hex_encode(content.0.into()), all_nodes.len());
```
jacobkaufmann, line 316: nit: this line seems long, and I'm not sure why `cargo fmt` did not pick this up.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                continue;
+            }
+
+            // Get log2 random enrs to gossip
+            let random_nodes = log2_random_enrs(interested_nodes);
+
+            // Temporally Store all random selected nodes with interested content.
```
jacobkaufmann, line 323:
```suggestion
            // Temporarily store all randomly selected nodes with the content of interest.
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+
+    /// Try to store accepted content into database
+    fn store_accepted_content(&self, content_key: &impl OverlayContentKey, value: ByteList) {
+        match self.storage.read().should_store(content_key) {
+            Ok(_) => {
+                if let Err(err) = self.storage.write().store(content_key, &value.into()) {
+                    warn!("Unable to store accepted content: {err}");
+                }
+            }
+            Err(err) => error!("Should_store error on accepted content: {err}"),
```
jacobkaufmann, line 367:
```suggestion
            Err(err) => error!("Unable to determine whether to store accepted content: {err}"),
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            );
+
+            if let Err(err) = self.request_tx.send(overlay_request) {
+                error!("Unable to send OFFER request to overlay: {err}.")
+            }
+        }
+    }
+
+    /// Try to store accepted content into database
+    fn store_accepted_content(&self, content_key: &impl OverlayContentKey, value: ByteList) {
```
jacobkaufmann, line 360:
```suggestion
    fn store_accepted_content(&self, content_key: &TContentKey, value: ByteList) {
```

trin-core/src/portalnet/types/content_key.rs
```diff
@@ -1,14 +1,39 @@
...
+use std::ops::Deref;
 
 /// SSZ encoded overlay content key as bytes
 pub type RawContentKey = Vec<u8>;
 
+/// uTP content payload list represented as SSZ encoded Container[payloads: List[ByteList, max_length=64]]
+#[derive(Clone, Debug, Encode, Decode)]
+pub struct ContentPayloadList {
+    payloads: Vec<ByteList>,
+}
```
jacobkaufmann, line 18: We can replace with [`VariableList`](https://docs.rs/eth2_ssz_types/0.2.1/ssz_types/struct.VariableList.html).

trin-core/src/portalnet/overlay.rs
```diff
@@ -490,6 +675,24 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
     }
 }
 
+/// Randomly select log2 nodes. log2() is stable only for floats, that's why we cast it first to f32
+fn log2_random_enrs(enrs: Vec<Enr>) -> Vec<Enr> {
+    assert!(!enrs.is_empty());
```
jacobkaufmann, line 680: If there is a panic here, then we should document that in doc comment. I don't think an empty input should be a panic. We can return `Option<Vec<Enr>>` instead. Alternatively, we can return the input if the input is empty.

trin-core/src/portalnet/overlay.rs
jacobkaufmann, line 335: Ideally we don't have to perform serialization to and from base 64, but I understand why it's being done. There may be a cleaner way, but it's not obvious to me right now.

jacobkaufmann (changes requested):
Looks good, but there are a few things we should change. We should use the generic type parameter for the content key. We should also use a `VariableList` SSZ type if that list has a maximum length. Most other comments are minor.

trin-core/src/portalnet/overlay.rs
carver, line 203: Add an issue that's part of a milestone. Alpha?

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
-            self.protocol
-        );
+    pub fn process_utp_event(&self, event: UtpListenerEvent) -> anyhow::Result<()> {
+        match event {
+            UtpListenerEvent::ClosedStream(utp_payload, _, stream_id) => {
+                // Handle all closed uTP streams, currently we process only AcceptStream here.
+                // FindContent payload is processed explicitly when we send FindContent request.
+                if let AcceptStream(content_keys) = stream_id {
+                    self.process_accept_utp_payload(content_keys, utp_payload)?;
+                }
```
carver, line 200: These seems to be an easy place to accidentally forget to process something. Maybe a good option would be a match that explicitly does nothing on the `FindContent`. Later, if there's a new entry in the enum, then we get a compiler error here for a missing branch in the match. That way we are forced to consider whether to process it here or not.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        if content_keys.len() != content_values.len() {
+            return Err(anyhow!(
+                "Content keys len doesn't match content values len."
+            ));
+        }
+
+        // Try to store accepted content into the database and propagate gossip the content
+        match self.protocol {
+            ProtocolId::History => {
+                // TODO: Verify history content data with Oracle
```
carver, line 232: Add an issue that's part of the alpha milestone?

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                            warn!("Unable to decode history content key: {err}");
+                            continue;
+                        }
+                    }
+                }
+                // Propagate gossip accepted content
+                self.propagate_gossip(content_keys_values);
+            }
+            ProtocolId::State => {
+                // TODO: Verify state content data with Oracle
```
carver, line 256: Add an issue, but no milestone defined for state data.

trin-core/src/portalnet/types/content_key.rs
carver, line 18: Is this the only non-key in the file? Seems like it either belongs in a different file, or this file needs a rename.

trin-core/src/portalnet/overlay.rs
```diff
@@ -490,6 +675,24 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+    // log2(1) is zero but we want to propagate to at least one node
+    if enrs.len() == 1 {
+        return enrs;
+    }
+
+    // Greedy gossip by ceiling the log2 value
+    let log2_value = (enrs.len() as f32).log2().ceil();
+
+    let random_enrs: Vec<Enr> = enrs
+        .clone()
```
carver, line 690: I'm still slowly building Rust intuition: why do we need to clone here? I would have thought that with ownership of `enrs` we could destroy it with `into_iter()` and not have a problem.

trin-core/src/portalnet/overlay.rs
```diff
@@ -490,6 +675,24 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
     }
 }
 
+/// Randomly select log2 nodes. log2() is stable only for floats, that's why we cast it first to f32
+fn log2_random_enrs(enrs: Vec<Enr>) -> Vec<Enr> {
```
carver, line 679: My guess is that we might want to do something different later, but this seems like a fine starting point. IIRC, geth uses sqrt instead as a kind of preference for over-propagating. There also might be even more special cases when only connected to a small number of peers. We'll see as we experiment with the network...

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+                "Content keys len doesn't match content values len."
+            ));
+        }
+
+        // Try to store accepted content into the database and propagate gossip the content
+        match self.protocol {
+            ProtocolId::History => {
+                // TODO: Verify history content data with Oracle
+
+                // Temporally store content key/value pairs to propagate here
```
carver, line 234:
```suggestion
                // Temporarily store content key/value pairs to propagate here
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+
+                    match history_key {
+                        Ok(key) => {
+                            // Store accepted content in DB
+                            self.store_accepted_content(&key, content_value.clone());
+                            content_keys_values.push((key, content_value))
+                        }
+                        Err(err) => {
+                            warn!("Unable to decode history content key: {err}");
+                            continue;
```
carver, line 248: If any one piece of data is invalid above, I expect we might drop the whole stream. So it stands out that this is weirdly inconsistent, that it would only individually drop malformed keys. Also, this is especially weird right? Shouldn't these be the keys that we already sent ourselves in an ACCEPT? It would be highly strange for them to be undecodable. I think this error message doesn't convey the strangeness.

trin-core/src/portalnet/overlay.rs
carver, line 286: I think this variable name assumes too much about the caller. Wouldn't we call the same endpoint after a `portal_historyStore` for example? So maybe something like:
```suggestion
    fn propagate_gossip(&self, content: Vec<(impl OverlayContentKey, ByteList)>) {
```

trin-core/src/portalnet/overlay.rs
carver, line 323:
```suggestion
            // Temporarily store all random selected nodes with interested content.
```

trin-core/src/portalnet/overlay.rs
carver, line 335: I take it this conversion is because an `Enr` instance is not hashable for the `HashMap`? Otherwise would be nice to avoid these steps of `.to_base64()` and `::from_str()`.

carver (approved):

trin-core/src/portalnet/overlay.rs
ogenev: I think `TODO` is confusing here. The intention for this is to be a placeholder if we want to do something with a `Reset` stream in the future. At this point, I think we don't need to do anything in the overlay, the stream is removed from the active connections in `UtpListener`.

I will remove the TODO and just leave it empty for now, no need to create an issue IMO. We may also want to remove this event or refactor the `UtpListener` events but I prefer to do this incrementally.

ogenev: Same answer as https://github.com/ethereum/trin/pull/328/files#r894254880

trin-core/src/portalnet/overlay.rs
ogenev: Good catch, no need to clone here, I wish clippy can catch those unnecessary clones.

trin-core/src/portalnet/overlay.rs
ogenev: Yeah, good points, updating the warning and returning an error makes more sense here :+1:

trin-core/src/portalnet/overlay.rs
ogenev: I finally learned the difference between temporally and temporarily :)

trin-core/src/portalnet/overlay.rs
ogenev: Yeah, it is because `discv5::enr::Enr<CombinedKey>` doesn't implement the `Hash` trait. The first workaround that came to my mind was to convert it to String.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        }
+        // Accepted content keys len should match content value len
+        if content_keys.len() != content_values.len() {
+            return Err(anyhow!(
+                "Content keys len doesn't match content values len."
+            ));
+        }
+
+        // Try to store accepted content into the database and propagate gossip the content
+        match self.protocol {
```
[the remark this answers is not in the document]
ogenev: I forgot about this type  :smile:  `TContentKey` is definitely a better solution.

trin-core/src/portalnet/overlay.rs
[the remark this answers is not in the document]
ogenev: We can't use Enr because it does not implement the `Hash` trait.

trin-core/src/portalnet/overlay.rs
```diff
@@ -175,12 +190,182 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
     /// Process overlay uTP listener event
-    pub fn process_utp_event(&self, event: UtpListenerEvent) {
-        // TODO: Handle overlay uTP events
-        debug!(
-            "Got overlay uTP event: {event:?}, protocol id: {:?}",
-            self.protocol
-        );
+    pub fn process_utp_event(&self, event: UtpListenerEvent) -> anyhow::Result<()> {
+        match event {
+            UtpListenerEvent::ClosedStream(utp_payload, _, stream_id) => {
```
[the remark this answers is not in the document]
ogenev: It is used by `PortalnetEvents` to dispatch the event to the overlay network [here](https://github.com/ethereum/trin/blob/62734135570a7ca9ee0515dc124c61281863c6f5/trin-core/src/portalnet/events.rs#L131).

The current event types are not perfect, we may need to refactor them to remove some redundant matches as we have more visibility now but I prefer to have something working first.

trin-core/src/portalnet/overlay.rs
ogenev: I think returning an `Error` and safe `unwrap` also makes sense?

trin-core/src/portalnet/types/content_key.rs
ogenev: We can :scissors: this in favor of `VariableList`.

trin-core/src/portalnet/overlay.rs
```diff
@@ -563,4 +740,25 @@ mod test {
             Err(err) => assert!(err.to_string().contains(&msg)),
         }
     }
+
+    #[test]
```
njgheorghita, line 744: *nit* This test seems like it could be cleaned up a bit by using `rstest` like we do [here](https://github.com/ethereum/trin/blob/37c1931a2f2eb9e4832d980574070e8ff1a5311e/trin-core/src/utils/bootnodes.rs#L53).

trin-core/src/portalnet/overlay.rs
```diff
@@ -192,12 +201,160 @@ where
...
+            return Err(anyhow!("Received content values with len > 64."));
+        }
+        // Accepted content keys len should match content value len
+        if content_keys.len() != content_values.len() {
+            return Err(anyhow!(
+                "Content keys len doesn't match content values len."
+            ));
+        }
+
+        // TODO: Verify overlay content data with an Oracle
```
njgheorghita, line 244: This should just require an update from the current `phantom_validator` field to be a handle to the actual `HeaderOracle`, which we can then use to validate the content.

njgheorghita (approved):
Looks great, stoked to get this functionality in!

trin-core/src/portalnet/overlay.rs
```diff
@@ -192,12 +201,160 @@ where
...
+        content_keys: Vec<RawContentKey>,
+        payload: UtpPayload,
+    ) -> anyhow::Result<()> {
+        // Process accepted uTP payload as SZZ decoded Container[payloads: List[ByteList, max_length=64]]
+        let content_values = ContentPayloadList::from_ssz_bytes(&payload)
+            .map_err(|err| anyhow!("Unable to decode uTP content payload: {err:?}"))?;
+        // Max content value length should be 64
+        if content_values.len() > 64 {
+            return Err(anyhow!("Received content values with len > 64."));
+        }
```
jacobkaufmann, line 236: I don't think we need this check here, because the type does not allow a length beyond 64.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: I prefer `Option` where `None` represents "undefined", but I also see that `Error` could make sense. We could use something like `unwrap_or_else` and return/continue on `None` or `Err`.

jacobkaufmann (approved):
Logic looks solid to me, but to the extent that it's possible, I would like to see some tests in a subsequent PR.

trin-core/src/portalnet/overlay.rs
```diff
@@ -192,12 +201,160 @@ where
...
+            for enr in random_enrs {
+                enrs_and_content
+                    .entry(enr.to_base64())
+                    .or_default()
+                    .push(content_key_value.clone().0.into());
+            }
+        }
+
+        // Create and send OFFER overlay request to the interested nodes
+        for (enr_string, interested_content) in enrs_and_content.drain() {
```
jacobkaufmann, line 323: I think we can consume the map here with `into_iter`. I don't see why we need to preserve the allocated memory with `drain`.

trin-core/src/portalnet/overlay.rs
```diff
@@ -192,12 +202,156 @@ where
...
+                // Handle all closed uTP streams, currently we process only AcceptStream here.
+                // FindContent payload is processed explicitly when we send FindContent request.
+                // We don't expect to process a payload for ContentStream and OfferStream.
+                match stream_id {
+                    UtpStreamId::AcceptStream(content_keys) => {
+                        self.process_accept_utp_payload(content_keys, utp_payload)?
+                    }
+                    UtpStreamId::ContentStream(_) => {}
+                    UtpStreamId::FindContentStream => {}
+                    UtpStreamId::OfferStream => {}
```
carver, line 217: Maybe worth a short `warn!()` on `ContentStream` and `OfferStream` then, since it's an unexpected code path.

carver (approved):
Cool, looks good! One last suggestion, for when you're back.

mrferris (approved):
👍

trin-core/src/portalnet/overlay.rs
ogenev: No need to warn here because we will receive `OfferStream` and `ContentStream` events but we don't want to process the payload because there should be no payload received (those streams are used to send uTP DATA packets). Current uTP listener events are maybe not with the optimal structure but we can refactor this at some point.

Issue #329, opened 2022-05-18
Discord invite in README is invalid or expired


ryankav:
Tried to join the discord channel today but the link in the README is either invalid or has expired.  

<img width="439" alt="Screenshot 2022-05-18 at 17 17 45" src="https://user-images.githubusercontent.com/39402882/169092120-ae69e973-380f-435c-8868-4902008c22db.png">

ogenev:
Hey, thanks for letting us know that the link expired, https://github.com/ethereum/trin/pull/330 should fix it!

PR #330, opened 2022-05-18
Set non-expire Discord link in README

ogenev:
### What was wrong?
Fixes https://github.com/ethereum/trin/issues/329

### How was it fixed?
Update Discord link to non-expire

### To-Do

- [x] Clean up commit history


njgheorghita (approved):
:boat:

ogenev:
> boat - is this something we're just going to have to continually update?

I don't think so, the current link should not expire.

PR #331, opened 2022-05-19
Initial ValidationOracle implementation

njgheorghita:
### What was wrong?

Initial datatype to help with portal network data validation. This is a fairly minimal implementation, but the general structure that I'm considering is there. Looking for feedback / suggestions / improvements on this layout before continuing to finish out the todos.

#### Validation needs to happen..
- [ ] Before anything is stored in the db (direct injection via jsonrpc)
- [x] For all content fetched on demand from a network (regardless of whether or not we store it)
- [ ] For all content fetched on demand from a network via utp
- [ ] For `Accept`ed content that's been `Offer`ed to us (via utp)
- [ ] For `BlockBodies` & `Receipts` in the Chain History network

#### Assumptions
- If we cannot validate a piece of content (for whatever reason) - it's considered invalid, discarded and not stored. 
- All supported networks will always be run simultaneously, no need to worry about a modular design for the validation at the moment.

#### Where to validate?
Validation could happen "on top" of an overlay network.
- Request a piece of content from an overlay network. then perform the validation afterwards, by requesting whatever data is needed from another network (similar to how we handle composing `eth_*` jsonrpc endpoints).

But since validation needs to happen "within" an overlay network (we might as well perform all the validation at that level).
- aka. In order to validate `Accept`ed content before storing it locally, `OverlayService` must be capable for validating data. 
- Resulting behavior is that all data requested from an overlay network has already been validated and doesn't require any further validation. 

#### Current Implementation
`ValidationOracle`
- A struct (shared between the networks) that maintains & uses jsonrpc tx channels (and infura capability) to request validation data from other overlay networks (or infura for now).
- This is passed between every network's intialization steps, where it's used to initialize that networks `Validator` and it collects a handle to that networks jsonrpc tx channel.
  - Using jsonrpc tx channels to communicate this data seemed the simplest option, but I'm open to alternative suggestions. 
    - I considered using an overlay networks `events` channel, but jsonrpc seemed sufficient & simpler
- It seems possible that sharing this object between all of the overlay networks is not a sustainable design, and at some point we might need to implement message channels to coordinate communication to/from the `ValidationOracle`
- Currently we're only validating chain history network data.

`Validator`
- A trait that must be implemented for every overlay service to use respectively to validate whatever `content_key`/`content` types that network supports.

### To-Do

- [x] Clean up commit history


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -916,17 +926,49 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            Err(_) => {
+                error!("Unable to process received content: Invalid content key.");
+                return;
+            }
+        };
+        let ck_clone = content_key.clone();
+        let content_clone = content.clone();
+        self.validator
+            .validate_content(ck_clone, content_clone)
+            .await;
```
carver, line 954: So here is where I expect an `Err` check and then an early return before storage if it was invalid.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -916,17 +926,49 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            Ok(val) => val,
+            Err(_) => {
+                error!("Unable to process received content: Invalid content key.");
+                return;
+            }
+        };
+        let ck_clone = content_key.clone();
+        let content_clone = content.clone();
+        self.validator
+            .validate_content(ck_clone, content_clone)
```
carver, line 953: Can these args be borrowed instead of cloned?

trin-core/src/types/validation.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub history_jsonrpc_tx: Option<tokio::sync::mpsc::UnboundedSender<HistoryJsonRpcRequest>>,
+    pub header_gossip_jsonrpc_tx: Option<bool>,
+    pub block_indices_jsonrpc_tx: Option<bool>,
+    pub state_jsonrpc_tx: Option<bool>,
+}
+
+impl Default for ValidationOracle {
+    fn default() -> Self {
+        Self {
+            infura_url: "https://mainnet.infura.io::443/v3/".to_string(),
```
carver, line 29: This extra colon looks wrong
```suggestion
            infura_url: "https://mainnet.infura.io:443/v3/".to_string(),
```

carver (commented):
A few funky things. Generally I think it's okay, but I flew through it pretty quickly.

With the move of the accumulator into the history network, we may not actually need the cross-network validation super soon, but I think we'll want it eventually anyway.

trin-core/src/types/validation.rs
```diff
@@ -0,0 +1,101 @@
...
+        content_key: TContentKey,
+        content: ByteList,
+    ) -> anyhow::Result<()>
+    where
+        TContentKey: 'async_trait;
+}
+
+// This is a mock Validator for use in tests where no validation is required.
+pub struct IdentityValidator {
+    pub header_oracle: HeaderOracle,
```
mrferris, line 86: Any reason this needs to have a `HeaderOracle`, instead of nothing?

trin-core/src/types/validation.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Clone)]
+pub struct HeaderOracle {
+    pub infura_url: String,
+    // We could simply store the main portal jsonrpc tx channel here, rather than each
+    // individual channel. But my sense is that this will be more useful in terms of
+    // determining which subnetworks are actually available.
+    pub history_jsonrpc_tx: Option<tokio::sync::mpsc::UnboundedSender<HistoryJsonRpcRequest>>,
+    pub header_gossip_jsonrpc_tx: Option<bool>,
+    pub block_indices_jsonrpc_tx: Option<bool>,
+    pub state_jsonrpc_tx: Option<bool>,
```
mrferris, line 22: I agree with having these individually instead of the main portal jsonrpc tx channel for the reason you said, but down the line maybe we want to wrap these `jsonrpc_tx`s up into a wrapper struct to avoid duplicating the same four `jsonrpc_tx` members in the `BlockBodyOracle` and `BlockReceiptsOracle`. Though that's only if all oracles need access to all networks. If not, I think the clearest thing would be to have each oracle have access to the minimum number of networks necessary.

trin-core/src/types/validation.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::jsonrpc::service::dispatch_infura_request;
+use crate::jsonrpc::types::{HistoryJsonRpcRequest, JsonRequest, Params};
+use crate::portalnet::types::content_key::IdentityContentKey;
+use crate::portalnet::types::messages::ByteList;
+
+// This struct is responsible for dispatching cross-overlay-network requests
+// for data to perform validation. Currently, it just proxies these requests
+// on to infura.
+#[derive(Clone)]
+pub struct HeaderOracle {
```
mrferris, line 14: nit: I'd call this `BlockHeaderOracle` to be consistent with other places where `BlockHeader` is used instead of just `Header`, like `ContentKey::BlockHeader`.

mrferris (commented):
I approve of the general pattern of a `Validator` instance per network, containing shared `Oracle` instances which are each able to validate a specific piece of network data. 👍

In your initial comment, the checklist says that validation is currently only done when requesting on-demand data (item 2), but it seems like the code is only validating content before storing in the DB (item 1). Not sure if that's a mistake or misunderstanding on my part.

trin-history/src/lib.rs
```diff
@@ -73,11 +75,16 @@ pub async fn main() -> Result<(), Box<dyn std::error::Error>> {
         events.process_discv5_requests().await;
     });
 
+    // todo: Update to a functional validation oracle.
+    // Currently running HistoryNetwork solo will not perform any validation,
+    // since supporting running individual networks is not currently a priority.
```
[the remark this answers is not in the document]
njgheorghita: I went ahead and added an infura project id requirement for running the history network solo, so now it'll perform validation normally.

trin-core/src/types/validation.rs
njgheorghita: There may be some confusion in how I explained this, but the idea was to only have a single "Oracle" - the object that has access to the various tx channels. This will be shared (via `Arc<RwLock>`) between the different "Validators" (eg. `BlockBodyValidator` and `ReceiptsValidator`) which will contain all the type-specific logic for validation. Iiuc - then the `HeaderOracle` is already behaving like the wrapper struct you're describing here?

src/lib.rs
```diff
@@ -69,29 +75,38 @@ pub async fn run_trin(
...
+        .networks
+        .iter()
+        .any(|val| val == HISTORY_NETWORK)
+    {
+        initialize_history_network(
+            &discovery,
+            utp_listener_tx,
+            portalnet_config.clone(),
+            storage_config.clone(),
+            header_oracle.clone(),
```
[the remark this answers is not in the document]
njgheorghita: Yeah good call, I had some contrived idea of how this was going to happen with a sort of path-dependent initialization process, but this is a much better solution - though had to go with `Arc<RwLock<HeaderOracle>>` to be able to get a mut ref.

trin-core/src/types/validation.rs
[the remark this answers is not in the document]
njgheorghita: Not off the top of my head. So I understand that this field (along with `header_gossip_jsonrpc_tx` and `block_indices_jsonrpc_tx`) are not necessary atm, but personally I don't mind leaving them in there to maintain some context for the object's purpose

trin-core/src/types/validation.rs
njgheorghita: in conversation during devconnect/syncs we've been throwing around this `HeaderOracle` term, which is a main reason I chose that name. I also considered `ValidationOracle`. Imo the "block" part of "BlockHeader" is somewhat superfluous since there is only one kind of header, therefore no need to distinguish between header types. Also, `ContentKey::BlockHeader` is a defined type from the specs, while this oracle object isn't strictly defined, a bit more general in purpose, and could end up performing various non-header related functions down the road. These are just some thoughts I had for why I went with `HeaderOracle`, but definitely lmk if you still feel strongly in favor of `BlockHeaderOracle`

njgheorghita:
> In your initial comment, the checklist says that validation is currently only done when requesting on-demand data (item 2), but it seems like the code is only validating content before storing in the DB (item 1). Not sure if that's a mistake or misunderstanding on my part.

@mrferris Gotcha, In my head # 1 was a bit more general ( and probably redundant now that you point this out) referring to all instances in which we'd store something in the db (findcontent / findcontent via utp / offer / direct injection). Since those aren't all complete, I left that checkmark blank for now

trin-core/src/types/validation.rs
mrferris: Hmm, I'm not seeing why there would be a `BlockBodyValidator` and `ReceiptsValidator` (instead of `BlockBodyOracle` and `ReceiptsOracle`) when currently there's `ChainHistoryValidator` and `StateValidator`. Seems like it's currently set up so that each network has one validator, and each data type has an oracle, but you're saying there will be a single oracle (in which case, `HeaderOracle` sounds like it's overly specific to headers) and a validator per data type.

njgheorghita: Ahh, my bad I got mixed up. You're right. The `ChainHistoryValidator` will do all validation for chain history network types (headers, bodies, and receipts). But, yeah the name `HeaderOracle` isn't necessarily an oracle for just the `BlockHeader` type. `HeaderOracle` refers more to the idea of an oracle that can answer questions like `is_hash_canonical`, `get_hash_at_height`, `get_latest_header` (etc). Which might be used to validate state network data for example. This is one reason why I considered the name `ValidationOracle` - which seems like a more generic name, but since we've used `HeaderOracle` in discussion I went that way. However, we don't need a `BlockBodyOracle` and `ReceiptsOracle` - we can simply use the `HeaderOracle` to validate those data types.

jacobkaufmann (commented):
I have not been able to come up with a better way to perform validation, but I can't help but feel like 3 `PhantomData` fields in `OverlayProtocol` is something indicative of an anti-pattern.

The only structural alternative I can imagine is some channel to bubble up unvalidated content items from the overlay, and some channel to send validation confirmations back down to the overlay.

Another thought I had was the addition of validation metadata to the `OverlayContentKey`, such as which other content keys are necessary to validate this content key, but this would potentially require that different implementations of `OverlayContentKey` be aware of each other, which we don't want. In general, I was thinking about ways to make the `OverlayContentKey` trait provide richer behavior. I would be interested to know if you have any ideas on that topic.

If possible, I would love to stay away from `async-trait`, but bypassing that might require some unwelcome complexity.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -935,17 +952,54 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+        let find_content = match request {
+            Request::FindContent(find_content) => find_content,
+            _ => {
+                error!("Unable to process received content: Invalid request message.");
+                return;
+            }
+        };
+        let content_key = match TContentKey::try_from(find_content.content_key) {
+            Ok(val) => val,
+            Err(_) => {
```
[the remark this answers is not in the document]
njgheorghita: Updated the one below, but I don't think it's currently possible with the message on `971` - due to [this bug](https://github.com/rust-lang/rust/issues/82252). Basically the compiler bugs out and says `Debug` isn't implemented for the error message, but `Debug` is most definitely implemented for an [anyhow Error](https://docs.rs/anyhow/latest/anyhow/struct.Error.html)

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -259,10 +268,15 @@ pub struct OverlayService<TContentKey, TMetric> {
...
+    /// Validator for overlay network content.
+    validator: TValidator,
 }
 
-impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
-    OverlayService<TContentKey, TMetric>
+impl<
+        TContentKey: OverlayContentKey + Send + Sync,
+        TMetric: Metric + Send,
+        TValidator: 'static + Validator<TContentKey> + Send,
```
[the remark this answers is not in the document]
njgheorghita: From what I can tell, `async_trait` makes this difficult if not impossible

njgheorghita:
@jacobkaufmann 

> I have not been able to come up with a better way to perform validation, but I can't help but feel like 3 PhantomData fields in OverlayProtocol is something indicative of an anti-pattern.

Yeah I can see that, but I don't have any strong instinct as to why it's problematic now or down the road. 

> The only structural alternative I can imagine is some channel to bubble up unvalidated content items from the overlay, and some channel to send validation confirmations back down to the overlay.

Iiuc, something like this was considered. But due to the requirement that we want to validate content that we've `Accept`ed, it makes sense to me to handle that validation within the overlay. Otherwise complexity is going to notch up quickly (that's a whole round trip that's somewhat redundant). And since we need validation in the overlay for those messages, we might as well perform all validation at this level.

> Another thought I had was the addition of validation metadata to the OverlayContentKey, such as which other content keys are necessary to validate this content key, but this would potentially require that different implementations of OverlayContentKey be aware of each other, which we don't want. In general, I was thinking about ways to make the OverlayContentKey trait provide richer behavior. I would be interested to know if you have any ideas on that topic.

I considered this as well. It would have been lovely to essentially merge the `TContentKey` and `TValidator` generics. The way I see it, the validation mechanism is going to need access to all (or most) of the other overlay networks. Whether it's the `OverlayContentKey` type or the `HeaderOracle`. Imo the separation of responsibilities in this case makes sense and justifies the new type. I think `OverlayContentKey` would just get too complex if we try to put it all in there.

> If possible, I would love to stay away from async-trait, but bypassing that might require some unwelcome complexity.

I'm just not sure if this is possible. Requests to networks for data used in validation **should** be async imo. But I'm curious why you want to stay away from `async-trait`

trin-core/src/portalnet/overlay_service.rs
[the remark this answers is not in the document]
jacobkaufmann: Hmm well the compiler only knows that `TContentKey` implements `OverlayContentKey`, but in the trait definition for `OverlayContentKey`, we do not specify that the `Error` associated type for `TryFrom` implements `Debug`. I think we could fix this with a trait bound on the associated type.

jacobkaufmann:
If the `PhantomData` becomes a problem later, then we can revisit. I know that it affects [the drop checker and variance](https://doc.rust-lang.org/nomicon/phantom-data.html), but that does not seem to be an issue for now.

My aversion to `async-trait` was based on my limited understanding of how it works, and how it moves some of our previously non-async functions to async. After some [subsequent reading](https://smallcultfollowing.com/babysteps/blog/2019/10/26/async-fn-in-traits-are-hard/) I think usage of `async-trait` should be okay.

trin-history/src/validation.rs
```diff
@@ -0,0 +1,202 @@
...
+    pub header_oracle: Arc<RwLock<HeaderOracle>>,
+}
+
+#[async_trait]
+impl Validator<HistoryContentKey> for ChainHistoryValidator {
+    async fn validate_content(
+        &mut self,
+        content_key: &HistoryContentKey,
+        content: &ByteList,
+    ) -> anyhow::Result<()>
```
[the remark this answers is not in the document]
njgheorghita: I'm not so sure. Imo `validate_xxx` is simply asking to validate the content or return an error. Pretty much every time I've used this naming, I've stuck with this basic pattern. I've read some posts (that I can't find now) which say returning `Result<(), Err>` is the preferred way to handle this in rust. 

I get that it seems kind of useless, but I'd also argue against implementing unnecessary logic (like returning data that we don't currently need). Now ... maybe down the road we do need this data. At which point, I'd say that's fine but we should change the name of the validation function to something like `validate_and_get_content_key()`.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -967,17 +997,46 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
             Content::Enrs(enrs) => self
                 .process_discovered_enrs(enrs.into_iter().map(|ssz_enr| ssz_enr.into()).collect()),
         }
     }
 
+    async fn process_received_content(&mut self, content: ByteList, request: FindContent) {
+        let content_key = match TContentKey::try_from(request.content_key) {
+            Ok(val) => val,
+            Err(msg) => {
+                error!("Unable to process received content: Invalid content key: {msg:?}");
```
carver, line 1010: This is a weird situation. We requested content with an invalid content ID and then find out about it during processing the response, yeah?

The error message doesn't really convey how weird that is. I think a log like this might give me more up-front understanding:
```suggestion
                error!("Unable to process received content: we sent a request with an invalid content key: {msg:?}");
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -967,17 +997,46 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
...
+            .validate_content(&content_key, &content)
+            .await
+        {
+            error!("Unable to validate received content: {err:?}");
+            return;
+        };
+        match self.storage.read().should_store(&content_key) {
+            Ok(should_store) => {
+                if should_store {
+                    if let Err(err) = self.storage.write().store(&content_key, &content.into()) {
```
carver, line 1025: Since validation might be expensive (literally, at first), maybe we should be validating after testing `should_store()`, but obviously before `store()`.

trin-core/src/types/validation.rs
```diff
@@ -0,0 +1,103 @@
...
+            header_gossip_jsonrpc_tx: None,
+            block_indices_jsonrpc_tx: None,
+            state_jsonrpc_tx: None,
+        }
+    }
+}
+
+impl HeaderOracle {
+    // Currently falls back to infura, to be updated to use canonical block indices network.
+    pub fn get_hash_at_height(&mut self, hex_number: String) -> anyhow::Result<String> {
```
carver, line 43: Passing in the block number as a hex string at this level of the API is surprising to me. I guess it's because Infura/json-rpc expects a hex string, but that seems like an implementation detail that's leaking up into the API. I would expect a uint here, then converting to a hex string inside of `get_hash_at_height()` before building the `JsonRequest`.

trin-history/src/validation.rs
```diff
@@ -0,0 +1,205 @@
...
+        match content_key {
+            HistoryContentKey::BlockHeader(key) => {
+                let rlp = Rlp::new(content);
+                let header = Header::decode_rlp(&rlp)?;
+                let number = format!("0x{:02X}", header.number);
+                let expected_hash = &self
+                    .header_oracle
+                    .write()
+                    .unwrap()
+                    .get_hash_at_height(number)?;
```
carver, line 38: ... continuing from the previous comment, I'd expect to be able to do this here:
```suggestion
                    .get_hash_at_height(header.number)?;
```

carver (approved):

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -363,7 +363,10 @@ pub fn make_ipc_request(ipc_path: &str, request: &JsonRpcRequest) -> anyhow::Res
     stream.flush().unwrap();
     let deser = serde_json::Deserializer::from_reader(stream);
     let next_obj = deser.into_iter::<Value>().next();
-    let response_obj = next_obj.ok_or(anyhow!("Empty JsonRpc response"))?;
+    let response_obj = match next_obj {
+        Some(val) => val,
+        None => return Err(anyhow!("Empty JsonRpc response")),
+    };
```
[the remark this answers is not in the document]
njgheorghita: Clippy was complaining about an `ok_or()` containing a function, but it looks like `ok_or_else(|| ...)` works as well, not sure why I went so verbose the first time around...

PR #332, opened 2022-05-20
Refactor FindNodeQuery and QueryPool with Query trait

jacobkaufmann:
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/find/iterators/findnodes.rs
+++ b/trin-core/src/portalnet/find/iterators/findnodes.rs
@@ -20,9 +20,12 @@
 
 // This basis of this file has been taken from the rust-libp2p codebase:
 // https://github.com/libp2p/rust-libp2p
-//
+
 use super::super::query_pool::QueryState;
+use super::query::{Query, QueryPeer, QueryPeerState, QueryProgress};
+
 use discv5::kbucket::{Distance, Key};
+
 use std::{
     collections::btree_map::{BTreeMap, Entry},
     time::{Duration, Instant},
@@ -32,6 +35,9 @@ use std::{
 pub struct FindNodeQuery<TNodeId> {
     /// The target key we are looking for
     target_key: Key<TNodeId>,
+
+    /// The instant when the query started.
+    started: Option<Instant>,
 
     /// The current state of progress of the query.
     progress: QueryProgress,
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/find/iterators/findnodes.rs
```diff
@@ -83,85 +89,76 @@ impl FindNodeQueryConfig {
...
         match self.closest_peers.entry(distance) {
             Entry::Vacant(..) => return,
-            Entry::Occupied(mut entry) => match entry.get().state {
+            Entry::Occupied(mut entry) => match entry.get().state() {
                 QueryPeerState::Waiting(..) => {
                     assert!(
                         self.num_waiting > 0,
                         "Query has invalid number of waiting queries"
                     );
                     self.num_waiting -= 1;
```
mrferris, line 153: A `decrement_num_waiting` method might be nice to wrap up that check that's done multiple times. Though as far as I can see it would require the same implementation on both `Query` trait implementations, which wouldn't be very nice. I'm guessing that's why you didn't.

mrferris (changes requested):
The `Query` trait definitely cleans things up. 

I was thinking that this approach might allow us to have a single `QueryPool` instance in `OverlayService` handle both `FindNodes` & `FindContent` queries, which would be elegant. But my understanding upon review is that this design will require two separate `QueryPool` instances in `OverlayService`, because the generics will be different. Do you agree? 
I think it's fine to have two instances, we just need to refactor the integration logic [here](https://github.com/ethereum/trin/blob/981b58dbf15e30f339aba2104117536eb2d6cb37/trin-core/src/portalnet/overlay_service.rs#L464) and [here](https://github.com/ethereum/trin/blob/981b58dbf15e30f339aba2104117536eb2d6cb37/trin-core/src/portalnet/overlay_service.rs#L591) to avoid duplication/bloat in `OverlayService`. Worst case we can put those in separate files if the separate types don't allow single functions to handle both. Hard to tell just by looking.

The setters/getters are a nice touch 👍

I don't quite agree with changing `next` to `poll`. I think that that `poll` makes sense for checking for updates to a data structure that might not have updates, while `next` makes sense for getting the next entry in an iterator that we know has a next entry (because of the result of `poll`). 

I think that the only thing needed to reach feature parity with the current implementation is the `request()` method on the `Query` trait. Let me know if I've misunderstood.

trin-core/src/portalnet/find/iterators/findnodes.rs
jacobkaufmann: Yeah, it could be a useful helper function, but I don't think it makes much of a difference either way. I will probably leave it for now.

trin-core/src/portalnet/find/iterators/query.rs
```diff
@@ -0,0 +1,156 @@
+use std::time::Instant;
+
+use super::super::query_pool::QueryState;
+
+use discv5::kbucket::Key;
+
+pub trait Query<TNodeId, TResponse, TResult> {
```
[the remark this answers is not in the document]
jacobkaufmann: That's a great callout, because I totally missed that. I need to go back, and account for the functionality provided by `QueryInfo`. Both the `untrusted_enrs` and `callback` fields are also important. I would prefer not to have to add any more methods to the `Query` trait, because it's already sort of my large. I will need to explore some options here.

jacobkaufmann:
Thanks for the feedback! You are correct that we will require separate `QueryPool` instances for each `Query` implementor, but I prefer this to having a single `QueryPool` that requires separate logic for each `Query` implementor. Open to pushback on that though.

With multiple query maps, it may happen that we simultaneously have a query in each map with the same `QueryId`. However, we will always be able to map a `QueryId` to the correct map, because there is a one-to-one correspondence between the type of the query and the type of the request that belongs to that query. A `FindNodes` will map to some `FindNodeQuery`, and a `FindContent` will map to some `FindContentQuery`.

We should be able to pull out the `query_event_poll` logic and make it generic over `Query`.

On `poll` versus `next`, I see a `Query` as more of a [generator](https://doc.rust-lang.org/std/ops/trait.Generator.html), sort of like a [`Future`](https://doc.rust-lang.org/std/future/index.html). Hence the name change to `poll`. I do not feel strongly about this though, so I don't have a problem changing it back if desired.

Regarding `request()`, please see related comment.

jacobkaufmann:
For now, I think we can move forward with what we have here. As discussed offline, I will keep `poll` in place of `next`.

Regarding `TTarget` and `QueryInfo` and the functionality there, we can store tuples `(QueryInfo, TQuery)` as entries in the query map.

At the time of writing, `QueryInfo` only contains logic for `FindNode` queries, but we can extend in a subsequent PR, where we define `FindContent` queries.

PR #334, opened 2022-05-25
Decode json-rpc hex with prefix

carver:
### What was wrong?

When doing manual testing to push content with `portal_historyStore`, I was getting the error "Unable to decode content_key".

I thought that was a bytes -> key issue, but it was actually a hex -> bytes issue. I updated the error message to be a bit more clear on that point.

It looks like most of the json-decoding was assuming hex without a prefix. The portal json-rpc spec has all content starting with 0x, so this was a bug to fix.

### How was it fixed?

- Added a new `hex_decode` which strips away the "0x" -- plus a little hacky error if it's not there.
- Use the `hex_decode()` method instead of `hex::decode()` (which can't handle a prefix), basically everywhere in the json-rpc type decoding.

Some bonus fixes:
- Log out the database location in `info`. There was no other indication where the data directory is, at the info level.
- Use `trin_` instead of `Trin_` for the data directory prefixes. `trin` is written lowercase roughly everywhere.

With this fix in place, the history data is being stored in response to the `portal_historyStore` call.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/jsonrpc/types.rs
+++ b/trin-core/src/jsonrpc/types.rs
@@ -13,6 +13,7 @@ use crate::{
         content_key::OverlayContentKey,
         messages::{ByteList, CustomPayload, SszEnr},
     },
+    utils::bytes::hex_decode,
 };
 
 type Responder<T, E> = mpsc::UnboundedSender<Result<T, E>>;
@@ -230,7 +231,7 @@ impl TryFrom<[&Value; 2]> for FindContentParams {
         let content_key = params[1]
             .as_str()
             .ok_or_else(|| ValidationError::new("Empty content key param"))?;
-        let content_key = match hex::decode(content_key) {
+        let content_key = match hex_decode(content_key) {
             Ok(val) => VariableList::from(val),
             Err(_) => return Err(ValidationError::new("Unable to decode content_key")),
         };
@@ -274,8 +275,10 @@ impl TryFrom<[&Value; 2]> for OfferParams {
                     .collect();
 
                 if let Ok(content_keys) = content_keys {
-                    let content_keys: Result<Vec<Vec<u8>>, _> =
-                        content_keys.iter().map(hex::decode).collect();
+                    let content_keys: Result<Vec<Vec<u8>>, _> = content_keys
+                        .iter()
+                        .map(|s| hex_decode(s.as_str()))
+                        .collect();
 
                     if let Ok(content_keys) = content_keys {
                         Ok(Self {
@@ -322,7 +325,7 @@ impl TryFrom<&Value> for RecursiveFindContentParams {
         let content_key = param
             .as_str()
             .ok_or_else(|| ValidationError::new("Empty content key param"))?;
-        let content_key = match hex::decode(content_key) {
+        let content_key = match hex_decode(content_key) {
             Ok(val) => VariableList::from(val),
             Err(_) => return Err(ValidationError::new("Unable to decode content_key")),
         };
@@ -423,7 +426,7 @@ impl<TContentKey: OverlayContentKey> TryFrom<&Value> for LocalContentParams<TCon
         let content_key = param
             .as_str()
             .ok_or_else(|| ValidationError::new("Empty content key param"))?;
-        let content_key = match hex::decode(content_key) {
+        let content_key = match hex_decode(content_key) {
             Ok(val) => match TContentKey::try_from(val) {
                 Ok(val) => val,
                 Err(_) => return Err(ValidationError::new("Unable to decode content_key")),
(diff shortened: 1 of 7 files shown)


carver:
Presumably I just have to update the test fixtures to use 0x-prefixed hex-encoding. Will look tomorrow...

trin-core/src/utils/bytes.rs
```diff
@@ -27,6 +27,18 @@ pub fn random_32byte_array(leading_bit_zeros: u8) -> [u8; 32] {
...
 
+/// Encode hex with 0x prefix
+pub fn hex_encode<T: AsRef<[u8]>>(data: T) -> String {
+    format!("0x{}", hex::encode(data))
+}
+/// Decode hex with 0x prefix
+pub fn hex_decode(data: &str) -> Result<Vec<u8>, hex::FromHexError> {
+    match &data[..2] {
+        "0x" => hex::decode(&data[2..]),
+        _ => Err(hex::FromHexError::InvalidHexCharacter { c: 'x', index: 0 }),
```
njgheorghita, line 38: I'm a bit confused by this error message. Isn't it saying basically that it found an `x` at index `0`? Which doesn't really correspond with the match statement. But, also I'm not sure if it matters all that much. Is there any reason to maintain the `FromHexError` here? Using `anyhow` here might be a better solution

njgheorghita (approved):
This is great. Pretty much because of the `hex` crates behavior, I've been using a combo of `0x` prefixed and non-`0x` prefixed hexstrings. It seems like a good idea to start enforcing exclusively `0x` prefixed strings throughout the codebase

trin-core/src/utils/bytes.rs
carver: Yeah, hence my:
> plus a little hacky error if it's not there.

> Is there any reason to maintain the `FromHexError` here? Using `anyhow` here might be a better solution

No good reason. It was really just convenience and my lack of anyhow familiarity. I'll dig in to fix it, because it will bother me too, someday...

trin-core/src/utils/bytes.rs
```diff
@@ -27,6 +29,21 @@ pub fn random_32byte_array(leading_bit_zeros: u8) -> [u8; 32] {
...
 
+/// Encode hex with 0x prefix
+pub fn hex_encode<T: AsRef<[u8]>>(data: T) -> String {
+    format!("0x{}", hex::encode(data))
+}
+/// Decode hex with 0x prefix
+pub fn hex_decode(data: &str) -> anyhow::Result<Vec<u8>> {
+    let first_two = &data[..2];
+    match first_two {
+        "0x" => hex::decode(&data[2..]).map_err(|e| e.into()),
```
carver, line 40: The other option I considered for this was:
```suggestion
        "0x" => Ok(hex::decode(&data[2..])?),
```

It's a little shorter, but a little less obvious to my eyes. (Though that may still just be my newness). So it was my second favorite option. But open to other opinions/suggestions!

trin-core/src/jsonrpc/types.rs
```diff
@@ -274,8 +275,11 @@ impl TryFrom<[&Value; 2]> for OfferParams {
                     .collect();
 
                 if let Ok(content_keys) = content_keys {
-                    let content_keys: Result<Vec<Vec<u8>>, _> =
-                        content_keys.iter().map(hex::decode).collect();
+                    let content_keys: Result<Vec<Vec<u8>>, _> = content_keys
+                        .iter()
+                        .map(|s| s.as_str())
+                        .map(hex_decode)
```
ogenev, line 281: I guess we can get rid of the second `map` with something like `.map(|s| hex_decode(s.into())`.

ogenev (approved):
Looks good :+1: .

trin-core/src/jsonrpc/types.rs
```diff
@@ -460,17 +464,21 @@ impl<TContentKey: OverlayContentKey> TryFrom<[&Value; 2]> for StoreParams<TConte
...
             .as_str()
             .ok_or_else(|| ValidationError::new("Empty content key param"))?;
-        let content_key = match hex::decode(content_key) {
+        let content_key = match hex_decode(content_key) {
             Ok(val) => match TContentKey::try_from(val) {
                 Ok(val) => val,
                 Err(_) => return Err(ValidationError::new("Unable to decode content_key")),
             },
-            Err(_) => return Err(ValidationError::new("Unable to decode content_key")),
+            Err(_) => {
```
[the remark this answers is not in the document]
carver: Hm, I had a fight with `ValidationError` library, and it nearly won. It was telling me:
```rust
474 |                     format!("Cannot convert content_key hex to bytes: {}", err),
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `&str`, found struct `std::string::String`
```
and providing a reference didn't work, because the `String` would drop out of scope at return, while the `ValidationError` was still holding it. So I come up with this ugly monster after digging into the VE docs:
```rs
                let mut ve = ValidationError::new("Cannot convert content_key hex to bytes");
                ve.add_param(std::borrow::Cow::Borrowed("hex_decode_exception"), &err.to_string());
                return Err(ve);
```

trin-core/src/jsonrpc/types.rs
carver: `into()` complained, but `as_str()` seems to work:
```rs
error[E0277]: the trait bound `&str: std::convert::From<&std::string::String>` is not satisfied
   --> trin-core/src/jsonrpc/types.rs:280:47
    |
280 |                         .map(|s| hex_decode(s.into()))
    |                                               ^^^^ the trait `std::convert::From<&std::string::String>` is not implemented for `&str`
    |
    = help: the following implementations were found:
              <&'a str as std::convert::From<&'a ascii::ascii_str::AsciiStr>>
              <&'a str as std::convert::From<webpki::name::dns_name::DnsNameRef<'a>>>
              <&'t str as std::convert::From<regex::re_unicode::Match<'t>>>
    = note: required because of the requirements on the impl of `Into<&str>` for `&std::string::String`
```

carver:
I'll merge and address any suggestions in a follow-up.

PR #335, opened 2022-05-26
Implement Receipts datatype

njgheorghita:
### What was wrong?
We need a `Receipts` object that can be validated against a header for the history network.

### How was it fixed?
- Adds a `Receipts` datatype for the [receipts](https://github.com/ethereum/portal-network-specs/blob/master/history-network.md#receipts) object in the history spec.
  - Implement `root()` to calculate the receipts root that can be validated against the header
- Imports `TypedReceipt` code (w/ some small modifications) from [openethereum](https://github.com/openethereum/openethereum/blob/main/crates/ethcore/types/src/receipt.rs)

### To-Do

- [x] Clean up commit history


trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+
+// Based on https://github.com/openethereum/openethereum/blob/main/crates/ethcore/types/src/receipt.rs
+
+/// A record of execution for a `LOG` operation.
+#[derive(Default, Debug, Clone, PartialEq, Eq, RlpEncodable, RlpDecodable)]
+pub struct LogEntry {
+    /// The address of the contract executing at the point of the `LOG` operation.
+    pub address: Address,
+    /// The topics associated with the `LOG` operation.
+    pub topics: Vec<H256>,
```
carver, line 46: Not sure the ramification exactly, but it's represented as integers, ie~ `U256`, in python.

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+            },
+        )
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TransactionOutcome {
+    /// Status and state root are unknown under EIP-98 rules.
+    Unknown,
+    /// State root is known. Pre EIP-98 and EIP-658 rules.
```
carver, line 68: As far as I can tell, EIP-98 was never deployed. We went straight to EIP-658. Would means this seems sufficient:
```suggestion
    /// State root is known, before EIP-658 is enabled.
```

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct LegacyReceipt {
+    pub cumulative_gas_used: U256,
+    pub log_bloom: Bloom,
+    pub logs: Vec<LogEntry>,
+    pub outcome: TransactionOutcome,
+}
+
+impl LegacyReceipt {
+    pub fn decode_rlp(rlp: &Rlp) -> anyhow::Result<Self> {
```
carver, line 83: This whole method seems like a duplicate of `LegacyReceipt::decode()`. Can it be removed?

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+            _ => Err(anyhow!("Invalid byte selector for tx type.")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TypedReceipt {
+    Legacy(LegacyReceipt),
+    AccessList(LegacyReceipt),
+    EIP1559Transaction(LegacyReceipt),
```
carver, line 193: It looks a little awkward to me to call this a `Transaction` when it's a receipt. Maybe just `EIP1559`, or `DynamicFee`?

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+
+    pub fn receipt_mut(&mut self) -> &mut LegacyReceipt {
+        match self {
+            Self::Legacy(receipt) => receipt,
+            Self::AccessList(receipt) => receipt,
+            Self::EIP1559Transaction(receipt) => receipt,
+        }
+    }
+
+    fn decode(tx: &[u8]) -> Result<Self, DecoderError> {
```
carver, line 231: `tx` name is confusing here. Should be `receipt`?

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+        if tx.is_empty() {
+            // at least one byte needs to be present
+            return Err(DecoderError::RlpIncorrectListLen);
+        }
+        let id = TypedTxId::try_from_wire_byte(tx[0]);
+        if id.is_err() {
+            return Err(DecoderError::Custom("Unknown transaction"));
+        }
```
carver, line 239: I think the underlying error is clearer here, no reason to wrap it. Plus then it lets you:
```suggestion
        let id = TypedTxId::try_from_wire_byte(tx[0])?;
```

carver (approved):
Yup, generally looks good to me, with some nits!

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+impl Receipts {
+    pub fn root(&self) -> anyhow::Result<H256> {
+        let memdb = Arc::new(MemoryDB::new(true));
+        let mut trie = EthTrie::new(Arc::clone(&memdb));
+
+        // Insert receipts into receipts tree
+        for (index, receipt) in self.receipt_list.iter().enumerate() {
+            let path = rlp::encode(&index).freeze().to_vec();
+            let encoded_receipt = receipt.encode();
+            match trie.insert(&path, &encoded_receipt) {
```
ogenev, line 25:
```suggestion
            trie.insert(&path, &encoded_receipt).map_err(|err| anyhow!("Error calculating receipts root: {err}"))?;
```

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+                Ok(_) => (),
+                Err(msg) => return Err(anyhow!("Error calculating receipts root: {msg:?}")),
+            }
+        }
+
+        match trie.root_hash() {
```
ogenev, line 31:
```suggestion
        trie.root_hash().map_err(|err| anyhow!("Error calculating receipts root: {err}"))?
```

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+use std::sync::Arc;
+
+use anyhow::anyhow;
+use bytes::Bytes;
+use eth_trie::{EthTrie, MemoryDB, Trie};
+use ethereum_types::{Address, Bloom, BloomInput, H256, U256};
+use rlp::{self, DecoderError, Rlp, RlpStream};
+use rlp_derive::{RlpDecodable, RlpEncodable};
+
+/// Represents the `Receipts` datatype used by the chain history wire protocol
```
ogenev, line 11: It is nice to have at least `Debug` and `Clone` implemented for this struct.

trin-core/src/types/receipts.rs
ogenev: We can also `map_err` here before `?`.

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+        if tx.is_empty() {
+            // at least one byte needs to be present
+            return Err(DecoderError::RlpIncorrectListLen);
+        }
+        let id = TypedTxId::try_from_wire_byte(tx[0]);
+        if id.is_err() {
+            return Err(DecoderError::Custom("Unknown transaction"));
+        }
+        //other transaction types
+        match id.unwrap() {
```
ogenev, line 241: We can remove this unwrap if we use bubble the error with `?` above.

ogenev (approved):
Looks good to me, left some nitty comments :+1: .

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+impl TypedTxId {
+    pub fn try_from_wire_byte(n: u8) -> anyhow::Result<Self> {
+        match n {
+            x if x == TypedTxId::EIP1559Transaction as u8 => Ok(Self::EIP1559Transaction),
+            x if x == TypedTxId::AccessList as u8 => Ok(Self::AccessList),
+            x if (x & 0x80) != 0x00 => Ok(Self::Legacy),
+            _ => Err(anyhow!("Invalid byte selector for tx type.")),
+        }
+    }
+}
```
jacobkaufmann, line 187: We can remove the `#[repr(u8)]` and custom function in favor of `TryFrom<u8>`.

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+            cumulative_gas_used,
+            log_bloom: logs.iter().fold(Bloom::default(), |mut b, l| {
+                b.accrue_bloom(&l.bloom());
+                b
+            }),
+            logs,
+            outcome,
+        }
+    }
+    pub fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
```
jacobkaufmann, line 125: Is there a reason that we aren't implementing the [`Decodable`](https://docs.rs/rlp/0.5.1/rlp/trait.Decodable.html) and [`Encodable`](https://docs.rs/rlp/0.5.1/rlp/trait.Encodable.html) traits here?

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+        }
+        s.append(&self.cumulative_gas_used);
+        s.append(&self.log_bloom);
+        s.append_list(&self.logs);
+    }
+}
+
+#[derive(Eq, Hash, Debug, Copy, Clone, PartialEq)]
+#[repr(u8)]
+pub enum TypedTxId {
```
jacobkaufmann, line 172: Can we just call this `TxId` or `TxType`? Also okay with verbose spelling (e.g. `TransactionId`).

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+            x if x == TypedTxId::EIP1559Transaction as u8 => Ok(Self::EIP1559Transaction),
+            x if x == TypedTxId::AccessList as u8 => Ok(Self::AccessList),
+            x if (x & 0x80) != 0x00 => Ok(Self::Legacy),
+            _ => Err(anyhow!("Invalid byte selector for tx type.")),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TypedReceipt {
```
jacobkaufmann, line 190: Can we just call this `Receipt`?

trin-core/src/types/receipts.rs
```diff
@@ -0,0 +1,523 @@
...
+            }
+            TypedTxId::AccessList => {
+                let rlp = Rlp::new(&tx[1..]);
+                Ok(Self::AccessList(LegacyReceipt::decode(&rlp)?))
+            }
+            TypedTxId::Legacy => Ok(Self::Legacy(LegacyReceipt::decode(&Rlp::new(tx))?)),
+        }
+    }
+
+    pub fn decode_rlp(rlp: &Rlp) -> Result<Self, DecoderError> {
```
jacobkaufmann, line 254: Same comment as above for `LegacyReceipt`. Why not a separate `impl` block for the `rlp` traits?

jacobkaufmann (commented):
Looks good, but there seems to be some cleanup in order before we merge. Most of my comments here are minor, but there are comments from other reviewers that we should address (e.g. duplicated code).

I'm not sure what the best customs are for OSS, but should we copy the license from `openethereum`? I ported the basis of `header.rs` from there, but I was not sure at the time, so I did not include the license there.

carver:
> I'm not sure what the best customs are for OSS, but should we copy the license from `openethereum`?

It depends on what their license is (and also what our license is). I didn't realize OE was GPL, until you mentioned it. It appears we don't even have a specific license added, which leaves us in an ambiguous place.

That leaves us with two main options:
1. Make `trin` a `GPL` license. That would diverge from the Snake Charmers history of MIT License, but could be worth a discussion.
2. License our code as something non-GPL (like MIT), and make sure that there is no code copied from OE. The API definitions (like the list of fields for a receipt or transaction) are copied in from the specification and so couldn't really be copyrighted by my understanding, so I don't expect that to be a problem. But we should be thoughtful about making sure we aren't copying any of the logic for processing/encoding/etc.

trin-core/src/types/receipts.rs
carver: As long as the docstring mentions that this is the typed version, then 👍🏻

trin-core/src/types/receipts.rs
njgheorghita: Hmm, I don't have a strong argument for either type here, but since both openethereum and the [ethers-rs library](https://docs.rs/ethers/latest/ethers/types/struct.Log.html) use `H256` here, I'm inclined to follow their lead.

njgheorghita:
Thanks for the feedback everyone. I've implemented the suggestions. However, I do consider the license question to be open still. I'll leave this pr here for now and we can discuss the license question at sync next tuesday.

trin-core/src/types/receipts.rs
njgheorghita: I've implemented them for `LegacyReceipt`, but turns out it's not so easy for `Receipt`. The problem is with encoding. The encoding api...
```rust
fn rlp_append(&self, s: &mut RlpStream);
```
does not play nicely with prefixing the typed transaction id to the rlp encoded transaction. I've tried using some of the available methods on `RlpStream` to make it work. Some methods like `append_raw()` and `append_internal()` seemed promising, but I had no luck. If you add an item to `RlpStream`, it always treats it as an encoded item, and won't simply allow a byte prefix along with an rlp-encoded hex string. Obvs, it's possible I'm missing something... But for now it doesn't seem worth spending more time getting those traits supported (even our `Header` implementation doesn't use `Decodable`). Implementing `Decodable` and not `Encodable` for `Receipt` seems a bit wonky to me, mixing up apis. So for `Receipt` I'd prefer to just leave the `Receipt::encode` and `Receipt::decode` implementations.

trin-core/src/types/receipts.rs
njgheorghita: Cool, I've implemented `TryFrom<u8>` instead of `try_from_wire_byte()` - but also leaving `#[repr(u8)]` in for now. I messed around with implementing `Into<u8>` ... but it makes the api a bit messy in the place where we do explicit conversion from `TransactionId` to `u8`...
eg.
`TransactionId::AccessList as u8` > `Into::<u8>::into(TransactionId::AccessList)`

PR #337, opened 2022-05-30
Dispatch uTP listener events to overlay protocol

ogenev:

### What was wrong?
In a previous [PR](https://github.com/ethereum/trin/pull/325), we implemented an uTP listener architecture that allows us to emit global uTP listener events when a uTP stream is closed.

To be able to process those events in the overlay layer, we need a way to handle them.

### How was it fixed?

The current design decision is to handle the uTP listener events in the global portal event handler and dispatch them to the overlay network handlers.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/events.rs
+++ b/trin-core/src/portalnet/events.rs
@@ -63,100 +63,100 @@ impl PortalnetEvents {
         loop {
             tokio::select! {
                 Some(event) = self.protocol_receiver.recv() => {
-                let request = match event {
-                    Discv5Event::TalkRequest(r) => r,
-                    Discv5Event::NodeInserted { node_id, replaced } => {
-                        match replaced {
-                            Some(old_node_id) => {
-                                debug!(
-                                    "Received NodeInserted(node_id={}, replaced={})",
-                                    node_id, old_node_id,
-                                );
+                    let talk_req = match event {
+                        Discv5Event::TalkRequest(r) => r,
+                        Discv5Event::NodeInserted { node_id, replaced } => {
+                            match replaced {
+                                Some(old_node_id) => {
+                                    debug!(
+                                        "Received NodeInserted(node_id={}, replaced={})",
+                                        node_id, old_node_id,
+                                    );
+                                }
+                                None => {
+                                    debug!("Received NodeInserted(node_id={})", node_id);
+                                }
                             }
-                            None => {
-                                debug!("Received NodeInserted(node_id={})", node_id);
-                            }
+                            continue;
                         }
-                        continue;
-                    }
-                    event => {
-                        debug!("Got discv5 event {:?}", event);
-                        continue;
-                    }
-                };
+                        event => {
+                            debug!("Got discv5 event {:?}", event);
+                            continue;
+                        }
+                    };
+                    self.dispatch_discv5_talk_req(talk_req);
+                },
+                Some(event) = self.utp_listener_receiver.recv() => self.handle_utp_listener_event(event)
+            }
+        }
+    }
 
-                let protocol_id = ProtocolId::from_str(&hex::encode_upper(request.protocol()));
+    /// Dispatch Discv5 TalkRequest event to overlay networks or uTP listener
+    fn dispatch_discv5_talk_req(&self, request: TalkRequest) {
+        let protocol_id = ProtocolId::from_str(&hex::encode_upper(request.protocol()));
 
-                match protocol_id {
-                    Ok(protocol) => match protocol {
-                        ProtocolId::History => {
-                            match &self.history_overlay_sender {
-                                Some(tx) => tx.send(request).unwrap(),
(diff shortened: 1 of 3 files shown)


trin-history/src/events.rs
```diff
@@ -3,35 +3,43 @@ use discv5::TalkRequest;
...
-                .await
-            {
-                Ok(response) => {
-                    debug!("Sending reply: {:?}", response);
-                    Message::from(response).into()
+    pub async fn start(mut self) {
+        loop {
+            tokio::select! {
+                Some(talk_request) = self.event_rx.recv() => {
+                let reply = match self
```
njgheorghita, line 19: *nit* shouldn't this block be indented once more

trin-state/src/events.rs
```diff
@@ -3,36 +3,44 @@ use discv5::TalkRequest;
...
-                Err(error) => {
-                    error!("Failed to process portal state event: {}", error);
-                    error.to_string().into_bytes()
-                }
-            };
+    pub async fn start(mut self) {
+        loop {
+            tokio::select! {
+                    Some(talk_request) = self.event_rx.recv() => {
+                let reply = match self
```
njgheorghita, line 19: Indents here also seem off

trin-core/src/portalnet/events.rs
```diff
@@ -33,68 +49,113 @@ impl PortalnetEvents {
...
+                    },
+                    Err(_) => warn!("Unable to decode protocol id"),
+                }
+            },
+                Some(event) = self.utp_listener_receiver.recv() => {
+                    match event {
+                        UtpListenerEvent::ClosedStream(_, ref protocol_id, _) => {
+                            match protocol_id {
+                                ProtocolId::History =>{
+                                match &self.history_utp_sender {
```
njgheorghita, line 126: moar indent here and below - also `->{` on line 125 and 142

njgheorghita (approved):
Mostly formatting issues. (did `cargo fmt` not pick those up?). And a general thought about improving how we pass around overlay network specific channels - I'd be curious to hear if anyone has suggestions on how we can improve that

carver:
> did `cargo fmt` not pick those up?

Apparently not, or CI should have complained.

trin-core/src/portalnet/events.rs
```diff
@@ -33,68 +49,113 @@ impl PortalnetEvents {
...
-                                "Received NodeInserted(node_id={}, replaced={})",
-                                node_id, old_node_id,
-                            );
-                        }
-                        None => {
-                            debug!("Received NodeInserted(node_id={})", node_id);
+    /// Main loop to dispatch `Discv5` and `UtpListener` events to overlay networks
+    pub async fn start(mut self) {
+        loop {
+            tokio::select! {
```
carver, line 64: This select is huge. I expect it to be much easier to read if it calls some functions here at each of the branches. Even some of the branches seem large enough to break down into a couple functions.

I don't feel confident about having reviewed this section until that happens.

trin-core/src/portalnet/overlay.rs
```diff
@@ -174,6 +174,15 @@ impl<TContentKey: OverlayContentKey + Send, TMetric: Metric + Send>
         self.send_overlay_request(request, direction).await
     }
 
+    /// Process overlay uTP listener event
+    pub fn process_utp_event(&self, event: UtpListenerEvent) {
+        // TODO: Handle overlay uTP events
```
carver, line 179: Normally I would say to move this into an issue. But if you already have the PR up in draft form, I am fine leaving an in-code todo comment. 👍🏻

trin-history/src/events.rs
carver: Seems like `cargo fmt` doesn't know how to handle `select!` properly? The other select was funky too.

One way to avoid the issue is by doing:
```rs
Some(talk_request) = self.event_rx.recv() => self.handle_history_talk_request(talk_request),

**Review by carver — APPROVED** [review]
I'll probably want to take another look at the refactored `select!`, but I'd rather not block your forward progress, so I'll give a 👍🏻 with an understanding of maybe suggesting some rework on it in a followup PR.

File: src/lib.rs [hunk 6]
```diff
@@ -72,26 +72,31 @@ pub async fn run_trin(
...
-            .await
-        } else {
-            (None, None, None, None)
-        };
+    let (
+        history_handler,
+        history_network_task,
+        history_event_tx,
+        history_utp_tx,
+        history_jsonrpc_tx,
```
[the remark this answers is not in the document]
ogenev: Yeah, maybe we should open an issue and try to refactor this at some point...

trin-core/src/portalnet/events.rs
ogenev: Refactored this a bit, it should be more readable now.

PR #338, opened 2022-06-06
Include QueryId in QueryPoolState enum variants

mrferris:
### What was wrong?

With the `Query` trait setup, there had become no way to get a query's ID when calling `poll` on the `QueryPool` and receiving a `QueryPoolState`. Having the ID is necessary in `OverlayService` for keeping track of which request corresponds to a given query, as well as for logging query completion/timeout.

### How was it fixed?

A field for the query's `QueryId` was added to each `QueryPoolState` enum variant.

An alternative would have been to add a `fn id() -> QueryId` method to the `Query` trait (like previously existed on the `FindNodeQuery` type in #265), but this would also require that the trait have a `fn set_id(query_id: QueryId)` method because the query is now already created by the time the `QueryId` generated in the `QueryPool`'s `add_query` method.

With the taken approach, only the `QueryPool` knows about a given query's `QueryId`. Having both the `QueryPool` and the `Query` keep track of the `QueryId` doesn't seem as clean, likewise for having `QueryPool` and `QueryInfo` both keep track if we were to go that route. Open to pushback on that. 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/find/query_pool.rs
+++ b/trin-core/src/portalnet/find/query_pool.rs
@@ -53,11 +53,11 @@ pub enum QueryPoolState<'a, TNodeId, TQuery> {
     Idle,
     /// At least one query is waiting for results. `Some(request)` indicates
     /// that a new request is now being waited on.
-    Waiting(Option<(&'a mut QueryInfo, &'a mut TQuery, TNodeId)>),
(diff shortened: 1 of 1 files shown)


jacobkaufmann (approved):
I think this is the least invasive change that we can make here. There will definitely be opportunities to make the query and query pool structure for elegant in the future, but for now we can move forward with this.

PR #339, opened 2022-06-08
Add BlockBody type

njgheorghita:
### What was wrong?
- Add `BlockBody` type, with `uncles` and `transactions` hash calculation.
- Update `Header` to use `rlp::Decodable` trait.

### How was it fixed?
- Implemented types, external api for `BlockBody` is left undone, to be implemented while validation of these types are integrated into the overlay service.

### To-Do

- [x] Clean up commit history


trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn root(&self) -> anyhow::Result<H256> {
+        // generate rlp encoded list of uncles
+        let mut stream = RlpStream::new();
+        stream.append_list(&self.uncle_list);
+        let uncles_rlp = stream.out().freeze();
+
+        // hash rlp uncles
+        let mut keccak = Keccak256::new();
+        keccak.update(uncles_rlp);
+        let hash = keccak.finalize().to_vec();
```
carver, line 57: When there's just one chunk to hash, I think [`digest()`](https://docs.rs/sha3/0.9.1/sha3/trait.Digest.html#tymethod.digest) ought to work:
```suggestion
        let hash = Keccak256::digest(uncles_rlp);
```

trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+
+impl Transaction {
+    pub fn decode(transaction: &[u8]) -> Result<Self, DecoderError> {
+        if transaction.is_empty() {
+            // at least one byte needs to be present
+            return Err(DecoderError::RlpIncorrectListLen);
+        }
+        let id = TransactionId::try_from(transaction[0]);
+        if id.is_err() {
+            return Err(DecoderError::Custom("Unknown transaction type"));
```
carver, line 78: Maybe an `if let` so we can format the unknown byte into this error message.

trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+#[derive(Default, Debug, PartialEq, Eq, Clone)]
+pub struct AccessList {
+    pub list: Vec<AccessListItem>,
+}
+
+impl Decodable for AccessList {
+    fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
+        let list = rlp
+            .iter()
+            .map(|v| rlp::decode(v.as_raw()).unwrap())
```
carver, line 159: This seems confusing: variable `rlp` and module `rlp` used in the same method. Maybe `rlp_obj` for the former?

trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+        Ok(Self { list })
+    }
+}
+
+impl Encodable for AccessList {
+    fn rlp_append(&self, stream: &mut RlpStream) {
+        stream.begin_list(self.list.len());
+        for access in self.list.iter() {
+            stream.append(access);
+        }
```
carver, line 170: This is the same, right?
```suggestion
        stream.append_list(&self.list);
```
or does that not work with a list of encodable items?

trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+}
+
+impl Encodable for AccessListItem {
+    fn rlp_append(&self, stream: &mut RlpStream) {
+        stream.begin_list(2);
+        stream.append(&self.addr);
+        stream.begin_list(self.keys.keys.len());
+        for location in &self.keys.keys {
+            stream.append(location);
+        }
```
carver, line 198: `keys.keys` is surprising here. Maybe the `AccessListItem` variable could be `storage`. If `append_list` works, then this could be:
```suggestion
        stream.append_list(self.storage.keys);
```

trin-core/src/types/block_body.rs
```diff
@@ -0,0 +1,307 @@
...
+            Transaction::decode(&hex::decode(TX10.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX11.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX12.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX13.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX14.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX15.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX16.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX17.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX18.to_string()).unwrap()).unwrap(),
+            Transaction::decode(&hex::decode(TX19.to_string()).unwrap()).unwrap(),
```
carver, line 266: Not a big deal, but it was surprising that this list went to 19, and the earlier list went to 18. Makes me wonder if something is wrong...

carver (approved):
LGTM

trin-core/src/types/block_body.rs
njgheorghita: Ahh, I just forgot to include tx 19 on the first test - but I've added it now- which simply tests that each tx individually decodes properly (maybe somewhat redundant to have that many individual txs tested?). The latter test is what's testing proper calculation of the trie root - if I remove a tx from that list the test will fail.

trin-core/src/types/block_body.rs
njgheorghita: `DecoderError::Custom()` only accepts a `str` and not a `String` - making it tricky to include the error message in there since the variable doesn't live long enough

trin-core/src/types/block_body.rs
carver: Yeah, I think it would be fine to have fewer tests in the first test, it just made me cringe to have all but one 😆 .

trin-core/src/types/block_body.rs
carver: Oh I've run into that with a couple other APIs too, there must be some better solution that we're missing...

carver:
Reminder not to check off "Clean up commit history" until the commits are squashed. 😄 It's a bit of a drag to run into "Pr feedback" commits when running `git blame`

PR #343, opened 2022-06-15
Refactor Query trait with associated types

jacobkaufmann:
### What was wrong?

The `Query` trait uses generic type parameters for the response and result types, but for a given generic node ID type, there will only ever be one implementation of the `Query` trait for each type of query. Thus, we can use associated types for the response and result types.

This change also cleans up the type signature of `Query` implementers and `QueryPool`.

### How was it fixed?

Move the response and result generic type parameters to associated types of the `Query` trait.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/find/iterators/findcontent.rs
+++ b/trin-core/src/portalnet/find/iterators/findcontent.rs
@@ -74,11 +74,13 @@ pub struct FindContentQuery<TNodeId> {
     config: QueryConfig,
 }
 
-impl<TNodeId> Query<TNodeId, FindContentQueryResponse<TNodeId>, FindContentQueryResult<TNodeId>>
-    for FindContentQuery<TNodeId>
+impl<TNodeId> Query<TNodeId> for FindContentQuery<TNodeId>
 where
     TNodeId: Into<Key<TNodeId>> + Eq + Clone,
 {
+    type Response = FindContentQueryResponse<TNodeId>;
+    type Result = FindContentQueryResult<TNodeId>;
+
     fn target(&self) -> Key<TNodeId> {
         self.target_key.clone()
     }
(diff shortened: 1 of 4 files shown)


mrferris (approved):
👍 Associated types look like the way to go in this instance because we never implement the `Query` trait more than once for a given type. 

Also definitely cleaner. 

My only question though is why we don't do the same for `TNodeID`? We don't ever, for example, implement `Query<NodeId>` for `FindNodeQuery` and then also implement `Query<SomethingElse> for FindNodeQuery`. Which, to my understanding, is the time to use associated types. 

And then going even farther (maybe too far for the scope of this PR), why have `NodeId` be a generic at all, when we know each `Query` implementation will be built on the same one? The actual type we use to represent a Node ID could change in the future, but it should change the same way in every `Query` implementation. Seems like the use of a generic type (or an associated type) for `NodeId` is just using generics for the sake of using generics.

jacobkaufmann:
Those are all valid considerations for the generic node ID parameter. Nothing is done in this PR to address those considerations because it's not clear to me why the same holds for queries in [`sigp/discv5`](https://github.com/sigp/discv5/blob/dbf6879dfc103d9ea8a32dc0ddb44a68254c40f3/src/query_pool.rs#L194), and I assume that there was some reason for that.

I expect the query structure to evolve, even in the near term. For the find content query, we represent the target content ID with the generic node ID parameter, because we can represent both with `[u8; 32]`. This is not ideal though. For both find node and find content queries, we rely on `Key` to [compute distances](https://github.com/ethereum/trin/blob/fe1ee1bfbb84aec69225aee27948022621bfc538/trin-core/src/portalnet/find/iterators/findnodes.rs#L104), but the [distance function](https://github.com/sigp/discv5/blob/dbf6879dfc103d9ea8a32dc0ddb44a68254c40f3/src/kbucket/key.rs#L81) is the XOR distance function, which does not extend to the Portal state subnetwork.

I expect the existing query structure to change further, and since the ultimate structure is not clear, I would prefer to impose as little change as possible at this time. If in the future we decide to remove the generic node ID parameter, then that should not be a difficult change.

carver (approved):
Yeah, seems like a straightforward improvement!

mrferris:
@jacobkaufmann good points 

> I expect the existing query structure to change further, and since the ultimate structure is not clear, I would prefer to impose as little change as possible at this time. If in the future we decide to remove the generic node ID parameter, then that should not be a difficult change.

Agreed

PR #344, opened 2022-06-16
Add RPC endpoints for retrieving routing table info for state and history networks

mrferris:
### What was wrong?

Trin has no way of retrieving a list of the nodes in the routing tables of each overlay network.

### How was it fixed?

- Added `portal_historyRoutingTableInfo` & `portal_stateRoutingTableInfo` endpoints. 
- Add corresponding tests to `ethportal-peertest`.

The json output format (under the "result" key) of these endpoints is: 

```
"buckets": {
      "246": [
        {
          "enr": "enr:-IS4QHq0W0mPXc5L5FkWnSB1vygIhnhSUh_xD2ddIZYCD3ycdBeHfU4S7wKLOB20zi93GEjfAa4qHHcE_i6eJaIwJPIBgmlkgnY0gmlwhEFsKgOJc2VjcDI1NmsxoQMlZcyp8md8ISz7m6TQlyYqSZPvZgRmtkV3MYgKLexwzIN1ZHCCI6Y",
          "node_id": "0x5e5feed5e2835682437c7bcae4e56ef55079bb80ace5d4482e2a2b684c86dcfb",
          "radius": "18446744073709551615",
          "status": "Connected"
        }
      ],
      "249": [
        {
          "enr": "enr:-IS4QI3GqhE7c_ufMjFYT7MXz98MqmzNlenWBKqvtFKvNqFHYVwMdM6jhtq8y4NzVYWUaJETm6jeK_V1nmmP0gJsXRgBgmlkgnY0gmlwhEFsKgOJc2VjcDI1NmsxoQNR97HGmoGUsgxt497YSsVl6RqIkrJHobZxBvBX14gEdoN1ZHCCI5I",
          "node_id": "0x5c43a5d46e81bb8cd329230ebc5d9db032c66c9ae97110c9b49e28de94b51f66",
          "radius": "18446744073709551615",
          "status": "Connected"
        }
      ],
      "252": [
        {
          "enr": "enr:-IS4QOA4voX3J7-R_x8pjlaxBTpT1S_CL7ZaNjetjZ-0nnr2VaP0wEZsT2KvjA5UWc8vi9I0XvNSd1bjU0GXUjlt7J0BgmlkgnY0gmlwhEFsKgOJc2VjcDI1NmsxoQI7aL5dFuHhwbxWD-C1yWH7UPlae5wuV_3WbPylCBwPboN1ZHCCI44",
          "node_id": "0x4f42329a2f4aebf4e488b4daa639f5bd03230865e4f5b4a232234618025b0ba8",
          "radius": "18446744073709551615",
          "status": "Connected"
        },
        {
          "enr": "enr:-IS4QNd_PFZkR7zizTltwbLbnYmDhT3TgH3xXH9ai_abUofLBkWr0_8kEdiapiR67Aurecu-E0C6-VPW4HFrHb-eENoBgmlkgnY0gmlwhE7JWtOJc2VjcDI1NmsxoQK8aBdTCqEfDB8D1xucsLrSzg_o7mOL5f6CR7QUNWXLfoN1ZHCCIzE",
          "node_id": "0x550b28434b38cf04cbd745e461ca62a81ec018e16c991d0d6a4a1aa6bb2708e0",
          "radius": "18446744073709551615",
          "status": "Disconnected"
        }
      ],
    },
    "numBuckets": 3,
    "numConnected": 3,
    "numNodes": 4
  }
}
```
I weighed a few different formats, and I think that this format strikes the best balance between human-readable and client-parseable. The previous formats were:
- flat structure where the top level key is `nodes` and each node object has a `bucket_index` key
- keyless structure where each node is represented by a 5-element list. It could be argued that this would save bandwidth by removing all of the repeated keys and just using known indices. 

Open to reverting to either if anyone has strong opinions.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/jsonrpc/utils.rs
+++ b/trin-core/src/jsonrpc/utils.rs
@@ -1,17 +1,16 @@
-use serde_json::{json, Value};
-use discv5::{
-    kbucket::{
-        ConnectionState, NodeStatus
-    },
-    enr::NodeId,
-};
 use crate::portalnet::Enr;
+use discv5::{
+    enr::NodeId,
+    kbucket::{ConnectionState, NodeStatus},
+};
+use serde_json::{json, Value};
 
-pub fn bucket_entries_to_json(bucket_entries: Vec<(usize, Vec<(NodeId, Enr, NodeStatus)>)>) -> Value {
-    let mut node_count: u16 = 0; 
+pub fn bucket_entries_to_json(
+    bucket_entries: Vec<(usize, Vec<(NodeId, Enr, NodeStatus)>)>,
+) -> Value {
+    let mut node_count: u16 = 0;
     let mut connected_count: u16 = 0;
-    let buckets_indexed: Vec<(String, Vec<(String, String, String)>)> = 
-        bucket_entries
+    let buckets_indexed: Vec<(String, Vec<(String, String, String)>)> = bucket_entries
         .into_iter()
         .map(|(index, bucket)| {
             (
@@ -20,7 +19,9 @@ pub fn bucket_entries_to_json(bucket_entries: Vec<(usize, Vec<(NodeId, Enr, Node
                     .iter()
                     .map(|(node_id, enr, node_status)| {
                         node_count += 1;
-                        if node_status.state == ConnectionState::Connected { connected_count += 1 }
+                        if node_status.state == ConnectionState::Connected {
+                            connected_count += 1
+                        }
                         (
                             format!("0x{}", hex::encode(node_id.raw())),
                             enr.to_base64(),
@@ -40,5 +41,4 @@ pub fn bucket_entries_to_json(bucket_entries: Vec<(usize, Vec<(NodeId, Enr, Node
             "numConnected": connected_count
         }
     )
-
 }
--- a/trin-core/src/jsonrpc/utils.rs
+++ b/trin-core/src/jsonrpc/utils.rs
@@ -1,0 +1,44 @@
+use serde_json::{json, Value};
+use discv5::{
+    kbucket::{
+        ConnectionState, NodeStatus
+    },
+    enr::NodeId,
+};
+use crate::portalnet::Enr;
+
+pub fn bucket_entries_to_json(bucket_entries: Vec<(usize, Vec<(NodeId, Enr, NodeStatus)>)>) -> Value {
+    let mut node_count: u16 = 0; 
+    let mut connected_count: u16 = 0;
+    let buckets_indexed: Vec<(String, Vec<(String, String, String)>)> = 
+        bucket_entries
+        .into_iter()
+        .map(|(index, bucket)| {
+            (
+                format!("{}", index),
+                bucket
+                    .iter()
+                    .map(|(node_id, enr, node_status)| {
+                        node_count += 1;
+                        if node_status.state == ConnectionState::Connected { connected_count += 1 }
+                        (
+                            format!("0x{}", hex::encode(node_id.raw())),
+                            enr.to_base64(),
+                            format!("{:?}", node_status.state),
+                        )
+                    })
+                    .collect(),
+            )
+        })
+        .collect();
+
+    json!(
+        {
+            "buckets": buckets_indexed,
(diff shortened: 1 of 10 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -233,6 +233,34 @@ where
             .collect()
     }
 
+    /// Returns a Vec of tuples, where each tuple represents a node in the kbuckets table.
+    ///     the usize represents the index (1-256) of the bucket containing
+    ///     the node with NodeId, Enr, and NodeStatus
+    pub fn bucket_entries(&self) -> HashMap<usize, Vec<(NodeId, Enr, NodeStatus, U256)>> {
+        self.kbuckets
+            .write()
```
carver, line 241: Will this work with a read lock?

trin-core/src/jsonrpc/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+                        map
+                    })
+                    .collect(),
+            )
+        })
+        .collect();
+
+    json!(
+        {
+            "buckets": buckets_indexed,
```
carver, line 46: I see that the example has buckets sorted by index, and I agree that it's something we want. I didn't see any explicit sorting here. On a quick search, I wasn't able to confirm that serde's `json!` uses a serialization that orders by keys, either.

I found some references to `#[serde(serialize_with="ordered_map")]` online, if that helps.

carver (approved):

trin-core/src/portalnet/overlay.rs
mrferris: sure will, good catch

trin-core/src/jsonrpc/utils.rs
mrferris: Looks like that approach was deprecated, and now you just choose map ordering behavior by enabling a feature in `Cargo.toml`.

trin-core/src/jsonrpc/utils.rs
```diff
@@ -0,0 +1,52 @@
...
+use std::collections::HashMap;
+
+type NodeMap = HashMap<String, String>;
+type NodeTuple = (NodeId, Enr, NodeStatus, U256);
+
+/// Converts the output of the Overlay's bucket_entries method to a JSON Value
+pub fn bucket_entries_to_json(bucket_entries: HashMap<usize, Vec<NodeTuple>>) -> Value {
+    let mut node_count: u16 = 0;
+    let mut connected_count: u16 = 0;
+    let buckets_indexed: HashMap<usize, Vec<NodeMap>> = bucket_entries
```
carver, line 17: When the buckets are collected into a `HashMap` here, I think we also risk losing the ordering. Maybe it's sufficient to do a `Vec<(usize, Vec<NodeMap>)>`?

If that doesn't work, it appears that a [BTreeMap](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html) might do the trick.

njgheorghita (approved):
:ship:

trin-core/src/jsonrpc/utils.rs
mrferris: iirc, I was originally doing that `Vec` based suggestion and `clippy` yelled at me for trying to use "too complex of a type".

carver:
> ### To-Do
>     - [ ] Clean up commit history

Just a reminder to check the to-dos before merging, to avoid having commits like "Fix PR Feedback" in master :)

Other than that, looks all good 👍🏻

mrferris:
I figured squashing it all into one commit was sufficient cleanup. Though that still leaves the commit messages in the extended single commit's message, so I'll clean that message up too going forward.

carver:
> Yea, I figured squashing it all into one commit was sufficient cleanup.

Oh, maybe I'm misreading the PR somehow, I thought it was a new merge commit with these 9 commits in the history

carver:
Interesting! I can't really tell by looking at this PR history that it was a squash commit. I could only tell by navigating away to master to see the latest commit and double-checking it's parent. That feels like a github UI bug to me...

PR #345, opened 2022-06-21
Process ACCEPT response in overlay service

ogenev:
### What was wrong?

When we send `OFFER` request via the `portal_*Offer` json-rpc endpoint, we are waiting for the `ACCEPT` response and process it here:
https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/overlay.rs#L556

However, this is not the case when a node receives an incoming `ACCEPT` request which is not a result of a `portal_*Offer` json-rpc request (for example when gossiping the content to a third node). 

All incoming requests are handled in `overlay_service` except `ACCEPT`: 

https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/overlay_service.rs#L1071

This breaks the propagated gossip flow when a node receives `ACCEPT` response from a third node.

### How was it fixed?
1. Move the processing of the `ACCEPT` response from `overlay` to `overlay_service`.
2. FIxes also a data radius to distance(content_id, node_id) comparison bug.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -122,11 +122,11 @@ pub struct OverlayProtocol<TContentKey, TMetric, TValidator> {
 
 impl<
         TContentKey: OverlayContentKey + Send + Sync,
-        TMetric: Metric + Send,
-        TValidator: 'static + Validator<TContentKey> + Send,
+        TMetric: Metric + Send + Sync,
+        TValidator: 'static + Validator<TContentKey> + Send + Sync,
     > OverlayProtocol<TContentKey, TMetric, TValidator>
 where
-    <TContentKey as TryFrom<Vec<u8>>>::Error: Debug + Display,
+    <TContentKey as TryFrom<Vec<u8>>>::Error: Debug + Display + Send,
 {
     pub async fn new(
         config: OverlayConfig,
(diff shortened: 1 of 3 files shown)


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1082,8 +1094,85 @@ where
...
+    /// Process ACCEPT response
+    pub async fn process_accept(
+        &self,
+        response: Accept,
+        enr: Enr,
+        content_keys: Vec<RawContentKey>,
+    ) -> anyhow::Result<Accept> {
+        let content_keys_offered: Result<Vec<TContentKey>, TContentKey::Error> = content_keys
+            .into_iter()
+            .map(|key| (TContentKey::try_from)(key))
```
carver, line 1125: What's with the parens here, would it work this way, too?
```suggestion
            .map(|key| TContentKey::try_from(key))
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1082,8 +1094,85 @@ where
...
+            .collect();
+
+        let content_keys_offered: Vec<TContentKey> = match content_keys_offered {
+            Ok(val) => val,
+            Err(_) => {
+                return Err(anyhow!("Unable to decode offered content keys"));
```
carver, line 1131: Just to highlight how weird this is:
```suggestion
                return Err(anyhow!("Unable to decode our own offered content keys"));
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1272,6 +1361,33 @@ where
...
 
+    /// Provide the requested content key and content value for the acceptor
+    fn provide_requested_content(
+        &self,
+        response: &Accept,
+        content_keys_offered: Vec<TContentKey>,
+    ) -> Vec<ByteList> {
+        let mut content_items: Vec<ByteList> = Vec::new();
+
+        for (i, key) in response
```
carver, line 1372: The name `response` here is confusing to me to read in this context. Naturally, it makes sense in the context of an `accept` responding to an `offer`. But here, it seems like the response ought to be the thing we are building to respond to the `accept`. Maybe we could call this variable `accept` or `accept_message`?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1272,6 +1361,33 @@ where
...
+            .content_keys
+            .clone()
+            .iter()
+            .zip(content_keys_offered.iter())
+        {
+            if i == true {
+                match self.storage.read().get(&key.clone()) {
+                    Ok(content) => match content {
+                        Some(content) => content_items.push(content.into()),
+                        None => {}
```
carver, line 1382: Since we offered this content, a `None` is a bit of a surprise, right? I'm thinking it's worth a log here. Maybe even a `warn!`.

Also, on the receiving side, if a peer skips over some content that we accepted, I think we (and others) might cause a validation failure. So at that point, it might even just be fine to give up altogether and stop collecting items. But I don't have a strong intuition about that yet.

I guess that would change the returned type of this method into a `Result`, and return an `Err` if there was a failure to collect any of the content.

I guess all of this applies to getting an `Err` on the `get()`, also. (except that we already have a `warn!`)

carver (approved):
I would typically expect the bonus radius fix to be in its own commit.

I didn't catch anything worth holding up the PR. 👍🏻

trin-core/src/portalnet/overlay.rs
```diff
@@ -305,7 +305,9 @@ where
                     XorMetric::distance(
                         &content_key_value.0.content_id(),
                         &node.key.preimage().raw(),
-                    ) < node.value.data_radius()
+                    )
+                    .0[3]
+                        < node.value.data_radius().as_u64()
```
jacobkaufmann, line 310: I don't understand what bug we are fixing here. IIUC, this changes the comparison to a comparison between `u64`, where we take the most significant 64 bits of the distance as a `u64` and convert the data radius from a `U256` to a `u64`. This seems like it would introduce a bug.

Also, the call to `U256::as_u64` will [panic](https://docs.rs/ethereum-types/0.13.1/ethereum_types/struct.U256.html#method.as_u64) if the data radius value is greater than `u64::MAX`.

We should not merge this as is.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -309,6 +316,8 @@ impl<
     > OverlayService<TContentKey, TMetric, TValidator>
 where
     <TContentKey as TryFrom<Vec<u8>>>::Error: Debug,
+    TValidator: Sync,
+    TMetric: Sync,
```
jacobkaufmann, line 320: nit: Why not add these bounds in the `impl<..>` above where we already bound with `Send`?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1272,6 +1361,33 @@ where
...
+        let mut content_items: Vec<ByteList> = Vec::new();
+
+        for (i, key) in response
+            .content_keys
+            .clone()
+            .iter()
+            .zip(content_keys_offered.iter())
+        {
+            if i == true {
+                match self.storage.read().get(&key.clone()) {
```
jacobkaufmann, line 1379: Do we need this `clone` here? It looks like we should be able to remove it.

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann, line 1131: nit: We can tidy this up with a `map_err(..)?`.

jacobkaufmann (changes requested):
Looks good aside from the change to the distance-radius comparison, which we should correct before merge. As @carver mentioned, we should isolate these types of changes. Aside from that, nothing much to add beyond existing comments.

trin-core/src/portalnet/overlay.rs
ogenev: I think there are some confusing discrepancies in data radius in our codebase:

Data radius is defined as a `U256` value in the portal network[ specs](https://github.com/ethereum/portal-network-specs/blob/master/history-network.md#radius) but in portal storage, we use u64: https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/storage.rs#L41.
https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/storage.rs#L132

Another confusing thing is how we set the default data radius values in `overlay_service`:
https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/ethportal-peertest/src/jsonrpc.rs#L20

https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/overlay_service.rs#L394

https://github.com/ethereum/trin/blob/43c2527c7f941a015b9d7b322f1a139fcf9a53c0/trin-core/src/portalnet/overlay_service.rs#L1039

I guess the default data radius value should be `U256::MAX` instead of `U256::from(u64::MAX)` and we need to update the storage value to `U256`?

`XorMetric::distance` returns a `U256` value which most of the time is bigger than u64 which makes the filter above ignore all nodes in the routing table because of their default data radius as `u64::MAX`.

I guess reverting this change and adding a default data radius constant as `U256::MAX` + portal storage fix will solve the issue.

jacobkaufmann: This is what #305 was meant to address. I can resurrect that PR, or close it and open a more isolated change. I would like to revisit that and see if it can be improved in any way. I sort of lost track of it after the start of Devconnect.

jacobkaufmann (approved):

PR #346, opened 2022-06-22
Fix seed script bug by properly skipping bodies/receipts

njgheorghita:
### What was wrong?
Recently updated `Header` rlp decoding caused a bug in the `seed-database.rs` script, crashing all of the testnet nodes.

### How was it fixed?
Properly handle bodies / receipts when decoding datastore.

### To-Do

- [x] Clean up commit history


mrferris (approved):
Looks good to me 👍

PR #347, opened 2022-06-22
Validation

njgheorghita:
### What was wrong?
Basic infura validation was required.

### How was it fixed?
Updated the `HeaderOracle` to support validation of `BlockBody` and `Receipts` types.
- Also updated `BlockBody` and `Receipts` types to encode/decode properly according to the [updated spec](https://github.com/ethereum/portal-network-specs/blob/master/history-network.md#content-keys-and-values)

### To-Do

- [x] Clean up commit history


trin-core/src/types/receipts.rs
```diff
@@ -32,6 +33,62 @@ impl Receipts {
...
+}
+
+impl TryFrom<ReceiptsHelper> for Receipts {
+    type Error = ssz::DecodeError;
+
+    fn try_from(helper: ReceiptsHelper) -> Result<Self, Self::Error> {
+        let receipt_list: Vec<Receipt> = helper
+            .rlp_receipt_list
+            .iter()
+            .map(|bytes| rlp::decode(bytes).unwrap())
```
carver, line 86: To my eyes, it's very strange to me that this works. The typed receipts are specifically built so that they cannot be rlp-decoded. I guess it's because there is this magic inside of the `Receipt` class that clips the type byte out before rlp-decoding.

At some point there could be receipts that don't have any RLP encoding in them at all. Following this pattern we've started would still call it `rlp::decode()` here anyway, which feels wrong. So something like `Receipt::decode(bytes)` here would be more satisfying to my eyes.

trin-core/src/types/receipts.rs
```diff
@@ -276,42 +345,72 @@ mod tests {
...
+                rlp::decode(&hex::decode(RECEIPT_9).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_10).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_11).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_12).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_13).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_14).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_15).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_16).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_17).unwrap()).unwrap(),
+                rlp::decode(&hex::decode(RECEIPT_18).unwrap()).unwrap(),
```
carver, line 375: I guess I have the same issue here, with `rlp` not feeling right. If I'm not making sense, then maybe we can have a call?

trin-core/src/types/block_body.rs
```diff
@@ -57,6 +47,73 @@ impl Uncles {
...
+    }
+
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, ssz::DecodeError> {
+        let helper = BlockBodyHelper::from_ssz_bytes(bytes)?.try_into()?;
+        Ok(helper)
+    }
+}
+
+#[derive(Debug, Decode, Encode)]
+struct BlockBodyHelper {
```
carver, line 95: Whenever I write a ...`Helper` name, I'm pretty sure that I'm not done naming it.

But yeah, clearly a more challenging name. `EncodedBlockBodyParts?` ick. `EncodedBlockElements` maybe?

trin-core/src/types/block_body.rs
```diff
@@ -57,6 +47,73 @@ impl Uncles {
...
 }
 
+impl TryFrom<BlockBodyHelper> for BlockBody {
+    type Error = ssz::DecodeError;
+
+    fn try_from(helper: BlockBodyHelper) -> Result<Self, Self::Error> {
+        let txs: Vec<Transaction> = helper
+            .rlp_txs
+            .iter()
+            .map(|bytes| rlp::decode(bytes).unwrap())
```
carver, line 57: Same `rlp` naming awkwardness here.

trin-core/src/types/block_body.rs
```diff
@@ -57,6 +47,73 @@ impl Uncles {
...
+            .collect();
+        let uncles: Vec<Header> = helper
+            .rlp_uncles
+            .iter()
+            .map(|bytes| rlp::decode(bytes).unwrap())
```
carver, line 62: We actually don't need to individually encode each uncle header (and probably shouldn't).

Note that the uncles are not included individually, but as a single encoded element (the list is rlp-encoded in aggregate):
https://github.com/ethereum/portal-network-specs/blob/9c7ad893f52e71ecb3e95f81c14710b2253aff51/history-network.md#block-body

I think this is going to give you different SSZ encodings, looking at how it's built below.

trin-core/src/types/block_body.rs
```diff
@@ -200,80 +269,144 @@ pub struct StorageKeys {
...
     use super::*;
     use rstest::rstest;
+    use ssz::{Decode, Encode};
 
     // tx data from: https://etherscan.io/txs?block=14764013
     #[rstest]
-    #[case(TX1, 102)]
-    #[case(TX2, 1078)]
     #[case(TX3, 454)]
-    #[case(TX4, 455)]
```
carver, line 209: Maybe add a comment explaining why this keeps the transactions that it does. Want to make sure there's coverage of typed and un-typed of course.

trin-history/src/validation.rs
```diff
@@ -195,4 +228,66 @@ mod tests {
             .await
             .unwrap();
     }
+
+    #[tokio::test]
+    #[ignore]
```
carver, line 233: Just a reminder to remove this before merging.

trin-history/src/validation.rs
```diff
@@ -195,4 +228,66 @@ mod tests {
...
+            block_hash: block_hash.0,
+        });
+        chain_history_validator
+            .validate_content(&content_key, &block_body_bytelist)
+            .await
+            .unwrap();
+    }
+
+    #[tokio::test]
+    #[ignore]
```
carver, line 258: Removal reminder

carver (changes requested):
Pretty sure there's an SSZ bug in here, but I'm guessing you're already bumping into that locally...

njgheorghita:
@carver Sorry for any confusion and the premature review request, I had a bit of a brainfart and request the review here instead of on #346. 

Pr's been updated and your comments addressed. The encoding/decoding should be all sorted out now, since they're working with the test values you sent me.

carver:
> It seems like it's possible but not trivial to implement a workaround, but I was curious if this is a high-priority limit that we should have implemented from the start or if it can wait as an issue for now?



> @carver Sorry for any confusion and the premature review request, I had a bit of a brainfart and requested a review here instead of on #346.
> 
> Pr's been updated and your comments addressed. The encoding/decoding should be all sorted out now, since they're working with the test values you sent me.
> 
> One thing to mention that enforcing `MAX_LENGTH` for ssz encoding/decoding is not so straightforward with the library's api. The library defines `u64::MAX` as a constant that it uses for all encoding/decoding. It seems like it's possible but not trivial to implement a workaround, but I was curious if this is a high-priority limit that we should have implemented from the start or if it can wait as an issue for now?

As far as I can tell, it's only relevant when building a partial proof. So it *may* never be relevant. But since some implementations require specifying it up front (like the python one), I figured we might as well specify it now across clients.

So I think it's okay to punt that for now, and when we try to build partial proofs, we'll realize that the max sizes are wrong...

trin-core/src/types/receipts.rs
```diff
@@ -32,6 +35,88 @@ impl Receipts {
...
+        EncodedReceiptList::from_ssz_bytes(bytes)?
+            .try_into()
+            .map_err(|msg: DecoderError| ssz::DecodeError::BytesInvalid(msg.to_string()))
+    }
+}
+
+#[derive(Debug)]
+struct EncodedReceiptList {
+    // list ( rlp receipts )
+    rlp_receipts: Vec<Vec<u8>>,
```
carver, line 67: Encoded, instead of rlp

trin-core/src/types/receipts.rs
```diff
@@ -307,11 +392,45 @@ mod tests {
...
+                Receipt::decode(&hex::decode(RECEIPT_14).unwrap()).unwrap(),
+                Receipt::decode(&hex::decode(RECEIPT_15).unwrap()).unwrap(),
+                Receipt::decode(&hex::decode(RECEIPT_16).unwrap()).unwrap(),
+                Receipt::decode(&hex::decode(RECEIPT_17).unwrap()).unwrap(),
+                Receipt::decode(&hex::decode(RECEIPT_18).unwrap()).unwrap(),
+            ],
+        };
+        let encoded = receipts.as_ssz_bytes();
+
+        let expected = std::fs::read_to_string("./src/types/assets/receipts_14764013.hex").unwrap();
```
carver, line 427: Oh, and after re-encoding the file, it should probably have a suffix other than `.hex`. Maybe... `.data` or `.bin`?

trin-history/src/validation.rs
```diff
@@ -191,8 +275,149 @@ mod tests {
...
+
+        chain_history_validator
+            .validate_content(&content_key, &block_body_bytelist.to_vec())
             .await
             .unwrap();
     }
+
+    #[tokio::test]
+    #[should_panic]
+    async fn invalidate_block_body() {
```
carver, line 311: Probably want to simulate both kinds of failures in independent tests: a tx root failure and an uncle hash failure.

carver (approved):
Yup, looks good to me with some nits.

carver:
There was a CI failure, and I thought it was their problem, because of a weird caching error:
> skipping cache - error checking storage: AccessDenied: Access Denied
	status code: 403, request id: DZ01S50VH47WMPRY, host id: JigSJV0rar0qfI+O9VFCvj/0PSs/GglOMCy6Csy30aAQ/Ssz1TAE25ogayxW6/lDhEizFgZrwTw=


but actually it was the `utp_listener_events` test, which apparently only fails sporadically. I'll open an issue.

PR #348, opened 2022-06-22
Print error instead of crashing if bootnode responds with invalid packet

mrferris:
### What was wrong?

Currently, if any of our bootnodes respond to our initial `PING` with an invalid Discv5 packet, Trin crashes with:
`thread 'tokio-runtime-worker' panicked at 'called Result::unwrap() on an Err value: InvalidRemotePacket', trin-history/src/lib.rs:185:40`

Even if a user reads this and thinks "I should try a different bootnode" or "I should alert the Trin team that their default bootnodes are misbehaving", they don't know which ENR is at fault. Turning on `debug` only prints the Node ID and IP, and is lost in a lot of noise. 

To reproduce the error above, run trin with this particular ultralight bootnode (error is intermittent):
`cargo run -- --networks history --bootnodes enr:-IS4QKczGIvcJ5mpqvzQQGhSNIPhyoBqW5H9OEggKYwmgei7egF3VvXNoCohGEcyUUB9jyBh8qY2iN8sJdK7pRgPP-UDgmlkgnY0gmlwhKRc9EGJc2VjcDI1NmsxoQKKnSTsqwcBYg1atI7dlanT8Mo29std_701sLx0g09yXYN1ZHCCE74`

### How was it fixed?

- An `InvalidRemoteDiscv5Packet` type is added to the `OverlayRequestError` enum.

- All errors while initially pinging bootnodes are caught and an error message is displayed:

```Jun 22 17:46:05.324 ERROR trin_history::network: Received invalid remote discv5 packet while pinging bootnode: enr:-IS4QKczGIvcJ5mpqvzQQGhSNIPhyoBqW5H9OEggKYwmgei7egF3VvXNoCohGEcyUUB9jyBh8qY2iN8sJdK7pRgPP-UDgmlkgnY0gmlwhKRc9EGJc2VjcDI1NmsxoQKKnSTsqwcBYg1atI7dlanT8Mo29std_701sLx0g09yXYN1ZHCCE74```

- This PR also simplifies the catching/printing of these errors in both `trin-state` and `trin-history`.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-history/src/network.rs
+++ b/trin-history/src/network.rs
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use log::debug;
+use log::{debug, error};
 use std::sync::{Arc, RwLock as StdRwLock};
 
 use parking_lot::RwLock;
@@ -107,6 +107,10 @@ impl HistoryNetwork {
                 Err(OverlayRequestError::Discv5Error(error)) => {
                     debug!("Unexpected error while bonding with {} => {:?}", enr, error);
                     return Err(anyhow!(error.to_string()));
+                }
+                Err(OverlayRequestError::InvalidRemotePacket) => {
+                    error!("Received invalid response received after pinging a potentially faulty bootnode: {enr:?}");
+                    continue;
                 }
                 _ => {
                     let msg = format!("Unexpected error while bonding with {enr}");
--- a/trin-history/src/network.rs
+++ b/trin-history/src/network.rs
@@ -1,4 +1,3 @@
-use anyhow::anyhow;
 use log::{debug, error};
 use std::sync::{Arc, RwLock as StdRwLock};
 
@@ -8,7 +7,7 @@ use tokio::sync::mpsc::UnboundedSender;
 use trin_core::{
     portalnet::{
         discovery::Discovery,
-        overlay::{OverlayConfig, OverlayProtocol, OverlayRequestError},
+        overlay::{OverlayConfig, OverlayProtocol},
         storage::{PortalStorage, PortalStorageConfig},
         types::{
             content_key::HistoryContentKey,
@@ -65,7 +64,7 @@ impl HistoryNetwork {
         // The overlay ping via talkreq will trigger a session at the base layer, then
         // a session on the (overlay) portal network.
         for enr in self.overlay.discovery.discv5.table_entries_enr() {
-            debug!("Pinging {} on portal history network", enr);
+            debug!("Attempting bond with bootnode {}", enr);
             let ping_result = self.overlay.send_ping(enr.clone()).await;
 
             match ping_result {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: trin-history/src/network.rs [hunk 1]
```diff
@@ -65,53 +64,16 @@ impl HistoryNetwork {
...
-                Err(OverlayRequestError::Discv5Error(error)) => {
-                    debug!("Unexpected error while bonding with {} => {:?}", enr, error);
-                    return Err(anyhow!(error.to_string()));
-                }
-                _ => {
-                    let msg = format!("Unexpected error while bonding with {enr}");
-                    debug!("{msg}");
-                    return Err(anyhow!(msg));
+                Err(err) => {
+                    error!("{err} while pinging bootnode: {enr:?}");
```
njgheorghita, line 76: I agree with the general spirit of this pr. Displaying the enr is definitely useful. I have a question about `ping_bootnodes()` returning `Ok(())` regardless of whether or not it successfully bonded with the bootnodes. [This comment](https://github.com/ethereum/trin/blob/master/trin-history/src/lib.rs#L184) implies that we should crash if any bootnode bond attempt is unsuccessful. I'm inclined to agree. iiuc the code here negates that behavior. Are you thinking that it's not worth "panicking" if a bootnode bond attempt is unsuccesful? Imo, still panicking but with a more descriptive error message is the ideal solution.

mrferris: Yea, that's a good point. I'll have it panic as well.

mrferris:
@njgheorghita What do you think of just doing this (latest commit)? 

The reason I don't want to call `panic!` here is that I think that it looks like a bug:
```
thread 'tokio-runtime-worker' panicked at 'Received invalid remote discv5 packet while pinging bootnode: enr:-IS4QJBALBigZVoKyz-NDBV8z34-pkVHU9yMxa6qXEqhCKYxOs5Psw6r5ueFOnBDOjsmgMGpC3Qjyr41By34wab1sKIBgmlkgnY0gmlwhKEjVaWJc2VjcDI1NmsxoQOSGugH1jSdiE_fRK1FIBe9oLxaWH8D_7xXSnaOVBe-SYN1ZHCCIyg', trin-history/src/network.rs:76:21
```

Whereas this looks like it's what we meant to do if a bootnode fails the ping: 

```
Jun 29 16:55:01.154 ERROR trin_history::network: Received invalid remote discv5 packet while pinging bootnode: enr:-IS4QJBALBigZVoKyz-NDBV8z34-pkVHU9yMxa6qXEqhCKYxOs5Psw6r5ueFOnBDOjsmgMGpC3Qjyr41By34wab1sKIBgmlkgnY0gmlwhKEjVaWJc2VjcDI1NmsxoQOSGugH1jSdiE_fRK1FIBe9oLxaWH8D_7xXSnaOVBe-SYN1ZHCCIyg
```

carver:
So there's two questions here:
1. Do we exit early if we get an invalid response from the bootnode? (just logging an error message doesn't exit, of course)
2. Do we use `panic!` to exit early?

For 1: I think for issues at the beginning of the the client running, that we might as well fail hard and fast. Of course, if only one of several bootnodes is behaving badly, then we probably don't want to exit out of the client early. We might save that for the case when all bootnodes are behaving badly. Of course that takes more work to implement, and I think it's fine to punt that a bit longer, till Beta, of that works for you.

For 2: Sure, I agree that a nicely-formatted error log and a cleaner exit than a panic would better convey that our problem is not having any valid bootnodes rather than a local execution bug.

mrferris:
Yea, I've started thinking similarly about distinguishing between a single bootnode failing the ping and all bootnodes failing the ping. I agree with punting that feature to Beta.

I'm not sure off the top of my head of the extent that running with only a portion of intended bootnodes affects a node's ability to interact with the entire network. I'd imagine that more bootnodes results in a "healthier" routing table, and faster content lookups. Maybe that's only true for a short period of time after the node boots, though. We can do some experiments with the alpha network.

trin-history/src/network.rs
```diff
@@ -65,53 +64,17 @@ impl HistoryNetwork {
...
+                    std::process::exit(1);
```
njgheorghita, line 77: My bad, I should've been a bit more descriptive with my words. By "panicking" I didn't necessarily mean an explicit panic, but just return an error which will cause a panic when in gets unwrapped [here](https://github.com/ethereum/trin/blob/master/trin-history/src/lib.rs#L100)

njgheorghita (approved):
lgtm! just a small change recommendation.

trin-history/src/network.rs
mrferris: Yea but regardless of where we panic (explicitly or due to unwrap or error) it will still appear to the user as an error on our end, rather than an issue with the bootnode. With this approach we get a clean end of the program. There's a crate called `human-panic` that could end our program nicely but that seems like overkill compared to this approach.

PR #353, opened 2022-06-23
Log during tests

carver:
### What was wrong?

Logs were not being initialized/generated during test runs. It's so much nicer to debug a failing test when the logs are shown on failure.

Also, bonus: **disable state network by default**. This was discussed on a call. Soon, history will be properly supported and state won't, so don't launch state by default.

### How was it fixed?

Add the test-log crate to use an easy attribute that initializes logging during tests.

Sometimes, when there are sibling `rstest` attributes, or a `tokio::test`, we need to manually wrap the macro.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-history/Cargo.toml
+++ b/trin-history/Cargo.toml
@@ -27,3 +27,8 @@ tracing = "0.1.26"
 tracing-subscriber = "0.2.18"
 tokio = { version = "1.8.0", features = ["full"] }
 trin-core = { path = "../trin-core" }
+
+[dev-dependencies]
+env_logger = "0.8.2"
+test-log = { version = "0.2.10", features = ["trace"] }
+tracing-subscriber = "0.2.18"
--- a/trin-state/Cargo.toml
+++ b/trin-state/Cargo.toml
@@ -24,3 +24,8 @@ tracing-futures = "0.2.5"
 tokio = {version = "1.8.0", features = ["full"]}
 trin-core = { path = "../trin-core" }
 rocksdb = "0.18.0"
+
+[dev-dependencies]
+env_logger = "0.8.2"
+test-log = { version = "0.2.10", features = ["trace"] }
+tracing-subscriber = "0.2.18"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,7 @@ jobs:
             tag: 1.58.0
         environment:
             RUSTFLAGS: '-D warnings'
+            RUST_LOG: 'debug'
         steps:
             - checkout
             - restore_cache:
@@ -45,7 +46,7 @@ jobs:
                 command: cargo build --workspace --jobs 2
             - run:
                 name: Test Trin workspace
-                command: cargo test --workspace --jobs 2 -- --nocapture
+                command: cargo test --workspace --jobs 2
             - save_cache:
                 key: cargo-{{ checksum "Cargo.lock" }}-v1
                 paths:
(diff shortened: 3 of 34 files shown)


carver:
I don't really know who to send this to, so I just went with the github recommendation 🤷🏻‍♂️

trin-history/Cargo.toml
```diff
@@ -27,3 +27,8 @@ tracing = "0.1.26"
 tracing-subscriber = "0.2.18"
 tokio = { version = "1.8.0", features = ["full"] }
 trin-core = { path = "../trin-core" }
+
+[dev-dependencies]
+env_logger = "0.8.2"
```
jacobkaufmann, line 32: For `trin-core` we only import `test-log` and `tracing-subscriber`. Based on the [`test-log` crate docs](https://crates.io/crates/test-log), we only need one of `test-log` or `env_logger`, so I'm not sure why we add both as dependencies here.

trin-state/Cargo.toml
```diff
@@ -24,3 +24,8 @@ tracing-futures = "0.2.5"
 tokio = {version = "1.8.0", features = ["full"]}
 trin-core = { path = "../trin-core" }
 rocksdb = "0.18.0"
+
+[dev-dependencies]
+env_logger = "0.8.2"
```
jacobkaufmann, line 29: Same question as above regarding dependency on `env_logger`.

jacobkaufmann (commented):
Just some minor questions in the comments. Happy to approve once I have clarity there. This is definitely an improvement for debugging failing tests. I assume that we validated that this does indeed work for failed tests whose output is not otherwise captured.

One thing that I did notice in the [CI test runs](https://app.circleci.com/pipelines/github/ethereum/trin/1538/workflows/a2181bcb-9ba0-4790-9508-cecdd9190772/jobs/3044) is that there was some printout under `test utp::packets::tests::test_reversible ... ok` that I do not see when I run the test suite locally. I'm not sure why this is, but we should make sure that we don't see log output on successful tests.

carver:
> One thing that I did notice in the [CI test runs](https://app.circleci.com/pipelines/github/ethereum/trin/1538/workflows/a2181bcb-9ba0-4790-9508-cecdd9190772/jobs/3044) is that there was some printout under `test utp::packets::tests::test_reversible ... ok` that I do not see when I run the test suite locally. I'm not sure why this is, but we should make sure that we don't see log output on successful tests.

Yup, I'm guessing you were not running tests locally with `-- --nocapture`. What you're seeing is actually not logging, it's some other stdout output. But it did highlight another bug which is that `RUST_LOG` has to be set for the logs to show up during testing. I'll add that, and turn off `--nocapture` in CI, so the output isn't overwhelming.

> I assume that we validated that this does indeed work for failed tests whose output is not otherwise captured.

Yes, I did run tests locally to see the log output on failed tests. But I'll also run a test in CI to confirm there, with the above change.

trin-history/Cargo.toml
jacobkaufmann: I see. I figured that there was some reason for the inclusion of `env_logger`.

carver:
Well, I figured everything out except for `tests/self_peertest.rs`. If I turn off `--nocapture` and use `test_log`, then it doesn't show anything at all on failure. Seems better to just show too much for now, until we figure out what's going on there. But it seems like a distraction to keep digging right now. I'll write up what I know in an issue.

carver:
> Happy to take another look whenever it's ready.

Ok, ready for your eyes again, thanks!

jacobkaufmann (approved):

Issue #355, opened 2022-06-24
Why does peertest::launch_peertest_nodes() break the capturing of test output?


carver:
See this PR to play along: #356 

In these three scenarios:
1. remove both `assert!(false);` lines
1. remove second `assert!(false);` line (or really, no change at all)
1. remove first `assert!(false);` line, before the `launch_peertest_nodes()` call

I expect these three outcomes:
1. all tests pass, no log output
2. one test fails, showing `Jun 24 13:41:58.030 DEBUG self_peertest::test: debug log`
3. one test fails, showing the above log, and several more from the launched peers

I actually get these three outcomes:
1. all tests pass, no log output
2. one test fails, showing `Jun 24 13:41:58.030 DEBUG self_peertest::test: debug log`
3. one test fails, no output captured at all

njgheorghita:
looks like this was resolved in #353

PR #356, opened 2022-06-24
[demo] weird crashing test behavior

carver:
The test does not capture log/stdout on a failure, if it fails after
launching the peertest nodes...

### What was wrong?

#355 

### How was it fixed?

This is not a fix, it's a problem demonstration


KolbyML:
Closing as this bug was resolved already https://github.com/ethereum/trin/issues/355

PR #357, opened 2022-06-27
Add node in overlay routing table before returning overlay response

ogenev:
### What was wrong?

We are responding to an overlay request before updating the overlay routing table with the remote `NodeId`.

This can lead to some unexpected behavior, for example when we receive an `OFFER` request, we want to check the overlay routing table if the node already exists and start listening for incoming uTP packets, **before** sending the `ACCEPT` response.

For reference, the discv5 implementation first adds the remote node to the routing table (after a successful handshake) and then responds with a talk response.

### How was it fixed?
1. Extract a new method from existing code encapsulating the logic for overlay routing table update.
2. Call the method before building the overlay response.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -694,6 +694,8 @@ where
         // channel if the request was initiated internally (e.g. for maintenance).
         match request.direction {
             RequestDirection::Incoming { id, source } => {
+                self.register_node_activity(source);
+
                 let response = self.handle_request(request.request.clone(), id.clone(), &source);
                 // Send response to responder if present.
                 if let Some(responder) = request.responder {
@@ -939,6 +941,14 @@ where
 
     /// Processes an incoming request from some source node.
     fn process_incoming_request(&mut self, request: Request, _id: RequestId, source: NodeId) {
+        match request {
+            Request::Ping(ping) => self.process_ping(ping, source),
+            _ => {}
+        }
+    }
+
+    /// Register source NodeId activity in overlay routing table
+    fn register_node_activity(&mut self, source: NodeId) {
         // Look up the node in the routing table.
         let key = kbucket::Key::from(source);
         let is_node_in_table = match self.kbuckets.write().entry(&key) {
@@ -975,11 +985,6 @@ where
                 };
                 self.connect_node(node, ConnectionDirection::Incoming);
             }
-        }
-
-        match request {
-            Request::Ping(ping) => self.process_ping(ping, source),
-            _ => {}
         }
     }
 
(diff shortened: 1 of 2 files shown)


carver (approved):
Makes sense to me

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -939,6 +941,14 @@ where
...
     /// Processes an incoming request from some source node.
     fn process_incoming_request(&mut self, request: Request, _id: RequestId, source: NodeId) {
+        match request {
+            Request::Ping(ping) => self.process_ping(ping, source),
+            _ => {}
+        }
+    }
+
+    /// Add source NodeId to overlay routing table or update its connection state
+    fn update_routing_table(&mut self, source: NodeId) {
```
jacobkaufmann, line 951: This function name is quite vague, so if you see it being called somewhere, then it's not clear what it's doing. A more specific name would improve readability. We already have a function named `connect_node`, so we cannot use that. This function is sort of like a "touch" because we also want to update the node's position in its k-bucket if it's already present.

That said, I don't feel great about `touch_node` either, but hopefully that idea suggests a more useful name.

jacobkaufmann (commented):
The rationale for this logic ordering was that we should not delay the response for any processing. In other words, all processing associated with an incoming request should be taken care of after we respond to the request to minimize latency.

ogenev:
> The rationale for this logic ordering was that we should not delay the response for any processing. In other words, all processing associated with an incoming request should be taken care of after we respond to the request to minimize latency.

We still process the incoming requests after we respond, the only "processing"  before a response is adding/updating the node in the overlay routing table.

In the current implementation, considering the following scenario, there are no guarantees that Node A will be added to Node B's routing table after step 2) and before 3):

1. Node A sends OFFER request to Node B
2. Node B responds with ACCEPT message.
3. Node A sends uTP SYN packet over talk req to node B to initialize uTP stream.

jacobkaufmann:
Yes I understand the reason for the change. I would note that, if the node is not already in the routing table, then the function to add the node to the routing table would only be able to add that node if the node exists in the underlying `Discv5` routing table. A similar lookup could be done in the handle offer flow, and we could keep things the way they are now. That fallback logic will soon be replaced by a lookup to the ENR cache, which I am working on currently.

jacobkaufmann:
@carver the only concern is performance, and those concerns are probably negligible, at least at this time. The PR for the ENR cache and its integration is further away than a day or two, so we can move forward with this change. I think that is more preferable than duplicating the fallback lookup logic in the handle offer flow.

trin-core/src/portalnet/overlay_service.rs
ogenev: I like `register_node_activity` most :1st_place_medal:.

PR #359, opened 2022-06-28
Add OFFER/ACCEPT test

ogenev:
### What was wrong?
Integration tests for offer/accept flow are missing.

### How was it fixed?
Add offer/accept scenario as part of `ethportal-peertest`

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -11,7 +11,7 @@ use crate::{cli::PeertestConfig, Peertest};
 use trin_core::{
     jsonrpc::types::{NodesParams, Params},
     portalnet::types::{
-        content_key::{AccountTrieNode, BlockHeader, HistoryContentKey, StateContentKey},
+        content_key::{AccountTrieNode, StateContentKey},
         messages::SszEnr,
     },
     utils::bytes::hex_encode,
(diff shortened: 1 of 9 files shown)


ethportal-peertest/src/scenarios.rs
```diff
@@ -0,0 +1,42 @@
...
+            Value::Array(vec![json!(HISTORY_CONTENT_KEY)]),
+        ]),
+    };
+
+    let accept = make_ipc_request(&peertest_config.target_ipc_path, &offer_request).unwrap();
+
+    // Check accept response
+    validate_portal_offer(&accept, peertest);
+
+    // TODO: Check if the stored content item in bootnode's DB matches the offered
```
ogenev, line 41: This is blocked by https://github.com/ethereum/trin/pull/277.

carver: Why? Is it to make sure every test starts with an empty database?

ogenev: > Why? Is it to make sure every test starts with an empty database?

Yeah, on OFFER/ACCEPT the receiving node stores the content and it seems we don't clean the peertest databases after all tests are done.

ethportal-peertest/src/lib.rs
```diff
@@ -64,8 +66,7 @@ pub async fn launch_node(id: u16, bootnode_enr: Option<&SszEnr>) -> anyhow::Resu
             TrinConfig::new_from(trin_config_args.iter()).unwrap()
         }
         None => {
-            let ip_addr =
-                socket::find_assigned_ip().expect("Could not find an IP for local connections");
```
carver, line 68: I think this change isn't necessary, if we change the `--external-address` in tests/self_peertest.rs to use the same IP as the peertest nodes, similar to this derivation:
https://github.com/ethereum/trin/blob/65665e39c4dd831ed0df5f4cd5f753aaf1b9febb/ethportal-peertest/src/lib.rs#L43-L47

If it *is* necessary, then there are some other cleanups we should do, including this ^ code.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -364,6 +370,15 @@ fn validate_portal_routing_table_info(result: &Value, _peertest: &Peertest) {
     assert!(result.get("numConnected").unwrap().is_u64());
 }
 
+pub fn validate_portal_offer(result: &Value, _peertest: &Peertest) {
+    println!("result: {}", result);
```
carver, line 374: Maybe left in by accident, and should be deleted? If you think we should keep it, then I might expect it to be a `debug` log, with a bit more info.

tests/self_peertest.rs
```diff
@@ -16,13 +17,18 @@ mod test {
 
         let peertest_config = peertest::PeertestConfig::default();
 
+        let test_ip_addr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+        let test_port = 9000;
+        let external_addr = format!("{}:{}", test_ip_addr, test_port);
```
carver, line 22: Could instead extract a matching IP from something like: `peertest.bootnode.enr.ip4()`

ethportal-peertest/src/lib.rs
ogenev: There are some things that concern me:

1. `find_assigned_ip` is not stable, it is not returning the same IP interface on every run. What is the point then of using this method if it is returning random results? I'm more in favor to get rid of this method and adding helper method that always returns `127.0.0.1` for example.

2.  Running `find_assigned_ip` just to get some random IP for one node and then use the same IP for the next node is kinda confusing, at least for me.
I think it is less mentally overwhelming to just run every node with a fixed static address which is not a result of a random method.

ethportal-peertest/src/scenarios.rs
```diff
@@ -0,0 +1,42 @@
...
+        params: Params::Array(vec![
+            Value::String(HISTORY_CONTENT_KEY.to_string()),
+            Value::String(HISTORY_CONTENT_VALUE.to_string()),
+        ]),
+    };
+
+    let store_result = make_ipc_request(&peertest_config.target_ipc_path, &store_request).unwrap();
+    assert_eq!(store_result.as_str().unwrap(), "true");
+
+    // Send offer request
```
njgheorghita, line 26: A bit more of a descriptive comment could be nice ... `// Send offer request from testnode to bootnode`?

ethportal-peertest/src/scenarios.rs
```diff
@@ -0,0 +1,42 @@
...
+        id: 11,
+        params: Params::Array(vec![
+            Value::String(peertest.bootnode.enr.to_base64()),
+            Value::Array(vec![json!(HISTORY_CONTENT_KEY)]),
+        ]),
+    };
+
+    let accept = make_ipc_request(&peertest_config.target_ipc_path, &offer_request).unwrap();
+
+    // Check accept response
```
njgheorghita, line 38: `// Check that accept response sent by bootnode accepted the offered content`?

njgheorghita (approved):
This generally checks out to me. I'm not too excited about the way this is laid out structurally. Things are getting a bit messy, with various functions in different places. What exactly distinguishes a "scenario" from a regular test? (just a comment at the top of `scenarios.rs` would help in the meantime). Maybe all the `validate_*` functions belong in their own file? This disorganization is not the fault of this pr, and was bound to emerge as more test cases are added. But, I think we're quickly approaching the point where some organization is needed as we're adding more and more complex tests. It would also be nice to see a tests in a later pr for the cases where the offer is not accepted, and where the testnode cannot offer the content since it's not stored locally in its db.

ogenev:
> Things are getting a bit messy, with various functions in different places. What exactly distinguishes a "scenario" from a regular test? (just a comment at the top of scenarios.rs would help in the meantime).

I agree, `ethportal-peertest` started as a way to test a single json-rpc endpoint and now we want to add more complex tests that involve a combination of multiple json-rpc endpoints that lead to desired behavior which we can call "scenarios" or test cases or whatever, any suggestions are welcome.

I'm more believer in first creating a mess by hacking something that works and is valuable and then refactoring the system to make it more maintainable and readable. I think this way we can iterate faster, and get more information along the way because it is hard to predict all use cases and bottlenecks. So I guess at this stage a little mess is acceptable but we should start to refactor as soon as it gets unbearable and returns diminishing results.

ogenev:
Do not merge before https://github.com/ethereum/trin/pull/364.

ethportal-peertest/src/lib.rs
carver: Yeah, I think I used to have a reason I didn't like using 127.0.0.1 here (maybe I wanted to be able to connect to non-local clients?). But I can't seem to defend it now, so I'm good with using localhost. As more tests get added, it's clearer that the complexity cost of picking a constant IP is growing...

There's probably some more work that can be done to simplify, and use 127.0.0.1 everywhere in the tests.

PR #360, opened 2022-06-28
cleanup: remove dead code for node discovery

jacobkaufmann:
### What was wrong?

I came across some dead code for `Discovery`. Based on the blame, it was added around May 2021.

### How was it fixed?

Remove dead code.

### To-Do

- [x] Clean up commit history


carver (approved):
LGTM!

PR #361, opened 2022-06-29
refactor: remove binaries for trin-history and trin-state

jacobkaufmann:
### What was wrong?

The binaries for `trin-state` and `trin-history` are unused and are not necessary to run the individual subnetworks.

### How was it fixed?

Remove `trin-state` and `trin-history` binaries so that `trin` is the single entry point.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-history/src/lib.rs
+++ b/trin-history/src/lib.rs
@@ -15,98 +15,13 @@ use tokio::{
 
 use crate::{events::HistoryEvents, jsonrpc::HistoryRequestHandler};
 use trin_core::{
-    cli::TrinConfig,
     jsonrpc::types::HistoryJsonRpcRequest,
     portalnet::{
-        discovery::Discovery,
-        events::PortalnetEvents,
-        storage::{PortalStorage, PortalStorageConfig},
-        types::messages::PortalnetConfig,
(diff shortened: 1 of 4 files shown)


njgheorghita (approved):
:ship: Just curious how far you're thinking of taking this refactor? Did you give any thought to removing the `--networks` cli arg as well? It could help clean up some of the code in `src/lib.rs`. The default is currently to only activate `history` network... which works... but we technically support activating just the state network, which is something I don't think we will ever want to support (for validation reasons, etc.). At some point down the road, we might find some nice groupings of networks that a user can activate without needing to activate every network, but currently it seems like we're moving ahead with the heavy option of requiring users to join every network trin supports. Imo.. it's not an urgent issue, but that could be a source for more more cleanup / code removal.

carver:
> we might find some nice groupings of networks that a user can activate without needing to activate every network

Yeah, I can imagine us heading this direction. Arguably, we don't know what pairings these are yet, so I'm 👍🏻 for punting on trying to make any changes to the network options for now.

jacobkaufmann:
@njgheorghita I will limit the refactor here to the initial changes. I would like to do a focused refactor on network initialization in the future.

PR #369, opened 2022-07-01
Add LRU cache of content rejected by database

### LINKED ISSUE #368: Propagating content that's not in the DB
Currently, Offer/Accept requires the data to be in the database. There is at least one situation where we want to offer content that may not be in the local database, like with the bridge node(s).

The code that makes the offer and handles the accept is decoupled, so it's not straightforward to hold onto the values in a variable between the two steps, without the database. Currently, the best idea I have is to keep an LRU cache of recently used content. Then we could serve an `Accept` from the cache. It's a little awkward that it's *possible* to offer the content and then have it revoked by the LRU, but I think we could come up with configurations that make that chance vanishingly small in the short time between offer & accept.

As a bonus, having a general LRU cache is something we've occasionally talked about for other cases (like helping the network by storing really hot data outside of your radius). So it might have side benefits.

Side note: the bridge currently uses `portal_historyStore`. Maybe we should make a distinct `portal_historyOfferValue` endpoint that accepts the value to send, rather than add propagate as a sequential step after storing?

carver:
### What was wrong?

Fix #368 

### How was it fixed?

Add a cache of content, even if it's not stored durably to disk. This is an easy way for content to be re-offered to peers, even if outside of the radius.

It has the fairly obvious problem, as a limited cache, of potentially having the content drop out of the cache between offer and accept. The most troubling part of that is that the cache might be externally controlled somehow, by flooding with content.

Given that issue, we may conclude in the future that there is no reasonable way to use the same data structure to both: 1) serve the hotspot data, and 2) save data between offer & accept. But I think this solution should work well enough to get an alpha network running.

### To-Do

- [x] Clean up commit history
- [x] move magic number to constant


trin-core/src/portalnet/storage.rs
```diff
@@ -16,6 +18,9 @@ use super::types::{
 };
 use crate::utils::db::get_data_dir;
 
+// Assuming large 15kB content values, 5k elements consume 75MB of data
+const CACHE_NUM_ELEMENTS_MAX: usize = 5000;
```
mrferris, line 22: @carver What do you think of calculating this as a percentage of the `PortalStorage`'s storage capacity? Our default storage capacity is currently 100MB, almost doubling that for the cache seems like a lot. Maybe make this 15-25% of storage capacity?

carver: Since we don't offer any limit on RAM usage in trin, there's no obvious setting to calculate this limit from. I'm much more hesitant to permanently use up disk space than I am to temporarily use up memory. Plus, 75MB is a worst-case scenario of storing huge 15kb items, I expect most items would be significantly smaller.

mrferris (approved):
LGTM!

trin-core/src/portalnet/storage.rs
```diff
@@ -141,6 +148,24 @@ impl PortalStorage {
...
+    ) -> Result<(), PortalStorageError> {
+        match self.durable_store(key, value) {
+            Ok(()) => Ok(()),
+            Err(e) => {
+                // When the data cannot be stored locally, save it to an LRU. This supports
+                // a few use cases, like:
+                // - offering all portal_historyStore content to peers, ignoring radius
+                // - serving hotspots in network, outside of radius
+                self.cache.write().put(key.content_id(), value.clone());
+                Err(e)
```
ogenev, line 160: I think returning an error here is a little dangerous because the caller will not expect this method to change any external state (in this case LRU) when this method returns an error.

carver: Yeah, I'm considering reverting this whole PR anyway. If it was going to work without any fiddling I might have been happy, but there are just too many little quirks like this one. Another one I've seen is some kind of surprising race condition or something that causes the data to be missing, occasionally.

I think I see the path to the alternative now. To thread the values through in memory, so that we don't rely on the database to answer the Accept.

PR #370, opened 2022-07-01
Remove generate data script

### LINKED ISSUE #366: Can we remove the `generate-data.rs` script?
iiuc, this is an outdated script from when we first were testing out portal storage. I'd be surprised if there's still a need for this? Can anybody think of a reason to keep this script around? 

https://github.com/ethereum/trin/blob/master/trin-core/src/bin/generate-data.rs

njgheorghita:
### What was wrong?
Fixes #366

### How was it fixed?
- Removes no longer needed `generate-data.rs` script
- Adds quickcheck to fuzz storing data in `PortalStorage`

### To-Do

- [x] Clean up commit history


njgheorghita:
> I think that something inspired by it could make a useful test in trin-core/tests, though.

@mrferris What kind of a useful test were you thinking of? afaict, all the script does is config a `PortalStorage` and then shove random data into there. I added `quickcheck` to the `PortalStorage::store()` test - which I think covers what you're talking about? But lmk if I'm missing your point.

PR #371, opened 2022-07-06
feat: add content lookup method to overlay

jacobkaufmann:
### What was wrong?

There does not exist a way to perform a content lookup external to `OverlayService`.

### How was it fixed?

- Refactor the overlay request channel to accept "commands" for both requests and find content queries.
- Add a public method on `OverlayProtocol` to perform a content lookup for a target content key.

### To-Do

- [x] Clean up commit history


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -79,6 +79,20 @@ const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
 /// Bucket refresh lookup interval in seconds
 const BUCKET_REFRESH_INTERVAL_SECS: u64 = 60;
 
+/// An action that the overlay may perform.
+#[derive(Debug)]
```
ogenev, line 83: I think it is a good practice to implement also `Clone` and `PartialEq/Eq` traits.

ogenev (approved):
LGTM :+1:

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -286,10 +300,10 @@ pub struct OverlayService<TContentKey, TMetric, TValidator> {
     peers_to_ping: HashSetDelay<NodeId>,
     // TODO: This should probably be a bounded channel.
     /// The receiver half of the service request channel.
```
njgheorghita, line 302: `half of the service command channel.`?

njgheorghita (approved):
No concerns with this code, but I find myself being confused with the "command" vs "request" naming. imo these words are pretty much interchangeable, so something extra to help distinguish them might be useful. tbh, nothing obvious stands out to me as a good option. Even something like "NativeRequest" vs "ExternalRequest" is tough... what is native, a request sent over the wire, or a request initiated over rpc? "UserRequest"... might be good for a request over rpc, but it's not satisfying tbh. Explicitly defining the difference between a "command" and "request" and how they are used would be sufficient.

`/// An action that the overlay may perform.` - is pretty vague imo...
`/// Send a request through the overlay.` - also somewhat vague...
`/// Perform a find content query through the overlay.`  - iiuc, this refers to actions initiated from jsonrpc? might be worth mentioning that.

Maybe we reserve & define a `"query"` as any action originating from jsonrpc?
I'm assuming we'll end up adding more queries to `OverlayCommand`? 

Anyways, sorry for complaining and not having a better solution. Essentially, "command" and "request" and their context confuse me as I read through the code.

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann: I don't think those traits apply here. You cannot clone `oneshot::Sender`, and I cannot think of any circumstances where we would compare two commands.

jacobkaufmann:
@njgheorghita yea I agree that the naming is not great. I would like to refactor `OverlayRequest` to something like `PortalRequest`, because those messages correspond to the portal wire spec. I'm not sure what a better name than "command" or "action" would be. The "command" name is purposefully general to accommodate a variety of possible messages in the channel.

I would like to defer a naming refactor to a later PR. Here, I can't think of a better name than "command" so I'm going to leave this as is for now, unless you have a suggestion.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -79,6 +79,20 @@ const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
 /// Bucket refresh lookup interval in seconds
 const BUCKET_REFRESH_INTERVAL_SECS: u64 = 60;
 
+/// An action that the overlay may perform.
+#[derive(Debug)]
+pub enum OverlayCommand<TContentKey> {
+    /// Send a request through the overlay.
+    Request(OverlayRequest),
+    /// Perform a find content query through the overlay.
+    FindContentQuery {
```
carver, line 88: > Here, I can't think of a better name than "command" so I'm going to leave this as is for now, unless you have a suggestion.

Yeah, I don't have a better suggestion either, so no objection to leaving it as command. But I highly recommend the following suggestion:

> Explicitly defining the difference between a "command" and "request" and how they are used would be sufficient.

The comments in 82 and 85 could use some love. How does a request differ from a command here, semantically? Is it that a Command is any kind of action we might ask the overlay service loop to perform (which might include multiple back-and-forth network requests), and a Request always has a single network request that it maps to 1:1?

If I have that wrong and it's hard to explain even with a couple sentences, that could be an important signal that we haven't found the right architectural choice yet.

jacobkaufmann: Agreed. You've described the basic idea that I had in mind.

PR #372, opened 2022-07-08
Improve Routing Table Initialization 

mrferris:
### What was wrong?

A trin node currently initializes its overlay routing table at boot by running a find nodes query for only its own node ID, with queries for node IDs in farther buckets being run once per minute. With `EXPECTED_NON_EMPTY_BUCKETS` at 17, it will take 17 minutes from boot to run a query for node IDs in all of the expected buckets.

### How was it fixed?

A FindNode query for each of the expected non-empty buckets is run at boot.

**Also**: 

- Cleans up `overlay_service`'s `!select` loop by breaking out findnode query logic.
- Consolidates multiple variations of `init_find_nodes_query` functions into single function.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -1764,8 +1764,24 @@ where
         Ok(closest_nodes)
     }
 
-    /// Starts a FindNode query to find nodes with IDs closest to target, starting with given enrs.
-    fn init_find_nodes_query_with_initial_enrs(&mut self, target: &NodeId, enrs: Vec<Enr>) {
+    /// Starts a FindNode query to find nodes with IDs closest to `target`.
+    fn init_find_nodes_query(&mut self, target: &NodeId) {
+        let target_key = Key::from(*target);
+        let mut closest_enrs: Vec<Enr> = self
+            .kbuckets
+            .write()
+            .closest_values(&target_key)
+            .map(|closest| closest.value.enr)
+            .collect();
+
+        // `closest_enrs` will be empty if `target` is our local node ID
+        // due to the behavior of `closest_values`. In this case, set closest_enrs
+        // to be all ENRs in routing table.
+        if closest_enrs.is_empty() {
+            let mut all_enrs: Vec<Enr> = self.table_entries_enr();
+            closest_enrs.append(&mut all_enrs);
+        }
+
         let query_config = QueryConfig {
             parallelism: self.query_parallelism,
             num_results: self.query_num_results,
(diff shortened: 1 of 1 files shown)


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1773,7 +1773,29 @@ where
     }
 
     /// Starts a FindNode query to find nodes with IDs closest to target, starting with given enrs.
-    fn init_find_nodes_query_with_initial_enrs(&mut self, target: &NodeId, enrs: Vec<Enr>) {
+    fn init_find_nodes_query(&mut self, target: &NodeId) {
```
carver, line 1781: docstring got a little stale (no `given enrs` are supplied). Anything else that should be added?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -640,7 +645,7 @@ where
                         None,
                         Some(query_id),
                     );
-                    let _ = self.request_tx.send(request);
+                    let _ = self.command_tx.send(OverlayCommand::Request(request));
```
carver, line 648: Can you help me understand why this change is needed or better? Something feels wrong about them both being valid options. Or is this also fixing some kind of bug?

carver (approved):
Yup, LGTM!

>  With `EXPECTED_NON_EMPTY_BUCKETS` at 17, it will take 17 minutes from boot to run a query for node IDs in all of the expected buckets.

This is awkward of course, it seems most likely that we won't get this number exactly right most of the time. So we'll be spamming with too many requests when the network is smaller than expected, and we'll be leaving some buckets undiscovered when the network is bigger than expected. I don't have any great solution here, it's just something that we'll probably want to revisit with a more general solution. (Like a more aggressive bucket filling strategy while our buckets are near-empty, that backs off when they get fuller)

trin-core/src/portalnet/overlay_service.rs
mrferris: Upon investigation, that was the third commit making up for a mistake during a rebase on the first commit. I re-did the commits to fix that, thanks!

mrferris:
@carver Agreed, I did a good chunk of work on dynamic initialization while working on this PR but decided to merge this step first. 

The dynamic approach I'm going to do in a next PR is the one that the Kademlia paper specifies: doing only a query for your own local ID, and then once that's done starting queries targeting only the buckets that were populated with at least one node during the first query.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1744,8 +1769,30 @@ where
...
+            .write()
+            .closest_values(&target_key)
+            .map(|closest| closest.value.enr)
+            .collect();
+
+        // `closest_enrs` will be empty if `target` is our local node ID
+        // due to the behavior of `closest_values`. In this case, set closest_enrs
+        // to be all ENRs in routing table.
+        if closest_enrs.is_empty() {
+            let mut all_enrs: Vec<Enr> = self
```
njgheorghita, line 1786: It seems to me you can use `self.table_entries_enr()` here

njgheorghita (approved):
:ship:

PR #373, opened 2022-07-08
Remove blocking async from overlay service loop

### LINKED ISSUE #342: Decouple blocking content validation from main Overlay Service loop
This thought relates to content validation code that was merged in #331. Let me know if I'm wrong, or if this is a duplicate/already discussed. 

I think that we should remove any blocking async code in the main `overlay_service.rs` loop, which currently occurs [here](https://github.com/ethereum/trin/blob/62734135570a7ca9ee0515dc124c61281863c6f5/trin-core/src/portalnet/overlay_service.rs#L419). This line's `await` was added in order to call the chain of `process_content`->`process_received_content`->`validate_content`, where `validate_content` is a blocking async function that makes this entire chain of functions need to be async.

The reasoning for avoiding this is that as long as this function is being `await`ed on, the `select!` loop cannot proceed. This means that no queries can progress, requests from the overlay cannot be acted upon, our responses to other nodes cannot happen (leading to peer requests timing out / us being marked disconnected), bucket refreshes can't happen, etc. 

Under nominal conditions this shouldn't block for long, but an infura request taking longer than expected would turn our node into a brick for that time. And later it gets even worse if we need to request a new accumulator snapshot in order to validate data: the content query to do that won't be able to progress, leaving us deadlocked unable to get the content that we need in order to validate. 

An alternative would be to send our unvalidated content to a dedicated content validation thread, which will then store it if valid.

njgheorghita:
### What was wrong?
Fixes #342 

`async` expressions in the handlers of main `OverlayService` loop block the loop from continuing while they are awaited
- validation of content
- processing an accept message

### How was it fixed?
Spawn new tasks for `async` expressions.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -68,6 +68,7 @@ use ssz::Encode;
 use ssz_types::{BitList, VariableList};
 use thiserror::Error;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::Mutex;
 
 /// Maximum number of ENRs in response to FindNodes.
 pub const FIND_NODES_MAX_NODES: usize = 32;
@@ -317,11 +318,11 @@ pub struct OverlayService<TContentKey, TMetric, TValidator> {
     /// Metrics reporting component
     metrics: Option<OverlayMetrics>,
     /// Validator for overlay network content.
-    validator: TValidator,
+    validator: Arc<Mutex<TValidator>>,
 }
 
 impl<
-        TContentKey: OverlayContentKey + Send + Sync,
+        TContentKey: 'static + OverlayContentKey + Send + Sync,
         TMetric: Metric + Send + Sync,
         TValidator: 'static + Validator<TContentKey> + Send + Sync,
     > OverlayService<TContentKey, TMetric, TValidator>
@@ -343,7 +344,7 @@ where
         protocol: ProtocolId,
         utp_listener_sender: UnboundedSender<UtpListenerRequest>,
         enable_metrics: bool,
-        validator: TValidator,
+        validator: Arc<Mutex<TValidator>>,
         query_timeout: Duration,
         query_peer_timeout: Duration,
         query_parallelism: usize,
@@ -486,7 +487,7 @@ where
 
                         // Perform background processing.
                         match response.response {
-                            Ok(response) => self.process_response(response, active_request.destination, active_request.request, active_request.query_id).await,
+                            Ok(response) => self.process_response(response, active_request.destination, active_request.request, active_request.query_id),
                             Err(error) => self.process_request_failure(response.request_id, active_request.destination, error),
                         }
 
@@ -1108,7 +1109,7 @@ where
     }
 
     /// Processes a response to an outgoing request from some source node.
-    async fn process_response(
+    fn process_response(
         &mut self,
         response: Response,
         source: Enr,
@@ -1171,7 +1172,6 @@ where
                     }
                 };
                 self.process_content(content, source, find_content_request, query_id)
-                    .await
             }
             Response::Accept(accept) => {
                 let offer_request = match request {
@@ -1182,10 +1182,7 @@ where
                     }
                 };
 
-                if let Err(err) = self
-                    .process_accept(accept, source, offer_request.content_keys)
-                    .await
-                {
+                if let Err(err) = self.process_accept(accept, source, offer_request.content_keys) {
                     error!("Error processing ACCEPT response in overlay service: {err}")
                 }
             }
@@ -1193,7 +1190,7 @@ where
     }
 
     /// Process ACCEPT response
-    pub async fn process_accept(
+    pub fn process_accept(
         &self,
         response: Accept,
         enr: Enr,
@@ -1227,17 +1224,17 @@ where
         self.utp_listener_tx
             .send(utp_request).map_err(|err| anyhow!("Unable to send Connect request to UtpListener when processing ACCEPT message: {err}"))?;
 
-        let mut conn = rx.await?;
-        // Handle STATE packet for SYN
-        let mut buf = [0; BUF_SIZE];
-        conn.recv(&mut buf).await?;
-
         let content_items = self.provide_requested_content(&response, content_keys_offered)?;
 
         let content_payload = ContentPayloadList::new(content_items)
             .map_err(|err| anyhow!("Unable to build content payload: {err:?}"))?;
 
         tokio::spawn(async move {
+            let mut conn = rx.await.unwrap();
+            // Handle STATE packet for SYN
+            let mut buf = [0; BUF_SIZE];
+            conn.recv(&mut buf).await.unwrap();
+
             // send the content to the acceptor over a uTP stream
             if let Err(err) = conn.send_to(&content_payload.as_ssz_bytes()).await {
                 warn!("Error sending content {err}");
@@ -1297,7 +1294,7 @@ where
     }
 
     /// Processes a Content response.
-    async fn process_content(
+    fn process_content(
         &mut self,
         content: Content,
         source: Enr,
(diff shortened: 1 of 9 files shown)


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1227,17 +1224,17 @@ where
...
         let content_items = self.provide_requested_content(&response, content_keys_offered)?;
 
         let content_payload = ContentPayloadList::new(content_items)
             .map_err(|err| anyhow!("Unable to build content payload: {err:?}"))?;
 
         tokio::spawn(async move {
+            let mut conn = rx.await.unwrap();
+            // Handle STATE packet for SYN
+            let mut buf = [0; BUF_SIZE];
+            conn.recv(&mut buf).await.unwrap();
```
njgheorghita, line 1236: @ogenev Are there any consequences I'm missing from moving this login into the spawned task?

ogenev: It should be fine but I think we have to call `provide_requested_content` after we initiate the uTP handshake. If we can't establish an uTP connection with the remote node, it doesn't make sense to try to get the requested content from the db. 

So I will move those lines after `conn.recv(&mut buf).await.unwrap()` :

```rust
let content_items = self.provide_requested_content(&response, content_keys_offered)?;

let content_payload = ContentPayloadList::new(content_items)
     .map_err(|err| anyhow!("Unable to build content payload: {err:?}"))?;
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -486,7 +487,7 @@ where
 
                         // Perform background processing.
                         match response.response {
-                            Ok(response) => self.process_response(response, active_request.destination, active_request.request, active_request.query_id).await,
+                            Ok(response) => self.process_response(response, active_request.destination, active_request.request, active_request.query_id),
```
ogenev, line 490: Is it possible to put `self.process_response` in `tokio::spawn` block here and keep the `await`?. I think this way it will be more clear that the whole task is on a different thread?

njgheorghita: Not really without majorly complicating things. I spent probably too much time trying this approach. The problem is that since we have a loop continually spawning async tasks in parallel, these tasks each require a mutable reference to `self`. This requires introducing `'static' lifetime on `self`, wrapping `self` in a arc/mutex, and even then I wasn't able to satisfy the compiler when we spawn an overlay service. @jacobkaufmann pointed out that if we only spawn what is strictly necessary (which only requires cloning two `Arc<RwLock<>>`s) in each new task, then it simplifies everything.

jacobkaufmann: At first glance, it looks like you could implement `provide_requested_content` where the storage is passed as an argument. I'm not sure to what extent it's an anti-pattern to pass `Arc` around whenever you run into an issue with `self` though.

njgheorghita: I'm not sure I understand... For sure `provide_request_content` could decouple from `self` and accept an `Arc` for `storage`. Whether or not that's an anti-pattern... I don't have any strong instinct, but it seems to be the best option from what I have come across. I actually also use it in #376, to achieve the same goal (decoupling from `self`). 

But, I'm not sure I understand the benefit here. Simply decoupling `provide_requested_content` à la your suggestion wouldn't allow us to move the `tokio::spawn` block up to `self.process_response` as @ogenev requested. Assuming, I'm understanding correctly, something like this will not fly for the reasons explained above.
```rust
// line 489ish
match response.response {
    Ok(response) => {
        tokio::spawn( async move {
            self.process_response(response, active_request.destination, active_request.request, active_request.query_id).await
        });
    },
    Err(error) => self.process_request_failure(response.request_id, active_request.destination, error),
}
```
Within `process_response()` we access multiple properties on `self`, so we'd need to `arc<rwlock>>`ify all of those, which imo is overkill

Another option would be to move the `tokio::spawn()` down a level, something like ....
```rust
            // line 1170-ish
            Response::Content(content) => {
                let find_content_request = match request {
                    Request::FindContent(find_content) => find_content,
                    _ => {
                        error!("Unable to process received content: Invalid request message.");
                        return;
                    }
                };
                // need to clone some arcs here
                tokio::spawn( async move {
                    OverlayService::<TContentKey, TMetric, TValidator>::process_content(content, source, find_content_request, query_id)
                });
            }
            Response::Accept(accept) => {
                let offer_request = match request {
                    Request::Offer(offer) => offer,
                    _ => {
                        error!("Unable to process received content: Invalid request message.");
                        return;
                    }
                };
                // need to clone some arcs here
                tokio::spawn( async move {
                    OverlayService::<TContentKey, TMetric, TValidator>::process_accept(accept, source, offer_request.content_keys) {
                });
            }

```
Now, as @jacobkaufmann suggests, `process_accept()` appears fairly minimal in terms of the arc-lock-clones it would need (`self.protocol`, `self.utp_listener_tx` & `self.storage` for `self.provide_requested_content`). However, `process_content()` needs to access...
- `self.process_received_content()`
- `self.advance_find_content_query_with_enrs()`
- `self.advance_find_content_query_with_content()`
- `self.process_discovered_enrs()`

This seems prohibitive to me (in terms of introduced complexity) in terms of justifying moving the `tokio::spawn` to a slightly more readable location. And the compromise of spawning tasks at different levels for `process_accept()` and `process_content()` doesn't seem ideal, and could lead to more confusion as one who is reading the code might not recognize that there is another async task spawned at a deeper level than the other.

Idk, that's kinda what I'm thinking here, but definitely open to pushback or something I'm misunderstanding.

jacobkaufmann: @njgheorghita sorry for not clarifying. My comment was isolated w.r.t. `provide_requested_content`. I was not pushing for the call to `process_response` to move into a tokio task.

njgheorghita: I'm not sure I understand your suggestion here... the call to `self.provide_requested_content` (on line 1227) is not async, and it seems to me like there's no reason to move it into the `tokio::spawn({})` block on line 1232

jacobkaufmann: If there is a concern about the possibility of hitting the database despite a failed connection, then the suggestion would be to move that call into the tokio task after the connection is established. This would require that we refactor `provide_requested_content`.

ogenev: Yeah, spawning self with  `tokio::spawn` seems to add a lot of complexity. I'm not sure what pattern is used in Rust to minimize such complexities but my feeling is that we are probably missing something.

Anyways, we can try Jacob's suggestion above and pass the storage as an argument to `provide_requested_content` and remove the self. This way we can add it inside `tokio::spawn`.

jacobkaufmann (changes requested):
Requesting some minor changes here, mainly the use of a synchronous locking mechanism. I'm also against the CI change that removes checksum checks. I think we should resolve this independently.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -68,6 +68,7 @@ use ssz::Encode;
 use ssz_types::{BitList, VariableList};
 use thiserror::Error;
 use tokio::sync::mpsc::{self, UnboundedReceiver, UnboundedSender};
+use tokio::sync::Mutex;
```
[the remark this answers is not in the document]
njgheorghita: I might be wrong here, but re: `"asynchronous mutex would only be necessary if we need to hold a lock across await points"` ... isn't that exactly what we're doing [here](https://github.com/ethereum/trin/pull/373/files#diff-354721dfd15f4b0d0f5df418d0ac4118592bf96bda720655548e380faf9f1b37R1368)?

`I'm not sure why that method takes &mut self in the first place` - agreed, but updating the self mut ref on validator and changing the `Mutex` to a `RwLock` doesn't change that since `validator.validate_content().await` is async, and a future created by an async block is not `Send`, which is a required trait by `tokio::spawn`. lmk if I'm missing something about this situation.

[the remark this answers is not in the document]
jacobkaufmann: Ah yes, so I was wrong above on that. But on the second point, if we change the validator to require `&self` instead of `&mut self`, then we can just wrap the validator in an `Arc` (without a lock) which is `Send`.

Looking at the [`ChainHistoryValidator`](https://github.com/ethereum/trin/blob/d6d8f722cb773ac4ca51874429072d49dc363ea4/trin-history/src/validation.rs#L23), I don't see why we need a `&mut self` receiver.

jacobkaufmann (approved):

PR #374, opened 2022-07-10
[bugfix] End program using method that properly cleans up IPC file

mrferris:
### What was wrong?

In #348, the line `std::process::exit(1)` was added in order to end the program gracefully in case an initial bootnode ping fails. Turns out that this does not call the `ctrlc` handler setup in `jsonrpc/service.rs` which cleans up the IPC socket file (`/tmp/trin-jsonrpc.ipc` by default). This resulted in subsequent runs of `trin` throwing a `thread 'tokio-runtime-worker' panicked at 'Could not serve from IPC path '/tmp/trin-jsonrpc.ipc': Os { code: 48, kind: AddrInUse, message: "Address already in use" }', trin-core/src/jsonrpc/service.rs:139:13`. This wasn't ending the entire process, so it wasn't noticed until I went to use the IPC file and found that trin wasn't listening on it under the conditions described above.

### How was it fixed?

I had to give up my dreams of cleanly ending the program in a way that communicates "we meant to do that" to the user. Now an error is returned and simply `unwrap`'d in the case of failure, instead of `std::process::exit(1)`. 

Any suggestions on a more purposeful-looking solution welcome. The `human-panic` crate seems close but also seems like overkill to me. 

**Bonus:** In [discussion](https://github.com/ethereum/trin/pull/348#issuecomment-1171795384), it was agreed that ending the program only if _all_ bootnodes failed initial ping would be ideal. With the taken approach, this isn't any more work than ending if one of many bootnodes fail, so it's implemented in this PR as well.

### To-Do

- [ ] Clean up commit history


trin-state/src/network.rs
```diff
@@ -72,20 +74,32 @@ impl StateNetwork {
         // Trigger bonding with bootnodes, at both the base layer and portal overlay.
         // The overlay ping via talkreq will trigger a session at the base layer, then
         // a session on the (overlay) portal network.
-        for enr in self.overlay.discovery.discv5.table_entries_enr() {
+        // Will return error (and end execution) if zero bootnodes succeed bonding.
+        let mut successfully_bonded_bootnode = false;
+        let bootnodes = self.overlay.discovery.discv5.table_entries_enr();
+        if bootnodes.is_empty() {
+            return Ok(());
```
mrferris, line 81: I'm not sure why this needs to be `return Ok(());`, whereas in the `if` clause on line 97/98 I can just do `Ok(())`. Compiler complains with `expected (), found enum Result`.

Issue #349, opened 2022-06-23
Add varint prefix to offer/accept content items send over uTP stream


ogenev:
Implement LEB128 encode/decode on content items send over uTP stream via OFFER/ACCEPT.

DIscussion: https://github.com/ethereum/portal-network-specs/pull/148

PR #375, opened 2022-07-11
Add varint prefix to offer/accept content items send via uTP stream

ogenev:
### What was wrong?
Fixes  https://github.com/ethereum/trin/issues/349

Trin uses SSZ encoding to send multiple content items over uTP stream. 
As discussed in https://github.com/ethereum/portal-network-specs/pull/148, we agreed to implement a varint prefix.

### How was it fixed?
A variable length unsigned integer (varint) is prefixed to each content item. The varint holds the size, in bytes, of the consecutive content item.
The varint encoding used is [Unsigned LEB128](https://en.wikipedia.org/wiki/LEB128#Encode_unsigned_integer) and the maximum content size allowed for this application is limited to `uint32`.

### To-Do

- [x] Clean up commit history


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1929,6 +1930,24 @@ fn limit_enr_list_to_max_bytes(enrs: Vec<SszEnr>, max_size: usize) -> Vec<SszEnr
...
+    let mut content_payload = BytesMut::new().writer();
+
+    for content_item in content_items {
+        leb128::write::unsigned(&mut content_payload, content_item.len() as u64)
+            .map_err(|err| anyhow!("Unable to encode LEB128 varint: {err}"))?;
+        content_payload
+            .write(&content_item)
+            .map_err(|err| anyhow!("unable to write to content payload buf: {err}"))?;
+    }
+    Ok(content_payload.get_ref().clone())
```
jacobkaufmann, line 1948:
```suggestion
    Ok(content_payload.into_inner())
```

trin-core/src/portalnet/overlay.rs
```diff
@@ -627,6 +624,43 @@ fn log2_random_enrs(enrs: Vec<Enr>) -> anyhow::Result<Vec<Enr>> {
...
+    while !payload.get_ref().is_empty() {
+        // Read LEB128 index
+        let (bytes_to_read, var_int) = read_varint(&mut payload.get_ref())?;
+        let mut buf = vec![0u8; bytes_to_read];
+        payload
+            .read_exact(&mut buf)
+            .map_err(|err| anyhow!("Error reading varint index: {err}"))?;
+
+        // Read the content item
+        let mut buf = vec![0u8; var_int as usize];
```
jacobkaufmann, line 644: This cast could mask errors from some other node. I would recommend that we handle the situation where the varint is greater than `u32::MAX`. If that varint is too large, it seems very likely that the read would fail somewhere down the line, but we would lose the reason for the failure.

jacobkaufmann (changes requested):
Requesting some minor changes w.r.t. error handling. Other than that, looks good!

mrferris (approved):
Looks good, I approve pending the changes request by the other reviews.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -3045,6 +3071,25 @@ mod tests {
...
 
+    #[test]
+    fn test_encode_content_payload() {
+        let content_items: Vec<Bytes> = vec![vec![1, 1].into(), vec![2, 2, 2].into()];
+        let expected_content_payload = hex::decode("02010103020202").unwrap();
+
+        let content_payload = encode_content_payload(content_items).unwrap().to_vec();
+
+        assert_eq!(expected_content_payload, content_payload);
+    }
```
jacobkaufmann, line 3082: It's not obvious to the reader of the test that the method behaves correctly, because the hex-string is not easy to interpret. For encoding/decoding tests, I think it's best to test them together. You start with some value `X`. Then you encode `X`. Then you decode the encoded value, and the output should be `X`. If we had test vectors, then we could use those. If we colocate the encoding and decoding functions, then these tests can reside in the same module.

jacobkaufmann (changes requested):
Requesting that we relocate the encoding/decoding functions given that this is more wire-specific and that we would like to reduce size of `overlay_service.rs`.

.circleci/config.yml
```diff
@@ -43,7 +43,7 @@ jobs:
                 command: cargo build --workspace --jobs 2
             - run:
                 name: Test Trin workspace
-                command: cargo test --workspace --jobs 2
+                command: cargo test --workspace --jobs 2 -- --nocapture
```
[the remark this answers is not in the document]
ogenev: Removed it.

trin-core/src/utils/portal_wire.rs
```diff
@@ -0,0 +1,148 @@
...
+    #[test]
+    fn test_encode_decode_content_payload() {
+        let content_items: Vec<Bytes> = vec![vec![1, 1].into(), vec![2, 2, 2].into()];
+
+        let content_payload = encode_content_payload(&content_items).unwrap().to_vec();
+        let expected_content_items: Vec<Bytes> = decode_content_payload(content_payload)
+            .unwrap()
+            .into_iter()
+            .map(|content| Bytes::from(content.to_vec()))
+            .collect();
```
jacobkaufmann, line 135: nit: I think the names should be reversed here. The "expected" content items would be those that we begin with.

trin-core/src/utils/portal_wire.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        assert_eq!(expected_content_items, content_items);
+    }
+
+    #[test]
+    #[ignore] // This test is ignored for now because CI fails for unknown reason
```
jacobkaufmann, line 141: If we ignore the test, then I don't see the point of including it in the PR. Also, "fails for unknown reason" suggests that the test is incorrect, or the logic is incorrect. If this test is run on a 32-bit target, then

```rust
u32::MAX as usize + 1
```

will overflow and panic.

Let's either remove the test, or much more preferably, figure out why it does not pass and fix it.

ogenev: This test fails with the SIGKILL signal which doesn't give us any information, check [here](https://app.circleci.com/pipelines/github/ethereum/trin/1696/workflows/4ebf3eb7-17af-428e-a5fa-24b5013dfad7/jobs/3394). It is failing only in the CI linux pipeline. I can't reproduce the error locally, the `win` pipeline also passes, so it seems the test and the logic are correct.

I don't care if we are going to remove this test or not, I thought this fulfilled one of the use cases of the `#[ignore]` flag.

jacobkaufmann: I would prefer that we remove this rather than keep with `#[ignore]`. I don't think it makes sense to include a test that does not run. It's essentially dead code.

ogenev: Removed the test :+1:

trin-core/src/utils/portal_wire.rs
```diff
@@ -0,0 +1,141 @@
...
+        let content_payload = encode_content_payload(&expected_content_items)
+            .unwrap()
+            .to_vec();
+        let content_items: Vec<Bytes> = decode_content_payload(content_payload)
+            .unwrap()
+            .into_iter()
+            .map(|content| Bytes::from(content.to_vec()))
+            .collect();
+
+        assert_eq!(expected_content_items, content_items);
```
jacobkaufmann, line 139: nit: For equality assertions, I find it easier to read `assert_eq!(actual, expected)`

jacobkaufmann (approved):
LGTM. I think we will want to eventually refactor this into a `wire` module that also holds the wire spec message types, etc.

Issue #286, opened 2022-03-31
test_response_to_triple_ack local test failure


carver:
Repeatable test failure (3 of 3 attempts on a fresh checkout of master at c00e99c6b

CI appears green at the commit, so I'm guessing it's some kind of configuration issue. I just added here for tracking (and also a reminder to add a README section about running the local test).

Captured output:
```
$ TRIN_INFURA_PROJECT_ID=1 RUST_LOG=debug cargo test -p trin-core
    Finished test [unoptimized + debuginfo] target(s) in 0.17s
     Running unittests (target/debug/deps/trin_core-19ec7be1499b1d43)

running 127 tests
test cli::test::test_custom_discovery_port ... ok
✂️ ✂️ ✂️ 
test portalnet::types::content_key::test::test_distance_to_key ... ok
[2022-03-31T22:40:03Z DEBUG trin_core::portalnet::storage] Setting up RocksDB at path: "/tmp/trin.8RiPOGjQik5o/rocksdb"
[2022-03-31T22:40:03Z DEBUG trin_core::portalnet::storage] Setting up SqliteDB at path: "/tmp/trin.8RiPOGjQik5o/trin.sqlite"
[2022-03-31T22:40:03Z DEBUG discv5::discv5] Discv5 service already shutdown
[2022-03-31T22:40:03Z DEBUG discv5::discv5] Discv5 service already shutdown
test utp::stream::tests::test_response_to_triple_ack ... FAILED
test portalnet::types::content_key::test::test_should_store ... ok
[2022-03-31T22:40:03Z INFO  trin_core::portalnet::discovery] Starting discv5 with local enr encoded=enr:-IS4QJIjtY6mIYt5tnOIa89gnMsZu76Bd5Bg45gMaUKJW-RoAHq6nMg0r4G4APtTTduRlnekuWgp9QKyZKnIt78zBfsBgmlkgnY0gmlwhMCoAcCJc2VjcDI1NmsxoQJ9TpUJQOxBIiqmFdLWSXi81LLm9HrWkpwH_JTIaQuhU4N1ZHCCEJI decoded=ENR: NodeId: 0x5b61..bb5f, Socket: Some(192.168.1.192:4242)
[2022-03-31T22:40:03Z DEBUG trin_core::portalnet::storage] Setting up RocksDB at path: "/tmp/trin.8RiPOGjQik5o/rocksdb"
[...]

PR #377, opened 2022-07-12
remove discv5 routing table bucket refresh

jacobkaufmann:
### What was wrong?

`trin` performs routing table bucket refreshes at both the portal subnetwork layer and the discv5 layer. The discv5 refreshes are not necessary because the subnetwork refreshes also populate the discv5 routing table. Furthermore, the subnetwork refreshes would only establish contact with other portal nodes, as opposed to arbitrary discv5 nodes.

### How was it fixed?

Remove spawned task for discv5 routing table bucket refresh and associated code.

### To-Do

- [ ] Clean up commit history


jacobkaufmann:
The CI is failing for a flaky test, but that failure is unrelated to the removal of code here.

carver:
Roughly how often does it fail? 1/2 the time?

jacobkaufmann:
@carver yea so there are a few different reasons for failure in the CI history of this PR.

- First three failures: Error on checksum (addressed by #378)
- Next two failures: `self_peertest`
- Last failure: `utp::stream::tests::test_response_to_triple_ack`

My comment was not very clear. The last failure appears to be a flaky test, but that's not clear by looking only at this PR. Also, the more consistent failure of `self_peertest` is a concern. Given that this PR is not urgent, we can resolve the `self_peertest` failure before review/merge.

carver:
Cool, I just tried re-running the test. I was curious how many times it would take to turn green again.

carver (approved):
My first thought is that it might be sort of "bad citizen" behavior to not keep the buckets full of non-Portal nodes. If Portal nodes only track other Portal nodes at the base layer, then anyone who finds us (but isn't a Portal node) might get stuck on our little partitioned island of nodes.

I'm not really convinced that it is actually a problem though, and it is easy enough to fix if we are wrong. So 👍🏻 , but with a strong preference for tracking down a solution to the flaky triple_ack test.

jacobkaufmann:
@carver how do you feel about merging this? I haven't been able to figure out the flaky uTP test, and I would like @ogenev's help there, but for now we can remove some unrelated LOC. Should we reopen #286?

On the "bad citizen" issue, I agree that is okay to be selfish for now.

jacobkaufmann:
@carver for sure. I like to leave those in for diff audibility prior to the final squash + merge.

PR #378, opened 2022-07-13
Add --ignore-checksums flag to win-build rustup install step

njgheorghita:
### What was wrong?
Broken step in ci `win-build`, when trying to install rustup, mismatching checksums are found between the package that's installed and the published checksum. This indicates a security concern that we're actually downloading a possible compromised version, rather than the official release. I'm skeptical, given that this bug has appeared out of the blue. We've been installing `rustup.install` in the same manner and without any errors since we introduced `win-build` step. 

I've tried specifically installing rustup versions...
`choco install rustup.install --version 1.24.3` &
`choco install rustup.install --version 1.22.1` and they both result in a hash mismatch. I'm guessing there's a bug somewhere in how the hash is generated. 

A common solution to this problem that I've come across implies that it's an error on the distributor's end, and that they forgot to update the published checksum. The solution is to reach out to the distributor and ask them to update the published checksum. I'm skeptical that this is the cause for our error, since we have been installing `v1.24.3` without a problem until recently. This version has been [stable](https://community.chocolatey.org/packages/rustup.install/#versionhistory) for about a year. 

Another solution is to pass in `--checksum ... --checksum-64 ...` with the invalid hashes rather than use `--ignore-checksums`. But, I'm not convinced this is any better, since it'll require passing in invalid checksums, compared to those published in the official release.
```
    checksum       = '33DDB81079F94150E162FA00A84A00FF4649419FCE5AA54116CB4147785732D1'
    checksumType   = 'sha256' 
    checksum64     = '780B445EE18BDC5A31AD6A4A63943927E5C0BBA8BFF664E8ECC8D5BCA3906C7B'
    checksumType64 = 'sha256' 
````

### How was it fixed?
Added `--ignore-checksums` flag to ci step. This is definitely `"A less secure option"`. It's not an ideal solution. But it's simple and works. This error currently blocks every single pr, since ci can't run. imo getting a fix through is fairly urgent, and not worth the time to figure out what's going wrong. If there are concerns about this, I could create an issue to track this and make sure it doesn't get dropped.

(on another note.... I'd also be willing to talk about remove the `win-build` entirely/temporarily from our CI. Or run it only in certain situations, like a pre-release check. [35 minutes is somewhat absurd for a ci process](https://app.circleci.com/pipelines/github/ethereum/trin/1625/workflows/d6f4f504-88e0-4cfd-b5ce-32bc63c57a51/jobs/3233))

### To-Do

- [x] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -67,7 +67,8 @@ jobs:
             - cargo-{{ checksum "Cargo.lock" }}-v1
       - run:
           name: Install rustup
-          command: choco install rustup.install
+          # todo: Remove --ignore-checksums flag
+          command: choco install rustup.install --ignore-checksums
       - run:
           name: Install clang
           command: choco install llvm


ogenev (approved):
I'm also in favor of removing the `win-build` pipeline, I think we currently don't support Windows anyways and I don't see any benefits of having such a CI pipeline at this point of development.

mrferris (approved):
I spent some time looking for a reason this suddenly started happening and I'm not seeing anything either. In the meantime I think `--ignore-checksums` is fine, if trin were being widely used and was worth somebody compromising our CI (let alone the windows version) obviously we'd all think differently. 

And 👍 for an issue for discussion of windows CI, as well as maybe for support in general.

PR #381, opened 2022-07-18
History network types for accumulator content

njgheorghita:
### What was wrong?
Added basic content key types for `EpochAccumulator` and `MasterAccumulator` content in the history network, as defined [here](https://github.com/ethereum/portal-network-specs/pull/155/files).

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/types/content_key.rs
+++ b/trin-core/src/portalnet/types/content_key.rs
@@ -1,5 +1,5 @@
 use anyhow::anyhow;
-use ethereum_types::{U256, U512};
+use ethereum_types::{H256, U256, U512};
 use sha2::{Digest as Sha2Digest, Sha256};
 use sha3::{Digest, Keccak256};
 use ssz::{self, Decode, Encode};
@@ -66,7 +66,7 @@ impl OverlayContentKey for IdentityContentKey {
 //
 
 /// A content key in the history overlay network.
-#[derive(Clone, Debug, Decode, Encode)]
+#[derive(Clone, Debug, Decode, Encode, PartialEq)]
 #[ssz(enum_behaviour = "union")]
 pub enum HistoryContentKey {
     /// A block header.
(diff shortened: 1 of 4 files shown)


trin-core/src/portalnet/types/content_key.rs
```diff
@@ -96,14 +100,67 @@ pub struct BlockBody {
...
+#[ssz(enum_behaviour = "union")]
+pub enum MasterAccumulator {
+    Latest(SszNone),
+    MasterHash(H256),
+}
+
+/// Struct to represent encodable/decodable None value for an SSZ enum
+#[derive(Clone, Debug, PartialEq)]
+pub struct SszNone {
+    value: Option<()>,
```
carver, line 122: What is a an `SszNone { value: None }` vs an `SszNone { value: Some(()) }`? Probably deserves a comment.

trin-history/src/validation.rs
```diff
@@ -106,6 +107,10 @@ impl Validator<HistoryContentKey> for ChainHistoryValidator {
                 }
                 Ok(())
             }
+            _ => {
```
carver, line 110: A catch-all here seems likely to bite us later. It looks like it would be an improvement to specifically include the accumulators, so that we notice when a new un-handled enum type is added.

trin-core/src/portalnet/types/content_key.rs
```diff
@@ -658,4 +715,73 @@ mod test {
...
+            &hex::decode("9fb2175e76c6989e0fdac3ee10c40d2a81eb176af32e1c16193e3904fe56896e")
+                .unwrap();
+
+        let content_key = HistoryContentKey::EpochAccumulator(EpochAccumulator {
+            epoch_hash: H256::from_slice(&epoch_hash),
+        });
+        assert_eq!(&content_key.content_id().to_vec(), expected_content_id);
+
+        let actual_content_key: Vec<u8> = content_key.clone().into();
+        assert_eq!(actual_content_key, expected_content_key);
```
carver, line 738: I found `content_key` vs `actual_content_key` didn't help me understand what was happening. This would have helped me more:
```suggestion
        let encoded_content_key: Vec<u8> = content_key.clone().into();
        assert_eq!(encoded_content_key, expected_content_key_encoding);
```

(Same in the other 2 tests)

carver (approved):

PR #382, opened 2022-07-20
Remove blocking async from overlay events loop

njgheorghita:
### What was wrong?
Similar to #373 We had blocking async inside of the overlay events processing loop.

### How was it fixed?
Spawned new non-blocking tokio task to process an event when it's received.

### To-Do

- [x] Clean up commit history


mrferris (approved):
LGTM! 🎖

jacobkaufmann (approved):

PR #383, opened 2022-07-20
Only send content keys into propagate_gossip()

carver:
### What was wrong?

It was confusing/misleading that the `propagate_gossip` method accepted content values but didn't use them (instead it relies on the content being available in the database).

The only reason we might not want to do this is if we want to thread the content values through here, instead of depending on them being available in the database.

### How was it fixed?

Only pass in the keys instead. Updated the only callsite to only pass in keys.

### To-Do

- [ ] Clean up commit history


carver:
Going the direction of #388 which requires keeping the values in the argument...

PR #384, opened 2022-07-22
Propagate after rpc history store

carver:
### What was wrong?

We need to propogate data to peers after calling `portal_historyStore`, in order for the bridge node to work.

### How was it fixed?

Add a propagate, to the end of the rpc handling for `portal_historyStore`

Bonus: Includes #383 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -255,8 +255,8 @@ where
 
         // TODO: Verify overlay content data with an Oracle
 
-        // Temporarily store content key/value pairs to propagate here
-        let mut content_keys_values: Vec<(TContentKey, ByteList)> = Vec::new();
+        // Track which content keys were successfully stored, for propagation
+        let mut stored_content_keys: Vec<TContentKey> = Vec::new();
 
         // Try to store the content into the database and propagate gossip the content
         for (content_key, content_value) in content_keys.into_iter().zip(content_values.to_vec()) {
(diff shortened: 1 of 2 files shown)


ogenev:
I'm not sure if I like coupling `portal_historyStore` with `propagate_gossip`.

For example, when I use `portal_historyStore` in testing, I expect this method to just try to store some data into DB, I don't expect it to have any other side effects like propagating content.

I think keeping all endpoints' functionalities as decoupled as possible is beneficial for writing more complex test scenarios where multiple endpoints and assertions are involved.

carver:
Yeah, this objection is similar to a direction I was talking about in #368 , with `portal_historyOfferValue`. I also started a conversation in discord, to see if we could reuse `portal_historyOffer` and `portal_historySendOffer`, rather than make a new endpoint that's not defined yet.

carver:
> I would prefer to add an additional endpoint for propagating the content or an endpoint for storing **and** propagating (like the implementation here).

Note that it would currently be very awkward to implement an endpoint that doesn't store, but does propagate, because the data is served from the database, when Accept is received. I'm probably going to look at that as the next alternative: to include the content values in the internal Offer request, so they can be used to respond without looking up the content in the DB.

carver:
> Note that it would currently be very awkward to implement an endpoint that doesn't store, but does propagate, because the data is served from the database, when Accept is received. I'm probably going to look at that as the next alternative: to include the content values in the internal Offer request, so they can be used to respond without looking up the content in the DB.

Merged #388 which now allows you to propagate without storing to disk

PR #385, opened 2022-07-22
Update accumulator

njgheorghita:
### What was wrong?
Implement `master_accumulator.update()` as defined in the [specs](https://github.com/ethereum/portal-network-specs/pull/155/files#diff-9564247101afbd35379ccdd1fdb4b21b1f4529467c0724234c6bf7d2f3120106R93).

### How was it fixed?
- Add support for updating the master accumulator
- Tests are based off [fluffy tests](https://github.com/status-im/nimbus-eth1/blob/master/fluffy/tests/test_accumulator.nim)
- Updated some invalid rlp values used in `Header` tests
- Updated `Header::from_get_block_jsonrpc_response()` to generate a valid header

### To-Do

- [x] Clean up commit history


trin-core/src/types/header.rs
```diff
@@ -61,14 +57,14 @@ where
...
-    }
-
-    /// Returns the RLP representation of the header.
-    pub fn rlp(&self, with_seal: bool) -> Bytes {
-        let mut s = RlpStream::new();
-        self.stream_rlp(&mut s, with_seal);
-        s.out().freeze()
+        match self.number {
+            // return hard coded mainnet genesis block hash
+            0 => H256::from_slice(
```
njgheorghita, line 62: @carver Am I missing something obvious here? I cannot for the life of me get the genesis block hash to match. I've seen it hardcoded in the geth codebase... Is it a hardcoded value? Is its header encoded in any special way as opposed to a regular block?

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,214 @@
...
+            },
+        }
+    }
+
+    // as defined in: https://github.com/ethereum/portal-network-specs/blob/master/history-network.md
+    fn update_accumulator(&mut self, new_block_header: &Header) {
+        // get the previous total difficulty
+        let last_total_difficulty = match self.current_epoch.header_records.len() {
+            // genesis
+            0 => U256::from_dec_str("0").unwrap(),
```
jacobkaufmann, line 36:
```suggestion
            0 => U256::zero(),
```

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,214 @@
+use ethereum_types::U256;
+use ssz::Encode;
+use ssz_derive::{Decode, Encode};
+use tree_hash::{MerkleHasher, TreeHash};
+use tree_hash_derive::TreeHash;
+
+use crate::types::header::Header;
+
+const EPOCH_SIZE: u32 = 8192; // number of blocks / epoch
+const MAX_HISTORICAL_EPOCHS: u32 = 131072; // 2 ** 17
```
jacobkaufmann, line 10:
```suggestion
const EPOCH_SIZE: usize = 8192; // number of blocks / epoch
const MAX_HISTORICAL_EPOCHS: usize = 131072; // 2 ** 17
```

With this change, we can remove casts below.

If we go with a `VariableList` (suggested below), then we may need to use [`typenum::U8192`](https://docs.rs/typenum/1.15.0/typenum/type.U8192.html) and [`typenum::U131072`](https://docs.rs/typenum/1.15.0/typenum/type.U131072.html).

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,214 @@
...
+        tree_hash::TreeHashType::List
+    }
+
+    fn tree_hash_packed_encoding(&self) -> Vec<u8> {
+        unreachable!("List should never be packed.")
+    }
+
+    fn tree_hash_packing_factor() -> usize {
+        unreachable!("List should never be packed.")
+    }
```
jacobkaufmann, line 125: Worth a comment to explain why each method is `unreachable!`.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,214 @@
...
+        self.current_epoch.header_records.push(header_record);
+    }
+}
+
+// ssz list
+// max_length = MAX_HISTORICAL_EPOCHS
+#[derive(Clone, Debug, Decode, Encode)]
+pub struct HistoricalEpochs {
+    epochs: Vec<tree_hash::Hash256>,
+}
```
jacobkaufmann, line 72: Can we use [`VariableList`](https://docs.rs/eth2_ssz_types/0.2.1/ssz_types/struct.VariableList.html) to enforce max length?

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,214 @@
...
+        tree_hash::mix_in_length(&historical_epochs_root, hash_tree_roots.len())
+    }
+}
+
+// ssz list
+// max_length = EPOCH_SIZE
+#[derive(Clone, Debug, Decode, Encode)]
+pub struct EpochAccumulator {
+    header_records: Vec<HeaderRecord>,
+}
```
jacobkaufmann, line 112: Same as above. Can we use [`VariableList`](https://docs.rs/eth2_ssz_types/0.2.1/ssz_types/struct.VariableList.html) to enforce max length?

jacobkaufmann (changes requested):
Looks good! Requesting some minor changes.

trin-core/src/types/header.rs
carver: It's not hardcoded in py-evm AFAICT. Seems like it's calculated normally from:
https://github.com/ethereum/py-evm/blob/7674e2e0caefa8b1208f650bb71a07c65ef2f305/eth/chains/mainnet/__init__.py#L140-L155

trin-core/src/types/header.rs
```diff
@@ -194,6 +188,12 @@ fn try_value_into_u64(val: &Value) -> anyhow::Result<u64> {
 impl Decodable for Header {
     /// Attempt to decode a header from RLP bytes.
     fn decode(rlp: &Rlp) -> Result<Self, DecoderError> {
+        // decode raw big-endian encoded bytes for nonce
+        let mut nonce: [u8; 8] = Default::default();
+        let raw_nonce: Vec<u8> = rlp.val_at(14)?;
+        nonce.copy_from_slice(&raw_nonce[..]);
+        let nonce = u64::from_be_bytes(nonce);
```
carver, line 195: Ah, was this the problem with the genesis header decoding? That the actual RLP header definition expects the nonce to be treated as bytes, but the Rust implementation is treating it as a uint?

Maybe we should just change the Rust `Header` to treat the nonce as bytes internally?

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    fn tree_hash_root(&self) -> tree_hash::Hash256 {
+        // list of composite objects
+        // mix_in_length(merkleize([hash_tree_root(element) for element in value], limit=chunk_count(type)), len(value))
+        let hash_tree_roots: Vec<tree_hash::Hash256> = self
+            .clone()
+            .epochs
+            .into_iter()
+            .map(|epoch_root| epoch_root.tree_hash_root())
```
carver, line 110: Something feels a little funny here about cloning self instead of the epochs. Actually, can the clone be skipped entirely like this?
```suggestion
        let hash_tree_roots: Vec<tree_hash::Hash256> = self
            .epochs
            .iter()
            .map(|epoch_root| epoch_root.tree_hash_root())
```

Ditto on `EpochAccumulator`

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+            .collect();
+        let mut historical_epochs_hasher = MerkleHasher::with_leaves(MAX_HISTORICAL_EPOCHS);
+        for root in &hash_tree_roots {
+            let hash = tree_hash::merkle_root(root.as_bytes(), 0);
+            historical_epochs_hasher.write(hash.as_bytes()).unwrap();
```
carver, line 115: I thought you decided this extra line was unnecessary in testing. Also, maybe a little silly, but nice to read why the unwrap might fail.
```suggestion
            historical_epochs_hasher.write(root.as_bytes()).expect("Invalid accumulator state: too many epochs");
```

Ditto on `EpochAccumulator`

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+        let hash_tree_roots = vec![
+            hex::decode("b629833240bb2f5eabfb5245be63d730ca4ed30d6a418340ca476e7c1f1d98c0")
+                .unwrap(),
+            hex::decode("00cbebed829e1babb93f2300bebe7905a98cb86993c7fc09bb5b04626fd91ae5")
+                .unwrap(),
+            hex::decode("88cce8439ebc0c1d007177ffb6831c15c07b4361984cc52235b6fd728434f0c7")
+                .unwrap(),
+        ];
+        let mut master_accumulator = MasterAccumulator::new();
+        for (index, header) in headers.iter().enumerate() {
```
carver, line 195: *nit* maybe the more natural representation is to `zip(hash_tree_roots)` with `headers`, instead of `enumerate`.

trin-core/src/types/header.rs
```diff
@@ -100,8 +90,8 @@ impl Header {
             .append(&self.extra_data);
 
         if with_seal && self.mix_hash.is_some() && self.nonce.is_some() {
-            s.append(&self.mix_hash.unwrap())
-                .append(&self.nonce.unwrap());
+            let nonce = Bytes::copy_from_slice(&self.nonce.unwrap().to_be_bytes());
```
carver, line 93: Referring to my comment from yesterday: this is another place that suggests to me that we should internally represent the nonce as bytes instead of a u64.

trin-core/src/types/header.rs
```diff
@@ -231,26 +255,44 @@ mod tests {
...
     fn decode_and_encode_header() {
-        let header_rlp = hex::decode("f901f9a0d405da4e66f1445d455195229624e133f5baafe72b5cf7b3c36c12c8146e98b7a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347948888f1f195afa192cfee860698584c030f4c9db1a05fb2b4bfdef7b314451cb138a534d225c922fc0e5fbe25e451142732c3e25c25a088d2ec6b9860aae1a2c3b299f72b6a5d70d7f7ba4722c78f2c49ba96273c2158a007c6fdfa8eea7e86b81f5b0fc0f78f90cc19f4aa60d323151e0cac660199e9a1b90100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000008302008003832fefba82524d84568e932a80a0a0349d8c3df71f1a48a9df7d03fd5f14aeee7d91332c009ecaff0a71ead405bd88ab4e252a7e8c2a23").unwrap();
+        // Mainnet block #1 rlp encoded header
+        // sourced from mainnetMM data dump
```
carver, line 261: `mainnetMM`?

trin-history/src/validation.rs
```diff
@@ -188,7 +188,8 @@ mod tests {
                         "gasLimit": "0x1c9c364",
                         "gasUsed": "0x140db1",
                         "hash": "0x720704f3aa11c53cf344ea069db95cecb81ad7453c8f276b2a1062979611f09c",
-                        "logsBloom": "", // logs removed since they're unnecessary for validation
+                        // Incorrect logs value used since they're unnecessary for validation
```
carver, line 191: This wasn't clear to me on first read. Maybe:
```suggestion
                        // Using an empty bloom that doesn't match the real logs, because it is easy and isn't validated
```
Is that what you mean?

carver (changes requested):
If changing the `nonce` data type makes sense to you, then 👍🏻 to ship.

Otherwise let's talk it out. Flagging as request changes to highlight this ^.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+use ssz_types::{typenum, VariableList};
+use tree_hash::{MerkleHasher, TreeHash};
+use tree_hash_derive::TreeHash;
+
+use crate::types::header::Header;
+
+const EPOCH_SIZE: usize = 8192; // number of blocks / epoch
+const MAX_HISTORICAL_EPOCHS: usize = 131072; // 2 ** 17
+
+// ssz list[Hash256, max_length = MAX_HISTORICAL_EPOCHS]
```
ogenev, line 13: Let's make those doc comments with `///` and probably add some short descriptions.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+pub type HistoricalEpochsList = VariableList<tree_hash::Hash256, typenum::U8192>;
+
+// ssz list[HeaderRecord, max_length = EPOCH_SIZE]
+pub type HeaderRecordList = VariableList<HeaderRecord, typenum::U131072>;
+
+// ssz container
```
ogenev, line 19: This is not a doc comment and doesn't give us any information on what `MasterAccumulator` is and what it is doing.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+            historical_epochs: HistoricalEpochs {
+                epochs: HistoricalEpochsList::empty(),
+            },
+            current_epoch: EpochAccumulator {
+                header_records: HeaderRecordList::empty(),
+            },
+        }
+    }
+
+    // as defined in:
```
ogenev, line 40: nit: doc comment with a short description would be nice.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+        let mut current_epoch_hasher = MerkleHasher::with_leaves(EPOCH_SIZE);
+        for root in &hash_tree_roots {
+            current_epoch_hasher.write(root.as_bytes()).unwrap();
+        }
+        let current_epoch_root = current_epoch_hasher.finish().unwrap();
+        tree_hash::mix_in_length(&current_epoch_root, hash_tree_roots.len())
+    }
+}
+
+// Individual record for a historical header.
```
ogenev, line 161:
```suggestion
/// Individual record for a historical header.
```

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+// Every HeaderRecord is 64bytes.
```
ogenev, line 162:
```suggestion
/// Every HeaderRecord is 64bytes.
```

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+            let hash = tree_hash::merkle_root(root.as_bytes(), 0);
+            historical_epochs_hasher.write(hash.as_bytes()).unwrap();
+        }
+        let historical_epochs_root = historical_epochs_hasher.finish().unwrap();
+        tree_hash::mix_in_length(&historical_epochs_root, hash_tree_roots.len())
+    }
+}
+
+#[derive(Clone, Debug, Decode, Encode)]
+pub struct EpochAccumulator {
```
ogenev, line 123: nit: doc comment with a short description would be nice.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+        };
+        self.current_epoch
+            .header_records
+            .push(header_record)
+            .expect("Invalid accumulator state, more current epochs than allowed.");
+    }
+}
+
+#[derive(Clone, Debug, Decode, Encode)]
+pub struct HistoricalEpochs {
```
ogenev, line 84: nit: doc comment with a short description would be nice.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+        let current_epoch_root = current_epoch_hasher.finish().unwrap();
+        tree_hash::mix_in_length(&current_epoch_root, hash_tree_roots.len())
+    }
+}
+
+// Individual record for a historical header.
+// Every HeaderRecord is 64bytes.
+#[derive(Debug, Decode, Encode, Clone, Copy, TreeHash)]
+pub struct HeaderRecord {
```
ogenev, line 164: nit: doc comment with a short description would be nice.

trin-core/src/types/accumulator.rs
njgheorghita: This one I slightly disagree. I find it somewhat useful as I'm putting together the accumulator logic to remember that this object is an ssz-encoded container, not a list.

njgheorghita:
@ogenev Totally fair, I tend to slack on commenting my code. Feel free to request more comments anytime you think it's insufficient. I'm curious as to docstrings vs. regular comments. Since trin's not a library, I was operating under the assumptions that the audience for comments are going to be mainly developers working on the code, rather than users. Which is why I tend to use regular comments rather than docstrings. But, I'm happy to switch it up to docstrings if that's preferred.

trin-core/src/types/header.rs
njgheorghita: Andrew's data dump for the first 10000 blocks, added a link to the source.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,203 @@
...
+
+// ssz container
+#[derive(Clone, Debug, Decode, Encode, TreeHash)]
+pub struct MasterAccumulator {
+    historical_epochs: HistoricalEpochs,
+    current_epoch: EpochAccumulator,
+}
+
+// todo: remove dead-code exception
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
njgheorghita: I expect it to get removed pretty quickly, as soon as we connect the accumulator to the header oracle and we're actually maintaining a master accumulator. The warning is for the functions, which are only used in the tests for now. I updated the comment

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,219 @@
...
+/// List of historical epoch accumulator merkle roots preceding current epoch.
+pub type HistoricalEpochsList = VariableList<tree_hash::Hash256, typenum::U8192>;
+
+/// SSZ List[HeaderRecord, max_length = EPOCH_SIZE]
+/// List of (block_number, block_hash) for each header in the current epoch.
+pub type HeaderRecordList = VariableList<HeaderRecord, typenum::U131072>;
+
+/// SSZ Container
+#[derive(Clone, Debug, Decode, Encode, TreeHash)]
+pub struct MasterAccumulator {
```
jacobkaufmann, line 23: nit: "SSZ Container" does not describe what the `MasterAccumulator` is. I would prefer something more descriptive.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,219 @@
...
+const EPOCH_SIZE: usize = 8192; // number of blocks / epoch
+const MAX_HISTORICAL_EPOCHS: usize = 131072; // 2 ** 17
+
+/// SSZ List[Hash256, max_length = MAX_HISTORICAL_EPOCHS]
+/// List of historical epoch accumulator merkle roots preceding current epoch.
+pub type HistoricalEpochsList = VariableList<tree_hash::Hash256, typenum::U8192>;
+
+/// SSZ List[HeaderRecord, max_length = EPOCH_SIZE]
+/// List of (block_number, block_hash) for each header in the current epoch.
+pub type HeaderRecordList = VariableList<HeaderRecord, typenum::U131072>;
```
jacobkaufmann, line 19: nit: It seems like these types can be private.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,219 @@
...
+
+        let block_0_header: Header = rlp::decode(&block_0_rlp).unwrap();
+        let block_1_header: Header = rlp::decode(&block_1_rlp).unwrap();
+        let block_2_header: Header = rlp::decode(&block_2_rlp).unwrap();
+        let headers = vec![block_0_header, block_1_header, block_2_header];
+
+        // test values sourced from: https://github.com/status-im/nimbus-eth1/blob/master/fluffy/tests/test_accumulator.nim
```
jacobkaufmann, line 198: nit: use permalink.

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    fn tree_hash_packing_factor() -> usize {
+        // in ssz merkleization spec, only basic types are packed
+        unreachable!("List should never be packed.")
+    }
+
+    fn tree_hash_root(&self) -> tree_hash::Hash256 {
+        // list of composite objects
+        // mix_in_length(merkleize([hash_tree_root(element) for element in value], limit=chunk_count(type)), len(value))
```
jacobkaufmann, line 154: Is this comment leftover from some previous implementation?

trin-core/src/types/accumulator.rs
```diff
@@ -0,0 +1,219 @@
...
```
jacobkaufmann, line 113: Same comment as below. This comment is more confusing than helpful to me.

jacobkaufmann (commented):
Sorry I didn't notice this on the first go around, but `TreeHash` is already implemented for [`VariableList`](https://docs.rs/eth2_ssz_types/0.2.1/ssz_types/struct.VariableList.html#impl-TreeHash), so is there a reason for the wrapped types and new implementations? It seems like we can remove a lot of the code here.

njgheorghita:
> Sorry I didn't notice this on the first go around, but TreeHash is already implemented for [VariableList](https://docs.rs/eth2_ssz_types/0.2.1/ssz_types/struct.VariableList.html#impl-TreeHash), so is there a reason for the wrapped types and new implementations? It seems like we can remove a lot of the code here.

That would be nice, but for some reason, automatically deriving the `TreeHash` trait on those types (whether we use `HistoricalEpochs` & `EpochAccumulator` or just `HistoricalEpochsList` & `HeaderRecordList` directly gives us the wrong tree-hash-root value on the master accumulator (at least compared to the values sourced from fluffy).

ogenev:
> Since trin's not a library, I was operating under the assumption that the audience for comments is going to be mainly developers working on the code, rather than users. This is why I tend to use regular comments rather than docstrings. But, I'm happy to switch it up to docstrings if that's preferred.

I think using doc comments even for binaries is a good practice, the only downside I see is adding additional `/` but it is more readable and serves as documentation for the developers. I would expect any significant type to have a doc comment with a short description.

For reference, Rust's [Cargo](https://github.com/rust-lang/cargo/blob/master/src/bin/cargo/main.rs), [Lighthouse](https://github.com/sigp/lighthouse) and [Alacritty](https://github.com/alacritty/alacritty)use doc comments.

trin-core/src/types/accumulator.rs
jacobkaufmann: Can we remove?

trin-core/src/types/accumulator.rs
jacobkaufmann: Can we remove?

njgheorghita: Same reply as above.

trin-core/src/types/accumulator.rs
jacobkaufmann: I agree that it's useful (and desired) to describe the logic, but I don't think that this comment accomplishes that. If we keep the comment, then can we translate the pseudocode into more of a description?

carver: It's not pseudocode, it's legit python. I agree that it would be nice to add more context for folks who haven't been hanging out in py-ssz, though. Something like:
```rs
// When generating the hash root of a list of composite objects, the complete procedure is here
// https://github.com/ethereum/py-ssz/blob/36f3406f814a5e5f4efb059a6928afc2d9d253b4/ssz/sedes/list.py#L113-L129
// Since we know each element is a composite object, we can simplify the python by removing some branches, to:
// mix_in_length(merkleize([hash_tree_root(element) for element in value], limit=chunk_count(type)), len(value))
```

PR #387, opened 2022-07-27
refactor: move Discv5Event handling into Discovery

jacobkaufmann:
### What was wrong?

Predecessor of #386. The upcoming ENR cache should be managed centrally within `Discovery`.

Most `Discv5Event` variants are irrelevant to the various portal subnetworks. In fact, the only relevant variant is the `TalkRequest` variant. The result is that we have unnecessary matching logic on the `Discv5Event` receiver.

The other `Discv5Event` variants can be handled centrally, so that any relevant information is available through `Discovery`, which we already pass around inside an `Arc`. Otherwise, we would have to pass around `Arc<RwLock<..>>` to whatever structs we would need to update based on those other events.

### How was it fixed?

- Spawn tokio task to manage the `Discv5Event` stream and dispatch `TalkRequest` through a `mpsc::Sender`.
- Return `mpsc::Receiver<TalkRequest>` from `Discovery::start` to listen for incoming talk requests.

Bonuses:

- Make `discv5` a private field of `Discovery` and add `pub` methods where necessary.
- Move logs from `Discovery::new` to `Discovery::start`.
  - It doesn't make sense to me to emit logs in a call to `new`.

### To-Do

- [x] Clean up commit history


ogenev (approved):
Nice, this is a great improvement :100:!

trin-core/src/portalnet/discovery.rs
```diff
@@ -105,19 +107,13 @@ impl Discovery {
...
-        info!(
-            "Starting discv5 with local enr encoded={:?} decoded={}",
-            enr, enr
-        );
-
         let discv5 = Discv5::new(enr, enr_key, config.discv5_config)
             .map_err(|e| format!("Failed to create discv5 instance: {}", e))?;
 
         for enr in config.bootnode_enrs {
-            info!("Adding bootnode {}", enr);
```
carver, line 117: There's probably some value in logging the bootnode IP addresses at an info level, which seems to get lost in this change.

trin-core/src/portalnet/discovery.rs
```diff
@@ -127,22 +123,47 @@ impl Discovery {
...
+
+        let mut event_rx = self.discv5.event_stream().await.unwrap();
+
+        let (talk_req_tx, talk_req_rx) = mpsc::channel(TALKREQ_CHANNEL_BUFFER);
+
+        tokio::spawn(async move {
+            while let Some(event) = event_rx.recv().await {
+                // Forward all TALKREQ messages.
+                if let Discv5Event::TalkRequest(talk_req) = event {
+                    let _ = talk_req_tx.send(talk_req).await;
```
carver, line 148: Do we even have to `await` here, since we're ignoring the result? Seems like we could do a more fire-and-forget type option, which might improve performance in some scenarios.

carver (approved):
Yup, looks good!

trin-core/src/portalnet/discovery.rs
carver: Nevermind, I was thinking in python again.

> This method will never block the current thread.
https://doc.rust-lang.org/std/sync/mpsc/struct.Sender.html#method.send

trin-core/src/portalnet/discovery.rs
jacobkaufmann: I think it's okay to keep this removed given that both the [history](https://github.com/ethereum/trin/blob/765d75b0ecc7986e678b2f260477aac1afd39369/trin-history/src/lib.rs#L85) and [state](https://github.com/ethereum/trin/blob/765d75b0ecc7986e678b2f260477aac1afd39369/trin-state/src/lib.rs#L79) network initializations log the boot nodes.

carver: Yup, my issue was specifically that those logs don't include the IP addresses.

jacobkaufmann: Oh whoops I totally overlooked that. Created #430.

PR #388, opened 2022-07-28
Serve content from memory

carver:
### What was wrong?

May not always want to offer content from database, sometimes want to offer it from memory. See #368 

LRU cache is a hacky solution, this is ... a little less hacky.

### How was it fixed?

Added a new `PopulatedOffer` which is just like an `Offer` but with the data included.

This is a special-cased out-bound only offer that allows us to skip all the database lookups. Even more importantly, we can offer data that *never* gets stored locally, like in the bridge nodes.

There are a couple of commits added at the end, related to the peertest failure in CI. The commits are pretty tidy, for reviewing one at a time.

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [ ] Clean up commit history
- [x] Add commit that reverts LRU cache

PR #390, opened 2022-07-28
We don't need to exit during panic handling

carver:
... I think

The panic hook only needs to display information we want, and delete the
IPC file. The Rust machinery will handle the rest on its own.

But, it does seem to cause the json-rpc server to hang on panic. Which I don't understand yet...

### To-Do

- [ ] Clean up commit history


carver:
Merged in #388

PR #392, opened 2022-07-30
Print errors without exiting on bootnode failures

mrferris:
### What was wrong?

In #348, the line `std::process::exit(1)` was added in order to end the program "gracefully" in case initial bootnode pings fail. Turns out this does not gracefully clean things up at all, and was leaving an orphaned IPC file.

Additionally, #348 had replaced the panic that occurred if a bootnode failed ping with an error printout, which made the `network.ping_bootnodes()` method returning a `Result` pointless because it always returned `Ok(())`. 

### How was it fixed?

This PR makes it so that specific errors are printed with no panic or exit for the cases in which:
- a single bootnode fails ping
- all bootnodes fail ping
- no bootnodes were provided

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -639,6 +639,44 @@ where
             Err(error) => Err(OverlayRequestError::ChannelFailure(error.to_string())),
         }
     }
+
+    pub async fn ping_bootnodes(&self) {
+        // Trigger bonding with bootnodes, at both the base layer and portal overlay.
+        // The overlay ping via talkreq will trigger a session at the base layer, then
+        // a session on the (overlay) portal network.
+        let mut successfully_bonded_bootnode = false;
+        let enrs = self.discovery.discv5.table_entries_enr();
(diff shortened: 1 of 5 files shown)


carver (approved):
LGTM. Obviously it would be nice if this `ping_bootnodes` logic weren't duplicated. It would have been very easy to miss in this change. If there's a straightforward way to refactor `ping_bootnodes` into a core, generic function that takes `overlay` as an argument, that would be cool. I trust you to decide if it's worth it after trying it out for a few minutes.

mrferris:
@carver Yea, been thinking I'd addressing that redundancy (and the redundancies in `events.rs` if possible) in a subsequent PR, but I agree this one's a low enough hanging fruit to address here 👍

carver:
> Yea, been thinking I'd addressing that redundancy (and the redundancies in `events.rs` if possible) in a subsequent PR

Totally understand, and the approach you took made it easy for me to see what changed in a commit, separate from the refactor. So win-win! 👍🏻

PR #395, opened 2022-08-02
utp: Fix packet's `wnd_size` value and handle bidirectional FIN packets

ogenev:
### What was wrong?
During interop testing with Fluffy, the following issues were found:

- `wnd_size` value in the uTP packet header was set to 0 when sending SYN, DATA, and FIN packets.
-  we don't check for active uTP connection matching remote packet `connection_id` when we receive the FIN packet on the same socket where we are the senders of the DATA packet.

### How was it fixed?
- Set `wnd_size` to max window size for `SYN` packets and `max_window_size - curr_window` for DATA and FIN packets.
- Handle FIN packets when we are senders of the data (bidirectional FIN).

### To-Do
- [x] Clean up commit history

### DIFF
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -273,7 +273,9 @@ impl UtpListener {
                                 };
                             }
                         } else {
-                            warn!("Received SYN packet for an unknown active uTP stream");
+                            warn!(
+                                "Received SYN packet for an unknown active uTP stream: {packet:?}"
+                            );
                         }
                     }
                     // Receive DATA and FIN packets
@@ -300,7 +302,9 @@ impl UtpListener {
                                 Err(err) => error!("Unable to receive uTP DATA packet: {err}"),
                             }
                         } else {
-                            warn!("Received DATA packet for an unknown active uTP stream")
+                            warn!(
+                                "Received DATA packet for an unknown active uTP stream: {packet:?}"
+                            )
                         }
                     }
                     PacketType::Fin => {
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
🚢

trin-core/src/utp/stream.rs
```diff
@@ -606,6 +623,7 @@ impl UtpStream {
             let mut packet = Packet::with_payload(chunk);
             packet.set_seq_nr(self.seq_nr);
             packet.set_ack_nr(self.ack_nr);
+            packet.set_wnd_size(WINDOW_SIZE.wrapping_sub(self.cur_window));
```
carver, line 626: When we allow an underflow like this, it seems worth at least a short comment to explain why we're not worrying about it. Something like:

> If `WND_SIZE` underflows to end up higher than we intended, then we would be advertising for the peer to send us more data than we would theoretically prefer. That's not a fatal situation, so we don't protect against it or warn.

trin-core/src/utp/stream.rs
```diff
@@ -312,13 +320,22 @@ impl UtpListener {
...
                                 error!("Unable to receive uTP FIN packet: {msg}")
                             }
+                        } else if let Some(conn) = self
+                            .utp_connections
+                            .get_mut(&ConnectionKey::new(*node_id, connection_id))
+                        {
+                            // Do not handle the packet here, send it to uTP socket layer
+                            if conn.discv5_tx.send(packet).is_err() {
+                                error!("Unable to send FIN packet to uTP stream handler");
+                            }
```
carver, line 334: I was surprised in this PR that the utp socket layer already handles this new case smoothly, but I'm assuming you tried it. Cool that the change ended up being pretty small!

carver (approved):
LGTM, just a comment suggestion

trin-core/src/utp/stream.rs
ogenev: yeah, the uTP socket is designed in such a way to allow bidirectional packets, It is tested but we may want to add also some uTP tests in CI.

trin-core/src/utp/stream.rs
ogenev: Actually, I changed this to [saturated_sub](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_sub) which advertises a remote window size of 0, instead of allowing underflow. I think this is better than asking for more data than we would theoretically prefer.

Ellajoke:
Randall

PR #397, opened 2022-08-03
Add client info to ENR

mrferris:
### What was wrong?

Knowing which client and version of that client your node is failing to connect to is helpful for debugging issues on the portal network.

Getting a node's info via RPC is either cumbersome or impossible, depending on the node's settings.

### How was it fixed?

- A client string (e.g. `t 0.1.0` for Trin v0.1.0) is added  to trin's ENR as a value under the key `client`.
- If a portal node has a `c` key in its ENR, it is displayed in the output of the `portal_*RoutingTableInfo` RPC endpoint, like so: 
```
{
  "client": "trin 0.1.0",
  "enr": "enr:-Je4QFui-99WVnY8Q4C3rR6iZokv_rK1d7Q4fUaWHEd6PEqPcPbVmSoaXOnHG-dFZd0L0v5GWIJXhz7XP3unaqBJv0YBhmNsaWVudIt0cmluIHYwLjEuMIJpZIJ2NIJpcIR_AAABiXNlY3AyNTZrMaEC64Hx6X6On6OQSGP-X5PaXZsCkEzUrimYclGbmwhFAXyDdWRwgiMq",
  "node_id": "0x0241dc8d483d628cbad18704e3a9b094642a07882d329c92e25fa5da4eb4155c",
  "radius": "18446744073709551615",
  "status": "Connected"
}
```

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/397.added.md
+++ b/newsfragments/397.added.md
@@ -1,0 +1,3 @@
+- Add `c` (short for `client`) key to ENR, whose value contains a client info string, eg `t v0.1.0` (as shorthand for `trin v0.1.0`)
+- Add `client` key to node objects within output of portal `RoutingTableInfo` endpoints
+- Add `address` key containing IP:Port 
--- a/trin-core/src/jsonrpc/utils.rs
+++ b/trin-core/src/jsonrpc/utils.rs
@@ -10,7 +10,7 @@ use validator::ValidationError;
 use crate::{portalnet::types::content_key::OverlayContentKey, utils::bytes::hex_decode};
 
 type NodeMap = BTreeMap<String, String>;
-type NodeTuple = (NodeId, Enr, NodeStatus, Distance);
+type NodeTuple = (NodeId, Enr, NodeStatus, Distance, Option<String>);
 
 /// Converts the output of the Overlay's bucket_entries method to a JSON Value
 pub fn bucket_entries_to_json(bucket_entries: BTreeMap<usize, Vec<NodeTuple>>) -> Value {
@@ -23,7 +23,7 @@ pub fn bucket_entries_to_json(bucket_entries: BTreeMap<usize, Vec<NodeTuple>>) -
                 bucket_index,
                 bucket
                     .iter()
-                    .map(|(node_id, enr, node_status, data_radius)| {
+                    .map(|(node_id, enr, node_status, data_radius, client_info)| {
                         node_count += 1;
                         if node_status.state == ConnectionState::Connected {
                             connected_count += 1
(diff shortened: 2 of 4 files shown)


jacobkaufmann:
It seems like we don't have consensus among the client teams yet to add this, so I'm going to hold off on review for now.

mrferris:
@jacobkaufmann Fair, I'll re-request review if/when we do.

mrferris:
Updated this to use shorthand notation within the ENR's key and value (as per @carver's suggestion): "c" for "client" and "t" for trin".

carver:
Eh, I'm okay moving forward with it before we have other clients fully convinced. I haven't really heard strenuous objection, and maybe we'll convince them otherwise after we can demonstrate the value by having trin share the info.

newsfragments/397.added.md
```diff
@@ -0,0 +1,2 @@
+- Add `client` key to ENR, whose value contains a client info string, eg `trin v0.1.0`
```
carver, line 1:
```suggestion
- Add `client` key to ENR, whose value contains a client info string, eg `t 0.1.0` (as short-hand for trin v0.1.0)
```

trin-core/src/jsonrpc/utils.rs
```diff
@@ -36,6 +36,9 @@ pub fn bucket_entries_to_json(bucket_entries: BTreeMap<usize, Vec<NodeTuple>>) -
                         map.insert("enr".to_owned(), enr.to_base64());
                         map.insert("status".to_owned(), format!("{:?}", node_status.state));
                         map.insert("radius".to_owned(), format!("{}", data_radius));
+                        if let Some(client) = client_info {
+                            map.insert("client".to_owned(), client.to_string());
```
carver, line 40: If you were feeling fancy, you could expand the `t` to `trin` in the output

carver (approved):

trin-core/src/jsonrpc/utils.rs
mrferris: Yea, I did that translation at first but removed it because I felt it would be brittle (across version and clients) for not much gain. But now I'm realizing I can just try to expand it, and if that fails for whatever reason just leave the original.

carver:
I don't think you needed to wait for another 👍🏻 but just in case you are... you have another from me.!

trin-core/src/jsonrpc/utils.rs
```diff
@@ -36,6 +36,17 @@ pub fn bucket_entries_to_json(bucket_entries: BTreeMap<usize, Vec<NodeTuple>>) -
                         map.insert("enr".to_owned(), enr.to_base64());
                         map.insert("status".to_owned(), format!("{:?}", node_status.state));
                         map.insert("radius".to_owned(), format!("{}", data_radius));
```
njgheorghita, line 38: Could you also add the `ip` address in here (`enr.ip4()` ... I think)? I've been doing this locally and it helps identify the client you're talking to.... Actually, how about you only output the ip if there is not client info found, to reduce clutter?

njgheorghita (approved):
Yup, looks good to me! This will be super useful for debugging the testnet

trin-core/src/jsonrpc/utils.rs
mrferris: Good idea, done.

I added a note that it can be removed once we have a nice dashboard, because then we can move from reading raw RPC output (like barbarians) and just decode the ENR UI-client-side.

mrferris: @njgheorghita weird, with that change that you proposed I seem to get a failing utp test and failing peertest. That doesn't sound right, going to investigate further.

carver: > @njgheorghita weird, with that change that you proposed I seem to get a failing utp test and failing peertest in CI. That doesn't sound right, going to investigate further.

I'm curious, what was the resolution to this?

PR #400, opened 2022-08-04
Increase the default number of rpc serving threads

carver:
This may be responsible for the timeouts on Windows CI, maybe clients
aren't closing their connections in the same way?

Anyway, it's a small enough change that it shouldn't wreck performance,
and it's worth it to check if this fixes this pain in the butt CI issue.

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/cli.rs
+++ b/trin-core/src/cli.rs
@@ -44,7 +44,7 @@ pub struct TrinConfig {
     pub web3_ipc_path: String, // TODO: Change to PathBuf
 
     #[structopt(
-        default_value = "2",
+        default_value = "5",
         long = "pool-size",
         help = "max size of threadpool"
     )]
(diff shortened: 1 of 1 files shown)


njgheorghita:
I'm fairly confident that this solves the problem. Tacked it onto #376 and saw just 1/5 failed test runs (`win-build`). The one failed test run was a different error. Not sure how many successful test runs confirm "nonflakiness"?

ogenev:
It looks like the windows pipeline doesn't run all tests  :scream: . I'm trying to address this in https://github.com/ethereum/trin/pull/403.

carver:
Okay, as far as we can tell, this is reducing the flakiness. I think the best way to confirm this is to merge it aggressively, and confirm that it's not affecting PRs anymore.

Rage-merge!

PR #403, opened 2022-08-05
Test and build the Trin workspace in `win-build` CI pipeline

ogenev:
### What was wrong?
Since genesis, win-build workflow in CI doesn't run all tests in the workspace.

### How was it fixed?
- Add `--workspace` flag to `cargo build` and `cargo test`
- Update the` clippy` command to check for all features and remove depreciated `--all` flag.
-  Add windows support for `trin-cli`. (this issue was found with the updated clippy command)

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,7 +75,7 @@ jobs:
           # Remove the first two lines of gitconfig as work around
           command: |
             (gc ..\.gitconfig | select -Skip 2) | sc ..\.gitconfig
-            cargo clippy --package trin -- --deny warnings
+            cargo clippy --workspace --all-targets --all-features --no-deps -- --deny warnings
       - run:
           name: Build Trin workspace
           command: cargo build --workspace --target x86_64-pc-windows-msvc
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,11 +77,11 @@ jobs:
             (gc ..\.gitconfig | select -Skip 2) | sc ..\.gitconfig
             cargo clippy --package trin -- --deny warnings
       - run:
-          name: Cargo Build --target x86_64-pc-windows-msvc
-          command: cargo build --target x86_64-pc-windows-msvc
+          name: Build Trin workspace
+          command: cargo build --workspace --target x86_64-pc-windows-msvc
       - run:
-          name: Cargo Test --target x86_64-pc-windows-msvc
-          command: cargo test --target x86_64-pc-windows-msvc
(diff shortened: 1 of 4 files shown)


ogenev:
Should we merge this PR first and then ignore/temporarily stop the win pipeline until we fix all failing tests?

njgheorghita:
Nice catch! I'm of the opinion that it'd be better to bump the priority of this pr and fix failing tests here before merging. Any reason that you'd want to merge this with a failing `win-build`?

mrferris:
I agree with fixing failing tests here, as long we can in a reasonable amount of time. Otherwise I can see it being better to merge this (so that other PRs can make use of the genius idea of running unit tests on windows) and then fix the broken tests in a separate PR.

Anyway, the failures are all from one issue: the sqlite interface seems to want to have sole ownership of the entire `C:\Users\circleci\AppData\Local\Temp\trin\.tmp` directory instead of just the `\trin_abcdef\trin.sqlite` subdirectory of that directory that we're giving it. Notable that this only happens when using a temp directory, and not with the default data directory when running, for example, the peertest. I'm not yet quite seeing why that's happening. Any ideas @njgheorghita?

njgheorghita:
Hmmm... haven't pinned it down just yet, but I imagine that it's related to [this](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html#resource-leaking) and our use of `TempDir` in `trin_core::utils::db::setup_temp_dir` in those tests that are failing

mrferris:
I was thinking the same, but the fact that the peertest succeeds even though it also uses `setup_temp_dir` (because it uses the `--ephemeral` flag) makes me think that it's not related.

Do we have a reason for using such an old verison of the windows circle ci orb? We're using `2.1` and `5.0` is out. The unit tests should still pass on `2.1` (which uses windows server 2019) though, so I don't think that's related either.

mrferris:
@ogenev if you want to enable "allow edits by maintainers" on this PR I'll push the commit that fixes the unit tests to this PR, or we can just merge this and then I'll merge [this PR](https://github.com/ethereum/trin/pull/404) where I tested the fix

mrferris:
@ogenev alternatively, we can just merge #404 and close this. I'll open that one for review/merge unless we agree otherwise.

carver (approved):
Looks good. At this point, it's probably faster to just merge #404 which has imported all these commits, and fixed the windows tests.

ogenev:
Closing this in favor of #404.

PR #405, opened 2022-08-10
uTP Testing Framework

ogenev:
### What was wrong?
Fixes https://github.com/ethereum/trin/issues/232

The lack of uTP testing infrastructure prevents us from testing core uTP protocol under various network conditions close to reality.

### How was it fixed?
- A testing framework that enables to test uTP implementation over different
network conditions. Highly based on tools developed to test QUIC protocol:
   [quic-interop-runner](https://github.com/marten-seemann/quic-interop-runner)
   [quic-network-simulator](https://github.com/marten-seemann/quic-network-simulator)

- Add  a test suite to transfer 1000 bytes with no drop rate
Note: Drop rate will be enabled when we add code to resend lost uTP packets and the transfer size will be increased when we fix a bug with building selective acks found using this framework.

- CircleCI pipeline to run current uTP test suite automatically

### How it works

Test setup uses docker compose to start 3 docker containers:
- client - which is an instance of uTP test app
- server - which is an instance of uTP test app
- sim - which is an instance with ns3 network simulator with several pre-compiled scenarios

The networking is setup in such a way that network traffic is routed from client to server
and server to client through sim which decides what to do with flowing packets

Explanation from [quic-network-simulator](https://github.com/marten-seemann/quic-network-simulator):

```
The framework uses two networks on the host machine: `leftnet` (IPv4
193.167.0.0/24, IPv6 fd00:cafe:cafe:0::/64) and `rightnet` (IPv4
193.167.100.0/24, IPv6 fd00:cafe:cafe:100::/64). `leftnet` is connected to the
client docker image, and `rightnet` is connected to the server. The ns-3
simulation sits in the middle and forwards packets between `leftnet` and
`rightnet`
```

### To-Do

- [x] Clean up commit history


utp-testing/src/bin/test_suite.rs
```diff
@@ -0,0 +1,60 @@
...
+
+/// Test suite for testing uTP protocol with network simulator
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    send_1000_bytes().await?;
+
+    Ok(())
+}
+
+/// Send 10k bytes payload from client to server
```
njgheorghita, line 19: Looks like we're sending  `1k` bytes not `10k`?

utp-testing/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        let mut conn = rx.await.unwrap();
+
+        let mut buf = [0; 1500];
+        conn.recv(&mut buf).await.unwrap();
+
+        conn.send_to(&payload).await.unwrap();
+
+        tokio::spawn(async move {
+            conn.close().await.unwrap();
+            debug!("connecton state: {:?}", conn.state)
```
njgheorghita, line 89: *typo*: `connection`

.circleci/config.yml
```diff
@@ -84,8 +84,63 @@ jobs:
...
+          command: cargo build -p utp-testing --release
+      - run:
+          name: Move docker artifacts
+          command: mv -t utp-testing/docker/circleci/ target/release/utp-test-app utp-testing/docker/run_endpoint.sh utp-testing/docker/setup.sh
+      - run:
+          name: Build uTP Test App container
+          command: docker build -t test-utp --no-cache utp-testing/docker/circleci/
+      - run:
+          name: Run Test App with network simulator
+          command: SCENARIO="drop-rate --delay=15ms --bandwidth=10Mbps --queue=25 --rate_to_client=0 --rate_to_server=0" docker-compose -f utp-testing/docker/docker-compose.yml up -d
```
njgheorghita, line 127: It seems to me like we are only running the utp transfer test under a single specific scenario? Is the idea to eventually run the same utp transfer test under multiple scenarios with various network attributes (aka different delay, bandwidth, etc)?

utp-testing/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+        match utp_payload {
+            Some(payload) => Ok(hex_encode(payload)),
+            None => Ok("None".to_string()),
+        }
+    }
+
+    async fn prepare_to_recv(&self, enr: String, conn_id: u16) -> RpcResult<String> {
+        let enr = Enr::from_str(&*enr).unwrap();
+        self.prepare_to_receive(enr, conn_id).await;
+        Ok("OK".to_owned())
```
njgheorghita, line 54: I'm not so excited about using `"OK"` and `"None"` as return types... It looks like `RpcResult<bool>` might be a more suitable type?

njgheorghita (approved):
Cool! Excited to get some more robust testing around utp.

.circleci/config.yml
ogenev: It is not a problem to docker-compose multiple networks with different scenarios but I'm not sure if we will need that. 
If we can transfer a significant amount of bytes during worst case scenario possible (some drop rate, big delay, low bandwidth), I think this will be enough.

PR #411, opened 2022-08-17
Offer history content to all peers over rpc

carver:
### What was wrong?

Replacing #384 

Need a way to push content over RPC, to build bridge node.

### How was it fixed?

- Move current logic in `portal_historyOffer` to `portal_historySendOffer`
- Add `portal_historyOffer` to broadcast a piece of content to all relevant peers
- slightly more convenient public function for propagating gossip in the overlay
- When validating accepted content, and finding that it fails, skip it instead of panicking

It's probably easiest to read through each commit in the PR individually.

### To-Do

- [x] Clean up commit history
- [x] Confirmation from other teams about the shift of the meaning of `portal_historyOffer`
- [x] re-enable `test_offer_accept()`, with a retry after sleep

Probably as good as it gets for client team confirmation, over Discord:

> acolytec3 — 8/23 at 5:26 PM
> Should be fine for us.  I don't know that we've even implemented this endpoint since our RPC is woefully lacking
> deme — 8/24 at 1:20 AM
> We didn't implement exactly that one either. But we can add it, and I'd be fine with such change I think.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,18 +70,6 @@ jobs:
           name: Install target toolchain
           command: rustup toolchain install stable-x86_64-pc-windows-msvc
       - run:
-          name: Install Clippy
-          command: rustup component add clippy
-      - run:
-          name: Run Clippy
-          # Remove the first two lines of gitconfig as work around
-          command: |
-            (gc ..\.gitconfig | select -Skip 2) | sc ..\.gitconfig
-            cargo clippy --package trin -- --deny warnings
-      - run:
-          name: Cargo Build --target x86_64-pc-windows-msvc
-          command: cargo build --target x86_64-pc-windows-msvc
-      - run:
           name: Cargo Test --target x86_64-pc-windows-msvc
           command: cargo test --target x86_64-pc-windows-msvc
   utp-test:
--- a/trin-core/src/jsonrpc/types.rs
+++ b/trin-core/src/jsonrpc/types.rs
@@ -239,12 +239,12 @@ impl TryFrom<[&Value; 2]> for FindContentParams {
     }
 }
 
-pub struct OfferParams {
+pub struct SendOfferParams {
     pub enr: SszEnr,
     pub content_keys: Vec<ByteList>,
 }
 
-impl TryFrom<Params> for OfferParams {
+impl TryFrom<Params> for SendOfferParams {
     type Error = ValidationError;
 
     fn try_from(params: Params) -> Result<Self, Self::Error> {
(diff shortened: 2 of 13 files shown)


trin-core/src/jsonrpc/types.rs
```diff
@@ -239,12 +239,66 @@ impl TryFrom<[&Value; 2]> for FindContentParams {
...
+            .as_str()
+            .ok_or_else(|| ValidationError::new("Empty content param"))?;
+        let content = match hex_decode(content) {
+            Ok(val) => val,
+            Err(_) => return Err(ValidationError::new("Unable to decode content")),
+        };
+        Ok(Self {
+            content_key,
+            content,
+        })
```
carver, line 292: There's a whole lot of copy/paste in here from `StoreParams`. Would probably be nice to reduce that

trin-core/src/portalnet/overlay.rs
```diff
@@ -259,58 +259,68 @@ where
...
-    /// Propagate gossip accepted content via OFFER/ACCEPT:
-    fn propagate_gossip(
+    /// Propagate gossip accepted content via OFFER/ACCEPT, return number of peers propagated
+    pub fn propagate_gossip(&self, content: Vec<(TContentKey, ByteList)>) -> usize {
+        let kbuckets = Arc::clone(&self.kbuckets);
+        let command_tx = self.command_tx.clone();
+        Self::_propagate_gossip(content, kbuckets, command_tx)
+    }
+
+    fn _propagate_gossip(
```
ogenev, line 319: Not sure if it is idiomatic rust to underscore private methods, my understanding is that underscore is used as a way to ignore something in Rust? Took a quick look into some of the most popular rust crates and I can't find this pattern.

.circleci/config.yml
```diff
@@ -69,18 +69,6 @@ jobs:
...
           name: Install target toolchain
           command: rustup toolchain install stable-x86_64-pc-windows-msvc
-      - run:
-          name: Install Clippy
-          command: rustup component add clippy
-      - run:
-          name: Run Clippy
-          # Remove the first two lines of gitconfig as work around
-          command: |
-            (gc ..\.gitconfig | select -Skip 2) | sc ..\.gitconfig
```
ogenev, line 79: Removing this is causing the windows pipeline to fail. It looks like our solution is kinda hacky but it works, it is related to a known [issue](https://github.com/rust-lang/cargo/issues/2078) with CIrcleCI and Cargo.

Maybe a more convenient solution would be to add `- run: git config --global --unset url.ssh://git@github.com.insteadof` but haven't tried it.

ogenev (approved):
Looks good, we only need to fix the windows pipeline.

.circleci/config.yml
carver: Cool, thanks for that link. Trying the workaround posted in it first.

carver:
I'm going to see if I can get #424 fixed so that we can turn CI green (AFAICT it's unrelated, but having a flaky CI is a huge PITA that deserves our focused attention).

Issue #311, opened 2022-04-22
Validate via Oracle


njgheorghita:
Headers, block bodies, receipts

# Validation (via infura) needs to happen
- [x] For all content fetched on demand from a network (regardless of whether or not we store it)
- [x] For all content fetched on demand from a network via utp
- [x] For `Accept`ed content that's been `Offer`ed to us (via utp)
- [x] For `BlockBodies` & `Receipts` in the Chain History network

random todo:
- [x] Include how to use `web3.py` as general purpose jsonrpc server

PR #413, opened 2022-08-17
Validate content received over utp for findcontent requests

njgheorghita:
### What was wrong?
Validation via `HeaderOracle` needed for content returned from a `FindContent` request.

Wraps up #311 

### How was it fixed?
Added validation for returned content

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -497,7 +497,9 @@ where
         content_key: Vec<u8>,
     ) -> Result<Content, OverlayRequestError> {
         // Construct the request.
-        let request = FindContent { content_key };
+        let request = FindContent {
+            content_key: content_key.clone(),
+        };
         let direction = RequestDirection::Outgoing {
             destination: enr.clone(),
         };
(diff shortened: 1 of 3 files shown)


trin-core/src/portalnet/overlay.rs
```diff
@@ -514,8 +516,17 @@ where
                     // Init uTP stream if `connection_id`is received
                     Content::ConnectionId(conn_id) => {
                         let conn_id = u16::from_be(conn_id);
-
-                        self.init_find_content_stream(enr, conn_id).await
+                        let content = self.init_find_content_stream(enr, conn_id).await?;
+                        let content_key = TContentKey::try_from(content_key)
+                            .expect("Error decoding valid content key.");
```
ogenev, line 521: Should we handle gracefully this exception?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -119,6 +119,10 @@ pub enum OverlayRequestError {
     #[error("Invalid response")]
     InvalidResponse,
 
+    /// An invalid response was received.
+    #[error("Response failed validation")]
+    FailedValidation,
```
ogenev, line 124: IMO it will be nice if we include the actual content value in this error message. If validation fails, it will be useful to know the content value that we are receiving.

ogenev (approved):
Looks good, just a few comments.

trin-core/src/portalnet/overlay.rs
njgheorghita: My thinking was that it is excessive, since the content key will already have been validated when the overlay network initiates the request [here](https://github.com/ethereum/trin/blob/master/trin-history/src/jsonrpc.rs#L165) - but I can add it I'm missing another case where an invalid content key might sneak through up to this point?

njgheorghita: On second thought, might as well add in the graceful handling, just in case.

PR #420, opened 2022-08-22
Move node type out of overlay service

njgheorghita:
### What was wrong?
`overlay_service.rs` file is becoming quite cluttered.

### How was it fixed?
Moved the `Node` type into its own file.
Some refactoring of imports.

### To-Do

- [x] Clean up commit history


carver (approved):
Looks straightforward 👍🏻

PR #421, opened 2022-08-23
refactor: define trait for portal data store

jacobkaufmann:
### What was wrong?

There is no abstraction for the data store layer, so a database must be set up and torn down even in tests where the particular database is not relevant.

### How was it fixed?

- Define trait `ContentStore` to provide abstraction over the data store layer.
- Implement `ContentStore` for `PortalStorage`.
- Replace `PortalStorage` with `ContentStore` generic type parameter in `OverlayProtocol` and `OverlayService`.
- Define `MemoryContentStore` for use in tests.

### To-Do

- [x] Clean up commit history


jacobkaufmann:
Here I removed the tests for `should_store` and `distance_to_key`. For `should_store`, that method was removed, so there is no test to have for it. For `distance_to_key`, it was not clear to me what we test there.

I preferred the combination of `is_key_within_radius` and `get` in the `PortalContentStore` trait over `should_store`. Those two methods are useful in isolation, and when combined, they are equivalent to `should_store`. `get` would be in the trait regardless.

trin-core/src/portalnet/overlay.rs
```diff
@@ -277,11 +278,8 @@ where
                                     // Skip storing & propagating content if it's not valid
                                     .expect("Unable to validate received content: {err:?}");
 
-                                // Check if data should be stored, and store if true.
                                 // Ignore error since all validated content is propagated.
-                                let _ = storage
-                                    .write()
-                                    .store_if_should(&key, &content_value.to_vec());
+                                let _ = store.write().put(key.clone(), &content_value.to_vec());
```
njgheorghita, line 282: My concern with using `put()` here is that it's not obvious that the radius check is happening... which could lead to us performing a redundant radius check, or "putting" data into storage we expect to be stored.

imo, w/ databases generally "put" has a pretty cut & dry implication that we are simply "putting" data into the db, which isn't the case here.

Likewise, I believe the removed comment is still valid, and probably better off left in since (imo) it's not obvious that this check is happening.

trin-core/src/portalnet/storage.rs
```diff
@@ -24,82 +26,217 @@ pub enum DistanceFunction {
...
+        key: &K,
+    ) -> Result<Option<Vec<u8>>, PortalContentStoreError> {
+        let key = key.content_id();
+        let value = self.db.get(key)?;
+        Ok(value)
+    }
 
-    #[error("Content can not be stored since it falls outside our radius")]
-    OutsideDistanceError,
+    fn put<K: OverlayContentKey, V: AsRef<[u8]>>(
```
njgheorghita, line 209: This function feels somewhat redundant. Why aren't we just using `store()`?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -933,20 +934,26 @@ where
                 ))
             })?;
 
-            requested_keys
-                .set(
-                    i,
-                    self.storage.read().should_store(&key).map_err(|err| {
```
njgheorghita, line 939: In this case, `should_store()` seems preferable to encapsulate the [checking radius logic and checking if it's already been stored logic] at the storage layer.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1327,35 +1334,37 @@ where
...
+
+                    if let Err(err) = store.write().put(content_key, content.to_vec()) {
+                        error!("Content received, but not stored: {err}")
+                    }
+                });
             }
-            Err(_) => {
-                error!("Content received, but not stored: Error communicating with db.");
+            Ok(Some(_)) => {
+                debug!("Content received, but not stored: content already stored.");
```
njgheorghita, line 1364: Similar here, this error handling seems more appropriate to be handled at the storage layer.

trin-core/src/portalnet/storage.rs
```diff
@@ -24,82 +26,217 @@ pub enum DistanceFunction {
...
+
+    /// Returns the distance to `key` from the local `NodeId` according to the distance function.
+    fn distance_to_key<K: OverlayContentKey>(&self, key: &K) -> U256 {
+        match self.distance_fn {
+            DistanceFunction::Xor => XorMetric::distance(&key.content_id(), &self.node_id.raw()),
+        }
+    }
+}
+
+impl PortalContentStore for MemoryPortalContentStore {
```
njgheorghita, line 137: I'd prefer to see `MemoryPortalStorage` here. I thought the ideas was that it's a like-for-like replacement for `PortalStorage`?

njgheorghita (commented):
I'm a bit curious as to some of the api changes you made here? At first glance - and it was a fairly brief skim tbf - I'm not sure I understand some of the advantages gained by parts of this refactor? imo, if there's not a strong reason to change it, there's no need. It's just another change I'll have to remember/learn, and my brain capacity is limited.

I thought the trait would be confined to simply how `PortalStorage` gets & sets data into it's db. And with this new trait, now the db could be either rocks or a hashmap. aka, introducing this trait will not touch any of logic outside `PortalStorage`. 

The basic trait implementation here checks out, but I'm not convinced on the other api changes (eg. `should_store`, `store`, `store_if_should`). 

Maybe this is the part that's holding me up: `"Those two methods are useful in isolation"`. I don't really see where this applies to `is_key_within_radius()`. Imo, this logic is better encapsulated at the storage layer, and shouldn't need to be handled in the overlay logic. 

Idk, I'm probably missing some of the benefits here, curious to hear how you saw these api changes improving how we use `PortalStorage`

trin-core/src/portalnet/overlay.rs
```diff
@@ -91,11 +91,11 @@ impl Default for OverlayConfig {
...
 /// handling common network requests/responses.
 #[derive(Clone)]
-pub struct OverlayProtocol<TContentKey, TMetric, TValidator> {
+pub struct OverlayProtocol<TContentKey, TMetric, TValidator, TStore> {
     /// Reference to the underlying discv5 protocol
     pub discovery: Arc<Discovery>,
-    /// Reference to the database instance
-    pub storage: Arc<RwLock<PortalStorage>>,
+    /// The data store.
+    pub store: Arc<RwLock<TStore>>,
```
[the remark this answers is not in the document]
jacobkaufmann: A "data store" is a widely recognized concept for persistent data storage (e.g. key-value store). The motivation is to move to this more canonical naming. I only kept `PortalStorage` in other places to minimize the diff.

If we go with the "store" naming, then `PortalStorage` would become something like `PortalDataStore` or `SqlitePortalContentStore` (if there is a single database).

trin-core/src/portalnet/storage.rs
jacobkaufmann: I'm not sure what you mean. Are you asking why the name is not `MemoryPortalStorage`?

trin-core/src/portalnet/overlay.rs
jacobkaufmann: We can remove the radius check from `put`. Currently, [`store`](https://github.com/ethereum/trin/blob/8a9bb83a416aefbd0d095baa9346774d5ee97704/trin-core/src/portalnet/storage.rs#L164) behaves the same way. I would not be opposed to removing this check in `put`. If it falls outside of the radius, and the storage capacity is exceeded, then that data would be the farthest item and would be immediately purged. It might be useful to make the return type of `put` a list of evicted content keys.

As things are now, `put` may fail, and a possible error is `InsufficientRadius`, so a look at the API communicates that the radius may be a reason for failure. If that's not clear, and we want to keep the radius check there, then we can add more documentation.

For a key-value store, which is what the abstraction of the storage layer should be IMO, "get" and "put" are the canonical APIs (e.g. RocksDB).

Based on what we decide I can undelete the comment or add a new comment.

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann: The disadvantage of `should_store` is that the reason you should or should not store some data is opaque. You receive a boolean, a yes or no, but you do not know why. It seems useful to be able to differentiate between the radius check and the contains check. You could return an enum from `should_store`, but then you would need to do some matching and end up with something similar to what I have below.

This combination of logic would be less clunky without the `map_err`. A `DataStoreError` (or similar) should be added so that we can use the `?` syntax directly after `get`.

trin-core/src/portalnet/storage.rs
jacobkaufmann: Well whether we call it `store` or `put`, we need to implement the trait. I'm not sure if you mean that we should just call this method `store` in the trait definition. For a comment on that, see comment thread above.

jacobkaufmann:
@njgheorghita I appreciate the healthy skepticism! The intent here is to abstract the portal data store, which is effectively a key-value store. With this abstraction, we can use something like a `HashMap` in tests where we do not care about the particular database storing the data. This removes failure modes from certain tests and also improves the performance of tests where we do not need to set up and tear down database connections.

Based on the comments, a point of contention is the change in naming, i.e. why `put` instead of `store`? My main argument here is that this abstraction is a key-value store, and `get` and `put` are the canonical names there. As of now, `put` is a wrapper around the pre-existing `store` method of `PortalStorage`, but this wrapping was only done to minimize the diff. In a subsequent refactor, there would only be one method that contains this logic.

Another point of contention is the absence of `should_store` in the trait definition, and as a consequence, the need to call `is_key_within_radius` in combination with `get` to achieve the same thing. Why is `is_key_within_radius` useful in isolation? It seems useful to know that you are not storing something because it falls outside of the data store's radius versus because it already exists in the data store. This is also a question of API design, which is more subjective. My preference is for an API that provides the base functionality a caller needs to use the thing correctly. That might require more of the caller, for example to combine function calls, but it keeps the trait definition simple.

trin-core/src/portalnet/overlay.rs
mrferris: It also doesn't seem like we're checking whether we already have the data here, which will get expensive when we're writing to disk.

I'm not totally against simplifying `put` to not do any checks, but I'm not sure about the idea of letting things outside of the radius be added. Especially in this scenario, when we're looping through a bunch of data and `put`ing it. Stored items outside of the radius will be immediately bumped, wasting the I/O necessary to have stored them because there was never an opportunity to later share them with the network.

> If it falls outside of the radius, and the storage capacity is exceeded

nit: The radius doesn't begin decreasing from `U256::MAX` until the storage capacity is reached, so there is no case where something's outside of the radius and storage capacity isn't exceeded.

mrferris (changes requested):
I support the vision! 

Left a few notes and requested changes.

trin-core/src/portalnet/storage.rs
```diff
@@ -24,82 +26,217 @@ pub enum DistanceFunction {
...
+        let content_id = key.content_id();
+        let value: &[u8] = value.as_ref();
+        self.store.insert(content_id.to_vec(), value.to_vec());
+
+        Ok(())
+    }
+
+    fn is_key_within_radius<K: OverlayContentKey>(&self, key: &K) -> bool {
+        let distance = self.distance_to_key(key);
+        distance <= self.radius
```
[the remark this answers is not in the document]
jacobkaufmann: Right. I was not sure whether it would be useful to have `MemoryPortalContentStore::set_radius`. As things are, we can remove the check.

trin-core/src/portalnet/storage.rs
```diff
@@ -24,82 +26,217 @@ pub enum DistanceFunction {
...
+}
+
+impl From<std::io::Error> for PortalContentStoreError {
+    fn from(err: std::io::Error) -> Self {
+        Self::Io(err)
+    }
+}
+
+/// A data store for Portal Network content (data).
+pub trait PortalContentStore {
```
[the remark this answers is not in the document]
jacobkaufmann: I like this! I will keep the name `PortalStorage` for now though to minimize the diff.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: Yeah, without checks in `put`, it would be the responsibility of the caller to perform the check and avoid unnecessary I/O. I'm also curious on how we feel about `put` returning the list of evicted content keys/IDs @mrferris.

mrferris: @jacobkaufmann I meant that currently this code doesn't check whether an item is already in the DB. `PortalStorage::store` only checks radius. So currently it will write duplicate data, which is what I was referring to as expensive. I might be wrong, but I don't think that sqlite or rocksdb come with the optimization of not re-writing identical data. So I think that a `has_key` trait method (whose `PortalStorage` implementation uses the efficient `rocksdb::get_pinned`) is a good idea.

Which use-case do you have in mind for `put` returning evicted items?

trin-core/src/portalnet/storage.rs
njgheorghita: For sure, I'd lean towards using the same name for both the function in the trait, and the one it proxies towards, whatever naming we end up with

trin-core/src/portalnet/overlay_service.rs
njgheorghita: Just a standard word of caution, since it's something I'm guilty of all the time, but just b/c "It seems useful to be able to differentiate between the radius check and the contains check.", doesn't mean we need to introduce this complexity until it's actually required. Until we have a valid reason for knowing whether data wasn't stored b/c it already exists, or b/c it falls outside our radius, why not just try and push it in and ignore the result. But, maybe there is a valid reason use case now/down the road i'm missing

trin-core/src/portalnet/storage.rs
njgheorghita: Yeah pretty much, but you can ignore this if we're going with the `"...Store"` naming scheme

njgheorghita:
@jacobkaufmann Just playing devils advocate here: w/r to "It seems useful to know that you are not storing something because it falls outside of the data store's radius versus because it already exists in the data store." ... Ok, but why? What is a situation where this will be useful?

trin-core/src/portalnet/overlay_service.rs
jacobkaufmann: @njgheorghita well my argument is that it's not additional complexity, and that it could be considered to be removal of complexity. To me, it makes the storage behavior easier to reason about. The opaque and more complicated `should_store` function is broken down into simple components. Either way, we can still call `put` to attempt to store the data and ignore the result.

trin-core/src/portalnet/overlay.rs
jacobkaufmann: Right now I think it would mostly be useful for debugging. It would also be sort of similar to [`HashMap::insert`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.insert), which returns an `Option` that may hold the previous value for a particular key. That said, I don't think those are strong arguments in favor of the change. We can add it later if desired.

jacobkaufmann:
@njgheorghita I will reiterate my argument above that the new interface is actually a simplification. IMO this is the stronger part of the argument in favor of the decomposition.

Consider

```rust
if store.should_store(key) { .. }
```

and

```rust
if store.is_key_within_radius(key) && !store.contains_key(key) { .. }
```

It's very unclear what sort of logic is going on in the former. I understand that there is an argument that we should not care if that logic is opaque, and that is the benefit of the abstraction. I'm not very sympathetic to that argument here. The latter is only slightly more verbose, and much more straightforward to reason about. The latter is also easier to implement and test for some arbitrary data store.


I admit the utility of `is_key_within_radius` by itself is limited. One thing that I could imagine is that it would be helpful to assess whether our peers have an accurate view of our radius. If our peers often offer us data that is outside the radius, then it would seem that those peers are behaving incorrectly or do not have an accurate view of our radius.

The purpose of the PR is to define a solid abstraction for the data store layer. If we were to break things up, I would prefer to leave out the in-memory data store and focus on getting the abstraction right.

mrferris:
@jacobkaufmann Good point, calling code does need to know why we shouldn't store data. We should eventually be acting on the case where we're offered data outside our radius, and keeping these functions separate is a better solution than `should_store` returning a custom enum that specifies the reason or something like that.

I'm fine with either finalizing that change here (with `contains_key` added to the trait) or keeping `should_store` and then modifying the trait later when we modify the calling code to act on out-of-radius `OFFER`s. 

Two more questions: 

1. How can we test that these updates to the calling code don't break any `OFFER`/`ACCEPT` logic? 
2.  > For distance_to_key, it was not clear to me what we test there.

    Why not test `PortalStorage::distance_to_key`?

njgheorghita:
@jacobkaufmann Point taken, that api was somewhat obscure. I'm 👍 on  the api changes. Let's keep both the api changes and hashmap support changes in this pr.

jacobkaufmann:
@njgheorghita @mrferris so I was talking with @carver offline last week, and the middle ground that we reached was to keep the shape of the storage API the same, but to rename `should_store` to something more descriptive, and to remove the `store_if_should` pattern.

We agreed that the key-value store should be separate from the logic associated with radius management. Ultimately, this may involve a further abstraction for the key-value store within the content store, but for now we can use a single outer trait.

(I may have misremembered some of the details about the discussion referenced above.)

To address your questions @mrferris:

1. Can you clarify?
2. Are we just testing the conversion of the content key to a content ID? That is quite straightforward. The distance computation logic is already covered by the tests for the distance function.

mrferris:
@jacobkaufmann 

1. No longer relevant, but I was wondering if there's any testing that would validate that the `should_store` logic being moved out of `should_store` and made more granular into `overlay.rs` was behaving properly (eg wouldn't accidentally accept content we don't want to accept). This isn't relevant anymore with the direction you've gone with renaming `should_store` to `is_key_within_radius_and_unavailable`.

2. Yea good point about the distance functions themselves already being tested

carver:
@mrferris any objections? The PR is still showing you as 👎🏻

jacobkaufmann:
@carver I think @mrferris was waiting on the re-request for review (just done) to acknowledge all comments addressed.

mrferris:
@jacobkaufmann Re-reviewing now, and yea @carver I thought that doing a "comment" review after a "request changes" would change from a 👎 to a neutral, but apparently not!

mrferris (approved):
👍

Just want to note that I still think that, as @jacobkaufmann pointed out, we'll still ultimately need to break up `is_within_radius_and_unavailable` into its two components in order take different actions based on the reason for not storing.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -934,20 +935,21 @@ where
...
-                )
+            // Accept content if within radius and not already present in the data store.
+            let accept = self
+                .store
+                .read()
+                .is_key_within_radius_and_unavailable(&key)
                 .map_err(|err| {
                     OverlayRequestError::AcceptError(format!(
-                        "Unable to set requested keys bits: {err:?}"
+                        "unable to check content availability {err}"
```
njgheorghita, line 945: *nitpick* but in general, I think it's better overall if we consistently begin error messages with capital letters throughout the codebase

njgheorghita (approved):
🚢

jacobkaufmann:
@mrferris @njgheorghita I added some logic to accommodate the accumulator in the `MemoryContentStore` along with some tests for the `MemoryContentStore`. If there are any final comments (aside from resolving merge conflicts), please let me know. Otherwise I will merge by EOD.

Issue #417, opened 2022-08-18
Build selective acks bug in uTP


ogenev:
Empty selective acks field is returned when we receive uTP packet with seq number > ack_nr + 1.

PR #422, opened 2022-08-24
Fix some uTP issues

ogenev:
### What was wrong?
Fixes https://github.com/ethereum/trin/issues/417

Some uTP issues were observed when testing uTP protocol with 10k byte transfer and network delay.

### How was it fixed?
- Build selective ack for an empty incoming buffer, this case was not handled.
- Insert packets into the incoming buffer before packet handling.
- Remove all acknowledged packets from send window upon selective ack receive.
- Emit close event after FIN packet is fully handled.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/417.fixed.md
+++ b/newsfragments/417.fixed.md
@@ -1,0 +1,2 @@
+- Fix selective ask bug in uTP protocol.
+- Fix a bug caused by emitting uTP stream close event before the incoming buffer is flushed.
--- a/newsfragments/417.fixed.md
+++ b/newsfragments/417.fixed.md
@@ -1,2 +1,2 @@
-- Fix selective ask bug in uTP protocol.
+- Fix selective ack bug in uTP protocol.
 - Fix a bug caused by emitting uTP stream close event before the incoming buffer is flushed.
--- a/trin-core/src/utp/stream.rs
+++ b/trin-core/src/utp/stream.rs
@@ -927,7 +927,7 @@ impl UtpStream {
                 let mut reply = self.prepare_reply(packet, PacketType::State);
 
                 if packet.seq_nr().wrapping_sub(self.ack_nr) > 1 {
-                    debug!(
+                    warn!(
                         "current ack_nr ({}) is behind received packet seq_nr ({})",
                         self.ack_nr,
                         packet.seq_nr()
@@ -1325,6 +1325,14 @@ impl UtpStream {
 
         debug!("received {:?}", packet);
 
+        // Insert data packet into the incoming buffer if it isn't a duplicate of a previously
+        // discarded packet
+        if packet.get_type() == PacketType::Data
+            && packet.seq_nr().wrapping_sub(self.last_dropped) > 0
+        {
+            self.insert_into_buffer(packet.clone());
+        }
+
         // Process packet, including sending a reply if necessary
         if let Some(mut pkt) = self
             .handle_packet(&packet, self.connected_to.clone())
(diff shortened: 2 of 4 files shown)


newsfragments/417.fixed.md
```diff
@@ -0,0 +1,2 @@
+- Fix selective ask bug in uTP protocol.
```
jacobkaufmann, line 1:
```suggestion
- Fix selective ack bug in uTP protocol.
```

utp-testing/docker/circleci/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:bullseye-slim
+FROM ubuntu:22.04
```
jacobkaufmann, line 1: Why the change here?

jacobkaufmann (commented):
Nice! I would like to see some unit tests alongside the changes that cover these bugs.

utp-testing/docker/circleci/Dockerfile
ogenev: There are some issues with the Debian slim image when I try to run the test app container locally. Switching to the latest ubuntu LTS works like a charm.

trin-core/src/utp/stream.rs
```diff
@@ -320,9 +320,18 @@ impl UtpListener {
...
+                            let mut result = Vec::new();
                             let mut buf = [0; BUF_SIZE];
-                            if let Err(msg) = conn.recv(&mut buf).await {
-                                error!("Unable to receive uTP FIN packet: {msg}")
+
+                            match conn.recv(&mut buf).await {
+                                Ok(bytes_read) => {
+                                    if let Some(bytes) = bytes_read {
+                                        result.extend_from_slice(&buf[..bytes]);
+                                        conn.recv_data_stream.append(&mut result);
```
jacobkaufmann, line 330: Why do we need `result` here? We can just do the following:

```rust
conn.recv_data_stream.extend_from_slice(&buf[..bytes]);
```

trin-core/src/utp/stream.rs
```diff
@@ -789,6 +798,14 @@ impl UtpStream {
 
     /// Builds the selective acknowledgement extension data for usage in packets.
     fn build_selective_ack(&self) -> Vec<u8> {
+        // Build selective ack for empty incoming buffer
+        if self.incoming_buffer.is_empty() {
+            let mut sack: Vec<u8> = vec![0u8; 4];
+            sack[0] |= 1 << 0;
```
jacobkaufmann, line 804:
```suggestion
            let mut sack: Vec<u8> = vec![0u8; 4];
            sack[0] = 1;
```

trin-core/src/utp/stream.rs
```diff
@@ -1097,6 +1113,25 @@ impl UtpStream {
...
+                        .take_while(|&seq_nr| seq_nr <= last_seq_nr);
+
+                    for seq_nr in ack_packets {
+                        if let Some(position) = self
+                            .send_window
+                            .iter()
+                            .position(|packet| packet.seq_nr() == seq_nr)
+                        {
+                            let packet = self.send_window.remove(position);
+                            self.cur_window = self.cur_window.saturating_sub(packet.len() as u32);
```
jacobkaufmann, line 1131: Should we be panicking here instead of `saturating_sub`? If we overflow, does that reflect some incorrect accounting on our part, or is this defense from some incorrect behavior from the peer?

ogenev: Thinking about it now, panicking makes more sense because it will indicate a bug in our code.

njgheorghita (approved):
:ship:

PR #423, opened 2022-08-24
Add how to use encode-key cli tool to readme & include link to contributing docs on how to create a newsfragment

### LINKED ISSUE #428: How to write release notes info is missing in contributor guidelines
Add information in contributor [guidelines](https://github.com/ethereum/trin/blob/master/docs/contributing.md) on how to create newsfragments files.

njgheorghita:
### What was wrong?
Missing documentation on how to use `encode-key` utility in `trin-cli`

Fixes #428 

### How was it fixed?
Added a blurb to docs.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-cli/README.md
+++ b/trin-cli/README.md
@@ -25,3 +25,6 @@ $ cargo run -p trin-cli -- portal_statePing --params enr:....
 ```sh
 $ cargo run -p trin-cli -- discv5_routingTableInfo --ipc /tmp/trin-jsonrpc-2.ipc
 ```
+
+### To use trin-cli to encode content keys:
+Check out the `Encode Content Keys` section of the [Getting Started docs](../docs/getting_started.md#encode-content-keys).
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -35,7 +35,7 @@ through github via pull requests.
   * `cargo fmt --all -- --check` and `cargo clippy --all -- --deny warnings` for linting checks
   * `RUSTFLAGS='-D warnings' cargo test --workspace` to run all tests
   * Run the `ethportal-peertest` harness against a locally running node. Instructions
-	can be found in [README](ethportal-peertest/README.md).
+	can be found in [README](../ethportal-peertest/README.md).
 * Pull requests **should** always be reviewed by another member of the team
   prior to being merged.
     * Obvious exceptions include very small pull requests.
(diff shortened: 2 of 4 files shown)


jacobkaufmann:
So we have a [section in the getting started docs](https://github.com/ethereum/trin/blob/8a9bb83a416aefbd0d095baa9346774d5ee97704/docs/getting_started.md#encode-content-keys) for this. Do you still think we need this change in the `trin-cli` README?

njgheorghita:
@jacobkaufmann Ahh, didn't see that. But I don't see any reason not to add this? It seems like a "natural" spot for at least a quick example of how to use the tool (aka. it's where I looked for the documentation)

jacobkaufmann:
Yeah I don't have a problem with redundancy. I do think that the header could be more clear though. It would be worth adding section for seeing which keys are available to encode as well as a specific example (e.g. block body).

njgheorghita:
`I do think that the header could be more clear though.` @jacobkaufmann can you give an example?

jacobkaufmann:
I meant something like "Generate the content key for a block header".

I prefer more general guidance that leverages the existing help messages on the CLI.

Something like "View available content keys":

```sh
cargo run -p trin-cli -- encode-key --help
```

And "See required arguments for a particular content key":

```sh
cargo run -p trin-cli -- encode-key block-header --help
```

This general guidance, that does not need to change if we add more content keys, can be paired with a specific example (e.g. block header).

njgheorghita:
Makes sense. I made this pr just bc, me, as a dev on the project, went looking for a quick example on how to use the encode key util, and it didn't exist where I expected to find it. As a dev, I didn't really need full details on the cli tool, just a quick example. Imo, the `trin-cli/README.md` is not really "public-facing" documentation - more reference for devs - especially since we're no longer pursuing full development of `trin-cli`.

But, that's not to say the the more detailed documentation isn't useful. Rather than just copy & paste, I've changed this to just be a link to the proper documentation in `getting-started.md`.

jacobkaufmann:
Totally understandable. Yeah, I'm not sure what shape that `README` should take given that `trin-cli` is quite small in scope, and unlikely to expand. Looks good to me, except that when I try to access the link to the getting started docs, I get a 404. Can you confirm the same? If we resolve that, then happy to approve.

trin-cli/README.md
```diff
@@ -25,3 +25,6 @@ $ cargo run -p trin-cli -- portal_statePing --params enr:....
 ```sh
 $ cargo run -p trin-cli -- discv5_routingTableInfo --ipc /tmp/trin-jsonrpc-2.ipc
 ```
+
+### To use trin-cli to encode content keys:
+Check out the `Encode Content Keys` section of the [Getting Started docs](/docs/getting_started.md).
```
carver, line 30: I think you should be able to deep-link to that header

njgheorghita:
@jacobkaufmann Updated the links, which are working locally on my end. Also threw in a fix for #428

jacobkaufmann (approved):
LGTM. The links give me a 404 when viewed from the PR, but they work when looking at the source branch. We should change the title of the PR before merge to include the other doc improvements.

PR #429, opened 2022-08-29
Add compare fn to MasterAccumulator

njgheorghita:
### What was wrong?
Need a function to validate that one `MasterAccumulator` contains the same validation data as another. For example, when bootstrapping & fetching `N` master accumulators from network peers, we want to validate that all of these accumulators contain the same validation data, even if they are not at the same height.

This means that all historical epoch roots should match, up to the point that both accumulators have the same number of historical epoch roots.

If one accumulator is not in the same current epoch as the other, there is no reason to validate that the header records from different current epochs match. So we return a valid check.

If the accumulators are both in the same current epoch, then we validate that both sets of header records check, up to the point that both accumulators have the same number of header records in their current epoch.

### How was it fixed?
Added `master_accumulator.validation_data_matches(other_master_accumulator) -> anyhow::Result<()>`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/types/accumulator.rs
+++ b/trin-core/src/types/accumulator.rs
@@ -1,3 +1,4 @@
+use anyhow::anyhow;
 use ethereum_types::U256;
 use ssz::Encode;
 use ssz_derive::{Decode, Encode};
@@ -42,6 +43,69 @@ impl MasterAccumulator {
                 header_records: HeaderRecordList::empty(),
             },
         }
+    }
+
+    pub fn latest_height(&self) -> u64 {
+        let historical_height = self.epoch_number() * EPOCH_SIZE as u64;
+        historical_height + self.header_record_count()
+    }
+
+    fn header_record_count(&self) -> u64 {
+        u64::try_from(self.current_epoch.header_records.len())
+            .expect("Header Record count should not overflow u64")
+    }
+
+    fn epoch_number(&self) -> u64 {
+        u64::try_from(self.historical_epochs.epochs.len())
+            .expect("Historical Epoch count should not overflow u64")
+    }
+
+    /// Compare that one master accumulator contains the same validation data as another,
+    /// regardless of each accumulator's current height.
+    pub fn validation_data_matches(&self, other: &Self) -> anyhow::Result<()> {
+        if self.historical_epochs.tree_hash_root() != other.historical_epochs.tree_hash_root() {
+            let historical_epochs = self
+                .historical_epochs
+                .epochs
+                .into_iter()
+                .zip(other.historical_epochs.epochs.into_iter());
+            for (index, (self_epoch, other_epoch)) in historical_epochs.into_iter().enumerate() {
+                if self_epoch != other_epoch {
+                    return Err(anyhow!(
+                        "Accumulator mismatch at historical epoch #{:?}.
+                        Expected historical epoch root: {:?}.
+                        Found historical epoch root: {:?}",
+                        index,
+                        self_epoch,
+                        other_epoch
+                    ));
+                }
+            }
+        }
+        // only check header records if accs. in same epoch.
+        if self.epoch_number() != other.epoch_number() {
+            return Ok(());
+        }
+        if self.current_epoch.tree_hash_root() != other.current_epoch.tree_hash_root() {
+            let header_records = self
+                .current_epoch
+                .header_records
+                .into_iter()
+                .zip(other.current_epoch.header_records.into_iter());
+            for (index, (self_hr, other_hr)) in header_records.into_iter().enumerate() {
+                if self_hr != other_hr {
+                    return Err(anyhow!(
+                        "Accumulator mismatch at header record #{:?}.
+                        Expected header record: {:?}.
+                        Found header record: {:?}",
+                        index,
+                        self_hr,
+                        other_hr
+                    ));
+                }
+            }
+        }
+        Ok(())
     }
 
     /// Update the master accumulator state with a new header.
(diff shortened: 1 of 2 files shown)


trin-core/src/types/accumulator.rs
```diff
@@ -44,6 +45,59 @@ impl MasterAccumulator {
...
     }
 
+    pub fn latest_height(&self) -> i64 {
+        let historical_height = self.epoch_number() * EPOCH_SIZE as i64;
+        historical_height + self.current_epoch.header_records.len() as i64
+    }
+
+    pub fn epoch_number(&self) -> i64 {
+        self.historical_epochs.epochs.len() as i64
+    }
```
jacobkaufmann, line 55: Why do we need `i64` here?

trin-core/src/types/accumulator.rs
```diff
@@ -44,6 +45,59 @@ impl MasterAccumulator {
         }
     }
 
+    pub fn latest_height(&self) -> i64 {
+        let historical_height = self.epoch_number() * EPOCH_SIZE as i64;
+        historical_height + self.current_epoch.header_records.len() as i64
+    }
```
jacobkaufmann, line 51: Why do we need `i64` here? Can we use `usize` or `u64`?

trin-core/src/types/accumulator.rs
```diff
@@ -44,6 +45,59 @@ impl MasterAccumulator {
...
+                return Err(anyhow!(
+                    "Accumulator mismatch at header record #{:?}.
+                    Expected header record: {:?}.
+                    Found header record: {:?}",
+                    index,
+                    self_hr,
+                    other_hr
+                ));
+            }
+        }
```
jacobkaufmann, line 97: We should be able to simplify this quite a lot by comparing common prefix slices of the historical epochs and/or header records in the latest common epoch directly.

jacobkaufmann (commented):
I don't understand some of the choices here, particularly around typing for `latest_height` and `epoch_number`.

I don't understand why `compare` returns a `Result` as opposed to a `bool`.

I also think that the name `compare` does not really capture what the function does. It seems like we are checking whether one accumulator is a valid prefix of the other accumulator.

We can also simplify the implementation of `compare` with equality checks for slices. We would lose the granularity in the error message as to which entry is the first mismatch, but I'm not sure that we need an error message in the first place re: `Result` v `bool`.

trin-core/src/types/accumulator.rs
```diff
@@ -44,6 +45,59 @@ impl MasterAccumulator {
         }
     }
 
+    pub fn latest_height(&self) -> u64 {
+        let historical_height = self.epoch_number() * EPOCH_SIZE as u64;
+        historical_height + self.current_epoch.header_records.len() as u64
```
ogenev, line 50: Is it possible to overflow here?

njgheorghita: Imo, not really. Doubt the blockchain height will ever exceed the largest `u64` - or come even remotely close. I can update this to check if you think we should

ogenev: I'm happy to leave it as it is, IMO panicking is good when it indicates a bug in our code.

carver: So just to be specific @ogenev -- your suggestion here would be to use something like `u64::try_from(...).unwrap()`, right? Yeah, that totally works for me too.

carver:
Try rebasing in master, which fixes a couple flakiness issues on the windows test.

trin-core/src/types/accumulator.rs
ogenev: > So just to be specific @ogenev -- your suggestion here would be to use something like `u64::try_from(...).unwrap()`, right? Yeah, that totally works for me too.

Not sure if we need `try_from` and `unwrap`, the code will panic with an overflow error if the sum of both values exceeds the return type of u64.

trin-core/src/types/accumulator.rs
ogenev: I guess the design decision here - comparing header by header instead of comparing a slice of headers, is to know at which header record there is a mismatch?

trin-core/src/types/accumulator.rs
```diff
@@ -44,6 +45,65 @@ impl MasterAccumulator {
...
+    /// Compare that one master accumulator contains the same validation data as another,
+    /// regardless of each accumulator's current height.
+    pub fn validation_data_matches(&self, other: &Self) -> anyhow::Result<()> {
+        let historical_epochs = self
+            .historical_epochs
+            .epochs
+            .into_iter()
+            .zip(other.historical_epochs.epochs.into_iter());
+        for (index, (self_epoch, other_epoch)) in historical_epochs.into_iter().enumerate() {
+            if self_epoch != other_epoch {
```
ogenev, line 72: I guess comparing epoch by epoch here and header by header below, instead of comparing slices of epochs/headers, is to know at which epoch/header record there is a mismatch?

trin-core/src/types/accumulator.rs
njgheorghita: Fair enough. I don't mind reverting the latest commit if you don't think the `expect()` messages would be useful?

trin-core/src/types/accumulator.rs
njgheorghita: Yup, pretty much. I've already found this debugging information helpful. I have a feeling that it will continue be helpful as we start introducing real accumulators into the network, passing them around, and using them to validate real data.

I guess the point you & @jacobkaufmann are making is that this design can result a lot of unnecessary work. Although I'm still not sure I understand what a "slice" is? But, I've added a check to compare the tree root hashes for the two data types before iterating. So we only perform the iteration to find the offending mismatch if the two root hashes don't match.

ogenev: Cool, that makes sense to me.

> Although I'm still not sure I understand what a "slice" is?
Array or vector of epochs.

trin-core/src/types/accumulator.rs
ogenev: > Fair enough. I don't mind reverting the latest commit if you don't think the `expect()` messages would be useful?

Don't have any preference really.

ogenev (approved):
Looks good to me :+1:

PR #431, opened 2022-09-05
Replace infura url with general trusted provider

njgheorghita:
### What was wrong?
We want the ability to make http requests to a "trusted provider" (usually Infura - or one of the internal geth nodes devops setup) - as well as open up a websocket connection to use `eth_subscribe` for listening to new heads. Infura allows a ws connection, but I'm still waiting to hear from devops if our internal nodes support ws. Either way, support for geth over ws can come in a following pr.

This pr is more about creating a datatype: `TrustedProvider`, that will allow on-demand access to the providers wherever it may be required (inside `jsonrpc/service.rs` and the `HeaderOracle`). 

### How was it fixed?
Adds the ability to use a choose either Infura or geth as your trusted provider. The default is still infura, so nothing will change for users. I've added the `"geth"` option with the `"experimental"` description b/c it's not quite ready for use (aka documentation) and there's a chance things change as the ability to build up your own master accumulator &/or subscribe to new headers is implemented.

### To-Do

- [x] Clean up commit history


mrferris (changes requested):
Looks good! Just a few thoughts.

A couple of thoughts I left are related to letting `structopt` and the type system handle as much CLI logic as possible to avoid CLI logic needing to be done downstream within trin, which might turn into spaghetti as we add more CLI options/arguments that may or may not play nicely with each other. 

Also, keeping CLI logic upstream allows for the CLI to be easily-unit-testable without having to run an entire trin instance. On that note, this PR could use at least a unit test for the erroneous case where `infura` is the provider and `geth_url` is provided, similar to the existing `test_http_protocol_rejects_custom_web3_ipc_path` test.

trin-core/src/utils/provider.rs
```diff
@@ -0,0 +1,120 @@
...
+        Err(_) => panic!(
+            "Must supply Infura key as environment variable, like:\n\
+            TRIN_INFURA_PROJECT_ID=\"your-key-here\" trin"
+        ),
+    }
+}
+
+fn build_infura_http_client_from_env() -> ureq::Request {
+    let infura_project_id = get_infura_project_id_from_env();
+    let infura_url = format!("{}{}", INFURA_BASE_HTTP_URL, infura_project_id);
```
ogenev, line 84:
```suggestion
    let infura_url = format!("{INFURA_BASE_HTTP_URL}{infura_project_id}");
```

trin-core/src/utils/provider.rs
```diff
@@ -0,0 +1,120 @@
...
+    ureq::post(&infura_url)
+}
+
+fn build_infura_ws_url_from_env() -> String {
+    let infura_project_id = get_infura_project_id_from_env();
+    format!("{}{}", INFURA_BASE_WS_URL, infura_project_id)
```
ogenev, line 90:
```suggestion
    format!("{INFURA_BASE_WS_URL}{infura_project_id}")
```

trin-core/src/utils/provider.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+fn get_geth_client_id_from_env() -> String {
+    match env::var("GETH_CLIENT_ID") {
```
ogenev, line 94: I see we are starting to use a lot of envs, Are there any plans to introduce a configuration file in the future, something like `config.toml` for example?

njgheorghita:
@mrferris Yup, good call on the enum. I also added tests to make sure we're parsing correctly. Re: pushing all of the logic handling responsibility to structopt, I had some trouble getting it to play nicely w/ our tests (especially the `required_if()` rule). What I kept running into is that structopt will enforce those rules [here](https://github.com/ethereum/trin/pull/431/files#diff-fa6e7eb0eae58f0f6ad91959a7c98838009df0c8bf30a2622b10ad8949fcb6fbR174), and return an error. But our testing has the pattern to validate that we panic w/ invalid args. You can see we also have some custom logic inside `TrinConfig::new_from()` handling for the ipc path args.

So it seems to me, we have some discontinuity b/w expected behavior for invalid args. Sometimes structopt's parsing recognizes a mistake and prints out a nice, custom message to the user. Other times, our custom logic handling will recognize a mistake and panic. 

Is this a bad thing? Is it necessary/worthwhile that structopt handles all of the arg validation logic (aka `TrinConfig::new_from()` should have NO validation logic)?

I'm not convinced that we need to put forth the effort to have structopt responsible for all arg validation logic. Mostly since it is a library we don't control, and there are probably some real limitations to whether it can be used the way we need to. 

But, yeah I also see that handling different types of arg validation logic in different places can get messy / confusing. Though, this is where I think tests come in handy to ensure the behavior is as expected.

Either way, I do feel like the larger question of where do/should we perform all parsing validation is beyond the scope of this pr. If we come to the agreement that it's worth updating - then I say let's track it in a separate issue.

trin-core/src/utils/provider.rs
njgheorghita: I believe there used to be an issue for it, but it might have been closed. For the time being, I'm alright with env variables - especially since I feel less likely to leak secrets by including a `.env` file in a commit than a `config.toml` file. But that's just personal preference. At some point we'll definitely want a config file

ogenev (approved):
:boat:

tests/self_peertest.rs
```diff
@@ -39,7 +40,11 @@ mod test {
             .iter(),
         )
         .unwrap();
-        let test_client_exiter = trin::run_trin(trin_config, String::new()).await.unwrap();
+        let trusted_provider = TrustedProvider {
+            http: Some(ureq::post("")),
```
mrferris, line 44: Why is a `Request` with an empty URL passed here? I'm assuming it means there's no trusted provider in this case, but why not pass `None`?

mrferris (commented):
@njgheorghita Good point on just handling validation in `new_from`, that gives us a lot more control than `structopt`.

Did another look over and everything looks good to me, just wondering about the question below.

tests/self_peertest.rs
njgheorghita: Good catch, turns out there was a situation in our jsonrpc handling that would return an error if this value is `None`. But, I realized that it doesn't make sense for this to be an `Option`. `Option` makes sense for the `ws` provider, since we don't currently support `ws` for `--trusted-provider geth`. But for `http` we will always have an http client, whether `--trusted-provider` is set to `infura` or `geth`. So I removed the Option.

PR #432, opened 2022-09-05
Expand Trin Metrics and Grafana Dashboard

mrferris:
### What was wrong?

Trin's metrics dashboard only included incoming packet counts.

### How was it fixed?

- Adds metrics & graphs for outgoing packet counts.
- Adds metrics & graphs for total storage usage, storage usage for network data, and storage capacity.
- Standardizes setup documentation to use docker and a provided Prometheus config file instead of "BYO package managers and configurations".
- Adds `create-dashboard` subcommand of `trin-cli`. This generates the prometheus & json-rpc datasources within grafana, interpolates the trin dashboard config template with the datasource UIDs, creates the dashboard, and outputs the URL to view it.
- Adds "Routing Table Health" and "Storage Usage" current-view gauges to top of dashboard.
- Removes State panels from Grafana dashboard.

### Next Steps

- Have all trin developers deploy the metrics dashboard (only 3 commands!) and gather feedback on useful metrics to add.
- The next PR that updates metrics should include a `delete-dashboard` command implementation to delete the old before calling `create-dashboard` to get the new. 
- Create & deploy a version of the dashboard to testnet's Grafana instance that shows metrics for all testnet nodes with a selector at the top to toggle which testnet node, inspired by (and maybe integrated with) the Node Exporter dashboard that is [currently deployed on Grafana](https://trin-bench.ethdevops.io/d/rYdddlPWk/node-exporter-full).

### To-Do

- [x] Clean up metrics setup docs
- [x] Add outgoing packet metrics
- [x] Add routing table metrics via JSON-RPC over HTTP (instead of prometheus as originally done)
- [x] Add storage metrics
- [x] Clean up commit history


trin-core/src/portalnet/metrics.rs
```diff
@@ -54,4 +78,24 @@ impl OverlayMetrics {
...
+
+    pub fn report_outbound_find_content(&self) {
+        self.outbound_find_content.inc();
+    }
+
+    pub fn report_outbound_offer(&self) {
+        self.outbound_offer.inc();
+    }
+
+    pub fn report_new_node_in_routing_table(&self) {
```
njgheorghita, line 98: Idk if this is the best approach? I think I'd rather see an accurate count of how many nodes are currently in my routing table. So, we should also decrement the count whenever a node is evicted. Otherwise this number will just continue to rise on-and-on every time a node is added

mrferris: @njgheorghita Good point, that should have been labelled something like "nodes seen". But "nodes connected" is much more useful, so I'm adding that as a Prometheus Gauge type instead of a counter. Currently debugging why it's telling me I have -2 connected nodes in the routing table.

pipermerriam:
FYI, excited to see this land.  I got metrics setup yesterday and found myself wanting more.

trin-cli/src/dashboard/grafana.rs
```diff
@@ -0,0 +1,100 @@
+use base64;
+use nanotemplate::template;
+use serde::Deserialize;
+use std::fs;
+use ureq;
+
+const TEMPLATE_FILENAME: &str = "./trin-cli/src/dashboard/metrics_dashboard.json.template";
```
ogenev, line 7: Is this path safe under Windows? I think we should move to use [Path](https://doc.rust-lang.org/stable/std/path/struct.Path.html) instead of strings for folder paths.

trin-cli/src/dashboard/grafana.rs
```diff
@@ -0,0 +1,100 @@
...
+pub struct GrafanaAPI {
+    basic_auth_string: String,
+    address: String,
+}
+
+impl GrafanaAPI {
+    pub fn new(username: String, password: String, address: String) -> Self {
+        let basic_auth_string = format!("{}:{}", username, password);
```
ogenev, line 15:
```suggestion
        let basic_auth_string = format!("{username}:{password}");
```

trin-core/src/jsonrpc/service.rs
```diff
@@ -287,20 +287,25 @@ fn serve_http_client(
...
-            }
-        };
-        let formatted_response = match obj.validate() {
-            Ok(_) => process_http_request(obj, trusted_provider.clone(), portal_tx.clone()),
-            Err(e) => format!("HTTP/1.1 400 BAD REQUEST\r\n\r\n{}", e).into_bytes(),
-        };
-        stream.write_all(&formatted_response).unwrap();
-        stream.flush().unwrap();
+    let mut json_objects = deser.into_iter::<JsonRequest>();
+    let obj = json_objects.next();
```
ogenev, line 291: What is the motivation for the change here? `next()` advances the iterator and returns the next value. Should we iterate over all `JsonRequest` values, not only the next value?

mrferris:
Alt Text
Ok, this is ready for review. Above is a really crappy gif showing what to expect when running `create-dashboard` and opening the URL that it outputs. Please run through the setup steps in the docs (should take like 30 seconds) and make sure it works correctly.

I updated the scope and goals a bit since initially receiving feedback. The initial description has been updated to reflect that.

I'd recommend reviewing the individual commits, mostly because the Storage commit had to make changes to a lot of files in order to pass down the `ProtocolId` into `PortalStorage`'s `StorageMetrics` struct.  The reasoning for this change:
   -  Originally the storage metrics used the same `metrics.rs` file / `OverlayMetrics` struct that `overlay_service.rs` uses, and the struct was passed to `PortalStorage` via a new `PortalStorage::set_metrics(metrics: OverlayMetrics)` method called within `overlay_service.rs`.
   -  But upon rebasing in the new `ContentStore` interface changes, I found that I could no longer easily pass the `OverlayMetrics` into `PortalStorage` because I couldn't be sure that `OverlayService`'s instance of `ContentStore` was a `PortalStorage`, and the `ContentStore` interface doesn't seem like the right place for a `set_metrics` method. 
   - So I gave `PortalStorage` its own metrics wrapper that it initializes internally (`StorageMetrics`), but it needs to know the protocol in order to use the correct label (eg `trin_content_storage_usage_kb_History`), so that info is now getting passed down through the `PortalStorage::new(config, protocol)` function.
  
Any better ideas on this are welcome.

trin-core/src/jsonrpc/service.rs
mrferris: Good question, I should have left a comment or something explaining why.

Previously if an HTTP request contained multiple json command objects, Trin would send an HTTP response for each command, which doesn't really make sense because it's multiple responses for a single request. Also, I don't know why multiple RPC commands in a single request is necessary for anything at the moment, so I made it just take and respond to the first (and likely only) JSON object in the HTTP body to keep things simple.

If we find that it's necessary to take multiple RPC commands in a single HTTP body, then this should be refactored to coalesce all of their results into a single HTTP response, rather than one for every `JsonRequest`.

docs/getting_started.md
```diff
@@ -216,17 +216,19 @@ cargo run -p trin-cli -- json-rpc portal_historyFindContent --params <enr>,<cont
...
-2. Configure your Prometheus server to target an open port to where `prometheus_exporter` will export Trin metrics.
-3. Start your Trin process with `--enable-metrics-with-url 127.0.0.1:9100` as flags
-	- The `--enable-metrics-with-url` parameter is the address for `prometheus_exporter` to export metrics to, and should be equal to the port to which your Prometheus server is targeting.
-4. Navigate to the URL on which Grafana is running in your browser (probably `localhost:3000`) and login.
-5. Add your Prometheus server as a Data Source, using the URL on which your Prometheus server is running.
-6. Use the "Import" dashboard feature in Grafana to create your Trin dashboard, and copy and paste `metrics_dashboard.json` as the template.
-7. You should now be able to see live metrics reporting for your Trin node.
+1. Install Docker.
+2. Run Prometheus: `docker run -d -p 9090:9090 -v /absolute/path/to/trin/docs/metrics_config:/etc/prometheus prom/prometheus`. Set the correct absolute path to your copy of Trin's `docs/metrics_config/` directory so that the container's `/etc/prometheus/prometheus.yml`.
```
njgheorghita, line 220: I'm not sure that the end of this sentence makes sense? Or i'm just confused

docs/getting_started.md
```diff
@@ -216,17 +216,19 @@ cargo run -p trin-cli -- json-rpc portal_historyFindContent --params <enr>,<cont
...
+3. Run Grafana: `docker run -d -p 3000:3000 -e "GF_INSTALL_PLUGINS=yesoreyeram-infinity-datasource" grafana/grafana:latest`.
+4. Start your Trin process with `--enable-metrics-with-url 127.0.0.1:9100` and `--web3-transport http`.
```
njgheorghita, line 222: If you get rid of the "and" here it makes for an easier copy/paste

docs/getting_started.md
```diff
@@ -216,17 +216,19 @@ cargo run -p trin-cli -- json-rpc portal_historyFindContent --params <enr>,<cont
...
+	- The `--enable-metrics-with-url` parameter is the address that Trin exports metrics to, and should be equal to the port to which your Prometheus server is targeting at the bottom of `metrics_config/prometheus.yml`. 
+    - The `--web-transport http` will allow Grafana to request routing table information from Trin via JSON-RPC over HTTP.
+5. From the root of the Trin repo, run `cargo run -p trin-cli -- create-dashboard`. If you used different ports than detailed in the above steps, or you are not using docker, then this command's defaults will not work. Run the command with the `-h` flag to see how to provide non-default addresses or credentials.
```
njgheorghita, line 225: It makes a bit more sense to me to do this step before step 4. Unless there's a need to do 4 first that I'm missing.

src/lib.rs
```diff
@@ -58,11 +58,17 @@ pub async fn run_trin(
...
 
-    let storage_config =
-        PortalStorageConfig::new(trin_config.kb.into(), discovery.local_enr().node_id());
+    let storage_config = PortalStorageConfig::new(
+        trin_config.kb.into(),
+        discovery.local_enr().node_id(),
+        trin_config.enable_metrics_with_url.is_some(),
+    );
+
+    let mut oracle_storage_config = storage_config.clone();
```
njgheorghita, line 67: There might be a better way to work around this... but for now I'd just add a comment explaining why we're creating this extra copy.

src/lib.rs
```diff
@@ -58,11 +58,17 @@ pub async fn run_trin(
         setup_temp_dir();
     }
 
-    let storage_config =
-        PortalStorageConfig::new(trin_config.kb.into(), discovery.local_enr().node_id());
+    let storage_config = PortalStorageConfig::new(
+        trin_config.kb.into(),
+        discovery.local_enr().node_id(),
+        trin_config.enable_metrics_with_url.is_some(),
```
njgheorghita, line 64: At this point, is it worth just passing in `trin_config` directly?

trin-core/src/jsonrpc/service.rs
njgheorghita: I'd add in a brief comment explaining this just in case we do need to revisit down the road

njgheorghita (approved):
Nice! It was honestly surprisingly flawless getting these metrics up & running. Not that I doubted your code, i'm just used to at least some kind of bug tripping me up when getting these things going. But it's a definite improvement over the previous system.

ogenev (approved):
Looks like a nice improvement!

src/lib.rs
mrferris: I'd say yes if `PortalStorageConfig` was using 1.) all of the fields of `TrinConfig` and 2.) only the fields of `TrinConfig`, otherwise I think that would make this `new` function much more opaque as far as what's being passed, and much less flexible for future changes.

docs/getting_started.md
mrferris: @njgheorghita The main reason for the order here is that this way you see metrics when you open the dashboard, whereas if you open the dashboard before you've run trin you see an empty dashboard which is ugly. That order doesn't matter too much though.

PR #433, opened 2022-09-06
Resend lost uTP packets on timeout

### LINKED ISSUE #416: Resend lost uTP packets on timeout
We should add a mechanism to resend lost packets in the core uTP protocol caused by a timeout.

ogenev:
### What was wrong?
Closes #416 

A feature to resend lost uTP packets caused by a timeout was missing.

### How was it fixed?

If we don't receive acknowledgment for the sent uTP packet from remote, we mark the packet as lost and try to resend it.

### To-Do

- [x] Clean up commit history


njgheorghita (approved):
:ship:

PR #434, opened 2022-09-08
uTP: Do not return error when handling RESET packet

### LINKED ISSUE #414: Unable to receive uTP RESET packet
It seems like the longer I run the bridge node, the more of these messages I get:

```
trin_core::utp::stream: Unable to receive uTP RESET packet: Connection reset by remote peer
```

I haven't investigated it yet, just making a note to check it out.

ogenev:
### What was wrong?
Closes #414 

A bug caused a misleading error message when handling uTP RESET packet.

### How was it fixed?
Return `Ok(None)` instead of `Err(..)` when handling the RESET packet.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/414.fixed.md
+++ b/newsfragments/414.fixed.md
@@ -1,0 +1,1 @@
+uTP: Do not log error when handling RESET packet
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
🚢

mrferris (approved):
👍

PR #436, opened 2022-09-12
Move from `log` to `tracing`

ogenev:
### What was wrong?

The [tracing](https://docs.rs/tracing) crate is the state-of-the-art framework for instrumenting `tokio` programs to collect structured, event-based diagnostic information.

It is not necessary to import as a dependency the `log` crate, because tracing is built on top of it and includes all necessary event macros plus more features.

### How was it fixed?

Move from `log` to `tracing` events.

Bonus: Clean unused crates.

### To-Do

- [x] Clean up commit history


trin-core/Cargo.toml
```diff
@@ -77,7 +74,8 @@ version = "0.26.3"
 features = ["bundled"]
 
 [dev-dependencies]
+env_logger = "*"
```
jacobkaufmann, line 77: Why the wildcard here and elsewhere? I would prefer a more granular version.

jacobkaufmann (approved):
A step in the right direction for our logging 💯

trin-core/Cargo.toml
mrferris: agreed, this allows for any version (not just latest), which might bring unexpected changes

trin-core/Cargo.toml
```diff
@@ -56,8 +54,7 @@ threadpool = "1.8.1"
 tokio = { version = "1.8.0", features = ["full"] }
 tokio-test = "0.4.2"
 tokio-util = "0.6.8"
-tracing = "0.1.26"
-tracing-futures = "0.2.5"
```
mrferris, line 60: I'm surprised that `cargo clippy` or something doesn't tell us when we have completely unused crates, or was this being used somewhere where I'm missing?

mrferris (approved):
👍

trin-core/Cargo.toml
ogenev: Unfortunately, `cargo clippy` doesn't handle this for us but there is [cargo udeps](https://github.com/est31/cargo-udeps) which aims to solve this issue.

trin-core/Cargo.toml
ogenev: > Why the wildcard here and elsewhere? I would prefer a more granular version.

It is because `env_logger` is required only as a dependency of `test-log` and we don't really care about the exact version, but I agree it is more convenient to set a fixed version number.

PR #437, opened 2022-09-12
Follow blockchain head over http/ws

njgheorghita:
### What was wrong?
Introduce `Bridge` type that will facilitate requesting/receiving data from a trusted provider and pumping it into the network. Currently, all it does is listen for new headers to build up the local macc, and then offer them to the network.

Thought about combining bridge functionality w/ header oracle, but eventually ended up on separating b/c ....
- non-bridge nodes will still need to listen to header gossip
- bridge nodes will listen to header gossip & suck new data from trusted providers

This is sort of difficult to test. We don't want to introduce a reliance on infura in our tests. I feel the same way about our geth clients. Most of this logic has well-unit-tested components. "bridge" mode is still explicitly experimental. I could be convinced to use our geth clients in CI to test this logic, but I'm not sure it's necessary. Idk, happy to field other opinions on this. 

### How was it fixed?
- Introduced `Bridge` type.
- Reorganized `HeaderOracle` into its own file.

### To-Do

- [ ] Clean up commit history


njgheorghita:
Closing this for now since this functionality will no longer be needed to keep your macc up to date, since macc's will be frozen at the merge block

PR #441, opened 2022-09-13
feat: look up unknown ENRs in node address cache

jacobkaufmann:
### What was wrong?

For remote peers who are not in the subnetwork routing table, we rely on the `Discv5` routing table. However, the limited capacity of the `Discv5` routing table is shared among all active subnetworks.

### How was it fixed?

Look up ENR for unrecognized remote peer in the node address cache. The node address cache contains the ENR and observed socket address from the latest established session for a given `NodeId`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -1024,18 +1024,13 @@ where
                 }
             }
         } else {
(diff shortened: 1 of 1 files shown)


ogenev (approved):
Nice :+1:

PR #442, opened 2022-09-14
feat: Adds local check for content in lookup_content

mrferris:
### What was wrong?

Initiating a `RecursiveFindContent` via RPC does not first check whether the content is stored locally.

### How was it fixed?

`lookup_content` will immediately return content if it is stored locally. If the data is not stored locally or we get a storage error, proceed with network lookup.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-history/src/jsonrpc.rs
+++ b/trin-history/src/jsonrpc.rs
@@ -4,6 +4,7 @@ use rand::seq::SliceRandom;
 use serde_json::{json, Value};
 use ssz::Decode;
 use tokio::sync::mpsc;
+use tracing::error;
 
 use crate::network::HistoryNetwork;
 use trin_core::{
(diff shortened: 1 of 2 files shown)


jacobkaufmann:
Should this logic be in `lookup_content`, or should it be in the JSON-RPC request handling? If it goes in `lookup_content`, then we do not preserve a way to perform the content lookup regardless of whether we possess the content or not.

jacobkaufmann (commented):
Did you intentionally not implement for `trin-state`?

njgheorghita (approved):
Lookin' good to me - after adding it to trin-state à la @jacobkaufmann

jacobkaufmann (approved):
LGTM. And to me this is more of a feature than a bug fix, because the JSON-RPC endpoint is not defined in a way that would make the existing behavior "incorrect". Given that, I would suggest that we change the title of the PR/commit.

PR #443, opened 2022-09-15
bugfix: Removes quartering of capacity

mrferris:
### What was wrong?

`PortalStorage::setup_config` was being used in the main trin binary to create a `PortalStorageConfig`, but the `setup_config` method was dividing the storage capacity by 4. This was originally done for testing/debugging purposes but seems to have moved to the main-execution-path undetected.

### How was it fixed?

Removed the `setup_config` method, which was simply dividing the capacity by 4 and then returning a Result even though it couldn't fail. Replaced this method with direct calls to the method it had been calling, `PortalStorageConfig::new`.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/src/bin/seed-database.rs
+++ b/src/bin/seed-database.rs
@@ -9,7 +9,7 @@ use tracing::debug;
 
 use trin_core::{
     portalnet::{
-        storage::PortalStorage,
+        storage::{PortalStorage, PortalStorageConfig},
         types::content_key::{BlockHeader, HistoryContentKey, IdentityContentKey},
     },
     types::header::Header,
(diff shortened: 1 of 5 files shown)


ogenev (approved):
LGTM :+1:

jacobkaufmann (approved):
LGTM.

Since this quarter configuration was there deliberately, it doesn't seem like "bugfix" is the appropriate label though.

mrferris:
@jacobkaufmann What do you think would be a more appropriate label? 

It was deliberately put somewhere at some time, but I don't think here and definitely not at the time that it was merged.

jacobkaufmann:
It could go without a label, but it doesn't matter too much. My point was only that, when it was [introduced](https://github.com/ethereum/trin/pull/230/files#diff-3cb8583ae71ee50dba75cf78cbb3d4449cf61ba558411eba0c28a58e53d0e031R422), it appears to have been done deliberately based on the comment above assignment. Looks like it was transferred from the old `generate-data` binary. That may have been unintended. But this was probably not even worthing raising in the first place. No need to respond to this comment. Do whatever seems appropriate.

PR #444, opened 2022-09-15
Validate pre-merge headers via master acc.

### LINKED ISSUE #440: Lengthy accumulator `validation_data_matches_valid` test runs
Introduced in #429, the `validation_data_matches_valid` unit tests run for an excessively long time (longer than 60 seconds) as shown by the [logs emitted in the CI test runs](https://app.circleci.com/pipelines/github/ethereum/trin/2027/workflows/b7112523-b094-4df4-9eec-c67b8304d40a/jobs/4307). As far as I can tell, we are testing accumulators that are impractically large for unit tests (~819,200 shown below).

https://github.com/ethereum/trin/blob/d02340e3cf4f7d62cac47210de65050ee630c8a8/trin-core/src/types/accumulator.rs#L332-L336

I understand that we want to test accumulators whose length is on the epoch boundary. A configurable epoch size (e.g. via an [associated constant](https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants)) would allow us to achieve this with much smaller epoch lengths. Such a change might require logic to check that we only compare accumulators whose epoch lengths are equal.

njgheorghita:
### What was wrong?
Apologies for the massive pr. I did initially attempt to break this functionality up into multiple, smaller prs. But the pieces are all so intertwined, that I found it pretty challenging.

This pr updates the `HeaderOracle` to use the default master accumulator to validate all headers pre-merge. I generated this master accumulator, and accompanying epoch accumulators myself. And while they're tested & and do sufficiently validate / invalidate mainnet data, I've also done further testing locally, time will tell if they are accurate... (aka somebody can generate a master acc & we'll compare).

### How was it fixed?
There's a lot of little fixes to this pr. I'm more than happy to hop on a zoom chat anytime to discuss the decisions made rather than bloating this pr with excessive comments/discussion.

- removes master acc bootstrap process, and simply loads default master acc from disk
- removes accumulator db in portal storage, since we no longer need to respond to network requests
- adds cli option for specifying path to custom master acc, but functionality is not yet supported: 
- removes master acc "compare" fn since it's no longer needed
- improve master acc testing
- remove `portal_historySampleMasterAccumulator` jsonrpc functionality
- includes some raw master acc & a few epoch accs as assets for testing
- fixes #440 
- added `--no-capture` flag to ci test run

### To-Do

- [x] Clean up commit history


njgheorghita:
Dealing with a flaky `win-build`. Spoke with @mrferris and he said he recognized it as one of the earlier `win-build` bugs we were dealing with. Since debugging the windows build isn't as high of a priority as getting this functionality over the line, let's move forward with reviewing this pr for now. If the flakiness continues I'll open an issue. It's also very possible that I actually introduced a bug in here, so I'd keep that in mind as you review

.circleci/config.yml
```diff
@@ -44,7 +44,7 @@ jobs:
                 command: cargo build --workspace
             - run:
                 name: Test Trin workspace
-                command: cargo test --workspace
+                command: cargo test --workspace -- --nocapture
```
carver, line 47: Are you good with dropping this change, after the Discord conversation?

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -22,23 +22,13 @@ use trin_core::{
 const DATA_RADIUS: Distance = Distance::MAX;
 /// Default enr seq value
 const ENR_SEQ: &str = "1";
-/// Default history header content key presuming chain ID 3
+/// History header content key & value for merge block on mainnet
+/// Use merge block here so that we can use the default master accumulator for validation
+/// rather than needing to go lookup an epoch accumulator.
```
carver, line 27: Hm, don't we also want to test that validation works when looking up an old epoch? Even if we do something like drop the old epoch into the local data store, first, I'd feel better that we're testing that code path.

carver:
> Since debugging the windows build isn't as high of a priority as getting this functionality over the line, let's move forward with reviewing this pr for now. If the flakiness continues I'll open an issue. It's also very possible that I actually introduced a bug in here, so I'd keep that in mind as you review

Yeah, I'm inclined to not ignore this failure. It's a real PITA to merge a flaky test into master and start causing problems with all PRs. Master has been clean for a little while, let's keep it that way.

There are some changes in here that could certainly be related, so let's see if we can figure it out.

ethportal-peertest/src/lib.rs
```diff
@@ -76,7 +75,28 @@ impl Peertest {
...
+pub async fn launch_node(trin_config: TrinConfig) -> anyhow::Result<PeertestNode> {
+    let web3_ipc_path = trin_config.web3_ipc_path.clone();
+    let server = setup_mock_trusted_http_server();
+    let mock_trusted_provider = TrustedProvider {
+        http: ureq::post(&server.url("/")),
+        ws: None,
+    };
+    let exiter = trin::run_trin(trin_config, mock_trusted_provider)
+        .await
+        .unwrap();
```
carver, line 87: The short answer is that for some reason, panic handling is broken (especially on Windows?). So every `unwrap()` is a chance for the `win-build` to fail, and fail silently.

In order to [find and fix the last one of these](https://github.com/ethereum/trin/pull/411/commits/cec625ab242ea803a141e4a0db08017824b6fd68), I inserted `error!()` logs before every unwrap, to find the panic. It seems like a good place to start with that is any `unwrap`/`expect`/etc that is new, or moved in this PR.

carver (changes requested):
👎🏻 on merging a failing test. I can only do a partial review right now, but wanted to say that part before I take a break and come back to it.

jacobkaufmann (commented):
I agree with @carver on the failing test. I also have some high level comments.

- I don't see why we need to support a custom master accumulator at this time, so I propose that we remove all of that logic to reduce the size of the PR.
- Let's pass the master accumulator as an argument into the constructor for `PortalStorage` instead of looking for it in the file system post-construction.

.circleci/config.yml
njgheorghita: Tbh, I don't mind dropping it. But I'm a bit confused. `"oh yeah, I never figured out how to get peertest to capture on success and display on failure."` ... I'm interpreting this as meaning that, yes basically peertest doesn't capture all of the output as a normal test would? Shouldn't this be an argument for leaving the flag in? So that the missing output will be displayed in ci. Or am I misunderstanding here

ethportal-peertest/src/jsonrpc.rs
njgheorghita: Those scenarios are all unit-tested in `accumulator.rs`. I didn't feel like they needed to be replicated in `peertest` - where the test (imo) is more related to making sure the jsonrpc request/response happens successfully, and less about covering the edge cases.

njgheorghita:
@jacobkaufmann 
> I don't see why we need to support a custom master accumulator at this time, so I propose that we remove all of that logic to reduce the size of the PR.

Sounds good. We're in a bit of a grey area right now with the spec - at least until it's merged. But it does state that we should allow users the ability to choose a custom master acc, but via loading from disk. I removed all the logic for requesting a custom master acc from the network - but left the pieces in place to support loading a custom master acc from disk, to be handled in a later pr - and tracked in #449 

> Let's pass the master accumulator as an argument into the constructor for PortalStorage instead of looking for it in the file system post-construction.
Good call. Turns out, now that we no longer needed to serve master accumulator messages over the network, we can remove the concept of an accumulator from `PortalStorage` entirely.

njgheorghita:
@carver I ended up removing a fair bit of logic - thanks to @jacobkaufmann 's comments. 5 successful ci runs in a row feels sufficiently un-flaky to me. I'm hoping that implies that I removed the source of the flakiness, but if the error pops back up before this gets merged, I'll dig back into it.

trin-core/src/cli.rs
```diff
@@ -129,11 +130,13 @@ pub struct TrinConfig {
...
+    // todo: support custom master accs
     #[structopt(
-        long = "bridge",
-        help = "Activate to run this node in bridge mode (experimental)."
+        long = "master-accumulator-path",
+        help = "Path to master accumulator for validation (not currently supported)",
+        default_value(DEFAULT_MASTER_ACC_PATH)
     )]
-    pub bridge: bool,
+    pub master_acc_path: String,
```
ogenev, line 139: I think it is safer to convert to `PathBuf` as soon as possible, using:

 #[structopt(parse(from_os_str))]
    input: PathBuf,

trin-core/src/types/accumulator.rs
```diff
@@ -108,9 +99,16 @@ impl MasterAccumulator {
     /// preceding epoch's merkle root to historical epochs.
     // as defined in:
     // https://github.com/ethereum/portal-network-specs/blob/e807eb09d2859016e25b976f082735d3aceceb8e/history-network.md#the-header-accumulator
-    fn update_accumulator(&mut self, new_block_header: &Header) {
+    fn update_accumulator(&mut self, new_header: &Header) {
+        if new_header.number > MERGE_BLOCK_NUMBER {
+            panic!("Invalid master acc update: Header is post-merge.");
```
ogenev, line 104: I think returning an error here nd below makes more sense because trying to update an accumulator with the wrong header number is not unexpected behavior and will not signal a bug in our code.

ogenev (approved):
Looks good to me :boat:

trin-core/src/types/accumulator.rs
```diff
@@ -39,12 +47,24 @@ impl MasterAccumulator {
...
+    /// Number of the next block to be included in the accumulator
+    pub fn next_header_height(&self) -> u64 {
+        match self.height() {
+            Some(val) => val + 1,
+            None => 0,
+        }
     }
 
-    fn header_record_count(&self) -> u64 {
+    fn current_header_count(&self) -> u64 {
```
jacobkaufmann, line 67:
```suggestion
    fn current_epoch_len(&self) -> u64 {
```

trin-core/src/types/accumulator.rs
```diff
@@ -149,6 +147,67 @@ impl MasterAccumulator {
...
+/// accumulator.
+pub async fn validate_pre_merge_header(
+    header: &Header,
+    master_acc: &MasterAccumulator,
+    history_jsonrpc_tx: mpsc::UnboundedSender<HistoryJsonRpcRequest>,
+) -> anyhow::Result<bool> {
+    if header.number > MERGE_BLOCK_NUMBER {
+        return Err(anyhow!("Unable to validate post-merge block."));
+    }
+    if header.number >= master_acc.next_header_height() {
```
jacobkaufmann, line 162:
```suggestion
    if header.number > master_acc.height() {
```

jacobkaufmann (commented):
Looks good overall. Left some minor questions/suggestions. My biggest comment is that I would like to use a concrete error type instead of `anyhow` for the accumulator and validation. This would make it more straightforward to emit metrics for specific errors.

trin-core/src/cli.rs
```diff
@@ -173,6 +176,10 @@ impl TrinConfig {
     {
         let config = Self::from_iter(args);
 
+        if config.master_acc_path != DEFAULT_MASTER_ACC_PATH {
+            panic!("Custom master accumulators are not yet supported.")
+        }
+
```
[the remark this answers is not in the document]
njgheorghita: It's being tracked in an issue (#449) - supporting this is part of the spec. I plan to implement it as soon as this pr gets merged. This is basically a temporary measure to accommodate `TrinConfig`

trin-core/src/types/validation.rs
```diff
@@ -133,12 +72,49 @@ impl HeaderOracle {
...
+                    }
+                }
+            }
+        }
+        // either header is post-merge or there was an error trying to validate it via chain
+        // history network, so we fallback to infura
+        let trusted_hash = self.get_hash_at_height(header.number)?;
+        match trusted_hash == header.hash() {
+            true => Ok(()),
+            false => Err(anyhow!(
```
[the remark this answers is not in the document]
njgheorghita: I get the sense that you don't like matching on bools? Personally, I lean towards `match` in as many cases as possible. It's cleaner (imo) - with bools, it's one less line. I know, syntax nitpick. But if you want to update our style guide to recommend using an `if` statement for all bools - we can continue this syntax discussion in that pr.

trin-core/src/types/accumulator.rs
```diff
@@ -149,6 +147,67 @@ impl MasterAccumulator {
...
 
+/// Validation function for pre merge headers that will use the master accumulator to validate
+/// whether or not the given header is canonical. Returns true if header is validated, false if
+/// header is invalidated, or an err if it is unable to validate the header using the master
+/// accumulator.
+pub async fn validate_pre_merge_header(
+    header: &Header,
+    master_acc: &MasterAccumulator,
+    history_jsonrpc_tx: mpsc::UnboundedSender<HistoryJsonRpcRequest>,
+) -> anyhow::Result<bool> {
```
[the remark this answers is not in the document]
njgheorghita: good call, this decision was leftover from when we stored `MasterAccumulator`s in `PortalStorage` - mainly b/c there wasn't access to the `history_jsonrpc_tx` channel to lookup epoch accumulators.

trin-core/src/types/accumulator.rs
```diff
@@ -54,52 +74,23 @@ impl MasterAccumulator {
...
-                    ));
-                }
-            }
-        }
-        Ok(())
+    fn historical_header_count(&self) -> u64 {
+        self.epoch_number() * EPOCH_SIZE as u64
+    }
+
+    fn get_epoch_index(&self, header: &Header) -> u64 {
```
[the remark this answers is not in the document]
njgheorghita: I went with `get_epoch_index_of_header()`. Imo, `epoch_of_header()` implies that you're going to return the actual `EpochAccumulator` - so I left "index" in to help distinguish that you're just getting the index of the header

trin-core/src/types/accumulator.rs
njgheorghita: Imo, since we also have the method `historical_header_count()`, I actually prefer `current_header_count()` here. They seem to complement each other nicely

trin-core/src/types/accumulator.rs
njgheorghita: The problem is that `height()` returns an `Option` which makes this evaluation unnecessarily messy

njgheorghita:
@jacobkaufmann 
> My biggest comment is that I would like to use a concrete error type instead of anyhow for the accumulator and validation. This would make it more straightforward to emit metrics for specific errors.

I'd agree, that makes sense. But seems premature and not a high priority right now. Once we get metrics supported here, the error type can be easily updated then.

trin-core/src/types/validation.rs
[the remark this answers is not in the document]
jacobkaufmann: Yea it is a matter of preference and only a nit. When there are only two conditions (i.e. is true or is not true), I think it is easier to read `if condition { .. } else { .. }` than the equivalent `match`. Going to resolve this comment thread.

trin-core/src/types/accumulator.rs
```diff
@@ -123,7 +134,7 @@ impl MasterAccumulator {
         };
 
         // check if the epoch accumulator is full
-        if self.current_epoch.header_records.len() == EPOCH_SIZE {
+        if self.next_header_height() > 0 && self.next_header_height() % EPOCH_SIZE as u64 == 0 {
```
jacobkaufmann, line 137: Why do we need the more complicated condition?

njgheorghita: Can't remember exactly which one, but there was an edge case that wasn't being handled properly in the `validation_data_matches_valid()` test

trin-core/src/types/accumulator.rs
```diff
@@ -47,6 +49,15 @@ impl MasterAccumulator {
         Self::default()
     }
 
+    /// Load default trusted master acc
+    pub fn try_from_file(master_acc_path: PathBuf) -> anyhow::Result<MasterAccumulator> {
+        let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+        path.push(master_acc_path);
```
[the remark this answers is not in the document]
njgheorghita: `CARGO_MANIFEST_DIR` is a defined environment variable, set by cargo, which will always give us a path. Now, maybe `master_acc_path` is an invalid path... but that would cause an error on line 56. I don't see a side effect in this case that occurs before line 56 - where it might have trouble reading an invalid path, in which case we return an error.

trin-core/src/types/accumulator.rs
jacobkaufmann: I interpret `current_header_count` the same way I interpret `height`, the total number of headers in the accumulator.

If you prefer the `header_count` naming as opposed to `len`, what about `current_epoch_header_count` and `historical_epochs_header_count`?

trin-core/src/types/accumulator.rs
jacobkaufmann: Which test was failing? Looking at it more closely, this logic is incorrect.

- Assume `EPOCH_SIZE` is equal to `8192`.
- Suppose `next_header_height` is equal to `8192 * 2` such that we are about to fill the current (second) epoch.
- The condition returns `true` because `8192 * 2 > 0` and `(8192 * 2) % 8192` is equal to `0`.

njgheorghita: Good catch, you're right that the extra logic was unnecessary. But I believe the extra logic was correct. If we are trying to update the accumulator with block #`16384` (8192* 2) (i.e. `next_header_height == 16384`), this means that we already have `16384` total header records (blocks 0 -16383). In which case our current epoch accumulator is full and we want to clear it, update our historical headers and then push block #`16384` as the first record in the new current epoch. Regardless, that extra logic is redundant and I've reverted it back to the simpler check.

jacobkaufmann: I forgot that height starts at zero 🤦‍♂️

.circleci/config.yml
njgheorghita: Leaving this change in for now, but happy to remove it in a future pr if i'm misunderstanding the situation

njgheorghita:
Going ahead with the merge since I'd like to get this change up on the testnets. I've addressed every comment here, but if there are any other questions / changes requested, just ping me and i'll add them in a follow-up pr.

PR #447, opened 2022-09-20
feat: offer content from query to closest nodes

jacobkaufmann:
### What was wrong?

Content found via find content queries is not disseminated to other peers who may store it.

### How was it fixed?

Implement "poke" mechanism to offer the content found by a find content query to the nodes involved in the query who did not possess the desired content and whose radius is sufficiently large.

Also addressed:

- Fix `PartialEq` implementation for `Node` so that non-ENR changes are recognized by the routing table.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/447.added.md
+++ b/newsfragments/447.added.md
@@ -1,0 +1,1 @@
+Disseminate content found via find content queries to other peers who may store it (i.e. poke content).
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -47,7 +47,7 @@ use crate::{
             distance::{Distance, Metric},
             messages::{
                 Accept, ByteList, Content, CustomPayload, FindContent, FindNodes, Message, Nodes,
-                Offer, Ping, Pong, ProtocolId, Request, Response, SszEnr,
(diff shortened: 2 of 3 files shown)


jacobkaufmann:
I have limited the maximum number of poke offers to the `query_parallelism` configuration parameter. The [current default value](https://github.com/ethereum/trin/blob/9f87e9b829d096336d5f5582a4adc436a4459432/trin-core/src/portalnet/overlay.rs#L78) is 3. I'm not sure whether this limit should be higher, or whether this limit should not exist at all. The Kademlia paper specifies that the found content only be offered to the closest node. My interpretation of the portal spec is that the content should be offered to all peers with sufficient radius.

carver:
Yeah, this poke thing isn't in Kademlia at all IIRC. I would say to offer it to everyone you asked the content for along the way, who has a radius that is interested.

jacobkaufmann:
@carver from [Kademlia section 2.3](https://pdos.csail.mit.edu/~petar/papers/maymounkov-kademlia-lncs.pdf): "For caching purposes, once a lookup succeeds, the requesting node stores the (key, value) pair at the closest node it observed to the key that did not return the value."

I will change to offering found content to all encountered nodes with sufficient radius. We can always limit this spread in the future if desired.

mrferris (approved):

PR #451, opened 2022-09-26
Remove chain id from history network message

njgheorghita:
### What was wrong?
`chain_id` field has just been removed from history network keys in the latest spec update.

### How was it fixed?
Removed the chain id field from history network content key type & test values.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -203,7 +203,7 @@ cargo run -p trin-cli -- encode-key block-header -h
 Example:
 
 ```sh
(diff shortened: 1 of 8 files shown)


jacobkaufmann:
Waiting on test vectors to be confirmed/merged [here](https://github.com/ethereum/portal-network-specs/pull/167).

jacobkaufmann (approved):
Looks like the test vectors were confirmed in the portal spec.

nit: A more accurate PR title would be something like "remove chain ID from history content keys"

PR #452, opened 2022-09-26
cleanup: remove Option for non-nullable param

jacobkaufmann:
### What was wrong?

Method param `source` is non-nullable but has type `Option<&NodeId>`.

### How was it fixed?

Change param `source` to have type `&NodeId`.

### To-Do

- [x] Clean up commit history


ogenev (approved):
:+1:

PR #455, opened 2022-10-03
Validate pre-merge headers via master acc.

njgheorghita:
### What was wrong?
No validation for epoch accumulators received over network.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/types/accumulator.rs
+++ b/trin-core/src/types/accumulator.rs
@@ -38,12 +38,10 @@ type HeaderRecordList = VariableList<HeaderRecord, typenum::U131072>;
 /// Verifies canonical-ness of a given header.
 #[derive(Clone, Debug, Decode, Encode, Eq, PartialEq, Deserialize, Serialize, TreeHash)]
 pub struct MasterAccumulator {
-    historical_epochs: HistoricalEpochs,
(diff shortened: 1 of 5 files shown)


trin-core/src/types/accumulator.rs
```diff
@@ -110,6 +108,10 @@ impl MasterAccumulator {
     /// preceding epoch's merkle root to historical epochs.
     // as defined in:
     // https://github.com/ethereum/portal-network-specs/blob/e807eb09d2859016e25b976f082735d3aceceb8e/history-network.md#the-header-accumulator
+    // Allow dead code since this functionality is no longer needed in the code base, now that we
+    // use a pre-merge, frozen master accumulator. However, somebody might find this code useful to
```
ogenev, line 112: I'm more in favor of removing this. We can always revive it from history if needed.

ogenev (approved):
LGTM :+1:

trin-core/src/types/accumulator.rs
njgheorghita: Since we still use this multiple times in the testing, and I'd rather not remove those tests, I converted this functionality to a test utility function.

PR #456, opened 2022-10-03
Remove master accumulator content key

njgheorghita:
### What was wrong?
Master Accumulator content keys are no longer needed now that they are no longer a network-supported content type according to the latest spec update.

### To-Do

- [x] Clean up commit history


ogenev (approved):
:100:

jacobkaufmann (approved):

PR #457, opened 2022-10-03
Add support for eth_getBlockByNumber for pre-merge blocks.

njgheorghita:
### What was wrong?
Adds jsonrpc support for the `eth_getBlockByNumber` endpoint, but only for pre-merge blocks, where we can look up the header hash via the master accumulator, before retrieving the header from the chain history network.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -134,6 +134,7 @@ pub async fn run_trin(
         portal_jsonrpc_rx,
         state_jsonrpc_tx,
         history_jsonrpc_tx,
+        header_oracle: header_oracle.clone(),
     };
 
     tokio::spawn(rpc_handler.process_jsonrpc_requests());
--- a/trin-core/src/jsonrpc/eth.rs
+++ b/trin-core/src/jsonrpc/eth.rs
@@ -1,15 +1,18 @@
+use std::sync::Arc;
+
 use anyhow::anyhow;
 use serde_json::{json, Value};
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, RwLock};
 
 use crate::{
     jsonrpc::{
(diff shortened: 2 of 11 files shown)


src/lib.rs
```diff
@@ -134,6 +134,7 @@ pub async fn run_trin(
         portal_jsonrpc_rx,
         state_jsonrpc_tx,
         history_jsonrpc_tx,
```
njgheorghita, line 136: It seems like there's a potential refactor here where we just pass in the header oracle and use the tx channels on that, rather than passing in each channel specifically, but I'll save that for a follow-up pr to avoid cluttering this pr

trin-core/src/jsonrpc/eth.rs
```diff
@@ -43,3 +46,31 @@ pub async fn get_block_by_hash(
...
+    let block_hash = header_oracle
+        .read()
+        .await
+        .get_hash_at_height(params.block_number)
+        .await?;
+    let params = Params::Array(vec![
+        Value::String(format!("{:?}", block_hash)),
+        Value::Bool(false),
+    ]);
+    get_block_by_hash(
```
ogenev, line 71: I guess we are currently not returning transactions and this makes the transaction details flag not have any effect on the result from this endpoint?

njgheorghita: Yup, but good catch that this would lead to unintended behavior for users over jsonrpc. So, I added a check to return an error if users request full transactions, to prevent this behavior for now. It can easily be removed down the road when we support full transactions, but I don't consider implementing that support a priority atm.

ogenev (approved):
Looks solid to me :+1:

PR #458, opened 2022-10-04
Fix FindContent uTP stream issues

ogenev:
### To-Do

- [ ] Clean up commit history


ogenev:
> ship - Would it be possible to get a test added to the `utp-testing` module that would cover the edge cases fixed in this pr?

It is possible but maybe those tests would be more suitable for an interop testing framework. We can think about this and add them in a separate PR as it is not very trivial work.

PR #459, opened 2022-10-07
Fix missing master accumulator in dockerfile

njgheorghita:
### What was wrong?
Default master accumulator wasn't available in final docker image, causing docker images to fail when launching trin.

### How was it fixed?
Explicitly copy over the default master accumulator to the expected location. 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,10 +25,14 @@ RUN cargo build --all --release
 # final base
 FROM rust:1.62
 
-# copy build artifact from build stage
-COPY --from=builder /trin/target/release/trin .
+# copy build artifacts from build stage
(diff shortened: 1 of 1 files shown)


ogenev:
Why do we need the default master accumulator in the final docker image?

If it is because Trin fails when running the code and looking for the file, I think it is better to include those files in the binary itself. We can use [rust_embed](https://crates.io/crates/rust-embed) or [include_dir](https://crates.io/crates/include_dir) crates to achieve this.

carver:
Feel free to re-ping me after the change to eg~ `rust_embed`

njgheorghita:
closing this due to #468

PR #460, opened 2022-10-07
Handle broken ipc sockets due to timeouts

njgheorghita:
### What was wrong?
`web3.py`'s default timeout is 10seconds, which will break the ipc socket if its timeout is reached before the jsonrpc request responds. 

### How was it fixed?
Just handle the error, and we can make multiple, sequential jsonrpc requests to the same ipc socket that result in a timeout. 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/jsonrpc/service.rs
+++ b/trin-core/src/jsonrpc/service.rs
@@ -253,8 +253,10 @@ fn serve_ipc_client(
                     .to_string()
                     .into_bytes(),
                 };
-                tx.write_all(&formatted_response).unwrap();
-                tx.flush().unwrap();
+                match tx.write_all(&formatted_response) {
+                    Ok(_) => tx.flush().unwrap(),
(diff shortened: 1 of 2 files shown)


trin-core/src/jsonrpc/service.rs
```diff
@@ -253,8 +253,10 @@ fn serve_ipc_client(
                     .to_string()
                     .into_bytes(),
                 };
-                tx.write_all(&formatted_response).unwrap();
-                tx.flush().unwrap();
+                match tx.write_all(&formatted_response) {
+                    Ok(_) => tx.flush().unwrap(),
+                    Err(msg) => warn!("Unable to write bytes to unix stream, your request probably timed out: {:?}", msg),
```
mrferris, line 258: nit, to be clear that it's not a trin timeout but a caller (web3.py) timeout
```suggestion
                    Err(msg) => warn!("Unable to write bytes to unix stream, the requester has likely timed out: {:?}", msg),
```

mrferris (approved):

PR #464, opened 2022-10-13
Update merge master acc

njgheorghita:
### What was wrong?
The `MasterAccumulator` struct resulted in an invalid ssz encoding/decoding.

### How was it fixed?
Removed the redundant nested types from `MasterAccumulator`

### To-Do

- [x] Clean up commit history


trin-core/src/types/validation.rs
```diff
@@ -13,9 +13,9 @@ use crate::{
     utils::provider::TrustedProvider,
 };
 
-pub const MERGE_BLOCK_NUMBER: u64 = 15_537_394u64;
+pub const MERGE_BLOCK_NUMBER: u64 = 15_537_393u64;
```
carver, line 16: This change (and the other parts where the accumulator ends one header earlier) seems like it belongs in its own commit separate from the fix to the SSZ structure.

trin-core/src/types/accumulator.rs
```diff
@@ -23,23 +23,23 @@ use crate::utils::bytes::{hex_decode, hex_encode};
 pub const EPOCH_SIZE: usize = 8192;
 
 /// Max number of epochs - 2 ** 17
-const MAX_HISTORICAL_EPOCHS: usize = 131072;
+const _MAX_HISTORICAL_EPOCHS: usize = 131072;
```
ogenev, line 26: What is the reason for prefixing this with `_` instead of deleting the constant?

ogenev (approved):
:boat:

trin-core/src/types/accumulator.rs
njgheorghita: We reference the constant in the line below, so I thought to leave it in to have the actual int as a reference. But yeah, it is unnecessarily compiled, so I've commented it out

Issue #438, opened 2022-09-12
thread 'tokio-runtime-worker' panicked at 'attempt to multiply with overflow'


pipermerriam:
I get a lot of these when running the `eth-portal` bridge client:

```
thread 'tokio-runtime-worker' panicked at 'attempt to multiply with overflow', trin-core/src/utp/stream.rs:849:9
```

Unclear if this is trin <> trin or cross client behavior.

ogenev:
> I could fix this and refactor the utp stream implementation to use `tokio::time::Duration` across the board, instead of the current somewhat bespoke implementation. @ogenev objections? :) 

No objections, feel free to open a PR with anything you feel suitable. A lot of things can be improved in utp 😄

njgheorghita:
@erwanor Any progress on this? I'm still seeing a lot of these errors occurring. If not, no worries at all, we have plenty of other issues to tackle, hahaha. And in the meantime, maybe @ogenev can knock this one out

ogenev:
> I'm still seeing a lot of these errors occurring.

Just curious, do you use the latest master (with the FindContent uTP fix)?

njgheorghita:
@ogenev Yup, it's still occurring on the latest master for me, also seeing a lot of `WARN trin_core::utp::stream: Received STATE packet for an unknown active uTP stream: Packet {}` errors

njgheorghita:
Admittedly, I'm not intimately familiar with the utp spec, but it seems like this could be the source of the issue?
https://github.com/ethereum/trin/blob/master/trin-core/src/utp/stream.rs#L878

As it stands now, the congestion time is squared every time this method is hit, causing it to rise exponentially and eventually overflowing the multiplication buffer. Is it supposed to be squared every time? Or should it just be doubled?

ogenev:
> As it stands now, the congestion time is squared every time this method is hit, causing it to rise exponentially and eventually overflowing the multiplication buffer. Maybe it would be better to simply increment it with a constant value - like the `MIN_CONGESTION_TIMEOUT`?

Yes, incrementing with `MIN_CONGESTION_TIMEOUT` and [saturating_add](https://doc.rust-lang.org/std/primitive.u32.html#method.saturating_add) should solve the overflow panic which is the quickest fix.

However, why is this method called multiple times in a row?

We call this method every time we send a uTP packet and don't receive a response from the remote socket. This is part of the resend uTP logic which works as follows:

1. We send uTP packet.
[...]

PR #467, opened 2022-10-24
Retry limit when receiving acknowledgments for sent uTP packets

### LINKED ISSUE #462: Add a limit when resending uTP packets
We are [resending](https://github.com/ethereum/trin/blob/6af8f46ecfac3cf6772481fd5b51bd335d6b7e69/trin-core/src/utp/stream.rs#L877) a uTP packet every time when we [don't receive](https://github.com/ethereum/trin/blob/6af8f46ecfac3cf6772481fd5b51bd335d6b7e69/trin-core/src/utp/stream.rs#L1454) a response from the remote uTP socket.
The problem is that if we never receive a response from the remote peer, we keep resending that packet forever.

A reasonable solution would be to try to resend a uTP packet a set number of maximum times, for example, 4 times, and then give up on the uTP connection.

Relates to #438 .

ogenev:
### What was wrong?
When we miss acknowledgments for sent uTP packets, we keep trying to receive acknowledgments forever which leads to an infinite loop.

Closes #462, #438

### How was it fixed?
- Set a retry limit when receiving acknowledgments for sent uTP packets
- Fix overflow bug when updating congestion window

### To-Do

- [ ] Clean up commit history


trin-core/src/utp/stream.rs
```diff
@@ -875,7 +877,9 @@ impl UtpStream {
 
     /// Handle uTP socket timeout
     async fn handle_receive_timeout(&mut self) -> anyhow::Result<()> {
-        self.congestion_timeout *= 2;
+        self.congestion_timeout = self
+            .congestion_timeout
+            .saturating_add(MIN_CONGESTION_TIMEOUT);
```
jacobkaufmann, line 882: This change fixes the overflow, but it removes the exponential backoff.

Why not

```rust
self.congestion_timeout = self.congestion_timeout.saturating_mul(2);
```

trin-core/src/utp/stream.rs
```diff
@@ -1644,10 +1648,15 @@ impl UtpStream {
...
     pub async fn flush(&mut self) -> anyhow::Result<()> {
         let mut buf = [0u8; BUF_SIZE];
-        while !self.send_window.is_empty() {
+
+        let mut max_flush_retries: u8 = 0;
+
+        while !self.send_window.is_empty() && max_flush_retries < MAX_RECV_ACKS_RETRIES {
             self.recv(&mut buf).await?;
+            max_flush_retries += 1;
         }
```
jacobkaufmann, line 1660:
```suggestion
        let mut flush_retries: u8 = 0;

        while !self.send_window.is_empty() && flush_retries < MAX_RECV_ACKS_RETRIES {
            self.recv(&mut buf).await?;
            flush_retries += 1;
        }
```

trin-core/src/utp/stream.rs
```diff
@@ -60,6 +60,8 @@ const MAX_BASE_DELAY_AGE: Delay = Delay(60_000_000);
 const DISCV5_SOCKET_TIMEOUT: u64 = 25;
 /// uTP receive timeout in milliseconds
 const UTP_RECEIVE_TIMEOUT: u64 = 1000;
+/// Maximum retries trying to receive acknowledgments for the send packets
+const MAX_RECV_ACKS_RETRIES: u8 = 5;
```
jacobkaufmann, line 64: I don't have a sense of what a good limit is, but is there any empirical evidence for why we choose 5 here?

trin-core/src/utp/stream.rs
ogenev: After reviewing this, I don't think we need to double the congestion timeout every time when we resend a single packet. Increasing the congestion timeout with `MIN_CONGESTION_TIMEOUT` (500ms) seems more reasonable to me.

trin-core/src/utp/stream.rs
ogenev: No empirical evidence, Fluffy is using 4 but I'm happy to change this to any reasonable value (< 10).

trin-core/src/utp/stream.rs
jacobkaufmann: Based on [spec](https://www.bittorrent.org/beps/bep_0029.html): "For every packet consecutive subsequent packet that times out, the timeout is doubled."

I'm not sure what the behavior of the reference implementation is though.

I do think that exponential backoff is generally more effective than linear backoff, so I would be in favor of keeping the doubling. You seem to have some reason to favor the linear backoff, but I'm not sure what that is. Can you elaborate?

ogenev: Didn't know this was in the specs, let's keep it close to the specs then, changing back to `saturating_mul(2)`.

jacobkaufmann (approved):
LGTM. I'm not sure if we should only have a single news fragment with number of this PR instead of two separate ones for each issue, but I'll leave that up to you.

PR #468, opened 2022-10-27
Build and publish docker image to Docker Hub

### LINKED ISSUE #367: Sunsetting `seed-database.rs`
https://github.com/ethereum/trin/blob/master/src/bin/seed-database.rs

This script is actively under use by our testnet nodes every time they boot, to seed headers from the first 10000 blocks. It has been useful for testing data availability, and still might be in the near future. But at some point, I imagine it'll become obsolete.

1. Given that we're about to start pumping data into the network, do we still want to maintain this behavior?
2. There's been discussion about adding some kind of `--is-bootnode` (not a good name) flag that would disable any kind of content transmission. The idea being that bootnodes are only concerned with introducing peers to the network and maintaining a view of the network, and do not need the overhead of storing / propagating content.

ogenev:
### What was wrong?
In order to be able to download the Trin docker image for the interop testing framework, we want to build and push the image with a `latest` tag to DockerHub on a master commit.

Close #367

### How was it fixed?
This PR solves two issues:

1. Supersedes #459 (include master accumulator in docker image) and removes `seed-database` binary. It also reduces the final size of the image by switching the base to `ubuntu:22.04` instead of `rust:1.62`.
2. Creates `docker-build` and `docker-push` pipelines to build and push the image in CircleCI. For now, the image is pushed to my DockerHub account at `ogenev/trin`. Credentials are included in the CircleCI project secrets.
3. Removes seed-database script.

### To-Do

- [x] Clean up commit history


.circleci/config.yml
```diff
@@ -2,7 +2,42 @@ version: 2.1
...
+      - image: cimg/rust:1.62.0
 jobs:
+  docker-build:
+    executor: docker-publisher
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run:
+          name: Build Docker image
+          command: docker build -t $IMAGE_NAME:latest .
```
njgheorghita, line 19: `latest` is the tag that we use for our testnet nodes (though it's `ethpm/trin:latest` not `ogenev/trin:latest`. But I do think we'll want to introduce a `stable` or `testnet` tag, since it's not ideal to update every testnet node each time a new commit is pushed to master

njgheorghita (changes requested):
This all looks good to me, but I'd like to hold off for a moment. This image is what our testnet is configured to use. We could update our testnet config to use this image, but I don't think it's a good idea to update our testnet nodes on each change to master. 

More importantly, it's obvious to me that we need to setup a docker hub org to host our images. I'm going to work on setting that up, and then let's revisit this config. It's messy and confusing to maintain multiple types of images over multiple registries.

I'd also like to add a basic ci workflow check (maybe in a later pr) for the docker image itself in the `trin` repo. Something simple that builds the image and completes a simple interaction with the node. This is because, I've introduced a bug before that broke the testnet. The docker image still built successfully, but all of our testnet nodes broke. The interop test would catch something like this, but I'm not sure at what frequency we plan to run that test, and it seems worth having a simple check in this repo

ogenev:
Hey @njgheorghita, I would like to move this forward. I think we should update the docker file in any case and building and push with a `latest` flag is required for portal-hive.  
We can store this image in the same repo with the deployment image for the testnet (under a different flag) but I think we are not quite there yet and pushing for now to `ogenev/trin` makes sense for interop testing. It is trivial to change the repository link later.

Dockerfile
```diff
@@ -16,27 +16,22 @@ COPY ./trin-core ./trin-core
...
-RUN mkdir /mainnetMM
-COPY --from=builder /trin/mainnetMM ./mainnetMM
+# copy default merge master acc to expected location
+RUN mkdir -p /trin/trin-core/src/assets
+COPY --from=builder /trin/trin-core/src/assets/merge_macc.bin ./trin/trin-core/src/assets/merge_macc.bin
 
 ENV RUST_LOG=debug
 
-ENTRYPOINT ["./trin"]
+ENTRYPOINT ["./trin-main"]
```
njgheorghita, line 37: Could we just run line `32` before line `29` and keep the executable named `trin`?

njgheorghita (approved):
:ship:

PR #471, opened 2022-11-07
Support tracing recursive find content history network requests

### LINKED ISSUE #469: Recursive Find Content debug endpoint
Via a cli arg / custom jsonrpc endpoint, return "route" information along with the content for a `RecursiveFindContent` request.

Route information...
- Did it come from our local db?
- ENRs requested during the recursive lookup (including the one who responded with the content)
- Each ENR's distance from the final content

njgheorghita:
### What was wrong?
It's tricky to see the path a recursive find network request follows as it hops from enr to enr in the network looking up the target content. Also, it's useful to know which enr actually returns the target content.

### How was it fixed?
Add support for a `portal_traceHistoryRecursiveFindContent` request.

Fixes #469 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/docs/jsonrpc_api.md
+++ b/docs/jsonrpc_api.md
@@ -37,6 +37,7 @@ The specification for these endpoints can be found [here](https://eth.wiki/json-
 - [`portal_historyLocalContent`](#portal_historyLocalContent) 
 - [`portal_stateLocalContent`](#portal_stateLocalContent)
 - [`portal_historyRecursiveFindContent`](#portal_historyRecursiveFindContent)
+- [`portal_historyTraceRecursiveFindContent`](#portal_historyTraceRecursiveFindContent)
 - [`portal_paginateLocalContentKeys`](#portal_paginateLocalContentKeys)
 
 # History Overlay Network
@@ -80,13 +81,13 @@ Attempts to look up content key in Trin node's local db.
 ```
 
 ## `portal_historyRecursiveFindContent`
-This method is for development purposes and unstable. It's not fully recursive, but will perform a single lookup in the Trin client's routing table to find a peer close to the content key, and then request the content value from the found peer. 
+Traverses the network by recursively sending `FINDCONTENT` messages in order to retrieve the target content.
 
 ### Parameters
-- `content_key`: Target block header content key.
+- `content_key`: Target content key.
 
 ### Returns
-- Block header
+- Target content value, or `0x` if the content was not found.
 
 #### Example
 ```json
@@ -94,22 +95,32 @@ This method is for development purposes and unstable. It's not fully recursive, 
   "id": 1,
   "jsonrpc": "2.0",
   "result": {
-    "author": "0x2a65aca4d5fc5b5c859090a6c34d164135398226",
-    "base_fee_per_gas": null,
-    "difficulty": "0x710895564a0",
-    "extra_data": "0xd783010302844765746887676f312e352e31856c696e7578",
-    "gas_limit": "0x2fefd8",
-    "gas_used": "0x5208",
-    "log_bloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
-    "mix_hash": "0xc5e99c6e90fbdee5650ff9b6dd41198655872ba32f810de58acb193a954e1589",
-    "nonce": 4679748334323072988,
-    "number": 633130,
-    "parent_hash": "0x6add1c183f1194eb132ca8079197c7f2bc43f644f96bf5ab00a93aa4be499360",
-    "receipts_root": "0x59cf53b2f956a914b8360ea6fe271ebe7b10461c736eb16eb1a4121ba3abbb85",
-    "state_root": "0x5ae233f6377f0671c612ec2a8bd15c20e428094f2fafc79bead9c55a989294dd",
-    "timestamp": 1449116606,
-    "transactions_root": "0x64183d9f805f4aecbf532de75e6ad276dc281ba90947ff706beeaecc14eec6f5",
-    "uncles_hash": "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"
+	  "content": "0xf90217a06add1c183f1194eb132ca8079197c7f2bc43f644f96bf5ab00a93aa4be499360a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347942a65aca4d5fc5b5c859090a6c34d164135398226a05ae233f6377f0671c612ec2a8bd15c20e428094f2fafc79bead9c55a989294dda064183d9f805f4aecbf532de75e6ad276dc281ba90947ff706beeaecc14eec6f5a059cf53b2f956a914b8360ea6fe271ebe7b10461c736eb16eb1a4121ba3abbb85b9010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000860710895564a08309a92a832fefd882520884565fc3be98d783010302844765746887676f312e352e31856c696e7578a0c5e99c6e90fbdee5650ff9b6dd41198655872ba32f810de58acb193a954e15898840f1ce50d18d7fdc"
+  }
+}
+```
+
+## `portal_historyTraceRecursiveFindContent`
+Same as `portal_historyRecursiveFindContent`, but will also return a "route" with the content. The "route" contains all of the ENR's contacted during the lookup, and their respective distance to the target content. If the content is available in local storage, the route will contain an empty array.
+
+### Parameters
+- `content_key`: Target content key.
+
+### Returns
+- Target content value, or `0x` if the content was not found.
+- Network ENRs traversed to find the target content along with their base-2 log distance from the content. If the target content was found in local storage, this will be an empty array.
+
+#### Example
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+	  "content": "0xf90217a06add1c183f1194eb132ca8079197c7f2bc43f644f96bf5ab00a93aa4be499360a01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347942a65aca4d5fc5b5c859090a6c34d164135398226a05ae233f6377f0671c612ec2a8bd15c20e428094f2fafc79bead9c55a989294dda064183d9f805f4aecbf532de75e6ad276dc281ba90947ff706beeaecc14eec6f5a059cf53b2f956a914b8360ea6fe271ebe7b10461c736eb16eb1a4121ba3abbb85b9010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000860710895564a08309a92a832fefd882520884565fc3be98d783010302844765746887676f312e352e31856c696e7578a0c5e99c6e90fbdee5650ff9b6dd41198655872ba32f810de58acb193a954e15898840f1ce50d18d7fdc",
+	  "route": [{
+            "enr": "enr:-IS4QFoKx0TNU0i-O2Bg7qf4Ohypb14-jb7Osuotnm74UVgfXjF4ohvk55ijI_UiOyStfLjpWUZsjugayK-k8WFxhzkBgmlkgnY0gmlwhISdQv2Jc2VjcDI1NmsxoQOuY9X8mZHUYbjqVTV4dXA4LYZarOIxnhcAqb40vMU9-YN1ZHCCZoU",
+            "distance": 256
+	  }]
   }
 }
 ```
--- a/ethportal-peertest/src/scenarios/find.rs
+++ b/ethportal-peertest/src/scenarios/find.rs
@@ -1,0 +1,85 @@
+use serde_json::{json, Value};
+
+use crate::{
+    jsonrpc::{make_ipc_request, JsonRpcRequest, HISTORY_CONTENT_VALUE},
+    Peertest, PeertestConfig,
+};
+use trin_core::jsonrpc::types::Params;
+
+pub fn test_trace_recursive_find_content(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    let uniq_content_key = "0x0015b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286";
+    // Store content to offer in the testnode db
+    let store_request = JsonRpcRequest {
+        method: "portal_historyStore".to_string(),
+        id: 11,
+        params: Params::Array(vec![
+            Value::String(uniq_content_key.to_string()),
+            Value::String(HISTORY_CONTENT_VALUE.to_string()),
+        ]),
+    };
+
+    let store_result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &store_request).unwrap();
+    assert_eq!(store_result.as_str().unwrap(), "true");
+
+    // Send trace recursive find content request
+    let request = JsonRpcRequest {
+        method: "portal_historyTraceRecursiveFindContent".to_string(),
+        id: 12,
+        params: Params::Array(vec![json!(uniq_content_key)]),
+    };
+
+    let result = make_ipc_request(&peertest.nodes[0].web3_ipc_path, &request).unwrap();
+    assert_eq!(result["content"], json!(HISTORY_CONTENT_VALUE.to_string()));
+    assert_eq!(
+        result["route"],
+        json!([{"enr": &peertest.bootnode.enr.to_base64(), "distance": 256}])
+    );
+}
+
+// This test ensures that the jsonrpc channels don't close if there is an invalid recursive find
+// content request
+pub fn test_recursive_find_content_invalid_params(
+    _peertest_config: PeertestConfig,
+    peertest: &Peertest,
+) {
+    let invalid_content_key = "0x00";
+    // Send trace recursive find content request
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindContent".to_string(),
+        id: 12,
+        params: Params::Array(vec![json!(invalid_content_key)]),
+    };
+
+    let error = make_ipc_request(&peertest.nodes[0].web3_ipc_path, &request).unwrap_err();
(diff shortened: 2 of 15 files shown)


jacobkaufmann (commented):
I'm not convinced that this is the right approach. I would prefer that the trace information be maintained in the query internally, and only in the form of ENRs. The distances can be computed later based on the target.

`FindContentQuery` already maintains some notion of contacted peers in `closest_peers`. Those peers who were contacted successfully (except for the peer who responded with the content) are yielded in the query result. This information, perhaps with minor modifications, appears sufficient to satisfy the JSON-RPC endpoint.

docs/jsonrpc_api.md
```diff
@@ -79,36 +80,46 @@ Attempts to look up content key in Trin node's local db.
...
+  }
+}
+```
+
+## `portal_historyRecursiveFindContent`
+Same as `portal_historyRecursiveFindContent`, but will also return a "route" with the content. The "route" contains all of the ENR's contacted during the lookup, and their respective distance to the target content. If the content is available in local storage, the route will contain "local".
```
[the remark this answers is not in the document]
njgheorghita: Base-2 log distance is sufficient imo, updated the docs to specify. But I'm open to returning the actual distance

trin-history/src/jsonrpc.rs
```diff
@@ -73,54 +79,132 @@ impl HistoryRequestHandler {
                     };
                     let _ = request.resp.send(response);
                 }
+                // todo: combine with TraceRecursiveFindContent & refactor once `break` from
+                // labeled blocks becomes stable
+                // https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#break-from-labeled-blocks
```
njgheorghita, line 84: I spent quite some time (in vain) trying to combine & refactor these two rpc call execution blocks since they share a majority of code. However, it turns out that the early `return` pattern we used for parsing the params breaks the tx channel. I confirmed this by passing in invalid `RecursiveFindContent` params. Using an early `return` was about the only way I could think of to refactor this code. The other being the pattern linked above, but that's not a stable feature. If there are any other ideas on how to refactor this monstrosity of a nested code block, I'm all ears

njgheorghita:
@jacobkaufmann Thanks for the review, I wasn't too familiar with the inner workings of the query pool, and leveraging `closest_nodes` seems like a better, less complex mechanism than the `"route"` attribute I introduced in the first commit. 

A couple thoughts on using the `closest_nodes` approach...
- `closest_nodes` will only contain a max of [config.num_results](https://github.com/ethereum/trin/blob/master/trin-core/src/portalnet/find/iterators/findcontent.rs#L312) nodes. So, the returned trace will be incomplete if it took more than 20 (the default) "hops" to find the target content. 
  - One approach I tried to mitigate this was returning the full list from the query and truncating the result at the `overlay_service` level, but this broke the test [here](https://github.com/ethereum/trin/blob/master/trin-core/src/portalnet/find/iterators/findcontent.rs#L601) and I'm unsure if that's acceptable.
- Iiuc, `closest_nodes` will return all nodes that responded successfully during query execution. eg. we could receive a redundant `ClosestNodes` response right before receiving a `Content` response from a different node, resulting in returning `closest_nodes` that were never actually part of the "path" traversed towards the content. 
  - tbf... my `"route"` approach had this same issue.

imo, returning a precise trace will require additional, possibly complex, logic. also imo, I'm not sure that additional logic is very useful at this point, especially when we're dealing with a relatively small testnet. But, it's probably something we might want to consider as the network grows.

jacobkaufmann:
> `closest_peers` will only contain a max of [config.num_results](https://github.com/ethereum/trin/blob/master/trin-core/src/portalnet/find/iterators/findcontent.rs#L312) nodes. So, the returned trace will be incomplete if it took more than 20 (the default) "hops" to find the target content.

If you look at the [query polling logic](https://github.com/ethereum/trin/blob/f6771565ff61d04cdef1f9f9e55c010bdc3caf6d/trin-core/src/portalnet/find/iterators/findcontent.rs#L266-L267), once we have `config.num_results` successful responses, then the query is marked as finished, so there cannot be more than `config.num_results` successful "hops". If you try to [advance the query after it is marked as finished](https://github.com/ethereum/trin/blob/f6771565ff61d04cdef1f9f9e55c010bdc3caf6d/trin-core/src/portalnet/find/iterators/findcontent.rs#L97-L99), then it will be a no-op. Maybe there is an edge case, but at first glance the problem you describe does not seem to exist.

> Iiuc, `closest_peers` will return all nodes that responded successfully during query execution. eg. we could receive a redundant `ClosestNodes` response right before receiving a `Content` response from a different node, resulting in returning `closest_peers` that were never actually part of the "path" traversed towards the content.

I'm having a hard time understanding the scenario you are describing here. Specifically, I don't understand how we would return peers that were never contacted as part of the query. Can you clarify?

> imo, returning a precise trace will require additional, possibly complex, logic.

I agree. I think a more precise trace would yield a nested structure, which provides not just a list of nodes contacted but which nodes returned which other nodes, and allow you to follow the path traversed toward the content. This would require more complex changes.

njgheorghita:
>  Maybe there is an edge case, but at first glance the problem you describe does not seem to exist.

:+1:

> Can you clarify?

It seems to me like it's possible that a `ClosestNodes` response from peer "A" updates `closest_peers` with a set of peers, and immediately after we receive a `Content` response from peer "B", before we're able to contact the peers from peer "A"'s response. In this case, `closest_peers` would contain node info for some peers who were never actually contacted during query execution. 

It's also possible that I'm misunderstanding this?

> I agree. ... . This would require more complex changes.

Cool. Imo, introducing this complex logic isn't necessary at this point. It will surely have to manage many edge cases, which really isn't a huge priority. Imo, improving the accuracy of the trace can be tracked in an issue for the time being. What do you think? Is a "general approximation of the trace" sufficient for now?

jacobkaufmann:
> It seems to me like it's possible that a `ClosestNodes` response from peer "A" updates `closest_peers` with a set of peers, and immediately after we receive a `Content` response from peer "B", before we're able to contact the peers from peer "A"'s response. In this case, `closest_peers` would contain node info for some peers who were never actually contacted during query execution.

The [result](https://github.com/ethereum/trin/blob/f6771565ff61d04cdef1f9f9e55c010bdc3caf6d/trin-core/src/portalnet/find/iterators/findcontent.rs#L294) of a `FindContentQuery` will never contain peers who were not successfully contacted. A peer is only [marked](https://github.com/ethereum/trin/blob/f6771565ff61d04cdef1f9f9e55c010bdc3caf6d/trin-core/src/portalnet/find/iterators/findcontent.rs#L104) "successful" when the query is advanced with a response from that peer.

If that is not clear, then I can go into more detail.

jacobkaufmann (commented):
I think there are still some things to address here, beyond the merge conflicts.

trin-history/src/jsonrpc.rs
mrferris: I think that `continue` can do what we need here.

Here's an approach that I tested (with some awkward safe unwraps that should be cleaned up) that combines both endpoints and minimizes nesting. 
```suggestion
...
  HistoryEndpoint::RecursiveFindContent
  | HistoryEndpoint::TraceRecursiveFindContent => {
      let params = RecursiveFindContentParams::try_from(request.params);
      if let Err(msg) = params {
          let _ = request.resp.send(Err(format!(
              "Invalid RecursiveFindContent params: {:?}",
              msg.code
          )));
          continue;
      }
      let params = params.unwrap();
      let content_key = HistoryContentKey::try_from(params.content_key.to_vec());
      if let Err(msg) = content_key {
          let _ = request
              .resp
              .send(Err(format!("Invalid content key requested: {}", msg)));
          continue;
      }
      let content_key = content_key.unwrap();
      // Check whether we have the data locally.
      let local_content: Option<Vec<u8>> =
          match self.network.overlay.store.read().get(&content_key) {
              Ok(Some(data)) => Some(data),
              Ok(None) => None,
              Err(err) => {
                  error!(
                      error = %err,
                      content.key = %content_key,
                      "Error checking data store for content",
                  );
                  None
              }
          };

      if local_content.is_some() {
          let response = match request.endpoint {
              HistoryEndpoint::RecursiveFindContent => {
                  let content = local_content.unwrap();
                  Value::String(hex_encode(content))
              }
              HistoryEndpoint::TraceRecursiveFindContent => {
                  json!({
                      "content": Value::Null,
                      "route": []
                  })
              }
              _ => json!("none"),
          };
          let _ = request.resp.send(Ok(response));
          continue;
      }

      let (content, closest_nodes) = self
          .network
          .overlay
          .lookup_content(content_key.clone())
          .await;

      let value = if request.endpoint == HistoryEndpoint::TraceRecursiveFindContent {
          let closest_nodes = closest_nodes.unwrap_or_default();
          let closest_nodes: Vec<Value> = closest_nodes
              .iter()
              // Skip over enrs that are unable to be looked up
              .filter_map(|node_id| {
                  let content_id = content_key.content_id();
                  // Return None for enrs that cannot be looked up
                  match self.network.overlay.discovery.find_enr(node_id) {
                      Some(enr) => {
                          XorMetric::distance(&content_id, &enr.node_id().raw())
                              .log2()
                              .map(|distance| {
                                  json!({
                                      "enr": enr,
                                      "distance": distance
                                  })
                              })
                      }
                      None => None,
                  }
              })
              .collect();

          json!({
              "content": match content {
                  Some(content) => Value::String(hex_encode(content)),
                  None => Value::Null,

              },
              "route": closest_nodes,
          })
      } else {
          match content {
              Some(content) => Value::String(hex_encode(content)),
              None => Value::Null,
          }
      };

      let _ = request.resp.send(Ok(value));
}
```

mrferris (changes requested):

trin-core/src/portalnet/overlay.rs
```diff
@@ -662,7 +662,11 @@ where
     }
 
     /// Performs a content lookup for `target`.
-    pub async fn lookup_content(&self, target: TContentKey) -> Option<Vec<u8>> {
+    /// Returns the target content along with the peers traversed during content lookup.
+    pub async fn lookup_content(
+        &self,
+        target: TContentKey,
+    ) -> (Option<Vec<u8>>, Option<Vec<NodeId>>) {
```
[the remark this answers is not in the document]
njgheorghita: Always returning a list of contacted peers (empty or not) makes sense to me. It does seem like useful information to return a `Result` for `lookup_content()`, but I'd rather implement that in a follow up pr to avoid bloat if you 👍

trin-history/src/jsonrpc.rs
```diff
@@ -73,54 +79,132 @@ impl HistoryRequestHandler {
...
+                                                .collect();
+                                            Ok(json!({
+                                                "content": hex_encode(content),
+                                                "route": closest_nodes,
+                                            }))
                                         }
                                     }
-                                    Some(content) => Ok(Value::String(hex_encode(content))),
                                 }
+                                Err(msg) => Err(format!("Invalid content key requested: {}", msg)),
```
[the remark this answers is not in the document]
njgheorghita: Updated these nits in this pr, but if this is a change you feel strongly about, I'd recommend creating a pr to detail this in the contributing guidelines (`docs/contributing.md`) and then an issue/pr to grep through the repo and change all of these occurrences

trin-history/src/jsonrpc.rs
```diff
@@ -73,54 +79,132 @@ impl HistoryRequestHandler {
...
+                                                .iter()
+                                                // Skip over enrs that are unable to be looked up
+                                                .filter_map(|node_id| {
+                                                    let content_id = content_key.content_id();
+                                                    // Return None for enrs that cannot be looked up
+                                                    match self
+                                                        .network
+                                                        .overlay
+                                                        .discovery
+                                                        .find_enr(node_id)
```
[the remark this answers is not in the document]
njgheorghita: Hmm, good point, but if I'm understanding correctly, this will require a new message type to make this information available via `self.network.overlay` (aka `OverlayProtocol`). Seems out of scope for this pr to me, and imo not the highest priority (but lmk if you feel otherwise). I'd suggest we track this in an issue for now?

njgheorghita:
@mrferris Nice! yeah, `continue` was exactly what I was looking for. I added a test scenario with invalid find content params to ensure that the channels stay alive if the endpoint execution exits early.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -3305,14 +3314,9 @@ mod tests {
...
             .await
-            .expect("Expected result on callback channel receiver")
-        {
-            Some(result_content) => {
-                assert_eq!(result_content, content);
-            }
-            _ => panic!("Unexpected find content query result type"),
-        }
+            .expect("Expected result on callback channel receiver");
+        assert_eq!(response.0.unwrap(), content);
```
jacobkaufmann, line 3320: I would prefer that we keep the original `match` format here. It's easier to understand what the test case asserts.

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    let result = make_ipc_request(&peertest.nodes[0].web3_ipc_path, &request).unwrap();
+    assert_eq!(result["content"], json!(HISTORY_CONTENT_VALUE.to_string()));
+    assert_eq!(
+        result["route"],
+        json!([{"enr": &peertest.bootnode.enr.to_base64(), "distance": 256}])
+    );
+}
+
+// This test ensures that the jsonrpc channels don't close if there is an invalid RFC request
```
jacobkaufmann, line 39:
```suggestion
// This test ensures that the jsonrpc channels don't close if there is an invalid RPC request
```

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -0,0 +1,84 @@
+use serde_json::{json, Value};
+
+use crate::{
+    jsonrpc::{make_ipc_request, JsonRpcRequest, HISTORY_CONTENT_VALUE},
+    Peertest, PeertestConfig,
+};
+use trin_core::jsonrpc::types::Params;
+
+pub fn test_trace_recursive_find_content(_peertest_config: PeertestConfig, peertest: &Peertest) {
```
jacobkaufmann, line 9: why the unused `_peertest_config` here and in other functions?

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -771,6 +779,7 @@ where
                     // If content was found, then offer the content to the closest nodes who did
                     // not possess the content.
                     if let Some(content) = content {
+                        let closest_nodes = closest_nodes.into_iter().collect();
```
jacobkaufmann, line 782: I don't understand the reason for this change. It seems unnecessary.

trin-core/tests/overlay.rs
```diff
@@ -230,12 +230,6 @@ async fn overlay() {
...
-    match overlay_one.lookup_content(content_key).await {
-        Some(found_content) => {
-            assert_eq!(found_content, content);
-        }
-        None => {
-            panic!("Unable to find content stored with peer");
-        }
-    }
+    let response = overlay_one.lookup_content(content_key).await;
+    assert_eq!(response.0.unwrap(), content);
```
jacobkaufmann, line 234: Same comment as above w.r.t. `match` format.

trin-history/src/jsonrpc.rs
```diff
@@ -92,54 +98,95 @@ impl HistoryRequestHandler {
...
+                                    }))
                                 }
+                                false => Ok(Value::String(hex_encode(content))),
                             }
-                            Err(err) => Err(format!("Invalid content key requested: {}", err)),
-                        };
-
+                        }
+                        Some(content) => match is_trace {
+                            true => Ok(json!({
```
jacobkaufmann, line 183:
```suggestion
                        Some(content) => if is_trace {
                            Ok(json!({
```

trin-history/src/jsonrpc.rs
```diff
@@ -92,54 +98,95 @@ impl HistoryRequestHandler {
...
-                                    };
-                                match local_content {
-                                    None => {
-                                        match self.network.overlay.lookup_content(content_key).await
-                                        {
-                                            Some(content) => {
-                                                let value = Value::String(hex_encode(content));
-                                                Ok(value)
+                    let response = match local_content {
+                        None => {
```
jacobkaufmann, line 143: We could eliminate some nesting if we did something like the following:

```rust
let (content, closest_nodes) = match local_content {
  Some(content) => (Some(content), Vec::new()),
  None => {
    // Look up content via overlay
  }
}
// Handle trace if requested

File: trin-state/src/jsonrpc.rs [hunk 13]
```diff
@@ -102,12 +102,10 @@ impl StateRequestHandler {
...
                                     None => {
-                                        match self.network.overlay.lookup_content(content_key).await
-                                        {
-                                            Some(content) => {
-                                                let value = Value::String(hex_encode(content));
-                                                Ok(value)
-                                            }
+                                        let response =
+                                            self.network.overlay.lookup_content(content_key).await;
+                                        match response.0 {
```
jacobkaufmann, line 107: Again, I would prefer that we keep the match structure. We can always ignore irrelevant parts of the match pattern like the following:

```rust
match self.network.overlay.lookup_content(content_key).await {
  (Some(content), _) => { ... }
  ...
}
```

jacobkaufmann (commented):
I still have some remaining comments. Most of them are minor and syntactic, but I think there is room to simplify the JSON-RPC handler logic.

ethportal-peertest/src/scenarios/find.rs
njgheorghita: `RFC` = "Recursive Find Content" - since this test will only cover cases with invalid params for a recursive find content request, not all rpc request. But I see the point that this is confusing, so I just spelled it out

trin-history/src/jsonrpc.rs
```diff
@@ -92,54 +98,95 @@ impl HistoryRequestHandler {
...
+                            let (content, closest_nodes) = self
+                                .network
+                                .overlay
+                                .lookup_content(content_key.clone())
+                                .await;
+                            let content = content.unwrap_or_default();
+                            match is_trace {
+                                true => {
```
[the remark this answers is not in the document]
njgheorghita: Personally, I prefer to `match` on a bool. If you feel strongly about this *nit* - I think you should open a pr to the style guide - and we can discuss it there, and then "enshrine" a certain pattern in the docs, so it's no longer left to personal preference.

ethportal-peertest/src/scenarios/find.rs
njgheorghita: I'm just following the pattern here. If you think it's worth going back and refactoring the pattern, let's do it in another pr/issue.

trin-history/src/jsonrpc.rs
```diff
@@ -92,54 +98,87 @@ impl HistoryRequestHandler {
...
+                    let (content, closest_nodes) = match local_content {
+                        Some(val) => (Some(val), vec![]),
+                        None => {
+                            self.network
+                                .overlay
+                                .lookup_content(content_key.clone())
+                                .await
+                        }
+                    };
+                    let content = content.unwrap_or_default();
```
jacobkaufmann, line 151: IIUC, if `content` is `None`, then we will return a hex-encoded empty byte array in the RPC response. Is that the expected behavior?

mrferris (commented):
My main remaining feedback on this is that I don't think that outputting a flat list of the closest nodes is very useful. I proposed an alternative in #482.

Let's merge this (pending @jacobkaufmann's comments) and then I'll rebase #482 on top of it.

trin-history/src/jsonrpc.rs
njgheorghita: Yup, I think it's fine for now... until I address [this comment](https://github.com/ethereum/trin/pull/471#discussion_r1025579107) in a follow up pr.

jacobkaufmann: Based on the [JSON-RPC docs](https://github.com/njgheorghita/trin/blob/5d1bcde647f189884e8707f90990ef6a78fd6e30/docs/jsonrpc_api.md?plain=1#L90), it seems that the correct value in the response would be `null`. Looking at the current version of the portal network JSON-RPC spec, it's not clear to me what the expected value is when the requested content is not found. We can figure that out in a subsequent PR, but we should make sure our documentation is consistent with the behavior prior to merging.

jacobkaufmann (approved):

PR #472, opened 2022-11-07
[WIP] - Implement history recursive find nodes endpoint

pipermerriam:
### What was wrong?

No JSON-RPC endpoint for doing a recursive-find-nodes request.

### How was it fixed?

This is a WIP.  Mostly copy/pasta right now and untested.

### To-Do

- [ ] Clean up commit history


pipermerriam:
I no longer need this with any immediacy so... I'm unlikely to pick it back up in the next few weeks... however I would like to get this across the line, but I might need @jacobkaufmann to sit down with me some Wednesday and pair on this because I still have very little idea of the right way to do things.

PR #474, opened 2022-11-08
Paginate local content keys

### LINKED ISSUE #470: Paginate local content keys jsonrpc endpoint
Introduce a jsonrpc debug endpoint that returns a paginated list of all the content keys stored by a node.

possibly: `portal_debugLocalContent` / `portal_paginateLocalContent` ...

njgheorghita:
### What was wrong?
- fixes #470
- Rearrange ethportal-peertest scenarios

### To-Do

- [x] Clean up commit history

### DIFF
--- a/docs/jsonrpc_api.md
+++ b/docs/jsonrpc_api.md
@@ -37,6 +37,7 @@ The specification for these endpoints can be found [here](https://eth.wiki/json-
 - [`portal_historyLocalContent`](#portal_historyLocalContent) 
 - [`portal_stateLocalContent`](#portal_stateLocalContent)
 - [`portal_historyRecursiveFindContent`](#portal_historyRecursiveFindContent)
+- [`portal_paginateLocalContentKeys`](#portal_paginateLocalContentKeys)
 
 # History Overlay Network
 
@@ -149,6 +150,31 @@ Attempts to look up content key in Trin node's local db.
   "jsonrpc": "2.0",
   "result": {
 	  "content": "0x0217a06ebc43f644f96bf5ab00a93aa4be499360a01dcc4de8dec75d0a142fd40d49347942a65aca4d5fc5b5c859090a6c34d164135398226a05ae233f6377f0671c612ec2a8bd15c20e428094f2fafc79bead9c55a989294dda064183d9f805f4aecbf532de75e6ad276dc281ba90947ff706beeaecc14eec6f5a059cf53b2f956a914b8360ea6fe271ebe7b10461c736eb16eb1a4121ba3abbb85b90110895564a08309a92a832fefd882520884565fc3be98d783010302844765746887676f312e352e31856c696e7578a0c5e99c6e90fbdee5650ff9b6dd41198655872ba32f810de58acb193a954e15898840f1ce50d18d7fdc"
+  }
+}
+```
+
+# General
+
+## `portal_paginateLocalContentKeys`
+Return a paginated list of all of the content keys (from every subnetwork) currently available in local storage.
+
+### Parameters
+- `offset`: The number of records that need to be skipped.
+- `limit`: Number of entries to return.
+
+### Returns
+- `content_keys`: List of content keys.
+- `total_entries`: Total number of content keys in local storage.
+
+#### Example
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "content_keys": ["0x0055b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286"],
+    "total_entries": 1
   }
 }
 ```
--- a/ethportal-peertest/src/scenarios/paginate.rs
+++ b/ethportal-peertest/src/scenarios/paginate.rs
@@ -1,0 +1,103 @@
+use serde_json::{json, Value};
+
+use crate::{
+    jsonrpc::{make_ipc_request, JsonRpcRequest},
+    Peertest, PeertestConfig,
+};
+use trin_core::jsonrpc::types::Params;
+use trin_core::utils::bytes::random_32byte_array;
+
+pub fn test_paginate_local_storage(peertest_config: PeertestConfig, _peertest: &Peertest) {
+    // Test paginate with empty storage
+    let paginate_request = JsonRpcRequest {
+        method: "portal_paginateLocalContentKeys".to_string(),
+        id: 11,
+        params: Params::Array(vec![json!(0), json!(1)]),
+    };
+    let result = make_ipc_request(&peertest_config.target_ipc_path, &paginate_request).unwrap();
+    assert_eq!(result["total_entries"], 0);
+    assert_eq!(result["content_keys"].as_array().unwrap().len(), 0);
+
+    let mut content_keys: Vec<String> = (0..20_u8)
+        .map(|_| format!("0x00{}", hex::encode(random_32byte_array(0))))
+        .collect();
+
+    for (i, content_key) in content_keys.clone().into_iter().enumerate() {
+        // Store content to offer in the testnode db
+        let store_request = JsonRpcRequest {
+            method: "portal_historyStore".to_string(),
+            id: i as u8,
+            params: Params::Array(vec![
+                Value::String(content_key),
+                Value::String("0x00".to_string()),
+            ]),
+        };
+
+        let store_result =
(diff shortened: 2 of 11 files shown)


docs/jsonrpc_api.md
```diff
@@ -152,3 +153,32 @@ Attempts to look up content key in Trin node's local db.
...
+
+#### Example
+```json
+{
+  "id": 1,
+  "jsonrpc": "2.0",
+  "result": {
+    "content_keys": ["0x0055b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286"],
+    "offset": 0,
+    "limit": 5,
```
carver, line 180: Why are we returning the same offset and limit they just passed in? They should already have it, so it doesn't seem to add any value.

trin-core/src/portalnet/storage.rs
```diff
@@ -693,6 +737,11 @@ const XOR_FIND_FARTHEST_QUERY: &str = "SELECT
 
 const TOTAL_DATA_SIZE_QUERY: &str = "SELECT TOTAL(content_size) FROM content_metadata";
 
+const TOTAL_ENTRY_COUNT_QUERY: &str = "SELECT COUNT(content_id_long) FROM content_metadata";
+
+const PAGINATE_QUERY: &str =
+    "SELECT content_id_long FROM content_metadata ORDER BY content_id_long DESC LIMIT :limit OFFSET :offset";
```
ogenev, line 743: I'm not sure if we want to order the `content_id` before setting the `offset`. Maybe I'm missing something but what is the use case for using the offset on ordered content ids?

carver: Hm, doesn't that open up to returning in random order and making it impossible to page through all the results, because you get random selections at each page?

ogenev: My thinking process is that the DB order of the records is the order they were created (the last saved content id is on the last row). Then. if we want to get, let's say the last 10 saved content keys, we can use an `offset = total_records_number - 10`. 

However, If we order the records by `content_id_long` (sha256 hash), this will sort the records in an order that is not predictable and then the last 10 records will be what I would call random.

carver: Ah, so you're proposing a new use case. The only previous use case mentioned was to enumerate *all* keys. The one your suggesting is to retrieve the most recent N keys added. So I guess the question I want to make sure we're asking is when would we use that feature? Is that just something that seems like a good idea, or are we ready to use it now?

Second, if we decide it's a good idea, SQL doesn't have guarantees about the default ordering when you don't provide an `ORDER BY` (and I'd rather not make assumptions about which SQL implementation we're using). So in order to reliably provide the feature, I think we'd need to add a new field. Something like a switching the primary key to an autoincrement or adding a new `create_timestamp` field to order by.

I'm not opposed to the idea, but I would like to see the specific example. Also, it feels like something to address in a follow-up PR.

njgheorghita: One more point to consider after speaking with @mrferris is whether or not we want to return all content keys, or all content keys for a specific subnetwork. Returning only the content keys belonging to a specific subnetwork will require an additional table column, but iiuc we need to do this anyway to support different radius / subnetwork. 

I don't have a strong opinion on whether we want to return content keys on a per-subnetwork basis, or simply enumerate them all. However, if we choose per-subnetwork, it's best to implement this asap so that we avoid the need to tear-down our testnet dbs down the road if we implement this feature later.

ogenev: > Ah, so you're proposing a new use case. The only previous use case mentioned was to enumerate _all_ keys. The one your suggesting is to retrieve the most recent N keys added. So I guess the question I want to make sure we're asking is when would we use that feature? Is that just something that seems like a good idea, or are we ready to use it now?

I don't think there are strongly defined use cases here. My initial understanding is that this RPC endpoint will be used for testing purposes (as Piper and Nick suggested) i.e to get visibility over the content keys stored in the network.
My intuition is that we may also want to know **when** content has been stored in a node, which can open the room for more detailed analyses.

I'm in favor of adding `timestamp` and `subnetwork` fields, it doesn't cost too much and I think this will give us better visibility over the content in the network.

ethportal-peertest/src/scenarios/paginate.rs
```diff
@@ -0,0 +1,87 @@
...
+    let paginate_request = JsonRpcRequest {
+        method: "portal_paginateLocalContentKeys".to_string(),
+        id: 11,
+        params: Params::Array(vec![json!(0), json!(1)]),
+    };
+    let result = make_ipc_request(&peertest_config.target_ipc_path, &paginate_request).unwrap();
+    assert_eq!(result["offset"], 0);
+    assert_eq!(result["limit"], 1);
+    assert_eq!(result["total_entries"], 20);
+    assert_eq!(result["content_keys"].as_array().unwrap().len(), 1);
```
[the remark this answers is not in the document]
carver: I just wanted to highlight this ask, before merging.

trin-core/src/jsonrpc/endpoints.rs
```diff
@@ -34,6 +34,8 @@ pub enum HistoryEndpoint {
     RecursiveFindContent,
     Store,
     RoutingTableInfo,
+    // This endpoint is not History network specific
+    PaginateLocalContentKeys,
```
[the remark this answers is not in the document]
njgheorghita: The problem with making this a `PortalEndpoint` is that `PortalEndpoint`s get forwarded onto the `eth_handler` rather than directly to a subnetwork. This is a redundant layer of abstraction that is unnecessary for the function of this endpoint. Since, the `eth_handler` (which doesn't have direct access to storage) will just simply forward the request on to the history subnet. Rather than introduce that redundant forwarding, I thought it would be simpler to just process this endpoint under the guise of a `HistoryEndpoint` - even though it's technically not. 

I don't have a strong opinion on this, if you guys think it's worthwhile to introduce that extra forwarding, in order to maintain the integrity of the naming scheme, I can make the change.

docs/jsonrpc_api.md
njgheorghita: This was the recommended scheme in the blog I read about api response pagination. But, I agree it's redundant. Happy to remove it if you don't think it's useful.

trin-core/src/jsonrpc/endpoints.rs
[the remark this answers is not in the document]
ogenev: > I don't have a strong opinion on this, if you guys think it's worthwhile to introduce that extra forwarding, in order to maintain the integrity of the naming scheme, I can make the change.

This makes sense, we are duplicating many of the endpoints already in `StateEndpoint` and `HistoryEndpoint`. 
I think the simplest solution is to leave it as it is, if we want something more robust, we need to refactor some of the `JsonRpcHandler` code and the endpoints which are out of the scope of this PR.

docs/jsonrpc_api.md
njgheorghita: Went ahead and removed it for now, we can add it back if it ends up being useful

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -366,7 +366,7 @@ pub fn validate_portal_offer(result: &Value, _peertest: &Peertest) {
     let connection_id = result.get("connection_id").unwrap().as_str().unwrap();
     assert!(connection_id.parse::<u64>().is_ok());
     // Should accept the requested content
-    assert_eq!(result.get("content_keys").unwrap().as_str(), Some("0x03"))
+    assert_eq!(result.get("content_keys").unwrap().as_str(), Some("0x02"))
```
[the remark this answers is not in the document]
njgheorghita: I had switched up the order that some of the scenarios were run in, but that switch is no longer needed, so I undid this change.

ethportal-peertest/src/scenarios/paginate.rs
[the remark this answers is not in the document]
njgheorghita: Ask received, updated the tests to validate that the actual content keys being returned are what we expect.

trin-core/src/portalnet/storage.rs
njgheorghita: tracking this issue here: #480

Issue #473, opened 2022-11-08
Invalid message encoding for Portal Accept message


kdeme:
I've noticed invalid Portal protocol `Accept` messages. (Perhaps other responses might have such issue too)

While trying to offer content on the network, I received bogus data where I would expect a proper `Accept` response. Raw bytes received: `7554502072657175657374206572726f723a2043616e27742066696e6420454e5220696e206f7665726c617920726f7574696e67207461626c65206d61746368696e672072656d6f7465204e6f646549643a203078393037392e2e66666663`

The response I'm getting fails on the very first step, decoding the enum value for the Offer response.
The raw bytes I'm getting returned are always the same, which sounded weird, so I tried converting to ASCII, which results in:
`uTP request error: Can't find ENR in overlay routing table matching remote NodeId: 0x9079..fffc`

Which is an actual internal error from: https://github.com/ethereum/trin/blob/4cdea092f2c56f21815c5995ea956b02f31c85ed/trin-core/src/portalnet/overlay_service.rs#L529

This error appears to be propagated all the way to actually sending of the data: 
https://github.com/ethereum/trin/blob/4cdea092f2c56f21815c5995ea956b02f31c85ed/trin-core/src/portalnet/overlay_service.rs#L841 and I think further to https://github.com/ethereum/trin/blob/a9faaee8d4f0148b1b7603df38ee002e31e6cc86/trin-core/src/portalnet/overlay.rs#L195 and https://github.com/ethereum/trin/blob/a9faaee8d4f0148b1b7603df38ee002e31e6cc86/trin-history/src/events.rs#L31?
Not completely sure of the flow of data here as I didn't fully investigate, but it doesn't seem right.

njgheorghita:
this bug was resolved in #475

PR #475, opened 2022-11-08
Return an empty response for requests which result in an error during processing.

njgheorghita:
### What was wrong?
Related to #473 

If there is an error in processing a `TALKREQ`, we respond to the request with a bytes representation of the error message. 

This pr doesn't resolve the core bug, which is likely something to do with enr caching, but at least it prevents responding the the request with trin's internal error message.

### How was it fixed?
Respond to the request with an empty string rather than responding with an error message. 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/475.fixed.md
+++ b/newsfragments/475.fixed.md
@@ -1,0 +1,1 @@
+Respond with empty TALKRESP for TALKREQs that result in an error during processing.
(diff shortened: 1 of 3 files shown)


carver (approved):
Yup, seems reasonable to me!

ogenev (approved):
:+1:

jacobkaufmann (approved):

PR #476, opened 2022-11-09
Account for additional overhead when limiting size of Nodes/Content responses

mrferris:
### What was wrong?

`WARN discv5::socket::send: Sending packet larger than max size: 1300 max: 1280`

#450 fixed most of these errors, but they were still occasionally happening. The packets exceeding these limits cannot be decoded by the remote peer, so we are likely getting ourselves disconnected each time this happens.

Whereas #450 fixed an issue where we were often sending packets ~1000s of bytes over the limit, the issue now is that the limits are occasionally exceeded by 5-20 bytes.

### How was it fixed?

The `limit_enr_list_to_max_bytes` function needs to account for the extra 4 offset bytes per ENR that the SSZ representation of the list of ENRs includes, as well as the 2 bytes for the `selector` byte and the `total` field in a `NODES` response, and the `selector` and union bytes in a `CONTENT` response. 

Trin tends to be returning 9 ENRs when the issue occurs. With this # of ENRs, this change adds ~40 bytes of overhead, which makes me confident that this is what was causing the limit to be exceeded.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -987,7 +987,7 @@ where
                 let enrs = self.find_nodes_close_to_content(content_key);
                 match enrs {
                     Ok(mut val) => {
-                        pop_while_ssz_bytes_len_gt(&mut val, MAX_NODES_SIZE);
+                        pop_while_ssz_bytes_len_gt(&mut val, MAX_CONTENT_NODES_SIZE);
                         Ok(Content::Enrs(val))
                     }
                     Err(msg) => Err(OverlayRequestError::InvalidRequest(msg.to_string())),
@@ -2055,8 +2055,13 @@ const TALK_REQ_PACKET_OVERHEAD: usize = 16 + // IV
     16; // RLP HMAC
 const NODES_PACKET_OVERHEAD: usize = 1 + // Selector byte
     1; // `total` field
+const CONTENT_PACKET_OVERHEAD: usize = 1 + // Selector byte
+    4; // Union type index
 const MAX_NODES_SIZE: usize =
     MAX_DISCV5_PACKET_SIZE - TALK_REQ_PACKET_OVERHEAD - NODES_PACKET_OVERHEAD;
+const MAX_CONTENT_NODES_SIZE: usize =
+    MAX_DISCV5_PACKET_SIZE - TALK_REQ_PACKET_OVERHEAD - CONTENT_PACKET_OVERHEAD;
+
 /// Limits a to a maximum packet size, including the discv5 header overhead.
 fn pop_while_ssz_bytes_len_gt(enrs: &mut Vec<SszEnr>, max_size: usize) {
     while enrs.ssz_bytes_len() > max_size {
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -930,10 +930,10 @@ where
             .as_ref()
             .and_then(|m| Some(m.report_inbound_find_nodes()));
         let distances64: Vec<u64> = request.distances.iter().map(|x| (*x).into()).collect();
-        let enrs = self.nodes_by_distance(distances64);
+        let mut enrs = self.nodes_by_distance(distances64);
 
         // Limit the ENRs so that their summed sizes do not surpass the max TALKREQ packet size.
-        let enrs = limit_enr_list_to_max_bytes(enrs, MAX_NODES_SIZE);
+        pop_while_ssz_bytes_len_gt(&mut enrs, MAX_NODES_SIZE);
 
         Nodes { total: 1, enrs }
     }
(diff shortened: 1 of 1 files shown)


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -2067,9 +2068,12 @@ fn limit_enr_list_to_max_bytes(enrs: Vec<SszEnr>, max_size: usize) -> Vec<SszEnr
     }
 
     let mut total_size: usize = 0;
+
     enrs.into_iter()
         .take_while(|enr| {
-            let enr_size = enr.ssz_bytes_len();
+            // Add 4 bytes per ENR for SSZ offset overhead.
+            let enr_size = enr.ssz_bytes_len() + 4;
```
jacobkaufmann, line 2075: So IIUC the offset adjustment is not accounted for in the call to `ssz_bytes_len` because the offset only appears in the encoding of a list. Would it be simpler to gradually build the list, and then call `ssz_bytes_len` on the list? That would hurt performance, but it seems simpler and less error prone.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -2054,9 +2054,10 @@ const TALK_REQ_PACKET_OVERHEAD: usize = 16 + // IV
     9 + // Request ID, max 8 bytes + 1 for RLP encoding
     3 + // RLP Encoding of inner response
     16; // RLP HMAC
-        // ENR SSZ overhead empirically observed to be double.
-        // Todo: determine why this is. It seems too high.
-const MAX_NODES_SIZE: usize = MAX_DISCV5_PACKET_SIZE - TALK_REQ_PACKET_OVERHEAD;
+const NODES_PACKET_OVERHEAD: usize = 1 + // Selector byte
+    1; // `total` field
```
jacobkaufmann, line 2058: IIUC the overhead would not be the same for a `Content` and `Nodes` response because the `Content` message is a SSZ union, which has a `u32` type index. It seems like we only account for the `Nodes` message here.

trin-core/src/portalnet/overlay_service.rs
mrferris: @jacobkaufmann If we were seeing warnings that our packets were much larger than this overhead could account for I'd say yea let's sacrifice performance to guarantee that we account for all overhead. But I reproduced the warning a lot and the highest overflow was by 20 bytes, where this change accounts for an extra ~40 bytes of overhead for 9 ENRs (observed maximum # being returned). So I feel confident that these changes are the only thing missing and will fix the issue.

jacobkaufmann: My suggestion was instead to do something like the following:

```rust
while enrs.ssz_bytes_len() > max_size {
  enrs.pop();
}
```

If `ssz_bytes_len`, when called on the `Vec`, accounts for the offset, then that is less accounting for us to do.

mrferris: Good call, that's nice.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -2054,26 +2053,21 @@ const TALK_REQ_PACKET_OVERHEAD: usize = 16 + // IV
...
-            let enr_size = enr.ssz_bytes_len();
-            total_size = total_size + enr_size;
-            total_size < max_size
-        })
-        .collect()
+fn limit_enr_list_to_max_bytes(mut enrs: Vec<SszEnr>, max_size: usize) -> Vec<SszEnr> {
+    while enrs.ssz_bytes_len() > max_size {
+        enrs.pop();
+    }
+    enrs
```
jacobkaufmann, line 2070: Hmm, now this function signature seems a bit unclear. We take a mutable `Vec` and modify it, but we also return the modified `Vec`. I think we should do one or the other but not both. It seems more efficient to modify the existing `Vec` as opposed to allocating a new one, but it could be more clear to pass in a non-mutable `Vec` and return the shrunken result. I don't have a preference, but the function name should reflect that choice.

mrferris: @jacobkaufmann I don't think it's unclear because without returning the modified `Vec` its owernship would never be transferred back to the calling function and we'd get compile errors trying to use it. 

That said, for efficiency purposes I think we should pass by reference which does allow us to just pass in / modify the `Vec` without having to return it. But I'm not sure if "borrow instead of move" was your point. Let me know if I'm misunderstanding.

> but it could be more clear to pass in a non-mutable Vec and return the shrunken result

This strikes me as less clear because as a reader I'd just be wondering why an inefficient approach was taken instead of an easy, more efficient approach

> but the function name should reflect that choice.

Not sure what you mean, do you have a suggestion? `limit_enr_list_to_max_bytes_without_allocating_a_cloned_list_of_enrs`?

jacobkaufmann: Sorry, that was an unclear comment. And after a second look, the original implementation with the move of the ENR list makes sense. A more clear comment would have been on move versus mutable borrow. That said, I expect the reader's primary concern to be the clarity of the function's behavior, not its efficiency.

To me, a more descriptive name would be something like `pop_while_ssz_bytes_len_gt`.

PR #478, opened 2022-11-14
Support loading custom master accumulators

### LINKED ISSUE #449: Support loading a custom master accumulator from disk
According to the spec, we should allow users to specify a custom master accumulator to use if they might want to. The `--master-acc-path` cli arg introduced in #444 simply needs to be supported, and some validation to make sure that the master accumulator decodes correctly and is at the correct latest height.

njgheorghita:
### What was wrong?
Fixes #449 

While docker-izing [eth-portal](https://github.com/carver/eth-portal), it turns out it would actually be quite useful to support loading a custom macc, due to a naming conflict b/w the executable ethportal (`./trin`) tries to launch and the directory (`./trin/`) in which we look up the default master accumulator. 

This pr also updates the error message if trin is unable to find the master accumulator. Which before was simple (`Not a directory, OS error 20`) which wasn't explicit about what file was unable to be found. 

### How was it fixed?
Added support for loading a custom master acc from path. 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/types/accumulator.rs
+++ b/trin-core/src/types/accumulator.rs
@@ -51,9 +51,10 @@ impl MasterAccumulator {
     pub fn try_from_file(master_acc_path: PathBuf) -> anyhow::Result<MasterAccumulator> {
         let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         path.push(master_acc_path);
-        let raw = std::fs::read(path)?;
+        let raw = std::fs::read(&path)
+            .map_err(|_| anyhow!("Unable to find master accumulator at path: {:?}", path))?;
         MasterAccumulator::from_ssz_bytes(&raw)
-            .map_err(|msg| anyhow!("Unable to decode master accumulator: {:?}", msg))
+            .map_err(|err| anyhow!("Unable to decode master accumulator: {err:?}"))
     }
 
     /// Number of the last block to be included in the accumulator
--- a/newsfragments/478.added.md
+++ b/newsfragments/478.added.md
@@ -1,0 +1,1 @@
+Add support for loading a custom master acc from file via cli arg.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,6 +66,10 @@ pub async fn run_trin(
 
     // Initialize validation oracle
     let master_accumulator = MasterAccumulator::try_from_file(trin_config.master_acc_path.clone())?;
+    info!(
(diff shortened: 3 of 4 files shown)


trin-core/src/types/accumulator.rs
```diff
@@ -51,9 +52,12 @@ impl MasterAccumulator {
...
         let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         path.push(master_acc_path);
-        let raw = std::fs::read(path)?;
-        MasterAccumulator::from_ssz_bytes(&raw)
-            .map_err(|msg| anyhow!("Unable to decode master accumulator: {:?}", msg))
+        let raw = std::fs::read(&path)
+            .map_err(|_| anyhow!("Unable to find master accumulator at path: {:?}", path))?;
+        let master_acc = MasterAccumulator::from_ssz_bytes(&raw)
+            .map_err(|msg| anyhow!("Unable to decode master accumulator: {:?}", msg));
+        info!("Loaded master accumulator from: {:?}", path);
```
jacobkaufmann, line 59: We ought to move this log to the call-site of this function. The function may return an error, so the "loaded..." `info!` log may mislead the reader. We can log with `info!` or `error!` at the call-site based on the result.

ogenev (approved):
Looks good after fixing the logging :+1:

trin-core/src/types/accumulator.rs
```diff
@@ -51,9 +52,12 @@ impl MasterAccumulator {
     pub fn try_from_file(master_acc_path: PathBuf) -> anyhow::Result<MasterAccumulator> {
         let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
         path.push(master_acc_path);
-        let raw = std::fs::read(path)?;
-        MasterAccumulator::from_ssz_bytes(&raw)
-            .map_err(|msg| anyhow!("Unable to decode master accumulator: {:?}", msg))
+        let raw = std::fs::read(&path)
+            .map_err(|_| anyhow!("Unable to find master accumulator at path: {:?}", path))?;
+        let master_acc = MasterAccumulator::from_ssz_bytes(&raw)
+            .map_err(|msg| anyhow!("Unable to decode master accumulator: {:?}", msg));
```
ogenev, line 58:
```suggestion
            .map_err(|err| anyhow!("Unable to decode master accumulator: {err:?}"));
```

PR #481, opened 2022-11-21
Possible bug in uTP packet duplication check 

timoth-y:
### What was wrong?
The duplicate check with `wrapping_sub()` will miss packets whose `seq_nr` is less than the most recently handled one, ie. `self.last_dropped`. As a result, it writes the duplicated bytes into the buffer and messes up the state of UtpStream when performing `flush_incoming_buffer()`.

### How was it fixed?
Changed to `saturating_sub` so that any packet with `seq_nr` less or equal to `self.last_dropped` will be skipped.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
             - run:
                 name: Install rustfmt
                 command: rustup component add rustfmt
-            - run: 
+            - run:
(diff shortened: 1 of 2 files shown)


ogenev (approved):
Thanks!

PR #482, opened 2022-11-24
Query Search Path

mrferris:
### What was wrong?

`portal_historyTraceRecursiveFindContent` endpoint's output contains a list of nodes involved in a content query, but currently lacks a comprehensive trace of the query's search path.

### How was it fixed?

Each content query trace is kept track of by a `QueryTracer` struct, that ultimately serializes to the following JSON format:

```
    {
         "found_content_at": "G",
         "origin": "A",
         "responses":
               "A": {
                       timestamp_ms: 0,
                       responded_with: ["B", "C", "D"],
               },
               "B": {
                       timestamp_ms: 124,
                       responded_with: ["E", "F", "G"],
               },
              "C": {
                      timestamp_ms: 150,
                      responded_with: [],
               },
              "G": {
                       timestamp_ms: 200,
                       responded_with: [],
               },
     }
```

where each letter is an ENR.
Each entry in `responses` is a map of a remote node's ENR to the list of ENRs that it responded with because it didn't have the content. `origin` is the local node, and the `responses` entry for the origin node shows the nodes that were closest to the content in its own routing table. The entry for the `found_at` node should have an empty `responded_with` field, as it is only used to mark the timestamp at which the content was received.

Each node's `timestamp_ms` field contains the number of milliseconds that elapsed between the query beginning and the response from the node being received.

Note that only the first node that responded with a given node `A` will have `A` in their `responded_with` list, so only the actual route taken is present in the data rather than other hypothetical routes. This means that many nodes will have empty `responded_with` arrays. These nodes did respond, but not with anything that hasn't already been seen in the query. This is the case for `C` in the example above.

A full visualization of a query using this data format is done here: https://github.com/pipermerriam/glados/pull/28

Here's a screenshot:
Screen Shot 2023-01-04 at 12 21 11 PM

and here's the route highlighted on a successful query: 

Screen Shot 2022-12-05 at 5 12 48 PM

### To-Do

- [x] Add timestamps
- [x] Use ENRs instead of Node IDs
- [x] Include data to distinguish between no response from a node and no progress toward content from a node
- [x] Unit tests
- [x] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) 
- [x] Clean up commit history


trin-history/src/jsonrpc.rs
```diff
@@ -139,45 +133,29 @@ impl HistoryRequestHandler {
...
-                                .collect();
-                            Ok(json!({
+                        Some(content) => match is_trace {
+                            true => Ok(json!({
                                 "content": hex_encode(content),
-                                "route": closest_nodes,
-                            }))
-                        }
-                        false => Ok(Value::String(hex_encode(content))),
+                                "trace": {},
```
jacobkaufmann, line 155: Is the trace supposed to be empty if the content was found?

trin-core/src/portalnet/find/query_tracer.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// None if the content was not found.
+    pub found_content_at: Option<ENRString>,
+    /// The local node.
+    pub origin: ENRString,
+    /// Map of a node's ID to its corresponding `QueryResponse`
+    pub responses: HashMap<ENRString, QueryResponse>,
+    /// Time that the query started.
+    #[serde(skip_serializing)]
+    pub started_at: SystemTime,
+}
```
jacobkaufmann, line 22: A few comments on this struct.

I would suggest renaming to `QueryTrace`.

I am generally opposed to JSON-RPC types outside of the `jsonrpc` module. I would prefer that the API object types (and the logic to convert to them from "core" types) only exist in the `jsonrpc` module. That may mean some duplication, but it enables us to change the internal representations without changing the API response.

`found_content_at` is not as clear as it could be. It could be interpreted as the time at which we found the content, which would be a meaning consistent with the `started_at` field.

The `origin` field is redundant because we know our own ENR.

trin-core/src/portalnet/find/query_tracer.rs
```diff
@@ -0,0 +1,144 @@
...
+
+/// Represents the response from a single node.
+#[derive(Debug, Serialize)]
+pub struct QueryResponse {
+    /// Milliseconds since query started.
+    pub timestamp_ms: u64,
+    /// Node IDs that this node responded with.
+    pub responded_with: Vec<ENRString>,
+}
```
jacobkaufmann, line 99: A few comments on this struct.

It makes more sense to me to use the absolute timestamp at the time the response was received. The latencies can be computed later.

A minor thing is to rename the field to `timestamp_millis` to be consistent with `std::time` naming. This would apply here and other places where `ms` is used.

I would propose that we rename `responded_with` to `enrs` to be consistent with the portal wire protocol naming.

jacobkaufmann (commented):
Neither approving or requesting changes because I want to get your thoughts on the comments.

I'm also not sure how I feel about changing the query callback response type for `FindContent`. An alternative would be to create a struct for that type and include a field for the trace, which could be wrapped in an `Option`. I don't have a problem with doing the trace for all queries, because it could be useful for metrics, but I'm not sure it should be included in the response for non-trace find content queries.

I also find the structure of the response confusing, particularly the omission of previously returned ENRs from the arrays. If each response is timestamped, then you could infer the path if you assume that you only query a node the first time you see it. Alternatively, you could include a field in the response dedicated to representing the path.

trin-core/src/portalnet/find/query_tracer.rs
```diff
@@ -0,0 +1,144 @@
...
+            origin: local_enr.to_base64(),
+            responses: HashMap::new(),
+            started_at: SystemTime::now(),
+        }
+    }
+
+    /// Records the receipt of a node as part of a query in the trace.
+    /// Adds a new `QueryResponse` object for `node` if it does not exist, updates if it does.
+    /// Timestamp for the object is only set on the first call for a given `node`.
+    pub fn node_responded_with(&mut self, enr: &Enr, responded_with: &Enr) {
```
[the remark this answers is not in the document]
mrferris: Good point, I've squashed them into one method that's called once per response.

trin-core/src/portalnet/find/query_tracer.rs
```diff
@@ -0,0 +1,144 @@
+use crate::portalnet::Enr;
+
+use serde::Serialize;
+use std::collections::HashMap;
+use std::time::{Duration, SystemTime};
+
+///Store ENRs as strings for serialization simplicity
+type ENRString = String;
```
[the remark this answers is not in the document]
mrferris: If I remember right, `HashMap` couldn't use an Enr as a key.

Regardless, I've switched to using `NodeId`s as keys, and `NodeIdString` because `NodeId` doesn't implement `serde::Serialize`

trin-core/src/portalnet/find/query_tracer.rs
mrferris: > I would suggest renaming to QueryTrace.
Agreed. 👍
> I am generally opposed to JSON-RPC types outside of the jsonrpc module. I would prefer that the API object types (and the logic to convert to them from "core" types) only exist in the jsonrpc module. That may mean some duplication, but it enables us to change the internal representations without changing the API response.

I'd prefer we handle any decoupling of types on an as-needed basis in the hypothetical future PR that makes it necessary. For now I think the `peertest` is sufficient for not accidentally changing API responses while changing internal representations.

> found_content_at is not as clear as it could be.
Agreed. 👍 I went with `received_content_from_node` which feels like a bit much. Any better ideas?

> The origin field is redundant because we know our own ENR.
Yea, but whoever's parsing this data needs to get the `origin` node's ENR at some point, and I think it should be bundled with the rest of the data rather than requiring a separate RPC call.

trin-core/src/portalnet/find/query_tracer.rs
mrferris: > It makes more sense to me to use the absolute timestamp

Any reason? Seems like it would just add a bit more data & visual chaos to the JSON output. Though I think I should add a `started_at` field in the output so that the absolute time of any given request can be calculated.

> A minor thing is to rename the field to timestamp_millis to be consistent with std::time naming. This would apply here and other places where ms is used.

Done 👍

> I would propose that we rename responded_with to enrs to be consistent with the portal wire protocol naming.

👍

perama-v (approved):
Very nice. Looks good to me. Tested by making a query (`portal_historyTraceRecursiveFindContent`) and inspecting the results. 

I notice that the resulting `"responses"` includes the nodes own ID and includes `"responded_with": ["node_x", "node_y", ...]` for itself. This seems ok/useful, just wanted to note it.

ethportal-api/src/types/portal.rs
```diff
@@ -53,7 +54,7 @@ pub struct AcceptInfo {
 #[serde(rename_all = "camelCase")]
 pub struct TraceContentInfo {
     pub content: HistoryContentItem,
-    pub route: Vec<NodeInfo>,
+    pub trace: Value,
```
ogenev, line 57: I think, `Value`  is too broad type here. It is hard to know what is the type's exact format before serializing it.

We have already a `QueryTrace` struct which can be used here. Of course, it will need the appropriate serde features so we can implicitly Serialize/Deserialize to `QueryTrace`.

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -81,5 +109,12 @@ pub fn test_trace_recursive_find_content_local_db(
 
     let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
     assert_eq!(result["content"], json!(HISTORY_CONTENT_VALUE.to_string()));
-    assert_eq!(result["route"], json!([]));
+    assert_eq!(
+        result["trace"]["received_content_from_node"],
```
ogenev, line 113: This is an example of what I don't like about the `Value` type. Having a more specific type is easier to work with.

trin-core/src/portalnet/find/query_trace.rs
```diff
@@ -0,0 +1,219 @@
+use crate::portalnet::Enr;
+use crate::utils::bytes::hex_encode;
+
+use discv5::enr::NodeId;
+use serde::Serialize;
+// use ssz_types::typenum::Xor;
```
ogenev, line 6: :scissors:

trin-history/src/jsonrpc.rs
```diff
@@ -244,31 +248,9 @@ impl HistoryRequestHandler {
...
-                        }),
-                    None => None,
-                }
-            })
-            .collect();
-
-        Ok(json!(TraceContentInfo {
-            content: content.into(),
-            route: closest_nodes,
+        Ok(json!({
```
ogenev, line 251: I think it is more clear if we return the `TraceContentInfo` type, instead of json map.

ogenev (changes requested):
Blocking this until we agree on what to do with the `Value` type in `TraceContentInfo`.

mrferris:
> I'm also not sure how I feel about changing the query callback response type for FindContent. An alternative would be to create a struct for that type and include a field for the trace, which could be wrapped in an Option.

Agreed 👍

> I also find the structure of the response confusing, particularly the omission of previously returned ENRs from the arrays. If each response is timestamped, then you could infer the path if you assume that you only query a node the first time you see it.

I originally had it that way but didn't see any useful, non-chaotic way of visualizing that info on the front-end so it was unnecessary data & front-end parsing complexity. I'm open to a second iteration down the line that adds all response data.

mrferris:
Note that `node_meta_data` was added to the format, but will no longer be necessary once full ENR parsing is working on the front-end.

Also note that Node IDs are now being used as the primary ID for each node rather than ENRs. This was due to a bug caused by multiple ENRs for the same node ID making it into the `QueryTrace`.

trin-core/src/portalnet/find/query_info.rs
```diff
@@ -18,6 +19,8 @@ pub struct QueryInfo<TContentKey> {
 
     /// Temporary ENRs used when trying to reach nodes.
     pub untrusted_enrs: SmallVec<[Enr; 16]>,
+
+    pub trace: Option<QueryTrace>,
```
jacobkaufmann, line 23: I'm not sure that the trace belongs in `QueryInfo`, because the trace is specific to content queries, and `QueryInfo` exists for both node and content queries.

trin-history/src/jsonrpc.rs
```diff
@@ -243,33 +247,14 @@ impl HistoryRequestHandler {
...
-            route: closest_nodes,
-        }))
+        if let Some(trace) = trace {
+            Ok(json!({
+                "content": hex_encode(content),
+                "trace": trace,
+            }))
+        } else {
+            Err("Content query trace requested but none provided.".to_owned())
+        }
```
jacobkaufmann, line 257: if we cannot find a clean way to separate the trace and non-trace queries, then I think it would be preferable to always build the trace throughout the query so that we don't have to deal with this `None` possibility.

jacobkaufmann (commented):
It looks like there are some unaddressed comments from other reviewers, and I have some comments of my own.

At a high level, the following aspects of the trace structure do not make much sense to me:

* that the origin node has an entry in the `responses` field
* that only the first node to respond with some node `X` will be the only entry with `X` in that list

I believe we can find a better representation for the trace that doesn't warp the information in these ways. I would be more in favor of a straightforward representation that requires the caller to do whatever necessary disentanglement of the data on their end.

trin-core/src/portalnet/find/query_trace.rs
```diff
@@ -0,0 +1,219 @@
...
+/// Keeps track of query details.
+#[derive(Debug, Serialize)]
+pub struct QueryTrace {
+    /// None if the content was not found.
+    pub received_content_from_node: Option<NodeIdString>,
+    /// The local node.
+    pub origin: NodeIdString,
+    /// Map of a node's ID to its corresponding `QueryResponse`
+    pub responses: HashMap<NodeIdString, QueryResponse>,
+    pub node_metadata: HashMap<NodeIdString, NodeInfo>,
```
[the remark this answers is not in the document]
mrferris: From above: 

> Also note that Node IDs are now being used as the primary ID for each node rather than ENRs. This was due to a bug caused by multiple ENRs for the same node ID making it into the QueryTrace.

trin-core/src/portalnet/find/query_trace.rs
```diff
@@ -0,0 +1,219 @@
...
+}
+
+#[derive(Debug, Serialize)]
+pub struct NodeInfo {
+    pub enr: Enr,
+    pub ip: String,
+    pub port: String,
+    pub distance_to_content: u32,
+    pub distance_log2: usize,
+}
```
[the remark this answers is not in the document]
mrferris: From above:

> Note that node_meta_data was added to the format, but will no longer be necessary once full ENR parsing is working on the front-end.

There are some issues with ENR parsing on the front-end that I haven't sorted out. 

u32 distance and log2 distance so that we can sort nodes with the same log2 distance by distance. I could instead calculate log2 of distance on the frontend.

mrferris:
@jacobkaufmann 

> that the origin node has an entry in the responses field

Why would `origin` not have an entry? Where do we put the nodes that `origin` used to begin the query? 

> that only the first node to respond with some node X will be the only entry with X in that list

My previous response from above:

> I originally had it that way (all responses) but didn't see any useful, non-chaotic way of visualizing that info on the front-end so it was unnecessary data & front-end parsing complexity. I'm open to a second iteration down the line that adds all response data.

Why add unused complexity to both the back-end and front-end? We can iterate this format as our needs progress.

trin-core/src/portalnet/find/query_trace.rs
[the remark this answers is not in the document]
jacobkaufmann: I assume that we don't use `NodeId` because it doesn't implement the `serde` traits then? I would prefer we convert at the JSON-RPC layer as opposed to making the internal logic less straightforward.

trin-core/src/portalnet/find/query_trace.rs
```diff
@@ -0,0 +1,219 @@
...
+            distance_log2,
+        }
+    }
+}
+
+/// Represents the response from a single node.
+#[derive(Debug, Serialize)]
+pub struct QueryResponse {
+    /// Milliseconds since query started.
+    pub timestamp_millis: u64,
```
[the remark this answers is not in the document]
jacobkaufmann: that is not consistent with the [doc comment](https://github.com/mrferris/trin/blob/2a3ab9282608074b29c19951d95d9e39a5ddb149/trin-core/src/portalnet/find/query_trace.rs#L144) or the [logic which constructs the value](https://github.com/mrferris/trin/blob/2a3ab9282608074b29c19951d95d9e39a5ddb149/trin-core/src/portalnet/find/query_trace.rs#L77).

jacobkaufmann:
apologies for missing some of the original comments on rationale.

the origin would not have entry in `responses` because there is no notion of "response" for the originating node. that information (the bootstrap query peers) could exist in a separate field. it seems like we are trying to stuff information into this shape that does not really accommodate the info in a clean way, so we can explore a different structure.

my claim is that right now we are introducing complexity on the back-end where things could be much simpler, and it would be better for any complexity to reside on the front-end. the simplest thing to do on the back-end is to provide all the response data as we received it. as long as the front-end can tease out the structure from that data, then it is preferable to keep our logic simple at the expense of the consumer.

trin-core/src/portalnet/find/query_trace.rs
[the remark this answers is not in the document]
mrferris: Sure, it's an instant whose epoch is the beginning of the query rather than 1970, aka a duration. 

I can make them absolute.

mrferris:
@ogenev I've created `QueryTrace`, `NodeInfo`, and `QueryResponse` types to use in `TraceContentInfo` in `ethportal-api`, but kept the `QueryTrace` implementation that creates the `QueryTrace` in `trin-core`. Perhaps the one in `trin-core` should be named `QueryTracer` and the one in `ethportal-api` named `QueryTrace` to denote that one creates the other.

mrferris:
@jacobkaufmann 

> there is no notion of "response" for the originating node

The format is generalized: each node (including the origin node) responds to the query by either returning the requested content or by looking in its own routing table for closest ENRs to continue the query. I see no need to make the origin node's behavior an edge case and add complexity to the format instead of just describing its behavior the exact same way that the behavior of every other node is described. Seems cleaner than any non-generalized format to me. 

This is the same reason why I felt it was best to change the output of a trace where the content is found locally from `{}` to `{origin: origin_id, found_content_at: origin_id}`. The format describes what happened in a generalized way, rather than piecewise.

I suppose the semantics of the word `responses` could use improvement. Feel free to propose alternatives, maybe `closest_nodes_unseen`. 

> my claim is that right now we are introducing complexity on the back-end where things could be much simpler, and it would be better for any complexity to reside on the front-end. 

Could you elaborate on where complexity is being introduced to trin?

Are you referencing nodes only being included in the trace the first time they're seen? The single `if` statement that handles it is already there for other purposes, and when we make the 1 line modification to trin to add every response it will still be the same format, just with more items in each `responses` entry. I'm not seeing how there's any warping or complexity there. 

I agree with adding all responses in the near future. It will allow us to answer/visualize interesting questions like the degree of routing table overlap across the network. But currently nobody's asking those questions, whereas questions that can be answered by the implemented front-end that consumes this current format are being asked daily. So I'm against blocking deployment of current functionality to add unseen complexity to the front-end for zero gain. Not sure if you're saying we should block this and come up with a new format.

carver:
> @jacobkaufmann
> 
> > there is no notion of "response" for the originating node
> 
> The format is generalized: each node (including the origin node) responds to the query by either returning the requested content or by looking in its own routing table for closest ENRs to continue the query. I see no need to make the origin node's behavior an edge case and add complexity to the format instead of just describing its behavior the exact same way that the behavior of every other node is described. Seems cleaner than any non-generalized format to me.
> 
> This is the same reason why I felt it was best to change the output of a trace where the content is found locally from `{}` to `{origin: origin_id, found_content_at: origin_id}`. The format describes what happened in a generalized way, rather than piecewise.
> 
> I suppose the semantics of the word `responses` could use improvement. Feel free to propose alternatives, maybe `closest_nodes_unseen`.

Yeah, naming this as a DAGSearchPath instead of a query trace might help? This data structure is the nodes and directed edges that were followed in the path. That explains why we drop the duplicate responses, and why the origin node lists outbound edges just like all the intermediate nodes do.

The reason to favor something like Jacob is talking about would be if we want to use the query trace for something besides visualizing a search path. In which case, it makes sense to keep the data structure as clearly describing the series of events that occurred (nodes responded, etc), rather than overfitting to this particular usage.

I don't have a clear idea what else we would use this for right now. So maybe just "admitting" that we're building this data for a single purpose, by naming it as such, is the way to go. Until we have a specific idea of what else we would use a query trace for.

mrferris:
> Yeah, naming this as a DAGSearchPath instead of a query trace might help?

Sure.

> The reason to favor something like Jacob is talking about would be if we want to use the query trace for something besides visualizing a search path. In which case, it makes sense to keep the data structure as clearly describing the series of events that occurred (nodes responded, etc), rather than overfitting to this particular usage.

I would agree if we were locking ourselves into anything. Adding all of the responses is a ~1 line code change in trin, and is backward compatible on the front-end.

jacobkaufmann:
here is a concrete example of the sort of thing I had in mind:

```rust
struct QueryPeer {
  requested_at: Instant,
  responded_at: Instant,
  response: QueryResponse,
}

enum QueryResponse {
  Peers(HashMap<Enr, QueryPeer>),
  Content,
}

struct QueryTrace {
  // originator of query (i.e. local node)
  origin: Enr,
  // target content ID.
  target: ContentId,
  // UTC timestamp for query start.
  started_at: Instant,
  // UTC timestamp for query end (termination).
  ended_at: Instant,
  // first level of peers in the map are the bootstrap peers, or those initially queried.
  trace: HashMap<Enr, QueryPeer>,
}
```

the structure could be modified to include peers who do not respond or use nested arrays instead of maps, but I just want to get the general idea across.

having said that, I'm okay moving forward with the existing structure. like @carver said, we can move forward with this special-purpose design until something more general is required. there are still some outstanding comments that I would like to see addressed though.

ogenev (commented):
IMO, we don't need two QueryTrace types - one in ethportal-api and one in `query_trace.rs`.

We can leave only the `QueryTrace` type in `query_trace.rs` and use some serde primitives not to serialize the `started_at` and `target_id` fields. We can also use the `NodeId` type from `ethportal-api` which will make implementing Serialize/Deserialize for this `QueryTrace` type trivial. Then we should re-export this type via `ethportal-api` for external users to access it.

In summary, this would look something like this:

1. Remove the `QueryTrace` type from `ethportal-api`.
2. Move `QueryTrace` from `trin-core` to `trin-types`.
3. Use `ethportal-api::discv5::NodeId` type instead of  `discv5::enr::NodeId` in `QueryTrace`.
4. Implement Serialize/Deserialize for `QueryTrace` and add  `#[serde(skip_serializing)]` attributes for `started_at` and `target_id` fields.
5. Import  `QueryTrace`  in `ethportal-api` from `trin-types` and re-export it.

ethportal-api/src/types/query_trace.rs
```diff
@@ -0,0 +1,38 @@
+use discv5::Enr;
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+
+///Store ENRs as strings for hashability
+type NodeIdString = String;
```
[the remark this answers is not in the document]
mrferris: 👍

@ogenev I found that when using the `serde-hex` crate and its `#[serde(with = "hex::serde")]` directive (as `NodeId` does) that deserializing a `QueryTrace` would result in a `"Parse error: invalid type: string \"0x6b1538b52e3c1541ebbcea0a06a50d1e985000087ba5b3750b48466004db2a12\", expected a borrowed byte array"`.

I didn't figure out why that happened, though the fact that `serde-hex` [hasn't been maintained in years](https://github.com/fspmarshall/serde-hex/issues/13) probably doesn't help. I switched to use the `hex` crate's `serde` feature, which works but without the `0x` prefix. 

So I think I'll implement custom `Serialization`/`Deserialization` for `NodeId` (which I moved to `trin-types` and re-exported from `ethportal-api`).

ethportal-api/src/types/discv5.rs
ogenev, line 1: Let's delete this file and re-export those types in `ethportal-api` lib file with `pub use`, as in this example: https://github.com/ethereum/trin/blob/85a7d3cd59aea4ec84a8027d5889a3f4108f0be9/ethportal-api/src/lib.rs#L16

ethportal-api/src/types/query_trace.rs
ogenev, line 1: Same as above, we can delete this file and re-export the type in `lib.rs`

portalnet/src/utils/mod.rs
ogenev, line 1: what is the reason for moving those utils to `trin-utils`?

perama-v:
Tested, looks good. Three minor notes: 

1. One note is that the content is returned twice in the response. Perhaps the second `received_content_from_node` field could be skipped (`#[serde(skip_serializing)]`)?
```
{
  "jsonrpc": "2.0",
  "result": {
    "content": "0x", // here
    "trace": {
      "received_content_from_node": null, // here
      "origin":
```
2. Another note is that the timestamp that the request was initialised is returned back to the caller. Perhaps this could also be skipped as the caller will always know when they started this request. 
```
      "started_at": { // ?omit
        "secs_since_epoch": <unix secs>,
        "nanos_since_epoch": <nanos>
      },
```
3. The response differs from the (now outdated) response defined in `book/src/developers/protocols/json_rpc.md`.

portalnet/src/utils/mod.rs
mrferris: @ogenev So that it can be used from within `trin-types`

trin-core/src/portalnet/find/query_info.rs
mrferris: I think that having traces of node queries could be good too, so I'll keep it here for now and move it elsewhere if we decide to never create traces of node queries.

mrferris:
@perama-v 

> One note is that the content is returned twice in the response

The first is the content itself, whereas the second is optional ID of the node that returned the content.

> Perhaps this could also be skipped as the caller will always know when they started this request.

Good point, this was previously skipped, but with the way that `QueryTrace` is now defined, a default value is required to skip that field. The `SystemTime` type we're using doesn't implement Default, so I'm going to make a note to possibly remove this in the future when we decide whether to keep using relative timestamps or use absolute timestamps.

trin-history/src/jsonrpc.rs
mrferris: I'm hesitant to switch it back because of the tradeoff that previously if the query returned an error it also constructed and returned an empty query trace, which is kind of obnoxious. I think that between the two this is cleaner for that reason. If we can find a third even cleaner option in the future let's do that.

trin-core/src/portalnet/find/query_trace.rs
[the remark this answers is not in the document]
mrferris: I'm going to err on the side of human-parseable json on this for now and keep the timestamps as relative. If we have a need for absolute timestamps in the future we can switch to that.

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -1,61 +1,106 @@
...
 
-    // Send trace recursive find content request
-    let request = JsonRpcRequest {
-        method: "portal_historyTraceRecursiveFindContent".to_string(),
-        id: 12,
-        params: Params::Array(vec![json!(uniq_content_key)]),
-    };
+    let uniq_content_key =
+        "\"0x0015b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286\"";
+    let history_content_key: HistoryContentKey = serde_json::from_str(uniq_content_key).unwrap();
```
carver, line 33: This feels like it has extra steps that I don't understand. Why go into a json-encoded string and back out?

I see some other places that do this same concept like this, like:
```rs
    let content_key: HistoryContentKey =
        serde_json::from_value(json!(HISTORY_CONTENT_KEY)).unwrap();
```
which is a little better. But still, it seems like going in and out of json doesn't really do anything for us.

Something equivalent to this seems like the direct path that doesn't involve unnecessary json:
```rs
let key_bytes = "0015b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286";
let history_content_key = HistoryContentKey::from_bytes(hex::decode(key_bytes));
```

portalnet/src/overlay_service.rs
```diff
@@ -2171,7 +2179,10 @@ where
         &mut self,
         target: TContentKey,
         callback: Option<oneshot::Sender<FindContentResult>>,
+        is_trace: bool,
     ) -> Option<QueryId> {
+        info!("Starting query for content key: {}", target);
```
carver, line 2184: We might want to reduce this to debug or drop it altogether, if we notice it being too little noisy, but I'm pretty happy to let that evolve over time, and be just a little generous with logs, especially with new features.

trin-history/src/jsonrpc.rs
```diff
@@ -193,13 +192,20 @@ impl HistoryRequestHandler {
...
-            Some(val) => (Some(val), vec![]),
+        let (possible_content_bytes, trace) = match local_content {
+            Some(val) => {
+                let local_enr = self.network.overlay.local_enr();
+                let mut trace = QueryTrace::new(
+                    &self.network.overlay.local_enr(),
+                    NodeId::new(&content_key.content_id()).into(),
+                );
+                trace.node_responded_with_content(&local_enr);
+                (Some(val), if is_trace { Some(trace) } else { None })
```
carver, line 203: Seems like a bummer to do all the trace work every time, even when it wasn't requested. Maybe something like:
```suggestion
                let trace_option = if is_trace {
                    let mut trace = QueryTrace::new(
                        &self.network.overlay.local_enr(),
                        NodeId::new(&content_key.content_id()).into(),
                    );
                    trace.node_responded_with_content(&local_enr);
                    Some(trace)
                } else { None }
                (Some(val), trace_option)
```

trin-types/src/enr.rs
```diff
@@ -79,3 +81,61 @@ impl ssz::Encode for SszEnr {
...
+    let key = CombinedKey::generate_secp256k1();
+
+    let mut rng = rand::thread_rng();
+    let ip = Ipv4Addr::from(rng.gen::<u32>());
+
+    let enr = EnrBuilder::new("v4")
+        .ip(ip.into())
+        .udp4(8000)
+        .build(&key)
+        .unwrap();
```
carver, line 96: I can't really think of a time when it's better to use `clippy::unwrap_used` than to switch to `expect()` with an explanation of why it won't ever panic in reasonable scenarios of runtime. When does building the ENR fail? Why is that impossible now?

carver (approved):
Just did a quick scan. Seems like we're well into the territory of this being an improvement from where we were, with no major setbacks. I think Ognyan's ❌ is outdated and can be disregarded now. I say we address any last things, and I saw you put issues in for follow-up work, so it's good to go. I'm excited to see the results in glados!

PR #486, opened 2022-12-02
core/jsonrpc: Do not drop TCP stream before reaching EOF

ogenev:
### What was wrong?
When receiving HTTP JSON-RPC request,  Trin drops the TCP stream right after the request is processed (the stream goes out of scope). This is unexpected behavior because the stream may not reach EOF and more requests should be expected on the same stream.

### How was it fixed?
Add a loop to keep processing JSON-RPC requests until stream EOF.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/newsfragments/486.fixed.md
+++ b/newsfragments/486.fixed.md
@@ -1,0 +1,1 @@
+Keep processing JSON-RPC requests until TCP stream EOF. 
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
Nice! As I've working with the testnet nodes, I've noticed some odd bugs here & there with our jsonrpc server closing unexpectedly. It seems worthwhile to harden our server with some tests / fuzzing. Not necessarily for the content received, but the behavior under edge cases for both http/ipc. eg. timeouts, invalid params, etc... That's probably out of scope for this pr though..

Issue #466, opened 2022-10-24
Include master accumulator in Trin's binary


ogenev:
Instead of loading the pre-merge master accumulator from the file path (`trin-core/assets/merge_macc.bin`), Include it in trin binary with [rust-embed](https://crates.io/crates/rust-embed) crate.

PR #488, opened 2022-12-06
Add support for header with proof message type & validation

njgheorghita:
### What was wrong?
Add support for `HeaderWithProof` history network message type, along with proof generation/validation.

### How was it fixed?
- Added a `HeaderWithProof` message type to History network
- Implemented support for generating accumulator proofs via master accumulator
- Implemented support for validating accumulator proofs against master accumulator
- Found a bug in `EpochAccumulator` ssz encoding, which required updating the test accumulators
- Added tests against fluffy & ultralight (epoch & master) accumulators to validate that we are encoding / decoding our message values similarly
- Added tests against fluffy's header with proofs to validate that we are generating/validating our proofs similarly
- Reorganized some test assets

### To-Do

- [x] Clean up commit history


ogenev:
I think it would be useful to include all resources for testing in a separate `test` folder. for example `assets/test`. This way we will know what files are required for testing and what is required for running the binary (we can include them inside the binary, see  https://github.com/ethereum/trin/issues/466).

ogenev (approved):
Looks good, :+1: for optimistically merge.

PR #489, opened 2022-12-07
Run http endpoint tests in ci

njgheorghita:
### What was wrong?
We never run http jsonrpc endpoint tests in ci.

### How was it fixed?
- Update the `ethportal-peertest` scenarios to be transport agnostic (requires async-ification to support http)
- Add support for the http server to listen to the jsonrpc exiter
- add a test to `self_peertest.rs` that runs the jsonrpc endpoints over http. 

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/main.rs
+++ b/ethportal-peertest/src/main.rs
@@ -1,22 +1,53 @@
+use std::net::{IpAddr, Ipv4Addr};
 use std::{thread, time};
 use tracing::info;
 
 use ethportal_peertest::{
     cli::PeertestConfig,
     jsonrpc::{test_jsonrpc_endpoints_over_http, test_jsonrpc_endpoints_over_ipc},
-    launch_peertest_nodes,
+    launch_peertest_nodes, TransportProtocol,
 };
+use trin_core::cli::TrinConfig;
+use trin_core::utils::provider::TrustedProvider;
 
 #[tokio::main]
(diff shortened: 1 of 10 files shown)


ethportal-peertest/src/main.rs
```diff
@@ -1,22 +1,53 @@
...
     tracing_subscriber::fmt::init();
 
-    let peertest = launch_peertest_nodes(2).await;
+    let peertest = launch_peertest_nodes(2, TransportProtocol::IPC).await;
     // Short sleep to make sure all peertest nodes can connect
     thread::sleep(time::Duration::from_secs(1));
 
     tokio::spawn(async move {
-        let peertest_config = PeertestConfig::from_cli();
+        let test_ip_addr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
```
njgheorghita, line 22: `ethportal-peertest` was completely broken without this... I'm split on whether this means we should just remove `cargo run -p ethportal-peertest` functionality // add tests to make sure it's always functional. It is documented as functional, but we run the tests in ci via `tests/self_peertest.rs`, so I'm not convinced we need to duplicate the test run here...

carver: I'm okay updating the documentation to say it is a WIP, to skip the tests for now.

PR #490, opened 2022-12-09
docker: Save binaries to /usr/bin folder

ogenev:
### What was wrong?
When trying to use the docker image as a building block for another image, executing `./trin` command returns an error because there is a folder with the same name.

### How was it fixed?
Move trin binaries in `/usr/bin/` folder which allows running with `trin` command instead of `./trin`.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,9 +28,9 @@ FROM ubuntu:22.04
 RUN mkdir -p /trin/trin-core/src/assets
 COPY --from=builder /trin/trin-core/src/assets/merge_macc.bin ./trin/trin-core/src/assets/merge_macc.bin
 # copy build artifacts from build stage
-COPY --from=builder /trin/target/release/trin .
-COPY --from=builder /trin/target/release/trin-cli .
+COPY --from=builder /trin/target/release/trin /usr/bin/
+COPY --from=builder /trin/target/release/trin-cli /usr/bin/
 
 ENV RUST_LOG=debug
 
-ENTRYPOINT ["./trin"]
+ENTRYPOINT ["/usr/bin/trin"]


Dockerfile
```diff
@@ -28,9 +28,9 @@ FROM ubuntu:22.04
...
 COPY --from=builder /trin/trin-core/src/assets/merge_macc.bin ./trin/trin-core/src/assets/merge_macc.bin
 # copy build artifacts from build stage
-COPY --from=builder /trin/target/release/trin .
-COPY --from=builder /trin/target/release/trin-cli .
+COPY --from=builder /trin/target/release/trin /usr/bin/
+COPY --from=builder /trin/target/release/trin-cli /usr/bin/
 
 ENV RUST_LOG=debug
 
-ENTRYPOINT ["./trin"]
```
njgheorghita, line 36: Did you read my mind? Haha, I've been dealing with this in my docker image for running trin with a bridge node. Since I use the trin image as the base for the bridge image. But, in my case this pr doesn't help due to ... https://github.com/carver/eth-portal/blob/main/eth_portal/trin.py#L32 ... where it's hardcoded how to launch the executable as `./trin`.

njgheorghita: Of course, we could change how it's hardcoded in `eth-portal`...

njgheorghita (approved):
Actually, now that I think about it, this is no problem. I actually just copy the binary from the root image rather than running it, and I can easily just copy it from its new location.

Dockerfile
ogenev: This issue also breaks `portal-hive` because I can't run trin with `./trin` after we changed it from `./trin-main`. Fluffy also store its binary file in `/usr/bin` which I think is the correct way to deal with it.

PR #492, opened 2022-12-19
Add serde deserialization to header

njgheorghita:
### What was wrong?
The primary pr I'm working on (introducing some bridge capability to trin) has grown quite large. This pr is an attempt to decrease the size of the upcoming pr by extracting some simple value conversion types. These types are going to be shared amongst `block_body.rs` / `header.rs` / `receipts.rs`. The verbose structure of the old conversion fns became quite cumbersome, so I introduced these types to simplify reusing the conversion logic in the different modules.

### How was it fixed?
Added type conversion helper types. Didn't feel like unit tests are necessary for these, since they're being tested implicitly inside `Header`. 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/types/header.rs
+++ b/trin-core/src/types/header.rs
@@ -1,25 +1,25 @@
-use anyhow::anyhow;
-use bytes::Bytes;
-use ethereum_types::{Bloom, H160, H256, U256};
+use ethereum_types::{Bloom, H160, H256, H64, U256};
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
-use serde::{Serialize, Serializer};
-use serde_json::Value;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use ssz::SszDecoderBuilder;
 use ssz_derive::Decode;
 
 use crate::portalnet::types::messages::ByteList;
-use crate::utils::bytes::hex_decode;
+use crate::utils::bytes::{hex_decode, hex_encode};
 
 const LONDON_BLOCK_NUMBER: u64 = 12965000;
 
 /// A block header.
-#[derive(Debug, Clone, Eq, Serialize)]
+#[derive(Debug, Clone, Eq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
 pub struct Header {
     /// Block parent hash.
     pub parent_hash: H256,
     /// Block uncles hash.
+    #[serde(rename(deserialize = "sha3Uncles"))]
     pub uncles_hash: H256,
     /// Block author.
+    #[serde(rename(deserialize = "miner"))]
     pub author: H160,
     /// Block state root.
     pub state_root: H256,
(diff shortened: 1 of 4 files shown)


trin-core/src/types/json_helpers.rs
```diff
@@ -0,0 +1,416 @@
...
+        let mut hashes = vec![];
+        for val in result {
+            let hash = JsonH256::try_from(val)?.into();
+            hashes.push(hash);
+        }
+        Ok(Self(hashes))
+    }
+}
+
+#[allow(clippy::from_over_into)]
```
njgheorghita, line 44: The goal of these conversion functions is to go from `Value` -> `Primitive` (`u8`, `H256`, etc.). This requires some use of a wrapped type. If we listen to clippy and go with `impl From<Value> for Primitive` this will require more layers of wrapped types, eg. `impl From<WrappedValue> for WrappedPrimitive`, imo it's much preferable to minimize the amount of wrapped types we use and just ignore clippy in this one case.

ogenev (changes requested):
I'm afraid this change is not solving the right problem and is adding many boilerplate code lines.

What we would like is to be able to convert a JSON object (in our case response["result"]) to a block header. This would require implementing the `serde::Deserialize` trait on the `Header` struct. Then we will node need any of the `try_value_into_x` functions or any json helpers.

Then we will be able to say: 
```rust
let header: Header = serde_json::from_str(response["result"]).unwrap()
```

trin-core/src/types/header.rs
```diff
@@ -28,44 +29,76 @@ pub struct Header {
     /// Block receipts root.
     pub receipts_root: H256,
     /// Block bloom filter.
-    pub log_bloom: Bloom,
+    pub logs_bloom: Bloom,
     /// Block difficulty.
     pub difficulty: U256,
     /// Block number.
+    #[serde(deserialize_with = "de_hex_to_u64")]
     pub number: u64,
```
ogenev, line 37: We can get rid of `de_hex_to_u64` if we change this type to `ethereum_types::U64`.

trin-core/src/types/header.rs
```diff
@@ -28,44 +29,76 @@ pub struct Header {
...
+    #[serde(serialize_with = "se_hex")]
+    #[serde(deserialize_with = "de_hex_to_vec_u8")]
     pub extra_data: Vec<u8>,
     /// Block PoW mix hash.
     pub mix_hash: Option<H256>,
     /// Block PoW nonce.
-    #[serde(serialize_with = "raw_bytes")]
+    #[serde(deserialize_with = "de_hex_to_option_bytes")]
+    #[serde(serialize_with = "se_raw_bytes")]
     pub nonce: Option<Bytes>,
```
ogenev, line 54: Could we switch this type to `ethereum_types::H64` and get rid of the helper methods? I'm not sure if `Option` makes sense here, is it possible to have a header without nonce?

trin-core/src/types/header.rs
```diff
@@ -28,44 +29,76 @@ pub struct Header {
...
     pub difficulty: U256,
     /// Block number.
+    #[serde(deserialize_with = "de_hex_to_u64")]
     pub number: u64,
     /// Block gas limit.
     pub gas_limit: U256,
     /// Block gas used.
     pub gas_used: U256,
     /// Block timestamp.
+    #[serde(deserialize_with = "de_hex_to_u64")]
```
ogenev, line 43: Same as above, switch to `ethereum_types::U64`.

trin-core/src/types/header.rs
njgheorghita: Off the top of my head, I'm not sure. But we perform a check on this field on lines `117`, and `137`. So I think it's safe to assume that it is possible and we shouldn't change the core field type.

trin-core/src/types/header.rs
njgheorghita: Same as above, I think it's simpler to convert to `u64` here, rather than converting whenever this field is used in the codebase

trin-core/src/types/header.rs
njgheorghita: `Option<H64>` seems to be fine, gets rid of the helper methods, and mirrors what is done [here](https://github.com/vorot93/ethereum-jsonrpc/blob/master/src/types/block.rs#L102)

trin-core/src/types/header.rs
ogenev: Hmm, why do we deserialize from hex to u64 but keep serializing u64 to u64? I think it is better if we are consistent here and use [serde_hex](https://docs.rs/serde-hex/latest/serde_hex/) to ser/de from u64 to hex and the reverse instead of writing our own implementations...

njgheorghita: Well, on one hand we only serialize this to display the header to a user. In which case, I'm ok with showing the int and not the hex value, it's way easier than having to go convert the hex value to figure out the block number. Even web3.py does some formatting to hex values when they're returned to the user.

I also think it's preferable to avoid importing a whole library if we can replace it with a single method...

trin-core/src/types/header.rs
```diff
@@ -28,42 +28,53 @@ pub struct Header {
...
 }
 
-fn raw_bytes<S>(value: &Option<Bytes>, serializer: S) -> Result<S::Ok, S::Error>
+fn se_hex<S>(value: &[u8], serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
 {
-    let value = value.clone().unwrap();
-    serializer.serialize_str(format!("0x{}", hex::encode(&value)).as_str())
+    serializer.serialize_str(format!("0x{}", hex::encode(value)).as_str())
```
ogenev, line 60:
```suggestion
    serializer.serialize_str(hex_encode(value)).as_str())
```

ogenev (approved):
Looks good! :+1:

PR #493, opened 2022-12-20
jsonrpc LocalContent: Return `0x0` when content is not available

ogenev:
### What was wrong?
The Portal hive LocalContent test fails because we return an error instead of `0x0` when local content is not available.

### How was it fixed?
Return `0x0` when local content is not available according to https://github.com/ethereum/portal-network-specs/pull/176.

### To-Do

- [x] Clean up commit history


njgheorghita (approved):
It seems worth adding a quick test to `ethportal-peertest` to cover this case.

PR #494, opened 2022-12-20
Add json deserialization to receipts/block bodies

njgheorghita:
### What was wrong?
Deserializing from jsonrpc responses was necessary for some of our datatypes to support upcoming bridge functionality.

### How was it fixed?
- Added support for `serde::Deserialize` trait where appropriate
- Introduced some helper types to workaround conflicting rlp encodings
- Introduced a `FullHeader` datatype, which can be created from a "get header" response, and extracts all the information needed to generate `BlockBodies` & `Receipts`

### To-Do

- [x] Clean up commit history


trin-core/src/types/block_body.rs
```diff
@@ -222,6 +334,75 @@ pub struct EIP1559Transaction {
     pub s: U256,
 }
 
+#[derive(Eq, Debug, Clone, PartialEq, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct EIP1559TransactionHelper {
```
ogenev, line 339: If I understand the reasoning correctly, the only point for those helper types is to deserialize a transaction?

If this is the case, I think it will be more clear if we add serde's Serialize/Deserialise derive macros on the transaction enum and we can use `#[serde(tag = "type")]` config on the enum. This will convert the enum type to a `type` json field and if we want to rename the fields we can add `#[serde(rename = "0x0")]` (example [here](https://github.com/ogenev/ethportal-api/blob/9b060df420338f1c8aad28500e77a3b13d3404d5/src/types/block_body.rs#L195) ).

This will also require adding Serialize/Deserialize to all enum types, but this is trivial because all ethereum types we use implement them.

That way we will get rid of all helper structs and `Into` traits.

trin-core/src/types/block_body.rs
```diff
@@ -178,6 +180,42 @@ impl Transaction {
     }
 }
 
+impl<'de> Deserialize<'de> for Transaction {
```
ogenev, line 183: Why do we want to only deserialize a transaction? I think it is good practice to implement both Serialize and Deserialize traits.

trin-core/src/types/header.rs
```diff
@@ -180,6 +182,59 @@ impl PartialEq for Header {
...
+/// requests.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct FullHeader {
+    pub header: Header,
+    pub txs: Vec<Transaction>,
+    pub tx_hashes: TxHashes,
+    pub uncles: Vec<H256>,
+}
+
+// Prefer TryFrom<Value> over implementing Deserialize trait, since it's much simpler when
```
ogenev, line 197: I'm not sure if I agree with this. I think adding a lot of  `TryFrom` implementations build up with time and it becomes very hard to follow all the logic. I don't like the pattern of first converting json field to `Value` and then converting the Value to a type. I guess a more good practice is to directly convert a json string to type using Serialize/Deserialize.

If we have a json object which consists of multiple types, I think we should create a rust struct for the whole json object instead of implementing structs only for the subtypes and then doing some fancy conversion field by field to a Value type and then converting from Value to a type struct.

njgheorghita:
Moving ahead with this merge & not addressing the comments, even though I agree with them, as this pr is a minor blocker. Since #495 has landed, and has already incorporated the suggestions mentioned above by @ogenev, there's no need to replicate those changes in this pr. Since this pr is functional, we can move ahead with this merge, and as we incorporate the `ethportal-api` types into core `trin` functionality, we will remove any unnecessary helpers / fns that were added in this pr.

ogenev (approved):

PR #496, opened 2023-01-04
jsonrpc: Return a boolean instead of a string from the `portal_*Store endpoint`

ogenev:
### What was wrong?
According to JSON-RPC [specs](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false), Trin should return a boolean instead of a  string from the `portal_*Store` endpoint.

### How was it fixed?
Return JSON boolean.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -360,7 +360,7 @@ fn validate_portal_find_nodes_zero_distance(result: &Value, peertest: &Peertest)
 }
 
 fn validate_portal_store(result: &Value, _peertest: &Peertest) {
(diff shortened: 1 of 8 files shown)


jacobkaufmann (approved):

njgheorghita (approved):
🚢

PR #497, opened 2023-01-05
circleci: Add no_output_timeout value to docker-build job

ogenev:
### What was wrong?

The Docker build pipeline fails on the master branch.

### How was it fixed?
Increases the no_output_timeout ` setting.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,12 +10,14 @@ executors:
       - image: cimg/rust:1.62.0
 jobs:
   docker-build:
+    resource_class: xlarge
     executor: docker-publisher
     steps:
       - checkout
       - setup_remote_docker
       - run:
           name: Build Docker image
+          no_output_timeout: 30m
           command: docker build -t $IMAGE_NAME:latest .
       - run:
           name: Archive Docker image


njgheorghita:
> it takes around 25 min to build the image even on a powerful CircleCI instance
Are there any other alternatives? I'm not sure I'm on board with this much downtime.

njgheorghita:
> Only editing the Dockerfile is critical for the builds 
Now that I think about it, I'm not sure this is true... I've definitely broken the docker build before by changing some internal trin code (I believe it was an invalid path for loading the master acc) which caused every testnet node to crash. However, I'm not sure if the current `docker-build` step would catch that, since the error didn't break "building" the docker image, but broke running the docker image...

What do you think about adding a step to the docker ci step that ensures you can launch a trin node from the docker image and ping it... Although, this is exactly what portal-hive does, so it seems we should be able to catch any similar errors sufficiently fast enough via portal-hive... in which case (if you agree) i'm ok with `"keep the no_output_timeout to fix the master failures and proceed with caution when changing the dockerfile?"`

ogenev:
> What do you think about adding a step to the docker ci step that ensures you can launch a trin node from the docker image and ping it...

Yeah, I think portal-hive is doing that, and adding another job for pinging trin is unnecessary complexity.

njgheorghita (approved):
:shipit:

PR #498, opened 2023-01-09
Bridge mode

njgheorghita:
### To-Do

- [ ] Clean up commit history


jacobkaufmann (commented):
Before further review, why did you decide to make this a flag on the existing `trin` binary as opposed to its own separate binary?

njgheorghita:
Copying my comments from Discord here... 

> 1. Generating a `HeaderWithProof` requires access to the `HeaderOracle` in order to lookup the epoch accumulator (whether it's available locally, or sourced from the history network) and then generate the proof. To generate proofs with the "separate" architecture will require some extra complexity - either in the jsonrpc api (eg some kind of special `Offer` endpoint) or internal logic inside `trin-history` that will automatically add the proof. 

> 2. Trin's bridge takes advantage of rust's nice async api, and spawns many tasks concurrently. I'm not confident that our jsonrpc server can handle this (at least until we complete the transition to `jsonrpsee`) without sacrificing the gains from executing async tasks.

> I'd like to re-endorse my preference for leaving trin's `bridge` as "native" (as how it exists in the current pr). For the reasons above, along with others mentioned during the call. It reduces the overall complexity of the code. Imo, a single, moderately-sized file (`bridge.rs`) does not necessarily constitute "bloat" . We can always extract it down the road, if the necessity proves worthwhile, but atm I feel like it's perfectly viable to leave the `bridge` functionality as "native"

And adding some clarifications...

1. There are two ways I can see us doing this. Imo, neither is ideal. 
	1. Add a custom jsonrpc endpoint (eg `portal_historyOfferHeader`). Since we only get a `Header` from geth - this endpoint will take the `Header` as a param - and then generate a proof (using the `HeaderOracle`) before offering the `HeaderWithProof` to the network. My thoughts... Do we really want to add yet another custom `Offer` endpoint to our api?
	2. Inside the jsonrpc handling for `portal_historyOffer` - we add some custom logic to detect an eligible`Header`, and if so, we generate and attach a proof. But now that I type this out, this obfuscates some fairly significant logic, and I don't see this as a legitimate option anymore.
2. This bridge implementation fetches `128` geth headers across `8` spawned tasks concurrently. This repeats, until it has an epoch's worth of headers. Then it (more or less) repeats this process to fetch uncles & receipts. Within these concurrent tasks, the content is offered to the network over jsonrpc. 
- Our jsonrpc server - in its current state (pre #ogneev ) is fragile. I come across broken pipes fairly frequently, even with sending single commands. I'm skeptical that our jsonrpc server can withstand the assault of offer commands. Now, transitioning to `jsonpsee` will stabilize our server, and add support for batch commands. I don't see the benefit of wasting time, wrangling with a fussy jsonrpc server (in its current state), when we have a more stable implementation coming down the pipeline. This isn't an argument against the merits of extracting `bridge` into an "external" service, just an argument that it's not worth extracting until our jsonrpc server stabilizes.

njgheorghita:
@ogenev 

I do appreciate the time you took to type those out, thank you. I think it will be better if we continue this discussion during sync tomorrow, but for now / posterity, i'll jot down some quick thoughts.

1. Ok, but fluffy is also implementing a "native" bridge. So then, is the effort of a client-agnostic bridge worth it just for "ultralight"?
2. Every bridge node design we've considered is accompanied by a running node. Maybe this idea would be useful down the road (like years from now), but it doesn't even come close to a priority / important use case to consider imo.
3. ...
4. Yes, I'm against duplicating code, but I'm also hesitant to extract every little piece of `trin-core` into it's own package - until it's necessity is sufficiently justified. In the case of `ethportal-api` with `glados`, I believe it was sufficiently justified. In this case, I'm not so sure I agree, but I definitely don't see it as a priority.
5. Same as number 4
6. I don't agree. We expect all trin users to have some familiarity with the command line. This is how they are expected to manage allocated machine resources. 
7. I haven't thought this one through yet... but I don't see why this would be a problem with the "native" design.
8. Again, I don't see why this would be a problem with the "native" design

Ultimately, I think it's best if we continue this discussion in sync tomorrow, I find it much easier to come to a consensus via dialogue. But if I were to summarize my opinion on this matter...

Perfect is the enemy of good. "Native" bridge mode is not intended (yet) for users. We don't have users. Even if we did, I don't see a strong argument for why we should build it for their use. Bridge architecture should be considered "in-house" at least for this year until we can deliver an mvp. It should be fast, and functional. Our goal is to build a functional mvp network. Not a pretty, easy-to-use client for end users that supports their every dream (eg running their own bridge).

carver:
Following up from our 1-1 convo, do you still feel good about merging this into a branch instead of master? If so, you can create a branch and modify this PR or open a new PR against it. It's still nice to get a quick look at the code, but it doesn't have to be held to the same standard.

I also remember that you wanted to rebase on top of some other PR that was in progress. Which one was that?

njgheorghita:
@carver My thinking was that I would just leave this branch (`bridge`) as the main branch to track bridge development. All future updates / bugfixes to the bridge would be opened up in a pr against this branch. And then, when / if we're ready to merge bridge functionality into `master`, I'll open up that pr and the bridge code can be reviewed then.

carver:
I don't like that PRs against it would be in the "wrong" repo, and that releases would be coming from your personal fork. I'm okay with it as a temporary situation. If we're still using your repo branch in 3-6 months, then the priority to move it to an ethereum/trin branch will go up a lot, from my perspective.

njgheorghita:
@carver Sorry, I misunderstood your suggestion. No problem moving this into it's own branch. Closing this pr in favor of #552

PR #506, opened 2023-01-17
Remove ByteList as type for offered content

njgheorghita:
### What was wrong?
Using `ByteList` as the type for content we offer was a mistake... `ByteList` is just an alias for `VariableList` with max length `2048` - so every piece of content we offered was truncated to 2048... which obviously meant that it didn't pass validation on the receiving side, and is likely the reason for poor testnet performance recently.

### How was it fixed?
Changed use of `ByteList` -> `Vec<u8>`

Why didn't testing catch this? Well, utp testing isn't covered extensively inside trin, and that will be covered in `portal-hive` scenario testing. @ogenev (who helped debug this, thanks!) I would just make sure to document in some todo-list that this situation is one we should definitely cover as soon as `portal-hive` is capable of executing this type of test.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/506.fixed.md
+++ b/newsfragments/506.fixed.md
@@ -1,0 +1,1 @@
+Replaced ByteList with Vec<u8> as type for offered content values.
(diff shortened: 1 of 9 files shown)


jacobkaufmann (approved):

mrferris (approved):
Good catch.

Definitely an issue for receipts/bodies, though I don't expect it to fix issues with header gossip because headers are <2048 bytes.

Issue #509, opened 2023-01-19
Requires Inufra API


perama-v:
### Description

Currently the client requires information from an execution node for some content, and so requires a connection to Infura.
It would be good to be able to connect to a node that is running locally.

perama-v:
closed by #510

PR #510, opened 2023-01-19
add CLI flag and docs for local trusted provider

perama-v:
### What was wrong?

As noted in #509, Trin sometimes requires a connection to a trusted provider, however was no alternative to 
Infura. 

### How was it fixed?

Added:
- CLI flag `--trusted-provider`
    - New option `custom` 
    - Altered option `geth` -> `pandaops`
- `Url` library (see below)
- CLI flag `--trusted-provider-url` 
     - Accepts a url, such as `http://127.0.0.1:8545` that is parsed with `Url` library
     - Is required for the `custom` and `pandaops` trusted provider arguments
- Documentation in `./docs/getting_started.md`
- Flag testing for invalid combinations
- Requirement that `--trusted-provider custom` choice requires the `--trusted-provider-url`.
- ~~The ability to pass a either `http://127.0...` or `127.0...`~~. Urls are parsed for correctness, and must have a base otherwise the CLI parser will panic and return an appropriate parsing message.
- Url library is also used for `web3-http-address`
- Port clash detection for trusted provider (non-trin) and trin web3 ports.

### Validation
Started execution node on port 8545.

Started Trin with the following flags:
```sh
RUST_LOG=debug cargo run -- \
    --trusted-provider custom \
    --trusted-provider-url http://127.0.0.1:8545 \
    --web3-http-address http://127.0.0.1:8547 \
    --web3-transport http \
    --discovery-port 9009 \
    --bootnodes default \
    --kb 200000 \
    --no-stun
```
Called `eth_blockNumber` on port 8545 (Execution node's port) and received a response.

Called `eth_blockNumber` (requires trusted provider) on port 8547 (Trin's port) and received a response as follows:

```sh
curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc": "2.0", "method": "eth_blockNumber", "params": [], "id":1}' localhost:8547

> {"jsonrpc":"2.0","id":1,"result":"0xfada96"} 
```
Which is block `16_439_958`.
### Interface
Running `cargo run -- --help`, the changed and new flag docs respectively are shown:
```sh
...
        --trusted-provider <trusted-provider>
            Trusted provider to use. (options: 'infura' (default), 'pandaops' (devops) or 'custom') [default: infura]

        --trusted-provider-url <trusted-provider-url>
            URL for a trusted http provider. Must include a base, host and port (e.g., '<base>://<host>:<port>').
...
```
## Panics
The following panic scenarios are introduced to enforce correct flag use:

With: `cargo run -- --trusted-provider-url http://127.0.0.1:8546`
```sh
thread 'main' panicked at '--trusted-provider-url flag is incompatible with infura as the trusted provider.', trin-core/src/cli.rs:194:48
```
With `cargo run -- --trusted-provider custom`
```sh
thread 'main' panicked at ''--trusted-provider custom' choice requires the --trusted-provider-url flag.', trin-core/src/cli.rs:202:48
```
With `cargo run -- --trusted-provider custom --trusted-provider-url http://127.0.0.1:8545 --web3-transport http`
```sh
thread 'main' panicked at '--trusted-provider-url and --web3-http-address cannot have the same localhost port.', trin-core/src/cli.rs:215:21
```
# To-Do

- [x] Clean up commit history
- [x] Run cargo clippy and cargo fmt 
- [x] Get unit tests passing 
- ~~[ ] Run peertest~~ (deprecated)

### DIFF
--- a/trin-core/src/cli.rs
+++ b/trin-core/src/cli.rs
@@ -1,13 +1,14 @@
 use std::{env, ffi::OsString, fmt, net::SocketAddr, path::PathBuf};
 
 use structopt::StructOpt;
+use url::Url;
 
 use crate::portalnet::types::messages::HexData;
 use crate::utils::provider::TrustedProviderType;
 
 pub const DEFAULT_MASTER_ACC_PATH: &str = "src/assets/merge_macc.bin";
 pub const DEFAULT_WEB3_IPC_PATH: &str = "/tmp/trin-jsonrpc.ipc";
-pub const DEFAULT_WEB3_HTTP_ADDRESS: &str = "127.0.0.1:8545";
+pub const DEFAULT_WEB3_HTTP_ADDRESS: &str = "http://127.0.0.1:8545/";
 const DEFAULT_DISCOVERY_PORT: &str = "9000";
 pub const HISTORY_NETWORK: &str = "history";
 pub const STATE_NETWORK: &str = "state";
@@ -36,7 +37,7 @@ pub struct TrinConfig {
         long = "web3-http-address",
         help = "address to accept json-rpc http connections"
     )]
-    pub web3_http_address: String,
+    pub web3_http_address: Url,
 
     #[structopt(
         default_value(DEFAULT_WEB3_IPC_PATH),
@@ -118,16 +119,17 @@ pub struct TrinConfig {
 
     #[structopt(
         long = "trusted-provider",
-        help = "Trusted provider to use. (options: 'infura' (default) or 'geth' (experimental))",
+        help = "Trusted provider to use. (options: 'infura' (default), 'pandaops' (devops) or 'custom')",
         default_value(DEFAULT_TRUSTED_PROVIDER)
     )]
     pub trusted_provider: TrustedProviderType,
 
     #[structopt(
-        long = "geth-url",
-        help = "URL for geth client to use as trusted http provider (experimental)"
+        long = "trusted-provider-url",
+        help = "URL for a trusted http provider. Must include a base, host and port (e.g., '<base>://<host>:<port>').",
+        validator(check_url_format)
     )]
-    pub geth_url: Option<String>,
+    pub trusted_provider_url: Option<Url>,
 
     #[structopt(
         long = "master-accumulator-path",
@@ -142,7 +144,7 @@ impl Default for TrinConfig {
     fn default() -> Self {
         TrinConfig {
             web3_transport: "ipc".to_string(),
-            web3_http_address: DEFAULT_WEB3_HTTP_ADDRESS.to_string(),
+            web3_http_address: Url::parse(DEFAULT_WEB3_HTTP_ADDRESS).unwrap(),
             web3_ipc_path: DEFAULT_WEB3_IPC_PATH.to_string(),
             pool_size: 5,
             discovery_port: DEFAULT_DISCOVERY_PORT.parse().unwrap(),
@@ -158,7 +160,7 @@ impl Default for TrinConfig {
             enable_metrics_with_url: None,
             ephemeral: false,
             trusted_provider: TrustedProviderType::Infura,
-            geth_url: None,
+            trusted_provider_url: None,
             master_acc_path: PathBuf::from(DEFAULT_MASTER_ACC_PATH.to_string()),
         }
     }
@@ -180,26 +182,50 @@ impl TrinConfig {
                 DEFAULT_WEB3_IPC_PATH => {}
                 _ => panic!("Must not supply an ipc path when using http protocol for json-rpc"),
             },
-            "ipc" => match &config.web3_http_address[..] {
+            "ipc" => match config.web3_http_address.as_str() {
                 DEFAULT_WEB3_HTTP_ADDRESS => {}
-                _ => panic!("Must not supply an http address when using ipc protocol for json-rpc"),
+                p => panic!("Must not supply an http address when using ipc protocol for json-rpc (received: {p})"),
             },
             val => panic!("Unsupported json-rpc protocol: {}", val),
         }
 
-        match config.geth_url {
+        match config.trusted_provider_url {
             Some(_) => {
-                if config.trusted_provider != TrustedProviderType::Geth {
-                    panic!("--geth-url flag is incompatible without Geth as the trusted provider.")
+                if config.trusted_provider == TrustedProviderType::Infura {
+                    panic!("--trusted-provider-url flag is incompatible with infura as the trusted provider.")
                 }
             }
-            None => {
-                if config.trusted_provider == TrustedProviderType::Geth {
-                    panic!("Using Geth for a trusted provider requires a --geth-url flag.")
+            None => match config.trusted_provider {
+                TrustedProviderType::Infura => {}
+                TrustedProviderType::Pandaops => panic!(
+                    "'--trusted-provider pandaops' choice requires the --trusted-provider-url flag."
+                ),
+                TrustedProviderType::Custom => panic!(
+                    "'--trusted-provider custom' choice requires the --trusted-provider-url flag."
+                ),
+            },
+        }
+        // Should not serve http over same port as localhost provider.
+        if config.web3_transport.as_str() == "http"
+            && config.trusted_provider == TrustedProviderType::Custom
+        {
+            if let Some(url) = &config.trusted_provider_url {
+                let is_local_provider = url.host_str() == Some("127.0.0.1");
+                let port_clash = url.port() == config.web3_http_address.port();
+                if is_local_provider && port_clash {
+                    panic!("--trusted-provider-url and --web3-http-address cannot have the same localhost port.")
                 }
             }
         }
         Ok(config)
+    }
+}
+
+/// A validator function for CLI URL arguments.
+fn check_url_format(url: String) -> Result<(), String> {
+    match Url::parse(&url) {
+        Ok(_) => Ok(()),
+        Err(e) => panic!("Invalid URL '{url}', {e}"),
     }
 }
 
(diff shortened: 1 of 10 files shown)


perama-v:
New commits:
- [04fb4d1](https://github.com/ethereum/trin/pull/510/commits/04fb4d15bedc6ff2948a60c8f89e16236512b1df) fixes a bug introduced in the new check for port clashes. It skips port clash check if using IPC
- [50440ea](https://github.com/ethereum/trin/pull/510/commits/50440eab98f9dbafbca4457edc222c9216ab722c) fixes a bug where the tests falsely fail when using a local provider. The fail arises from an assertion that the env contains an infura key. Now the infura key check is not asserted if the provider flag argument is Local or Geth.

njgheorghita:
Thanks for the pr! It's nicely documented & tested. With the addition of a `local` flag, I think it introduces some confusion w/ regards to the `geth` flag. Since, a geth node can just as well be run locally and used under the `local` flag. The use of the `geth` flag is really for "internal" users only, since it's designed to be compatible with some, private, provisioned nodes that our devops team exposed for us. So, I'd recommend adding a commit to update the `geth` flag name to `pandaops` (which is the internal name for our node service), and updating the relevant documentation

perama-v:
Great idea - done

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -69,7 +69,6 @@ pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
     let received_content_str = received_content_value.as_str().unwrap();
     assert_eq!(
         HISTORY_CONTENT_VALUE, received_content_str,
-        "The received content {}, must match the expected {}",
-        HISTORY_CONTENT_VALUE, received_content_str,
+        "The received content {HISTORY_CONTENT_VALUE}, must match the expected {received_content_str}",
```
njgheorghita, line 72: These two values should actually be switched around

trin-core/src/cli.rs
```diff
@@ -184,20 +191,41 @@ impl TrinConfig {
                 DEFAULT_WEB3_HTTP_ADDRESS => {}
                 _ => panic!("Must not supply an http address when using ipc protocol for json-rpc"),
             },
-            val => panic!("Unsupported json-rpc protocol: {}", val),
+            val => panic!("Unsupported json-rpc protocol: {val}"),
         }
 
-        match config.geth_url {
+        match config.pandaops_url {
```
njgheorghita, line 197: I haven't done a thorough review of the logic here, but it seems to me like some of this logic can be handled by [clap](https://docs.rs/clap/latest/clap/builder/struct.Arg.html#method.requires_if) helpers.

njgheorghita (commented):
Sorry for taking a while to get back to this. I'd check out the clap helpers I mentioned, in case you haven't seen those yet. In fact, there might be some other arg parsing logic that we can remove with these helpers. 

Also, I think it's ok to leave it in for this pr. But, something like the formatting updates (eg. `format!("{}", var)` -> `format!("{var}")`) would be better off in a separate pr, including an update to the style guide in `docs/contributing.md`. Keeping it separate helps avoid bloat in this pr and makes it easier to review rather than tacking on all those changes here.

perama-v:
> Also, I think it's ok to leave it in for this pr. But, something like the formatting updates (eg. `format!("{}", var)` -> `format!("{var}")`) 

Ok, I worked out where the madness came from: I had run clippy on nightly. So I've reverted those changes now.

> Keeping it separate helps avoid bloat in this pr and makes it easier to review rather than tacking on all those changes here.

A good point, I'll keep this in mind

ethportal-peertest/src/scenarios/offer_accept.rs
perama-v: I think you're right. This was a nightly-clippy change so after reverting I will leave it untouched for now.

trin-core/src/cli.rs
perama-v: I really like the concept but ran into some trouble and so reverted the changes after not finding a solution. 
## Tests that should pass, fail
When I replaced the manual checks with clap helpers, I noticed that the unit tests that require a `should_panic` were not capturing the panic that clap triggers. 
1. A test had an incorrect flag combination 
2. Triggered a clap helper requirement 
3. Code panics (good)
4. Test (`should_panic`) does not pass (bad)

It seems that Clap triggers an early exit before the test can properly assess things. I also tried a `panic::catch_unwind(|| {})` based attempt, but no result. 
## Opacity during triggering
A separate but also important thing is that the panic messages Clap produced were without context and hard to determine where they came from (e.g,. because tests are run multithreaded, it would dump a the following (out of order, making it appear to be from a different (passing) test. 

```sh
test cli::test::test_http_protocol_rejects_custom_web3_ipc_path - should panic ... ok
error: The following required arguments were not provided:
    --local-provider-url <local-provider-url>

USAGE:
    trin --discovery-port <discovery-port> --kb <kb> --local-provider-url <local-provider-url> --master-accumulator-path <master-acc-path> --networks <networks>... --pandaops-url <pandaops-url> --pool-size <pool-size> --trusted-provider <trusted-provider> --web3-http-address <web3-http-address> --web3-ipc-path <web3-ipc-path> --web3-transport <web3-transport>
For more information try --helptest cli::test::test_ipc_with_custom_path ... ok
```
Running the test in single-thread mode allows the ordering to be correct. The error messages are still quite terse relative to the custom message, and are less useful to the end user.
## Outcome
I have kept the conditionals-based logic that I had previously

trin-core/src/cli.rs
```diff
@@ -494,59 +525,134 @@ mod test {
...
         )
         .unwrap();
-        assert_eq!(config.trusted_provider, TrustedProviderType::Geth);
+        assert!(env_is_set(&config));
+        assert_eq!(config.trusted_provider, TrustedProviderType::Pandaops);
+    }
+
+    #[test]
+    fn test_local_node_trusted_provider() {
+        use crate::utils::provider::TrustedProvider;
```
njgheorghita, line 552: I would move this up to the other test import statements around line `265`

trin-core/src/cli.rs
```diff
@@ -187,17 +194,38 @@ impl TrinConfig {
...
+            && config.trusted_provider != TrustedProviderType::Local
+        {
+            panic!("--local-provider-url flag is incompatible without 'local' as the trusted provider.")
+        }
+        if config.trusted_provider == TrustedProviderType::Local
+            && config.local_provider_url.is_none()
+        {
+            panic!("'--trusted-provider local' choice requires the --local-provider-url flag.")
+        }
+        match &config.local_provider_url {
```
njgheorghita, line 219: I think we can take advantage of the [url](https://docs.rs/url/2.3.1/url/index.html) library here (which `ureq` uses). This will help eliminate some tricky edge cases with matching schemes, domains, ports, etc.

trin-core/src/utils/provider.rs
```diff
@@ -112,31 +121,39 @@ fn build_infura_ws_url_from_env() -> String {
...
+    let client_id = get_pandaops_client_id_from_env();
+    let client_secret = get_pandaops_client_secret_from_env();
+    ureq::post(&pandaops_url)
         .set("Content-Type", "application/json")
         .set("CF-Access-Client-Id", client_id.as_str())
         .set("CF-Access-Client-Secret", client_secret.as_str())
 }
+
+fn build_local_node_http_client(node_url: String) -> ureq::Request {
+    let url = match node_url.starts_with("127.0.0.1") {
```
njgheorghita, line 154: I'd also use the [url](https://docs.rs/url/2.3.1/url/index.html) library here

trin-core/src/cli.rs
```diff
@@ -118,16 +118,22 @@ pub struct TrinConfig {
...
+        help = "URL for pandaops client to use as trusted http provider (experimental)"
     )]
-    pub geth_url: Option<String>,
+    pub pandaops_url: Option<String>,
+
+    #[structopt(
+        long = "local-provider-url",
+        help = "URL for local Ethereum execution client as '<ip_address>:<port>' (experimental)."
+    )]
+    pub local_provider_url: Option<String>,
```
njgheorghita, line 136: Any reason not to combine `pandaops-url` and `local-provider-url` into a `trusted-provider-url`? I think it would help clean up some of the following parsing logic. eg.
- if `infura` -> no `trusted-provider-url` is allowed
- else if `pandaops` -> validate `trusted-provider-url` is a valid url (using the `url` library)
- else if `local` -> validate `trusted-provider-url` is valid and perform the port collision check

njgheorghita (commented):
just a few more thoughts, lmk what you think

trin-core/src/cli.rs
perama-v: This makes sense. Not done initially to minimise disrupting the infura/pandaops logic

trin-core/src/cli.rs
perama-v: Good call

trin-core/src/cli.rs
perama-v: Done

trin-core/src/cli.rs
perama-v: Done

trin-core/src/cli.rs
perama-v: Done. I have changed all instances of the word "local" to "custom", because one is not limited to using a local node with this flag (it could be also be personal remote node)

trin-core/src/utils/provider.rs
perama-v: done :)

njgheorghita:
Nice! Looks good to me, thanks for taking care of all the comments!

perama-v:
Cheers 🥂

PR #514, opened 2023-01-20
Update master accumulator, replace Header with HeaderWithProof

njgheorghita:
### What was wrong?
Implement the change from using `Header` -> requiring `HeaderWithProof` - https://github.com/ethereum/portal-network-specs/pull/177

### How was it fixed?
- Replace `Header` with `HeaderWithProof`
- Update master accumulator to allow proofs to be generated for each pre-merge header
- Removed ultralight test assets, since those are no longer compatible with new accumulator
- Removed all logic for building / updating a master accumulator from `accumulator.rs` since those are no longer required, with a "finalized" master accumulator

### To-Do

- [x] Clean up commit history


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -975,7 +975,7 @@ where
             Ok(Some(content)) => {
                 // Check content size and initiate uTP connection if the size is over the threshold
                 // TODO: Properly calculate max content size
-                if content.len() < 1000 {
+                if content.len() <= 1028 {
```
njgheorghita, line 978: This increase in size was required so that the `eth_getBlockByHash` scenario in `ethportal-peertest` doesn't trigger a utp transfer. It seemed like an alright change imo, since it's only marginally larger, and we still have the todo to properly calculate max content size

ethportal-peertest/src/scenarios/eth_rpc.rs
```diff
@@ -25,50 +27,11 @@ pub fn test_eth_get_block_by_hash(_peertest_config: PeertestConfig, peertest: &P
...
+            json!("0x6251d65b8a8668efabe2f89c96a5b6332d83b3bbe585089ea6b2ab9b6754f5e9"),
             json!(false),
         ]),
     };
 
     let result = make_ipc_request(&peertest.nodes[0].web3_ipc_path, &request).unwrap();
-    assert_eq!(json!(15537393), result["number"]);
-}
-
-pub fn test_eth_get_block_by_number(_peertest_config: PeertestConfig, peertest: &Peertest) {
```
njgheorghita, line 37: I removed this test because the epoch accumulator was causing strange bugs, due to it's large size. A potential cause for this error could be our current jsonrpc server implementation, since the error came from `jsonrpc/service.rs`. I spent some time debugging, but didn't consider it to be worthwhile pursuing. Since @ogenev plans to remove support for these two scenarios, I figured we can just remove this test for now - and it will be re-introduced once we have an improved jsonrpc server / support for this endpoint

trin-core/src/types/accumulator.rs
```diff
@@ -116,45 +82,36 @@ impl MasterAccumulator {
...
+            }
+            BlockHeaderProof::None(_) => {
+                if hwp.header.number <= MERGE_BLOCK_NUMBER {
+                    return Err(anyhow!("Missing accumulator proof for pre-merge header."));
+                } else {
+                    // Skip validation for post-merge headers until proof format is finalized
+                    return Ok(());
+                }
+            }
         };
```
jacobkaufmann, line 101: nit: we could clean this up by rearranging the conditional statements

example:

```rust
if hwp.header.number > MERGE_BLOCK_NUMBER {
    if let BlockHeaderProof::AccumulatorProof(..) = hwp.proof {
        return Err(..);
    }
    return Ok(());
}

if let BlockHeaderProof::AccumulatorProof(proof) = hwp.proof {
    // validation...
} else {
    Err(..)
}
```

jacobkaufmann (commented):
looks good! a few minor questions left in comments

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -22,13 +22,11 @@ use trin_core::{
...
 /// Default enr seq value
 const ENR_SEQ: &str = "1";
-/// History header content key & value for merge block on mainnet
-/// Use merge block here so that we can use the default master accumulator for validation
-/// rather than needing to go lookup an epoch accumulator.
+/// History HeaderWithProof content key & value
 pub const HISTORY_CONTENT_KEY: &str =
-    "0x0055b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286";
+    "0x006251d65b8a8668efabe2f89c96a5b6332d83b3bbe585089ea6b2ab9b6754f5e9";
 pub const HISTORY_CONTENT_VALUE: &str =
```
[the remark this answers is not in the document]
njgheorghita: The reason we used the merge block before, was due to our structure of the `MasterAccumulator`, we could verify the merge block directly without needing to rely on an `EpochAccumulator`. Actually, I believe the choice for this block was even before we had support for validation via an accumulator proof. Basically, our validation required the `EpochAccumulator` in which it looked up the block hash directly and made sure they matched. Now, all headers have proofs, and we no longer require an `EpochAccumulator` for validation. This value is just one of the test vectors from `trin-core/src/assets/test/fluffy/header_with_proofs.json` since I didn't feel there was a need to generate a `HeaderWithProof` content value for the merge block. I added the block number for reference

ethportal-peertest/src/scenarios/eth_rpc.rs
```diff
@@ -7,6 +8,7 @@ use crate::{
 use trin_core::jsonrpc::types::Params;
 
 pub fn test_eth_get_block_by_hash(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    info!("Testing eth_getBlockByHash");
```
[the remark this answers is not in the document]
njgheorghita: I used it for debugging, but left it in purposefully since I think it's useful. All of our test runs in `jsonrpc.rs` have `info` logs stating which test is currently being run. Keeping this `info` log here helps communicate which test is currently being run / failing

trin-core/src/types/header.rs
```diff
@@ -244,7 +244,30 @@ pub struct HeaderWithProof {
...
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
+
+    fn ssz_append(&self, buf: &mut Vec<u8>) {
+        let header = rlp::encode(&self.header).to_vec();
+        let header: ByteList = ByteList::from(header);
+        let offset =
+            <ByteList as Encode>::ssz_fixed_len() + <AccumulatorProof as Encode>::ssz_fixed_len();
+        let mut encoder = SszEncoder::container(buf, offset);
```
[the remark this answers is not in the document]
njgheorghita: I'll leave it in for now. But, I'm not so convinced that intermediate types is a cleaner api, it can just be confusing to me to have intermediate / helper / conversion types lying around amongst the actual type definitions. Maybe that's just me, but simply encapsulating any conversion logic within the trait seems more clean-cut. But we can leave that discussion for a later pr

trin-core/src/types/accumulator.rs
njgheorghita: I don't think it's necessary, we're going to have some nested logic regardless (in your example, nesting all of the validation logic.

ogenev (approved):
Looks good to me :boat: .

trin-core/src/types/accumulator.rs
```diff
@@ -308,6 +232,12 @@ pub struct HeaderRecord {
     pub total_difficulty: U256,
 }
 
+fn calculate_generalized_index(header: &Header) -> usize {
+    // Calculate generalized index for header
+    let hr_index = header.number as usize % EPOCH_SIZE;
+    (EPOCH_SIZE * 2 * 2) + (hr_index * 2)
```
perama-v, line 238: It's a bit unclear what the calculation is here. 

```rs
(EPOCH_SIZE * 2 * 2) + (hr_index * 2)
```

Otherwise, all looks good to me. Tested locally, no issues found.

ethportal-peertest/src/scenarios/eth_rpc.rs
```diff
@@ -25,50 +27,11 @@ pub fn test_eth_get_block_by_hash(_peertest_config: PeertestConfig, peertest: &P
...
-        id: 11,
-        params: Params::Array(vec![
-            // Block #15537395
-            json!("0xed14f3"),
-            json!(false),
-        ]),
-    };
-
-    let result = make_ipc_request(&peertest.nodes[0].web3_ipc_path, &request).unwrap_err();
-    assert_eq!(result.to_string(), "JsonRpc response contains an error: String(\"Error while processing eth_getBlockByNumber: eth_getBlockByNumber not currently supported for blocks after the merge (#15537393)\")");
```
carver, line 73: I imagine this part of the test still works, and doesn't need to be thrown out with the bathwater.

ethportal-peertest/src/scenarios/eth_rpc.rs
carver: Can you add a tracking issue?

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -70,6 +70,6 @@ pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
     assert_eq!(
         HISTORY_CONTENT_VALUE, received_content_str,
         "The received content {}, must match the expected {}",
-        HISTORY_CONTENT_VALUE, received_content_str,
+        received_content_str, HISTORY_CONTENT_VALUE
```
carver, line 73: Nice. I bet that was annoying when debugging X)

tests/self_peertest.rs
```diff
@@ -59,10 +59,6 @@ mod test {
             peertest_config.clone(),
             &peertest,
         );
-        peertest::scenarios::eth_rpc::test_eth_get_block_by_number(
-            peertest_config.clone(),
-            &peertest,
-        );
```
carver, line 65: Just making sure this gets added back when the last part of the test gets added again.

trin-core/src/portalnet/overlay_service.rs
carver: Always nice to link the specific issue when mentioning it in the PR :)

carver (approved):
Looks good! Some suggestions

trin-core/src/portalnet/overlay_service.rs
njgheorghita: https://github.com/ethereum/trin/issues/523

ethportal-peertest/src/scenarios/eth_rpc.rs
njgheorghita: #524

PR #521, opened 2023-01-30
circleci: Remove windows pipeline

ogenev:
### What was wrong?

CircleCI windows pipeline is unstable and the general consensus from the Discord conversation is that we don't need to run it at this stage.

### How was it fixed?
- Remove win-build pipeline from CIrcleCI config

### To-Do

- [ ] Clean up commit history


njgheorghita:
We do have some logic throughout our codebase that was solely responsible for handling windows environments, like [here](https://github.com/ethereum/trin/blob/master/trin-core/src/jsonrpc/service.rs#L122) and [here](https://github.com/ethereum/trin/blob/cf18e67f7840c8bb4ff02cca050e01c4fa706be2/trin-core/src/socket.rs#L7) - just searching for "windows" in the codebase will reveal a couple others. It might be worth removing all of those in this pr, just to maintain order, especially while we deal with the jsonpsee refactor. What do you think?

ogenev:
> We do have some logic throughout our codebase that was solely responsible for handling windows environments, like [here](https://github.com/ethereum/trin/blob/master/trin-core/src/jsonrpc/service.rs#L122) and [here](https://github.com/ethereum/trin/blob/cf18e67f7840c8bb4ff02cca050e01c4fa706be2/trin-core/src/socket.rs#L7) - just searching for "windows" in the codebase will reveal a couple others. It might be worth removing all of those in this pr, just to maintain order, especially while we deal with the jsonpsee refactor. What do you think?

I think some of them may be useful in the future when we want to bring back windows support, so I prefer to leave them for now. IMO, removing the CI pipeline is enough to ease our daily workflow and not think about windows for a while.

README.md
```diff
@@ -6,6 +6,10 @@ The Portal Network is still in the research phase, and this client is experiment
 
 **Do not rely on Trin in a production setting.**
 
+**NOTE: Unix-only**
+
+Trin only runs on Unix-based platforms (Linux, macOS). Do not expect any help for issues on Windows.
```
njgheorghita, line 11: Maybe something more like `Trin currently only runs on Unix-based platforms (Linux, macOS). We plan to eventually implement support for Windows, but until then do not expect any support for issues on Windows.`

njgheorghita (approved):
🚢

carver (approved):
LGTM!

carver:
Just a reminder that Windows support won't be optional later. There's still a massive user base on Windows, as soon as you leave the developer community.

PR #522, opened 2023-01-30
bump minor ver for ethportal-api new method

perama-v:
### What was wrong?

New method was added, but neglected to bump `0.x.0` with the change

### How was it fixed?

`0.1.3` -> `0.2.0`


ogenev:
We don't need to bump the version on every change, it depends on how often we want to release. Any users still can use the latest changes by using the git link instead of crates.io.

perama-v:
Ok, that's good to know thanks

PR #525, opened 2023-02-02
Updates source node radius on `PING`/`PONG`

mrferris:
### What was wrong?

Trin's processing of `PING`/`PONG` messages is not updating the radius of the source node in the routing table. This results in Trin's gossip functionality sending `OFFER`s to nodes that are not interested in the data based on their radius. These nodes respond with `ACCEPT` messages, but they are empty implying no interest.

### How was it fixed?

Interpret `PING` and `PONG` custom payload field as a 32 byte, little endian representation of an unsigned 256 bit radius value and update the routing table entry.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -1383,6 +1383,13 @@ where
                 self.request_node(&node.enr());
             }
         }
+
+        let data_radius: Distance = pong.custom_payload.into();
+        let updated_node = Node {
+            enr: source,
+            data_radius: data_radius,
+        };
+        self.kbuckets.write().update_node(&key, updated_node, None);
     }
 
     /// Processes a Nodes response.
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -1145,7 +1145,7 @@ where
     }
 
     /// Processes a ping request from some source node.
-    fn process_ping(&mut self, ping: Ping, source: NodeId) {
+    fn process_ping(&self, ping: Ping, source: NodeId) {
         // Look up the node in the routing table.
         let key = kbucket::Key::from(source);
         let optional_node = match self.kbuckets.write().entry(&key) {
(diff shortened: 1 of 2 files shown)


trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1382,9 +1392,32 @@ where
...
     }
 
+    /// Update the recorded radius of a node in our routing table.
+    fn update_node_radius(&self, enr: Enr, data_radius: Distance) {
+        let node_id = enr.node_id();
+        let key = kbucket::Key::from(node_id);
+
+        let updated_node = Node { enr, data_radius };
+
+        match self.kbuckets.write().update_node(&key, updated_node, None) {
```
njgheorghita, line 1410:
```rust
if let UpdateResult::Failed(_) = self.kbuckets.write().update_node(..) {
    error!("..");
}
```

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1264,7 +1269,12 @@ where
     }
 
     // Process ACCEPT response
-    fn process_accept(&self, response: Accept, enr: Enr, offer: Request) -> anyhow::Result<Accept> {
+    fn process_accept(
+        &mut self,
```
njgheorghita, line 1273: It doesn't seem to me like this update to a mutable borrow is necessary - am I missing something?

njgheorghita (approved):
:ship:

trin-core/src/portalnet/overlay_service.rs
mrferris: yea, the `update_node_radius` requires a mutable `self` because it uses `self`'s `kbuckets` property mutably

mrferris: Good point, I think I had needed it previously because of something I was trying but ditched.

But this also made me realize that `process_pong` and `process_ping` are both borrowing `self` mutably when they don't need to be. Weird that the rust compiler doesn't complain about that.

PR #526, opened 2023-02-02
Add purge db script

njgheorghita:
### What was wrong?
After #514 we will have testnet nodes that contain invalid / outdated data. Also following the discussion [here](https://github.com/ethereum/portal-network-specs/pull/177#discussion_r1068003004) it is now expected that we might have to occasionally purge the data from our testnet due to spec updates / etc.

### How was it fixed?
Add a script that can be easily run in our testnet deployment ansible playbook to either flush all data from a node's db - or iterate through and evict only incorrectly encoded data.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/src/bin/purge_db.rs
+++ b/src/bin/purge_db.rs
@@ -1,0 +1,140 @@
+use std::str::FromStr;
+
+use anyhow::anyhow;
+use discv5::enr::{CombinedKey, EnrBuilder};
+use rocksdb::IteratorMode;
+use ssz::Decode;
+use structopt::StructOpt;
+use tracing::{info, warn};
+
+use ethportal_api::types::accumulator::EpochAccumulator;
+use ethportal_api::types::block_body::BlockBody;
+use ethportal_api::types::block_header::BlockHeaderWithProof;
+use ethportal_api::types::content_key::HistoryContentKey;
+use ethportal_api::types::receipts::BlockReceipts;
+use trin_core::portalnet::storage::{PortalStorage, PortalStorageConfig};
+use trin_core::portalnet::types::messages::ProtocolId;
+use trin_core::utils::db::get_data_dir;
+
+///
+/// This script will iterate through all content id / key pairs in rocksd & meta db.
+/// However, if it is run in "invalid-only" mode, it will error if it encounters any
+/// non-history network content. Since we only support history network content, this
+/// shouldn't be a problem, but as we add support for more sub-networks this script will
+/// need to be updated to avoid panicking.
+///
+pub fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let purge_config = PurgeConfig::from_args();
+
+    let mut encoded = hex::decode(&purge_config.private_key).unwrap();
+    let enr_key = CombinedKey::secp256k1_from_bytes(&mut encoded).unwrap();
+    let enr = EnrBuilder::new("v4").build(&enr_key).unwrap();
+    let node_id = enr.node_id();
+    let data_dir_path = get_data_dir(node_id);
+    info!("Purging data for NodeID: {}", node_id);
+    info!("DB Path: {:?}", data_dir_path);
+
+    // Capacity is 0 since it (eg. for data radius calculation) is irrelevant when only removing data.
+    let capacity = 0;
+    let protocol = ProtocolId::History;
+    let config = PortalStorageConfig::new(capacity, node_id, false);
+    let storage = PortalStorage::new(config.clone(), protocol).unwrap();
+    let iter = config.db.iterator(IteratorMode::Start);
+    let mut item_count = 0;
+    let mut remove_count = 0;
+    for (id, value) in iter {
+        item_count += 1;
+        let mut content_id = [0u8; 32];
+        content_id.copy_from_slice(&id);
+        match purge_config.mode {
+            PurgeMode::All => match storage.evict(content_id) {
+                Ok(_) => remove_count += 1,
+                Err(err) => warn!(
+                    "Error occurred while evicting content id: {:?} - {:?}",
+                    content_id, err
+                ),
+            },
+            PurgeMode::Invalid => {
+                // naked unwrap since we shouldn't be storing any invalid content keys
+                let key = match storage.lookup_content_key(content_id).unwrap() {
+                    Some(val) => val,
+                    None => {
+                        warn!(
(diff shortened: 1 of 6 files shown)


jacobkaufmann (commented):
I feel like a more appropriate name for the binary is `purge` and that the two modes would be something like `all` and `invalid-only`. to me, the `validate` mode does not convey that data will be deleted.

njgheorghita:
@jacobkaufmann Good call on rolling back rocksdb evictions if metadb eviction attempts failed.

src/bin/purge_db.rs
```diff
@@ -0,0 +1,111 @@
...
+    let config = PortalStorageConfig::new(capacity, node_id, false);
+    let storage = PortalStorage::new(config.clone(), protocol).unwrap();
+    let iter = config.db.iterator(IteratorMode::Start);
+    let mut item_count = 0;
+    let mut remove_count = 0;
+    for (id, value) in iter {
+        item_count += 1;
+        let mut content_id = [0u8; 32];
+        content_id.copy_from_slice(&id);
+        if purge_config.mode == "purge" {
```
jacobkaufmann, line 38:
```suggestion
        if purge_config.mode == "all" {
```

src/bin/purge_db.rs
```diff
@@ -0,0 +1,111 @@
...
+                ),
+            }
+        } else {
+            // naked unwrap since we shouldn't be storing any invalid content keys
+            let key = storage.lookup_content_key(content_id).unwrap();
+            let content_key = HistoryContentKey::try_from(key).unwrap();
+            if !is_content_valid(&content_key, &value.into_vec()) {
+                match storage.evict(content_id) {
+                    Ok(_) => remove_count += 1,
+                    Err(err) => println!(
```
jacobkaufmann, line 53:
```suggestion
                    Err(err) => eprintln!(
```

src/bin/purge_db.rs
```diff
@@ -0,0 +1,111 @@
...
+    let mut item_count = 0;
+    let mut remove_count = 0;
+    for (id, value) in iter {
+        item_count += 1;
+        let mut content_id = [0u8; 32];
+        content_id.copy_from_slice(&id);
+        if purge_config.mode == "purge" {
+            match storage.evict(content_id) {
+                Ok(_) => remove_count += 1,
+                Err(err) => println!(
```
jacobkaufmann, line 41:
```suggestion
                Err(err) => eprintln!(
```

trin-core/src/portalnet/storage.rs
```diff
@@ -462,6 +457,34 @@ impl PortalStorage {
...
+
+    /// Public method for looking up a content key by its content id
+    pub fn lookup_content_key(&self, id: [u8; 32]) -> anyhow::Result<Vec<u8>> {
+        let conn = self.sql_connection_pool.get()?;
+        let mut query = conn.prepare(CONTENT_KEY_LOOKUP_QUERY)?;
+        let id = id.to_vec();
+        let mut result = query.query_map([id], |row| Ok(ContentKey { key: row.get(0)? }))?;
+        let result = result
+            .next()
+            .ok_or(anyhow!("Unable to find content id in meta db"))?;
```
jacobkaufmann, line 483: instead of an error if not found, it would make more sense to return `Option`.

src/bin/purge_db.rs
```diff
@@ -0,0 +1,111 @@
...
+                return false;
+            }
+        }
+    }
+    true
+}
+
+// CLI Parameter Handling
+#[derive(StructOpt, Debug, PartialEq)]
+#[structopt(name = "Trin DB Util", about = "Manage trin db")]
```
jacobkaufmann, line 96: nit: we could be more specific on the description. one example:

```suggestion
#[structopt(name = "Trin DB Purge Util", about = "Remove undesired or invalid data from Trin DB")]
```

src/bin/purge_db.rs
```diff
@@ -0,0 +1,111 @@
...
+    );
+    Ok(())
+}
+
+fn is_content_valid(content_key: &HistoryContentKey, value: &[u8]) -> bool {
+    match content_key {
+        HistoryContentKey::BlockHeaderWithProof(_) => {
+            if BlockHeaderWithProof::from_ssz_bytes(value).is_err() {
+                return false;
+            }
```
jacobkaufmann, line 73: we could simplify all match arms and remove the `true` at the bottom

```suggestion
            BlockHeaderWithProof::from_ssz_bytes(value).is_ok()
```

jacobkaufmann (commented):
The main thing here is that the check on `mode` was not changed along with the new modes `all` and `invalid-only`. If there is a straightforward way to use an enum for those modes with `structopt`, then that would nice.

Also, instead of `println!` we can use the `tracing` crate to emit error/info logs.

trin-core/src/portalnet/storage.rs
njgheorghita: I'm not so sure, we really never expect to be unable to lookup a content key. Our metadb and rocksdb are expected to remain in sync, so if for some reason we're unable to find a content key for a given content id, that means something's out of sync, which imo warrants an `Error`.

jacobkaufmann: well that assumes that we only look up a content key for a content ID that we have retrieved from the database. if we are given some arbitrary content ID, and we want to look up the content key for some reason, then the caller would not expect an error if the content key was not found.

I submit that here we know that the method will only be called for previously retrieved content IDs, but that may not be true in some possible future use case for the method. I would prefer that we `Option::unwrap` (or emit a warn/error) in the case where the database not holding the content key represents some sort of data corruption.

njgheorghita: Ok, I've made the suggested change, since imo it's a fairly trivial point to go back and forth about. But, for the record, I don't feel like we should be building for "some possible future use case for the method". Especially without a strong example use case. I was under the impression that content ids are primarily for internal use within storage for management. I can't think of a use case where a public facing method on `PortalStorage` will be looking up a content id directly - without knowledge of the content key? Can you?

jacobkaufmann (approved):
it might be worth noting somewhere that this script assumes that only history data is present in the database. if there is non-history data, then it looks like the script (running in 'invalid' mode) will panic while iterating through the keys in RocksDB.

PR #527, opened 2023-02-02
Do not gossip to disconnected nodes

mrferris:
### What was wrong?

Trin currently gossips to nodes without regard for whether the routing table shows them as disconnected or not.

### How was it fixed

Nodes are filtered by connection status before gossiping.

### To-Do

- [x] Filter disconnected before gossiping
- [ ] Clean up commit history


njgheorghita (approved):
:ship:

PR #531, opened 2023-02-08
add hex string method to overlaycontentkey

### LINKED ISSUE #528: No OverlayContentKey method for getting hex string
### Description

`OverlayContentKey`s are the primary method for referring to specific elements in the Portal Network. 

In communications, such as logs, there are two options
- Manually encode every time (overhead per instance) such as `content.key=hex::encode::<Vec<u8>>(content_key.clone().into())`
- Log the bytes (low readability) currently no simple way to log the hex encoded key. 

### Solution
Add a method to get the hex string "0xabcd...1234" such as:
```rs
self.as_hex_str(&self) -> String {
    let hex = hex::encode(self.bytes())
    format!("0x{hex}")
}
```
Implement as a default method for OverlayContentKey that will apply to all future content key types

### Usage

```rs
debug!(content.key=content_key.as_hex_str(), "Key check passed")
```

perama-v:
### What was wrong?

Resolves #528. 

### How was it fixed?

Adds new default method to `OverlayContentKey` trait as follows:
```rust
/// Returns the content key as a hex encoded "0x"-prefixed string.
fn to_hex(&self) -> String {
    hex_encode(self.to_bytes())
}
```
Uses `to_*` naming convention for borrowed->owned methods according to [guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv)

Trin-native `hex_encode`/`hex_decode` are used instead of external `hex::encode`/`hex::decode`. These utils
were copied from trin-core to `ethportal-api`. 

### Testing

Added test cases for `.to_hex()` and, for comparison, the existing `.to_string()` 
```sh
// Existing .to_string()
"EpochAccumulator { epoch_hash: 0xe242..c491 }"
// New .to_hex()
"0x03e242814b90ed3950e13aac7e56ce116540c71b41d1516605aada26c6c07cc491"
```

### Usage
Used in `self.serialize()` to replace/condense existing logic.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/Cargo.toml
+++ b/ethportal-api/Cargo.toml
@@ -25,3 +25,7 @@ serde = { version = "1.0.150", features = ["derive"] }
 serde_json = "1.0.89"
 serde-hex = "0.1.0"
 sha2 = "0.10.1"
+thiserror = "1.0.38"
+
+[dev-dependencies]
+rstest = "0.16.0"
--- a/ethportal-api/src/lib.rs
+++ b/ethportal-api/src/lib.rs
@@ -6,6 +6,7 @@
 mod discv5;
 mod history;
 pub mod types;
+pub mod utils;
 mod web3;
 
 pub use discv5::{Discv5ApiClient, Discv5ApiServer};
--- a/ethportal-api/src/types/content_key.rs
+++ b/ethportal-api/src/types/content_key.rs
@@ -5,6 +5,8 @@ use ssz::{self, Decode, Encode};
 use ssz_derive::{Decode, Encode};
 use std::fmt;
 
+use crate::utils::bytes::{hex_decode, hex_encode, hex_encode_compact};
+
 /// Types whose values represent keys to lookup content items in an overlay network.
 /// Keys are serializable.
 pub trait OverlayContentKey:
@@ -15,6 +17,10 @@ pub trait OverlayContentKey:
     fn content_id(&self) -> [u8; 32];
     /// Returns the bytes of the content key.
     fn to_bytes(&self) -> Vec<u8>;
+    /// Returns the content key as a hex encoded "0x"-prefixed string.
+    fn to_hex(&self) -> String {
+        hex_encode(self.to_bytes())
+    }
 }
 
 /// A content key in the history overlay network.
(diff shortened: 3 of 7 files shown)


ethportal-api/src/types/content_key.rs
```diff
@@ -15,6 +15,11 @@ pub trait OverlayContentKey:
     fn content_id(&self) -> [u8; 32];
     /// Returns the bytes of the content key.
     fn to_bytes(&self) -> [u8; 33];
+    /// Returns the content key as a hex encoded "0x"-prefixed string.
+    fn to_hex_string(&self) -> String {
```
njgheorghita, line 19: I prefer `to_hex()` here. Imo, it's implicit that a fn named `to_hex()` will return a hex-encoded string. However, that might be a hangover from spending a fair amount of time from working with the [eth-utils python library](https://eth-utils.readthedocs.io/en/stable/utilities.html#to-hex-bytes-int-bool-text-str-hexstr-str-hexstr). I'm not sure if people share this convention?

ethportal-api/src/types/content_key.rs
```diff
@@ -15,6 +15,11 @@ pub trait OverlayContentKey:
...
+        let hex = hex::encode(self.to_bytes());
+        format!("0x{hex}")
+    }
```
njgheorghita, line 22: Imo, if we're adding a `to_hex` / `to_hex_string` fn then we should also include the complementary `from_hex` / `from_hex_string` fn

ethportal-api/src/types/content_key.rs
```diff
@@ -222,9 +202,8 @@ mod test {
 
     #[test]
     fn block_header() {
-        let expected_content_key =
-            hex::decode("00d1c390624d3bd4e409a61a858e5dcc5517729a9170d014a6c96530d64dd8621d")
-                .unwrap();
+        const KEY_STR: &str = "00d1c390624d3bd4e409a61a858e5dcc5517729a9170d014a6c96530d64dd8621d";
```
njgheorghita, line 205: Let's add `0x` prefixes to all these values and replace all usage of the `hex` crate for decoding below with the `hex_decode` util. Which, I'm now realizing is not available inside `ethportal-api` ... so my personal preference would be just to copy and paste the `hex_decode` & `hex_encode` utils from `trin_core` into a `ethportal::utils::bytes` file so they're available to use here

ethportal-api/src/types/content_key.rs
perama-v: No problem

ethportal-api/src/types/content_key.rs
perama-v: Completed:
- Copy utils to `ethportal-api`
- Replace use of `hex::*` with native version
- Update test values to comply

ethportal-api/src/types/content_key.rs
perama-v: No problem

ethportal-api/src/types/content_key.rs
njgheorghita: Yup, good point

ethportal-api/src/lib.rs
```diff
@@ -5,6 +5,7 @@
 mod discv5;
 mod history;
 pub mod types;
+mod utils;
```
ogenev, line 8: What is your opinion about exporting this module or parts of it as public?

ethportal-api/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license = "GPL-3.0"
 repository = "https://github.com/ethereum/trin/tree/master/ethportal-api"
 
 [dependencies]
+anyhow = "1.0.68"
```
ogenev, line 10: I think we need to switch to [thiserror](https://docs.rs/thiserror/latest/thiserror/) here.

It is suggested to use `anyhow` for applications and `thiserror` for libraries. This is because a library user can't match anyhow's result error variant.

ethportal-api/src/lib.rs
perama-v: I was on the fence but now favour `pub mod utils`. On the one hand exposing utils increases the scope of the crate's contract. On the other, it reduces friction for the `hex_*` methods, which is desirable

ethportal-api/Cargo.toml
perama-v: Removed `anyhow` and replaced with `thiserror`. The two error variants introduced are both captured in the existing test cases. 
```rs
#[derive(Error, Debug)]
pub enum UtilsError {
    #[error("Could not decode hex")]
    DecodeError(#[from] hex::FromHexError),
    #[error("Hex strings must start with 0x, but found {0}")]
    HexPrefixError(String),
}
```

ethportal-api/src/utils/bytes.rs
```diff
@@ -0,0 +1,66 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum UtilsError {
```
ogenev, line 4: I think it will be more clear if we rename this enum to something like `HexError`, and change the enum variant `HexPrefixError` to `PrefixError`.

Adding a short docstring for the enum will be a nice addition too.

perama-v: Done 👍

ogenev (approved):
Looks good :100: !

njgheorghita:
@perama-v Heyo, could you rebase this then let's get it merged!

PR #543, opened 2023-02-13
Beacon types

njgheorghita:
### What was wrong?
Initial implementation for some datatypes needed to support post-merge headers. Includes...
- Basic proof types copied over from fluffy.
- Beacon chain types that will be needed in the bridge node to construct post-merge proofs
- Reorganization of "wrapped" types

### How was it fixed?
Implemented the types
- This pr copied over all of the tests from the consensus specs, which I left in to show completeness, but I'm open to removing some of the test cases if they're considered unnecessary

### To-Do

- [x] Clean up commit history


ogenev (changes requested):
I think we don't need to implement those types, there are rust libraries that we can use.

Could we check if we can import them from https://github.com/ralexstokes/ethereum-consensus or  [lighthouse](https://github.com/sigp/lighthouse/tree/stable/consensus/types)?

njgheorghita:
I checked out if `lighthouse` was possible, but they don't export their types. And importing the entire lighthouse repo just to access a couple types is overkill imo. I wasn't aware about `ethereum-consensus` - and it seems a plausible candidate.

(IMO)
The benefits of importing types from `ethereum-consensus`...
- Removes some number of lines of code from our repo
- We don't have to maintain them as the consensus spec evolves

The negatives...
- Introduces an "inconsistency" across how our types are used & their api
  - eg. `ethereum-consensus` uses [a different ssz library](https://github.com/ralexstokes/ssz-rs) from the one we use (`eth2_ssz`) - which means to encode/decode will require a different api - which means i'll get quite frustrated when I try to call `.as_ssz_bytes()` on an `ssz_rs`-based type - and then have to go look up the correct api
- Also introduces unnecessary bloat. I'm not convinced that importing a library just to reduce lines of native code in our repo always justifies doing so (especially since this new dependency also brings new dependencies along with it). 

In this case, I think importing these types actually introduces some hidden complexity and dependencies, and is therefore not worth it, but I'm curious what you think

ogenev:
> * eg. `ethereum-consensus` uses [a different ssz library](https://github.com/ralexstokes/ssz-rs) from the one we use (`eth2_ssz`) - which means to encode/decode will require a different api - which means i'll get quite frustrated when I try to call `.as_ssz_bytes()` on an `ssz_rs`-based type - and then have to go look up the correct api

This is a good point to not import `ethereum-consensus` :+1: .

I would add types that are based on the Bellatrix [specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/bellatrix/beacon-chain.md) and create `bellatrix` folder inside `beacon_chain`. This is the current consensus upgrade.

njgheorghita:
@ogenev It was probably this comment that gave the impression the types were based off of the `phase0` specs ...
```
/// Types based off specs @
/// https://github.com/ethereum/consensus-specs/blob/5970ae56a1cd50ea06049d8aad6bed74093d49d3/specs/phase0/beacon-chain.md
```
Which is misleading. The types are based off of the latest version (Bellatrix) of the specs. They are tested against the [latest test vectors](https://github.com/ethereum/consensus-spec-tests/commit/c6e69469a75392b35169bc6234d4d3e6c4e288da) (excluding the recent "hotfix" release of the vectors).

When it comes to organizing these types, I'm not convinced we need release subfolders (eg `beacon/bellatrix/body.rs` / `beacon/phase0/header.rs`). The reason being that only some datatypes are updated in any given consensus spec update. So that would leave us w/ either...

datatypes in the folder of the last update when they were updated (eg...)
```
- beacon/
  - bellatrix/
    - body.rs
  - phase0/
    -header.rs
```
This feels like unnecessary maintenance imo.

Or we just define every type as "updated" in the lastest spec release - regardless of whether or not the datatype actually changed (eg...)
```
- beacon/
  - bellatrix/
    - body.rs
    - header.rs
```
Which is redundant imo

So, yes, we want to make sure that our types are up to date w/ the latest release of the consensus specs. However, I think the best way to do this is to update our test vectors with every consensus spec update. Then we can simply assume that every type in `beacon/*` is up to date with the latest spec update. I don't foresee any scenario when we will need to maintain multiple versions of the same type from different consensus spec updates, do you?

ogenev:
I don't know if we want to implement beacon types before having the required logic that uses them. This is because that way we don't know how to architecturally structure the modules and where to put them.

For example, if only a bridge node requires them, we can have a `bridge-node` crate with a `types` file for any types required by this module. This way we encapsulate the bridge logic and it is clear for contributors where to look.

I want to challenge the decision to implement those types in `ethportal-api`. The vision for `ethportal-api` is to be a standalone library for interacting with the portal network APIs. Does this require beacon types, if they are used only for a bridge service that is part of Trin?

njgheorghita:
@ogenev Makes sense. How would you feel about storing these types inside of the newly created `trin-types`? It seems like a plausible place for these to live until we're ready to use them.

ogenev:
> @ogenev Makes sense. How would you feel about storing these types inside of the newly created `trin-types`? It seems like a plausible place for these to live until we're ready to use them.

Yes, seems a reasonable place. I would put those types inside a `consensus` or `beacon-chain` folder for example.

njgheorghita:
Moved the beacon types into `trin-types/consensus`, so this pr is ready for a :eyes: @ogenev @jacobkaufmann

carver (approved):
Yup, seems fine to me. Were you able to confirm that the tests are live in practice? (as in, if you push an invalid implementation, Circle complains because of the new tests in this PR)

njgheorghita:
> Yup, seems fine to me. Were you able to confirm that the tests are live in practice? (as in, if you push an invalid implementation, Circle complains because of the new tests in this PR)

confirmed!

PR #544, opened 2023-02-14
jsonrpc: Rename `Offer` endpoint to `Gossip` and add new `Offer` call

### LINKED ISSUE #499: Update `portal_*SendOffer` and `portal_historyOffer` JSON-RPC endpoints
Update `portal_*SendOffer` and `portal_historyOffer` to match the newest specs updates in https://github.com/ethereum/portal-network-specs/pull/179.

ogenev:
### What was wrong?
We need to change Trin's offer endpoints to match recent [spec changes](https://github.com/ethereum/portal-network-specs/pull/179).

Closes #499 

### How was it fixed?
Rename the `Offer` endpoint to `Gossip` and add a new `Offer` call.

- `portal_*Offer` accepts now a single content key as a parameter, instead of a list of content keys.
- Do not return `connection_id` in the offer json-rpc result.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1273,7 +1273,7 @@ checksum = "0198b9d0078e0f30dedc7acbb21c974e838fc8fae3ee170128658a98cb2c1c04"
 
 [[package]]
 name = "ethportal-api"
-version = "0.1.4"
+version = "0.1.5"
 dependencies = [
  "bytes 1.3.0",
  "enr 0.7.0",
--- a/ethportal-api/Cargo.toml
+++ b/ethportal-api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethportal-api"
-version = "0.1.4"
+version = "0.1.5"
 edition = "2021"
 description = "Definitions for various Ethereum Portal Network JSONRPC APIs"
 license = "GPL-3.0"
--- a/ethportal-api/src/history.rs
+++ b/ethportal-api/src/history.rs
@@ -82,21 +82,17 @@ pub trait HistoryNetworkApi {
 
     /// Send the provided content item to interested peers. Clients may choose to send to some or all peers.
     /// Return the number of peers that the content was gossiped to.
-    #[method(name = "historyOffer")]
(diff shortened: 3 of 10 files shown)


njgheorghita (approved):
As we test the testnet, it becomes useful to have a jsonrpc endpoint where I can initiate an offer to a specific enr. The `portal_historyOffer` endpoint seems the best suited for this, but the ux is still a bit clunky. 
1. Lookup local content keys.
2. Send `portal_historyOffer`.

The problem is looking up the local content keys. Figuring out which ones the node should be interested in. And not having much insight into the result of the entire offer-accept transaction.

It would be useful to have an endpoint that consolidates these steps. Imo, something like the following would be ideal.

1. `portal_historyOfferAllInterested(enr)` (this name could use improvement)
2. When you send this request. Your local node looks up all content keys that the enr might be interested in (based on advertised radius) - and sends an `Offer`.
4. This request returns a list of all `Accept`ed content keys and which were successfully transmitted

Another point I've been discussing with @mrferris is that for our `Offer` jsonrpc requests, it would be quite useful to know (via the result) which offer/accept streams concluded with a successful transfer and which failed

So, obvs these things are out of scope for this pr, but just wanted to list them here to get thoughts, and if any ideas gain traction, I'll open an issue

ogenev:
> It would be useful to have an endpoint that consolidates these steps. Imo, something like the following would be ideal.
> 
> 1. `portal_historyOfferAllInterested(enr)` (this name could use improvement)
> 2. When you send this request. Your local node looks up all content keys that the enr might be interested in (based on advertised radius) - and sends an `Offer`.
> 3. This request returns a list of all `Accept`ed content keys and which were successfully transmitted
> 
> Another point I've been discussing with @mrferris is that for our `Offer` jsonrpc requests, it would be quite useful to know (via the result) which offer/accept streams concluded with a successful transfer and which failed

I think we can have as many debug/admin endpoints as we want. This doesn't mean we should include them in the specs. IMO, it makes sense to add anything that we find useful and help debug.

PR #546, opened 2023-02-15
Remove obsolete `JsonRpcExiter`

### LINKED ISSUE #542: Remove `JsonRpcExiter` struct
We don't need [JsonRpcExiter](https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/trin-core/src/jsonrpc/service.rs#L12) anymore, because jsonrpsee server handles this for us.

Instead of returning it from run_trin, we can return the jsonrspee server handler which for now runs forever in the background - [here](https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/src/lib.rs#L165) and [here](https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/src/lib.rs#L177).

ogenev:
### What was wrong?

Closes #542 

### How was it fixed?
Remove the obsolete `JsonRpcExiter` struct and instead return jsonrpsee server handle from the `run_trin` method.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/lib.rs
+++ b/ethportal-peertest/src/lib.rs
@@ -6,15 +6,15 @@ pub use cli::PeertestConfig;
 pub use jsonrpc::get_enode;
 
 use std::net::{IpAddr, Ipv4Addr};
-use std::{sync::Arc, thread, time};
+use std::{thread, time};
 
+use ethportal_api::jsonrpsee::server::ServerHandle;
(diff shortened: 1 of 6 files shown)


njgheorghita (approved):
:ship:

PR #551, opened 2023-02-17
migrate to separate uTP crate

jacobkaufmann:
### What was wrong?

`trin` maintains an internal uTP implementation based on the [`utp`](https://crates.io/crates/utp) crate. This approach was taken to support uTP in a timely manner, but now we have the opportunity to replace it with a more general-purpose library that is not so interwoven into other system components.

### How was it fixed?

Changes:

* remove the `trin_core::utp` module
* add a dependency on `utp-rs`
* use `UtpSocket` in place of `UtpListener`
* implement an abstract UDP socket over Discovery v5
* update `utp-testing` to use the transport from `utp-rs`

### To-Do

- [x] Clean up commit history


jacobkaufmann:
the version of `utp-rs` here is `0.1.0-alpha` for now. I would like to collect some feedback before moving `utp-rs` beyond a pre-release version. once I have confidence that the API is stable, I will make a `0.1.0` release and update the PR to depend on that version.

.circleci/config.yml
```diff
@@ -122,7 +122,7 @@ jobs:
       - run:
           name: Run Test App with network simulator
           # Dropping packets from both sides of the stream to test dealing with lost packets
-          command: SCENARIO="droplist --delay=15ms --bandwidth=10Mbps --queue=25 --drops_to_server=2,5,7 --drops_to_client=6,8" docker-compose -f utp-testing/docker/docker-compose.yml up -d
+          command: SCENARIO="droplist --delay=15ms --bandwidth=10Mbps --queue=25 --drops_to_server=3,5,7 --drops_to_client=6,8" docker-compose -f utp-testing/docker/docker-compose.yml up -d
```
jacobkaufmann, line 125: note: this change was made because the 2nd packet corresponds to the node's response to the discovery v5 `WHOAREYOU` challenge

trin-core/src/portalnet/discovery.rs
```diff
@@ -285,3 +304,86 @@ impl Discovery {
         Ok(response)
     }
 }
+
+pub struct Discv5UdpSocket {
+    // `RwLock` for interior mutability.
```
ogenev, line 309: Is this `tokio::sync::Mutex` instead of `RwLock`?

ogenev (commented):
Looks solid to me, I think it is a great improvement over what we had. :+1: 

Did we test OFFER/ACCEPT and FINDCONTENT?CONTENT compatibility with Fluffy?

Sadly, we don't have automatic tests for this and before approval, I think we should test this manually.

trin-core/src/portalnet/discovery.rs
jacobkaufmann: good catch! it was a `tokio::sync::RwLock` initially. I will update the comment.

.circleci/config.yml
jacobkaufmann: at this time I'm not sure, but I can look into it further.

jacobkaufmann:
I did some manual testing by running a bridge peered with a (uTP-modified) `trin` node, but I should do some more isolated testing with a local `fluffy` node. Once confirmed successful, I will leave a note here.

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1056,13 +1064,68 @@ where
...
+            let mut stream = match utp.accept_with_cid(cid.clone()).await {
+                Ok(stream) => stream,
+                Err(err) => {
+                    warn!(%err, cid.send, cid.recv, peer = %cid.peer.node_id(), "unable to accept uTP stream");
+                    return;
+                }
+            };
+
+            let mut data = vec![];
+            match stream.read_to_eof(&mut data).await {
```
njgheorghita, line 1106: `if let Err(err) = stream.read_to_eof(&mut data).await {..}`?

trin-core/src/portalnet/discovery.rs
```diff
@@ -285,3 +304,86 @@ impl Discovery {
...
+    pub fn new(discv5: Arc<Discovery>, talk_reqs: mpsc::UnboundedReceiver<TalkRequest>) -> Self {
+        Self {
+            discv5,
+            talk_reqs: tokio::sync::Mutex::new(talk_reqs),
+        }
+    }
+}
+
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct UtpEnr(pub Enr);
```
njgheorghita, line 325: Can you leave a comment on why this additional enr wrapped type is necessary

.circleci/config.yml
carver: Any luck tracking this down?

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -102,8 +99,8 @@ async fn wait_for_content(
 
     let mut counter = 0;
     while received_content_value == HistoryContentItem::Unknown("".to_owned()) && counter < 5 {
-        error!("Retrying after 0.5sec, because content should have been present");
-        thread::sleep(time::Duration::from_millis(500));
+        error!("Retrying after 1s, because content should have been present");
+        thread::sleep(time::Duration::from_millis(1000));
```
carver, line 103: A code comment explaining the intuition of why not to drop it to 0.5s would be nice.

carver (approved):
Hooray for a bunch of deleted code! 👍🏻 for the direction

> I should do some more isolated testing with a local fluffy node. Once confirmed successful, I will leave a note here.

I agree this is worth doing.

ethportal-peertest/src/scenarios/offer_accept.rs
jacobkaufmann: I was able to move the retry duration back to 0.5 seconds with no failure.

.circleci/config.yml
jacobkaufmann: unfortunately I have not been able to track this down. I'm not sure why things were working before. that said, I do not think this should block merge because we move the drop into the uTP data transfer, which is the aspect we are interested in testing.

jacobkaufmann:
I ran into some discv5-level issues testing a local `trin` and `fluffy` pair, but I have been able to peer a local `trin` node with the public portal network and successfully transmit data via uTP. I believe that we are in a place to move forward and diagnose/address issues as they crop up. there are further enhancements to make to automated uTP testing, but that is outside the present scope.

PR #555, opened 2023-02-20
Gossip to combination of closest nodes and random nodes, rather than just random nodes

mrferris:
### What was wrong?

Per piece of content to gossip, Trin currently gossips to 5-7 random nodes from the entire routing table. This results in data mostly being gossiped to nodes where the data won't reasonably be looked for during queries.

### How was it fixed?

Before selecting gossip recipients for a given piece of content, the routing table is sorted by proximity to that content. The content is then gossiped to (at most) the 4 closest nodes, and 4 other random nodes.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -385,7 +385,7 @@ where
                 continue;
             }
 
-            // Sort all eligible nodes by proximity to the content and take the closest k.
+            // Sort all eligible nodes by proximity to the content.
             interested_enrs.sort_by(|a, b| {
                 let distance_a = XorMetric::distance(&content_key.content_id(), &a.node_id().raw());
                 let distance_b = XorMetric::distance(&content_key.content_id(), &b.node_id().raw());
@@ -394,21 +394,13 @@ where
                     std::cmp::Ordering::Less
                 })
             });
-            let closest_k_enrs = interested_enrs[0..MAX_NODES_PER_BUCKET].to_vec();
 
-            // Get log2 random ENRs to gossip
-            let random_enrs = match log2_random_enrs(closest_k_enrs) {
-                Ok(val) => val,
-                Err(msg) => {
-                    warn!(error = %msg, "Error producing ENRs for gossip");
-                    return 0;
-                }
-            };
+            let gossip_recipients = select_gossip_recipients(interested_enrs);
 
             // Temporarily store all randomly selected nodes with the content of interest.
             // We want this so we can offer all the content to interested node in one request.
             let raw_item = (content_key.into(), content_value);
-            for enr in random_enrs {
+            for enr in gossip_recipients {
                 enrs_and_content
                     .entry(enr.to_base64())
                     .or_default()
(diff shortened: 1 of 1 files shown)


perama-v (approved):

njgheorghita (approved):
🚢

trin-core/src/portalnet/overlay.rs
```diff
@@ -376,6 +376,14 @@ where
                 .map(|node| node.value.enr())
                 .collect();
 
+            // Sort all eligible nodes by proximity to the content and take the closest k.
+            interested_enrs.sort_by(|a, b| {
+                let distance_a = XorMetric::distance(&content_key.content_id(), &a.node_id().raw());
+                let distance_b = XorMetric::distance(&content_key.content_id(), &b.node_id().raw());
+                distance_a.partial_cmp(&distance_b).unwrap()
```
ogenev, line 383: Is this a safe `unwrap`?

trin-core/src/portalnet/overlay.rs
```diff
@@ -376,6 +376,14 @@ where
...
+            });
+            let interested_enrs = interested_enrs[0..MAX_NODES_PER_BUCKET].to_vec();
```
ogenev, line 385: What is the reasoning for limiting the interested nodes to 16?

trin-core/src/portalnet/overlay.rs
mrferris: @ogenev Good question, I was wondering the same. I think the answer is yes but I updated it to provide a default value and a warning in case it's not.

trin-core/src/portalnet/overlay.rs
mrferris: @ogenev I updated this variable name to be `closest_k_enrs` so hopefully it's clearer now. 

The reasoning for this limit is that because once we've sorted the list, the first 16 will be the closest to the data. Without this limit, the majority of our gossip will be to nodes whose IDs are far away from the content ID. 

This (including the selection of k=16 nodes) is done according to [portal wire spec](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-protocol.md#neighborhood-gossip). 

Let me know if that does/doesn't answer your question.

ogenev: > This (including the selection of k=16 nodes) is done according to [portal wire spec](https://github.com/ethereum/portal-network-specs/blob/master/portal-wire-protocol.md#neighborhood-gossip).

I don't understand why k=16 in this case. There are 16 keys per bucket but why do we want to gossip to only 16 nodes instead of all interested nodes that are close to the content?

mrferris: @ogenev It's an arbitrary number. But the larger the number, the more we are gossiping to far-away nodes that will either never be asked or only be asked after asking the closer nodes are asked (and thus both wasting storage space and slowing down queries).

Do you have a different number in mind?

ogenev: I see, here is how Fluffy implements this:

https://github.com/status-im/nimbus-eth1/blob/a74bbef74c49676c8ac78ef4c4a1c8d94265894e/fluffy/network/wire/portal_protocol.nim#L1153

I think their algorithm is an improvement over ours but I guess we can iron it out more in another PR.

mrferris: @ogenev Yea, in a future PR we'll implement the case where there's too few interested nodes so we initiate a `FINDNODES` query.

Currently the max number of nodes we're gossiping to is 4, I'm good with doubling it to 8. You're right, 4 might be too low.

carver:
(Not a reason to block this PR, just something to watch for going forward)

- Refusing to consider the nodes that are far away has the effect of "not believing" nodes that claim they want to store a wide radius. If someone wants to run a node with complete data, this makes it harder for them. (This is probably fine. Someone who wants to support the network a lot would be better off hosting several different nodes with IDs spread around the network anyway, instead of one huge one. Maybe something worth pointing out in docs somewhere)
- One reason we don't just pick the n closest is to make it harder for a malicious actor to "claim" a certain key, and then act as a black hole. By spreading out who we gossip to more (and unpredictably), they have to do more sybil work, spreading nodes across more IDs. But if we end up tuning k too small, it could make this attack very cheap, again. I don't know what "too small" is here, yet.

perama-v:
This is interesting
> ... Someone who wants to support the network a lot would be better off hosting several different nodes with IDs spread around the network anyway, instead of one huge one. ...

I think there could be a moderate user base who thinks of a portal node as a regular node with a "slider". So if they have 50GB of disk they want to contribute, they can. If they have 100GB soon after, they can effectively "drag the slider". 

Spinning up additional nodes likely seems like more work for the user, and perhaps doesn't give them the benefit of faster lookups (which of their nodes should they ping?). So rather than recommending a new node, increase the capacity of the node (more threads, disk) and have the network believe that there really are nodes with large radii

trin-core/src/portalnet/overlay.rs
ogenev: 8 looks also better to me

> Currently the max number of nodes we're gossiping to is 4, I'm good with doubling it to 8. You're right, 4 might be too low.

Could you factor it into this PR?

mrferris:
@carver Those are good points.

> Refusing to consider the nodes that are far away has the effect of "not believing" nodes that claim they want to store a wide radius.

True, though I'd find wide radius nodes easier to believe if we were to start low capacity Trin instances with a smaller-than-infinite radius, like Fluffy does. Many of these faraway infinite radius nodes to which we're gossiping will delete this gossiped data and narrow their radius once their storage capacity is reached. 

> By spreading out who we gossip to more (and unpredictably), they have to do more sybil work

I've modified this PR to take this into account. 

Now we gossip to up to 8 nodes: the 4 closest nodes plus 4 other random nodes from the entire set of interested nodes.

trin-core/src/portalnet/overlay.rs
```diff
@@ -801,22 +804,36 @@ where
...
-    }
-    // log2(1) is zero but we want to propagate to at least one node
-    if enrs.len() == 1 {
-        return Ok(enrs);
-    }
-
-    // Greedy gossip by ceiling the log2 value
-    let log2_value = (enrs.len() as f32).log2().ceil();
-
+fn random_enrs(num_enrs: usize, enrs: Vec<Enr>) -> Vec<Enr> {
```
ogenev, line 807: Maybe rename this to something like: 
```suggestion
fn select_random_enrs(num_enrs: usize, enrs: Vec<Enr>) -> Vec<Enr> {
```

ogenev (approved):
Looks good :+1: .

mrferris:
> Spinning up additional nodes likely seems like more work for the user, and perhaps doesn't give them the benefit of faster lookups (which of their nodes should they ping?).

@perama-v Hypothetical, but we could write a wrapper that runs multiple nodes for the user, and splits the provided storage space between them. We would just start the query from the node that's closest to the content. The user wouldn't have to know that's happening.

perama-v:
That sounds pretty good. One thought is that each would have a discrete discovery port. If a user needs to modify ufw or similar this could be a challenge

perama-v:
Or maybe this is possible over a single discovery port. If for example a single trin node maintains a concept of multiple virtual nodes. The discovery port clash detector ttrigger could be avoided.

carver:
> That sounds pretty good. One thought is that each would have a discrete discovery port. If a user needs to modify ufw or similar this could be a challenge

Ideally, our #596 updates should prevent users from manually having to fiddle with firewalls. Also, I think that now that we're gossiping to some fully random nodes, the network can still support wide-radius nodes pretty cleanly.

PR #556, opened 2023-02-20
Add check for valid ENR signature

mrferris:
### What was wrong?

I don't think the signatures of the ENRs that Trin (or discv5) receives are being checked anywhere. Correct me if I'm wrong. 

Discv5 uses a `verify_enr` method, but it doesn't check the signature, only that the ENR's stated address matches the observed address.

### How was it fixed?

Using the `Enr` crate's [verify](https://docs.rs/enr/latest/enr/struct.Enr.html#method.verify) method.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -1543,6 +1543,11 @@ where
         for enr in enrs {
             let node_id = enr.node_id();
 
+            if !enr.verify() {
+                warn!("Received ENR with invalid signature: {}", enr);
+                continue;
+            }
+
             // Ignore ourself.
             if node_id == local_node_id {
                 continue;


jacobkaufmann:
the ENR signature is [verified when decoded from RLP](https://github.com/sigp/enr/blob/79d2df3cf2727386fc42f9233181af7ea17b2386/src/lib.rs#L1020), so I don't think there is a need to do separate verification.

mrferris:
> the ENR signature is [verified when decoded from RLP](https://github.com/sigp/enr/blob/79d2df3cf2727386fc42f9233181af7ea17b2386/src/lib.rs#L1020), so I don't think there is a need to do separate verification.

Ah there we go, thanks

pipermerriam:
Maybe we want to add a test for this that simply gives us verification that if trin receives ENR with invalid signature it gets rejected...  not 100% sure this is valuable, so consider this an uninformed suggestion.

jacobkaufmann:
I don't think a check is necessary so long as we convert whatever ENR we receive to an [`Enr`](https://github.com/sigp/enr/blob/79d2df3cf2727386fc42f9233181af7ea17b2386/src/lib.rs#L225) before doing anything with it. That would be necessary to send a talk request to that ENR or to add that ENR to the routing table. AFAIK there are no places where we exclusively (with no conversion(s)) work with "raw" base64 or RLP-encoded ENRs.

From the doc comment linked above:

> This struct will always have a valid signature, known public key type, sequence number and `NodeId`.

PR #560, opened 2023-02-21
Output purge logs to stdout

njgheorghita:
### What was wrong?
The strategy we use in Ansible for capturing log output from the purge step only works for `stdout`.

### How was it fixed?
Use `println!` / `eprintln!` for log output rather than tracing library

### To-Do

- [x] Clean up commit history

### DIFF
--- a/src/bin/purge_db.rs
+++ b/src/bin/purge_db.rs
@@ -5,7 +5,7 @@ use discv5::enr::{CombinedKey, EnrBuilder};
 use rocksdb::IteratorMode;
 use ssz::Decode;
 use structopt::StructOpt;
-use tracing::{debug, info, warn};
+use tracing::{info, warn};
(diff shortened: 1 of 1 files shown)


jacobkaufmann:
for this, we just need to create a [`Subscriber`](https://docs.rs/tracing/0.1.37/tracing/trait.Subscriber.html) like we do for [`trin`](https://github.com/ethereum/trin/blob/e0304a3c48988d3b85833ee0890a0a5fb79f826f/src/main.rs#L8) right?

pipermerriam:
I think this might be what @jacobkaufmann already said, but can't we fix this by just ensuring that the tracing output is configured to go to stdout?

njgheorghita:
@pipermerriam Yup, I just included that change

PR #562, opened 2023-02-22
Add condition for clippy to warn on naked unwraps

njgheorghita:
### What was wrong?
Naked unwraps are bad, and have a tendency slip through the cracks. Case in point, I missed one in #555. Even though @ogenev spotted it, I think it's a good idea to add a check for unwraps to clippy so we don't have to rely on humans.

Furthermore, while working with the testnet nodes, I've experienced multiple instances where nodes are crashing / behaving incorrectly due to these unwraps. 
- eg. I spent a solid day once trying to figure out why a node was running but wouldn't respond to `portal_historyPaginate*` endpoints. Turns out the history thread was crashing due to a naked unwrap.

Our style guide also [dictates](https://github.com/ethereum/trin/blob/master/docs/contributing.md#error-handling) that naked unwraps are not allowed and if there is an exception (eg parsing a static value), the justification must be labeled.

In this pr, I added a clippy check for `unwrap_used`. I went through and fixed all of the simple cases. I plan to go through (before merging this pr) and revisit the more complex cases once I get :+1: on adopting this lint from the others

### How was it fixed?
- Added `#[warn(clippy::unwrap_used)]` to each workspace, except...
  - `trin-cli` - this library is no longer "maintained"
  - `utp-testing` - since it's a "testing" library
- Introduced a `clippy.toml` file to allow for `allow-unwrap-in-tests` configuration 
- Currently, the `allow-unwrap-in-tests` lint is [only available in rust nightly, not stable](https://github.com/rust-lang/rust-clippy/issues/9062). Which requires an explicit `#[allow(clippy::unwrap_used)]` for each test block that contains unwraps - since our ci environment uses stable. Imo, this is a bit wordy, but a fair trade-off since...
  - we'll be able to remove these explicit `allows` once the feature gets in stable
  - according to the [clippy docs](https://doc.rust-lang.org/clippy/usage.html?highlight=meant#lint-configuration)
    - `Clippy is meant to be used with a generous sprinkling of #[allow(..)]s through your code.`
  - in the meantime we get the benefit of automatically checking for naked unwraps

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/overlay_service.rs
+++ b/trin-core/src/portalnet/overlay_service.rs
@@ -312,7 +312,7 @@ where
         query_parallelism: usize,
         query_num_results: usize,
         findnodes_query_distances_per_peer: usize,
-    ) -> Result<UnboundedSender<OverlayCommand<TContentKey>>, String>
+    ) -> UnboundedSender<OverlayCommand<TContentKey>>
     where
         <TContentKey as TryFrom<Vec<u8>>>::Error: Send,
     {
@@ -363,7 +363,7 @@ where
             service.start().await;
         });
 
-        Ok(command_tx)
+        command_tx
     }
 
     fn add_bootnodes(&mut self, bootnode_enrs: Vec<Enr>) {
@@ -1302,10 +1302,19 @@ where
         let response_clone = response.clone();
 
         tokio::spawn(async move {
-            let mut conn = rx.await.unwrap();
+            let mut conn = match rx.await {
+                Ok(conn) => conn,
+                Err(err) => {
+                    error!(error = %err, "Error receiving UTP stream from UtpListener");
+                    return;
+                }
+            };
             // Handle STATE packet for SYN
             let mut buf = [0; BUF_SIZE];
-            conn.recv(&mut buf).await.unwrap();
+            if let Err(err) = conn.recv(&mut buf).await {
+                error!(error = %err, "Error receiving STATE packet from UtpListener");
+                return;
+            }
 
             let content_items = match offer {
                 Request::Offer(offer) => {
@@ -2099,6 +2108,7 @@ fn pop_while_ssz_bytes_len_gt(enrs: &mut Vec<SszEnr>, max_size: usize) {
 }
 
 #[cfg(test)]
+#[allow(clippy::unwrap_used)]
 mod tests {
     use super::*;
 
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -206,17 +206,21 @@ pub struct PortalStorageConfig {
 }
 
 impl PortalStorageConfig {
-    pub fn new(storage_capacity_kb: u64, node_id: NodeId, metrics_enabled: bool) -> Self {
-        let db = Arc::new(PortalStorage::setup_rocksdb(node_id).unwrap());
-        let sql_connection_pool = PortalStorage::setup_sql(node_id).unwrap();
-        Self {
+    pub fn new(
+        storage_capacity_kb: u64,
+        node_id: NodeId,
+        metrics_enabled: bool,
+    ) -> anyhow::Result<Self> {
+        let db = Arc::new(PortalStorage::setup_rocksdb(node_id)?);
+        let sql_connection_pool = PortalStorage::setup_sql(node_id)?;
+        Ok(Self {
             storage_capacity_kb,
             node_id,
             distance_fn: DistanceFunction::Xor,
             db,
             sql_connection_pool,
             metrics_enabled,
-        }
+        })
     }
 }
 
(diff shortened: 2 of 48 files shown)


carver:
> In this pr, I added a clippy check for `unwrap_used`. I went through and fixed all of the simple cases. I plan to go through (before merging this pr) and revisit the more complex cases once I get +1 on adopting this lint from the others

So right now you're just looking on +1 to adopt the lint, then you'll come back and ask for another review later?

If so, then 👍🏻 for adding this lint.

mrferris:
👍 from me as well

ogenev:
:+1:

njgheorghita:
@carver @mrferris @ogenev ready for an :eyes: on this one

trin-core/src/portalnet/overlay_service.rs
```diff
@@ -1302,10 +1302,19 @@ where
...
+                Err(err) => {
+                    error!(error = %err, "Error receiving UTP stream from UtpListener");
+                    return;
+                }
+            };
             // Handle STATE packet for SYN
             let mut buf = [0; BUF_SIZE];
-            conn.recv(&mut buf).await.unwrap();
+            if let Err(err) = conn.recv(&mut buf).await {
+                error!(error = %err, "Error receiving UTP stream from UtpListener");
```
carver, line 1315: It's almost always helpful to have the string be unique. To make this different from 1308:
```suggestion
                error!(error = %err, "Error receiving STATE packet from UtpListener");
```

trin-core/src/portalnet/storage.rs
```diff
@@ -427,7 +431,8 @@ impl PortalStorage {
 
         // Delete furthest data until our data usage is less than capacity.
         while self.capacity_reached()? {
-            // Unwrap because this was set in the block before this loop.
+            // Naked unwrap ok because this was set in the block before this loop.
+            #[allow(clippy::unwrap_used)]
             let id_to_remove = self.farthest_content_id.clone().unwrap();
```
carver, line 436: There are so many places this could go wrong in future code changes. Maybe it's fine to panic here, but I think an expect is clearer, and means we can skip the allow and comment. Something like:

```suggestion
            let id_to_remove = self.farthest_content_id.clone()
                .expect("farthest_content_id is always set immediately before this block");
```

trin-core/src/portalnet/storage.rs
```diff
@@ -689,7 +696,9 @@ impl PortalStorage {
 
     /// Helper function for opening a RocksDB connection for the radius-constrained db.
     pub fn setup_rocksdb(node_id: NodeId) -> Result<rocksdb::DB, ContentStoreError> {
-        let mut data_path: PathBuf = get_data_dir(node_id);
+        let mut data_path: PathBuf = get_data_dir(node_id).map_err(|err| {
+            ContentStoreError::Database(format!("unable to get data dir: {:?}", err))
```
carver, line 700:
```suggestion
            ContentStoreError::Database(format!("unable to get data dir for rocksdb: {:?}", err))
```

for error uniqueness

trin-core/src/portalnet/storage.rs
```diff
@@ -700,7 +709,9 @@ impl PortalStorage {
 
     /// Helper function for opening a SQLite connection.
     pub fn setup_sql(node_id: NodeId) -> Result<Pool<SqliteConnectionManager>, ContentStoreError> {
-        let mut data_path: PathBuf = get_data_dir(node_id);
+        let mut data_path: PathBuf = get_data_dir(node_id).map_err(|err| {
+            ContentStoreError::Database(format!("unable to get data dir: {:?}", err))
```
carver, line 713:
```suggestion
            ContentStoreError::Database(format!("unable to get data dir for sql: {:?}", err))
```

for error uniqueness

trin-core/src/utp/packets.rs
```diff
@@ -142,7 +142,7 @@ impl PacketHeader {
 
     /// Returns the packet's type.
     pub fn get_type(&self) -> PacketType {
-        PacketType::try_from(self.type_ver >> 4).unwrap()
+        PacketType::try_from(self.type_ver >> 4).expect("Invalid packet type")
```
carver, line 145: This string is used already. For uniqueness:

```suggestion
        PacketType::try_from(self.type_ver >> 4).expect("Can't parse packet type")
```

trin-state/src/utils.rs
```diff
@@ -50,6 +50,8 @@ impl ToU256 for BigInt {
 // distance function as defined at...
 // https://notes.ethereum.org/h58LZcqqRRuarxx4etOnGQ#Storage-Layout
 // todo: measure performance & optimize if necessary
+// naked wraps ok here because we're operating on static values
+#[allow(clippy::unwrap_used)]
```
carver, line 54: Slight preference to move this down to the `unwrap()` usages, rather than allowing the whole method.

Also, between the overhead of adding allow, and the comment, I am thinking it might be better to generally prefer `expect`s, even when doing static parsing. Like:

```rs
   let modulo = BigInt::parse_bytes(&MODULO, 10).expect("Parse static MODULO always works");
   let mid = BigInt::parse_bytes(&MID, 10).expect("Parse static MID always works");
```

This kind of thing could even be added to the style guide as a recommendation.

carver (approved):
Love it! 🚢 

It's great how often unwraps could be replaced by `?`, or removed altogether.

PR #563, opened 2023-02-22
Add optional `content_value` param in `portal_*Offer` rpc

### LINKED ISSUE #559: Allow `content_value` param in `Offer` json-rpc endpoint
According to JSON-RPC [specs](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) `portal_*Offer` accept `contentKey` and `contentValue` params. In our implementation we allow only `contentKey` param and we check for a content value in the database.

From discord:

deme: _The benefit of passing also the content_value is that you can use this portal\_*Offer call for any kind of node to inject data or to test with. If you don't do this, you need to do a portal\_*Store first, which might result in the node not actually storing the data due to its NodeId & radius combination, resulting in a failed Offer. So you practically need a Full node (as in, full radius) running to inject data into the network, or to test these offers in Hive tests and such._

This may require some architectural overlay changes in Trin to allow sending a "direct" offer message and initializing uTP transfer without storing the content value in the database.

ogenev:
### What was wrong?
Portal json-rpc [specs](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) require that a `content_value` param is accepted in `portal_*Offer` endpoint. 

Closes #559

### How was it fixed?

Add an optional `content_value` param to `Offer` rpc. If it is not provided, we look for the content in the database, otherwise, we keep the content value in memory and send it using the `PopulateOffer` overlay request.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/scenarios/offer_accept.rs
+++ b/ethportal-peertest/src/scenarios/offer_accept.rs
@@ -1,70 +1,116 @@
+use std::str::FromStr;
 use std::{thread, time};
 
-use serde_json::Value;
+use serde_json::json;
 use tracing::{error, info};
 
+use ethportal_api::jsonrpsee::async_client::Client;
+use ethportal_api::types::discv5::Enr;
+use ethportal_api::{HistoryContentItem, HistoryContentKey, HistoryNetworkApiClient};
+
 use crate::{
-    jsonrpc::{
-        make_ipc_request, validate_portal_offer, JsonRpcRequest, HISTORY_CONTENT_KEY,
-        HISTORY_CONTENT_VALUE,
-    },
+    jsonrpc::{validate_portal_offer, HISTORY_CONTENT_KEY, HISTORY_CONTENT_VALUE},
     Peertest, PeertestConfig,
 };
-use trin_core::jsonrpc::types::Params;
 
-pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
-    info!("Testing OFFER/ACCEPT flow");
+pub async fn test_unpopulated_offer(peertest_config: PeertestConfig, peertest: &Peertest) {
+    info!("Testing Unpopulated OFFER/ACCEPT flow");
+
+    let ipc_client = reth_ipc::client::IpcClientBuilder::default()
+        .build(&peertest_config.target_ipc_path)
+        .await
+        .unwrap();
+
+    let content_key: HistoryContentKey =
+        serde_json::from_value(json!(HISTORY_CONTENT_KEY)).unwrap();
+    let content_value: HistoryContentItem =
+        serde_json::from_value(json!(HISTORY_CONTENT_VALUE)).unwrap();
 
     // Store content to offer in the testnode db
-    let store_request = JsonRpcRequest {
-        method: "portal_historyStore".to_string(),
-        id: 11,
-        params: Params::Array(vec![
-            Value::String(HISTORY_CONTENT_KEY.to_string()),
-            Value::String(HISTORY_CONTENT_VALUE.to_string()),
-        ]),
-    };
+    let store_result = ipc_client
+        .store(content_key.clone(), content_value.clone())
+        .await
+        .unwrap();
 
-    let store_result = make_ipc_request(&peertest_config.target_ipc_path, &store_request).unwrap();
-    assert!(store_result.as_bool().unwrap());
+    assert!(store_result);
 
-    // Send offer request from testnode to bootnode
-    let offer_request = JsonRpcRequest {
-        method: "portal_historyOffer".to_string(),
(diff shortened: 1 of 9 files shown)


ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -68,3 +72,55 @@ pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
...
+            .local_content(serde_json::from_value(json!(HISTORY_CONTENT_KEY)).unwrap())
+            .await
+            .unwrap();
+
+        counter += 1;
+    }
+
+    assert_eq!(
+        content_value, received_content_value,
+        "The received content {received_content_value:?}, must match the expected {HISTORY_CONTENT_VALUE}",
```
njgheorghita, line 124: I'm assuming it doesn't make much of a difference in the displayed value, but I'd update `HISTORY_CONTENT_VALUE` -> `content_value` to avoid slight confusion

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -68,3 +72,55 @@ pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
         "The received content {received_content_str}, must match the expected {HISTORY_CONTENT_VALUE}",
     );
 }
+
+pub async fn test_populate_offer(peertest_config: PeertestConfig, peertest: &Peertest) {
+    info!("Testing Populate Offer/ACCEPT flow");
+
+    let ipc_client = reth_ipc::client::IpcClientBuilder::default()
```
njgheorghita, line 79: This new way of testing jsonrpc methods over ipc differs from that in the previous test (`test_offer_accept`). I'd update the other test to match this new method, for congruency &  so it's easy to tell from reading the file that the same endpoint is being tested.

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -68,3 +72,55 @@ pub fn test_offer_accept(peertest_config: PeertestConfig, peertest: &Peertest) {
         "The received content {received_content_str}, must match the expected {HISTORY_CONTENT_VALUE}",
     );
 }
+
+pub async fn test_populate_offer(peertest_config: PeertestConfig, peertest: &Peertest) {
+    info!("Testing Populate Offer/ACCEPT flow");
```
njgheorghita, line 77: `test_populated_offer` & `Testing Populated Offer/Accept flow`? And rename previous test to `test_unpopulated_offer`?

trin-history/src/jsonrpc.rs
```diff
@@ -184,16 +184,34 @@ impl HistoryRequestHandler {
...
 
-                    let content_key: Vec<RawContentKey> = vec![content_key.as_ssz_bytes()];
-                    let response = match self.network.overlay.send_offer(content_key, enr).await {
-                        Ok(accept) => Ok(json!(AcceptInfo {
-                            content_keys: accept.content_keys,
-                        })),
-                        Err(msg) => Err(format!("SendOffer request timeout: {msg:?}")),
+                    let response = if let Some(content_value) = content_value {
+                        let mut content_item = vec![];
+                        content_value.encode(&mut content_item);
```
njgheorghita, line 192: Would it be preferable to have `content_value.into()` functionality rather than doing this little dance? I'm also not a huge fan of `content_value` and `content_item` naming convention. I think it's a result of #504? Imo it's not obvious from their name what they refer to / what is the difference between them. We've been using `content_value` pretty consistently throughout this codebase to refer to a `content_key`'s complement, which is the cause of my confusion. Moving forward I'd like to adopt one or the other ... unless I'm missing something here, which is totally possible

njgheorghita (approved):
:ship: the only thing I want to say is to re-iterate my desire for some kind of automated docs build / lint check to make sure we have a consistent location for supported jsonrpc api endpoints and their params. It's out of scope for this pr obvs, but if you have any ideas about the best way to go about it and don't mind opening an issue describing how it can be done that would be 🥇 - I can also create a generic issue describing the request if you don't have an obvious contender for how to do this

PR #564, opened 2023-02-23
Use zero radius if empty storage & 0kb config

njgheorghita:
### What was wrong?
`PortalStorage` has some inconsistency w/ respect to the `--kb` flag, which revealed itself while running bridge nodes w/ `--kb 0`. 

Scenario 1
- User runs a fresh `trin` node (eg. random private key) w/ `--kb 0`
  - Problem: Storage would be configured w/ `Distance::MAX` as it's radius instead of `0`
  - Solution: Set radius to `Distance::ZERO` in this scenario

Scenario 2
- User runs an old `trin` node w/ `--kb 1` (when `--kb` was set to a greater number in a previous run)
  - Problem: There was no pruning until the user tried to "store" a new piece of data
  - Solution: Prune db inside `PortalStorage::new()` if capacity is reached on startup, & set relevant fields

Scenario 3
- User runs an old `trin` node w/ `--kb 0`
  - Problem: There was no pruning until the user tried to "store" a new piece of data & storage would be configured with radius set to `Distance::MAX`
  - Solution: Same as Scenario 2 but also set radius to `Distance::ZERO`

### How was it fixed?
- Also removed `farthest_content_id` field from `PortalStorage` struct which served no apparent purpose

### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/564.fixed.md
+++ b/newsfragments/564.fixed.md
@@ -1,0 +1,1 @@
+Update storage to skip pruning db if 0kb selected.
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -17,7 +17,7 @@ use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 use rocksdb::{Options, DB};
 use rusqlite::params;
-use tracing::{debug, error, info};
+use tracing::{debug, info};
 
 use super::types::{
     content_key::OverlayContentKey,
@@ -230,7 +230,6 @@ pub struct PortalStorage {
     node_id: NodeId,
     storage_capacity_in_bytes: u64,
     radius: Distance,
-    farthest_content_id: Option<[u8; 32]>,
     db: Arc<rocksdb::DB>,
     sql_connection_pool: Pool<SqliteConnectionManager>,
     distance_fn: DistanceFunction,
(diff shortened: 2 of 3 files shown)


mrferris (commented):
1. 👍
2. This fixes part of #124. As piper pointed out in that issue, it would be better UX to not block the main thread while pruning. That can be left to a future PR.
3. 👍

4. I'm not sure when `farthest_content_id` stopped being used, but I think that we should re-add it to make our `prune_db` function require half as many SQL calls because we can store the `farthest_content_id` between loop iterations rather than calling `find_farthest_content_id()` twice within the loop like we do now.

njgheorghita:
@mrferris Ah, fair point about unnecessary sql queries. I hadn't really considered that. However, I'm still not a fan of keeping track of this `farthest_content_id = Option<[u8;32]>` attribute. Imo, it's a bit clunky, unwrapping the option is annoying, and it's just another attribute that needs updating whenever we prune....

That being said, I'm not against it and don't think that now is the right time to spend hours refactoring `PortalStorage` (this will come at a later time when performance is important and we have benchmarks that we can test against). But I made a couple changes and wanted to get your thoughts on them. Some things I noticed...
- I introduced a `total_entry_count()` method to check if the store is empty or not. Using `find_farthest_content_id()` for this function seems overkill imo and the method's being used inappropriately
- I moved radius calculation inside `prune_db()`. Imo it makes more sense to handle it there, since everytime we prune the db we should recalculate the radius.
- I introduced a "cache" `farthest_content_id` variable inside prune db. This helps cut down on sql queries when we're pruning lots of data. But we still have two queries when pruning a single value. I'm not sure if this is a big deal or not... these are the kinds of finer details I think we leave till we are in the optimizing stage

mrferris (approved):
👍 on all of the points above.

> But we still have two queries when pruning a single value. I'm not sure if this is a big deal or not

I think it's 👍 and necessary to have two queries in that case

PR #567, opened 2023-02-24
Update discv5 version and add `serde` feature

ogenev:
### To-Do

- [ ] Clean up commit history


trin-history/src/jsonrpc.rs
```diff
@@ -185,8 +180,6 @@ impl HistoryRequestHandler {
                     }
```
njgheorghita, line 180: Can we also simplify the conversion up on line `159`? And lower on `269`?

njgheorghita (approved):
:ship:

PR #572, opened 2023-03-01
Switch native trin-core types for ethportal-api

njgheorghita:
### What was wrong?
I've been trying to use `ethportal-api` types inside `trin-core` - but I've found that since they're all inextricably linked, we can't transition gradually and need to convert them all simultaneously. 

### To-Do

- [ ] Clean up commit history


njgheorghita:
Imo, #504 was rushed, had insufficient test coverage and code review. Resulting in unnecessary work & debate. Relying heavily upon `reth` for types was never fully justified imo. Such a significant refactor that affects the entire codebase requires a :+1: from everybody involved in the project. At the very least - the transition to `reth`-based types should have lived on a separate branch until the entire transition was complete, with passing tests (similar to my `bridge` work).

We now have 18 failing tests after integrating `reth`-based types. These are tests I've debugged before and tbh I have no interest in debugging them again on a whole new set of types. Nor do I think it's a worthwhile use of time.

To be honest, after working on this pr, I've developed some fairly strong opinions against using `reth`-based types. But, I am aware that there are blind spots in my understanding and that I could be the cause of these failing tests. So, I'd like to get a review from @jacob and @ognyan to make sure that I'm using the new types correctly, before we continue with the debate as to whether saving some lines of code justifies introducing these new types & new external dependencies.

jacobkaufmann:
I'm sorry about the frustration that the `reth` refactor is causing you. The scope of #504 was limited to `ethportal-api` and `rpc`, so that was my justification for foregoing a broader set of approvers. I still stand behind the decision to move to `reth`.

Looking at the test failures, it looks like those failures are for tests that should not even exist in `trin`. For example, `receipts::tests::calculate_receipts_root` tests the computation of the receipts Merkle root, but this is not logic that we maintain. I'm not sure why we opted to keep the tests in those files for which all type definitions were removed. The point of moving to `reth` is so that we do not have to have such tests. Ideally, tests are only necessary to test portal-specific encoding of the types.

re: usage of `reth` types, it looks like you have duplicated code that is already provided by `reth` (e.g. [receipts root computation](https://github.com/paradigmxyz/reth/blob/2884eae0755ea2809c4d70dd39863997ca55dfa4/crates/primitives/src/proofs.rs#L63)).

njgheorghita:
@jacobkaufmann No problemo, it's inevitable when it comes to working on a shared codebase. 

> receipts::tests::calculate_receipts_root tests the computation of the receipts Merkle root, but this is not logic that we maintain. 

We do maintain that logic [here](https://github.com/ethereum/trin/blob/master/trin-core/src/types/receipts.rs#L26-L40)

> I'm not sure why we opted to keep the tests in those files for which all type definitions were removed. 

Those type definitions exist in `master`. They were removed in this particular pr as part of the transition over to the `ethportal-api` types. The tests were left in their original location to minimize the diff in this pr. 

> The point of moving to reth is so that we do not have to have such tests.

Sure, but I don't feel comfortable switching out the native `trin-core` types (& tests) for the `reth`-based types until we can plug in the new `reth` based types into the old `trin-core` tests and get green. Then, I'm ok with removing the old types & tests completely. 

> so that was my justification for foregoing a broader set of approvers.

Fair enough, no harm done. Just wanted to point out that since the purpose of the `ethportal-api` types is to eventually replace all of the types inside `trin-core`, I think the scope of the consequences of the refactor is broader than you initially anticipated. 

> re: usage of reth types, it looks like you have duplicated code that is already provided by reth (e.g. [receipts root computation](https://github.com/paradigmxyz/reth/blob/2884eae0755ea2809c4d70dd39863997ca55dfa4/crates/primitives/src/proofs.rs#L63)).

Good point, hadn't noticed those. Integrated that code and got rid of a few failing tests, but we're still dealing with 15 or so failing tests. I'll continue to investigate those

> I still stand behind the decision to move to reth.

For the record, I'm not against this ... eventually... I just want to make sure our old tests are passing with the new types and that we've evaluated all the pros & cons of such a move before committing to it

njgheorghita:
closing since we're moving ahead with #615

PR #574, opened 2023-03-02
Consensus light client implementation

### LINKED ISSUE #570: Research light client integration
Research on beacon chain light client integration with Trin.

ogenev:
### What was wrong?
Closes #570 

Note:  I don't expect a detailed review, as this PR is quite large. The goal is to try to merge the `light-client` code in order to start its integration with Trin in future PRs. This crate encapsulates the light client logic and types and doesn't interfere with other Trin crates (at least for now). It is added as a workspace member in other to manage better the cargo dependencies with the rest of the project.

### How was it fixed?
Most of the code in this crate is a fork of [Helios](https://github.com/a16z/helios) but it doesn't require running 
an execution client alongside the light client.

This light client is modified in a way to be compatible with Portal Network.

This crate can be used as binary with the default configuration:

```
cargo run -p light-client
```

or library:

```rust
use eyre::Result;
use light_client::config::networks;
use light_client::database::FileDB;
use light_client::{Client, ClientBuilder};
use std::path::PathBuf;
#[tokio::main]
async fn main() -> Result<()> {
    // Create a new Helios Client Builder
    let mut builder = ClientBuilder::new();
    // Set the network to mainnet
    builder = builder.network(networks::Network::Mainnet);
    // Set the consensus rpc URL (works with Nimbus node too)
    builder = builder.consensus_rpc("https://www.lightclientdata.org");
    // Set the checkpoint to the last known checkpoint
    builder =
        builder.checkpoint("85e6151a246e8fdba36db27a0c7678a575346272fe978c9281e13a8b26cdfa68");
    // Set the rpc port
    builder = builder.rpc_port(8545);
    // Set the data dir
    builder = builder.data_dir(PathBuf::from("/tmp/light-client"));
    // Set the fallback service
    builder = builder.fallback("https://sync-mainnet.beaconcha.in");
    // Enable lazy checkpoints
    builder = builder.load_external_fallback();
    // Build the client
    let mut client: Client<FileDB> = builder.build().unwrap();
    // Run the client
    client.start().await.unwrap();
    tokio::signal::ctrl_c()
        .await
    Ok(())
}
```

### To-Do

- [x] Clean up commit history


ogenev:
Moving this back to draft until I test its compatibility with the Shapella upgrade after 2 days.

ogenev:
Capella upgrade is supported now.

trin-cli/Cargo.toml
```diff
@@ -28,3 +28,4 @@ ureq = { version = "2.5.0", features = ["json"] }
 [[bin]]
 name="trin-cli"
 path="src/main.rs"
+
```
carver, line 31: *nit* maybe we can remove this change that doesn't do anything?

carver (approved):
Yeah, as you mentioned on the call, there's no way we're going to be able to meaningfully review all 11k lines of code.

I guess the next thing this makes me wonder is: how do we use it?

You mentioned using it in the bridge, maybe to push some version of the consensus header into the history network? Or were you thinking something else?

The burning question on my mind that I'm not sure is answered either way: can we run light-client alongside a regular trin node as a way to verify canonicalness of a header? Surely light-client is answering that question itself in some way. So can we run light-client and hack it to expose an API, so we can ask light-client that question?

docker/Dockerfile
```diff
@@ -18,6 +18,7 @@ COPY ./Cargo.toml ./Cargo.toml
 COPY ./ethportal-api ./ethportal-api
 COPY ./ethportal-peertest ./ethportal-peertest 
 COPY ./portalnet ./portalnet
+COPY ./light-client ./light-client
 COPY ./rpc ./rpc
```
njgheorghita, line 22: Let's also update `docker/Dockerfile.bridge`

light-client/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+eyre = "0.6.8"
+figment = { version = "0.10.7", features = ["toml", "env"] }
+jsonrpsee = { version = "0.15.1", features = ["full"] }
+serde = { version = "1.0.143", features = ["derive"] }
+serde_yaml = "0.9.14"
+tracing-subscriber = "0.3.15"
+hex = "0.4.3"
+reqwest = { version = "0.11.13", features = ["json"] }
+serde_json = "1.0.85"
+ssz-rs = { git = "https://github.com/ralexstokes/ssz-rs", rev = "d09f55b4f8554491e3431e01af1c32347a8781cd" }
```
njgheorghita, line 24: Hahaha, so much for our efforts to use a single ssz library in trin - I guess this is unavoidable though

njgheorghita (approved):
> You mentioned using it in the bridge, maybe to push some version of the consensus header into the history network? 

I'm sure it's possible... but my initial reaction is that this isn't necessary since the bridge already has access to that data via full consensus clients via pandaops

light-client/Cargo.toml
carver: haha, oh noes!

ogenev:
> The burning question on my mind that I'm not sure is answered either way: can we run light-client alongside a regular trin node as a way to verify canonicalness of a header? Surely light-client is answering that question itself in some way. So can we run light-client and hack it to expose an API, so we can ask light-client that question?

Yeah, I think this would be the main use case. As Nick mentioned, for a bridge node we can probably use pandaops.

We are discussing different mechanics with Kim to verify canonicalness with the Beacon chain light client network and a slightly modified/hacky version of a light client (for example storing recent X number of `LightClientUpdate`/`LightClientHeader`) seems a reasonable shortcut for such verification, before the actual launch of the Beacon chain light client network (however, after launching of the network, every portal node would need to implement most of the light client mechanics to follow the chain anyways, so some parts of this implementation can be reused too).

PR #575, opened 2023-03-02
Print version

njgheorghita:
### To-Do

- [ ] Clean up commit history


Cargo.toml
```diff
@@ -1,17 +1,20 @@
...
 edition = "2021"
 rust-version = "1.66.1"
 default-run = "trin"
 repository = "https://github.com/ethereum/trin"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+license = "GPL-3.0"
+readme = "README.md"
+keywords = ["ethereum", "portal-network"]
+categories = ["cryptography::cryptocurrencies"]
```
njgheorghita, line 12: Other options [here](https://crates.io/category_slugs)

Cargo.toml
```diff
@@ -1,17 +1,20 @@
 [package]
 name = "trin"
 version = "0.1.0"
-authors = ["Jacob Kaufmann <jacobkaufmann18@gmail.com>", "Jason Carver <ut96caarrs@snkmail.com>"]
+authors = ["https://github.com/ethereum/trin/graphs/contributors"]
```
njgheorghita, line 4: Not trying to step on any toes w/ this change, happy to revert to individual contributors, but this seems like a more sustainable approach

ethportal-api/Cargo.toml
```diff
@@ -2,30 +2,27 @@
 name = "ethportal-api"
 version = "0.2.0"
 edition = "2021"
+rust-version = "1.66.1"
```
ogenev, line 5: Not sure if we want to add this requirement on a crate that is intended for external use too.

ethportal-peertest/Cargo.toml
```diff
@@ -2,12 +2,17 @@
 name = "ethportal-peertest"
 version = "0.1.0"
 edition = "2021"
-rust-version = "1.62.0"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+rust-version = "1.66.1"
```
ogenev, line 5: Do we need this in every sub-crate Cargo file? Haven't tried it, but maybe only include rust-version in the main Cargo.toml will be enough.

Cargo.toml
```diff
@@ -1,17 +1,20 @@
...
 edition = "2021"
 rust-version = "1.66.1"
 default-run = "trin"
 repository = "https://github.com/ethereum/trin"
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+license = "GPL-3.0"
```
carver, line 9: I don't think we use GPL anywhere else. Are we using libraries that force us to use GPL (because they are GPL too)? Most everything in snake charmers uses the MIT license. I know Piper feels strongly about not using GPL. We can make our own decision, of course, but we should decide it explicitly all together.

ethportal-api/Cargo.toml
```diff
@@ -5,27 +5,23 @@ edition = "2021"
 description = "Definitions for various Ethereum Portal Network JSONRPC APIs"
 license = "GPL-3.0"
 repository = "https://github.com/ethereum/trin/tree/master/ethportal-api"
+readme = "README.md"
```
carver, line 8: Weird, I must have missed it when someone added GPL to this file. I don't think we've actually agreed on this as a group anywhere. The most important place for it to be would be the LICENSE file in the repo, which is missing.

carver (commented):
I'll make a tracking issue for resolving the OS license we choose.

trin-core/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 #![warn(clippy::unwrap_used)]
+#![warn(clippy::cargo)]
```
carver, line 2: Any reason this is just in trin-core? Could/should we enable for every module?

ethportal-api/Cargo.toml
njgheorghita: We discussed it on the sync, I think it was last thursday, when you were out. But happy to revisit this decision

trin-core/src/lib.rs
njgheorghita: I was getting some inconsistent behavior tbh. Which is why I only left it in here. However, I was also using `#![warn(unused_crate_dependencies)]` which is a rustc lint. This might have been contributing to the inconsistencies, since I don't believe `rustc` lints and `clippy` lints always play nicely. I could be very wrong about that though. Regardless, yeah both of those lints are very useful and should be enabled everywhere imo - trac

PR #576, opened 2023-03-03
Replace r2d2 connection pool with rusqlite connection

njgheorghita:
### What was wrong?
<s>`r2d2_sqlite` is heavy. When building a docker image, compilation for `r2d2_sqlite` alone accounts for ~45% of the compilation time</s>. We don't require the connection pool, and can use `rusqlite`'s `Connection` instead. Wrapped in a `Mutex` because [Connection is not Send + Sync](https://github.com/rusqlite/rusqlite/issues/342). It seems to me like this change should have no impact on performance as we start supporting multiple sub-networks.

### How was it fixed?
- Switch out `Pool<SqliteConnectionManager>` in favor of `Arc<Mutex<Connection>>`
- Remove dependency on `r2d2_sqlite`

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3943,27 +3943,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "r2d2"
-version = "0.8.10"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "51de85fb3fb6524929c8a2eb85e6b6d363de4e8c48f9e2c2eac4944abc181c93"
-dependencies = [
- "log 0.4.17",
- "parking_lot 0.12.1",
- "scheduled-thread-pool",
-]
-
-[[package]]
-name = "r2d2_sqlite"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "54ca3c9468a76fc2ad724c486a59682fc362efeac7b18d1c012958bc19f34800"
-dependencies = [
- "r2d2",
- "rusqlite",
-]
-
-[[package]]
 name = "radium"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


njgheorghita:
Well, on further inspection... it seems like `r2d2_sqlite` is not the problem. Docker has a strange habit of appending a line to the compilation logs only after the step is complete...
```
   Compiling uds_windows v1.0.2
   Compiling rusqlite v0.26.3
   Compiling r2d2_sqlite v0.19.0
```
It was hanging here, but it turns out that at this point, `r2d2_sqlite` has already been built, and the culprit is actually `rocksdb` - which makes more sense and is unavoidable.

However, imo this pr is still a (very, very slight) improvement - so I'll leave it open for review

ogenev:
By removing the connection pool, we no longer support concurrent sqlite access from multiple threads. I'm not sure if the benefits outweigh the cons here.

njgheorghita:
@ogenev I'm not sure that's the case here. The new sql `Connection` is wrapped in `Arc<Mutex<..>>`. Which allows for concurrent use across threads. We create the connection inside `PortalStorageConfig::new()` - and then clone the `Arc<..>` to provide a handle for threads to access. So, given the current setup, it seems to me like we do support concurrent sqlite access from multiple threads. Unless I'm misunderstanding here?

I think you would be correct if we changed our db setup. eg. we instantiate a new `Config` on every subnetwork/thread. However, this doesn't seem desirable since it would break our rocksdb setup (which doesn't use a pool and creates a single connection and then clones an `Arc` around to different threads. Which is essentially what this pr is doing to our sql db).

But, if I'm misunderstanding and this pr does break concurrent sqlite access from multiple threads, then yes I definitely agree the benefits do not outweigh the cons here.

ogenev:
> @ogenev I'm not sure that's the case here. The new sql `Connection` is wrapped in `Arc<Mutex<..>>`. Which allows for concurrent use across threads.

I meant multi-threaded access to SQLite or parallelism. Mutexes allow access to data from one thread at a time. To access the data in a mutex, a thread must first signal that it wants to access by asking to acquire the mutex’s lock.

A connection pool allows more than one thread at the same time to access the DB without waiting to acquire the mutex lock.

PR #578, opened 2023-03-04
Change web3-transport cli argument to enum instead of string

### LINKED ISSUE #535: Change `web3-transport` cli argument to enum instead of string
Web3 transport is either `ipc` or `http`, so it makes sense to use an enum instead of a string:

https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/trin-core/src/cli.rs#L33

julio4:
### What was wrong?

Fixes #535
Web3 transport is either `ipc` or `http`

### How was it fixed?

Wrap it in a `Web3TransportType` enum in `trin_core::utils::transport`

### To-Do

- [x] Is `trin_core::utils::transport` the best location ?
- [x] Clean up commit history

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,7 @@ use tracing::info;
 
 use trin_core::jsonrpc::types::HistoryJsonRpcRequest;
 use trin_core::{
-    cli::{TrinConfig, HISTORY_NETWORK, STATE_NETWORK},
+    cli::{TrinConfig, Web3TransportType, HISTORY_NETWORK, STATE_NETWORK},
     portalnet::{
         discovery::Discovery, events::PortalnetEvents, storage::PortalStorageConfig,
         types::messages::PortalnetConfig,
@@ -161,8 +161,8 @@ async fn launch_jsonrpc_server(
     let history_handler = history_handler.ok_or_else(|| {
         "History network must be available to use IPC transport for JSON-RPC server".to_string()
     })?;
-    match trin_config.web3_transport.as_str() {
-        "ipc" => {
+    match trin_config.web3_transport {
+        Web3TransportType::IPC => {
             // Launch jsonrpsee server with IPC transport
             let rpc_handle =
                 JsonRpcServer::run_ipc(trin_config.web3_ipc_path, discv5, history_handler)
@@ -171,7 +171,7 @@ async fn launch_jsonrpc_server(
             info!("IPC JSON-RPC server launched.");
             Ok(rpc_handle)
         }
-        "http" => {
+        Web3TransportType::HTTP => {
             // Launch jsonrpsee server with http and WS transport
             let rpc_handle =
                 JsonRpcServer::run_http(trin_config.web3_http_address, discv5, history_handler)
@@ -180,6 +180,5 @@ async fn launch_jsonrpc_server(
             info!("HTTP JSON-RPC server launched.");
             Ok(rpc_handle)
         }
-        val => panic!("Unsupported web3 transport: {val}"),
     }
 }
--- a/trin-core/src/cli.rs
+++ b/trin-core/src/cli.rs
@@ -1,4 +1,4 @@
-use std::{env, ffi::OsString, fmt, net::SocketAddr, path::PathBuf};
+use std::{env, ffi::OsString, fmt, net::SocketAddr, path::PathBuf, str::FromStr};
 
 use structopt::StructOpt;
 use url::Url;
@@ -15,6 +15,46 @@ pub const STATE_NETWORK: &str = "state";
 const DEFAULT_SUBNETWORKS: &str = "history";
 pub const DEFAULT_STORAGE_CAPACITY: &str = "100000"; // 100mb
 pub const DEFAULT_TRUSTED_PROVIDER: &str = "infura";
+pub const DEFAULT_WEB3_TRANSPORT: &str = "ipc";
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum Web3TransportType {
+    HTTP,
+    IPC,
+}
+
+impl fmt::Display for Web3TransportType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::HTTP => write!(f, "http"),
+            Self::IPC => write!(f, "ipc"),
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct ParseWeb3TransportError;
+
+impl fmt::Display for ParseWeb3TransportError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
+            f,
+            "Invalid web3-transport arg. Expected either 'http' or 'ipc'"
+        )
+    }
+}
+
+impl FromStr for Web3TransportType {
+    type Err = ParseWeb3TransportError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "http" => Ok(Web3TransportType::HTTP),
+            "ipc" => Ok(Web3TransportType::IPC),
+            _ => Err(ParseWeb3TransportError),
+        }
+    }
+}
 
 #[derive(StructOpt, Debug, PartialEq, Clone)]
 #[structopt(
@@ -25,12 +65,11 @@ pub const DEFAULT_TRUSTED_PROVIDER: &str = "infura";
 )]
 pub struct TrinConfig {
     #[structopt(
-        default_value = "ipc",
-        possible_values(&["http", "ipc"]),
+        default_value(DEFAULT_WEB3_TRANSPORT),
         long = "web3-transport",
         help = "select transport protocol to serve json-rpc endpoint"
     )]
-    pub web3_transport: String,
+    pub web3_transport: Web3TransportType,
 
     #[structopt(
         default_value(DEFAULT_WEB3_HTTP_ADDRESS),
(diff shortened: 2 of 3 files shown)


trin-core/src/utils/mod.rs
```diff
@@ -4,3 +4,4 @@ pub mod db;
 pub mod node_id;
 pub mod portal_wire;
 pub mod provider;
+pub mod transport;
```
ogenev, line 7: We don't have clear architecture patterns and it is confusing where to put those kinds of types but I think his utils module is more for helper methods used for some kind of calculations/conversions.

I would put this type in `trin-core/src/cli.rs` for now. We are planning to decomposite `trin-core` in smaller crates and at some point, we would like all the CLI stuff to be in its own crate. Then having a `types.rs` file in `cli` crate seems reasonable.

trin-core/src/utils/transport.rs
```diff
@@ -0,0 +1,31 @@
...
+impl fmt::Display for Web3TransportType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Http => write!(f, "http"),
+            Self::Ipc => write!(f, "ipc"),
+        }
+    }
+}
+
+type ParseError = &'static str;
```
ogenev, line 19:
```suggestion
struct ParseWeb3TransportError;
```

trin-core/src/utils/transport.rs
```diff
@@ -0,0 +1,31 @@
...
+
+impl FromStr for Web3TransportType {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "http" => Ok(Web3TransportType::Http),
+            "ipc" => Ok(Web3TransportType::Ipc),
+            _ => panic!("Invalid web3-transport arg: {s} is not an option."),
```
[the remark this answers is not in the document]
julio4: I agree that error is better, maybe we can fallback to default value "ipc" and issue a warning/error ?
But error handling with StructOpt are a bit tricky.

trin-core/src/cli.rs
```diff
@@ -15,6 +16,43 @@ pub const STATE_NETWORK: &str = "state";
...
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct ParseWeb3TransportError;
+
+impl fmt::Display for ParseWeb3TransportError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Invalid web3-transport arg.")
```
njgheorghita, line 41: `"Invalid --web3-transport arg. Expected either 'http' or 'ipc'"`

trin-core/src/cli.rs
```diff
@@ -180,18 +218,28 @@ impl TrinConfig {
...
+            Err(error) => {
+                match error.kind {
+                    ErrorKind::ValueValidation => {
+                        println!("Fallback to default value, cause: {}", error);
+                    }
+                    _ => panic!("Could not create configuration with arguments"),
+                };
+                Default::default()
+            }
+        };
```
julio4, line 232: Allow error handling when creating `TringConfig` from `structopt::StructOpt::from_iter(args)`. 
There are downsides to this:
- It will fallback to all default values, even if only one argument is wrong.
-  We should switch from using `println` to `tracing::{warn/error}`, but the error message from `clap::Error` uses Unicode for coloring, which can make it difficult to read.

I'm sure there might be a better solution.

trin-core/src/cli.rs
```diff
@@ -180,18 +218,28 @@ impl TrinConfig {
         I: Iterator<Item = T>,
         T: Into<OsString> + Clone,
     {
-        let config = Self::from_iter(args);
+        let config = match Self::from_iter_safe(args) {
```
njgheorghita, line 221: I'm not sure I agree with this approach. If a user provides invalid flags, it seems to me like we should panic, and tell them explicitly what went wrong, rather than silently falling back to using the default flags

julio4: I agree that this is not the perfect solution. See my previous comment. 
I wasn't sure if we should panic, but it makes sense to panic to ensure that the user's provided config is used.

njgheorghita: Yup, I agree panic is the way to go here. Seems to me like we can just leave it as `let config = Self::from_iter(args);` which will [panic on failure](https://docs.rs/structopt/latest/structopt/trait.StructOpt.html#method.from_iter).

julio4: I thought the same first, but it don't really panic, [it prints the error and call `process::exit(1);`](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from).
As we handle other cases with panic, it might be better to panic here as well, so we can use `should_panic` in tests.

julio4: > Yup, I agree panic is the way to go here. Seems to me like we can just leave it as `let config = Self::from_iter(args);` which will [panic on failure](https://docs.rs/structopt/latest/structopt/trait.StructOpt.html#method.from_iter).

I thought the same first, but it don't really panic, [it prints the error and call process::exit(1);](https://docs.rs/clap/2.33.3/clap/struct.App.html#method.get_matches_from).
As we handle other cases with panic, it might be better to panic here as well, so we can use should_panic in tests.

trin-core/src/cli.rs
```diff
@@ -301,7 +344,7 @@ mod test {
         let expected_config = TrinConfig {
             web3_http_address: Url::parse("http://0.0.0.0:8080/").unwrap(),
             pool_size: 3,
-            web3_transport: "http".to_string(),
+            web3_transport: Web3TransportType::HTTP,
```
njgheorghita, line 347: You should be able to delete all of these `web3_transport` attribute definitions since it's already defined in `default()`. And while you're at it, might as well remove all of the `web3_http_address` definitions as well. In the cases where we're just re-defining the default value.

njgheorghita: Ahh, nvm. Just noticed that default is actually ipc. But you can still get rid some of the values in the tests below.

trin-core/src/cli.rs
njgheorghita: Ahh, gotcha. Yeah, it seems like we panic on those other cases explicitly inside `from_cli()`. But, in this case it makes sense to use structopt's native type casting. Could we update it to something like...
```
// Panic on invalid unwrap of `from_iter_safe()` to force panic for testing purposes.
let config = Self::from_iter_safe(args).unwrap_or_else(|e| panic!("{e}"));
```

trin-core/src/cli.rs
julio4: I see what you mean but are you sure it's necessary ? Are we sure we will never change default values ?

I feel like we should write tests while abstracting default values. Also it helps make tests more readable as we see what we are testing.

trin-core/src/cli.rs
julio4: Thank you for the rust syntax, it's much cleaner!

trin-core/src/cli.rs
njgheorghita: It's not necessary, but it helps remove unnecessary, redundant visual clutter from the tests. Imo, the default behavior is just that, the default behavior. It should be assumed in all of these tests. No doubt, default values will be changed, but the necessary adjustments to these tests can be made then.

Ultimately, I'd evaluate it on a case-by-case basis. If you think that redefining a default attribute is critical information to communicate to a reader trying to understand the test, then definitely leave it in. Otherwise, I'd cut it. Up to you how you want to handle these cases.

njgheorghita:
Looks good! Take a look at my comment, then rebase off `master` to get rid of the merge conflicts, and squash your commits, then we should be good to :ship: thanks!

njgheorghita (approved):

julio4:
Thanks for all the comments, they were very helpful. It should be ready to go !

src/lib.rs
```diff
@@ -161,17 +161,17 @@ async fn launch_jsonrpc_server(
     let history_handler = history_handler.ok_or_else(|| {
         "History network must be available to use IPC transport for JSON-RPC server".to_string()
     })?;
-    match trin_config.web3_transport.as_str() {
-        "ipc" => {
-            // Launch jsonrpsee server with IPC transport
```
ogenev, line 166: This comment looks correct to me, `Web3TransportType::IPC` means we run the server with IPC transport, not HTTP and WS.

trin-core/src/cli.rs
```diff
@@ -172,18 +212,17 @@ impl TrinConfig {
         I: Iterator<Item = T>,
         T: Into<OsString> + Clone,
     {
-        let config = Self::from_iter(args);
+        let config = Self::from_iter_safe(args).unwrap_or_else(|e| panic!("{e}"));
```
ogenev, line 215: Why are we changing this from [from_iter](https://docs.rs/structopt/latest/structopt/trait.StructOpt.html#method.from_iter)?

Also `unwrap_or_else` and then panicking doesn't make sense to me, isn't the same as just `.unwrap()`?

src/lib.rs
julio4: Typo, my bad

trin-core/src/cli.rs
julio4: I think this is a way to explicitly state that we have handled the `unwrap` here and are using the default error message instead of a custom `expect` message ?

It was suggested by @njgheorghita

perama-v (approved):

ogenev (approved):

trin-core/src/cli.rs
carver: This introduced a panic during `--help` or `--version`.See https://docs.rs/structopt/0.3.7/structopt/trait.StructOpt.html#method.from_iter_safe

New test to detect the panic, plus the fix, in #617

PR #585, opened 2023-03-09
Create trin-utils workspace

njgheorghita:
### What was wrong?
Trin is moving towards a more modular structure, eg. splitting out `trin-core` into separate workspaces. This pr creates `trin-utils` - a crate for shared utils among trin workspaces. If a util is only used by a single workspace members, it should live in a root-level `utils` module. 

This pr only moves the `bytes` utils from `trin-core` into `trin-utils`. This is to..
- minimize the diff - since this is the most widely used util
- to stabilize our dependence on `0x` prefixed hexstrings. Our previous convention of using both `0x`-prefixed and non-prefixed hexstrings has caused some bugs. 

Other `trin-core` utils should be moved accordingly as we work towards breaking up `trin-core`

### How was it fixed?
- Moved `trin_core::bytes` utils into `trin-utils`
- Removed all dependencies on `hex` crate (except inside `trin-utils`)

### To-Do

- [x] update dockerfile
- [x] Clean up commit history


carver (approved):
LGTM

PR #588, opened 2023-03-10
Extract core jsonrpc types & create trin-types

njgheorghita:
### What was wrong?
Extract `trin-core/src/jsonrpc/*` & create `trin-types` workspace
Motivation. To extract our validation logic, we need to first extract some of these types so they're accessible from a different workspace without introducing a circular dependency.
...
- `trin_core/src/jsonrpc/endpoints.rs` -> `ethportal_api/src/endpoints.rs`
- `trin_core/src/jsonrpc/service.rs` -> `ethportal_api/src/service.rs`
- `trin_core/src/jsonrpc/types.rs`
  - `Params` -> `ethportal-api/src/types/params.rs`
  - `Request` -> `ethportal-api/src/types/request.rs`
- `trin_core/src/jsonrpc/utils/rs` -> `trin_utils/src/bucket.rs`
  - remove unused `parse_content_item()` util
- `trin_core/src/portalnet/types/messages.rs`
  - `SszEnr` -> `trin_types/src/enr.rs`
- `trin_core/src/portalnet/mod.rs`
  - `Enr` -> `trin_types/src/enr.rs`
- `trin_core/src/portalnet/types/distance.rs` -> `trin-types/src/distance.rs`

### To-Do

- [x] Clean up commit history


carver (approved):
I don't have a strong opinion on how rust types should be organized. Seems fine to me, as a way toward adding a trin bridge app. 👍🏻

njgheorghita:
Going ahead with the merge on this one - since it's a blocker on extracting the validation types. But it's worth pinging @ogenev for a retrospective review on this. Mostly since I've moved some types into `ethportal-api`. Since they were all jsonrpc related, it felt like an appropriate home for them. But, I'd like to get confirmation on that. If you feel like they are better housed inside `trin-types`, I'm happy to make that change in a future pr

ogenev:
@njgheorghita  I think we should keep `ethportal-api` as encapsulated as possible and not couple it with Trin (i.e. trin specific types).

It seems for types that are Trin specific, a better place would be the newly created `trin-types` crate in `jsonrpc.rs` file for example.

PR #589, opened 2023-03-11
GitHub Action to deploy mdbook docs from master

carver:
### What was wrong?

mdbook isn't building. 404 at https://ethereum.github.io/trin

### How was it fixed?

Add the default GitHub Action recipe for mdbook. It's the first time I'm doing it, so this may not work. 🤷🏻‍♂️ 

I don't really have a theory for how to test a GitHub Action, besides merging it and seeing what happens.

### DIFF
--- a/.github/workflows/mdbook.yml
+++ b/.github/workflows/mdbook.yml
@@ -1,0 +1,62 @@
+# Sample workflow for building and deploying a mdBook site to GitHub Pages
+#
+# To get started with mdBook see: https://rust-lang.github.io/mdBook/index.html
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
Sounds reasonable to push this through until we figure out why it's working / not working. Is this workflow meant to replace our `.github/workflows/book.yml` workflow? There seems to be a lot of overlap between the two. But, as long as we consolidate the workflows once everything's working, I'm 👍

carver:
Closing in favor of #590

PR #591, opened 2023-03-14
Remove websocket from trusted provider

njgheorghita:
### What was wrong?
The logic for generating `TrustedProvider` when using panda ops caused a panic, which occurred when deploying the bridge nodes to the testnet. Imo, it's worth removing the `ws` attribute entirely from `TrustedProvider`
- The pandaops nodes don't support ws
- The decision to have a ws option was back when we were targeting wallet users - and expected to proxy to infura quite frequently. Now, we're targeting research and expect to remove any infura relience in the near future.

If we want to leave in the `ws` attribute, it's fairly simple bugfix to avoid the panic. But, imo since we don't need `ws` we might as well remove it entirely

### How was it fixed?
Removed `ws` attribute from `TrustedProvider`

### To-Do

- [x] Clean up commit history


carver (approved):
🚢

PR #593, opened 2023-03-15
Move params, provider and cli to trin-types

njgheorghita:
### To-Do

- [ ] Clean up commit history


carver (approved):
LGTM

👍🏻 for breaking up into smaller PRs

ethportal-api/src/types/request.rs
ogenev, line 1: As mentioned in https://github.com/ethereum/trin/pull/588, I think we should move the types in this file to `trin-types` crate and not couple `ethportal-api` with trin.

trin-types/src/cli.rs
ogenev, line 1: I'm more in favor of creating `trin-cli` crate responsible for all CLI logic and types because the CLI functionality is by itself an important component of Trin.

trin-types/src/params.rs
ogenev, line 1: Could we be more specific on the file name, what kind of `params` is this?

trin-types/src/receipts.rs
ogenev, line 1: I think it is worth creating an `execution` folder for example inside `trin-types` where we will put all execution types.

ogenev (commented):
Nice direction forward decomposition of `trin-core` :+1: ! Left some comments to discuss before approval.

ethportal-api/src/types/request.rs
njgheorghita: So this pr should be thought of more as the "low-hanging" / "easy to extract" types refactor. It certainly won't be the last one. The problem with moving `ethportal-api/src/types/request.rs` types into `trin-types` is that it would also require moving a lot more types (since `ethportal-api` depends on `trin-types`, so `trin-types` cannot have access to any types inside `ethportal-api`) - at least all of the endpoint types and other downstream type dependencies. I don't disagree that this move should happen, but I'd rather explore the change/extraction in a following pr to have a cleaner understanding of the changes required.

trin-types/src/cli.rs
njgheorghita: I don't have a **strong** feeling on this one, but I do have a feeling. Your suggestion here feels like "over-modularization" to me. Essentially `trin-cli` would be a workspace with only a single file. 

Plus we already have a `trin-cli` workspace (and iirc, we decided to leave that in for now since we might be able to put it to use later).

So at the very least, I'd suggest leaving this type here (for this pr) until we figure out what to do with the existing `trin-cli`.

ethportal-api/src/types/request.rs
ogenev: In theory,  `ethportal-api` should not depend on trin internals - i.e.  `trin-types` but I agree that this is out of the scope of this PR.

ogenev (approved):
:boat:

PR #598, opened 2023-03-17
Extract validation logic into trin-validation

njgheorghita:
### What was wrong?
- `trin_core::types::validation::HeaderOracle` -> `trin_validation::oracle::HeaderOracle`
- `trin_core::types::validation::Validator` -> `trin_validation::validator::Validator`
- `trin_core::types::accumulator::MasterAccumulator` -> `trin_validation::accumulator::MasterAccumulator`
- consolidated `ethportal_api::OverlayContentKey` & `trin_core::..::OverlayContentKey` definitions into `ethportal_api::content_key::OverlayContentKey` definition

blocked by #593 

### How was it fixed?
- [x] move associated test assets

### To-Do

- [x] Clean up commit history


njgheorghita:
@ogenev w/ regards to the question of whether we keep `ethportal-api` standalone or not... imo, for the sake of moving forward let's ignore that question for this particular pr - since it's not the primary focus of this pr. We'll continue the discussion, and we can move types back into `ethportal-api` in a future pr if that's the course we decide upon

ethportal-api/src/types/content_key.rs
ogenev, line 1: I don't understand why we want to remove those types from `ethportal-api`. Those are  already used in Glados for example [here](https://github.com/pipermerriam/glados/blob/8ad0255c78ceffdd91ecc8a30c1f237efd916df6/glados-core/src/jsonrpc.rs#L13) and [here](https://github.com/pipermerriam/glados/blob/8ad0255c78ceffdd91ecc8a30c1f237efd916df6/glados-audit/src/lib.rs#L9). Having them in `ethportal-api` is critical for the external usage of the library.

ogenev (changes requested):

ethportal-api/src/types/content_key.rs
njgheorghita: True, but we could just as easily re-export them in `ethportal-api` to be used in `glados` if they're defined inside `trin-types`. Either way... I've updated this pr, so that content keys are defined inside `ethportal-api` and I moved `Request` to `trin-types`. It honestly doesn't seem to make much of a difference so I'm probably being pedantic here but it just makes so much more sense to me to define all of our types inside `trin-types`....

njgheorghita: It seems the issue at debate here is whether or not we want to maintain the current definition for the `trin-types` crate. (from its readme).
```
Library for shared types across trin workspace members. Types only used by a single workspace member belong in a root-level types module.
```

carver (approved):
No objection 👌🏻

ethportal-api/src/types/content_key.rs
```diff
@@ -1,12 +1,86 @@
...
 
 use trin_utils::bytes::{hex_decode, hex_encode, hex_encode_compact};
 
+/// SSZ encoded overlay content key as bytes
+pub type RawContentKey = Vec<u8>;
+
+/// A content key type whose content id is the inner value. Allows for the construction
+/// of a content key with an arbitary content ID.
+#[derive(Clone, Debug)]
+pub struct IdentityContentKey {
```
ogenev, line 20: I don't think we want to support and expose this struct to the library users of `ethportal-api`. If I get this right, `IdentityContentKey` is used only for testing in Trin and it is not part of portal network specs and the global protocol.

As long as `ethportal-api` expose the `OverlayContentKey` trait, any user has all the tools to create its own internal content key structs (for testing or etc).

I would move this type to `trin-types` crate, some for the `RawContentKey` type above.

ogenev, line 20: I don't think we want to support and expose this struct to the library users of `ethportal-api`. If I get this right, `IdentityContentKey` is used only for testing in Trin and it is not part of portal network specs and the global protocol.

As long as `ethportal-api` expose the `OverlayContentKey` trait, any user has all the tools to create its own internal content key structs (for testing and etc).

I would move this type to `trin-types` crate, same for the `RawContentKey` type above.

ethportal-api/src/endpoints.rs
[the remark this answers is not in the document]
njgheorghita: No worries, I missed this one. I have no problem moving more types into `trin-types` 😅

ogenev (approved):
Looks good! :+1:  :boat:

Issue #599, opened 2023-03-17
Enable `clippy::cargo` and/or `unused_crate_dependencies` lints for every workspace


njgheorghita:
As the title says, it would be nice to have these lints enabled for every workspace
more context : https://github.com/ethereum/trin/pull/575#discussion_r1140504504

[UNUSED_CRATE_DEPENDENCIES](https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint_defs/builtin/static.UNUSED_CRATE_DEPENDENCIES.html) seems to have some peculiar edge cases

e3243eric:
I tried to enable `clippy::cargo` or `unused_crate_dependencies` for all workspaces. But in some cases, both linters have false positives, and this cause a bunch of lint errors in Trin.

- `clippy::cargo`
  - [multiple_crate_versions](https://rust-lang.github.io/rust-clippy/stable/index.html#multiple_crate_versions) is one of lint in `clippy::cargo`. It has a known issue for errors caused by the dependencies themselves.
  - Like in Trin, `reth-primitives` has a dependency on `hashbrown@0.13.2`, but `eth_trie` has a dependency on `hashbrown @0.3.1`, then the lint will issue `multiple versions for dependency hashbrown`.
  - Related [Issue#9756](https://github.com/rust-lang/rust-clippy/issues/9756) on clippy.
- [`unused_crate_dependencies`](https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint_defs/builtin/static.UNUSED_CRATE_DEPENDENCIES.html)
  - As docs said, it has a known issue for multiple create in a cargo package.
  - Like in the Trin package (trin/src), the package has a dependency on `anyhow`. `anyhow` is used in `bin/purge_db.rs` but not in `lib.rs`, then the linter will incorrectly issue `anyhow` is unused in the `lib.rs`.
  - Related [Issue#95513](https://github.com/rust-lang/rust/issues/95513) on rust.

njgheorghita:
@e3243eric Thanks for looking into this! Yeah, I've played around with these and was never able to find a stable configuration, I ran into a lot of the errors you're describing. So it's nice to get some confirmation. I wasn't aware of the `multiple_crate_versions` lint, and it would be so nice to get something like that in...

[...]

PR #601, opened 2023-03-21
Add 'clippy::cargo' and 'unused_crate_dependencies' lints to Makefile

e3243eric:
### What was wrong?

Related to Issue #599.

### How was it fixed?

Add a new Makefile `lint-optional` command. This command includes the original lint checks, and adds the `clippy::cargo` and the rustc lint `unused_crate_dependencies`. It will bump up warnings and includes some false positives as #599 said.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 help:
 		@echo "lint - run clippy and rustfmt"
+		@echo "lint-unstable - run clippy, rustfmt and rustc lints with unstable features. expect errors, which cannot be resolved, so the user must step through and evaluate each one manually."
 		@echo "notes - generate release notes"
 		@echo "release - publish a new release"
 		@echo "create-docker-image - create docker image"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -9,6 +10,11 @@ lint:
 		cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
 		cargo fmt --all -- --check
 
+lint-optional:
```
njgheorghita, line 13: I would add a comment here explaining this lint - why these lint levels are left out of the main lint config

njgheorghita (commented):
Since we're relying on unstable linting features, it seems to me that `lint-unstable` might be a more suitable name for this command

Makefile
```diff
@@ -9,6 +10,11 @@ lint:
...
+		cargo clippy --all --all-targets --all-features --no-deps -- -Wclippy::cargo
+		cargo fmt --all -- --check
+		RUSTFLAGS="-W unused_crate_dependencies" cargo build
```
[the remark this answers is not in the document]
e3243eric: It seems it is not a similar lint in [rustc](https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html).

e3243eric:
Nice naming! I got stuck for about five minutes on naming.

Makefile
```diff
@@ -9,6 +10,11 @@ lint:
 		cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
```
e3243eric, line 10: I didn't copy `--deny warnings` because it will stop the script. I think it's better to list all warnings at once. Another reason is that lint has expected errors.

njgheorghita:
Thanks @e3243eric 🥳

PR #602, opened 2023-03-21
Shortening version hash bugfix

### LINKED ISSUE #600: `client_version` json-rpc endpoint panic
#595 introduced a bug that causes the `client_version` json-rpc endpoint to panic with the following error:

```sh
thread 'main' panicked at 'byte index 6 is out of bounds of `HEAD`', trin-utils/src/version.rs:9:9
```

This is because the latest git hash on master is `HEAD` which is less than 6 characters specified [here.](https://github.com/ethereum/trin/blob/5a3ae14ec9b11ae92cf047d1d7d86aa50a7bfde1/trin-utils/src/version.rs#L9).

This also causes portal-hive to [fail](https://portal-hive.ethdevops.io/viewer.html?file=results/trin/client-cf0b45b49168f443d008a8bceaaab61440c99eafe8a29bad9d4be86a0851c2eb.log) to launch the Trin client.

njgheorghita:
### What was wrong?
Fixes #600 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-utils/src/version.rs
+++ b/trin-utils/src/version.rs
@@ -3,10 +3,10 @@ pub const TRIN_VERSION: &str = env!("CARGO_PKG_VERSION");
 /// Returns the trin version and git revision.
 pub fn get_trin_version() -> String {
     let git_hash = env!("GIT_HASH");
-    let git_revision_short = if git_hash.is_empty() {
-        "unknown".to_string()
-    } else {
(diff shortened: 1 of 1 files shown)


ogenev (approved):

njgheorghita:
Merging this so we can fix the failing run in `portal-hive` but this doesn't solve the fact that our docker image doesn't support the git commit inside enr feature - tracking in #603

PR #604, opened 2023-03-21
Build docker image with git hash env var

njgheorghita:
### What was wrong?
`git` is not available from within the docker build context. Which means that all of our docker-based nodes (eg. hive & testnet) are built w/o the git hash in their enr (`trin v0.1.0-unknown`). The latest commit hash must be passed in manually using the `--build-arg` flag. 

### How was it fixed?
Added logic to check for the docker build argument if git is not available to produce the git hash.

[This docker image](https://hub.docker.com/layers/portalnetwork/trin/latest/images/sha256-13fbae2c1ebedfebc8d989d5e7f5ff338bb01377aff7d8e4fbe60f64048a0295?context=explore) contains the build of this pr. If you pull it down locally and run it, you can see the git hash in the displayed version when you launch trin.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,11 @@ FROM rust:1.66.1 AS builder
 RUN USER=root cargo new --bin trin
 WORKDIR /trin
 
+# Docker build command *SHOULD* include --build-arg GIT_HASH=...
+# eg. --build-arg GIT_HASH=$(git rev-parse HEAD)
+ARG GIT_HASH=unknown
+ENV GIT_HASH=$GIT_HASH
+
 RUN apt-get update && apt-get install clang -y
 
 # copy over manifests and source to build image
--- a/trin-utils/build.rs
+++ b/trin-utils/build.rs
@@ -1,11 +1,29 @@
+use std::env;
 use std::process::Command;
+
 fn main() {
     let output = Command::new("git")
         .args(["rev-parse", "HEAD"])
         .output()
         .expect("Unable to get git hash");
-    let git_hash = String::from_utf8(output.stdout)
-        .expect("git rev-parse output must be a valid utf8 encoding");
+    let output = String::from_utf8(output.stdout)
+        .expect("git rev-parse output must be a valid utf8 encoding")
+        .replace('\n', "");
+    let git_hash = match output.len() {
+        40 => output,
+        _ => {
+            // If the git hash is not 40 characters, then we are probably in a
+            // docker build context, in which case we access the git hash via the
+            // environment variable set in the Docker build command
+            match env::var("GIT_HASH") {
+                Ok(val) => match val.len() {
+                    40 => val,
+                    _ => "".to_string(),
+                },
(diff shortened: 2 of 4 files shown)


carver (approved):
Hah, deploying is never quite as easy as it seems it ought to be...

Dockerfile
```diff
@@ -5,6 +5,11 @@ FROM rust:1.66.1 AS builder
 RUN USER=root cargo new --bin trin
 WORKDIR /trin
 
+# Docker build command *MUST* include --build-arg GIT_HASH=... to set this value
+# eg. --build-arg GIT_HASH=$(git rev-parse HEAD)
+ARG GIT_HASH
```
ogenev, line 10: I think it is better if we set a default value for GIT_HASH:

```sh
ARG GIT_HASH=<default_value>
```

This will use the default value **if** `GIT_HASH` is not provided via `--build-arg`.

I'm not sure what this default value should be but I think we don't want to have this hard requirement to always provide a git hash when building the docker image. It is adding additional overhead over the typical use case for just building the docker image and not caring for the git hash.

njgheorghita: From what I can tell, it's not technically a hard requirement. W/o a default value, the arg is set to an empty string. But, it's not a huge deal to put a default value in there. I updated it to `GIT_HASH=unknown`. I've also updated the `build.rs` script to set the rustc env var `GIT_HASH` to an empty string rather than panicking. Then the actual util function will set the hash or "unknown" accordingly.

trin-utils/build.rs
```diff
@@ -1,11 +1,29 @@
...
+    let git_hash = match output.len() {
+        40 => output,
+        _ => {
+            // If the git hash is not 40 characters, then we are probably in a
+            // docker build context, in which case we access the git hash via the
+            // environment variable set in the Docker build command
+            match env::var("GIT_HASH") {
+                Ok(val) => match val.len() {
+                    40 => val,
+                    _ => "".to_string(),
```
jacobkaufmann, line 21: IIUC we should never encounter this `match` arm, so it could be useful to log some warning here.

jacobkaufmann (approved):

trin-utils/build.rs
njgheorghita: There's a case when `docker build ...` is used *without* passing in a `--build-arg GIT_HASH=$CIRCLE_SHA1` arg. Then, the default `GIT_HASH=unknown` env var will be set in the Dockerfile. Which will lead to this `match` arm, and in which case we want to set the rustc env var to `""` to have some sort of consistent behavior for our `get_trin_version()` util

Also, according to the [docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script) - `The output of the script is hidden from the terminal during normal compilation.` unless you use `"very verbose" with the -vv flag.`. Which I'd expect to rarely happen if ever, so I'm not sure how useful a logged warning is in this case

Issue #530, opened 2023-02-08
Better INFO level logging output


pipermerriam:
This is what  I get today when I run `trin` with INFO level logging.  (I've removed duplicate messages for conciseness)

```
2023-02-08T18:04:04.964773Z  INFO trin: Launching trin config=TrinConfig { networks: ["history"], capacity_kb: 100000, ephemeral: false, json_rpc_url: /tmp/trin-jsonrpc.ipc, pool_size: 5, metrics_enabled: false }
2023-02-08T18:04:04.971829Z  INFO trin_core::socket: Connecting to STUN server to find public network endpoint
2023-02-08T18:04:05.112540Z  INFO trin_core::portalnet::discovery: Starting discv5 enr.encoded=enr:-I24QOfv0-qpAnmtVIWWsRVcuGSH0XqJ8t2VXpmtBl_dzwriAIktNIlTRNLJOz24I_jUXppVc51_tQWp0BW8VJ8jbYYBY4d0IDAuMS4wgmlkgnY0gmlwhEkODAWJc2VjcDI1NmsxoQPtVMsZC2JJRMSgN-egNOgLZ71VebaH5QfCi6FYsgT9L4N1ZHCCIyg enr.decoded=ENR: NodeId: 0x2c2c..5fce, IpV4 Socket: Some(73.14.12.5:9000) IpV6 Socket: None
2023-02-08T18:04:05.113211Z  INFO discv5::service: Discv5 Service started
2023-02-08T18:04:05.113282Z  INFO discv5::service: Ip4
2023-02-08T18:04:05.113610Z  INFO trin_core::portalnet::storage: Setting up RocksDB path=/home/pipermerriam/.local/share/trin/trin_2c2c408f/rocksdb
2023-02-08T18:04:05.222498Z  INFO trin_core::portalnet::storage: Setting up SqliteDB path=/home/pipermerriam/.local/share/trin/trin_2c2c408f/trin.sqlite
2023-02-08T18:04:05.251853Z  INFO trin: Loaded master accumulator from: "src/assets/merge_macc.bin"
[...]

PR #605, opened 2023-03-22
Add simple heartbeat log to history network

carver:
### What was wrong?

Related to #530 

### How was it fixed?

Added heartbeat log that shows radius as a percentage of data covered
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-history/src/lib.rs
+++ b/trin-history/src/lib.rs
@@ -13,6 +13,7 @@ use network::HistoryNetwork;
 use tokio::{
     sync::{mpsc, RwLock},
     task::JoinHandle,
+    time::{interval, Duration},
 };
 use tracing::info;
 use utp_rs::socket::UtpSocket;
@@ -66,6 +67,7 @@ pub async fn initialize_history_network(
         portalnet_config,
         history_event_rx,
     );
+    spawn_history_heartbeat(Arc::clone(&history_network));
     Ok((
         Some(history_handler),
         Some(history_network_task),
@@ -106,4 +108,18 @@ pub fn spawn_history_network(
             .await
             .expect("failed to pause until ctrl-c");
     })
+}
+
+pub fn spawn_history_heartbeat(network: Arc<HistoryNetwork>) {
+    tokio::spawn(async move {
+        let mut heart_interval = interval(Duration::from_millis(30000));
+
+        loop {
+            let radius = network.overlay.store.read().radius();
+            // calculate the percentage of the whole data ring covered by the data radius
+            let coverage_percent = radius.byte(31) as f32 * 100.0 / 255.0;
+            info!("report: radius={coverage_percent:.1}%");
+            heart_interval.tick().await;
+        }
+    });
 }
--- a/trin-core/src/portalnet/overlay.rs
+++ b/trin-core/src/portalnet/overlay.rs
@@ -86,7 +86,7 @@ pub struct OverlayProtocol<TContentKey, TMetric, TValidator, TStore> {
     pub discovery: Arc<Discovery>,
     /// The data store.
     pub store: Arc<RwLock<TStore>>,
-    /// The data radius of the local node.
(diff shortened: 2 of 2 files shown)


carver:
Obviously a long way to go before finishing up 530, but who doesn't love a tiny PR?

trin-history/src/lib.rs
```diff
@@ -113,3 +116,16 @@ pub fn spawn_history_network(
...
 }
+
+pub fn spawn_history_heartbeat(network: Arc<HistoryNetwork>) {
+    tokio::spawn(async move {
+        let mut heart_interval = interval(Duration::from_millis(30000));
+
+        loop {
+            let radius = network.overlay.data_radius();
+            let coverage_percent = *radius / U256::MAX * 100;
+            info!("status: coverage={coverage_percent:.2}%");
```
njgheorghita, line 127: At some point, I'd expect we'd want these heartbeat logs to be their own "level" - so a user can run them at just that level, hiding the rest of the `info!` logs.

For now, it might be worth something slightly more descriptive like `status: history network coverage=..` ... or maybe even `heartbeat: history network coverage=..` - so that we can distinguish the heartbeat logs from regular `info!` logs.

njgheorghita (approved):
:heart:

trin-history/src/lib.rs
njgheorghita: Thinking a bit more about this, just wanted to mention that it's probably worth moving towards a structure where the heartbeat log is in some way connected to our metrics - or at the very least they share the data which they report. Since something like this `coverage_percent` (as well as most of the heartbeat log data I'd imagine) would be nice to have displayed on the metrics dashboard

pipermerriam, line 127: Any thoughts on using structured logging here or whether that's the correct approach?

carver: I didn't find obvious google hits on "rust structured logging." What do you mean?

I imagine whatever it is will be easy to move to going forward, so I won't consider it a blocker for the PR.

carver: I can't believe that "rust structured logging" search doesn't take you straight here: https://docs.rs/log/latest/log/#structured-logging

trin-history/src/lib.rs
```diff
@@ -113,3 +116,16 @@ pub fn spawn_history_network(
             .expect("failed to pause until ctrl-c");
     })
 }
+
+pub fn spawn_history_heartbeat(network: Arc<HistoryNetwork>) {
+    tokio::spawn(async move {
+        let mut heart_interval = interval(Duration::from_millis(30000));
+
+        loop {
+            let radius = network.overlay.data_radius();
```
carver, line 125: Pretty sure this is wrong, and it needs to be:
```suggestion
            let radius = network.overlay.store.read().data_radius();
```

It seems that the radius on the overlay is just the user-supplied thing, and never changes.

carver: There we go, there's a rounding bug right now, which I'd like to fix before merging.

trin-history/src/lib.rs
carver: Ok, so the reason I'm not doing this before merge is that I want to use the float rounding display. That means presumably doing a format inside the structured value, which feels kind of awkward, like:
```rust
info!(radius = format!("{coverage_percent:.1}%"); "report");
```

So if someone comes up with a better idea, I'll include it in the next PR.

trin-history/src/lib.rs
pipermerriam: Maybe the `network.overlay.data_radius()` should be renamed to `hardcoded_data_radius()` or something.

carver: Yup! Tracking that thought (or removing it entirely) in https://github.com/orgs/ethereum/projects/33/views/1?pane=issue&itemId=23912052

Issue #610, opened 2023-03-25
Reduce uTP log verbosity


carver:
After #551 there's some logspam of seemingly normal activity that doesn't belong at `INFO` level. For example:
```
2023-03-25T02:40:33.299468Z  INFO uTP{send=18394 recv=18393}: utp_rs::conn: uTP conn starting...
2023-03-25T02:40:33.473132Z  INFO uTP{send=18394 recv=18393}: utp_rs::conn: uTP conn initiating shutdown...
2023-03-25T02:40:33.473293Z  INFO uTP{send=57369 recv=57368}: utp_rs::conn: uTP conn initiating shutdown...
2023-03-25T02:40:33.473285Z  INFO uTP{send=9503 recv=9502}: utp_rs::conn: uTP conn initiating shutdown...
2023-03-25T02:40:33.473565Z  INFO uTP{send=48483 recv=48482}: utp_rs::conn: uTP conn initiating shutdown...
2023-03-25T02:40:33.996850Z  INFO uTP{send=9503 recv=9502}: utp_rs::conn: uTP conn closing... err=None
2023-03-25T02:40:33.996980Z  INFO utp_rs::socket: uTP conn shutdown cid.send=9503 cid.recv=9502
```

These should probably all be DEBUG level logs.

carver:
Even the warnings are a bit noisy, though those might be actual issues to look into.

As a workaround for the noise, you can launch trin with `RUST_LOG='info,utp_rs=ERROR'`.

PR #611, opened 2023-03-25
bump utp-rs dependency version

jacobkaufmann:
### What was wrong?

#610 

### How was it fixed?

bump `utp-rs` version to [`0.1.0-alpha.3`](https://github.com/jacobkaufmann/utp/pull/42), which includes changes to address verbosity issue above

### To-Do

- [ ] Clean up commit history


carver (approved):
Thanks!

PR #612, opened 2023-03-27
Remove history content key/item `Unknown` variants

### LINKED ISSUE #538: ethportal-api: Add `is_unknown` method to `HistoryContentItem` and `HistoryContentKey`.
As Jacob mentioned:

_it might be worth adding a method to HistoryContentItem like is_unknown given that now the conversion from a byte vector never fails. with such a method, you would be able to short-circuit your logic without doing a match if you require a valid content item. ultimately, I would prefer that we find some way to remove the Unknown variant while still getting the behavior we want from jsonrpsee. The same goes for HistoryContentKey._

### LINKED ISSUE #540: Refactor hacky errors in `trin-core/portalnet/storage`
Errors [here](https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/trin-core/src/portalnet/storage.rs#L355) and [here](https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/trin-core/src/portalnet/storage.rs#L496) are hacks to get around the fact that rusqlite doesn't support returning a custom error type.

perama-v:
## What was wrong?

- Resolves #538
- Fixes #540

## How was it fixed?

- `thiserror` is used to capture errors 
- Problematic bytes are stored in an error variant:
    - Where decoding fails to produce a known content type (item/key), represent the input as an error variant rather than a content variant.
- `From` replace with `TryFrom`:
    - When a content key decode failure was encountered via `From`, the `Unknown` variant was used. Now, `From` is replaced with `TryFrom`.
- RPC responses are wrapped in an Option-like enum for `HistoryContentValue `called `PossibleHistoryContentValue`. This allows the None variant to be represented as `0x`. The serialization and deserializeation of this new enum could be probably be shortened by pointing to the implementations for `HistoryContentValue`.

```rs
// ethportal-api/src/types/content_item.rs

/// An error decoding a portal network content item.
#[derive(Clone, Debug, Error, PartialEq)]
pub enum ContentValueError {
    #[error("unable to decode value SSZ bytes {input} due to {decode_error:?}")]
    DecodeSsz {
        decode_error: ssz::DecodeError,
        input: String,
    },
    #[error("could not determine content type of {bytes} from {network} network")]
    UnknownContent { bytes: String, network: String },
    /// The content value is the "0x" absent content message rather than data.
    ///
    /// This error implies that handling of the "content absent" response was skipped.
    #[error("attempted to deserialize the '0x' absent content message")]
    DeserializeAbsentContent,

    /// The content value is the "0x" absent content message rather than data.
    ///
    /// This error implies that handling of the "content absent" response was skipped.
    #[error("attempted to decode the '0x' absent content message")]
    DecodeAbsentContent,
}
```
Errors have two representations and a test is introduced to check both kinds:
- Debug: `UnknownContent { bytes: "0x010203040506070809", network: "history" }`
- Display: `"could not determine content type of 0x010203040506070809 from history network"`

### co-resolution of rusqlite mixed error handling
Handling of the `TryFrom` code touched some previously unresolved `rusqlite` error workaround where multiple error
kinds were occurring within closure that had to return a rusqlite error. This was fixed by dividing the closure into two and mapping the rusqlite error to a native error
```rs
 #[error("rusqlite error")]
 Rusqlite(#[from] rusqlite::Error),
```

### Automatic implementation of existing error traits

The existing errors (`ContentKeyError`, `ContentItemError`) had manual implementations of `From` and `Display`. These
have been moved to the error enum via the derive syntax that `thiserror` provides.

### Replaced anyhow use in trin-utils with custom error types

The hex utils had opaque anyhow errors, which were changed to specific error kinds for greater transparency upon failure.
This is accompanied by additional unit tests and handling of cases where the `hex_decode()` function could panic.

This could be pulled out as a separate PR. I included it because a) handling no content responses with hex_decode induced a panic that was unhandled due to string slice indexing and b) newly created error variants elsewhere would have been created with opaque internal anyhow that would later need to be undone.

### Treat invalid keys in requests as error

Peertest was responding with `0x` when passed an invalid key (odd length, too short, etc). Now it responds with invalid parameter jsonrpc response and includes a message with an explanation of the reason.
```console
{"code": Number(-32602), "message": String("unable to decode key SSZ bytes 0x0055b11b918355b1ef9c5db810302ebad0bf2544255b530cdce9 due to InvalidByteLength { len: 25, expected: 32 }")}
```

### Check content tracing of absent-content response

Previously there was no peertest to check the peers involved when a particular content value was absent. A test is
added, with a check that the ENR list has at least one peer (the specific peer ENR is not asserted as the content is
not posted to a particular peer.)

## To-Do

- [x] Clean up commit history
- [x] Resolve remaining issues related to replacing `From` with `TryFrom` for the `HistoryContentKey`
- [x] Testing

### DIFF
--- a/ethportal-peertest/src/jsonrpc.rs
+++ b/ethportal-peertest/src/jsonrpc.rs
@@ -8,6 +8,7 @@ use hyper::{self, Body, Client, Method, Request};
 use serde_json::{self, json, Value};
 use ssz::Encode;
 use tracing::{error, info};
+use trin_types::constants::CONTENT_ABSENT;
 
 use crate::{cli::PeertestConfig, Peertest};
 use trin_types::distance::Distance;
@@ -203,7 +204,7 @@ fn validate_discv5_node_info(val: &Value, _peertest: &Peertest) {
 fn validate_discv5_routing_table_info(val: &Value, _peertest: &Peertest) {
     let local_key = val.get("localNodeId").unwrap();
     assert!(local_key.is_string());
-    assert!(local_key.as_str().unwrap().contains("0x"));
+    assert!(local_key.as_str().unwrap().starts_with("0x"));
     assert!(val.get("buckets").unwrap().is_array());
 }
 
@@ -254,7 +255,7 @@ pub fn validate_portal_offer(result: AcceptInfo, _peertest: &Peertest) {
 }
 
 pub fn validate_portal_local_content(result: &Value, _peertest: &Peertest) {
-    assert_eq!(result.as_str().unwrap(), "0x0");
+    assert_eq!(result.as_str().unwrap(), CONTENT_ABSENT);
 }
 
 #[cfg(unix)]
--- a/ethportal-peertest/src/scenarios/offer_accept.rs
+++ b/ethportal-peertest/src/scenarios/offer_accept.rs
@@ -1,12 +1,13 @@
-use std::{str::FromStr, thread, time};
+use std::str::FromStr;
 
 use serde_json::json;
 use tracing::{error, info};
 
-use ethportal_api::jsonrpsee::async_client::Client;
-use ethportal_api::HistoryContentKey;
-use ethportal_api::{HistoryContentValue, HistoryNetworkApiClient};
-use trin_types::enr::Enr;
+use ethportal_api::{
+    jsonrpsee::async_client::Client, HistoryContentKey, HistoryContentValue,
+    HistoryNetworkApiClient,
+};
+use trin_types::{content_value::PossibleHistoryContentValue, enr::Enr};
 
 use crate::{
     jsonrpc::{validate_portal_offer, HISTORY_CONTENT_KEY, HISTORY_CONTENT_VALUE},
@@ -48,7 +49,11 @@ pub async fn test_unpopulated_offer(peertest_config: PeertestConfig, peertest: &
     validate_portal_offer(result, peertest);
 
     // Check if the stored content value in bootnode's DB matches the offered
-    let received_content_value = wait_for_content(ipc_client, content_key).await;
+    let response = wait_for_content(ipc_client, content_key).await;
+    let received_content_value = match response {
+        PossibleHistoryContentValue::ContentPresent(c) => c,
+        PossibleHistoryContentValue::ContentAbsent => panic!("Expected content to be found"),
+    };
     assert_eq!(
         content_value, received_content_value,
         "The received content {received_content_value:?}, must match the expected {content_value:?}",
@@ -84,7 +89,11 @@ pub async fn test_populated_offer(peertest_config: PeertestConfig, peertest: &Pe
     validate_portal_offer(result, peertest);
 
     // Check if the stored content value in bootnode's DB matches the offered
-    let received_content_value = wait_for_content(ipc_client, content_key).await;
+    let response = wait_for_content(ipc_client, content_key).await;
+    let received_content_value = match response {
+        PossibleHistoryContentValue::ContentPresent(c) => c,
+        PossibleHistoryContentValue::ContentAbsent => panic!("Expected content to be found"),
+    };
     assert_eq!(
         content_value, received_content_value,
         "The received content {received_content_value:?}, must match the expected {content_value:?}",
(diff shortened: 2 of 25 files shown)


ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -84,31 +83,9 @@ pub async fn test_populated_offer(peertest_config: PeertestConfig, peertest: &Pe
...
-    let received_content_value = wait_for_content(ipc_client, content_key).await;
+    let received_content_value = ipc_client.local_content(content_key).await.unwrap();
     assert_eq!(
         content_value, received_content_value,
         "The received content {received_content_value:?}, must match the expected {content_value:?}",
     );
 }
-
-/// Wait for the content to be transferred
-async fn wait_for_content(
```
perama-v, line 95: Not sure about the status of this function. Is it being used? I was not clear on how to retain the function in the context of removing the `::Unknown` variant

src/bin/purge_db.rs
```diff
@@ -97,10 +97,6 @@ fn is_content_valid(content_key: &HistoryContentKey, value: &[u8]) -> bool {
         HistoryContentKey::BlockBody(_) => BlockBody::from_ssz_bytes(value).is_ok(),
         HistoryContentKey::BlockReceipts(_) => Receipts::from_ssz_bytes(value).is_ok(),
         HistoryContentKey::EpochAccumulator(_) => EpochAccumulator::from_ssz_bytes(value).is_ok(),
-        HistoryContentKey::Unknown(_) => {
```
ogenev, line 100: I think this change will break the script when on line 72 if an invalid content value is present.

trin-types/src/content_value.rs
```diff
@@ -26,8 +27,93 @@ pub enum HistoryContentValue {
     BlockBody(BlockBody),
     Receipts(Receipts),
     EpochAccumulator(EpochAccumulator),
-    /// A placeholder for data that could not be interpreted as any of the valid content types.
-    Unknown(String),
+}
+
+/// A content response from the RPC server.
+///
+/// This type allows the RPC response to be non-error, byt
```
ogenev, line 34: nit pick:
```suggestion
/// This type allows the RPC response to be non-error,
```

rpc/src/history.rs
```diff
@@ -192,18 +196,20 @@ impl HistoryNetworkApiServer for HistoryNetworkApi {
         Ok(result)
     }
 
-    /// Get a content from the local database
+    /// Get a content from the local database.
+    ///
+    /// If content is absent, Ok(None) is returned.
```
ogenev, line 201: Confused about this comment, if `PossibleHistoryContentValue::ContentAbsent` is returned, this is serialized to `0x` instead of ` None`?

portalnet/src/storage.rs
```diff
@@ -482,17 +460,16 @@ impl PortalStorage {
...
             .query_map([id], |row| {
                 let row: String = row.get(0)?;
-                // value is stored without 0x prefix, so we must add it
-                let row = format!("0x{}", row);
-                let bytes: Vec<u8> = hex_decode(&row).map_err(|err|
-                    // TODO: This is a hack to get around the fact that rusqlite doesn't
-                    // support returning a custom error type. We should fix this.
-                    rusqlite::Error::InvalidParameterName(err.to_string()))?;
-                Ok(HistoryContentKey::from(bytes))
+                Ok(row)
```
ogenev, line 466: :tada:

trin-types/src/jsonrpc/request.rs
```diff
@@ -8,6 +8,9 @@ use crate::jsonrpc::params::Params;
 
 type Responder<T, E> = mpsc::UnboundedSender<Result<T, E>>;
 
+/// Portal network defines "content absent" response as "0x".
+pub const CONTENT_ABSENT: &str = "0x";
```
ogenev, line 12: How do we feel about moving this constant in a separate file `constants.rs` for example in `trin-types` crate?

ogenev (approved):
Looks great to me :+1:, :heart:  , left some minor comments to address/discuss before merging.

ethportal-peertest/src/jsonrpc.rs
```diff
@@ -203,7 +204,7 @@ fn validate_discv5_node_info(val: &Value, _peertest: &Peertest) {
 fn validate_discv5_routing_table_info(val: &Value, _peertest: &Peertest) {
     let local_key = val.get("localNodeId").unwrap();
     assert!(local_key.is_string());
-    assert!(local_key.as_str().unwrap().contains("0x"));
+    assert!(local_key.as_str().unwrap().contains(CONTENT_ABSENT));
```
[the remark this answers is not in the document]
perama-v: Yes this is an error on my part

rpc/src/history.rs
perama-v: Good find. This was a doc comment I made but failed to update as more changes were made.

trin-types/src/jsonrpc/request.rs
perama-v: I thought about it, but did not see a benefit. It is a value inherent to the JSON-RPC, so thought it appropriate here. 
I guess it is good convention though, so I'll move it.

perama-v:
> Looks great to me 👍 ❤️ , left some minor comments to address/discuss before merging.

Thanks. Changes made and squashed

PR #614, opened 2023-03-28
Storage metrics for log report

carver:
### What was wrong?

Progress on #530 

Want to report more storage usage metrics in the logs

Bonus issue: There was going to be some duplication of effort, to build the metrics for logs. So we combine the collection/calculation to one spot.

### How was it fixed?

Changes are small, but there are several of them. I recommend reviewing commit-by-commit.

1. Always track the database statistics, whether or not we report them online (since we always report them to log now)
2. Combine all storage metrics into a single loggable string
3. Internally track whenever the radius changes (both, setting `storage.radius` and updating `storage.metrics.report_radius()`)
4. Correctly generate the starting radius, when reusing an existing database
5. Bonus fix: the first log was getting double-reported for unknown reasons, this went away in local testing.

Example report:
```
2023-03-28T01:17:43.692962Z  INFO trin_history: storage-report: radius=22.7% content=9.8/10kb disk=147.1kb
```

Low urgency follow-up work: convert large storage values to MB or GB

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-core/src/portalnet/storage.rs
+++ b/trin-core/src/portalnet/storage.rs
@@ -10,13 +10,13 @@ use discv5::enr::NodeId;
 use ethportal_api::types::portal::PaginateLocalContentInfo;
 use prometheus_exporter::{
     self,
-    prometheus::{register_gauge, Gauge},
+    prometheus::{opts, register_gauge, register_gauge_with_registry, Gauge, Registry},
 };
 use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 use rocksdb::{Options, DB};
 use rusqlite::params;
-use tracing::{debug, info};
+use tracing::{debug, error, info};
 
 use crate::{portalnet::types::messages::ProtocolId, utils::db::get_data_dir};
 use ethportal_api::types::content_key::{HistoryContentKey, OverlayContentKey};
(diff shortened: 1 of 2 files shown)


carver:
Failure is:
```
test test::test_launches ... FAILED

failures:

---- test::test_launches stdout ----
thread 'test::test_launches' panicked at 'called `Result::unwrap()` on an `Err` value: AlreadyReg', trin-core/src/portalnet/storage.rs:741:10


failures:
    test::test_launches

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.20s

error: test failed, to rerun pass `-p trin --test self_peertest`
```

I believe this is the first test that runs multiple instances with metrics turned on. I think registering the same metric twice in the same process is causing a duplicate metrics error.

trin-core/src/portalnet/storage.rs
```diff
@@ -280,54 +279,63 @@ impl PortalStorage {
...
+                    storage.set_radius(Distance::ZERO);
                 }
             }
             // Only prunes data when at capacity. (eg. user changed it via kb flag)
-            _ => storage.prune_db()?,
+            _ => match storage.prune_db()? {
+                0 => {
+                    // No items were pruned, so the radius was never calculated.
+                    // Calculate current radius now, rather than waiting for the next overfill.
+                    match storage.find_farthest_content_id()? {
```
njgheorghita, line 301: `if let Some(farthest) == storage.find_farthest_content_id()? {..}`

trin-core/src/portalnet/storage.rs
```diff
@@ -280,54 +279,63 @@ impl PortalStorage {
...
             0 => {
                 // Default radius is left in place, unless user selected 0kb capacity
                 if storage.storage_capacity_in_bytes == 0 {
-                    storage.radius = Distance::ZERO;
+                    storage.set_radius(Distance::ZERO);
                 }
             }
             // Only prunes data when at capacity. (eg. user changed it via kb flag)
-            _ => storage.prune_db()?,
+            _ => match storage.prune_db()? {
```
njgheorghita, line 297: `_ => if storage.prune_db()? == 0 {.. }`

trin-core/src/portalnet/storage.rs
```diff
@@ -708,34 +717,68 @@ impl PortalStorage {
...
+            Ok(gauge) => (gauge, Registry::default()),
+            Err(_) => {
+                error!("Failed to register prometheus gauge with default registry, creating new");
+                let custom_registry = Registry::new_custom(None, None)
+                    .expect("Prometheus docs don't explain when it might fail to create a custom registry, so... hopefully never");
+                let gauge = register_gauge_with_registry!(
+                    content_storage_usage_kb_options,
+                    custom_registry
+                )
+                .unwrap();
```
njgheorghita, line 753: Hmmm, this `unwrap()` shouldn't be possible without the linter complaining... Though based off my work in #616 - it seems like the linting process isn't hitting these files... But, either way this unwrap should be handled

trin-core/src/portalnet/storage.rs
```diff
@@ -708,34 +717,68 @@ impl PortalStorage {
...
+                (gauge, custom_registry)
+            }
+        };
+
+        let total_storage_usage_kb = register_gauge_with_registry!(
+            format!("trin_total_storage_usage_kb_{:?}", protocol),
```
njgheorghita, line 759: for this and the next 2 `format!` statements ... `format!("trin_...{protocol:?}")`

trin-history/src/lib.rs
```diff
@@ -113,12 +113,11 @@ pub fn spawn_history_network(
...
+        heart_interval.tick().await;
 
         loop {
-            let radius = network.overlay.store.read().radius();
-            // calculate the percentage of the whole data ring covered by the data radius
-            let coverage_percent = radius.byte(31) as f32 * 100.0 / 255.0;
-            info!("report: radius={coverage_percent:.1}%");
+            let storage_log = network.overlay.store.read().get_summary_info();
+            info!("storage-report: {storage_log}");
             heart_interval.tick().await;
```
njgheorghita, line 121: *nit* if you want to get rid of an extra line of code here, it seems like you could just move `heart_interval.tick().await;` to the beginning of the `loop { .. }`

njgheorghita (approved):
It's definitely a move in the right direction. But, I think our overall metrics strategy could use some definition. eg. How does this interact w/ `OverlayMetrics`? Should they be separate? Share a common trait? Do all metrics belong in an isolated workspace? Some of these questions will be easier to answer once all the dust settles from the recent re-org. 

We have [this clause](https://github.com/ethereum/trin/blob/master/trin-core/src/portalnet/metrics.rs#L8-L12) which is a very loose definition. But as our metrics evolves, it would be nice to make the definition a bit more robust.

trin-history/src/lib.rs
carver: Oh yeah, it evolved in a weird way, because I didn't want to wait 30s to get the first log line, but there is some weird thing (bug?) where the first tick expires roughly immediately (at least on my runs). 👍🏻 for simplifying, though I'll add a comment so I may not kill a line.

pipermerriam:
This is a lovely improvement.

PR #613, opened 2023-03-27
Add receipts & body round trip tests

njgheorghita:
### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [ ] Clean up commit history

Issue #606, opened 2023-03-22
Long-term home for core Portal Network types


njgheorghita:
@ogenev and have been discussing this for the past days, and we think it's time to open it up with the group to get feedback and decide on a path forward.

The core question is where will we define core portal network types that we specifically want to expose over jsonrpc. 
1. Inside `ethportal-api`
2. Inside `trin-types`. 

For example. Let's take `HeaderWithProof`. It's required by `ethportal-api` as well as `trin-history`, `trin-validation` and `trin-bridge` as well as 3rd party libraries using `ethportal-api`.

(Currently this type is defined twice, once inside `ethportal-api` and once inside `trin-types`. but @ogenev and I are in agreement that we should move towards a single definition.)

The question is essentially, which workspace should act as the "root" level package where we store these type definitions.
`ethportal-api` -> `trin-types` -> `trin-...`
or
`trin-types` -> `ethportal-api` -> `trin-...`

We'll leave some thoughts below for everyone to ponder and then we can discuss it tomorrow during sync

njgheorghita:
IMO, we should follow the convention set in the `README` for `trin-types`.
`Library for shared types across trin workspace members.`

If we start defining core types inside `ethportal-api`. Then it will become a de-facto "secondary" `trin-types`, where we will end up defining many of our core types. It seems much more organized to simply define all shared types in one place. 

One other way to frame this question is...
My opinion:  the `trin` in `trin-types` refers to all kinds of types used inside this project directory.
Oggie's opinion: the `trin` in `trin-types` refers to shared types specifically used by the `trin` client. The core types we're debating are better categorized as "portal" types - since they're also used by external users of `ethportal-api` (eg glados).

Tbf there is more nuance in this decision, including versioning dependencies, the `reth` situation & how we plan to resolve that. But I'll leave that for discussion tm.

carver:
Preliminary proposal from the Thursday call, waiting on @ogenev 's thoughts on Monday: define all common types in trin-types. Don't *implement* any types in ethportal-api, only re-export them from their trin-types implementation.

perama-v:
> Preliminary proposal from the Thursday call, waiting on @ogenev 's thoughts on Monday: define all common types in trin-types. Don't _implement_ any types in ethportal-api, only re-export them from their trin-types implementation.

This sounds good. Defining all in one place is good and `glados` is flexible to such a change

carver:
[...]

PR #615, opened 2023-03-28
Replace reth-based types with trin-native types

njgheorghita:
### What was wrong?

This pr is not a foregone conclusion, but a proposal to remove our dependency on `reth` - at least until it's stable.

In my opinion, the reth-based types experiment was not successful and has wasted a significant amount of time debugging these issues. Personally, I do not see any way in which the benefits outweigh the negatives. (in fact, I'm not sure I see many benefits at all, since I prefer direct control over these core types rather than outsourcing them).

#### 3 direct, technical blockers from using reth
- lack of support for pre-byzantium receipts
- incompatible versioning - requiring us to pin to a specific git revision
- rlp encoding / decoding - #613

#### Other qualms with using reth
- it's officially "unstable" / not in production, yes so are we, but i'd rather build on top of stable types rather than unstable types
- it has no documentation, requires diving into the codebase to figure out what's going on
- introduces new apis for rlp encoding/decoding (`reth_rlp`), which conflicts with existing apis (`rlp`) in our codebase
- introduces new primitives (`reth_primitives`) which conflicts with existing primitives in our codebase (`ethereum_types`)

#### In this case. I prefer to manage our own types.
- we don't have to worry about getting bugfixes / pushes merged into the reth repo
- we have control over our types, this one is quite important imo, in terms of tradeoffs, I'm more than happy to deal with extra lines of code if that means we have direct control over these very core types to our project
- removes unstable dependencies, don't have to juggle multiple apis, conflicting dependency libraries
- our types themselves are not that "heavy" in terms of raw code. it's 3, moderately sized files

I do like the concept of the `ContentItem` trait. Though, I feel like `ContentValue` might be a better name - since that's the verbiage we've been using up to this point around the codebase. Also, the trait fn names of `encode` and `decode` are not ideal, imo. We have so many encode / decode traits floating around this codebase, and it's confusing. (but this can be discussed / addressed in a later pr).

This pr also addresses the dependency chain we've been discussing recently, by implementing all types inside `trin-types` and then re-exporting them from `ethportal-api`

### How was it fixed?
- removed `reth-primitives` types
- moved `ContentItem` & `ContentKey` -> `trin-types`
- re-export relevant types from within `ethportal-api` so they remain available to external library users

### To-Do

- [x] Clean up commit history


carver:
Yup, this makes a strong case for switching back to internal types.

For thoroughness, one argument missing here is that we might want to use reth for EVM execution at some point. (though I don't think that's enough on its own to outweigh the points made so far)

I'd like to hear one last time from @jacobkaufmann on whether we are missing any upsides to using reth in particular. I'd love to have the choice on direction made by tomorrow, but no later than the Thursday meeting.

ethportal-api/src/lib.rs
```diff
@@ -10,16 +10,18 @@ mod web3;
...
-pub use web3::{Web3ApiClient, Web3ApiServer};
+pub use trin_types::execution::accumulator::{EpochAccumulator, HeaderRecord};
+pub use trin_types::execution::block_body::{BlockBody, EncodableHeaderList};
+pub use trin_types::execution::header::{Header, HeaderWithProof};
+pub use trin_types::execution::receipts::{Receipt, Receipts};
 
 // Re-exports jsonrpsee crate
 pub use jsonrpsee;
-
-pub use reth_primitives::{Header, Receipt, TransactionSigned};
```
ogenev, line 25: is it a concern if we don't export`TransactionSigned` anymore?

ogenev (approved):
I agree that `reth` dependency at this stage brings us only headaches :+1: .

trin-types/src/content_item.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::execution::header::HeaderWithProof;
+use crate::execution::receipts::Receipts;
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use ssz::{Decode, Encode};
+use trin_utils::bytes::{hex_decode, hex_encode};
+
+/// An encodable portal network content item.
+pub trait ContentItem: Sized {
+    /// Encodes the content item, appending the encoded bytes to `buf`.
+    fn encode(&self) -> Vec<u8>;
```
carver, line 12: docstring is out of date, there is no `buf` here.

ethportal-api/src/lib.rs
njgheorghita: Ahh, ya good catch. I skipped that one since it was a type that's specific to `reth_primitives`. But, I've gone ahead and re-exported our implementation of `Transaction`. 

I've also checked w/ glados, and they aren't reliant upon `TransactionSigned`. However, they will need to change some of their import strings when they update the version dependency since most of their `OverlayContentKey` imports are using a deeply nested statement rather than the shallow statement. 
eg. `use ethportal_api::types::content_key::OverlayContentKey;` -> `use ethportal_api::OverlayContentKey`.

Also, it seems to me like we'll certainly want to release `trin-types` as a standalone crate sometime in the near future. Since it's an enum, our `Transaction` type alone isn't necessarily that useful without the `LegacyTransaction` / `Eip1559Transaction` / `AccessListTransaction` types.

trin-types/src/content_item.rs
```diff
@@ -0,0 +1,158 @@
...
+            Self::BlockHeaderWithProof(item) => item.as_ssz_bytes(),
+            Self::BlockBody(item) => item.as_ssz_bytes(),
+            Self::Receipts(item) => item.as_ssz_bytes(),
+            Self::EpochAccumulator(item) => item.as_ssz_bytes(),
+            Self::Unknown(item) => {
+                if item.is_empty() {
+                    return serializer.serialize_str("0x0");
+                }
+                item.clone().into_bytes()
+            }
```
carver, line 80: So yeah, I'd be curious what goes wrong if we return an `Err` here, instead? I'm assuming that introduces challenges somewhere that I can't see by reading this code block.

Even if it's not changed during this merge, a code comment and some info in the tracking issue about what happens when attempting this change would be valuable.

ethportal-api/src/lib.rs
carver: > Also, it seems to me like we'll certainly want to release `trin-types` as a standalone crate sometime in the near future. Since it's an enum, our `Transaction` type alone isn't necessarily that useful without the `LegacyTransaction` / `Eip1559Transaction` / `AccessListTransaction` types.

Hm, this seems to go against the path I thought we were on, that `trin-types` is for internal usage, and `ethportal-api` is for other projects to use.

carver (commented):
I'm waiting to give a final 👍🏻 to give @jacobkaufmann a chance to convince me otherwise by Thursday morning. :)

At first glance, I agree that `ContentValue` is better than `ContentItem` because it's more consistent with how it's used in specs and elsewhere. Without any context besides coming from Python, I would expect an item to be a `(key, value)` pair.

ethportal-api/src/lib.rs
njgheorghita: iiuc, the idea behind including these types is that a 3rd party consumer of `ethportal-api` might want access to our core types - eg `Header` / `Transaction`. I've updated this pr to only re-export the content key & content item types. The reason being that these are the only types that are currently consumed by `glados`. 

With regards to the core types - whether we re-export those or force libraries to also depend upon `trin-types` -
 let's continue the conversation in #606

trin-types/src/content_item.rs
njgheorghita: I believe this is handling the case of a failed attempt to lookup a content key in local storage. In which case we're supposed to return `0x0`. But, iiuc, this will be handled in #612

PR #617, opened 2023-03-30
Panicking during: trin --help

carver:
### What was wrong?

trin was panicking when being called with `--help` or `--version`. This was introduced by #578 

### How was it fixed?

Add a test that `-h` doesn't panic. Return the error instead of panicking. Then, when parsing args for CLI, use `Clap::Error`'s built-in `.exit()`, which prints the appropriate message and exits immediately.

### To-Do

- [ ] Clean up commit history
- [x] Add GFI to switch all the cli arg panics to custom `clap:Error` messages #618

### DIFF
--- a/trin-types/src/cli.rs
+++ b/trin-types/src/cli.rs
@@ -205,14 +205,14 @@ impl Default for TrinConfig {
 
 impl TrinConfig {
     pub fn from_cli() -> Self {
-        Self::new_from(env::args_os()).expect("Could not parse trin arguments")
+        Self::new_from(env::args_os()).unwrap_or_else(|e| e.exit())
     }
     pub fn new_from<I, T>(args: I) -> Result<Self, clap::Error>
     where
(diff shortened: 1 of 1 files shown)


trin-types/src/cli.rs
```diff
@@ -205,14 +205,14 @@ impl Default for TrinConfig {
 
 impl TrinConfig {
     pub fn from_cli() -> Self {
-        Self::new_from(env::args_os()).expect("Could not parse trin arguments")
+        Self::new_from(env::args_os()).unwrap_or_else(|e| e.exit())
```
njgheorghita, line 208: Ahh, nice! I think this is the magic. We've had some difficulty with these tests over the past. The reason for the explicit `panic!`s was due to our use of `should_panic(...)` in the tests. But, if we can use `clap`s native errors, I'm all for it.

njgheorghita (approved):
:shipit:

PR #620, opened 2023-03-30
Pass real radius during ping/pong

### LINKED ISSUE #619: Bug: Reported radius is always the user-configured value
Remove duplicate `data_radius` in overlay and overlay_service, pass through reads to the storage instead.

This should reduce the number of times that peers send us data outside our radius (which I've noticed happening a lot when setting the storage value low).

carver:
### What was wrong?

Fix #619 

### How was it fixed?

Rremove any duplicate values of `data_radius` and pass all reads through to storage. Then pings/pongs should use the real value from our data store.

### To-Do

- [x] Clean up commit history


carver:
Someday, we'll have to deal with different radii per network, but for today, at least we'll stop sending the MAX radius in all PING/PONG messages.

PR #625, opened 2023-03-31
Increase radius when storage limit increases

### LINKED ISSUE #624: Increase the radius at launch, when the storage limit increases
Before #614, the radius was always resetting to 100% at the beginning of every run. This felt wrong when we had already identified the radius on previous runs with the same storage limit, so #614 identifies the radius based on the existing storage.

This works fine when the storage limit stays the same between runs, but when the storage increases on a new run, it means the radius stays too small. The node will sub-optimally avoid storing things that it ought to be interested in storing.

Before and after #614, trin correctly handles when you decrease the storage limit: it will prune down the database and reduce the radius as needed.

Some options:
1. Keep the previous storage limit in the database, to detect if there was a change on launch and adjust the radius
2. Adjust the radius as a percentage increase based on the amount used vs the limit
3. ?

I think # 2 is the best option here

carver:
### What was wrong?

Fix #624

### How was it fixed?

Detect the storage usage on launch, and if it's significantly under the limit (2x or more), then increase the radius.

Implement `saturating_mul` on distance, plus some tests

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -285,6 +285,13 @@ impl PortalStorage {
         // Set the metrics to the default radius, to start
         storage.metrics.report_radius(storage.radius);
 
+        // Report total storage used by network content.
+        let network_content_storage_usage =
+            storage.get_total_storage_usage_in_bytes_from_network()?;
+        storage
+            .metrics
+            .report_content_data_storage(network_content_storage_usage as f64 / 1000.0);
+
         // Check whether we already have data, and use it to set radius
         match storage.total_entry_count()? {
             0 => {
@@ -299,7 +306,16 @@ impl PortalStorage {
                     // No items were pruned, so the radius was never calculated.
                     // Calculate current radius now, rather than waiting for the next overfill.
                     if let Some(farthest) = storage.find_farthest_content_id()? {
-                        storage.set_radius(storage.distance_to_content_id(&farthest));
+                        // Set the radius to attempt to fill the unused storage capacity
+                        // For example, if storage is half-full, then the radius should be 2x the
+                        // distance to the furthest currently-stored item.
+                        let radius_multiple = storage.storage_capacity_in_bytes as f64
+                            / network_content_storage_usage as f64;
+                        let current_span = storage.distance_to_content_id(&farthest);
+                        // For simplicity, convert to integer multipe, which means radius won't
+                        // bump up unless it needs to go up by 2x.
+                        let target_radius = current_span.saturating_mul(radius_multiple as usize);
+                        storage.set_radius(target_radius);
                     }
                 }
             }
@@ -315,13 +331,6 @@ impl PortalStorage {
         storage
             .metrics
             .report_total_storage_usage(total_storage_usage as f64 / 1000.0);
-
-        // Report total storage used by network content.
-        let network_content_storage_usage =
-            storage.get_total_storage_usage_in_bytes_from_network()?;
-        storage
-            .metrics
-            .report_content_data_storage(network_content_storage_usage as f64 / 1000.0);
 
         Ok(storage)
     }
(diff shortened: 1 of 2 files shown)


portalnet/src/storage.rs
```diff
@@ -299,7 +306,16 @@ impl PortalStorage {
...
                     // Calculate current radius now, rather than waiting for the next overfill.
                     if let Some(farthest) = storage.find_farthest_content_id()? {
-                        storage.set_radius(storage.distance_to_content_id(&farthest));
+                        // Set the radius to attempt to fill the unused storage capacity
+                        // For example, if storage is half-full, then the radius should be 2x the
+                        // distance to the furthest currently-stored item.
+                        let radius_multiple = storage.storage_capacity_in_bytes as f64
+                            / network_content_storage_usage as f64;
+                        let current_span = storage.distance_to_content_id(&farthest);
+                        // For simplicity, convert to integer multipe, which means radius won't
```
jacobkaufmann, line 315:
```suggestion
                        // For simplicity, convert to integer multiple, which means radius won't
```

portalnet/src/storage.rs
```diff
@@ -285,6 +285,13 @@ impl PortalStorage {
         // Set the metrics to the default radius, to start
         storage.metrics.report_radius(storage.radius);
 
+        // Report total storage used by network content.
+        let network_content_storage_usage =
+            storage.get_total_storage_usage_in_bytes_from_network()?;
+        storage
+            .metrics
+            .report_content_data_storage(network_content_storage_usage as f64 / 1000.0);
```
jacobkaufmann, line 293: unrelated, but it would add clarity for the `report_` methods to have a unit suffix (e.g. kb) where the expected unit is not obvious

jacobkaufmann (approved):
nice! this definitely has some advantages over the naive/simple logic to reset the radius to the maximum if there is available capacity. the maximum radius solution would create additional churn as the node converges toward the appropriate radius. however, the advantage of the maximum radius solution is that you would never under-utilize your capacity. here, there might be some pathological situations where the scaling of the storage capacity does not correspond with an equal scaling of the radius, but the existence and severity of those situations are not immediately obvious.

portalnet/src/storage.rs
carver: Maybe this can go into the PR to switch to MB in #634 -- it seems silly to add kb now just for @mrferris to churn it to MB in a day or two :)

carver:
> here, there might be some pathological situations where the scaling of the storage capacity does not correspond with an equal scaling of the radius, but the existence and severity of those situations are not immediately obvious.

Yeah! I thought about this a little bit. Here is where I got... There are two possible issues to contend with: our estimate of the ideal radius is too large, or too small. Obviously, being an estimate, it will have small-medium variance from the target. But it's reasonable to suppose that there's some variance that we think is too big.

If the estimate of the radius is too large, that's a problem that will fix itself quickly by eventually accepting content until full and then trimming down. No issue there.

The real hairy situation is if the estimate of the radius is too small. Let's say that at the beginning of a client's life, it accepts a single piece of data and then shuts down. That content happens to be shockingly close to the node ID. Like 0.0001% of the radius away (Let's roughly call that a 1 in a million shot). We'll say that we Now, on reboot. Maybe we chose a data size limit that is 1000x bigger than the content already stored. That still only bumps the radius to 0.1%. While storing one piece of content, and at 1/1000th data capacity.

It's unlikely, but probably bad enough that we want to avoid it. Some options, varying levels of hackiness:
- we will probably get pushed data outside our radius anyway, so we could just aggressively store data that is outside our radius, when the database is low
- we could choose to only scale the database this way if we are at least n% full (maybe half?)
- we could go with Piper's suggestion in #624 

I wish I'd done these out of order, because I think I'm going to say that his solution is cleanest, after I rejected it at first look. :)

carver:
Yeah, I think #647 is the best option. Coding in progress, to be resumed tomorrow...

PR #626, opened 2023-04-03
Remove unused cargo dependencies

ogenev:
### What was wrong?
Unused dependencies in some workspace crates.

### How was it fixed?
Ran [cargo-udeps](https://github.com/est31/cargo-udeps) to remove unused dependencies.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -225,17 +225,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-recursion"
-version = "1.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


njgheorghita (approved):
🥳 is there any way to leverage `cargo-udeps` in our ci pipeline?

ogenev:
> partying_face is there any way to leverage `cargo-udeps` in our ci pipeline?

It requires rust nightly, not sure if we want to add a nightly pipeline only for `cargo udeps`, shall we?

njgheorghita:
@ogenev Nah, i don't think it's worth it for now. periodically running `cargo udeps` locally seems sufficient

PR #629, opened 2023-04-03
bump utp-rs to 0.1.0-alpha.4

jacobkaufmann:
### What was wrong?

* incorrect uTP ACK num initialization upon SYN-ACK
* unexpected sequence number in uTP `STATE` packets

### How was it fixed?

bump `utp-rs` to `0.1.0-alpha.4` to include [fix](https://github.com/jacobkaufmann/utp/pull/43)

### To-Do

- [ ] Clean up commit history


jacobkaufmann:
it would be nice if we could run the portal hive test suite before merge

ogenev (approved):
Tested with `portal-hive` locally, this fixes the recent uTP issues with Fluffy :+1:

PR #633, opened 2023-04-03
Show number of storage items in heartbeat log

### LINKED ISSUE #630: Log how many content items we are storing, in heartbeat

carver:
### What was wrong?

Fix #630

### To-Do
- [x] Clean up commit history

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -10,7 +10,10 @@ use discv5::enr::NodeId;
 use ethportal_api::types::portal::PaginateLocalContentInfo;
 use prometheus_exporter::{
     self,
-    prometheus::{opts, register_gauge, register_gauge_with_registry, Gauge, Registry},
+    prometheus::{
+        opts, register_gauge, register_gauge_with_registry, register_int_gauge_with_registry,
+        Gauge, IntGauge, Registry,
+    },
 };
 use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
@@ -270,7 +273,9 @@ impl PortalStorage {
                 }
             }
             // Only prunes data when at capacity. (eg. user changed it via kb flag)
-            _ => {
+            entry_count => {
+                storage.metrics.report_entry_count(entry_count);
+
                 if storage.prune_db()? == 0 {
                     // No items were pruned, so the radius was never calculated.
                     // Calculate current radius now, rather than waiting for the next overfill.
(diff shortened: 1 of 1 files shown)


portalnet/src/storage.rs
```diff
@@ -444,6 +451,7 @@ impl PortalStorage {
                 debug!("Error writing content ID {id_to_remove:?} to meta db. Reverted: {err:?}",);
             } else {
                 num_removed_items += 1;
+                self.metrics.decrease_entry_count();
```
jacobkaufmann, line 454: I think this ought to go at the bottom of `evict` since that method is public and the container for deletion logic

carver: Ah yes, good catch!

njgheorghita (approved):
LGTM! Just thinking out loud, but it seems like we've more or less adopted this architecture of `Storage` consuming `Metrics`. This will lead to a fair amount of these `self.metrics.increase_*` / `self.metrics.decrease_*` statements throughout the `Storage` code. These are fairly difficult to keep track of. To know when / where to insert them. Changes we make in `Storage` code might have consequential effects on these reported metrics that are not so obvious...

I'm not sure if this design was ever discussed, evaluated, or just came to be. 

It seems to me, like it would be cleaner if we could have `Metrics` consume `Storage`. For example, with respect to tracking the entry count, we could take advantage of the already existing `Storage::total_entry_count()` - as opposed to inc/dec a counter with every insertion / deletion. 

Again, this is more just thinking out loud than a request for changes

pipermerriam:
When I wrote this in python forever ago, I had an API that let me fire "events" from just about anywhere and then a mirror'd API that let me consume those events from just about anywhere else.  This let me do things like:

```
# fire the event
self.events.offer_received.trigger(content_key)
self.events.offer_accepted.trigger(content_key)
self.events.offer_rejected.trigger(content_key, reason)
self.events.offer_accept_complete.trigger(content_key, payload)
self.events.offer_accept_failed.trigger(content_key, reason)

# listen elsewhere
self.events.offer_accept_completed.subscribe(|content_key, payload| {
    ...  # maybe record some metrics data...
})
```

This architecture made for a really clean way to deal with getting access to events that happen in very disparate parts of the codebase.... just something for you guys to consider.

carver:
Yeah, I think ultimately we'll want to have both directions: emitting events *and* a way to request summaries, like `total_entry_count()`.

carver:
I'm also 👍🏻 to move away from the indirection and emit the metrics straight into the Prometheus Gauges. There's already plenty of friction to move away from Prometheus. I don't mind choosing to just bind ourselves to using their API. If we want to switch to another metrics collector later, we can make an internal class that meets the same API.

PR #639, opened 2023-04-04
Wrap NodeId over Enr::NodeId

e3243eric:
### What was wrong?
https://github.com/ethereum/trin/blob/0b1ed5b4ce9837160b74308c027ef74520d70ef0/ethportal-api/src/types/discv5.rs#L9-L12
NodeId is implemented from scratch because Enr::NodeId did not support Serialize/Deserialize traits.

### How was it fixed?

- Enr::NodeId support Serialize/Deserialize in https://github.com/sigp/enr/pull/27.
- Wrap NodeId over Enr::NodeId, like https://github.com/ethereum/trin/blob/0b1ed5b4ce9837160b74308c027ef74520d70ef0/trin-types/src/wrapped/bloom.rs.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ description = "A Rust implementation of the Ethereum Portal Network"
 [dependencies]
 anyhow = "1.0.68"
 discv5 = { version = "0.2.1", features = ["serde"] }
-enr = { version = "=0.7.0", features = ["k256", "ed25519"] }
+enr = { git = "https://github.com/sigp/enr.git", features = ["k256", "ed25519"], rev = "220daa0" }
 eth2_ssz = "0.4.0"
 ethereum-types = "0.12.1"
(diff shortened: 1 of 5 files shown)


pipermerriam:
Will this approach work nicely with `enr.node_id()`?  Will people have to up-cast that value to our `NodeId` type?

e3243eric:
> Will people have to up-cast that value to our `NodeId` type?

It looks like yes. I tried to implement `get_enr` in JSON-RPC, and I still got mismatched types from rustc.

> struct `NodeId` and struct `enr::node_id::NodeId` have similar names, but are actually distinct types

e3243eric:
Perhaps we could wait until the discv5 include the sigp/enr#27, and use discv5::enr directly without wrapping it.

Issue #621, opened 2023-03-30
Persistence between runs without providing private key


pipermerriam:
### What is wrong

Running the trin client in a manner that re-uses node-id and storage requires providing manual CLI args and is a barrier to entry.

### How can it be fixed.

I should be able to run the bare `./trin` binary without any additional arguments and it should:

- [x] default to the hardcoded bootnodes if none are explicitly provided
- [ ] generate a new private key for `node-id` if none exists.  Otherwise, it should re-use the previous one
- [ ] same for storage, lazily make a new one if needed, otherwise re-use.

carver:
> same for storage, lazily make a new one if needed, otherwise re-use.

This should already be done, that trin will default to using the same storage directory, if you are reusing the same node id.

pipermerriam:
I'm not likely to work on this anytime soon so I unassigned.

carver:
Ah yes, in the pared-down project view of the issue, I didn't see the link to Nick's PR to see that he's the one who started in on it.

KolbyML:
After this issue is solved we can potentially look at also closing this issue https://github.com/ethereum/trin/issues/320 since I believe it is generally targeting the same issue.

PR #640, opened 2023-04-04
Change default bootnode arg to testnet bootnodes

njgheorghita:
### What was wrong?
Addresses part of #621 

`--bootnodes default` / `--bootnodes none` / `--bootnodes enr:-...` / `--bootnodes enr:-...,enr:-...` / no `--bootnodes` argument uses default testnet bootnodes

### To-Do

- [x] Clean up commit history


book/src/users/startup.md
```diff
@@ -22,7 +22,7 @@ cargo run -p trin -- --help
 ```
 ### Bootnodes
 
-To quickly connect to the testnet use the `--bootnodes default` flag.
+Trin automatically connects to the testnet bootnodes. Use the `--bootnodes` cli flag to connect to a specific node or to none.
```
pipermerriam, line 25: I think we should start moving away from "testnet" terminology...
```suggestion
Trin automatically connects to a set of hardcoded bootnodes. Use the `--bootnodes` cli flag to connect to a specific node or to none.
```

njgheorghita: as in moving away from "testnet" towards "mainnet"? The assumption in your suggested change being that the hardcoded bootnodes are on the mainnet? or maybe rather the "Portal Network" to avoid confusion with the execution network

pipermerriam: - "portal network"
- "live network"
- "main network"

Basically, we just aren't operating in a "testnet".  It's the main/live/official/real network.

ogenev (approved):
LGTM :+1:

PR #645, opened 2023-04-05
Only use ANSI colors if output to colored terminal

### LINKED ISSUE #638: Turn off log styling for trin when not in colored terminal, especially in portal-hive tests
tracing-subscriber doesn't do any smart detection of output, when deciding whether to add ANSI coloring to log output. A brief look at the years of [ansi issues in tracing-subscriber](https://github.com/tokio-rs/tracing/issues?q=is%3Aissue+is%3Aopen+ansi) suggests that we can't get anything upstreamed at a reasonable speed. The [suggested approach to handling color](https://github.com/tokio-rs/tracing/issues/1160) is to manually turn it off with [`with_ansi()`](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_ansi) in trin.

Some options for how to decide whether to colorize in trin:

1. Turn off all color, always (but I like color in the terminal!)
2. Reproduce [env_logger's color detection behavior](https://docs.rs/env_logger/latest/env_logger/#disabling-colors), which auto-detects `TERM=dumb` for example and disables color
3. Reproduce env_logger's behavior, which detects `RUST_LOG_STYLE=never` as an override (which then requires the portal-hive change)
4. Use an example in the tracing-subscriber repository, which detects whether the output is to a terminal or piping to a file: https://github.com/engula/engula/pull/909

Some of these options may be used together, reasonably.

If necessary, add environment variable hint, like this: https://github.com/ogenev/portal-hive/pull/20

carver:
### What was wrong?

Fix #638 

### How was it fixed?

Only import the libraries as needed, per OS.

Unfortunately, increases the size of trin-utlis a bit, to add log tracing.

On *nix, check if output is going to terminal (or for example, piped to a file). Also, if going to a terminal, make sure the terminal supports color (ie~ the environment variable TERM is not "dumb").

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5661,8 +5661,12 @@ dependencies = [
 name = "trin-utils"
 version = "0.1.0"
 dependencies = [
+ "ansi_term",
+ "atty",
  "hex",
  "thiserror",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/ethportal-peertest/src/main.rs
+++ b/ethportal-peertest/src/main.rs
@@ -2,6 +2,7 @@
 
 use std::{thread, time};
 use tracing::info;
+use trin_utils::log::init_tracing_logger;
 
 use ethportal_peertest::{
     cli::PeertestConfig,
@@ -11,7 +12,7 @@ use ethportal_peertest::{
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
-    tracing_subscriber::fmt::init();
+    init_tracing_logger();
 
     let peertest = launch_peertest_nodes(2).await;
     // Short sleep to make sure all peertest nodes can connect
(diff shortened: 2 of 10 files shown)


carver:
I confirmed locally that this turns off the ANSI coloring in portal-hive runs.

carver:
I tried this out using Copilot. Not a bad first impression!

ogenev (approved):
Looks good :100:

njgheorghita (approved):
seems like it might be worth checking for unused dependencies in this pr - https://crates.io/crates/cargo-udeps is the easiest way imo - since we're extracting the `tracing` & `tracing-subscriber` construction into `trin-utils`

carver:
Nothing log-related, but it claimed a few libraries are unused. Pushing up the commit to see if the build fails with them missing...

carver:
Nope, all the libraries were false positives.

PR #646, opened 2023-04-05
refactor constants for portal wire message limits

### LINKED ISSUE #523: Properly calculate max content size for utp transmission
Properly calculating whether a message is large enough to require transmission over utp implementation. Implementer should determine whether this calculation belongs in `trin-core` code or inside `utp` / [utp](https://github.com/jacobkaufmann/utp).
- [link to todo](https://github.com/ethereum/trin/blob/5d1c1bfcbb58f18ec4ea035b7b2d6fe87e02b1d5/trin-core/src/portalnet/overlay_service.rs#L966)

jacobkaufmann:
### What was wrong?

fixes #523

### How was it fixed?

* move constants for portal wire message limits to `portalnet::types::messages.rs`
* renaming:
  * from `MAX_NODES_SIZE` to `MAX_PORTAL_NODES_ENRS_SIZE`
  * from `MAX_CONTENT_NODES_SIZE` to `MAX_PORTAL_CONTENT_PAYLOAD_SIZE`
* use `MAX_PORTAL_CONTENT_PAYLOAD_SIZE` when constructing a portal content response to determine whether uTP transmission is necessary
* the following values were changed:
  * `discv5` talk request overhead: from `103` to `109`
  * `portal` nodes `enrs` overhead: from `2` to `6`
  * `portal` content `enrs` overhead: from `5` to `6`

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -64,13 +64,59 @@ use trin_validation::validator::Validator;
 
 /// Maximum number of ENRs in response to FindNodes.
 pub const FIND_NODES_MAX_NODES: usize = 32;
+
 /// Maximum number of ENRs in response to FindContent.
 pub const FIND_CONTENT_MAX_NODES: usize = 32;
+
 /// With even distribution assumptions, 2**17 is enough to put each node (estimating 100k nodes,
 /// which is more than 10x the ethereum mainnet node count) into a unique bucket by the 17th bucket index.
 const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
+
 /// Bucket refresh lookup interval in seconds
 const BUCKET_REFRESH_INTERVAL_SECS: u64 = 60;
+
+/// The maximum size of a Discv5 packet.
+const MAX_DISCV5_PACKET_SIZE: usize = 1280;
+
+/// The maximum size of a Discv5 talk request payload.
+///
+/// Discv5 talk request overhead:
+///   * masking IV length: 16
+///   * static header (protocol ID || version || flag || nonce || authdata-size) length: 23
+///   * authdata length: 32
+///   * HMAC length: 16
+///   * (max) talk request ID length: 8
+///   * (max assumed) talk request protocol length: 8
+///   * RLP byte array overhead: 6
+const MAX_DISCV5_TALK_REQ_PAYLOAD_SIZE: usize =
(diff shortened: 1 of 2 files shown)


jacobkaufmann:
it would be appreciated if someone could sanity check the SSZ overhead calculations

ogenev:
How do we feel about moving those constants to a separate file inside the `portalnet` crate?

jacobkaufmann:
I don't mind doing that, but where else would the constants be used? if the only use is within `OverlayService`, then it makes sense to me for those constants to reside in the same file. I thought about defining the message-specific constants as associated constants for the message structs (e.g. `Content::MAX_ENRS_SIZE`), but I decided against it.

mrferris (approved):
Looks good 👍

Maybe we think of a name for that var that doesn't have `CONTENT_CONTENT`? I felt like my brain was skipping when I read that. Maybe just a single `MAX_PORTAL_CONTENT_PAYLOAD_SIZE` for both vars? Not that big a deal though.

portalnet/src/overlay_service.rs
```diff
@@ -64,14 +64,60 @@ use trin_validation::validator::Validator;
...
 /// which is more than 10x the ethereum mainnet node count) into a unique bucket by the 17th bucket index.
 const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
+
 /// Bucket refresh lookup interval in seconds
 const BUCKET_REFRESH_INTERVAL_SECS: u64 = 60;
 
+/// The maximum size of a Discv5 packet.
+const MAX_DISCV5_PACKET_SIZE: usize = 1280;
+
+/// The maximum size of a Discv5 talk request payload.
```
njgheorghita, line 81: Could you also include the calculated size of these constants in the comments, so we don't have to do math / go look up the `ssz` constants if we want to know

njgheorghita (approved):
Looks good! Though, my 2 cents w/ regards to the constants location. I'd like to see them in a workspace-root level constants file - eg `portalnet/src/constants.rs`. Agreed that we're unlikely to use them elsewhere. But, this file is way too big already. It's a bit of a nightmare to navigate, and I think that's reason enough to move the constants & their somewhat lengthy docstrings.

On the other hand, it's possible that this can wait, since i'd like to propose the norm of a workspace-root level `constants.rs` file to house all constants, get it in our style guide, then in a following pr i'll comb through and implement it. 

So, it's up to you how you want to leave it in this pr

PR #647, opened 2023-04-05
Only prune until DB is slightly overfull

### LINKED ISSUE #624: Increase the radius at launch, when the storage limit increases

carver:
### What was wrong?

Second attempt to fix #624 

### How was it fixed?

Following Piper's suggestion:
https://github.com/ethereum/trin/issues/624#issuecomment-1495106383

We leave DB slightly overfull to detect that we are at capacity. Then we set the radius there.

On `trin` reboot, we either end up with 100% radius, or we are full, and set the radius to exactly the furthest content id.

### To-Do
- [x] Clean up commit history
- [x] Add test that overfull database is not under-capacity after pruning
- [x] Add test that radius is set after pruning
- [x] Set radius when reaching DB capacity during prune

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -421,6 +421,16 @@ impl PortalStorage {
             let id_to_remove =
                 // Always expect a content id if capacity_reached(), even at 0kb capacity
                 farthest_content_id.expect("Capacity reached, but no farthest id found!");
+            // Test if removing the item would put us under capacity
+            if self.does_eviction_cause_under_capacity(&id_to_remove)? {
+                // If so, we're done pruning
+                debug!(
+                    "Removing item would drop us below capacity. We target slight overfilling. {}",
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
looks good!

PR #650, opened 2023-04-07
Display content key rather than id

njgheorghita:
### What was wrong?
Imo - content keys are far more useful to display - especially for debugging purposes.

### To-Do

- [x] Clean up commit history


carver (approved):
LGTM

PR #651, opened 2023-04-08
Switch from StructOpt to Clap

### LINKED ISSUE #627: Switch from StructOpt to Clap
StructOpt is being deprecated. Switch to the features which are now integrated into Clap.

This is being updated one crate at a time.

Update: The last one left is utp-testing.

e3243eric:
### What was wrong?

Fix #627 

### How was it fixed?

Switch from StructOpt to Clap.

### To-Do

- [x] ethportal-peertest
- [x] purge_db
- [ ] trin-bridge
- [ ] trin-cli
- [ ] trin-types
- [ ] utp-testing
- [ ] Clean up commit history

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ description = "A Rust implementation of the Ethereum Portal Network"
 
 [dependencies]
 anyhow = "1.0.68"
+clap = { version = "4.2.1", features = ["derive"] }
 discv5 = { version = "0.2.1", features = ["serde"] }
 enr = { version = "=0.7.0", features = ["k256", "ed25519"] }
 eth2_ssz = "0.4.0"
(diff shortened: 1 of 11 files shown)


carver:
If you can, I'm in favor of breaking this up into multiple PRs. No need to get the whole repo into one PR. That might mean depending on both clap and structopt in some places, for a little while. But I prefer that to having a massive PR to review.

carver:
Thanks for taking this on!

perama-v:
One tip from using clap without structopt in Glados is that when passing args the first arg is treated as the app name. So, `""` or `"test"` as the first argument allows for reading of arguments:

```rs
let result = Args::parse_from(["test", "--transport", "ipc", "--ipc-path", IPC_PATH]);
```
https://github.com/ethereum/glados/blob/776d2356ab26d659d74db11dce784345586b7ce8/glados-audit/src/cli.rs#L33

PR #653, opened 2023-04-10
Add a SQL index to speed up data size checks

### LINKED ISSUE #649: Investigate the speed of querying the total database size
Starting hypothesis: SELECT TOTAL() with many rows will probably be too slow

The storage uses a TOTAL to track whether it's full, using:
https://github.com/ethereum/trin/blob/e2e96adc8ceee16774289f834681f8c4b4704e31/portalnet/src/storage.rs#L839

It's not totally unreasonable to hit a million rows eventually. Scanning 1M rows in a database could take a significant amount of time. This is a bigger concern, because the lookup happens regularly and inline for every insert.

Let's see how much faster we can make it by adding an index on the column. Normally, it doesn't help SUM, but in the very special case of summing the full index, it could. That's only if the query optimizer is smart enough though, per https://stackoverflow.com/a/437924/8412986

Try adding an index on the column. Then, run the TOTAL() query using an explain, to see if SQLite is smart enough to use the index https://www.sqlite.org/eqp.html

If so, let's make an index standard now, to have it be part of as many installs as possible. If not, we probably need to find a different way to keep track of the total content size. (It still may not be fast enough with the index, but it should buy us some time)

carver:
In a local test with 2M rows, adding this index sped up a SELECT TOTAL(content_size) query time from 125ms to 75ms.

### What was wrong?

Fix #649 

### How was it fixed?

Add an index and shelve the issue, for now.

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -868,7 +868,8 @@ const CREATE_QUERY: &str = "create table if not exists content_metadata (
                                 content_id_short INTEGER NOT NULL,
                                 content_key TEXT NOT NULL,
                                 content_size INTEGER
-                            )";
+                            );
+                            create index content_size_idx on content_metadata(content_size);";
 
 const INSERT_QUERY: &str =
     "INSERT OR IGNORE INTO content_metadata (content_id_long, content_id_short, content_key, content_size)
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
:shipit:

carver:
I added some more indices. This is a small, low-risk, change, so I'll rage-merge after the tests pass. I'll quickly follow up if anyone notices something off.

PR #656, opened 2023-04-10
Track and display network message stats

### LINKED ISSUE #655: Track count of all sent/received messages in overlay (service)
There is already a tracking of most (all?) of these messages, but...

- [x] need an actual description for metrics, instead of "help"
- [x] always track metrics, not just if CLI flag is used
- [x] use labels for sent/received
- [x] use labels for message type (eg~ ping/pong)
- [x] use labels for subprotocol name
- [x] add unit to end of name: `_total`
- [x] use an IntCounter instead, it ["Provides better performance if metric values are all positive integers (natural numbers)."](https://docs.rs/prometheus/latest/prometheus/type.IntCounter.html)
- [x] print out *something* to the logs to show some progress
- [x] ~test that metrics are making it to a local dashboard~ *tracking in a separate issue*
- [x] ~maybe~ switch to using custom registries ~if necessary~, like in earlier PRs
- [x] A number of metrics were not being collected, like responses in either direction (like never collecting `Accept`, `Nodes`, or `Content`), make sure to collect all requests and responses in both directions

It's nice to use labels because they make it easier to combine data in single charts. It also makes it easier to report the stat, and enabled some significant code deletion. Finally, it's the [recommended best practice](https://prometheus.io/docs/practices/naming/#labels).

carver:
### What was wrong?

Fix #655 

### How was it fixed?

Combine metrics into one, using labels to differentiate. Display offers and accepts in the heartbeat log. Other things are being tracked, this is shown just to give you a flavor of what's going on.

The heartbeat log now looks like:
```
2023-04-13T04:09:28.335984Z  INFO trin_history: reports~ data: radius=0.4% content=2086.4/2000kb #=27 disk=2509.3kb; msgs: offers=7/8, accepts=3/4
```

### To-Do

- [x] Clean up commit history
- [x] Add the custom registry trick to pass tests.

### DIFF
--- a/portalnet/src/metrics.rs
+++ b/portalnet/src/metrics.rs
@@ -1,9 +1,13 @@
 use prometheus_exporter::{
     self,
-    prometheus::{register_counter, Counter},
+    prometheus::{
+        opts, register_int_counter_vec, register_int_counter_vec_with_registry, IntCounterVec,
+        Registry,
+    },
 };
+use tracing::error;
 
-use crate::types::messages::ProtocolId;
+use crate::types::messages::{ProtocolId, Request, Response};
 
 /// General Metrics Strategy (wip)
 /// - Each module should maintain its own metrics reporter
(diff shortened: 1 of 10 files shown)


njgheorghita (approved):
Nice! I really like how it's looking in `overlay_service.rs` after this

portalnet/src/overlay.rs
```diff
@@ -540,6 +545,20 @@ where
             );
         }
     }
+
+    pub fn get_summary_info(&self) -> String {
+        format!(
+            "offerin={}, acceptout={}, offerout={}, acceptin={}",
```
jacobkaufmann, line 551: IMO the extra character is worth the legibility

```suggestion
            "offer-in={}, accept-out={}, offer-out={}, accept-in={}",
```

portalnet/src/metrics.rs
```diff
@@ -14,79 +18,99 @@ use crate::types::messages::ProtocolId;
...
+    fn increment_message_count(&self, protocol: &ProtocolId, direction: &str, message: Message) {
+        let protocol_label = Self::protocol_to_label(protocol);
+        let message_type_label = Self::message_to_label(message);
+        let labels = [protocol_label, direction, message_type_label];
+        self.message_count.with_label_values(&labels).inc();
     }
 
-    pub fn report_outbound_offer(&self) {
-        self.outbound_offer.inc();
+    fn message_to_label(message: Message) -> &'static str {
```
jacobkaufmann, line 87: to avoid cloning each message to increment the corresponding metric, we could implement conversions from the `&Request` and `&Response` types

jacobkaufmann (commented):
it would be good to define enums for the various label categories with conversions from existing enums where possible

portalnet/src/overlay.rs
carver: Hah, I was just thinking about how to make it even shorter, since we'll end up with even more stuff in there. I landed on:
```
023-04-13T01:45:10.714353Z  INFO trin_history: storage-report: radius=0.0% content=2086.4/2000kb #=30 disk=5407.0kb; msg-report: offered=523/526, accepted=656/656
```

portalnet/src/metrics.rs
```diff
@@ -14,79 +18,99 @@ use crate::types::messages::ProtocolId;
...
+        match protocol {
+            ProtocolId::State => "state",
+            ProtocolId::History => "history",
+            ProtocolId::TransactionGossip => "transaction_gossip",
+            ProtocolId::HeaderGossip => "header_gossip",
+            ProtocolId::CanonicalIndices => "canonical_indices",
+            ProtocolId::Utp => "utp",
+        }
+    }
+}
```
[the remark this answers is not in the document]
carver: 👍🏻 yeah, much better, especially when using directly with the borrowed request & response, which means we can avoid the allocations for cloning the Req/Resp, and for creating the message.

As @jacobkaufmann implied, it's totally worth optimizing those away when we're potentially calling into these metrics quite a bit.

portalnet/src/overlay.rs
carver: Yup, exactly! I feel like it's a good sign that you intuited the explanation.

PR #657, opened 2023-04-12
Recursive find nodes jsonrpc api endpoint

### LINKED ISSUE #628: JSON-RPC support for `recursiveFindNodes`
### What is wrong

It seems that `recursive_find_nodes` isn't yet implemented

https://github.com/ethereum/trin/blob/d8c15e76bdc7331680440df1d45699d61cbcc7d9/rpc/src/history.rs#L98-L100

This is a blocker to getting network exploration done in glados

### How can it be fixed

~~Looks like it just needs to be wired up.~~

Depends on #637

### LINKED ISSUE #637: Implement Recursive Find Nodes in overlay.rs

njgheorghita:
### What was wrong?
Fixes #637 
Fixes #628 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/scenarios/find.rs
+++ b/ethportal-peertest/src/scenarios/find.rs
@@ -1,3 +1,4 @@
+use discv5::enr::NodeId;
 use serde_json::{json, Value};
 
 use crate::{
@@ -5,6 +6,7 @@ use crate::{
     Peertest, PeertestConfig,
 };
 use trin_types::{constants::CONTENT_ABSENT, jsonrpc::params::Params};
+use trin_utils::bytes::hex_encode;
 
 pub fn test_trace_recursive_find_content(_peertest_config: PeertestConfig, peertest: &Peertest) {
     let uniq_content_key = "0x0015b11b918355b1ef9c5db810302ebad0bf2544255b530cdce90674d5887bb286";
@@ -106,4 +108,41 @@ pub fn test_trace_recursive_find_content_local_db(
     let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
     assert_eq!(result["content"], json!(HISTORY_CONTENT_VALUE.to_string()));
     assert_eq!(result["route"], json!([]));
+}
+
+pub async fn test_recursive_find_nodes_self(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    let target_node_id = hex_encode(peertest.bootnode.enr.node_id().raw());
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindNodes".to_string(),
+        id: 15,
+        params: Params::Array(vec![json!(target_node_id)]),
+    };
+    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
+    assert_eq!(result, json!([peertest.bootnode.enr.to_base64()]));
+}
+
+pub async fn test_recursive_find_nodes_peer(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    let target_enr = peertest.nodes[0].enr.clone();
+    let target_node_id = hex_encode(target_enr.node_id().raw());
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindNodes".to_string(),
+        id: 15,
+        params: Params::Array(vec![json!(target_node_id)]),
+    };
+    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
+    assert_eq!(result, json!([target_enr.to_base64()]));
+}
+
+pub async fn test_recursive_find_nodes_random(
+    _peertest_config: PeertestConfig,
+    peertest: &Peertest,
+) {
+    let target_node_id = hex_encode(NodeId::random());
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindNodes".to_string(),
+        id: 15,
+        params: Params::Array(vec![json!(target_node_id)]),
+    };
+    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
+    assert_eq!(result.as_array().unwrap().len(), 2);
 }
--- a/portalnet/src/overlay.rs
+++ b/portalnet/src/overlay.rs
@@ -441,6 +441,51 @@ where
         }
     }
 
+    pub async fn lookup_node(&self, target: NodeId) -> Vec<Enr> {
+        if target == self.local_enr().node_id() {
+            return vec![self.local_enr()];
+        }
+        let connected_peer = self
+            .kbuckets
+            .write()
+            .iter()
+            .filter(|entry| entry.status.is_connected())
+            .map(|entry| *entry.node.key.preimage())
+            .find(|node_id| node_id == &target);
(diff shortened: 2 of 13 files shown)


njgheorghita:
Ready for review - though for now let's just ignore the hacky `NodeId` type I created temporarily until we finalize our `trin-types::NodeId` implementation

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -107,3 +109,39 @@ pub fn test_trace_recursive_find_content_local_db(
...
+
+pub async fn test_recursive_find_nodes_peer(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    let target_node_id = hex_encode(peertest.nodes[0].enr.node_id().raw());
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindNodes".to_string(),
+        id: 15,
+        params: Params::Array(vec![json!(target_node_id)]),
+    };
+    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
+    assert_eq!(result, json!([peertest.nodes[0].enr.to_base64()]));
```
jacobkaufmann, line 132:
```suggestion
    let target_enr = peertest.nodes[0].enr;
    let target_node_id = hex_encode(target_enr.node_id().raw());
    let request = JsonRpcRequest {
        method: "portal_historyRecursiveFindNodes".to_string(),
        id: 15,
        params: Params::Array(vec![json!(target_node_id)]),
    };
    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
    assert_eq!(result, json!([target_enr.to_base64()]));
```

portalnet/src/overlay.rs
```diff
@@ -441,6 +441,39 @@ where
...
+        if target == self.local_enr().node_id() {
+            return vec![self.local_enr()];
+        }
+        if self.discovery.connected_peers().contains(&target) {
+            let enr = self
+                .discovery
+                .find_enr(&target)
+                .expect("cannot find enr expected to exist in routing table");
+            return vec![enr];
+        }
```
jacobkaufmann, line 454: IIUC, we do not do the following because the node may not be in the connected state?

```suggestion
        if let Some(enr) = self.discovery.find_enr(&target) {
            return vec![enr];
        }
```

why not look directly at our own routing table?

```rust
self.kbuckets
  .write()
  .iter()
  .filter(|entry| entry.status.is_connected())
  .find(|entry| *entry.node.key.preimage() == target)
```

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -107,3 +109,39 @@ pub fn test_trace_recursive_find_content_local_db(
...
 }
+
+pub async fn test_recursive_find_nodes_self(_peertest_config: PeertestConfig, peertest: &Peertest) {
+    let target_node_id = hex_encode(peertest.bootnode.enr.node_id().raw());
+    let request = JsonRpcRequest {
+        method: "portal_historyRecursiveFindNodes".to_string(),
+        id: 15,
+        params: Params::Array(vec![json!(target_node_id)]),
+    };
+    let result = make_ipc_request(&peertest.bootnode.web3_ipc_path, &request).unwrap();
```
mrferris, line 120:
```suggestion
   let client = reth_ipc::client::IpcClientBuilder::default()
        .build(&peertest.bootnode.web3_ipc_path)
        .await
        .unwrap();
  let result: Vec<Enr> = client.recursive_find_nodes(node_id).await.unwrap();
```
I'm partial to this syntax over the `JsonRpcRequest`/`make_ipc_request` syntax because it looks nicer and the deserialization from json is tested as well.

ethportal-peertest/src/scenarios/find.rs
ogenev, line 1: As @mrferris pointed out, while we are on this, we may want to convert this scenario to use the jsonrpsee framework.

njgheorghita: Left a response on the other comment

ethportal-peertest/src/scenarios/find.rs
carver: Sure, might as well create the issue now. We don't really need agreement about whether we're going to do it, it's just good to have a place to discuss the refactor. It's even valuable to have an issue where we decide not to do something and simply close it, for our future reference.

portalnet/src/overlay.rs
```diff
@@ -441,6 +441,43 @@ where
...
+        let connected_peer = self
+            .kbuckets
+            .write()
+            .iter()
+            .filter(|entry| entry.status.is_connected())
+            .map(|entry| *entry.node.key.preimage())
+            .find(|node_id| node_id == &target);
+        if let Some(entry) = connected_peer {
+            let enr = self.discovery.find_enr(&entry).expect("ENR should exist");
```
carver, line 456: Panicking here seems a little extreme. Sure, it shouldn't happen, but do we really want to crash trin if we have an entry in our kbuckets that the discv5 lib forgot about?

I'd say an `error!()` log and an empty `vec` would be sufficient.

trin-history/src/jsonrpc.rs
```diff
@@ -170,6 +170,17 @@ impl HistoryRequestHandler {
...
                     let _ = request.resp.send(response);
                 }
+                HistoryEndpoint::RecursiveFindNodes(node_id) => {
+                    let node_id = discv5::enr::NodeId::from(node_id.0);
+                    let mut nodes = self.network.overlay.lookup_node(node_id).await;
+                    nodes.sort_by(|a, b| {
+                        XorMetric::distance(&node_id.raw(), &a.node_id().raw())
+                            .cmp(&XorMetric::distance(&node_id.raw(), &b.node_id().raw()))
+                    });
+                    let nodes: Vec<Enr> = nodes.into_iter().take(16).collect();
```
carver, line 180: 16 is used across a few methods that have this behavior. It probably belongs in some rpc constants file. I don't see one yet, and this is the only method written so far (there's going to be the same methods for state and basic discv5). I am fine punting for now I suppose.

trin-types/src/node_id.rs
```diff
@@ -0,0 +1,24 @@
+use serde::{Deserialize, Serialize};
+use serde_hex::{SerHex, StrictPfx};
+use std::ops::Deref;
+
+type RawNodeId = [u8; 32];
+
+/// Discv5 NodeId
+// TODO: Wrap this type over discv5::NodeId type
```
carver, line 8: We're more likely to make sure this gets followed up on, if we move it out of a TODO and into a Ready issue in the tracker.

carver (approved):
AIUI, this PR is a requirement before glados can use trin again (b/c of the network explorer). So I'm definitely in favor of linking the jsonrpsee discussion into a new issue, to help get it merged. 👍🏻

ethportal-peertest/src/scenarios/find.rs
njgheorghita: I went ahead and implemented the refactor here #666

trin-history/src/jsonrpc.rs
njgheorghita: Yeah, seems like it's fine to save this consolidation until we build out the next network and have a more clear idea of where to house things like these constants.

PR #658, opened 2023-04-12
configure storage capacity in megabytes

### LINKED ISSUE #634: Switch storage limit from --kb to --mb
Block bodies and receipts can already take up hundreds of kb (or more?). I don't think we have a real use case where users need to store less than 1MB of data (besides 0MB), so it's probably cleaner to use MB the unit of choice. Off the top of my head, that means changing:

- CLI arg from kb to mb
- heartbeat log from kb to mb
- prometheus metrics from kb to mb

This will be a win in more readable output and easier-to-enter values anyway.

jacobkaufmann:
### What was wrong?

fixes #634

### How was it fixed?

* change CLI argument from `kb` to `mb`, keeping the default value at 100MB
* change `PortalStorageConfig` to configure capacity in MB
* change prometheus gauges to report storage in bytes
* refactor storage tests
* refactor mentions of `kb` throughout repo

### To-Do

- [x] Clean up commit history


portalnet/src/storage.rs
```diff
@@ -27,6 +27,9 @@ use trin_types::content_key::{ContentKeyError, HistoryContentKey, OverlayContent
 use trin_types::distance::{Distance, Metric, XorMetric};
 use trin_utils::bytes::{hex_decode, hex_encode, ByteUtilsError};
 
+const BYTES_IN_MB_U64: u64 = 1000 * 1000;
+const BYTES_IN_MB_F64: f64 = 1000.0 * 1000.0;
```
pipermerriam, line 31: If we use numbers like 1024 here we end up squeaking out just a little extra storage from each user....

jacobkaufmann: I'm not opposed to that

portalnet/src/storage.rs
```diff
@@ -546,7 +549,7 @@ impl PortalStorage {
         })?;
         let storage_usage = Self::get_total_size_of_directory_in_bytes(data_dir)?;
         self.metrics
-            .report_total_storage_usage(storage_usage as f64 / 1000.0);
+            .report_total_storage_usage(storage_usage as f64 / BYTES_IN_MB_F64);
```
carver, line 552: This is my mistake from before: this line duplicates the stat reporting, since it's also reported other times the method is called internally. I remember thinking that I didn't like it being a side effect inside this `get_` method, but must have forgotten to remove it.  Seems like it can be deleted.

portalnet/src/storage.rs
```diff
@@ -419,7 +422,7 @@ impl PortalStorage {
         // Delete furthest data until our data usage is less than capacity.
         while self.capacity_reached()? {
             let id_to_remove =
-                // Always expect a content id if capacity_reached(), even at 0kb capacity
+                // Always expect a content id if capacity_reached(), even at 0mb capacity
```
carver, line 425: I realized when I read this comment that I don't understand: why won't it be None with a 0mb capacity?

After digging a bit, I think it was a more helpful comment before the commit: 201dba58

Now, I'd say it's more accurate to have a comment like:
```suggestion
                // An empty database cannot have .capacity_reached() be true, because it's only true when content size > capacity,
                // so content ID will always be Some() inside the loop.
```

portalnet/src/storage.rs
```diff
@@ -754,45 +758,45 @@ impl PortalStorage {
 
 #[derive(Debug)]
 struct StorageMetrics {
-    content_storage_usage_kb: Gauge,
-    total_storage_usage_kb: Gauge,
-    storage_capacity_kb: Gauge,
+    content_storage_usage_mb: Gauge,
+    total_storage_usage_mb: Gauge,
+    storage_capacity_mb: Gauge,
```
carver, line 763: Sorry, I gave the suggestion to switch the suffixes to `_mb` before I read [best practices](https://prometheus.io/docs/practices/naming/#base-units).

I think the best option would be to switch it to `_bytes` in the metric, and convert from MB for the CLI arg, and to MB for the heartbeat log.

I'm going to go in eventually and follow more of these metrics to best practices (like using labels for the protocol, instead of adding it onto the metric name), so either way it will get done eventually. But if you feel up to making the switch that would be appreciated.

portalnet/src/storage.rs
```diff
@@ -1073,15 +1079,12 @@ pub mod test {
...
-        // change will make the minimum storage size 1MB, so 32 keys should still be sufficient then.
+        // Fill up the storage.
         for _ in 0..32 {
             let content_key = generate_random_content_key();
-            let value: Vec<u8> = "OGFWs179fWnqmjvHQFGHszXloc3Wzdb4".into();
+            let value: Vec<u8> = vec![0; 32000];
             storage.store(&content_key, &value)?;
             // Speed up the test by ending the loop as soon as possible
-            // At 1kb, that should be immediately after the first item is stored.
-            // At 1MB, it will take the full 32 items.
```
carver, line 1084: lol, why did I write that 32 items at 32 bytes would fill 1MB? 😅 Just off by a factor of 1000x :P

Thanks for deleting this garbage.

trin-types/src/cli.rs
```diff
@@ -129,13 +129,13 @@ pub struct TrinConfig {
...
     pub networks: Vec<String>,
 
-    /// Number of Kilobytes to store in the DB
+    /// Storage capacity specified in megabytes.
     #[structopt(
-        default_value(DEFAULT_STORAGE_CAPACITY),
+        default_value(DEFAULT_STORAGE_CAPACITY_MB),
         long,
-        help = "Maximum number of kilobytes of total data to store in the DB"
+        help = "Maximum number of megabytes of total data to store in the DB"
```
carver, line 136: Might be nice to clarify to users something like:
```suggestion
        help = "Maximum number of megabytes of total data to store in the DB. The actual disk size will exceed the limit, because of overhead."
```

carver (approved):
LGTM, nothing but a few suggestions offered.

njgheorghita (approved):
Looks good! There are just a few dangling `kb` instances that should be addressed in this pr. Searching for `"kb"` through the repo should reveal them. I'm seeing 3 files in `/book/` and `/trin-bridge/src/main.rs` which need some `kb` -> `mb` conversions. The only one I'm unsure of is `trin-cli/src/dashboard/metrics_dashboard.json.template` ... technically we're not supporting `trin-cli` anymore - and i'm sure this will get addressed as we continue to develop our metrics strategy so imo it's ok to just leave this one for now and not interfere with the auto-generated config.

portalnet/src/storage.rs
jacobkaufmann: after some discussion with the team, I opted to keep the multiples of 1000. a higher default is an alternative to elicit more storage from the median user

portalnet/src/storage.rs
jacobkaufmann: went ahead and took care of this, using `bytes` for the individual gauges and then converting to `mb` in the summary

jacobkaufmann:
@njgheorghita good catch! I left the uses of `kb` in `trin-cli` untouched.

PR #660, opened 2023-04-13
Radius metric: higher precision, switch to ratio, log approprate number of digits

### LINKED ISSUE #659: Update radius log to show more significant digits
A tenth of a percent is not small enough. Probably will want to aim for parts per million. Even cooler would be to switch from % to ppm when the percent goes under ~0.5%.

This requires more changes:
- update prometheus metric name suffix to `_ratio` (per [best practices](https://prometheus.io/docs/practices/naming/#base-units))
- store radius as a ratio (0-1) - also best practice
- collect more precise measurement of radius (only using top byte of precision now, but need top 3 bytes to get to 1ppm precision)
- in the log display, multiply ratio by 1M
- bonus: show as a percent still, if radius is large

carver:
### What was wrong?

Fix #659 

It won't take long for 1 digit of percentage to be rounded down to 0, for even moderately sized nodes. So we need to generate and display more precision for the logs.

### How was it fixed?

Get more bytes of the radius, to get a higher precision estimate. We don't expect to display a higher precision than 1 in a million, so 3 bytes would be sufficient. The nearest supported integer is at u32 with 4 bytes.

Ratio (0-1) is favored over % (0-100) in the prometheus system. Switch to saving that value, plus the corresponding metric name change.

Depending on the size of the radius, grow or shrink the number of displayed digits in the heartbeat log.

### To-Do

- [ ] Clean up commit history
- [x] Show more digits in the log, maybe even dynamically as the number changes?

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -751,7 +751,7 @@ struct StorageMetrics {
     content_storage_usage_kb: Gauge,
     total_storage_usage_kb: Gauge,
     storage_capacity_kb: Gauge,
-    radius_percent: Gauge,
+    radius_ratio: Gauge,
     entry_count: IntGauge,
 }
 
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
🚢

Issue #407, opened 2022-08-12
Refactor DB utils


jacobkaufmann:
The functions in [`trin-core/src/utils/db.rs`](https://github.com/ethereum/trin/blob/c07fce832ce735faa346d932d996833869cf9b9b/trin-core/src/utils/db.rs) do not behave the way that a caller would expect. For instance, `get_data_dir` returns a `PathBuf` but may also create a directory. `setup_temp_dir` returns a `TempDir` but also sets an environment variable (which is used in `get_root_path`).

These functions should be free of side effects and pure to the extent possible.

KolbyML:
Closing this cause doing it would increase code complexity. For example when getting the directory we would have to call ``fs::create_dir_all`` everywhere we call  ``get_data_dir`` so they are pretty codependent. Also this saves us from having to call ``fs::create_dir_all`` multiple times through the code which helps with readability as there is less cluter.

The same codependence case applies with setup_temp_dir and setting the enviroment variable

If anyone has a clear good reason or thinks they can make a beneficial change we can reopen it.

PR #665, opened 2023-04-14
Refactor DB utils to decrease side effects

KolbyML:
### What was wrong?
Jacob made this issue https://github.com/ethereum/trin/issues/407 and I thought I would take a jab at it. After I clicked "New Pull Request" I seen he made a draft PR for it roughly 8 months ago. I found it after looking through the other PR's to see what the report format looked like. So if there is no need for this PR that is fine I will close it, I wish the issue was marked cause I don't want to step on anybody's toes. But since that PR had not been touched for so long I thought I would create my pull request anyways and see how it goes.

### How was it fixed?
I refactored ``create_dir_all`` out of ``get_data_dir`` and placed it after ``setup_rocksdb`` and ``setup_sql`` since that is where it seemed best suited looking to remove side effects in mind. It being best suited in those places since it would only need to create them possibly in setup was my assumption.

Then for removing the side effect list in the issue from ``setup_temp_dir``. I noticed that setup_temp_dir is only used in 2 places in storage tests, and ``trin/src/lib.rs`` so since the storage tests didn't need ``get_root_path()`` which used the environment variable ``TRIN_DATA_ENV_VAR`` it only made set to set the variable in ``trin/src/lib.rs``. Which eliminates the side effect.

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -718,6 +718,7 @@ impl PortalStorage {
         let mut data_path: PathBuf = get_data_dir(node_id).map_err(|err| {
             ContentStoreError::Database(format!("Unable to get data dir for rocksdb: {err:?}"))
         })?;
+        fs::create_dir_all(&data_path).expect("Unable to create data directory folder");
         data_path.push("rocksdb");
         info!(path = %data_path.display(), "Setting up RocksDB");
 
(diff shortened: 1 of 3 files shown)


carver:
Ah, interesting. Thanks for taking a look! Let's see how #654 shakes out

PR #666, opened 2023-04-14
Use jsonrpsee based testing in scenarios

njgheorghita:
### What was wrong?
`ethportal-peertest` was out of date. Needed some spring cleaning to take advantage of the new jsonrpsee based client.

### How was it fixed?
- Moved `NodeId` type to `trin-types`
- Removed no-longer needed jsonrpc utils
- Removed no-longer needed cli utils
- Updated tests to use a single, shared ipc client

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [x] Clean up commit history

PR #667, opened 2023-04-15
Switch from StructOpt to Clap in ethportal-peertest

e3243eric:
### What was wrong?

Related to #627. A tiny PR of #651.

### How was it fixed?

Switched from StructOpt to Clap in `ethportal-peertest`.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/cli.rs
+++ b/ethportal-peertest/src/cli.rs
@@ -1,5 +1,4 @@
-use std::{env, ffi::OsString};
-use structopt::StructOpt;
+use clap::{Parser, ValueEnum};
 use trin_types::cli::{
     DEFAULT_WEB3_HTTP_ADDRESS as DEFAULT_TARGET_HTTP_ADDRESS,
     DEFAULT_WEB3_IPC_PATH as DEFAULT_TARGET_IPC_PATH,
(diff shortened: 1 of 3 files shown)


KolbyML (approved):

ethportal-peertest/src/cli.rs
```diff
@@ -1,68 +1,81 @@
...
 
-    pub fn new_from<I, T>(args: I) -> Result<Self, String>
-    where
-        I: Iterator<Item = T>,
-        T: Into<OsString> + Clone,
-    {
-        let config = Self::from_iter(args);
+#[cfg(test)]
+mod test {
+    use super::*;
```
KolbyML, line 55: A wildcard import shouldn't be used here. If you could fix that everything else seems good. Good addition with the tests

KolbyML (changes requested):
Overall the PR looks good. The addition of the tests is very nice. If you could resolve the wildcard import everything else looks good 👍

KolbyML (approved):
Overall the PR looks good. The addition of the tests is very nice. Using use super::* gives me weird vibes but looking into the rust book and the rest of trin it seems like a normal practice so it looks good.

✔️

ethportal-peertest/src/cli.rs
```diff
@@ -1,68 +1,81 @@
...
         long = "listen-port",
         help = "The UDP port to listen on."
     )]
     pub listen_port: u16,
 
-    #[structopt(
+    #[arg(
         default_value = "ipc",
-        possible_values(&["http", "ipc"]),
+        value_parser = ["http", "ipc"],
```
e3243eric, line 26: I found `value_parser` was not for possible values. It will deal with the input string and compare with it. The above code will get an error when the input is "ipc", so this should be fixed.
I think a good way is to use `enum` to list all the possible values and `derive` by `clap::ValueEnum`.

e3243eric:
I agree that I should only `use` the necessary modules in tests.

ogenev:
Sorry for the misunderstanding, but our plan is to depreciate CLI in ethportal-peertest.

Closing this PR as it becomes obsolete after https://github.com/ethereum/trin/pull/666 is merged.

Issue #539, opened 2023-02-13
ethportal-api: Update `NodeId` type


ogenev:
Right now we have our own custom type for NodeId in `ethportal-api`:

https://github.com/ethereum/trin/blob/7c475988d20bfe2d3461b5bfe1b812647dea8d69/ethportal-api/src/types/discv5.rs#L11

Instead of this, I think it is more robust if we wrap the `NodeId` type over  `discv5::NodeId` and implement all necessary traits to the newtype.

KolbyML:
I will work on this if nobody else is

PR #669, opened 2023-04-15
replace wrapped NodeId with native discv5 type (except for caveat in description)

KolbyML:
### What was wrong?
Fix https://github.com/ethereum/trin/issues/539 

### How was it fixed?
Update discv5 to ``v0.3.0``

Use discv5 node_id type and remove our encapsulated one

#### Known issue
We can't fully remove it till we add ``#[serde(with = "SerHex::<StrictPfx>")]`` to ``sigp/enr`` it is used for recursive query tracing. It is a issue with serialization with hashmaps, if we don't have this flag checked.

https://github.com/sigp/enr/pull/34

^ here is a PR I made to resolve it.

I think we can still merge this even though we still have this 1 reliance, since our goal is to only have to use ``discv5::enr::NodeId`` and not maintain our own and this is one step closer. In the next version of discv5 we should be able to fully remove our wrapper.
### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/src/types/node_id.rs
+++ b/ethportal-api/src/types/node_id.rs
@@ -1,7 +1,18 @@
-use crate::types::distance::{Metric, XorMetric};
-use discv5::enr::NodeId;
+use super::enr::Enr;
+use serde::{Deserialize, Serialize};
+use stremio_serde_hex::{SerHex, StrictPfx};
 
-pub fn generate_random_node_id(target_bucket_idx: u8, local_node_id: NodeId) -> NodeId {
+use discv5::enr::NodeId as EnrNodeId;
+
+use super::distance::{Metric, XorMetric};
+
+type RawNodeId = [u8; 32];
+
+/// Generate random NodeId based on bucket index target and a local node id.
+/// First we generate a random distance metric with leading zeroes based on the target bucket.
+/// Then we XOR the result distance with the local NodeId to get the random target NodeId
+// TODO: We should be able to make this generic over a `Metric`.
+pub fn generate_random_node_id(target_bucket_idx: u8, local_node_id: EnrNodeId) -> EnrNodeId {
     let distance_leading_zeroes = 255 - target_bucket_idx;
     let random_distance = crate::utils::bytes::random_32byte_array(distance_leading_zeroes);
 
@@ -10,4 +21,27 @@ pub fn generate_random_node_id(target_bucket_idx: u8, local_node_id: NodeId) -> 
     let raw_bytes = raw_node_id.big_endian();
 
     raw_bytes.into()
+}
+
+/// Discv5 NodeId
+// TODO: Wrap this type over discv5::NodeId type
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
+
+impl From<&Enr> for NodeId {
+    fn from(enr: &Enr) -> Self {
+        NodeId(enr.node_id().raw())
+    }
+}
+
+impl From<Enr> for NodeId {
+    fn from(enr: Enr) -> Self {
+        NodeId(enr.node_id().raw())
+    }
+}
+
+impl From<EnrNodeId> for NodeId {
+    fn from(value: EnrNodeId) -> Self {
+        NodeId(value.raw())
+    }
 }
--- a/ethportal-api/src/types/node_id.rs
+++ b/ethportal-api/src/types/node_id.rs
@@ -1,115 +1,13 @@
-use super::enr::Enr;
-use serde::{Deserialize, Serialize};
-use std::ops::Deref;
-use stremio_serde_hex::{SerHex, StrictPfx};
+use crate::types::distance::{Metric, XorMetric};
+use discv5::enr::NodeId;
 
-use discv5::enr::NodeId as EnrNodeId;
+pub fn generate_random_node_id(target_bucket_idx: u8, local_node_id: NodeId) -> NodeId {
+    let distance_leading_zeroes = 255 - target_bucket_idx;
+    let random_distance = crate::utils::bytes::random_32byte_array(distance_leading_zeroes);
 
-use super::distance::{Metric, XorMetric};
+    let raw_node_id = XorMetric::distance(&local_node_id.raw(), &random_distance);
 
-type RawNodeId = [u8; 32];
+    let raw_bytes = raw_node_id.big_endian();
 
-/// Discv5 NodeId
-// TODO: Wrap this type over discv5::NodeId type
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
-pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
-
-impl NodeId {
-    pub fn random() -> Self {
-        Self(rand::random())
-    }
-
-    pub fn raw(self) -> RawNodeId {
-        *self.deref()
-    }
-
-    /// Generate random NodeId based on bucket index target and a local node id.
-    /// First we generate a random distance metric with leading zeroes based on the target bucket.
-    /// Then we XOR the result distance with the local NodeId to get the random target NodeId
-    // TODO: We should be able to make this generic over a `Metric`.
-    pub fn generate_random_node_id(target_bucket_idx: u8, local_node_id: NodeId) -> NodeId {
-        let distance_leading_zeroes = 255 - target_bucket_idx;
-        let random_distance = crate::utils::bytes::random_32byte_array(distance_leading_zeroes);
-
-        let raw_node_id = XorMetric::distance(&local_node_id.raw(), &random_distance);
-
-        let raw_bytes = raw_node_id.big_endian();
-
-        raw_bytes.into()
-    }
-}
-
-impl From<RawNodeId> for NodeId {
-    fn from(item: RawNodeId) -> Self {
-        NodeId(item)
-    }
-}
-
-impl From<EnrNodeId> for NodeId {
-    fn from(value: EnrNodeId) -> Self {
-        NodeId(value.raw())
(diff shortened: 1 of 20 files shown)


trin-types/src/node_id.rs
```diff
@@ -1,24 +1,63 @@
...
-// TODO: Wrap this type over discv5::NodeId type
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NodeId(discv5::enr::NodeId);
 
-impl From<RawNodeId> for NodeId {
-    fn from(item: RawNodeId) -> Self {
-        NodeId(item)
+impl NodeId {
```
ogenev, line 10: Let's implement [Deref](https://doc.rust-lang.org/std/ops/trait.Deref.html) trait instead of explicitly implementing those methods.

trin-types/src/node_id.rs
```diff
@@ -1,24 +1,63 @@
...
+use trin_utils::bytes::{hex_decode, hex_encode};
 
 type RawNodeId = [u8; 32];
 
 /// Discv5 NodeId
-// TODO: Wrap this type over discv5::NodeId type
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
-pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct NodeId(discv5::enr::NodeId);
```
ogenev, line 8:
```suggestion
pub struct NodeId(pub discv5::enr::NodeId);
```

trin-types/src/node_id.rs
ogenev, line 1: It is also useful to implement `From<discv5::enr::NodeId) for NodeId` so we can convert between these types with ease.

ogenev (commented):
Let's wait for https://github.com/ethereum/trin/pull/666 to land first as there is some overlap with the `NodeId` move to `trin_type`

**Note:** This is a temporal solution until discv5 updates their enr  crate version to include this update: https://github.com/sigp/enr/pull/27. Then we can use the native discv5 NodeId without a wrapper.

trin-types/src/node_id.rs
KolbyML: image
Seems like my editor isn't happy with that

trin-types/src/node_id.rs
KolbyML: image
image
How would that work with Rust? Sorry I must not be familiar enough with how Deref works, but that is what happens when I attempt to use this for example

trin-types/src/node_id.rs
KolbyML: resolved

trin-types/src/node_id.rs
ogenev: On line 30 you want to return `Self` which is the wrapped structure NodeId, so it should be something like:

```rust          Ok(NodeId(discv5::enr::NodeId::parse(result.as_slice()).expect("Could not parse decoded NodeID")))
```

trin-types/src/node_id.rs
KolbyML: Ty for the pointers and the help resolved

KolbyML:
@ogenev ok your issue's have been resolved

ogenev:
> @ogenev ok your issue's have been resolved

Need to rebase this before the final review.

trin-history/src/jsonrpc.rs
```diff
@@ -171,7 +171,7 @@ impl HistoryRequestHandler {
                     let _ = request.resp.send(response);
                 }
                 HistoryEndpoint::RecursiveFindNodes(node_id) => {
-                    let node_id = discv5::enr::NodeId::from(node_id.0);
+                    let node_id = discv5::enr::NodeId::from(node_id.0.raw());
```
ogenev, line 174: I think we can remove this conversion and use `node_id.into()` on the next.

KolbyML: solved

ogenev:
@njgheorghita I think we used some node id hacks to convert between values, could you remind me where those are, and should we update them in this PR?

trin-types/src/node_id.rs
```diff
@@ -39,41 +34,56 @@ impl NodeId {
...
+
+impl<'de> Deserialize<'de> for NodeId {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let result: String = Deserialize::deserialize(deserializer)?;
+        let result = hex_decode(&result).map_err(serde::de::Error::custom)?;
+        Ok(NodeId(
+            discv5::enr::NodeId::parse(result.as_slice()).expect("Could not parse decoded NodeID"),
```
njgheorghita, line 55: We should return an error here instead of panicking with `expect`

trin-types/src/node_id.rs
```diff
@@ -1,26 +1,21 @@
...
-// TODO: Wrap this type over discv5::NodeId type
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
-pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct NodeId(pub discv5::enr::NodeId);
 
 impl NodeId {
     pub fn random() -> Self {
-        Self(rand::random())
+        Self(discv5::enr::NodeId::random())
```
njgheorghita, line 14: I prefer to see all type imports at the top of the file. I'd go with something like `use discv5::enr::NodeId as Discv5NodeId;`.  However, an exception would be something like line 53, where I think it's fine to have an in-line import inside `map_err(..)`

njgheorghita (changes requested):
@ogenev I'm not sure I remember what kind of hacks you're referencing...

Imo, the purpose of this issue should really be to consolidate all of our usage of `NodeId` into a single type. This will make the transition to using discv5's native `NodeId` much simpler once it implements serde natively. @KolbyML If you can go through the repo and convert all usage of `NodeId` to use `trin_types::node_id::NodeId` then it will be much easier to have a complete understanding of the consequences of the changes implemented to the type in this pr

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -81,16 +81,19 @@ pub async fn test_trace_recursive_find_content(peertest: &Peertest) {
...
 
-    let query_origin_node: NodeId = peertest.nodes[0].enr.node_id().into();
-    let node_with_content: NodeId = peertest.bootnode.enr.node_id().into();
+    let query_origin_node: NodeId = peertest.nodes[0].enr.node_id();
+    let node_with_content: NodeId = peertest.bootnode.enr.node_id();
 
     // Test that `origin` is set correctly
     let origin = trace.origin;
-    assert_eq!(origin, query_origin_node);
+    assert_eq!(origin, ethportal_api::NodeId::from(query_origin_node));
```
ogenev, line 89: I think if we replace the `NodeId`  type in  `ethportal-api/src/types/query_trace.rs`, we will not need this conversion.

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -81,16 +81,19 @@ pub async fn test_trace_recursive_find_content(peertest: &Peertest) {
...
 
     // Test that `received_content_from_node` is set correctly
     let received_content_from_node = trace.received_content_from_node.unwrap();
-    assert_eq!(node_with_content, received_content_from_node);
+    assert_eq!(
+        ethportal_api::NodeId::from(node_with_content),
```
ogenev, line 94: Same comment as above,

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -158,6 +161,6 @@ pub async fn test_trace_recursive_find_content_local_db(peertest: &Peertest) {
         origin
     );
 
-    let expected_origin_id: NodeId = peertest.bootnode.enr.node_id().into();
-    assert_eq!(expected_origin_id, origin);
+    let expected_origin_id: NodeId = peertest.bootnode.enr.node_id();
+    assert_eq!(ethportal_api::NodeId::from(expected_origin_id), origin);
```
ogenev, line 165: Same comment as above,

ethportal-api/src/types/node_id.rs
```diff
@@ -9,53 +8,25 @@ use super::distance::{Metric, XorMetric};
...
 
-impl From<NodeId> for EnrNodeId {
-    fn from(val: NodeId) -> Self {
-        EnrNodeId::new(val.deref())
-    }
-}
+/// Discv5 NodeId
+// TODO: Wrap this type over discv5::NodeId type
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
+pub struct NodeId(#[serde(with = "SerHex::<StrictPfx>")] pub RawNodeId);
```
ogenev, line 29: We don't need this structure anymore, we can use `discv5::enr::NodeId`.

KolbyML: https://github.com/sigp/enr/pull/34 here I list why we need it only for tracing. Without ``#[serde(with = "SerHex::<StrictPfx>")`` ``json!(Hashmap<NodeId, xyz>) will panic. This is used in the query tracing code.

If the concensus is to just wait till the next discv5 till this is resolved, that is fine. I just thought we could do most of the transition.

KolbyML: I updated my write up with the information as well, https://github.com/ethereum/trin/pull/669#issue-1669551007 . It is unfortunate.

ogenev: I see, ok, no need to wait for IMO but we need to create an issue after this is merged.

ogenev (approved):
Looks good :+1:

PR #44, opened 2021-07-18
Add StructOpt for CLI parsing

ogenev:
As mentioned in https://github.com/carver/trin/issues/32, it is good idea to use `StructOpt` for parsing cli arguments into `TrinConfig`.

Fixes #32

PR #671, opened 2023-04-15
change web3_ipc_path from String to PathBuf

KolbyML:
### What was wrong?
Fixes todo added in this https://github.com/ethereum/trin/pull/44 PR
### How was it fixed?
Instead of using String/&str I converted the types of Path or PathBuf where it made sense. Basically just using PathBuf instead of using a String

### DIFF
--- a/ethportal-peertest/src/lib.rs
+++ b/ethportal-peertest/src/lib.rs
@@ -2,6 +2,7 @@ pub mod constants;
 pub mod scenarios;
 
 use std::net::{IpAddr, Ipv4Addr};
+use std::path::PathBuf;
 use std::{thread, time};
 
 use ethportal_api::jsonrpsee::server::ServerHandle;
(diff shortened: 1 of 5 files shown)


ogenev (commented):
I would change all path function arguments to accept the trait `AsRef<Path>` which will allow `Path`, `PathBuf`, and strings.

KolbyML:
> I would change all path function arguments to accept the trait `AsRef<Path>` which will allow `Path`, `PathBuf`, and strings.

Should we do that though for example 
image
this has to use a Box wouldn't that be unidealistic?

ogenev:
> this has to use a Box wouldn't that be unidealistic?

We use `Box` because the function arguments must be with known size on compile time; in this case, a trait is with unknown size and that's why we wrap it in a Box which is a smart pointer.

We can also use generics, this would look like:

```rust
fn foo<T: AsRef<Path>>(path: T)
```

ogenev (approved):
Looks good :+1: .

Issue #618, opened 2023-03-30
If there's a CLI parsing problem, then return a Clap:Error, instead of panicking


carver:
It's better to have a clean exit. It also makes testing easier.

This means:
- updating the tests to detect an error from `TrinConfig::new_from()`
- replacing all `panic!` calls with returning a `Clap::Error`

That probably looks something like:
```diff
         match config.web3_transport {
             Web3TransportType::HTTP => match &config.web3_ipc_path[..] {
                 DEFAULT_WEB3_IPC_PATH => {}
-                _ => panic!("Must not supply an ipc path when using http protocol for json-rpc"),
+                _ => {
+                    return clap::Error::with_description(
+                        "Must not supply an ipc path when using http protocol for json-rpc",
+                        clap::ErrorKind::ArgumentConflict,
+                    );
+                }
             },
```

KaydenML:
I want to do this issue

PR #675, opened 2023-04-18
Clap:Error instead of panicking

KaydenML:
### What was wrong?
Fixs https://github.com/ethereum/trin/issues/618
### How was it fixed?
I changed all the panic! to use Clap::Error instead. I also added .exit() to the end of it because that is what panic does but this want displayed in the issue, probably because it was assumed it exits by default.

for ErrorKind I matched it to what I thought best suited the error 
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3823,6 +3823,7 @@ version = "0.1.0"
 dependencies = [
  "async-trait",
  "chrono",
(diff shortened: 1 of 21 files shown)


KaydenML:
i made a new pr that actually fix's the problem

Issue #358, opened 2022-06-27
`--no--stun` doesn't work for all test cases


ogenev:
Adding a `--no-stun` flag allows us to deal with the unstable [find_assigned_ip](https://github.com/ethereum/trin/blob/5a52c9068cd9209b4e7d87a86dada952c9a51f51/trin-core/src/socket.rs#L36) method and send a discv5 request with empty ENR IP address and port.

`Discv5` doesn't add the node in the routing table of the receiver when the ENR address is empty.

This worked fine for now in [self_peertest](https://github.com/ethereum/trin/blob/master/tests/self_peertest.rs) but it prevents us to add new test cases where we expect the test node to be added to the remote peer's overlay routing table.

One possible solution is to try to make the `find_assigned_ip` stable. Any more opinions?

ogenev:
I think another reasonable solution is to depreciate [find_assigned_ip](https://github.com/ethereum/trin/blob/5a52c9068cd9209b4e7d87a86dada952c9a51f51/trin-core/src/socket.rs#L36) and use `127.0.0.1:port` as external socket address in tests.

carver:
> I think another reasonable solution is to depreciate [find_assigned_ip](https://github.com/ethereum/trin/blob/5a52c9068cd9209b4e7d87a86dada952c9a51f51/trin-core/src/socket.rs#L36) and use `127.0.0.1:port` as external socket address in tests.

Yeah, I think forcing an address/port in the test is an option that's worked well-enough, like in:
https://github.com/ethereum/trin/blob/7ba304a08784bbd6ed195d304775657b012365ed/trin-core/tests/overlay.rs#L112-L120

or

https://github.com/ethereum/trin/blob/7ba304a08784bbd6ed195d304775657b012365ed/ethportal-peertest/src/lib.rs#L67-L83

Note that we're not forced to use 127.0.0.1 here, as long as we grab the IP once at the beginning of the test and then reuse it.

PR #680, opened 2023-04-19
depreciate find_assigned_ip

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/issues/358
### How was it fixed?
By depreciating ``find_assigned_ip()``. ``find_assigned_ip()`` was only still used for this one test case and is no longer part of anything else so if using 127.0.0.1 is a solution to the problem it doesn't make much sense keeping the function around. This also lowers the dependency count which is nice.
### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8,8 +8,8 @@ version = "0.11.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
 dependencies = [
- "lazy_static 1.4.0",
- "regex 1.7.3",
+ "lazy_static",
+ "regex",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


ogenev (approved):
:+1: for depreciating the unstable `find_assigned_ip`.

carver:
Just a vocabulary note that a "deprecated" API is one that is *still in* the project, but includes a warning that it will be removed soon. `find_assigned_ip()` was removed by this PR, not deprecated. (and I have no objections to that 👍🏻 )

KolbyML:
> Just a vocabulary note that a "deprecated" API is one that is _still in_ the project, but includes a warning that it will be removed soon. `find_assigned_ip()` was removed by this PR, not deprecated. (and I have no objections to that 👍🏻 )

Thank you for letting me know, I didn't know the nuance to the word, I always thought it was a synonym for remove. Going forward I will I will keep that in mind so I can more accurately betray what my commit is doing in the description. I really appreciate feedback!

PR #681, opened 2023-04-20
embed pre-merge master accumulator

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/issues/466
### How was it fixed?
by utilizing the [rust-embed](https://crates.io/crates/rust-embed) crate
### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-validation/src/accumulator.rs
+++ b/trin-validation/src/accumulator.rs
@@ -1,3 +1,4 @@
+use rust_embed::RustEmbed;
 use std::path::PathBuf;
 
 use anyhow::anyhow;
@@ -26,6 +27,11 @@ use trin_utils::bytes::hex_decode;
 /// List of historical epoch accumulator merkle roots preceding current epoch.
 pub type HistoricalEpochRoots = VariableList<tree_hash::Hash256, typenum::U131072>;
 
+#[derive(RustEmbed)]
+#[folder = "src/assets/"]
+#[prefix = "validation_assets/"]
+struct TrinValidationAssets;
+
 /// SSZ Container
 /// Primary datatype used to maintain record of historical and current epoch.
 /// Verifies canonical-ness of a given header.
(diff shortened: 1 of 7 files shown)


trin-validation/src/accumulator.rs
```diff
@@ -37,11 +43,8 @@ pub struct MasterAccumulator {
 impl MasterAccumulator {
     /// Load default trusted master acc
     pub fn try_from_file(master_acc_path: PathBuf) -> anyhow::Result<MasterAccumulator> {
-        let mut path = PathBuf::from(env!("CARGO_MANIFEST_DIR"));
-        path.push(master_acc_path);
-        let raw = std::fs::read(&path)
-            .map_err(|_| anyhow!("Unable to find master accumulator at path: {:?}", path))?;
-        MasterAccumulator::from_ssz_bytes(&raw)
+        let raw = TrinValidationAssets::get(&(*master_acc_path).display().to_string()[..]).unwrap();
```
ogenev, line 46: Let's handle this unwrap too, and map the error to anyhow (example in the line below). We don't want naked unwraps in our code, except for tests.

KolbyML: resolved

trin-validation/src/accumulator.rs
```diff
@@ -37,11 +43,16 @@ pub struct MasterAccumulator {
...
-        MasterAccumulator::from_ssz_bytes(&raw)
+        let raw = match TrinValidationAssets::get(&(*master_acc_path).display().to_string()[..]) {
+            Some(raw) => Ok(raw),
+            None => Err({
+                anyhow!(
+                    "Unable to find master accumulator at path: {:?}",
+                    master_acc_path
+                )
+            }),
+        }?;
```
ogenev, line 54:
```suggestion
        let raw = TrinValidationAssets::get(&(*master_acc_path).display().to_string()[..])
            .map_err(|_| anyhow!("Unable to find master accumulator at path: {master_acc_path:?}"))?;
```

KolbyML: .map_err cant be used on an option apparently it can only be used on a result.

KolbyML: The match cases does what map_err() would do though

ogenev: Ahh ok, let's use  [ok_or](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or), this will transform the option to Result and we can return the error with `?`.

KolbyML: that is a very cool function I didn't know about ty for letting me know :). Resolved

ogenev (approved):
Looks good! :+1:

PR #683, opened 2023-04-20
Implement non-implemented discv5 api endpoints for jsonrpsee

KolbyML:
### What was wrong?
quite a few api endpoints aren't implemented yet, this was discussed in the meeting on 4/20/2023, so I decided to do it for discv5, hopefully history when I get time
### How was it fixed?
By calling the discv5 functions (the wrapper in trin I did have to add a quick one that wasn't added yet) for the endpoints

I removed enr_seq from the api endpoints because after what I said and deme said it was removed from the spec.

I implemented ``addEnr``, ``getEnr``, ``deleteEnr``, ``talkReq``, ``updateNodeInfo``, ``recursiveFindNodes``

### edit
I am going to wait till sigp/discv5 makes there new release it should be done after Emilia's Nat PR goes through it is pretty big so I am not sure how long that will take, but it shouldn't be forever.

Here is a issue outlining my PR https://github.com/ethereum/trin/issues/689 and the 3 requests I have left to add
### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/src/discv5.rs
+++ b/ethportal-api/src/discv5.rs
@@ -1,7 +1,9 @@
 use crate::types::discv5::{NodeInfo, RoutingTableInfo};
 use crate::types::enr::Enr;
 use discv5::enr::NodeId;
+use discv5::service::Pong;
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use crate::types::portal::FindNodesInfo;
 
 /// Discv5 JSON-RPC endpoints
 #[rpc(client, server, namespace = "discv5")]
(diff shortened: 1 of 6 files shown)


KolbyML:
I can rebase them into 1 comment. I tried to get it to detect the discv5_rpc was a rename in the first comment it shows it but when you view the PR's second comment it still shows that discv5 was destroyed and _rpc is brand new which I am not sure why. Especially since locally I don't have that issue so maybe it is a github detection issue? I am not sure why.

rpc/src/discv5.rs
ogenev, line 1: Let's do it in a separate PR if we want to rename this file. Including this here makes tracking the changes in the file harder and we want to minimize the reviewer's overheat.

KolbyML: Resolved

njgheorghita (changes requested):
Tests for all of these endpoints should be added to `ethportal-peertest/src/scenarios/basic.rs`

KolbyML:
Closing this as it is stale and not that high of a priority here is a issue with for https://github.com/ethereum/trin/issues/689 if someone wants to fix it in the future

PR #685, opened 2023-04-21
Metrics on linux

### LINKED ISSUE #684: Get metrics working on linux, with example dashboard
The metrics setup only worked on Mac, so identify, fix and document the appropriate places to get it working in both.

Additionally, build an example dashboard with the latest metric names.

carver:
### What was wrong?

Fix #684 

### How was it fixed?

Mostly documentation in the trin-cli readme, by adding a few important flags, and related explanations.

Bonus fixes:
- Some storage metrics were being dropped. Fixed
- Radius wasn't increasing on relaunch, even when capacity was available. Add test and the follow-up fix for #624 
 
### To-Do

- [ ] Clean up commit history
- [x] Create dashboard automatically with new template


carver:
I'm happy with the dashboard I built (at least as a starting point), I'm just trying to get the cli tool to auto-populate it into grafana.

I'm running into some funky issue when trying to push the dashboard, where the cli tool is reporting a 404 from grafana when contacting the `api/dashboards/db/` URL, but it works fine on the old template. I don't understand why changing the template could possibly make the url 404. So that will take a little more digging...

mrferris:
> I'm running into some funky issue when trying to push the dashboard, where the cli tool is reporting a 404 from grafana when contacting the api/dashboards/db/ URL, but it works fine on the old template. I don't understand why changing the template could possibly make the url 404.

For posterity for anyone hitting this in the future: the root level `id` value in the dashboard's `.json` should be `null`, otherwise grafana thinks you're trying to update a dashboard that it can't find.

mrferris (approved):
Looks good, works on mac too. 

One thing I noticed is that the routing table section in the original `trin` dashboard is broken with a `415 unsupported media type`. Most likely an update to grafana broke that plugin, so I think we should not try to do that in grafana and just do it in glados instead. So I'd lean toward removing that original dashboard altogether.

carver:
> One thing I noticed is that the routing table section in the original `trin` dashboard is broken with a `415 unsupported media type`. Most likely an update to grafana broke that plugin

Ah, my guess was that it was the switch to jsonrpsee, that it didn't like the way the request was formed.

But yeah, I'm fine dropping the old dashboard. I guess I'll leave in the multi-dashboard logic, because I think we'll want more again soon.

PR #688, opened 2023-04-21
Switch from StructOpt to Clap in trin-cli

e3243eric:
### What was wrong?
Related to #627. A tiny PR of #651.

### How was it fixed?
Switched from StructOpt to Clap in trin-cli.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-cli/src/main.rs
+++ b/trin-cli/src/main.rs
@@ -1,4 +1,5 @@
 pub mod dashboard;
+use clap::{Args, Parser, Subcommand};
 
 #[cfg(unix)]
 use std::os::unix::net::UnixStream;
@@ -9,7 +10,6 @@ use std::path::{Path, PathBuf};
 
 use ethereum_types::H256;
 use serde_json::value::RawValue;
-use structopt::StructOpt;
 use thiserror::Error;
 
 use dashboard::grafana::{GrafanaAPI, DASHBOARD_TEMPLATES};
@@ -17,8 +17,8 @@ use ethportal_api::{BlockBodyKey, BlockHeaderKey, BlockReceiptsKey, HistoryConte
 use trin_types::cli::DEFAULT_WEB3_IPC_PATH;
 use trin_utils::bytes::hex_encode;
 
-#[derive(StructOpt)]
-#[structopt(
+#[derive(Parser, Debug, PartialEq)]
+#[command(
     name = "trin-cli",
     version = "0.0.1",
     author = "Ethereum Foundation",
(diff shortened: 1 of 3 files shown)


trin-cli/src/main.rs
```diff
@@ -9,92 +10,94 @@ use std::path::{Path, PathBuf};
...
+    #[arg(default_value = "http://host.docker.internal:9090")]
     prometheus_address: String,
 }
 
-fn main() -> Result<(), Box<dyn std::error::Error>> {
-    match Trin::from_args() {
-        Trin::JsonRpc(rpc) => json_rpc(rpc),
-        Trin::EncodeKey(content_key) => encode_content_key(content_key),
-        Trin::CreateDashboard(dashboard_config) => create_dashboard(dashboard_config),
+fn main() {
```
KolbyML, line 95:
```suggestion
fn main() -> Result<(), Box<dyn std::error::Error>> {
    match Trin::parse() {
        Trin::JsonRpc(rpc) => json_rpc(rpc),
        Trin::EncodeKey(content_key) => encode_content_key(content_key),
        Trin::CreateDashboard(dashboard_config) => create_dashboard(dashboard_config),
    }
```

We should only use unwrap during tests is there any reason you added unwrap and removed the return?

If you want you can just add the Result back and do match Trin::parse() and the rest of the PR looks good

KolbyML (changes requested):

trin-cli/src/main.rs
```diff
@@ -9,92 +10,94 @@ use std::path::{Path, PathBuf};
...
+    let args = Trin::parse();
+    match args {
+        Trin::JsonRpc(rpc) => json_rpc(rpc).unwrap(),
+        Trin::EncodeKey(content_key) => encode_content_key(content_key).unwrap(),
+        Trin::CreateDashboard(dashboard_config) => create_dashboard(dashboard_config).unwrap(),
     }
 }
```
KolbyML, line 102:
```suggestion
fn main() -> Result<(), Box<dyn std::error::Error>> {
    match Trin::parse() {
        Trin::JsonRpc(rpc) => json_rpc(rpc),
        Trin::EncodeKey(content_key) => encode_content_key(content_key),
        Trin::CreateDashboard(dashboard_config) => create_dashboard(dashboard_config),
    }
}
```
We should only use unwrap during tests is there any reason you added unwrap and removed the return?

If you want you can just add the Result back and do match Trin::parse() and the rest of the PR looks good

KolbyML (changes requested):

KolbyML (approved):

PR #692, opened 2023-04-24
Add purge db executable to bridge docker image

njgheorghita:
### What was wrong?
We never added the `purge_db` script to the docker bridge image.
Also removed some unnecessary lines post #681 

### To-Do

- [x] Clean up commit history


njgheorghita:
Quick docker images fix here - moving ahead with the un-approved merge so I can get this up & running on the testnet

PR #694, opened 2023-04-25
Update trin-types dependency on trin-utils

njgheorghita:
### What was wrong?
To cut a release for `trin-types` we need to update the `trin-utils` dependency

### To-Do

- [x] Clean up commit history


njgheorghita:
another rage merge.. apologies this should be the last one...

PR #695, opened 2023-04-25
Bump trin-types and trin-utils dependencies

njgheorghita:
### What was wrong?
Update `trin-types` and `trin-utils` dependencies to use the published versions of the crate rather than directly linking to the project directories.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2164,8 +2164,8 @@ dependencies = [
  "serde",
  "serde_json",
  "tracing",
- "trin-types",
- "trin-utils 0.1.1-alpha.1",
+ "trin-types 0.1.1-alpha.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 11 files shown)


KolbyML:
> Not sure if I'm on board with this change.
> 
> Depending on published versions means that we need to always publish new versions after a change in the published crates.

I kind of agree with ogenev cause if we are refactoring code lets say normally we do this in pairs so we modify a few crates at once in a PR. So wouldn't that mean if we are modifying multiple crates or refactoring then we will either have to edit the toml's each time we want to see our change locally or publish a new version of the crate then update the toml to the released version?

Maybe crates work differently, but if it is the case above that could become pretty tedious for whoever is working on updates locally or the person who has to update the crate each time. Something like this seems like little work up front but really adds up over time, so if crates work differently and that isn't a concern I think it is fine. But if that tedious example above is the reality it will probably just slow people down and be a little spike people step over where it isn't annoying at first, but over time it becomes horrible.

Issue #581, opened 2023-03-07
Investigate VPN compatibility


njgheorghita:
As @perama-v points out [here](https://github.com/ethereum/trin/pull/548#discussion_r1122961745) - trin should be able to connect to the portal network from behind a VPN. However, in practice, myself and others have experienced difficulty connecting to the network from behind a VPN. This should be investigated, resolved, and docs updated to reflect new findings.

PR #698, opened 2023-04-26
Allow a query to finish if threshold is met with a subset of peers

perama-v:
### What was wrong?

Investigating
- #581

Noticed that when polling peers there is a `result_counter` for number of peers that have returned a result. 
This `result_counter` is reset if there is a peer that meets the following criteria:
- Waiting for peer
- Peer has not timed out
- Not at parallelism capacity

The `result_counter` is used to test if it has reached: "the number of closest peers that a query must obtain successful results for before it terminates. Kademlia paper specifices that this should be equal to k, the max number of entries in a k-bucket. Currently defaults to 20.". If so, `QueryState::Finished` is returned.

In other words, the function requires that all peers are not waiting before considering that the successful count is sufficient. 
However, there may be peers contacted than the desired `result_counter` threshold. 

It follows that if the counter is allowed to continue incrementing as successes are identified, enough may be observed to meet the threshold for `QueryState::Finished`.

### How was it fixed?

Remove the zero-ing of the counter.

When a sufficient number of peers have returned a successful result, the query can now be considered finished.

### Effect
For a node that never encounters pending/timed-out peers, this change may have no effect. For a node that experiences poor connectivity, the query may be prematurely terminated in a significant number of cases (termination was being triggered as a 1 of n threshold).

### To-Do

- ~[ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)~
- [x] Clean up commit history

### DIFF
--- a/portalnet/src/find/iterators/findnodes.rs
+++ b/portalnet/src/find/iterators/findnodes.rs
@@ -29,6 +29,7 @@ use std::{
     collections::btree_map::{BTreeMap, Entry},
     time::Instant,
 };
+use tracing::debug;
 
 #[derive(Debug, Clone)]
 pub struct FindNodeQuery<TNodeId> {
@@ -213,10 +214,9 @@ where
                         return QueryState::WaitingAtCapacity;
                     } else {
                         // The query is still waiting for a result from a peer and the
-                        // `result_counter` did not yet reach `num_results`. Therefore
-                        // the query is not yet done, regardless of already successful
+                        // `result_counter` did not yet reach `num_results`. There may
+                        // be more peers than `num_results` so continue to check
                         // queries to peers farther from the target.
-                        result_counter = None;
                     }
                 }
                 QueryPeerState::Succeeded => {
@@ -226,6 +226,12 @@ where
                         // closest peers, the query is done.
                         if *count >= self.config.num_results {
                             self.progress = QueryProgress::Finished;
+                            let peers = self.closest_peers.len();
+                            debug!(
+                                peers.queried = peers,
+                                result.threshold = self.config.num_results,
+                                "Query finished after meeting threshold"
+                            );
                             return QueryState::Finished;
                         }
                     }


perama-v:
Some charts without this PR (left) and with this PR (right), syncing a fresh node

Data accumulates faster:
data

More `PING` messages are received:
rx

More `PONG` messages are sent
tx

pipermerriam:
I think the "optimal" early return state will be when our  `k`  closest peers have have all answered our find_nodes request.  At that point, any "pending" requests are coming from peers that are further away than the `k` closest.  While it would still be possible for a peer that is further away to respond with a closer node that we haven't heard about yet, this would be a rare thing and likely only happen due to something like a new node having just joined the network.

I'm concerned about us exiting early if we haven't actually gotten responses from some of the closest peers... because those peers are the ones that know the most about the area of the DHT that we are exploring.

perama-v:
I looked at `self.closest_peers` in practice for find_nodes, and found that this value is not limited to k (range ~`[1, 70]`), so this is a list of peers, ordered closest-first rather than being a list of "the k closest peers". With an observed `self.config.num_results=20` threshold

For find_content, the `self.closest_peers` was in the range ~`[16, 27]`, with an observed `self.config.num_results=16` threshold.

## Interpretation 1
> I think the "optimal" early return state will be when our `k` closest peers have have all answered our find_nodes request. 

I think this is what happens. Rephrased, when k of the closest peers (an uncapped ordered collection) have responses, return a success.

Go through closest peers and check responses: https://github.com/ethereum/trin/blob/d11ac5c5e4fc00d913a5e2a08dfb86c30c376a30/portalnet/src/find/iterators/findnodes.rs#L188

For every new success, detect if `k` is reached: https://github.com/ethereum/trin/blob/d11ac5c5e4fc00d913a5e2a08dfb86c30c376a30/portalnet/src/find/iterators/findnodes.rs#L227

> And maybe we can call it a win if some majority subset of the closest have responded so that we don't hang waiting for a single response with 19/20 have already responded. I'd say something like waiting for 4/5ths to respond might be a good starting point.
### With the PR
Wait for k/k (e.g., 20/20) responses amongst n closest peers.
### Without the PR
Wait for k/k (e.g., 20/20) responses amongst n closest peers, but reset the counter when a peer is pending.

For example if there were 21 closest peers (20 responses, 1 pending), the order is important: That group has a 1/21 chance of success count reaching k threshold (`[success, success, success, ... success, pending]`), all other combinations reset the counter mid-iteration `[success, pending (reset count), success, ... success, success]` max counter = 19, despite 20 successes. Hence the requirement is "a contiguous set of k successes", which could be satisfied by peers with indices: `[0, 20)` or `[10, 30)`, or similar.


## Interpretation 2
Or are you saying that because the closest peers are ordered by distance, we want to wait until every single peer that is less than k in the list responds. In other words: Truncate the closest peer list to k entries and require that none are pending.

### Summary

Overall, my concern is that there are instances where a query with a few pending peers could result in the query being dropped completely from the querypool.

pipermerriam:
Ok, i think your approach is :+1: and does roughly what we need and improves the overall state of things.  I'm not convinced that I actually have a better solution anymore and am gonna leave this as a future thing to figure out if there's further room for optimization.

KolbyML:
Closing this as we ended up going with https://github.com/ethereum/trin/pull/726 option almost a year ago now

PR #697, opened 2023-04-25
Remove data radius from json-rpc ping

carver:
Spec changed here:
https://github.com/ethereum/portal-network-specs/pull/202

### What was wrong?

Fix #696 

### How was it fixed?

Remove the paramater. We just ignored it anyway, so it's pretty straightforward.

### To-Do

- [x] Clean up commit history

PR #700, opened 2023-04-26
Depend on released trin-types and trin-utils

carver:
### What was wrong?

We can't release ethportal-api with a relative dependency.

### How was it fixed?

Switch to depending on the released versions.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2175,8 +2175,8 @@ dependencies = [
  "serde",
  "serde_json",
  "tracing",
- "trin-types",
- "trin-utils 0.1.1-alpha.1",
+ "trin-types 0.1.1-alpha.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


carver:
So trin-bridge adds both trin-types and ethportal-api as dependencies. That causes a conflict. The most obvious way to fix it is to depend on the released version of trin-types everywhere. (Although that sucks, because it forces releases when changes would normally happen in convenient lockstep.

We can already see an example of that with #697 which could happen atomically before this change, but requires a trin-types release after this change.

I'll try to get the next trin-types release out today.

carver:
I'm actually not certain that we have to do this with trin-utils everywhere, too. So that's something else I want to experiment with.

ogenev:
Can we change only the `ethportal-api` and `trin-bridge` dependencies to the released versions?

njgheorghita:
@carver Seems we can close this post #723 ?

PR #701, opened 2023-04-27
spawn thread for each history request

### LINKED ISSUE #652: RPC is not serving requests concurrently

perama-v:
### What was wrong?

- closes #652 

### How was it fixed?

When a request is encountered, a thread is spawned to handle it. 
The main changes are:
- Network is placed inside a mutex
- Handling for each variant is moved to a dedicated function to reduce nesting

### To-Do

- [x] Clean up commit history


trin-history/src/jsonrpc.rs
```diff
@@ -1,245 +1,304 @@
...
+            network.read().await.overlay.bucket_entries(),
+        )),
+        HistoryEndpoint::RecursiveFindNodes(node_id) => {
+            recursive_find_nodes(network, node_id).await
+        }
+    };
+    let _ = request.resp.send(response);
+}
+
+// Constructs a JSON call for the method.
```
ogenev, line 85: Could we change those comments to doc comments and additional info for the endpoint, something like:

```suggestion
/// Constructs a JSON call for the RecursiveFindContent method.
```

ogenev (approved):
Looks good, thanks! :heart:

trin-history/src/jsonrpc.rs
perama-v: done

PR #706, opened 2023-04-28
Add support for Shanghai block bodies

njgheorghita:
### What was wrong?
Add support for Shanghai block bodies which contain a new `Withdrawals` field

### How was it fixed?
- Updated block body type to consist of 3 variations (`Legacy`, `Merge`, `Shanghai`)
- move transaction related types to separate `transaction.rs` module
- removed `EncodedBlockBodyParts` helper type, since it only added more confusion to what was going on
- experimented with using [superstruct](https://lib.rs/crates/superstruct) but I wasn't able to find a suitably non-complex way to leverage the library. not to say that it cannot be done just that I wasn't able to figure it out
- updated bridge node to support latest block body type & fetch withdrawals from pandaops

### To-Do

- [x] Clean up commit history


ogenev (approved):
Looks solid to me :ship:. I guess we can create an issue if somebody have enthusiasm to experiment further with [superstruct](https://crates.io/crates/superstruct).

PR #711, opened 2023-05-02
Re-exports trin-types's Enr type from `ethportal_api`

mrferris:
### What was wrong?

`trin-types`'s `Enr` type was not re-exported from `ethportal_api`, like `NodeId` was

### How was it fixed?

re-exported it

### To-Do

- [ ] Clean up commit history


njgheorghita (approved):
:ship:

PR #713, opened 2023-05-03
add guide for grafana setup

perama-v:
### What was wrong?

No guide for grafana setup without docker

### How was it fixed?

Guide added to user/monitoring section of book.

Includes setup of node-exporter system data so that this data may be used, but does not plug in/expose those data to the user.

### To-Do

- ~[ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)~
- [x] Clean up commit history

### DIFF
--- a/book/src/users/monitoring.md
+++ b/book/src/users/monitoring.md
@@ -37,3 +37,241 @@ du -sh
 ```sh
 htop
 ```
+
+## Metrics
+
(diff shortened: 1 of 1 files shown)


ogenev (approved):
Looks useful :100:

PR #714, opened 2023-05-03
Add beacon network endpoints and content items

ogenev:
### What was wrong?

Integrating the beacon network crate requires some preliminary primitives.

### How was it fixed?
1. Add content key and content value object for the beacon network.
- Those are minimal implementations and they are not tested against test vectors because test vectors still don't exist :).
The goal is to have a compiler valid  `BeaconContentKey` and `BeaconContentValue` which is required for the RPC and overlay integration with the upcoming beacon network crate.
2. Add beacon endpoints to trin endpoints. Those are required for the integration of the beacon network with the jsonrpsee RPC server.

### To-Do

- [ ] Clean up commit history


trin-types/src/consensus/header_proof.rs
```diff
@@ -29,12 +32,74 @@ pub struct BeaconChainBlockProof {
     pub slot: u64,
 }
 
-#[derive(Debug, PartialEq, Clone, Default)]
-pub struct HistoricalRoots(VariableList<tree_hash::Hash256, typenum::U16777216>);
+type HistoricalSummary = tree_hash::Hash256;
```
njgheorghita, line 35: Where does the name change `Root` -> `Summary` come from?

trin-types/src/consensus/header_proof.rs
```diff
@@ -29,12 +32,74 @@ pub struct BeaconChainBlockProof {
...
+type HistoricalSummaries = VariableList<HistoricalSummary, typenum::U16777216>;
 
-pub struct HistoricalRootsProof([H256; 5]);
+// HistoricalSummariesProof* = array[5, Digest]
```
njgheorghita, line 38: Is this commented out line supposed to be left in?

njgheorghita (approved):
Looks good! Though yeah, there's a definite need for test vectors in here asap

trin-types/src/consensus/header_proof.rs
ogenev: `HistoricalSummary` is introduced in Cappella with the `historical_summaries` as `BeaconState` field which replaces the pre-Capella `hisotrical_roots` field.

PR #719, opened 2023-05-05
Remove not used import and fix a windows clippy issue

KolbyML:
### What was wrong?
``use std::env`` is only used on none windows so it breaks clippy since it doesn't have a cfg.

Also we have 2 not needed use's which don't do anything
### How was it fixed?

I changed ``env::var(`` to ``std::env::var(`` then I removed not 2 not used uses
### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/719.fixed.md
+++ b/newsfragments/719.fixed.md
@@ -1,0 +1,1 @@
+Remove not used import and fix a windows clippy issue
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
:shipit:

PR #720, opened 2023-05-05
Add optimizations for CI tests

KolbyML:
### What was wrong?
The CI takes so long most of this is just building libraries!

The cause? The mass amount of libraries Rust projects use as a side affect of an easy to use package manager!!!
### How was it fixed?

1. By adding [Mozilla'a sccache](https://github.com/mozilla/sccache) library.
2. Splitting lint-build-test CI into lint, build, and test

### Before
image
### After
image

### To-Do

- [x] Clean up commit history


ogenev (approved):
Nice improvement! :+1:

PR #723, opened 2023-05-06
Move trin-types into a module of ethportal-api

### LINKED ISSUE #710: Rethinking Repository Structure, amid Release Planning
## Plan?

TODO

## Background thoughts

When looking at our repository structure, and how crates are organized, I had thought that we could "vendor" one crate into another. It's possible to to *locally* use the source crate, but at the time of publishing any crate, we need each of the dependencies to be published too.

This personally alters my idea about what is a reasonable organizational structure. Some examples of what comes up for me:
 - `trin-types` was our internal implementation of types, and ethportal-api is our "external" one, but now they both have to be published anyway. Maybe we should just recombine them, with modules for what's meant to be internal and external?
 - `rpc` obviously is not a good name for a crate to be published externally. Maybe it should just be a module inside another crate
 - maybe `trin-core` was a better structure

Things that probably don't need to change are:
- `trin-cli` might go away anyhow, or never be published?
- `trin-bridge` feels like a different enough tool to be published separately

Even if we trim down the total number of crates, we will have a few left, and probably don't want to manage them independently. We probably want to use tools to help us avoid mistakes around releases. Ideally those tools would help us:
- build release notes for each released crate
- track whether a crate needs to have its version bumped
- bump and increase the version of the crate in all its dependent crates, as needed
- only publish crates that have changed since the last publishing
- publish all crates in dependency order, from a single command

At least one tool requires [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) to do a good job of auto-generating release notes.

Some example tools that might help us do this, either by combining their usage, or using a single tool that does everything:
- https://crates.io/crates/cargo-workspaces
- https://crates.io/crates/release-plz
- https://github.com/Byron/gitoxide/tree/main/cargo-smart-release
- https://github.com/crate-ci/cargo-release

carver:
### What was wrong?

Fix #710  -- for now, ignoring the larger repository structure question, just looking at ethportal-api.

### How was it fixed?

Combine trin-types and the `bytes` module of trin-utils into ethportal-api, so that we don't need to publish them as dependencies.

### To-Do

- [x] Does it makes sense to keep the "types" module in ethportal-api or is it redundant after this change?
- [x] Merge trin-types module into types
- [x] Move necessary trin-utils components into ethportal-api
- [ ] Clean up commit history, lots of squashing


carver:
So we have this funny situation of a trin_types and types module. I was thinking of trin-types as the "internal" version, not intended for public consumption. But the more I look at it, the more I think there should be just a single types module.

The API exposes a bunch of things from the trin_types anyway. It's not obvious from a user perspective when to pull from one vs another. (it requires reasoning about what we, the maintainers, think "should" be public or not)

So I'm going to take a pass at merging them into just `types`. Then I think that's the last cleanup.

ethportal-api/src/discv5.rs
```diff
@@ -1,7 +1,7 @@
+use crate::types::discv5::{NodeInfo, RoutingTableInfo};
+use crate::types::enr::Enr;
+use crate::types::node_id::NodeId;
```
njgheorghita, line 3: This brings up the question of import order formatting... technically it's defined in the style guide to follow the style we used back in python land.. but we aren't that consistent across our codebase. It's definitely not a blocker for this pr, but maybe something we should discuss in our next sync to reach upon an agreed upon scheme.

ethportal-api/Cargo.toml
```diff
@@ -11,15 +11,47 @@ categories = ["cryptography::cryptocurrencies"]
 authors = ["https://github.com/ethereum/trin/graphs/contributors"]
 
 [dependencies]
+anyhow = "1.0.68"
```
njgheorghita, line 14: It seems like this pr is a good candidate to run [udeps](https://crates.io/crates/cargo-udeps) locally if you haven't already

njgheorghita (approved):
Well, I can't say I love it (i was somewhat partial to the order `trin-types` imposed on our workspaces), but I agree this is the simplest way forward...
Some thoughts...
- maybe add a line to ethportal-api's readme about why we placed these types there, and what kind of types belong in this workspace
- what to do about the `trin-types` version we've already released? is there any official deprecation that needs to happen.
- what to do about `trin-utils`... imo we should also just move it to `ethportal-api`

ethportal-api/src/discv5.rs
carver: Ah, I didn't do anything special here, just ran `cargo fmt --all`, which did reordering as it saw fit. What were you expecting here?

carver:
> Well, I can't say I love it (i was somewhat partial to the order `trin-types` imposed on our workspaces), but I agree this is the simplest way forward... Some thoughts...

Yeah I know. That's why I asked for your review, I figured you'd be the toughest on it. :)

> 
>     * maybe add a line to ethportal-api's readme about why we placed these types there, and what kind of types belong in this workspace

👍🏻 added something

> 
>     * what to do about the `trin-types` version we've already released? is there any official deprecation that needs to happen.

Yeah, since there's not any reasonable expectation that anyone started using it yet, I think we should just look for the biggest, reddest button that says delete or shut-down or whatever. I'll poke around the crates.io website.

> 
>     * what to do about `trin-utils`... imo we should also just move it to `ethportal-api`

Since trin-utils is no longer needed by ethportal-api, I think it's okay to leave it in place.

ethportal-api/src/discv5.rs
njgheorghita: AFAI, `cargo fmt --all` alphabetically orders every block of imports. So this is what I would expect to see. However, if we were "strictly" adhering to the style guide, I'd expect to see `jsonrpsee::...` on line 1 (since it's an external crate), then an empty line, then this block of `crate::...` (i.e. internal) imports. But, this style guide hasn't been enforced consistently across the codebase. This pr just reminded me that maybe it's something we want to discuss, or maybe it's not much of a problem at all

ethportal-api/README.md
```diff
@@ -46,5 +46,18 @@ async fn main() {
 }
 ```
 
+## Types
```
njgheorghita, line 49: Nice! Maybe it's worth updating to `## Types and Utils`? Technically we did move the hex utils into this crate and it seems to me like the same convention is worth adopting for shared utils

PR #725, opened 2023-05-09
Launch beacon portal network

ogenev:
### What was wrong?
We want to launch the beacon chain network

### How was it fixed?
This PR launches the beacon portal network **without** RPC support when "beacon" is included in the `networks` cli flag.

To enable RPC for the beacon network, we need first to refactor how JSON-RPC server endpoints are initialized based on the networks launched, which will follow in a separate PR.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portalnet/src/events.rs
+++ b/portalnet/src/events.rs
@@ -15,6 +15,8 @@ pub struct PortalnetEvents {
     pub history_overlay_sender: Option<mpsc::UnboundedSender<TalkRequest>>,
     /// Send overlay `TalkReq` to state network
     pub state_overlay_sender: Option<mpsc::UnboundedSender<TalkRequest>>,
+    /// Send overlay `TalkReq` to beacon network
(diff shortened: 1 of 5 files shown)


portalnet/src/events.rs
```diff
@@ -59,6 +63,18 @@ impl PortalnetEvents {
...
+                ProtocolId::Beacon => {
+                    match &self.beacon_overlay_sender {
+                        Some(tx) => {
+                            if let Err(err) = tx.send(request) {
+                                error!(
+                                    "Error sending discv5 talk request to beacon network: {err}"
+                                );
+                            }
+                        }
+                        None => warn!("Beacon event handler not initialized!"),
```
njgheorghita, line 75: Seems like this case might be worthy of an `error!`? eg a mistake has been made somewhere in the user's config / our architecture if a user tries to send a beacon msg and the subnetwork is not available

njgheorghita (approved):
👍

Issue #708, opened 2023-04-28
 `historyRecursiveFindNodes` endpoint takes too long to return a result


pipermerriam:
### What is wrong

The default http client for the portal JSON-RPC API provided by the `ethportal-api` crate ends up with a timeout during most calls to `historyRecursiveFindNodes`

### How can it be fixed

I'm guessing there's a default timeout parameter that needs to be extended to allow the http request to hang longer waiting for the response....

ogenev:
`ethportal-api` doesn't provide a default HTTP client,  it exposes API traits for the jsonrpsee HTTP clients to have information for all available endpoints.

Users of  `ethportal-api` are expected to create their own jsonrspee HTTP clients and they can set the timeout with [request_timeout](https://docs.rs/jsonrpsee-http-client/0.18.1/jsonrpsee_http_client/struct.HttpClientBuilder.html#method.request_timeout).

pipermerriam:
I still think there's a fix needed here since it appears that the client defaults aren't sufficient to use the client with `trin`.  Which suggests that maybe the thing that needs to be fixed is the response times on the `historyRecursiveFindNodes` endpoint so that it reliably returns in under the default request timeout?

KolbyML:
I am going to look into this since I am going through all the history RPC's anyways, @ogenev let me know if you want to look into it and that is fine.

I am going to implement history* ``add_enr``, ``get_enr``, ``delete_enr``, ``lookup_enr`` so I might as well also look into this

KolbyML:
@pipermerriam Could you send a few lines of the log with that error? I am interested in what you get for the "x: *" the star part.

PR #726, opened 2023-05-09
fix ``historyRecursiveFindNodes`` endpoint taking too long to return

KolbyML:
### What was wrong?
fix https://github.com/ethereum/trin/issues/708 supersedes https://github.com/ethereum/trin/pull/698 since this resolves the root of the problem
### How was it fixed?
By changing our ``query_peer_timeout`` from 10s to 2s.

Our recursiveFindNode code is heavily based off sigp/discv5 implementation. The biggest difference? Our ``query_peer_timeout`` is set 5 times higher. 

Our ``query_timeout`` is 60 seconds with 3 allowed async queries. The issue is if we ping 3 bad nodes that means 1/6th of our query time is gone. So worst case we can only make 18 queries. Where as if we match the timeout our implementation is based off of, of 2s our worst case is 90.

    /// The timeout after which a `QueryPeer` in an ongoing query is marked unresponsive.
    /// Unresponsive peers don't count towards the parallelism limits for a query.
    /// Hence, we may potentially end up making more requests to good peers. Default: 2 seconds.
    ^ Taken from the doc in sigp/dicv5 our base implementation for ours
   Having a lower timeout means we will accept there response if they do respond, we just get to make more calls to responsive nodes since they will no longer count towards are parallelism limit
   
   
Also having a high peer_timeout will prevent the call from returning since we wait have longstanding waiting requests 
```
                        // The query is still waiting for a result from a peer and the
                        // `result_counter` did not yet reach `num_results`. Therefore
                        // the query is not yet done, regardless of already successful
                        // queries to peers farther from the target.
                        result_counter = None;
```
                       
So this resolves that
### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6879,7 +6879,6 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
- "trin-beacon",
  "trin-bridge",
  "trin-history",
  "trin-state",
--- a/newsfragments/726.fixed.md
+++ b/newsfragments/726.fixed.md
@@ -1,0 +1,1 @@
+fix ``historyRecursiveFindNodes`` endpoint taking too long to return
--- a/portalnet/src/find/iterators/query.rs
+++ b/portalnet/src/find/iterators/query.rs
@@ -58,7 +58,7 @@ impl Default for QueryConfig {
         Self {
             parallelism: 3,
             num_results: 20,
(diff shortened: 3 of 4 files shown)


mrferris (approved):
> The issue is if we ping 3 bad nodes that means 1/6th of our query time is gone.

Good catch! Not sure what the reasoning for the 10 change was at the time. 

🚀

PR #727, opened 2023-05-10
Fix 2 issues with Monitoring docs

KolbyML:
### What was wrong?
The template prometheus config is missing 2 spaces so if a user follows the instructions Prometheus won't launch

Grafana isn't in the default package repos on Ubuntu 22.04
### How was it fixed?
I added 2 spaces so the config is valid yml and add 3 commands from https://grafana.com/docs/grafana/latest/setup-grafana/installation/debian/

### DIFF
--- a/book/src/users/monitoring.md
+++ b/book/src/users/monitoring.md
@@ -207,6 +207,9 @@ Node explorer uses port 9100 by default.
 
 Install
 ```sh
+sudo apt-get install -y apt-transport-https software-properties-common wget
+sudo wget -q -O /usr/share/keyrings/grafana.key https://apt.grafana.com/gpg.key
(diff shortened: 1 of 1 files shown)


book/src/users/monitoring.md
```diff
@@ -207,6 +207,9 @@ Node explorer uses port 9100 by default.
 
 Install
 ```sh
+sudo apt-get install -y apt-transport-https software-properties-common wget
+sudo wget -q -O /usr/share/keyrings/grafana.key https://apt.grafana.com/gpg.key
+echo "deb [signed-by=/usr/share/keyrings/grafana.key] https://apt.grafana.com stable main" | sudo tee -a /etc/apt/sources.list.d/grafana.list
 sudo apt install grafana
```
ogenev, line 213:
```suggestion
sudo apt update
sudo apt install grafana
```

ogenev (approved):

book/src/users/monitoring.md
KolbyML: resolved

PR #729, opened 2023-05-10
Make ``prometheus-address`` a parameter for create-dashboard and fix book for monitoring

KolbyML:
### What was wrong?
If we ran the instructed ``cargo run -p trin-cli -- create-dashboard`` it used the default docker address
### How was it fixed?
Making ``prometheus-address`` a setable argument and updating the book accordingly 

I also made it a bit more visible that ``admin`` are the username/password since I didn't read to the bottom and I assume others won't too?


book/src/users/monitoring.md
njgheorghita, line 1: Just a note that the yml is invalid in lines 95-113

KolbyML: image
I fixed the spacing in a different PR, @perama-v made this guide originally I have just been looking at fixing issues

njgheorghita: cool! which pr are you referencing?

KolbyML: https://github.com/ethereum/trin/pull/727

KolbyML:
I will just fix this in my PR removing cli, adding windows support, and moving create dashboard

book/src/users/monitoring.md
```diff
@@ -234,10 +234,10 @@ This will serve metrics over port 3000.
 
 Generate a grafana dashboard. From trin root directory:
 ```sh
-cargo run -p trin-cli -- create-dashboard
+cargo run -p trin-cli -- create-dashboard http://localhost:3000 admin admin http://127.0.0.1:9090
```
mrferris, line 237: I know this is closed, but just a note: I don't think that last argument is right, it's going to have grafana looking for the prometheus server within its own container rather than going out to the docker host and looking there. That's how it works on my machine anyway, does it work differently on yours?

KolbyML: @perama-v guide doesn't set it up with docker, it sets it as another user on the system.

Issue #730, opened 2023-05-10
Bring ``trin-cli`` back to life


KolbyML:
[Link this issue comment by Nick about reviving ``trin-cli``](https://github.com/ethereum/trin/issues/573#issuecomment-1452462454)

> We discussed this today & decided that it's worth investigating whether or not we should bring `trin-cli` back to life...
> 
> * A functional/extendable cli utility can improve the workflow
> * `trin-cli` is fairly useful when interacting with testnet nodes
> * We should be able to leverage `ethportal-api` jsonrpsee client functionality to simplify the existing `trin-cli` logic
> 
> Therefore... This issue remains valid. But, the solution should be to `"Mark the trin-cli crate as deprecated"`, and remove it from our documentation.
> 
> However, if anybody finds some time, it's worth a day/two to investigate whether we can simplify the existing code here & revive `trin-cli` in a functional, useful manner.
> 
> Possible sources of inspiration
> 
> * https://github.com/AgeManning/discv5-cli
> * ...

I want to work on this so I can use ``trin-cli`` to access the RPC since it would make it easier for me to play with ``portal_historyRecursiveFindNodes`` for testing/(quickly benchmarking) PR https://github.com/ethereum/trin/pull/726

KolbyML:
closing since I am going to discontinue CLI

PR #731, opened 2023-05-11
Revive ``trin-cli``, adding support for http, ipc, (and windows compilation)

KolbyML:
### What was wrong?
fix https://github.com/ethereum/trin/issues/730
### How was it fixed?
By removing our old json-rpc functions and replacing it with jsonrpcee html client and reth-ipc Unix Domain Sockets IPC (there are cfg tags so it will panic if they attempt to use it on windows since it isn't support) So for now windows will be HTTP client only
### To-Do

- [ ] Clean up commit history


KolbyML:
closing since I am going to discontinue CLI

PR #732, opened 2023-05-11
Implement Beacon network RPC endpoints

ogenev:
### What was wrong?
No Beacon network json-rpc endpoints were implemented.

### How was it fixed?
Implement Beacon network RPC endpoints.

**Note**: The RPC server is still not running those endpoints, because some additional refactoring is needed on the server side which will follow in a separate PR. Then we can also add tests for those endpoints and refactor the duplication in the history and beacon crates.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -4,74 +4,71 @@ use crate::types::portal::{
 };
 use crate::{NodeId, RoutingTableInfo};
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
-use trin_types::content_key::HistoryContentKey;
-use trin_types::content_value::{HistoryContentValue, PossibleHistoryContentValue};
+use trin_types::content_key::BeaconContentKey;
+use trin_types::content_value::{BeaconContentValue, PossibleBeaconContentValue};
 use trin_types::enr::Enr;
 
-/// Portal History JSON-RPC endpoints
+/// Portal Beacon JSON-RPC endpoints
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
Looks good, maybe in another PR we could add a fmt for package import granularity so packages from the same package are onelined, but that isn't in the scope of this PR and also in other rpc files so.

:+1:

ethportal-api/src/beacon.rs
```diff
@@ -0,0 +1,111 @@
...
+use trin_types::enr::Enr;
+
+/// Portal Beacon JSON-RPC endpoints
+#[rpc(client, server, namespace = "portal")]
+pub trait BeaconNetworkApi {
+    /// Returns meta information about overlay routing table.
+    #[method(name = "beaconRoutingTableInfo")]
+    async fn routing_table_info(&self) -> RpcResult<RoutingTableInfo>;
+
+    /// Returns meta information about overlay routing table.
```
njgheorghita, line 18: This docstring is incorrect

njgheorghita (approved):
Just a couple of thoughts... 
- until we figure out how multiple subnetworks share / don't share storage, paginate local content keys is not going to work as expected. do you have any ideas as to how we'll handle that?
- i'd definitely expect to see tests for these endpoints added to `ethportal-peertest` eventually, the sooner the better obvs

No recommended changes though, just something to keep in mind since we're still building out this subnetwork, :shipit:

PR #735, opened 2023-05-11
Add utp tx metric

njgheorghita:
### What was wrong?

after running in bridge mode for 15 mins...
Screen Shot 2023-05-11 at 3 36 49 PM
success: 12391
failure: 10284

### To-Do

- [ ] update default grafana dashboard template
- [ ] Clean up commit history


portalnet/src/metrics.rs
```diff
@@ -34,6 +34,14 @@ pub enum MessageDirectionLabel {
     Received,
 }
 
+/// Utp Transfer Direction Labels
+pub enum UtpTxDirectionLabel {
```
njgheorghita, line 38: Would it be better to re-use `MessageDirectionLabel` here?

portalnet/src/overlay_service.rs
```diff
@@ -1432,25 +1447,30 @@ where
...
             if let Err(err) = stream.write(&content_payload).await {
+                metrics.report_outbound_utp_tx(&protocol, false);
                 warn!(
                     %err,
                     cid.send,
                     cid.recv,
                     peer = ?cid.peer.client(),
                     "Error sending content over uTP connection"
                 );
+                return;
```
njgheorghita, line 1458: I added a `return` here since I'd like to avoid falsely reporting two failures for a single tx should `stream.shutdown()` also fail. But, I'm not sure if that's ideal behavior, and we still want to explicitly shutdown the stream even if the tx failed. @jacobkaufmann ?

mrferris: I would get rid of the return and just use a local `mut bool` to make sure you only report once.
```
let mut reported_stream_failure: bool = false;

...

if !reported_stream_failure { 
     metrics.report_outbound_utp_tx(&protocol, false);
     reported_stream_failure = true;
 }
```

I'm assuming that calling `shutdown` after a failed write is intentional and correct, if it's not then I agree with your `return`.

mrferris (commented):
👍📊

portalnet/src/overlay_service.rs
ogenev: I think we still want to shut down the stream even if the write fails. I would remove the return.

portalnet/src/overlay_service.rs
```diff
@@ -1432,25 +1447,30 @@ where
...
             }
 
             // close uTP connection
             if let Err(err) = stream.shutdown() {
+                metrics.report_outbound_utp_tx(&protocol, false);
```
ogenev, line 1463: I'm not sure if we want to report this as uTP stream failure, I'm not very familiar with Jacob's uTP design but I think this shutdown is to make the stream inactive and disconnected, we have already sent/received all packets?

carver: Hm, yeah I think this looks like a failure, but I agree it's not 100%. AFAICT, there's no current API to be totally sure. That's discussed a little bit here: https://github.com/ethereum/utp/issues/89

carver (approved):
This PR is an example of the optimistic green check we talked about in Boulder: there are things that I really care about getting resolved. I trust you to make the call about how to implement things, to ask me for more conversation if that would be helpful, and to make a reasonable decision about when to merge.

trin-bridge/src/main.rs
```diff
@@ -29,6 +29,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             .args(["--unsafe-private-key", key])
             .args(["--web3-transport", "http"])
             .args(["--web3-http-address", &web3_http_address])
+            .args(["--enable-metrics-with-url", "127.0.0.1:9101"])
```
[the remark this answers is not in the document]
njgheorghita: 👍 tracking in #783

PR #739, opened 2023-05-12
Make Trin compilable on Windows

KolbyML:
### What was wrong?
fixes Trin so that it can compile to windows
### How was it fixed?
add cfg and add a few panics to clearly define what is not supported by windows so it can compile

### DIFF
--- a/rpc/src/server_rpc.rs
+++ b/rpc/src/server_rpc.rs
@@ -3,6 +3,7 @@ use crate::{Discv5Api, HistoryNetworkApi, Web3Api};
 use ethportal_api::types::jsonrpc::request::HistoryJsonRpcRequest;
 use ethportal_api::{Discv5ApiServer, HistoryNetworkApiServer, Web3ApiServer};
 use portalnet::discovery::Discovery;
+#[cfg(unix)]
 use reth_ipc::server::Builder as IpcServerBuilder;
 use std::net::SocketAddr;
 use std::path::Path;
@@ -31,6 +32,7 @@ impl JsonRpcServer {
         Ok(handle)
     }
 
+    #[cfg(unix)]
     pub async fn run_ipc(
         ipc_path: Box<dyn AsRef<Path>>,
         discv5: Arc<Discovery>,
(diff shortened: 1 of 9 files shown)


rpc/src/server_rpc.rs
```diff
@@ -46,4 +48,13 @@ impl JsonRpcServer {
...
         Ok(handle)
     }
+
+    #[cfg(windows)]
+    pub async fn run_ipc(
+        _ipc_path: Box<dyn AsRef<Path>>,
+        _discv5: Arc<Discovery>,
+        _history_handler: mpsc::UnboundedSender<HistoryJsonRpcRequest>,
+    ) -> anyhow::Result<ServerHandle> {
+        panic!("Windows doesn't support Unix Domain Sockets IPC, use --web3-transport http")
```
ogenev, line 58: It seems Windows support unix domain sockets since 2018: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ and there is a rust [crate](https://github.com/haraldh/rust_uds_windows) for it?

KolbyML: We used to use that crate. Currently Reth uses tokio UnixStream so we can't use that crate. For tokio to add windows UDS it has to be added to MIO first I made a PR for it https://github.com/tokio-rs/mio/pull/1667 but there is no telling when it will be merged if at all. This is a tokio windows UDS by Microsoft  https://github.com/Azure/tokio-uds-windows but it hasn't been updated really in 5 years so it is hard to know if it is even maintained.

If we want we could maintain our own Windows Tokio UDS library I can have that ready by this week, but then we would also need to see if Reth would want to use that.

So yes Windows supports UDS (without pair), but it really isn't supported by any Rust libraries what suit Tokio.

So I think this is the best solution for now and maybe we can talk if maintaining a windows tokio UDS library and working with reth in the future. So my idea was merge this now and see if that is a overhead we would want to support in the future. (Unless we get lucky and get support from the Tokio team, but they seem uninterested)

ogenev: I see, supporting windows at this stage is not a priority, so I don't think we should maintain our Windows Tokio UDS library.

KolbyML: I don't think so either, which is why I don't think windows should support RPC

PR #741, opened 2023-05-12
Remove ``trin-cli`` and move grafana to ``trin-types``

KolbyML:
### What was wrong?
removes ``trin-cli``
### How was it fixed?
removing ``trin-cli``
moving grafana to ``trin-types`` as discussed in our meeting on 3/11/2023


mrferris (commented):
I don't think that this belongs in `trin-types`, my understanding was that we want to move the `create-dashboard` subcommand to `src/bin/`, so that it can be its own script like `purge_db.rs`.

ogenev (approved):
Looks good, I'm a little concerned about moving everything (including the dashboard) to `ethportal-api` but let's see how this will turn out.

PR #742, opened 2023-05-15
JSON-RPC server support for multiple networks

### LINKED ISSUE #728: Spawn RPC server endpoints based on the portal networks launched
We need to refactor the way the JSON-RPC server endpoints are launched, so it can be aware of all the networks that are spawned inside Trin and expose only endpoints for those networks.

ogenev:
### What was wrong?
Closes #728 

### How was it fixed?

- Add a new` rpc_server` and `builder` module in the `rpc` crate. Those include builder structs to initialize  `ethportal-api` server APIs.
- Use the new `rpc` crate API to launch the JSON-RPC server based on a selected combination of portal networks and transports.

**Note:** 
- HTTP endpoints are tested locally with portal-hive.  `ethportal-peertest` IPC tests for the beacon network will be added in a separate PR.
- State network endpoints will be included in the server launch when we add the API trait for the state network in `ethportal-api`.

### To-Do

- [ ] Clean up commit history


rpc/src/lib.rs
```diff
@@ -1,13 +1,103 @@
...
+                modules.push(PortalRpcModule::History);
+                modules.push(PortalRpcModule::Web3);
+            }
+            STATE_NETWORK => {
+                // not implemented
+            }
+            BEACON_NETWORK => modules.push(PortalRpcModule::Beacon),
+            _ => panic!("Unexpected network type: {}", network),
+        }
+    }
```
KolbyML, line 55: Shouldn't ``modules.push(PortalRpcModule::Web3);`` be pushed if any network exists? and not just if the history network is specified

KolbyML (approved):
Very nice PR :) I have that small nit pick of the conditions for web3 to launch but other then that this PR is so cool.

I tested history, history+beacon, and beacon and it all worked as expected!. I used http, didn't test the PR using ipc or ws.

:+1: nice PR :)

rpc/src/lib.rs
ogenev: good catch!

rpc/src/builder.rs
```diff
@@ -0,0 +1,505 @@
...
+        self.ws.as_ref()
+    }
+
+    /// Returns the [RpcModuleSelection] for the ipc transport
+    pub fn ipc(&self) -> Option<&RpcModuleSelection> {
+        self.ipc.as_ref()
+    }
+
+    /// Ensures that both http and ws are configured and that they are configured to use the same
+    /// port.j
```
njgheorghita, line 116: trailing `j`

rpc/src/builder.rs
```diff
@@ -0,0 +1,505 @@
...
+    /// The default modules `discv5`, `history`, `web3`
+    #[default]
+    Standard,
+    /// Only use the configured modules.
+    Selection(Vec<PortalRpcModule>),
+}
+
+impl RpcModuleSelection {
+    /// The standard modules to instantiate by default `discv5`, `history`, `web3`
+    pub const STANDARD_MODULES: [PortalRpcModule; 3] = [
```
njgheorghita, line 153: Thinking further about these "default" modules... don't we **need** to instantiate the beacon network by default? Otherwise we have no way to validate data returned from the history network (post-merge)

rpc/src/cors.rs
```diff
@@ -0,0 +1,46 @@
...
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum CorsDomainError {
+    #[error("{domain} is an invalid header value")]
+    InvalidHeader { domain: String },
+    #[error("Wildcard origin (`*`) cannot be passed as part of a list: {input}")]
+    WildCardNotAllowed { input: String },
+}
+
+/// Creates a [CorsLayer] from the given domains
+pub(crate) fn create_cors_layer(http_cors_domains: &str) -> Result<CorsLayer, CorsDomainError> {
```
njgheorghita, line 14: What is the use case where we would need support for cors?

njgheorghita (commented):
Well, this looks pretty great. However, I will say that this is a fairly large pr which makes it tough to review thoroughly. I'd have liked to see it split into smaller chunks and at least extract the ws stuff into a later pr (though, maybe that wasn't possible).

I'm curious as to why we add support for ws here? Sure, it's a pretty common transport to support, but imo it falls in the category of "if we don't have a use case for it, there's no reason for the extra complexity atm". Is there a use case that requires ws? In general, this project is becoming increasingly complex, and the fewer moving pieces to keep track of, the better. I can imagine scenarios where ws support might cause unexpected behavior in trin. Now, maybe this is a good thing (it exposes bugs in the core code) or maybe it's a bit of a distraction until this library reaches a more mature stage.

rpc/src/builder.rs
ogenev: Hmm, yeah, I think launching Beacon API should be the default if we use to for history validation.

rpc/src/cors.rs
ogenev: Enabling CORS in Trin allows web applications hosted on specified domains to make HTTP requests to the Trin HTTP-RPC server and interact with the Portal network. 

Almost all of the Ethereum clients implement CORS which I think is good to have it also for Trin.

ogenev:
> Well, this looks pretty great. However, I will say that this is a fairly large pr which makes it tough to review thoroughly. I'd have liked to see it split into smaller chunks and at least extract the ws stuff into a later pr (though, maybe that wasn't possible).

I know it is a large PR but it is hard to extract it into smaller PRs because the building blocks are interconnected. If you insist I can try harder but I'm not sure if we will gain much.
> 
> I'm curious as to why we add support for ws here? Sure, it's a pretty common transport to support, but imo it falls in the category of "if we don't have a use case for it, there's no reason for the extra complexity atm". Is there a use case that requires ws? In general, this project is becoming increasingly complex, and the fewer moving pieces to keep track of, the better. I can imagine scenarios where ws support might cause unexpected behavior in trin. Now, maybe this is a good thing (it exposes bugs in the core code) or maybe it's a bit of a distraction until this library reaches a more mature stage.

As discussed in the sync call, we agreed that WS would be a useful thing and I don't think it adds too much complexity. For the record, we support WS since the first transition to jsonrpsee server because it comes built-in with the HTTP server but with this PR is possible to turn it off or launch it only for a specific network.

rpc/src/builder.rs
```diff
@@ -0,0 +1,505 @@
...
+        PortalRpcModule::Discv5,
+        PortalRpcModule::History,
+        PortalRpcModule::Web3,
```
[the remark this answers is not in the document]
ogenev: I agree that we are in unknown territory here,  but I think it is now easy to change the default modules when we have better visibility of the user requirements.

njgheorghita (approved):
ahhh, apologies, I forgot to return to this. no need to split it up 😅 🚢

PR #743, opened 2023-05-16
Remove unused sort and take 16 from recursive_find_nodes

KolbyML:
### What was wrong?
```
/// Constructs a JSON call for the RecursiveFindNodes method.
async fn recursive_find_nodes(
    network: Arc<RwLock<HistoryNetwork>>,
    node_id: ethportal_api::NodeId,
) -> Result<Value, String> {
    let node_id = discv5::enr::NodeId::from(node_id.0);
    let overlay = network.read().await.overlay.clone();
    let mut nodes = overlay.lookup_node(node_id).await;
    nodes.sort_by(|a, b| {
        XorMetric::distance(&node_id.raw(), &a.node_id().raw())
            .cmp(&XorMetric::distance(&node_id.raw(), &b.node_id().raw()))
    });
    let nodes: Vec<Enr> = nodes.into_iter().take(16).collect();
    Ok(json!(nodes))
}
```

```
    nodes.sort_by(|a, b| {
        XorMetric::distance(&node_id.raw(), &a.node_id().raw())
            .cmp(&XorMetric::distance(&node_id.raw(), &b.node_id().raw()))
    });
    let nodes: Vec<Enr> = nodes.into_iter().take(16).collect();
```

^ this part of the code does nothing because this work is already done lower in the chain.
https://hackmd.io/@ZaEk4HVeTgieC1VFF87xDA/Hy-KPQgB3 here is me running the command over 20 times (I got the result before this sort happened). This is just me testing proof.

I will now show code proof
https://github.com/ethereum/trin/blob/c20ee54897094293adf9297e031ab9b41985f22f/portalnet/src/find/iterators/findnodes.rs#L46
The data structure we get the list from is a BTreeMap ordered by increasing distance

For take 16 we do that in this function
https://github.com/ethereum/trin/blob/c20ee54897094293adf9297e031ab9b41985f22f/portalnet/src/find/iterators/findnodes.rs#L252
### How was it fixed?
removed the 2 redundant lines of code
### To-Do

- [x] Clean up commit history

### DIFF
--- a/newsfragments/743.fixed.md
+++ b/newsfragments/743.fixed.md
@@ -1,0 +1,1 @@
+Remove redundant code from the `recursive_find_nodes` API
--- a/trin-beacon/src/jsonrpc.rs
+++ b/trin-beacon/src/jsonrpc.rs
@@ -7,10 +7,7 @@ use ethportal_api::types::jsonrpc::endpoints::BeaconEndpoint;
 use ethportal_api::types::jsonrpc::request::BeaconJsonRpcRequest;
 use ethportal_api::types::portal::{AcceptInfo, FindNodesInfo, PongInfo, TraceContentInfo};
 use ethportal_api::types::{
-    constants::CONTENT_ABSENT,
-    content_key::RawContentKey,
-    enr::Enr,
-    query_trace::QueryTrace,
+    constants::CONTENT_ABSENT, content_key::RawContentKey, query_trace::QueryTrace,
 };
 use ethportal_api::utils::bytes::hex_encode;
 use portalnet::storage::ContentStore;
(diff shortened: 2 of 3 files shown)


newsfragments/743.fixed.md
```diff
@@ -0,0 +1 @@
+Remove unused sort and take 16 from recursive_find_nodes
```
ogenev, line 1: nitpick:

```suggestion
Remove redundant code from the `recursive_find_nodes` API
```

ogenev (approved):
NIce :+1: !

newsfragments/743.fixed.md
KolbyML: resolved

PR #745, opened 2023-05-16
Fix tracing subscriber init to support env log levels

### LINKED ISSUE #734: Investigate setting log level issue
Currently, setting rust log levels via environment variable or by prepending the level to the cargo command doesn't seem to have any actual impact on the log levels that are output. This should be investigated & resolved as it's very useful to be able to set log levels via these mechanisms. A potential culprit to investigate is `tracing-subscriber` and the way in which we initialize/use this library.

njgheorghita:
### What was wrong?
Fixes #734 

### How was it fixed?
Added explicit support for env-set log levels, which isn't supported when calling `.init()` as opposed to `::init()`.
https://github.com/tokio-rs/tracing/issues/1697#issuecomment-1003513224

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-utils/Cargo.toml
+++ b/trin-utils/Cargo.toml
@@ -12,7 +12,7 @@ authors = ["https://github.com/ethereum/trin/graphs/contributors"]
 build = "build.rs"
 
 [dependencies]
-tracing-subscriber = "0.3.15"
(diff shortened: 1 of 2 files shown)


mrferris (approved):
🦀👍

PR #748, opened 2023-05-18
CLI parsing return clap:Error instead of panicking

KaydenML:
### What was wrong?
fix's https://github.com/ethereum/trin/issues/618
### How was it fixed?
i replaced the panics in CLI with clap::Error
### To-Do

- [x] Clean up commit history


njgheorghita:
Thanks! @KaydenML

PR #749, opened 2023-05-18
Simplify state network distance code

KolbyML:
### What was wrong?
The distance function code for the state network was undated and hard to read.
### How was it fixed?
It was fixed by simplifying the way we do the calculation instead of using the nums library we natively use U256 to do the calcluation.

The result of this in much more readable code, as well as the code resembling the spec's implementation which increases readability

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-state/src/utils.rs
+++ b/trin-state/src/utils.rs
@@ -1,72 +1,22 @@
 use ethereum_types::U256;
-use num::{
-    bigint::{BigInt, Sign},
-    Signed,
-};
-
-// 2 ** 256
-const MODULO: [u8; 78] = [
-    49, 49, 53, 55, 57, 50, 48, 56, 57, 50, 51, 55, 51, 49, 54, 49, 57, 53, 52, 50, 51, 53, 55, 48,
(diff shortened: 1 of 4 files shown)


trin-state/src/utils.rs
```diff
@@ -75,84 +53,55 @@ mod test {
...
 
-    // 2 ** 256 - 1
-    const MOD_SUB_ONE: &str =
-        "115792089237316195423570985008687907853269984665640564039457584007913129639935";
-    // 2 ** 255
-    const MID_DEC_STR: &str =
-        "57896044618658097711785492504343953926634992332820282019728792003956564819968";
-
+    // all 7 of these test cases are from
+    // https://github.com/ethereum/portal-network-specs/blob/master/state-network.md
```
njgheorghita, line 57: Imo, it would be nicer to just name all the tests.. `case_one` .. `case_seven`, or something similar. And then in a comment just above each test paste the actual case over from the specs

njgheorghita (approved):
Nice! I can't quite remember why we used `BigInt`... but all the cases are passing and this definitely seems to be an improvement

PR #751, opened 2023-05-24
Update trin-bridge to use TrustedProvider + switch to surf for performance

KolbyML:
### What was wrong?
Trin-Bridge doesn't use TrustedProvider it uses custom functions to do the job instead
### How was it fixed?
Switch TrustedProvider to use surf since trin-bridge needs it for performance reasons.

Then I removed the custom fetch logic and moved it to trustedprovider
### To-Do

- [x] Clean up commit history


njgheorghita:
Holding off on reviewing this for now... One reason we use `ureq` inside `trin` and `surf` only inside `trin-bridge` is that `ureq` is a very lightweight library, and sufficient for the needs of `trin`. Whereas we only need `surf` which is a fairly heavy library, to handle the intense load of concurrent requests that happen inside the bridge code. We've discussed this before and decided against using `surf` inside `trin` to keep the dependencies light.... we can discuss this again during the sync tm, but I'm not sure if the benefits gained from consolidating http request code into a single type (`TrustedProvider`) in both `trin` & `trin-bridge` are worthwhile. In fact they might unnecessarily complicate things since we really don't care about making consensus requests inside `trin` at all...

We should also discuss what we want to do with infura. We've discussed completely dropping infura support before, and relying solely on validating data via data fetched from the network. This could lead us to dropping the `TrustedProvider` entirely. Just some thoughts, we'll discuss the options tm.

KolbyML:
closed for https://github.com/ethereum/trin/issues/755

PR #757, opened 2023-05-30
Update default ws port to avoid collisions when running multiple nodes

njgheorghita:
### What was wrong?
#742 introduced support for websockets. When running trin-bridge with multiple nodes running at once, we end up with a panic resulting from a port collision. Since, the websocket port is set by default, spinning up multiple nodes at the same time results in multiple nodes trying to open a ws connection on the same port. This causes the events channel to close, and results in a panic.

A long-term solution would be to enable configuration for custom ws ports via a cli argument.

### How was it fixed?
- Update default ws to use the same port as http
- Handle panics inside a subnetwork events loop

### To-Do

- [x] Clean up commit history

### DIFF
--- a/rpc/src/rpc_server.rs
+++ b/rpc/src/rpc_server.rs
@@ -6,9 +6,7 @@ use crate::jsonrpsee::server::{IdProvider, Server, ServerBuilder, ServerHandle};
 use crate::jsonrpsee::ws_client::{WsClient, WsClientBuilder};
 use crate::jsonrpsee::RpcModule;
 use crate::{RpcError, TransportRpcModuleConfig};
-use ethportal_api::types::cli::{
-    DEFAULT_WEB3_HTTP_PORT, DEFAULT_WEB3_IPC_PATH, DEFAULT_WEB3_WS_PORT,
-};
+use ethportal_api::types::cli::{DEFAULT_WEB3_HTTP_PORT, DEFAULT_WEB3_IPC_PATH};
 use reth_ipc::server::Builder as IpcServerBuilder;
 use reth_ipc::server::{Endpoint, IpcServer};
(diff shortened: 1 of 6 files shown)


rpc/src/rpc_server.rs
```diff
@@ -375,10 +373,13 @@ impl RpcServerConfig {
             ))
         });
 
-        let ws_socket_addr = self.ws_addr.unwrap_or(SocketAddr::V4(SocketAddrV4::new(
-            Ipv4Addr::UNSPECIFIED,
-            DEFAULT_WEB3_WS_PORT,
-        )));
+        // by default, we configure ws on the same port as http
```
ogenev, line 376: Can we run http and ws at the same port simultaneously? For example `geth` uses 8545 for hhtp and 8546 for ws.

njgheorghita: > Can we run http and ws at the same port simultaneously?

Maybe I'm confused here, but based on the comment below, on line 384, `// If both are configured on the same port, we combine them into one server.`, I assumed it was possible. Did I misinterpret your comment?

I guess the question is what should be the expected behavior from the user's perspective. In which case geth is probably a good example to follow. Based on their [cli options](https://geth.ethereum.org/docs/fundamentals/command-line-options)...
- they don't enable ws by default
- they allow the user to set a custom ws port

I'd suggest we add a `--ws` flag which will...
- enable the ws server if the flag is provided
- take a port, `--ws 8546`

But, in all honesty, this pr was meant to be more of a bugfix than adding a new feature. Depending on what we decide re: cli flags, I'd prefer to go ahead with merging this pr (assuming it's not a problem for http & ws to share the port) and then I'll open a tracking issue to update the cli flags.

PR #763, opened 2023-06-05
Add `network` cli flag to trin-bridge

ogenev:
### What was wrong?
No support for running a bridge node with more than one overlay network.

### How was it fixed?
Add 'network' cli flag to trin-bridge

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/trin-bridge/src/cli.rs
+++ b/trin-bridge/src/cli.rs
@@ -1,3 +1,4 @@
+use crate::types::NetworkKind;
 use clap::Parser;
 use std::path::PathBuf;
 use std::str::FromStr;
@@ -8,6 +9,7 @@ use std::str::FromStr;
 //   less and less evenly spread
 // - running more than 16 trin nodes simultaneously is not thoroughly tested
 pub const MAX_NODE_COUNT: u8 = 16;
+const DEFAULT_SUBNETWORK: &str = "history";
 
 #[derive(Parser, Debug, PartialEq)]
 #[command(name = "Trin Bridge", about = "Feed the network")]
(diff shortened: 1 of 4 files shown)


trin-bridge/src/types.rs
```diff
@@ -0,0 +1,32 @@
+use std::fmt;
+use std::str::FromStr;
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum NetworkKind {
```
njgheorghita, line 5: A bit of a nitpick, but I'd prefer something like `Subnetwork` here - not a blocker, so feel free to ignore me. I just hate types that contain `Type` / `Kind` in the name.

trin-bridge/src/cli.rs
```diff
@@ -34,6 +36,14 @@ pub struct BridgeConfig {
         help = "Path to epoch accumulator repo for bridge mode"
     )]
     pub epoch_acc_path: PathBuf,
+
+    #[arg(
+    long = "network",
```
njgheorghita, line 41: *nitpick* indentations

njgheorghita (approved):
Cool! My initial thought was that we shouldn't support this kind of feature. If a user wants to run the bridge, they must bridge data into all networks. Especially since a consensus client is required to create history network proofs, we might as well take advantage of the fact that any bridge user needs a connection to both consensus / execution.

The follow-up thought was that... we are the only users. We = those w/ pandaops privileges. I'm happy to move forward with evolving the bridge under this assumption.

And, I do see some value from a testing / debugging perspective of being able to run the bridge against a single subnetwork.

PR #767, opened 2023-06-06
Use default utp connection config

njgheorghita:
### What was wrong?
Rather than maintaining our own set of config constants, we should defer to the utp crate's default value for `ConnectionConfig`

The initial set of constant values was created in https://github.com/ethereum/trin/pull/551/files#diff-354721dfd15f4b0d0f5df418d0ac4118592bf96bda720655548e380faf9f1b37R77 - without any comment explaining as to why those values were chosen.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -20,6 +20,7 @@ use discv5::{
     rpc::RequestId,
 };
 use futures::{channel::oneshot, future::join_all, prelude::*};
+use lazy_static::lazy_static;
 use parking_lot::RwLock;
 use rand::seq::{IteratorRandom, SliceRandom};
 use smallvec::SmallVec;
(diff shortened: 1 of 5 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -77,17 +77,6 @@ const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
 /// Bucket refresh lookup interval in seconds
 const BUCKET_REFRESH_INTERVAL_SECS: u64 = 60;
 
-/// The default configuration to use for uTP connections.
-pub const UTP_CONN_CFG: ConnectionConfig = ConnectionConfig {
-    max_packet_size: 1024,
```
jacobkaufmann, line 82: at the very least you will want to set `max_packet_size` to the [`discv5` max packet size](https://github.com/sigp/discv5/blob/19342b374647c617eead7e88a4242e8cec68f7f9/src/packet/mod.rs#L44)

jacobkaufmann: or some value at most the `discv5` max packet size

jacobkaufmann: @njgheorghita you will actually want to set this strictly lower than the `discv5` max packet size to account for the `discv5` talk request overhead (e.g. header, request ID, etc), since `utp` is just packaging the payloads

njgheorghita: Gotcha, am I correct in understanding that the overhead is about `72` bytes? https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#ordinary-message-packet-flag--0

So `max_packet_size` here should be set to `1280-72 = 1208` bytes?

jacobkaufmann: I think there would be some more overhead for the talk request `request-id` and `protocol`. you might be able to just create an empty talk request and see what the overhead is

njgheorghita: Having trouble figuring out how to create an empty talkreq... any thoughts?
- `Request-id <= 8bytes` [from the spec](https://github.com/ethereum/devp2p/blob/master/discv5/discv5-wire.md#protocol-messages)
- The largest "protocol" we use in trin is `0x757470` for utp. (`3 bytes`)

If the additional overhead is at most 11 bytes, would `max_packet_size` of `1280 - 72 - 11 = 1197` bytes work in this case? Or is it preferable to get an exact count on this by creating an empty talk req? (and if so, any ideas on the best way to hook into the library to generate the empty talk req are 💯 )

njgheorghita:
closing in favor of #909

PR #771, opened 2023-06-07
Remove unused ENR dependency

KolbyML:
### What was wrong?
We import ENR depend in our Cargo.toml, but only use the one re-exported from discv5, so it is unused.
### How was it fixed?
by removing the unused depend


njgheorghita (approved):
:shipit:

PR #772, opened 2023-06-07
Refactor `trin-bridge` with PandaOps middleware

ogenev:
### What was wrong?

trin-bridge should provide data for more than one overlay subnetwork. Some refactoring is necessary to enable more solid architecture for adding support for the beacon and state overlay data.

### How was it fixed?
Create a Pandaops middleware structure and encapsulate requesting the data from the DevOps nodes. 

### To-Do

- [x]  Test this PR live when PandaOps nodes are back online
- [x] Clean up commit history

### DIFF
--- a/trin-bridge/src/bridge.rs
+++ b/trin-bridge/src/bridge.rs
@@ -1,9 +1,7 @@
 use crate::cli::BridgeMode;
-use crate::constants::PANDAOPS_URL;
-use crate::full_header::{FullHeader, FullHeaderBatch};
-use crate::{PANDAOPS_CLIENT_ID, PANDAOPS_CLIENT_SECRET};
+use crate::full_header::FullHeader;
+use crate::pandaops::PandaOpsMiddleware;
 use anyhow::{anyhow, bail};
-use ethereum_types::H256;
 use ethportal_api::jsonrpsee::http_client::HttpClient;
 use ethportal_api::types::content_key::{
     BlockBodyKey, BlockHeaderKey, BlockReceiptsKey, EpochAccumulatorKey, HistoryContentKey,
@@ -18,12 +16,9 @@ use ethportal_api::types::execution::header::{
     AccumulatorProof, BlockHeaderProof, Header, HeaderWithProof, SszNone,
 };
 use ethportal_api::types::execution::receipts::Receipts;
-use ethportal_api::types::jsonrpc::params::Params;
-use ethportal_api::types::jsonrpc::request::JsonRequest;
 use ethportal_api::utils::bytes::hex_encode;
 use ethportal_api::HistoryNetworkApiClient;
 use futures::stream::StreamExt;
-use serde_json::{json, Value};
 use ssz::Decode;
 use std::fs;
 use std::ops::Range;
@@ -40,19 +35,36 @@ use trin_validation::accumulator::MasterAccumulator;
 use trin_validation::constants::{EPOCH_SIZE as EPOCH_SIZE_USIZE, MERGE_BLOCK_NUMBER};
 use trin_validation::oracle::HeaderOracle;
 
-pub struct Bridge {
-    pub mode: BridgeMode,
-    pub portal_clients: Vec<HttpClient>,
-    pub header_oracle: HeaderOracle,
-    pub epoch_acc_path: PathBuf,
-}
-
 // todo: calculate / test optimal saturation delay
-// todo: single block option
 const HEADER_SATURATION_DELAY: u64 = 10; // seconds
 const LATEST_BLOCK_POLL_RATE: u64 = 5; // seconds
 const EPOCH_SIZE: u64 = EPOCH_SIZE_USIZE as u64;
 const FUTURES_BUFFER_SIZE: usize = 32;
+
+pub struct Bridge {
+    pub mode: BridgeMode,
+    pub portal_clients: Vec<HttpClient>,
+    pub pandaops: PandaOpsMiddleware,
+    pub header_oracle: HeaderOracle,
+    pub epoch_acc_path: PathBuf,
+}
+
+impl Bridge {
+    pub fn new(
+        mode: BridgeMode,
+        portal_clients: Vec<HttpClient>,
+        header_oracle: HeaderOracle,
+        epoch_acc_path: PathBuf,
+    ) -> Self {
+        Self {
+            mode,
+            portal_clients,
+            pandaops: PandaOpsMiddleware::default(),
+            header_oracle,
+            epoch_acc_path,
+        }
+    }
+}
 
 impl Bridge {
     pub async fn launch_single(&self, block_number: u64) {
@@ -75,13 +87,14 @@ impl Bridge {
 
         info!("Gossiping block: {block_number:?}");
         let gossip_stats = Arc::new(Mutex::new(GossipStats::new(gossip_range.clone())));
-        match Bridge::serve_full_block(
-            block_number,
-            epoch_acc,
-            gossip_stats,
-            self.portal_clients.clone(),
-        )
-        .await
+        match self
+            .serve_full_block(
+                block_number,
+                epoch_acc,
+                gossip_stats,
+                self.portal_clients.clone(),
+            )
+            .await
         {
             Ok(_) => info!("Successfully served block: {block_number:?}"),
             Err(msg) => warn!("Error serving block: {block_number:?}: {msg:?}"),
@@ -91,12 +104,12 @@ impl Bridge {
     // Devops nodes don't have websockets available, so we can't actually poll the latest block.
     // Instead we loop on a short interval and fetch the latest blocks not yet served.
     pub async fn launch_latest(&self) {
-        let mut block_index = get_latest_block_number().await.expect(
+        let mut block_index = self.pandaops.get_latest_block_number().await.expect(
             "Error launching bridge in latest mode. Unable to get latest block from provider.",
         );
         loop {
             sleep(Duration::from_secs(LATEST_BLOCK_POLL_RATE)).await;
-            let latest_block = match get_latest_block_number().await {
+            let latest_block = match self.pandaops.get_latest_block_number().await {
                 Ok(val) => val,
                 Err(msg) => {
                     warn!("error getting latest block, skipping iteration: {msg:?}");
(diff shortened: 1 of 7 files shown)


trin-bridge/src/bridge.rs
```diff
@@ -82,13 +100,14 @@ impl Bridge {
...
                     let gossip_stats = gossip_stats.clone();
                     async move {
-                        let _ = Bridge::serve_full_block(
-                            height,
-                            None,
-                            gossip_stats,
-                            self.portal_clients.clone(),
-                        )
-                        .await;
+                        let _ = self
```
njgheorghita, line 103: When I first implemented this functionality, I specifically tried to avoid using a shared reference, since I was using `tokio::spawn` and didn't want to have a shared reference across threads. Now, to be honest, I'm not sure what effect this actually has on performance, but my assumption was that we will eventually want to have fairly high concurrency in the bridge, so it's better to avoid shared refs across threads. I do agree that the shared ref does help simplify the code, but wonder if you have any opinions on the performance side of things?

ogenev: I see your concerns. I'm not sure what concurrency we will need and the future is unknown to me. I am more of an advocate of simple usage right now and adding complexity when the need arises.

trin-bridge/src/bridge.rs
```diff
@@ -40,19 +37,42 @@ use trin_validation::accumulator::MasterAccumulator;
 use trin_validation::constants::{EPOCH_SIZE as EPOCH_SIZE_USIZE, MERGE_BLOCK_NUMBER};
 use trin_validation::oracle::HeaderOracle;
 
+// todo: calculate / test optimal saturation delay
+// todo: single block option
```
njgheorghita, line 41: You can remove this line since it was implemented in #776

trin-bridge/src/bridge.rs
```diff
@@ -40,19 +37,42 @@ use trin_validation::accumulator::MasterAccumulator;
...
+    pub fn new(
+        mode: BridgeMode,
+        portal_clients: Vec<HttpClient>,
+        header_oracle: HeaderOracle,
+        epoch_acc_path: PathBuf,
+    ) -> Self {
+        Self {
+            mode,
+            portal_clients,
+            pandaops: PandaOpsMiddleware::new(
```
njgheorghita, line 65: Could we `impl Default` for `PandaOpsMiddleware` instead of using `new` here?

trin-bridge/src/bridge.rs
```diff
@@ -200,23 +217,34 @@ impl Bridge {
...
             full_header.epoch_acc = epoch_acc;
         }
         Bridge::gossip_header(&full_header, &portal_clients, &gossip_stats).await?;
         // Sleep for 10 seconds to allow headers to saturate network,
         // since they must be available for body / receipt validation.
         sleep(Duration::from_secs(HEADER_SATURATION_DELAY)).await;
-        Bridge::construct_and_gossip_block_body(&full_header, &portal_clients, &gossip_stats)
-            .await?;
-        Bridge::construct_and_gossip_receipt(&full_header, &portal_clients, &gossip_stats).await
+        if let Err(err) = self
```
njgheorghita, line 235: Curious why you changed to log the errors at this level? Imo, it's cleaner to log it at the top-level (aka inside the `launch_*` methods).
1. It removes the `let _ = ...` clause from the `launch_*` methods which is messy imo
2. In this case I think it's cleaner to let the error propagate all the way up and log them there.

If you're looking for some enhanced granularity in the log msg, maybe we could just `map_err()` at this level to add context to the error message...

```rust
        Bridge::construct_and_gossip_block_body(&full_header, &portal_clients, &gossip_stats)
            .await
            .map_err(|err| anyhow!("Error serving block body: {err:?}"))?;
```

trin-bridge/src/constants.rs
```diff
@@ -7,4 +7,7 @@ pub const BATCH_SIZE: u64 = 128;
 // These are only intended to be used by core team members who have access to the nodes.
 // If you don't have access to the PANDAOPS nodes, but still want to use the bridge feature, let us
 // know on Discord or Github and we'll prioritize support for any provider.
-pub const PANDAOPS_URL: &str = "https://archive.mainnet.ethpandaops.io/";
+/// Execution layer PandaOps endpoint
+pub const BASE_EL_ENDPOINT: &str = "https://geth-lighthouse.mainnet.ethpandaops.io";
```
njgheorghita, line 11: I think we should leave this as `archive`... it makes a difference when we're trying to access receipts/txs that have been pruned in older blocks

trin-bridge/src/constants.rs
```diff
@@ -7,4 +7,7 @@ pub const BATCH_SIZE: u64 = 128;
...
+/// Consensus layer PandaOps endpoint
+pub const BASE_CL_ENDPOINT: &str = "https://beacon.mainnet.ethpandaops.io";
```
njgheorghita, line 13: I'm assuming that you've tested this out - and removing the trailing `/` from the url didn't affect the http request? Sometimes these http libraries can be very particular about those kinds of things, but off the top of my head I can't remember if this was required or not with `surf`

trin-bridge/src/pandaops.rs
```diff
@@ -0,0 +1,135 @@
...
+            .header("Content-Type", "application/json".to_string())
+            .header("CF-Access-Client-Id", self.client_id.clone())
+            .header("CF-Access-Client-Secret", self.client_secret.clone())
+            .recv_string()
+            .await;
+
+        result.map_err(|_| anyhow!("Unable to request execution batch from pandaops"))
+    }
+
+    #[allow(dead_code)]
```
njgheorghita, line 52: Can you add a comment for why this is allowed / when can we expect the `allow` statement to be removed? (eg. when we actually implement some calls to a consensus client)

trin-bridge/src/pandaops.rs
```diff
@@ -0,0 +1,135 @@
...
+            client_secret,
+        }
+    }
+}
+
+impl PandaOpsMiddleware {
+    /// Used the "surf" library here instead of "ureq" since "surf" is much more capable of handling
+    /// multiple async requests. Using "ureq" consistently resulted in errors as soon as the number of
+    /// concurrent tasks increased significantly.
+    async fn batch_request(&self, obj: Vec<JsonRequest>) -> anyhow::Result<String> {
```
njgheorghita, line 38: I guess, if the middleware now has this concept of el & cl providers, we might want to change the names of `batch_request` / `request` to indicate that they are tied to a specific kind of provider. Maybe `el_batch_request` / `cl_request`? Idk, there's probably also a further refactoring opportunity here where we have an actual `Provider` trait that implements `request` & `batch_request` for both el & cl providers... but I'm not convinced that's something we need to implement at this point.

trin-bridge/src/pandaops.rs
```diff
@@ -0,0 +1,135 @@
...
+            .middleware(Retry::default())
+            .body_json(&json!(obj))
+            .map_err(|e| anyhow!("Unable to construct json post request: {e:?}"))?
+            .header("Content-Type", "application/json".to_string())
+            .header("CF-Access-Client-Id", self.client_id.clone())
+            .header("CF-Access-Client-Secret", self.client_secret.clone())
+            .recv_string()
+            .await;
+
+        result.map_err(|_| anyhow!("Unable to request execution batch from pandaops"))
```
njgheorghita, line 49: I'd say we probably want to preserve the error message here & on line 60, and interpolate the old error message into the new one... That kind of info can be very useful when debugging

trin-bridge/src/pandaops.rs
```diff
@@ -0,0 +1,135 @@
...
+        let result = surf::get(endpoint)
+            .header("Content-Type", "application/json".to_string())
+            .header("CF-Access-Client-Id", self.client_id.clone())
+            .header("CF-Access-Client-Secret", self.client_secret.clone())
+            .recv_string()
+            .await;
+        result.map_err(|_| anyhow!("Unable to request consensus block from pandaops"))
+    }
+
+    pub(crate) async fn get_trusted_receipts(
```
njgheorghita, line 63: OK... I have to ask bc I've seen this before but couldn't find in the docs what's going on. What exactly is `pub(crate)` doing? Is this only a public function within this crate? What's your reasoning behind using this vs a pure `pub async fn...`?

ogenev: Yes, it is a public function within the crate and because this function is not used outside of the `trin-bridge` crate, it makes more sense to be public only for the crate. We can change this to pure public func if we need it in the future imo.

trin-bridge/src/bridge.rs
ogenev: Default here would require Default implementations for `BridgeMode`, `HeaderOracle`, `epoch_acc_path`, and `HttpClient`. IMO, this will add more overheating to this PR...

trin-bridge/src/pandaops.rs
ogenev: Yeah, I'm planning to refactor this further when we implement the consensus data retrieval.

trin-bridge/src/bridge.rs
njgheorghita: I'm talking about `PandaOpsMiddleware::default()` not `Bridge::default()`. `PandaOpsMiddleware::default()` should be fairly trivial to implement

trin-bridge/src/constants.rs
ogenev: Yep, tested and working fine without trailing `/`.

trin-bridge/src/constants.rs
```diff
@@ -7,4 +7,7 @@ pub const BATCH_SIZE: u64 = 128;
 // These are only intended to be used by core team members who have access to the nodes.
 // If you don't have access to the PANDAOPS nodes, but still want to use the bridge feature, let us
 // know on Discord or Github and we'll prioritize support for any provider.
-pub const PANDAOPS_URL: &str = "https://archive.mainnet.ethpandaops.io/";
+/// Execution layer PandaOps endpoint
+pub const BASE_EL_ENDPOINT: &str = "https://archive.mainnet.eu1.ethpandaops.io";
```
ogenev, line 11: @njgheorghita FYI the archive node url is changed to this one.

PR #776, opened 2023-06-15
Single block mode support inside bridge

njgheorghita:
### What was wrong?
Add support for bridge mode to construct & gossip a single block

### To-Do

- [x] Clean up commit history

### DIFF
--- a/trin-bridge/src/cli.rs
+++ b/trin-bridge/src/cli.rs
@@ -59,11 +59,15 @@ fn check_node_count(val: &str) -> Result<u8, String> {
 /// - Latest: tracks the latest header
 /// - Backfill: starts at block 0
 /// - StartFromEpoch: starts at the given epoch
+///   - ex: "e123" starts at epoch 123
+/// - Single: executes a single block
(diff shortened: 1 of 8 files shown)


trin-bridge/src/cli.rs
```diff
@@ -59,11 +59,15 @@ fn check_node_count(val: &str) -> Result<u8, String> {
 /// - Latest: tracks the latest header
 /// - Backfill: starts at block 0
 /// - StartFromEpoch: starts at the given epoch
+///   - ex: "e123" starts at epoch 123
+/// - Single: executes a single block
+///   - ex: "b123" executes block 123
```
mrferris, line 64: Nit: In the future something like just `--single-block 123` and `--epoch 123` (that can't be used with `--latest` or `--backfill`) would be better CLI design.

mrferris (approved):
👍

trin-bridge/src/cli.rs
njgheorghita: agreed! tracking in #783

PR #777, opened 2023-06-16
Add catch for bootnode self-update

njgheorghita:
### What was wrong?
If a network bootnode uses the default bootnode list to bootstrap itself into the network, it will inevitably try to connect with itself, which will throw an `invalid self update` error.

### How was it fixed?
This clause checks to make sure that a node never attempts to add itself to its routing table.

### To-Do

- [x] Clean up commit history


njgheorghita:
going ahead with a bit of a rage-merge here to get this up on the network, but happy to revisit/address any posthumous comments

Issue #759, opened 2023-05-30
Clean up tracking from #756


KolbyML:
This issue is to track clean ups requested in https://github.com/ethereum/trin/pull/756 , but made sense to do them in there own PR

- [x] [remove TrinEndpoint](https://github.com/ethereum/trin/pull/756#discussion_r1210360796)
- [x] [Add a scenario for validating a pre-merge header-with-proof, bb, & receipts.](https://github.com/ethereum/trin/pull/756#discussion_r1210355170)

KolbyML:
resolved both tracked tasks closing issue

PR #778, opened 2023-06-19
Cleaning up unsupported/used jsonrpc utils

KolbyML:
### What was wrong?
fixes issue 1 in tracking pr https://github.com/ethereum/trin/issues/759
### How was it fixed?
by looking for unused code and removing it


njgheorghita (approved):
🚢

PR #781, opened 2023-06-22
Remove redundant withdrawals request

njgheorghita:
### What was wrong?
On the first go-round of implementing withdrawals, I just completely missed that they were available in the execution block and didn't require an additional call to a consensus client. This removes the redundant call. 

### How was it fixed?
- Add `withdrawals` field to `FullHeader`
- Updates a test asset to contain full transactions
- Add a test asset to check calculation of empty withdrawals root

### To-Do

- [x] Clean up commit history


mrferris (approved):
👍

Issue #755, opened 2023-05-26
Update trin's validation strategy


njgheorghita:
Following up with our recent discussion in weekly stand, we decided to update trin's validation strategy. 

### Step 1
Remove all dependence upon Infura / external provider to validate content data. Essentially this consists of removing the `TrustedProvider` type.

### Step 2
A cli flag that the user can set to choose the level of data validation that occurs within `HeaderOracle`
- `--validation strict` - the default setting. 
  - this attempts to validate all content by fetching validation data from the network. it will only return data that is successfully validated
- `--validation loose` 
  - return content that passes validation & also in the case where we are unable to acquire the data from the network required to validate the content
- `--validation none`
  - return all content retrieved from the network, regardless of validation

KolbyML:
> * `--validation loose`
>   
>   * return content that passes validation & also in the case where we are unable to acquire the data from the network required to validate the content

what do you mean by "also in the case where we are unable to acquire the data from the network required to validate the content"

Does it mean this flag will allow us to return data we validated, but also data we didn't?

Wouldn't that then be the same as none though since non will return both validated and non-validated data.

njgheorghita:
Tbh, I'm not convinced this is the ideal strategy that we want to support, so I'd wait until we get some other team members to chime in on how we want to build this out...

> Wouldn't that then be the same as none though since non will return both validated and non-validated data.

Almost the same, the only difference is that `loose` will **not** return data that has been invalidated. `none` will return any and all data retrieved from the network. Again, not even sure if this is something we want to support?

njgheorghita:
Actually, now that I'm thinking about it a bit more. It seems likely that a `none` flag could be useful in particular for debugging network data.

njgheorghita:
Closing this since we've adopted a de facto "strict" validation policy

PR #782, opened 2023-06-23
Add validate_content jsonrpc

KolbyML:
### What was wrong?
There wasn't a good way to validate if you had data other then push it onto the network and try and fetch it again. We should have a way for a user to validate data they have throw jsonrpc
### How was it fixed?
Add ``validate_content`` json rpc
### Use cases?

- debugging
- when we settle how we are going to handle verification https://github.com/ethereum/trin/issues/755
If someone chooses ``loose`` it might be useful for a user to check if we were able to validate the data for important information to them?
- Direct access to validate if a users data is valid instead of having to store then findcontent. Assuming they already have the data

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -95,6 +95,14 @@ pub trait BeaconNetworkApi {
         content_value: Option<BeaconContentValue>,
     ) -> RpcResult<AcceptInfo>;
 
+    /// Returns true or false depending on if the client was able to validate a given history content key and content data.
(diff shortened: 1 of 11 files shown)


pipermerriam:
Curious why this got closed?  seems like a valuable endpoint to have if for no other reason than to be able to have hive be able to easily test cross client test vectors for whether clients are implementing this correctly.

KolbyML:
https://github.com/ethereum/portal-network-specs/pull/214
^ here is context

PR #784, opened 2023-06-27
ethportal-api: Change `node_id` field of `NodeInfo` to `String`

ogenev:
### What was wrong?
The `NodeId `type as a `NodeInfo` field is incompatible with Fluffy and Ultralight and doesn't follow the json-rpc specs. This causes `portal-hive` tests to fail.

### How was it fixed?
Change node id field type to `String`.

### To-Do

- [ ] Clean up commit history


njgheorghita (approved):
:ship:

PR #785, opened 2023-06-27
Add support for fluffy-bridge

njgheorghita:
### What was wrong?
Add support for trin-bridge to use fluffy as the client it uses to gossip out data to the network.

### How was it fixed?
Added a subcommand to the `trin-bridge` command that requires a user to specify what kind of client is located at `--executable-path` 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7182,6 +7182,7 @@ dependencies = [
  "futures 0.3.28",
  "jsonrpsee 0.16.2",
  "lazy_static",
+ "portalnet",
  "rstest 0.11.0",
  "serde",
  "serde_json",
@@ -7192,6 +7193,7 @@ dependencies = [
  "tracing-subscriber",
  "trin-utils",
  "trin-validation",
+ "url",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


ogenev (approved):
I think that having `trin_handle` and `fluffy_handle` functions with the same arguments is a sign that we may want to leverage a trait here, for example, `ClientHandle`, and implement it for every client type.

njgheorghita:
@ogenev That's a fair point... however I'm not sure I see any massive improvements, just rearranging the complexity. Although, I might be misunderstanding your suggestion here. 

We don't have a need for a `struct Client` that contains client-related attributes. All that matters is the info from the `enum ClientType` (are we dealing with fluffy or trin). Then we know what kind of handle to build.

I don't see an advantage from `impl trait` on an `enum ClientType`. imo that's no less complex than simply `impl build_handle()` on `enum ClientType`. If we had a `struct TrinClient` & `struct FluffyClient`, I'd agree that we should use a trait here. 

I've spent a little time messing around with implementing a `Clients` & `Client` struct, but that didn't come to a satisfactory conclusion.

I agree this isn't the prettiest setup, and there is some repetition. I just don't see any obvious improvements without simply rearranging the complexity. I'm going to go ahead and merge this now to unblock my other pr. But I'm very open to revisiting this later especially if you can give a bit more detail of your idea of implementing your solution.

PR #790, opened 2023-07-07
Remove `get_block` RPC from `light-client` crate

ogenev:
### What was wrong?

We do not need anymore an RPC to get a beacon block from the light client network because this data will not be available.

### How was it fixed?
- Remove `get_block` rpc and the execution payload extraction from the block.
- Rename `ConsensusClient` to `ConsensusLightClient`

### To-Do

- [ ] Clean up commit history


light-client/src/consensus/consensus_client.rs
```diff
@@ -44,15 +43,15 @@ struct LightClientStore {
...
 }
 
-impl<R: ConsensusRpc> ConsensusClient<R> {
+impl<R: ConsensusRpc> ConsensusLightClient<R> {
     pub fn new(
         rpc: &str,
         checkpoint_block_root: &[u8],
         config: Arc<Config>,
-    ) -> Result<ConsensusClient<R>> {
+    ) -> Result<ConsensusLightClient<R>> {
```
njgheorghita, line 51: *nitpick* but it seems like we can just access the `rpc` value from the `Config` directly rather than requiring to pass it in as a separate argument

njgheorghita (approved):
:ship:

light-client/src/consensus/consensus_client.rs
ogenev: I see some benefits in leaving this as a separate argument, will merge it as it is for now.

PR #793, opened 2023-07-11
Introduce an event stream to `OverlayService`

ogenev:
### What was wrong?

 We want to introduce a separate Beacon network service that takes care of the light client mechanics and keeps the node in sync with the tip of the chain. This service needs to listen for the latest `LightClientFinality` and `LIghtClientOptimistic` content updates. Currently, there is no way to listen for an event from the overlay service.

### How was it fixed?

IMO, listening for an event is best done by an event stream. This PR adds an event stream to the `OverlayService` which can emit light client updates events or any other kind of events. As discussed in Boulder, this event stream can be also used for testing and debugging.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portalnet/src/events.rs
+++ b/portalnet/src/events.rs
@@ -1,4 +1,5 @@
 use std::str::FromStr;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use discv5::TalkRequest;
 use tokio::sync::mpsc;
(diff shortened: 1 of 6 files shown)


portalnet/src/events.rs
```diff
@@ -103,3 +103,10 @@ impl PortalnetEvents {
         }
     }
 }
+
+/// Events that can be produced by the `OverlayProtocol` event stream.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum OverlayEvent {
+    LightClientOptimisticUpdate,
```
ogenev, line 110: We can expand those updates by including more data like timestamps, content data and etc., but I'm keeping this simple for now until we know what we want exactly.

pipermerriam: Thoughts on implementing a simple "envelope" struct like this to go ahead and include timestamps.

```
struct EventEnvelope:
    timestamp: DateTime
    payload: OverlayEvent
```

portalnet/src/overlay.rs
```diff
@@ -701,6 +705,26 @@ where
...
+            let (callback_send, callback_recv) = oneshot::channel();
+
+            let command = OverlayCommand::RequestEventStream(callback_send);
+            channel
+                .send(command)
+                .map_err(|_| anyhow!("The Overlay Service channel has been closed early."))?;
+
+            callback_recv
+                .await
+                .map_err(|_| anyhow!("The Overlay Service channel has been closed early."))
```
njgheorghita, line 725: *nitpick* adjust this error message so it's distinguishable from the one directly above

portalnet/src/overlay_service.rs
```diff
@@ -487,6 +493,15 @@ where
                                 );
                             }
                         }
+                        OverlayCommand::RequestEventStream(callback) => {
```
njgheorghita, line 496: Am I missing something? This is only called once, iiuc. Why do we need to include it inside the `select` loop? Could we not just instantiate it once at the beginning of the `start()` method.

njgheorghita (commented):
In general, I'm ok with starting simple & evolving as we go. If you feel strongly that this is the right approach here, then I trust your judgement. *If this functionality isn't urgently needed*, would it be too much of a hassle to continue to build it out on a separate branch? At least until we're reporting every type of message that we eventually want to report (utp & protocol messages).

portalnet/src/overlay_service.rs
ogenev: This is a command that the overlay service receives from `OverlayProtocol` to create an event stream.

 If we include the creation of the event stream at the beginning of the `start()` method, it will create this stream for any `OverlayProtocol` instance, including tests, without giving us the opportunity to choose for which instance we want to create the stream.  The API is as follows:

 ```rust
 let overlay = OverlayProtocol::new(...);
 // Here we explicitly create the stream and we can pass it to another service if needed.
let stream = overlay.event_stream().await?'
````

For example, in this PR, we create the stream only for the Beacon Network because we will need the overlay protocol to signalize to the beacon service as soon as the node receives new `LightClientOptimisticUpdate` or `LightClientFinalityUpdate` content items.

**Comment by ogenev** [comment]
> _If this functionality isn't urgently needed_, would it be too much of a hassle to continue to build it out on a separate branch?

The light client sync process needs to receive every slot a `LightClientOptimisticUpdate` and `LightClientFinalityUpdate` objects. The beacon bridge will push those objects every slot (12s) and we want a way to notify the light client service as soon as the overlay protocol receives the content via the gossip mechanics of the network. This is why we want to create an event stream in the overlay service for the beacon network and emit those events to the light client service as soon as possible.

So I think this functionality is mandatory for the beacon network and because we use an `OverlayCommand` to create the service, we keep it optional for the rest of the networks.

File: portalnet/src/events.rs (hunk 1 above)
Comment (ogenev) [reply portalnet/src/events.rs:111]: I like it, implemented in https://github.com/ethereum/trin/pull/793/commits/63c77de1453f62765b3cb999f3b1174ea608ffa8.

**Review by njgheorghita — APPROVED** [review]
It still feels strange to me. Up to this point we've tried very hard to keep `OverlayService` generic over every subnetwork. This feels like a significant diversion from that pattern. Do you agree? Am I missing something?

Maybe it's something better discussed in sync later today. But under the guise of "move fast and break things" I'm :+1:

**Comment by ogenev** [comment]
> It still feels strange to me. Up to this point we've tried very hard to keep `OverlayService` generic over every subnetwork. This feels like a significant diversion from that pattern. Do you agree? Am I missing something?

Indeed, the goal is to keep `OverlayService` generic and that;s why we want to produce events - so we can handle and process them in separate services without coupling these logics with the generic pattern.

File: portalnet/src/events.rs [hunk 4]
```diff
@@ -103,3 +104,92 @@ impl PortalnetEvents {
...
+    LightClientFinalityUpdate,
+}
+
+/// Timestamp of an overlay event.
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Timestamp {
+    /// Timestamp not available.
+    NotAvailable,
+    /// Event creation time.
+    CreateTime(i64),
```
mrferris, line 121: What's the purpose of a signed number here? And what's the significance of the case of `-1` in `to_millis()`?

ogenev: Unix timestamps are in signed integers and  `-1` represents an error. It is mostly for historical reasons and deals with dates before January 1st, 1970. You can read more [here](https://unix.stackexchange.com/questions/25361/why-does-unix-store-timestamps-in-a-signed-integer) for example.

mrferris: Oh gotcha, I didn't know about the `-1` representing an error.

I would think that [this function](https://github.com/ethereum/trin/blob/73f3e1e6152de09cd56ad060aecbe3635701e94e/portalnet/src/events.rs#L167) should return a `-1` then, instead of a `0` in case of an error when getting the duration. Otherwise the signed-ness and `-1` case appear unused.

ogenev: `millis_to_epoch` makes sense to not be negative, because it is a duration from an anchor in time `UNIX_EPOCH`. 
`-1` is still used in the implementation of `From<i64>` .

I understand that we can get away with `u64`  but I think it is better to stick with the UNIX standards of timestamps which is `i64`.

PR #794, opened 2023-07-12
Make rpc use our wrapped NodeId and move some serialize tests

KolbyML:
### What was wrong?
https://github.com/ethereum/portal-hive/issues/60 caused by https://github.com/ethereum/trin/pull/669
### How was it fixed?
Moving the serialization tests (only a few for now) to a ``tests/serialize.rs``. This is meant to be a clear distinction of what these tests are for. If we ever wrap something in the future, it is likely we will also remove the file altogether since we are no longer using the wrapper. Putting theses type of tests in there own place with a common theme should better indicate there longevity, since whatever is replacing them should be able to pass the same tests.

This is meant to be a quick solution till ``sigp/enr & sigp/discv5`` update there releases with the new serialization changes. Then we can remove our wrapper altogether
### To-Do

- [ ] Clean up commit history


tests/serialize.rs
```diff
@@ -0,0 +1,129 @@
+// tests all Portal Network JSON-RPC Specification types
```
njgheorghita, line 1: I don't agree with moving these tests here. So far in this codebase we've followed a fairly regular convention of keeping unit tests in the same file as the types themselves. It would be confusing to break that convention for these few serde tests. Imo - `tests/` directory should be solely for the peertest launching logic. 

Personally, I would move these tests back to their original location.

njgheorghita (commented):
Any idea how long it'll take before the new `enr` library is published? I

KolbyML:
> Any idea how long it'll take before the new `enr` library is published?

I don't know if we are fine waiting (I am fine and mike is fine) I can close this PR I am just getting sentiment if this is something we want or we are fine waiting.

tests/serialize.rs
KolbyML: Eventually we are going to delete ``ethportal-api/src/types/node_id.rs`` when sigp make there releases. Where will we put the test to test NodeId serializes correct in our current format? What if we update discv5 and they change there library what if it breaks something. We should keep this test, but why are we going to have a file called node_id.rs with only one test in ``ethportal-api/src/types/node_id.rs`` ?

my concerns
- If we remove that file and the test then if we update discv5 it might break our rpc by serializing data wrong.
- It doesn't make sense to have a file in types which tests serialization for 1 type and nothing else.
- It is currently hard to track if we serialize all the the json rpc correctly or to keep track we have tests for the all. We currently have to read through like 20-30 different files. This leaves a lot of room to be missing these tests we should have or room to accidentally delete them.

So my point is testing json-rpc seralization spec coherence shouldn't be in the files of the types themselves as it doesn't make sense for rpc seralization tests to be strung across 20-30 files. It does make sense for specific tests to those types to be in those 30 files, but I think this case is different. 

Also it isn't a few tests we would be writing tests for 22 minimum (definitly more) [``ContentDescriptors``](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) I just moved a few as a starting place.

I am fine with either, I just think this will make it a lot easier to track/make sure we are encoding these 22+ types right

ogenev:
> > Any idea how long it'll take before the new `enr` library is published?
> 
> I don't know if we are fine waiting (I am fine and mike is fine (I believe)) I can close this PR I am just getting sentiment if this is something we want or we are fine waiting.

Waiting for this means that portal-hive tests will continue to fail and we are not sure how long this will take, so I"m more in favor of fixing the issue now.

tests/serialize.rs
ogenev: I don't have a strong opinion, I see some benefits of having all serialization/deserialization tests in one file but I think this file should be in `ethportal-api` not in the global tests folder.

However, given the current uncertainty, I prefer the minimal change possible which is leaving the tests in their original place.

njgheorghita: I'm confused... The unit tests you relocated in this pr have nothing to do with `NodeId` and are unit tests for serde on `HistoryContentKey` / `HistoryContentValue`. IMO, those should definitely remain in the `content_key.rs` file - where we implemented the serde logic ourselves.

We shouldn't have unit tests for logic that we do not maintain. There is an element of trust when using a library. It's completely reasonable to assume that a library has correctly implemented it's advertised functionality. If we run into a bug, we can re-evaluate whether or not we want to depend on that library or maintain our own version of the logic.

Maybe you can help me understand better, but you're looking for a kind of integration test to validate that our serde is working? Is this not something that we're already indirectly testing by using json-rpc in all of our peertest scenarios?

njgheorghita:
I pinged the sigp folks to see if we can get an eta on an `enr` release... so let's wait and hear back before moving ahead with this

PR #795, opened 2023-07-13
Prevent lookup_enr failing prematurely if internal FindNodes fails

KolbyML:
### What was wrong?
lookup_enr can prematurely fail if FindNodes errors out. It shouldn't be errored if that is the case, it should continue trying the other ways of finding an enr
### How was it fixed?
by using a if let Ok instead of using a ? (a try statement in rust). This prevents the potential case of lookup_enr failing prematurely


njgheorghita (approved):
:shipit:

PR #796, opened 2023-07-18
Add content validation result metrics

njgheorghita:
### What was wrong?
Add metrics reporting whether or not received content is being successfully validated or not. 

### How was it fixed?
Added metric reporting logic. Update to grafana dashboard template will come in a following pr along with other dashboard changes.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portalnet/src/metrics/overlay.rs
+++ b/portalnet/src/metrics/overlay.rs
@@ -26,6 +26,7 @@ pub struct OverlayMetrics {
     message_count: IntCounterVec,
     utp_outcome_count: IntCounterVec,
     utp_active_count: IntGaugeVec,
+    validation_count: IntCounterVec,
 }
 
 impl OverlayMetrics {
@@ -58,11 +59,22 @@ impl OverlayMetrics {
             utp_active_count_labels,
         );
 
+        let validation_count_options = opts!(
+            "trin_validation_total",
+            "count all content validations successful and failed"
+        );
+        let validation_count_labels = &["protocol", "success"];
+        let validation_count = OverlayMetrics::register_counter_metric(
+            validation_count_options,
+            validation_count_labels,
+        );
+
         Self {
             protocol: protocol.into(),
             message_count,
             utp_outcome_count,
             utp_active_count,
+            validation_count,
         }
     }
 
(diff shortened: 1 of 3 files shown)


KolbyML (approved):

portalnet/src/metrics/overlay.rs
```diff
@@ -216,12 +244,16 @@ impl OverlayMetrics {
     pub fn get_message_summary(&self) -> String {
         // for every offer you made, how many accepts did you receive
         // for every offer you received, how many accepts did you make
+        let successful_validations = self.validation_count_by_outcome(true);
+        let failed_validations = self.validation_count_by_outcome(false);
         format!(
-            "offers={}/{}, accepts={}/{}",
+            "offers={}/{}, accepts={}/{}, successful_validations={}/{}",
```
carver, line 250: This log line gets quite long, so maybe we can bend towards shorter labels here, like:
```suggestion
            "offers={}/{}, accepts={}/{}, validations={}/{}",
```

carver (approved):
`validate_content` is used so many places, it seems like it would be easy to miss one of them during a refactor or a new feature. Maybe we could pass the metric object into the metrics and increment it inside the validator once, instead of at every call site.

Or maybe a utility method that calls the validator, records the metric and returns the result, so the match calls can be replaced? Then we would generally make sure that no new direct calls to `validate_content` are created, though.

github tells me there are 9 call sites altogether, and I only see 4 places where the PR reports the validation result. So maybe there are still some missing?

njgheorghita:
@carver

> github tells me there are 9 call sites altogether

If I'm not missing something, the other call sites are from within tests, in which case it doesn't seem worthwhile to manage metrics

> validate_content is used so many places, it seems like it would be easy to miss one of them during a refactor or a new feature. 

I agree, it's not ideal, and it's easy to miss something. I do see this as a concern. 

But, this is how we handle **all** metrics inside `overlay.rs` / `overlay_protocol.rs`, just inserting them individually whenever a metric needs reporting. I'm uncertain about treating just the validation metric specially. And I'm not sure there's a better way to handle the metrics as a whole in this situation.

wrt the idea of passing metrics into the `validate_content()` method.
1. It'll require some finagling/dead code in the tests.
2. We'll have to import the `OverlayMetrics` arg in the `trin-validation`/`trin-beacon`/`trin-state`/`trin-history` crates.
3. It feels messy / poor separation of responsibility to let only one of the metrics bleed into other parts of the codebase

> Or maybe a utility method that calls the validator..

I went with a bit of a hybrid solution here. Added a utility method inside `overlay.rs`, but I opted to leave `overlay_service.rs` as-is. Mostly because that file is already quite complex, and I don't see much benefit of adding a utility function there. 

I'm going to go ahead with the merge here since you gave this pr :+1: - but I'm happy to come back to this if you have any further thoughts.

PR #797, opened 2023-07-18
Fix FindContent Enr failing to serialize and add test

KolbyML:
### What was wrong?
I was trying to write tests for portal-hive https://github.com/ethereum/portal-hive/pull/61

And I notice FindContent returning ENRS would fail everytime
### How was it fixed?
I did some debugging and found out it was an issue with serializing. So I used serde derivied serialization and it worked like a charm.
### To-Do

- [ ] Clean up commit history


ogenev (approved):
Good catch! :100:

PR #798, opened 2023-07-19
Add consensus light client types

ogenev:
Sorry for the huge PR, but most of the diffs are test specs yaml files. If you prefer, you can review the PR commit by commit.

### What was wrong?
No implementations of the light client consensus types in our repo. We need them to implement the beacon portal network [specs](https://github.com/ethereum/portal-network-specs/blob/master/beacon-chain/beacon-network.md#types).

### How was it fixed?
As discussed with Kim. we want to support `Bellatrix`(Altair) and `Capella` type variants for the beacon portal network. This will also make it feasible to switch to the next fork `Deneb`.

- introduce the [superstruct](https://github.com/sigp/superstruct) crate to deal with type variants based on the fork versions.
- Refactor the consensus test assets folder and add test cases for every implemented type. All tests are sourced from the [consensus-spec-test ](https://github.com/ethereum/consensus-spec-tests/tree/master)repo.
- Move the `ExecutionPayload` consensus type to a separate file.
- Add `ExecutionPayloadHeader` consensus type.
- Add `LightClientHeader` consensus type.
- Add `LightClientBootstrap` consensus type.
- Add `LightClientUpdate` consensus type.
- Add `LightClientOptimisticUpdate` consensus type.
- Add `LightClientFinalityUpdate` consensus type.

### To-Do

- [x] Clean up commit history


KolbyML (approved):
I took a look through it and it looks very clean. I didn't look at the test data files, but I did look at the actual code itself. There isn't really anything that stands out as off. I am going to hit approve, cause if anything is wrong it can be quickly fixed either way and beacon is high in R&D. It is easier to get this code in and fix it if a issue occurs (I don't think there will be, but it is about the mentality). Then blocking your work while you wait for a review because it is a big scary PR.

So I guess I am saying this PR looks good :+1:

PR #800, opened 2023-07-20
Update discv5 version to 0.3.1 (Fixing NodeId Serialization)

KolbyML:
### What was wrong?
Update discv5 version which has fixes for NodeID serialiation
### How was it fixed?
By updating the dependency.

``let string = base64::encode_config(bytes, base64::URL_SAFE);`` was switched to ``let string = base64::encode_config(bytes, base64::URL_SAFE_NO_PAD);`` since we just used a wrapped of sigp/enr Enr::from_str which uses "URL_SAFE_NO_PAD" I am not sure why we used URL_SAFE, but regardless if ``Enr::from_str()`` doesn't use URL_SAFE us using it will indeed throw an error, so if we want that we have to implement ``Enr::from_str()`` ourselves

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1639,9 +1639,9 @@ checksum = "212d0f5754cb6769937f4501cc0e67f4f4483c8d2c3e1e922ee9edbe4ab4c7c0"
 
 [[package]]
 name = "discv5"
-version = "0.3.0"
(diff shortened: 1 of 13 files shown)


ogenev (approved):
:boat:

PR #802, opened 2023-07-20
Filters disconnected nodes before beginning queries

mrferris:
### What was wrong?

`sigp/discv5` provides the `kbucket::closest_values` method, but the `ClosestValue` output of this method does not include information about the connection state of the nodes on which we can filter. So we haven't been filtering disconnected nodes.

Currently out of the 16 nodes that our queries begin with, only an average of 3 respond, which is  **19%**

### How was it fixed?

Iterating over the kbuckets directly and filtering out disconnected nodes, sorting, then taking the top 16.

After this change, an average of 13 / 16 respond which is **81%** 

Testing was done manually using a local glados instance and comparing batches of audit trace results with and without the change.

### To-Do

- [x] Breakout into a function and use the same function in find nodes
- [x] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2194,6 +2194,33 @@ where
         Ok(closest_nodes)
     }
 
+    /// Returns a vector of ENRs of the `max_nodes` closest connected nodes to the target from our routing table.
+    fn closest_connected_nodes(&self, target_key: &Key<NodeId>, max_nodes: usize) -> Vec<Enr> {
+        // Filter out all disconnected nodes
+        let kbuckets = self.kbuckets.read();
+        let mut all_nodes: Vec<&kbucket::Node<NodeId, Node>> = kbuckets
+            .buckets_iter()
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -2194,26 +2194,53 @@ where
...
     ) -> Option<QueryId> {
         let target_key = Key::from(*target);
-        let mut closest_enrs: Vec<Enr> = self
-            .kbuckets
-            .write()
-            .closest_values(&target_key)
-            .map(|closest| closest.value.enr)
-            .collect();
 
-        // `closest_enrs` will be empty if `target` is our local node ID
```
ogenev, line 2211: Do you know if this is still a valid case we must take care of?

mrferris: Yes, we still have to take care of this when all nodes are disconnected, good catch

ogenev (approved):
LGTM :+1:

PR #803, opened 2023-07-20
Add support for utp over findcontent

njgheorghita:
### To-Do

- [ ] Clean up commit history


njgheorghita:
sorry for the fake heads-up on the review. i'm running out of time for today but will look into the failing test case tm and have this ready for :eyes:

mrferris (changes requested):
We need to call `advance_find_content_query_with_content` so that the query will conclude

njgheorghita:
Closing this in favor of #807 since this became a bit of a mess as I was experimenting with different approaches. But I'll track all the comments here and make sure they're addressed over in the other pr

PR #806, opened 2023-07-24
ethportal-api: Refactor portal content keys and values architecture

ogenev:
### What was wrong?
All content key and value types, including history, state, beacon, and overlay were in the same files. This adds unnecessary mental overheat when adding new tests and new types for the beacon network.

### How was it fixed?
- Create a `content_key`  folder and add the types into separate files per network.
- Create a `content_value`  folder and add the types into separate files per network.

### To-Do

- [x] Clean up commit history


ethportal-api/src/types/content_key/beacon.rs
```diff
@@ -0,0 +1,136 @@
...
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct LightClientBootstrapKey {
+    /// Hash of the block.
+    pub block_hash: [u8; 32],
+}
+
+/// Key used to identify a set of light client updates.
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct LightClientUpdatesKey {
+    pub start_period: u64,
```
njgheorghita, line 28: Maybe add brief docstring comments here describing the fields

ethportal-api/src/types/content_key/beacon.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for BeaconContentKey {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let data = String::deserialize(deserializer)?.to_lowercase();
+        let first_two = &data[..2];
```
njgheorghita, line 113: might be a bit cleaner to use [starts_with](https://doc.rust-lang.org/std/primitive.str.html#method.starts_with)

ethportal-api/src/types/content_key/history.rs
```diff
@@ -0,0 +1,376 @@
...
+
+impl<'de> Deserialize<'de> for HistoryContentKey {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let data = String::deserialize(deserializer)?.to_lowercase();
+        let first_two = &data[..2];
+
+        if first_two != "0x" {
```
njgheorghita, line 46: I'm guessing this was just copy/pasted, but maybe update this to use `starts_with`

ethportal-api/src/types/content_key/history.rs
```diff
@@ -0,0 +1,376 @@
...
+                input: hex_encode(ssz_bytes),
+            })
+            .map_err(serde::de::Error::custom)
+    }
+}
+
+/// A key for a block header.
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct BlockHeaderKey {
+    /// Chain identifier.
```
njgheorghita, line 66: Seems like this docstring is now outdated, let's remove it

ethportal-api/src/types/content_value/constant.rs
```diff
@@ -0,0 +1,3 @@
+/// The length of the Merkle proof for the inclusion of a block header in a particular epoch
```
njgheorghita, line 1: Well, there's only one constant in this file, but usually in our codebase we've used the filename `constants.rs`, so I'd update it to remain consistent throughout the codebase, plus it seems reasonable that we'll add more constants to this file eventually

ethportal-api/src/types/content_value/value.rs
```diff
@@ -0,0 +1,9 @@
+use crate::ContentValueError;
```
njgheorghita, line 1: *nitpick* maybe just move this definition to `mod.rs`?

njgheorghita (approved):
:shipit:

PR #808, opened 2023-07-25
Add beacon network content types and keys

ogenev:
### What was wrong?
No content types implementation for the beacon portal network.

### How was it fixed?
Add beacon network content values and keys based on the [specs](https://github.com/ethereum/portal-network-specs/blob/master/beacon-chain/beacon-network.md) and [test vectors](https://github.com/ethereum/portal-network-specs/pull/219).

### To-Do

- [ ] Clean up commit history


KolbyML (approved):
:shipit: PR looks good

ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -99,6 +452,19 @@ impl<'de> Deserialize<'de> for BeaconContentValue {
...
             return Ok(Self::HistoricalSummariesWithProof(value));
         }
+        if let Ok(value) = ForkVersionedLightClientBootstrap::from_ssz_bytes(&content_bytes) {
+            return Ok(Self::LightClientBootstrap(value));
+        }
+        if let Ok(value) = LightClientUpdatesByRange::from_ssz_bytes(&content_bytes) {
+            return Ok(Self::LightClientUpdatesByRange(value));
+        }
+        if let Ok(value) = ForkVersionedLightClientOptimisticUpdate::from_ssz_bytes(&content_bytes)
+        {
```
carver, line 462: *nit* woah, did `rust fmt` bump the `{` here? It's making me wince to have every other block include the { at the end of the if, and only on this one to bump it to the next line...

ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -51,23 +67,348 @@ impl<'de> Deserialize<'de> for PossibleBeaconContentValue {
...
+
+    fn ssz_bytes_len(&self) -> usize {
+        self.encode().len()
+    }
+}
+
+/// Maximum number of `LightClientUpdate` instances in a single request is 128;
+/// Defined in https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/p2p-interface.md#configuration
+#[derive(Clone, Debug, PartialEq)]
+pub struct LightClientUpdatesByRange(VariableList<ForkVersionedLightClientUpdate, U128>);
```
carver, line 344: *nit* it would be nice to group up the `ForkVersionedLightClientUpdate` near this definition, instead of having the ...`FinalityUpdate` split them up. A code reader is probably interested in `ForkVersionedLightClientUpdate` and `LightClientUpdatesByRange` at the same time.

carver (approved):
Yup, looks good!

It's a bummer that the test json values are so large and opaque. I don't really have a suggestion to improve it right now besides maybe add a comment with a link to the portal-specs at a particular commit that include the test value? Then people can theoretically verify that it's the same value as the specified one, or maybe even notice if something has changed by diffing that commit against the same file in the spec master.

ethportal-api/src/types/content_value/beacon.rs
ogenev: yep, it's cargo fmt :innocent:

ethportal-api/src/types/jsonrpc/endpoints.rs
```diff
@@ -65,7 +65,7 @@ pub enum HistoryEndpoint {
 }
 
 /// Beacon network JSON-RPC endpoints. Start with "portal_beacon" prefix
-#[derive(Debug, PartialEq, Eq, Clone)]
```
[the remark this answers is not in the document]
ogenev: Yeah, we are not using it and it adds more complexity because all of the inner types must implement `Eq`  which is more restrictive and requires some additional tweaks.

ogenev:
> It's a bummer that the test json values are so large and opaque. I don't really have a suggestion to improve it right now besides maybe add a comment with a link to the portal-specs at a particular commit that include the test value? Then people can theoretically verify that it's the same value as the specified one, or maybe even notice if something has changed by diffing that commit against the same file in the spec master.

It is a bummer that the `json` file format doesn't allow comments and we started using such files some time ago for testing. I will add the specs permalink in the commit comment but If we want a more reasonable solution, maybe we need to rethink how we want to use JSON objects for testing.

PR #809, opened 2023-07-26
Handle utp connection id for recursive find content lookups

njgheorghita:
### What was wrong?
We weren't handling find content responses that required utp.

### How was it fixed?
- Validate data that's received from a `Content` response
- asyncify some `OverlayService` methods so that they can be used across threads.
  - `process_received_content`
  - `poke_content`
- Some modifications to the query pool
  - Handle all content processing (validation / storing) in the final stage of a find content query lifecycle
- Added some tests

### To-Do

- [x] Clean up commit history

#### in follow up prs

- [ ] boot peer if content validation fails
- [x] add utpTransfer field to return type
- [ ] add validation status to return type (?) 
- [ ] rename overlay_service -> service & protocol
- [x] unique content values for each test run
  - [x] add test recursive utp with trace scenario
- [ ] remove utp handling from protocol 
- [ ] prepend node id to peertest logs
 - https://github.com/ethereum/utp/blob/be858a8665276a4cff1c0c9bb78f55a1d87d0809/src/stream.rs#L42
- [ ] remove rwlock usage in favor of channels (see comment)
- [ ] https://github.com/ethereum/trin/pull/809/files#r1278470587
- [ ] https://github.com/ethereum/trin/pull/809/files/c4bff3fff4c683ec9cf4d0e1c2fa9ccd20463bac#r1278472910

### DIFF
--- a/portalnet/src/find/iterators/findcontent.rs
+++ b/portalnet/src/find/iterators/findcontent.rs
@@ -32,6 +32,7 @@ use super::query::{Query, QueryConfig, QueryPeer, QueryPeerState, QueryProgress}
 pub enum FindContentQueryResponse<TNodeId> {
     ClosestNodes(Vec<TNodeId>),
     Content(Vec<u8>),
+    ConnectionId(u16),
 }
 
 #[derive(Debug)]
@@ -39,13 +40,31 @@ pub enum FindContentQueryResult<TNodeId> {
     ClosestNodes(Vec<TNodeId>),
     Content {
         content: Vec<u8>,
-        closest_nodes: Vec<TNodeId>,
+        nodes_to_poke: Vec<TNodeId>,
+    },
+    Utp {
+        connection_id: u16,
+        nodes_to_poke: Vec<TNodeId>,
+        // peer used to create utp stream
+        peer: TNodeId,
     },
 }
 
 #[derive(Debug, Clone)]
-struct ContentAndPeer<TNodeId> {
+enum ContentAndPeer<TNodeId> {
+    Content(ContentAndPeerDetails<TNodeId>),
+    Utp(UtpAndPeerDetails<TNodeId>),
+}
+
+#[derive(Debug, Clone)]
+struct ContentAndPeerDetails<TNodeId> {
     content: Vec<u8>,
+    peer: TNodeId,
+}
+
+#[derive(Debug, Clone)]
+struct UtpAndPeerDetails<TNodeId> {
+    connection_id: u16,
     peer: TNodeId,
 }
 
@@ -166,11 +185,16 @@ where
                 };
             }
             FindContentQueryResponse::Content(content) => {
-                // Incorporate the found content into the query.
-                self.content = Some(ContentAndPeer {
+                self.content = Some(ContentAndPeer::Content(ContentAndPeerDetails {
                     content,
                     peer: peer.clone(),
-                });
+                }));
+            }
+            FindContentQueryResponse::ConnectionId(connection_id) => {
+                self.content = Some(ContentAndPeer::Utp(UtpAndPeerDetails {
+                    connection_id: u16::from_be(connection_id),
+                    peer: peer.clone(),
+                }));
             }
         }
     }
@@ -292,36 +316,30 @@ where
     }
 
     fn into_result(self) -> Self::Result {
-        match self.content {
-            Some(content) => {
-                let closest_nodes = self
-                    .closest_peers
-                    .into_values()
-                    .filter_map(|peer| {
-                        if let QueryPeerState::Succeeded = peer.state() {
-                            // Do not include the peer who returned the content.
-                            if *peer.key().preimage() == content.peer {
-                                None
-                            } else {
-                                Some(peer.key().clone().into_preimage())
-                            }
-                        } else {
-                            None
-                        }
-                    })
-                    .take(self.config.num_results)
-                    .collect();
-
-                FindContentQueryResult::Content {
-                    content: content.content,
-                    closest_nodes,
+        match self.content.clone() {
+            Some(content) => match content {
+                ContentAndPeer::Content(val) => {
+                    let nodes_to_poke = self.get_nodes_to_poke(&val.peer);
+                    FindContentQueryResult::Content {
+                        content: val.content,
+                        nodes_to_poke,
+                    }
                 }
-            }
+                ContentAndPeer::Utp(val) => {
+                    let nodes_to_poke = self.get_nodes_to_poke(&val.peer);
+                    FindContentQueryResult::Utp {
+                        connection_id: val.connection_id,
+                        nodes_to_poke,
+                        peer: val.peer,
+                    }
+                }
+            },
             None => {
                 let closest_nodes = self
                     .closest_peers
                     .into_values()
                     .filter_map(|peer| {
+                        // we don't filter out the source peer when returning the closest nodes
                         if let QueryPeerState::Succeeded = peer.state() {
                             Some(peer.key().clone().into_preimage())
                         } else {
@@ -388,6 +406,26 @@ where
             QueryProgress::Iterating { .. } => self.num_waiting >= self.config.parallelism,
             QueryProgress::Finished => true,
         }
+    }
+
+    /// Returns the nodes to poke after a successful query.
+    /// Do not include the peer who returned the content.
+    fn get_nodes_to_poke(&self, source_peer: &TNodeId) -> Vec<TNodeId> {
+        self.closest_peers
+            .iter()
+            .filter_map(|(_, peer)| {
+                if let QueryPeerState::Succeeded = peer.state() {
+                    if peer.key().preimage() == source_peer {
+                        None
+                    } else {
+                        Some(peer.key().clone().into_preimage())
+                    }
+                } else {
+                    None
+                }
+            })
+            .take(self.config.num_results)
+            .collect()
     }
 }
 
@@ -459,13 +497,11 @@ mod tests {
 
         let result = query.into_result();
         match result {
-            FindContentQueryResult::Content { .. } => {
-                panic!("Unexpected result variant from new query")
-            }
             FindContentQueryResult::ClosestNodes(closest_nodes) => assert!(
                 closest_nodes.is_empty(),
                 "Unexpected closest peers in new query"
             ),
+            _ => panic!("Unexpected result variant from new query"),
         }
     }
 
@@ -571,17 +607,16 @@ mod tests {
             match result {
                 FindContentQueryResult::Content {
                     content,
-                    closest_nodes,
+                    nodes_to_poke,
                 } => {
-                    let closest_nodes =
-                        closest_nodes.into_iter().map(Key::from).collect::<Vec<_>>();
-                    assert!(sorted(&target_key, &closest_nodes));
+                    let nodes_to_poke =
+                        nodes_to_poke.into_iter().map(Key::from).collect::<Vec<_>>();
+                    assert!(sorted(&target_key, &nodes_to_poke));
 
                     let content_peer = content_peer.unwrap();
 
-                    // The peer who returned the content should not be included in the closest
-                    // nodes.
-                    assert!(!closest_nodes.contains(&content_peer));
+                    // The peer who returned the content should not be included in the poke nodes
+                    assert!(!nodes_to_poke.contains(&content_peer));
 
                     assert_eq!(content, found_content);
                 }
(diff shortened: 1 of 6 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -758,58 +758,147 @@ where
...
+                        peer,
+                        nodes_to_poke,
+                    } => {
+                        let metrics = self.metrics.clone();
+                        let utp = self.utp_socket.clone();
+                        let id = u16::from_be(connection_id);
+                        let key = kbucket::Key::from(peer);
+                        let source = match self.kbuckets.write().entry(&key) {
+                            kbucket::Entry::Present(entry, _) => entry.value().enr(),
+                            _ => {
```
mrferris, line 815:
```suggestion
                        let source = match self.find_enr(&peer) {
                            Some(enr) => enr,
                            None => {
```

I was getting a lot of `Received uTP payload from unknown peer` warnings before making this change

njgheorghita: Cool, nice catch. `find_enr` seems like the better mechanism here. We also do `kbuckets.write()....` inside `poke_content` so it seems like it might be worth updating that also to use `find_enr`?

njgheorghita: @mrferris Ahh, nvm I was wrong, inside `poke_content` we don't have access to `self` so `find_enr()` is inaccessible

portalnet/src/overlay_service.rs
```diff
@@ -758,58 +758,146 @@ where
...
+                                            UtpOutcomeLabel::FailedConnection,
+                                        );
+                                        warn!(
+                                            %err,
+                                            cid.send,
+                                            cid.recv,
+                                            peer = ?cid.peer.client(),
+                                            "Unable to establish uTP conn based on Accept",
+                                        );
+                                        return;
```
carver, line 845: Just thinking this through: If someone offers content and then the connection step fails, we just stop trying to collect the content, right? It will time out eventually, I guess. I suppose we could at least "time out" right away so the caller can re-run the call if they want.

In an ideal world, we would probably continue asking other peers for the same content. 👍🏻 to put something like this in an issue, for follow-up work. Plus maybe a comment here and the other early returns, noting the current behavior.

portalnet/src/overlay_service.rs
```diff
@@ -758,58 +758,146 @@ where
...
+                                    }
+                                };
+
+                            let mut data = vec![];
+                            if let Err(err) = stream.read_to_eof(&mut data).await {
+                                metrics.report_utp_outcome(
+                                    UtpDirectionLabel::Inbound,
+                                    UtpOutcomeLabel::FailedDataTx,
+                                );
+                                error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), "error reading data from uTP stream");
```
carver, line 855: We almost never want to have identical log lines in multiple places in the code. Otherwise, tracking down a surprising log is harder, because it has multiple possible causes.

Plus, it seems like an opportunity to add more context, like "stream, while handling a FindContent query".

portalnet/src/overlay_service.rs
```diff
@@ -758,58 +758,146 @@ where
...
+                                UtpOutcomeLabel::Success,
+                            );
+
+                            let trace = query_info.trace;
+                            Self::process_received_content(
+                                kbuckets,
+                                command_tx,
+                                validator,
+                                store,
+                                (data.clone(), true),
```
carver, line 871: Can't this be moved instead of cloned? This isn't even just a stylistic question, since this data can be a pretty significant size: cloning it might have a measurable performance impact.

portalnet/src/overlay_service.rs
```diff
@@ -1692,27 +1767,19 @@ where
...
             protocol = %self.protocol,
             response.source = %source.node_id(),
             "Processing Content message",
         );
         match content {
-            Content::ConnectionId(id) => debug!(
-                protocol = %self.protocol,
-                "Skipping processing for content connection ID {}",
-                u16::from_be(id)
-            ),
```
carver, line 1712: Learning moment: this log should have probably been a `warn!` or higher, to indicate that core functionality was not implemented. We would have noticed it much earlier...

portalnet/src/overlay_service.rs
```diff
@@ -1692,27 +1767,19 @@ where
...
+                if let Some(query_id) = query_id {
+                    self.advance_find_content_query_with_connection_id(&query_id, source, id);
+                }
+            }
             Content::Content(content) => {
-                self.process_received_content(content.clone(), request);
-                // TODO: Should we only advance the query if the content has been validated?
                 if let Some(query_id) = query_id {
                     self.advance_find_content_query_with_content(&query_id, source, content);
                 }
```
carver, line 1785: In what situation would `query_id` be `None`? It feels strange to get this `Content()` message and then drop it on the floor without any log if the id is None.

portalnet/src/overlay_service.rs
```diff
@@ -1727,67 +1794,63 @@ where
...
+        store: Arc<RwLock<TStore>>,
+        content: (Vec<u8>, bool),
+        content_key: TContentKey,
+        responder: Option<oneshot::Sender<FindContentResult>>,
+        trace: Option<QueryTrace>,
+        nodes_to_poke: Vec<NodeId>,
+    ) {
+        // Spawn task that validates content before storing.
+        // Allows for non-blocking requests to this/other overlay services.
+        tokio::spawn(async move {
```
carver, line 1811: So yeah, based on a previous comment, I'd be curious to see if we can make things work smoothly, if this spawn is removed.

portalnet/src/overlay_service.rs
```diff
@@ -1727,67 +1794,63 @@ where
...
+            let content_id = content_key.content_id();
+            if let Err(err) = validator.validate_content(&content_key, &content.0).await {
```
carver, line 1813: Previously, the content was only validated after checking if it's in-radius and missing from the DB. That seems like the right choice. Why was it changed to happen before that check?

I guess it's because we might be poking other nodes with it, but I think there are better solutions. Something like:

```
if is_locally_stored:
  # ignore retrieved value (unless we want to punish the peer later, we can validate with `db.get(key) == content`
  content = db.get(key)
elif in_radius:
  validate(content)
  store(content)
poke(content)
```

carver (approved):

portalnet/src/overlay_service.rs
```diff
@@ -758,58 +758,146 @@ where
...
+                                        metrics.report_utp_outcome(
+                                            UtpDirectionLabel::Inbound,
+                                            UtpOutcomeLabel::FailedConnection,
+                                        );
+                                        warn!(
+                                            %err,
+                                            cid.send,
+                                            cid.recv,
+                                            peer = ?cid.peer.client(),
+                                            "Unable to establish uTP conn based on Accept",
```
mrferris, line 843:
```suggestion
                                            "Unable to establish uTP conn based on Content response",
```

mrferris (approved):

portalnet/src/overlay_service.rs
```diff
@@ -3657,6 +3776,71 @@ mod tests {
...
+            actual_connection_id,
+        );
+
+        let mut pool = service.find_content_query_pool.write();
+        let (_, query) = pool
+            .get_mut(query_id)
+            .expect("Query pool does not contain query");
+
+        // Query result should contain content.
+        match query.clone().into_result() {
```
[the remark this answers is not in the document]
njgheorghita: it looks like `into_result()` takes ownership of `query` and in this case `query` is behind a mut ref so the clone is required to take ownership

portalnet/src/find/iterators/findcontent.rs
```diff
@@ -602,6 +646,7 @@ mod tests {
                         assert_eq!(num_results, closest_nodes.len(), "Too  many results.");
                     }
                 }
+                _ => panic!("Unexpected result."),
```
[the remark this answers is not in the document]
njgheorghita: Since you mention "a crashed trin" makes me think that you might not have noticed this wildcard (and the earlier wildcard) are inside tests? When it comes to application code, I definitely agree that wildcards are poor patterns to rely upon, however when it comes to tests I tend to reach for them quite frequently (I'm sure I've left plenty around our codebase). It's less code clutter within each test and if the test fails I've found that it's typically not too complicated to hunt down the culprit.

portalnet/src/find/iterators/findcontent.rs
```diff
@@ -459,13 +506,11 @@ mod tests {
...
         let result = query.into_result();
         match result {
-            FindContentQueryResult::Content { .. } => {
-                panic!("Unexpected result variant from new query")
-            }
             FindContentQueryResult::ClosestNodes(closest_nodes) => assert!(
                 closest_nodes.is_empty(),
                 "Unexpected closest peers in new query"
             ),
+            _ => panic!("Unexpected result variant from new query"),
```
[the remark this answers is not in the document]
njgheorghita: See response on the other wildcard comment.

portalnet/src/overlay_service.rs
```diff
@@ -3657,6 +3776,71 @@ mod tests {
...
+            .write()
+            .insert_or_update(&bootnode_key, bootnode, status);
+
+        let target_content = NodeId::random();
+        let target_content_key = IdentityContentKey::new(target_content.raw());
+
+        let query_id = service.init_find_content_query(target_content_key, None, false);
+        let query_id = query_id.expect("Query ID for new find content query is `None`");
+
+        // update query in own span so mut ref is dropped after poll
```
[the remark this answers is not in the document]
njgheorghita: :+1: and this comment is more evidence to move away from `RwLock`, since it took me quite a while to discover that this lock wasn't being dropped and blocking the following code.

tests/self_peertest.rs
```diff
@@ -83,6 +83,7 @@ mod test {
         peertest::scenarios::validation::test_validate_pre_merge_block_body(&peertest, &target)
             .await;
         peertest::scenarios::validation::test_validate_pre_merge_receipts(&peertest, &target).await;
+        peertest::scenarios::find::test_recursive_utp(&peertest).await;
```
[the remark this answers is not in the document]
njgheorghita: Yup! And verified that it fails if I remove the `ConnectionId` handling code now.

portalnet/src/find/iterators/findcontent.rs
```diff
@@ -293,30 +319,51 @@ where
...
+                    }
                 }
-            }
+                ContentAndPeer::Utp(val) => {
+                    let nodes_to_poke = self
+                        .closest_peers
+                        .into_values()
+                        .filter_map(|peer| {
+                            if let QueryPeerState::Succeeded = peer.state() {
+                                Some(peer.key().clone().into_preimage())
```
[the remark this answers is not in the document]
njgheorghita: Actually, I think you're instinct is correct here. I can't think of any reason why we should not filter out the source peer for `nodes_to_poke` in either situation. Looking a little more closely, it looks like I might have copied this logic from below (the `None` arm) where it does make sense to include the source peer in the case that we return a list of closest nodes (although, now that I think about it, if the request fails maybe we shouldn't include the source peer?). But I've added a comment and moved the shared code into a separate method.

portalnet/src/overlay_service.rs
```diff
@@ -284,9 +284,9 @@ pub struct OverlayService<TContentKey, TMetric, TValidator, TStore> {
     /// A map of active outgoing requests.
     active_outgoing_requests: Arc<RwLock<HashMap<OverlayRequestId, ActiveOutgoingRequest>>>,
     /// A query pool that manages find node queries.
-    find_node_query_pool: QueryPool<NodeId, FindNodeQuery<NodeId>, TContentKey>,
+    find_node_query_pool: Arc<RwLock<QueryPool<NodeId, FindNodeQuery<NodeId>, TContentKey>>>,
```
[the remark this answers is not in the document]
njgheorghita: Agreed, see the other comment as evidence for moving away from the `RwLock` pattern. Also agreed that it'd be a fairly significant refactor, so adding it to the todo list in this comment's info - which I'm using to track the various tidying up tasks I'd like to tackle. (this list will move into it's own issue shortly)

portalnet/src/overlay_service.rs
njgheorghita: Honestly, I'm not intimately familiar with the query pool inner workings. @mrferris might have a better idea. Regardless, I've added this to my list of things to look into while refactoring, since I agree that it's not a intuitive code path here.

portalnet/src/find/iterators/findcontent.rs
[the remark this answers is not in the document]
carver: Yup, you're right. I missed that these were in tests!

PR #812, opened 2023-07-31
Add utp transfer to find content return

njgheorghita:
### What was wrong?
Add support for boolean `utpTransfer` field in json responses to find content & recursive find content jsonrpc requests.

### How was it fixed?
Repurpose `ContentInfo` response type to accommodate both find content and recursive find content responses. As well as adding the `utp_transfer` field.

### To-Do

- [x] support inside `trin-beacon`
- [x] add new field to spec (https://github.com/ethereum/portal-network-specs/pull/221)
- [x] Clean up commit history

### DIFF
--- a/trin-history/src/jsonrpc.rs
+++ b/trin-history/src/jsonrpc.rs
@@ -1,18 +1,17 @@
 use std::sync::Arc;
 
 use discv5::enr::NodeId;
-use ethportal_api::types::content_value::ContentValue;
 use ethportal_api::types::{
     constants::CONTENT_ABSENT, jsonrpc::endpoints::HistoryEndpoint,
     jsonrpc::request::HistoryJsonRpcRequest, query_trace::QueryTrace,
 };
 use ethportal_api::utils::bytes::hex_encode;
-use ethportal_api::OverlayContentKey;
 use ethportal_api::{
     types::portal::{AcceptInfo, ContentInfo, FindNodesInfo, PongInfo, TraceContentInfo},
-    ContentValue, {HistoryContentKey, OverlayContentKey},
+    ContentValue, {HistoryContentKey, OverlayContentKey, RawContentKey},
 };
 use portalnet::storage::ContentStore;
+use portalnet::types::messages::Content;
 use serde_json::{json, Value};
 use ssz::Encode;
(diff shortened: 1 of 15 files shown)


njgheorghita:
https://github.com/ethereum/portal-network-specs/pull/221 is merged so this is no longer blocked and ready for :eyes:

trin-history/src/jsonrpc.rs
```diff
@@ -124,11 +123,15 @@ async fn recursive_find_content(
 
     // If tracing is not required, return content.
     if !is_trace {
-        return Ok(content_response_string);
+        return Ok(json!(ContentInfo::Content {
+            content: serde_json::from_value(content_response_string).map_err(|e| e.to_string())?,
```
carver, line 127: Something here seems really off. Are we encoding to json string in `content_response_string`, then decoding it to re-encode it? ... I'm seeing now that maybe the same thing is happening in the trace below, but that also seems wrong.

Does this work?
```rust
    // Handle empty content
    let content_string = match possible_content_bytes {
        Some(bytes) => hex_encode(bytes),
        None => CONTENT_ABSENT.to_string(), // "0x"
    };

    // If tracing is not required, return content.
    if !is_trace {
        return Ok(json!(ContentInfo::Content {
            content: content_string,
```

trin-validation/src/oracle.rs
```diff
@@ -51,18 +52,37 @@ impl HeaderOracle {
...
-        let hwp_ssz = hwp_ssz
-            .as_str()
-            .ok_or_else(|| anyhow!("Invalid HWP format."))?;
-        let hwp_ssz = hex_decode(hwp_ssz)?;
-        HeaderWithProof::from_ssz_bytes(&hwp_ssz)
-            .map_err(|err| anyhow!("Invalid HWP received from chain history network: {err:?}"))
+        let hwp = match serde_json::from_value(hwp)? {
+            ContentInfo::Content { content, .. } => content,
+            ContentInfo::ConnectionId { .. } => {
+                return Err(anyhow!(
```
carver, line 64: I think you might be able to do roughly the same pattern as my comment before with "Nice cleanup".

carver (approved):
No blockers!

PR #814, opened 2023-08-01
Validate find_content result in non-utp transfers & use fresh testnet for each peertest scenario

njgheorghita:
### What was wrong?
Our testing setup for peertest was not ideal. We spin up a small testnet of nodes and use the same testnet for every scenario. This makes testing difficult & non-intuitive, when side-affects from certain scenarios can affect the outcome of later scenarios. Especially relevant for find content tests. 

Initially, I thought that it would be better to simply use unique test values for each scenario. However, this is an unsustainable approach, as we will need to continually generate new test values for each scenario we add. It turns out that the time overhead for spinning up a new peertest for each scenario is really not that bad, and avoids the unneeded complexity of maintaining unique content key/values for each scenario.

### How was it fixed?
- Updated any relevant scenarios to use a fresh peertest testnet.
- Added validation for content retrieved via a simple `FindContent` request.
- Improved header validation to also validate that its hash matches the hashed content.
- Added some test scenarios

### To-Do

- [x] bridge pandaops batch limit of 100
- [x] Clean up commit history

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ discv5 = { version = "0.3.1", features = ["serde"] }
 eth2_ssz = "0.4.0"
 ethereum-types = "0.12.1"
 ethportal-api = { path = "ethportal-api" }
+jsonrpsee = "0.16.2"
 parking_lot = "0.11.2"
 portalnet = { path = "portalnet" }
 prometheus_exporter = "0.8.4"
--- a/ethportal-peertest/src/constants.rs
+++ b/ethportal-peertest/src/constants.rs
@@ -1,3 +1,46 @@
+use ethportal_api::{HistoryContentKey, HistoryContentValue};
+use serde_json::json;
+
+/// History HeaderWithProof content key & value
+/// Block #1000010
+pub fn fixture_header_with_proof_1000010() -> (HistoryContentKey, HistoryContentValue) {
+    let content_key: HistoryContentKey =
+        serde_json::from_value(json!(HISTORY_CONTENT_KEY)).unwrap();
+    let content_value: HistoryContentValue =
+        serde_json::from_value(json!(HISTORY_CONTENT_VALUE)).unwrap();
+    (content_key, content_value)
+}
+
+/// History HeaderWithProof content key & value
+/// Block #14764013 (pre-merge)
+pub fn fixture_header_with_proof() -> (HistoryContentKey, HistoryContentValue) {
+    let content_key: HistoryContentKey =
+        serde_json::from_value(json!(HEADER_WITH_PROOF_CONTENT_KEY)).unwrap();
+    let content_value: HistoryContentValue =
+        serde_json::from_value(json!(HEADER_WITH_PROOF_CONTENT_VALUE)).unwrap();
+    (content_key, content_value)
+}
+
+/// History BlockBody content key & value
+/// Block #14764013 (pre-merge)
+pub fn fixture_block_body() -> (HistoryContentKey, HistoryContentValue) {
+    let content_key: HistoryContentKey =
+        serde_json::from_value(json!(BLOCK_BODY_CONTENT_KEY)).unwrap();
+    let content_value: HistoryContentValue =
+        serde_json::from_value(json!(BLOCK_BODY_CONTENT_VALUE)).unwrap();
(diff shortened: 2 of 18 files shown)


ethportal-peertest/src/constants.rs
```diff
@@ -1,3 +1,46 @@
...
+
+/// History HeaderWithProof content key & value
+/// Block #1000010
+pub fn test_item() -> (HistoryContentKey, HistoryContentValue) {
+    let content_key: HistoryContentKey =
+        serde_json::from_value(json!(HISTORY_CONTENT_KEY)).unwrap();
+    let content_value: HistoryContentValue =
+        serde_json::from_value(json!(HISTORY_CONTENT_VALUE)).unwrap();
+    (content_key, content_value)
+}
```
KolbyML, line 12: nitpick ``test_item()`` isn't that descriptive of a name. I have 2 ideas which could improve readability.

- use ``test_header_with_proof()`` instead and remove ``test_item()`` since they are both pre-merge header with proofs so maybe we no longer need ``test_item()``?
- try to use a more descriptive function name maybe like ``test_header_with_proof_block_1000010()`` it is a little long but readability is very nice.

This is a nitpick so if you want to leave it how it is already that is fine

Cargo.toml
```diff
@@ -19,6 +19,7 @@ discv5 = { version = "0.3.1", features = ["serde"] }
 eth2_ssz = "0.4.0"
 ethereum-types = "0.12.1"
 ethportal-api = { path = "ethportal-api" }
+jsonrpsee = {version="0.16.2", features = ["async-client", "client", "macros", "server"]}
```
KolbyML, line 22:
```suggestion
jsonrpsee = "0.16.2"
```

The features aren't needed to compile/run the testsuite

KolbyML (approved):
:partying_face:

carver:
> * bridge pandaops batch limit of 100

Not sure I know what this is, but it's unchecked. Does it need to be moved to a new issue?

carver (approved):
LGTM!

I like the shorter test fixture access with the methods.

Also, yay for running more validations! I updated the PR title, so that we can see at a glance that this includes the validations, which is worth mentioning at updates.

ethportal-peertest/src/constants.rs
```diff
@@ -1,3 +1,46 @@
+use ethportal_api::{HistoryContentKey, HistoryContentValue};
+use serde_json::json;
+
+/// History HeaderWithProof content key & value
+/// Block #1000010
+pub fn test_item() -> (HistoryContentKey, HistoryContentValue) {
```
ogenev, line 6: As we deal with more than one network, I would like to see some distinction between the tests and the networks. For example, we can rename this file to `history_tests` or something or extract the test methods in a separate file.

ogenev (approved):

KolbyML (approved):

ethportal-peertest/src/scenarios/validation.rs
```diff
@@ -53,45 +44,59 @@ pub async fn test_validate_pre_merge_header_with_proof(peertest: &Peertest, targ
...
+        block_hash: H256::random().into(),
+    });
+
+    let store_result = peertest
+        .bootnode
+        .ipc_client
+        .store(invalid_content_key.clone(), content_value.clone())
         .await
         .unwrap();
+    assert!(store_result);
```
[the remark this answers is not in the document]
njgheorghita: IIUC, this is the desired behavior for this endpoint. It's essentially a "testing" endpoint, so maybe we should move it to a `debug_`? But the endpoint's purpose is simply to insert data into the db without any validation.

ethportal-peertest/src/constants.rs
njgheorghita: I agree. But, this feels a bit out of scope for this pr. There does seem to be some important refactoring to be done inside peertest. Some tests can be shared across networks, while others are network specific. IMO, a big question we need to resolve before refactoring is do we / don't we support running different groupings of subnetworks? (eg history&state&beacon or history&beacon or history, etc). This has subtle but important ramifications for a lot of our architecture and now that we're starting to expand into multiple subnetworks I think we should settle on a decision.

Issue #687, opened 2023-04-21
Add `unwrap_used` lint warning to all workspaces


njgheorghita:
From https://github.com/ethereum/trin/pull/681#discussion_r1173387002 it appears like we're not linting for naked unwraps in all workspaces.

### How to fix it?
Add `#![warn(clippy::unwrap_used)]` to `lib.rs` for any workspace that doesn't already contain this lint. Example: https://github.com/ethereum/trin/blob/master/trin-history/src/lib.rs. And handle any unwraps that might be lurking

dhruvdabhi101:
I would like to contribute.

Eren-Yeaager:
@njgheorghita  Hi ,I would like to contribute

ogenev:
You are free to contribute, don't ask for permission :smile_cat: .

PR #819, opened 2023-08-07
Add unwrap_used lint and remove all naked unwrap usage 

Perelyn-sama:
### What was wrong?
https://github.com/ethereum/trin/issues/687
After I added the lint and ran `cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings` I found out there were a lot of places in the codebase that used `unwrap` besides from tests.

### How was it fixed?
I added the unwrap_used lint.

I also corrected the places that used `unwrap` with the `expect` method. I tried to use short and precise messages that could provide information about code failure but in places, I could not come up with anything I just used "operation failed". 

I also used the `?` operator and the `unwrap_or_default` method in places I could. 

On line 218 `light-client/src/client.rs`, I could not come up with anything so I used the  `#[allow(clippy::unwrap_used)]` attribute.

### To-Do

- [ ] Clean up commit history


ogenev (changes requested):
We don't need to handle naked unwraps for `ethportal-peertest` and `utp-testing` because those crates are used only for testing.

The error messages in `light-client` are too vague, we need to be more specific when using expect. However, I would remove also the requirement from the `light-client` crate because it is still in experimental state  and we will refactor it anyway pretty soon.

Perelyn-sama:
Okay, I guess I'll add the `[allow(clippy::unwrap_used)]` attribute to `ethportal-peertest` and `utp-testing`

Admittedly, some of the`expect` messages are vague, but what if right now, we just keep them as placeholders to be changed later on? My intention for adding them was just to stop the CI test from failing and there's no harm in keeping them for now since they don't have much of an impact on the codebase like `unwrap`.

PR #822, opened 2023-08-10
portal-bridge: Add support for beacon light client data

### LINKED ISSUE #820: `portal-bridge` support for beacon light client data
Add functionalities in `portal-bridge` to allow injecting beacon light client data into a test network.

ogenev:
### What was wrong?

We want to seed light client data into the beacon network. Current `portal-bridge` functionalities do not support beacon data. 

Closes #820 

### How was it fixed?
- Add a  `BeaconBridge` instance with functionalities related to the beacon network.
- Based on a `network` cli flag, the bridge launch beacon or/and history bridge instances on separate tokio threads.
- Support only the `latest` mode for now. Test mode and peertest scenario will be added in another PR because it requires some additional refactorings which will extend the scope of this PR (#848 ).
- Add support for the following light client data: `LightClientUpdatesByRange`, `LightClientOptimisticUpdate`, and `LightClientFInalityUpdate`.  Support for `LightClientBootstrap` will be added in a following PR (#849) because it requires some more research to find the most feasible way to deliver the latest bootstrap available.

### Known issues
- https://github.com/ethereum/trin/issues/850

### To-Do

- [x] Clean up commit history


ethportal-api/src/types/consensus/execution_payload.rs
```diff
@@ -18,9 +19,13 @@ pub struct ExecutionPayload {
     pub receipts_root: H256,
     pub logs_bloom: Bloom,
     pub prev_randao: H256, // 'difficulty' in the yellow paper
+    #[serde(deserialize_with = "as_u64")]
```
carver, line 22: I don't think this needs to be in the code, but can you give a brief explanation here of why we need this new library to parse u64?

carver (commented):
Hey, I only got to do a partial review before wrapping up today, so I just posted the comments I got to.

From a high level, the main thing I noticed was a lack of tests. Is there really nothing that makes sense to test here?

If you can't come up with anything, when I come back tomorrow, I'll see if I have any ideas

ethportal-api/src/types/consensus/execution_payload.rs
ogenev: > I don't think this needs to be in the code, but can you give a brief explanation here of why we need this new library to parse u64?

This allows us to deserialize the fields, `block_number` for example, from string, not only from a number. This is required because we receive the data from pandaops as JSON strings.

ogenev:
> From a high level, the main thing I noticed was a lack of tests. Is there really nothing that makes sense to test here?

I don't have great ideas other than adding an integration test in ethportal-peertest #848 . Most of the data is based on an internal clock and requests to pandaops. I guess we can mock the `ConsensusApi` but then I'm not sure what benefits we gain in regard to testing.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -0,0 +1,237 @@
...
+use serde_json::Value;
+use ssz::Encode;
+use ssz_types::VariableList;
+use std::cmp::Ordering;
+use std::sync::atomic::AtomicUsize;
+use std::sync::Arc;
+use std::thread::sleep;
+use tracing::{info, warn};
+
+pub struct BeaconBridge {
```
njgheorghita, line 28: nitpick: I haven't really seen us ever have non-pub fields on a pub struct. It seems like if a type is `pub` then every attribute should also be `pub`, unless there's a good reason for them not to be `pub` in which case I'd leave a comment as to why they're not `pub` so it doesn't get changed in the future.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -0,0 +1,237 @@
...
+            mode,
+            portal_clients,
+        }
+    }
+
+    pub async fn launch(&self) {
+        info!("Launching beacon bridge mode: {:?}", self.mode);
+
+        match self.mode.clone() {
+            BridgeMode::Latest => self.launch_latest().await,
```
njgheorghita, line 47: It feels like a custom `BridgeMode` might be appropriate here? Eventually, do you plan to support the same convention inside `BeaconBridge` that currently exists inside `BridgeMode`? eg. `--backfill:e100`, `--single:b10000`, etc.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -0,0 +1,237 @@
...
+use tracing::{info, warn};
+
+pub struct BeaconBridge {
+    pub api: ConsensusApi,
+    mode: BridgeMode,
+    portal_clients: Vec<HttpClient>,
+}
+
+impl BeaconBridge {
+    pub fn new(api: ConsensusApi, mode: BridgeMode, portal_clients: Vec<HttpClient>) -> Self {
```
njgheorghita, line 35: Something to think about, but not urgent to address is how you'd like to do metrics here? Can they both use the same shared types? And how `BeaconBridge` metrics will be reported alongside the regular `Bridge` metrics.

njgheorghita (approved):
🌉 I'm not sure what your plan is for getting this live. Do you want to use this feature in our live network bridge nodes? Or keep it experimental & just use locally until it's a bit more battle-tested?

portal-bridge/src/utils.rs
```diff
@@ -69,6 +71,36 @@ pub fn read_test_assets_from_file(test_path: PathBuf) -> TestAssets {
...
+/// Gets the duration until the next light client update
+/// Updates are scheduled for 4 seconds into each slot
+pub fn duration_until_next_update(genesis_time: u64) -> Duration {
+    let current_slot = expected_current_slot(genesis_time);
+    let next_slot = current_slot + 1;
+    let next_slot_timestamp = slot_timestamp(next_slot, genesis_time);
+
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("Time went backwards")
```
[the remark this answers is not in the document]
carver: "Timestamp overflow" is probably the more likely scenario, though I suppose it would happen if someone set their computer time to 1969, too.

It's not super urgent obviously, but as far as I understand, if someone runs trin on a computer with 32-bit timestamps in 2038, this will panic.

portal-bridge/src/consensus_api.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub async fn get_lc_bootstrap(&self, block_root: String) -> anyhow::Result<String> {
+        let endpoint = format!(
+            "{}/eth/v1/beacon/light_client/bootstrap/{}",
+            self.middleware.base_cl_endpoint, block_root
+        );
+        self.middleware.request(endpoint).await
+    }
+
+    /// Requests the LightClientUpdate instances in the sync committee period range
+    /// [start_period, start_period + count), leading up to the current head sync committee period
```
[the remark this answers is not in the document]
carver: Actually, I think this is correct. The interval is open on the upper end. If the start period is 10, and the count is 2, then `[10, 12)` means only include 10 and 11.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -0,0 +1,237 @@
...
+            "Got lc update for slot {:?}",
+            update.attested_header.beacon.slot
+        );
+
+        // Update the current known period if we successfully gossiped the latest data.
+        match Self::gossip_beacon_content(&portal_clients, content_key, content_value).await {
+            Ok(_) => {
+                current_period.store(
+                    expected_current_period as usize,
+                    std::sync::atomic::Ordering::Relaxed,
```
carver, line 171: I'm still figuring out the details of the atomic ordering semantics. So yeah, I'll admit some ignorance there. How sure are you that `Relaxed` reads and writes won't cause some kind of weird access order bug? I can't think of a specific problem, but I tend to just use the strictest-possible option, unless I take the time to prove to myself that it's definitely not a problem.

carver (approved):
> I don't have great ideas other than adding an integration test in ethportal-peertest https://github.com/ethereum/trin/issues/848

Cool, yeah, that integration test, plus some utility function cases sound great 👍🏻

portal-bridge/src/beacon_bridge.rs
ogenev: Yes, eventually we will need something for backfilling but I prefer to implement it when we have the exact use case and need.

portal-bridge/src/beacon_bridge.rs
ogenev: Good point about more restrictiveness. Changed to `Aquire` and `Release` according to https://stackoverflow.com/questions/30407121/which-stdsyncatomicordering-to-use.

portal-bridge/src/utils.rs
```diff
@@ -69,6 +71,36 @@ pub fn read_test_assets_from_file(test_path: PathBuf) -> TestAssets {
 
     assets
 }
+/// Gets the duration until the next light client update
+/// Updates are scheduled for 4 seconds into each slot
+pub fn duration_until_next_update(genesis_time: u64) -> Duration {
```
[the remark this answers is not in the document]
ogenev: Did some refactoring and added tests for those functions.

ogenev:
> 🌉 I'm not sure what your plan is for getting this live. Do you want to use this feature in our live network bridge nodes? Or keep it experimental & just use locally until it's a bit more battle-tested?

My plan is to first test it locally for a bit and then push it to the live bridge nodes.

Issue #821, opened 2023-08-09
Bug: "Error receiving content ... error=oneshot canceled"


carver:
> 2023-08-04T01:18:47.445374Z  WARN portalnet::overlay: Error receiving content from service protocol=History error=oneshot canceled content.id=0x79ba16c1c2bac068f5ccfb4dd72f2f39fd06434007fbc7c5fdb9bb2da958b7c2

Produced here (and seen in practice): https://portal-hive.ethdevops.io/viewer.html?file=results/trin/client-236f41433a53c24c88097f7d553655ef24769cc31cb2d267b5b318b09f51852f.log

Some to-dos:
- [ ] what exactly is happening?
- [ ] can we log something that more clearly explains the situation?
- [ ] it's causing an actual bug in portal-hive, how can that be prevented? (somewhat orthogonal)

KolbyML:
@carver yes and I am looking into why FC and RFC are timing out on a uTP transfer

KolbyML:
This issue can be closed after https://github.com/ethereum/trin/pull/825 is merged

PR #825, opened 2023-08-11
Respond to query if no nodes available

mrferris:
### What was wrong?
Queries were timing out when no nodes were available for the query. This was always the case, but having no nodes available  became something that started happening when #802 was merged, which only considers connected nodes.

### How was it fixed?

1. Use closest nodes regardless of connectivity if all nodes are in a disconnected state
2. If no nodes whatsoever are in our routing table, send a "no content found" response immediately instead of never responding. Currently change 1 makes it so that this should never happen, but better safe than sorry.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2402,7 +2402,35 @@ where
             peer_timeout: self.query_peer_timeout,
         };
 
-        let closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        let mut closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        if closest_enrs.is_empty() {
+            // If there are no connected nodes, use the closest nodes regardless of connectivity.
+            warn!("No connected nodes, using disconnected nodes for query.");
+            closest_enrs = self
+                .kbuckets
+                .write()
+                .closest_values(&target_key)
+                .map(|closest| closest.value.enr)
+                .take(query_config.num_results)
+                .collect();
+        }
+
+        if closest_enrs.is_empty() {
+            // If there are no nodes in the routing table the query cannot proceed.
+            warn!("No nodes in routing table, query cannot proceed.");
+            if closest_enrs.is_empty() {
+                if let Some(callback) = callback {
+                    let _ = callback.send((None, false, None));
+                }
+                return None;
+            }
+        }
+
+        // Convert ENRs into k-bucket keys.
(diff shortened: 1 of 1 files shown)


mrferris:
Note that the unit test that I added fails on master with a `canceled` error (like in #821 ), and it passes with this fix.

KolbyML (approved):
This is nice :+1:

portalnet/src/overlay_service.rs
```diff
@@ -2402,7 +2402,34 @@ where
             peer_timeout: self.query_peer_timeout,
         };
 
-        let closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        let mut closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        if closest_enrs.is_empty() {
+            // If there are no connected nodes, use the closest nodes regardless of connectivity.
+            closest_enrs = self
+                .kbuckets
+                .write()
```
carver, line 2410: Hm, I was wondering why this needs to be a write lock when it seems to be a read-only query. I'm assuming you tried a read lock first. Assuming you did, I think the `&mut self` on `closest_values` is why it's required. Though I can't tell from looking at the source why it couldn't be a `&self` instead:
https://github.com/sigp/discv5/blob/1439decd4e7d7c9de78ef61b5d67be3fee688510/src/kbucket.rs#L650-L670

Might be worth seeing if it's as easy as removing the `mut` in a draft PR against discv5. (To be clear, I *don't* mean that getting the discv5 change should block this PR).

portalnet/src/overlay_service.rs
```diff
@@ -2402,7 +2402,34 @@ where
             peer_timeout: self.query_peer_timeout,
         };
 
-        let closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        let mut closest_enrs = self.closest_connected_nodes(&target_key, query_config.num_results);
+        if closest_enrs.is_empty() {
+            // If there are no connected nodes, use the closest nodes regardless of connectivity.
```
carver, line 2407: This is an concerning enough situation that it seems worth a `warn!` log (or maybe at least `info!`). Having 0 known connected nodes is noteworthy.

carver (approved):
Yup, looks great. Yay for red/green testing!

portalnet/src/overlay_service.rs
mrferris: I wondered the same. 

I'll see if that discv5 change breaks anything after merging.

PR #828, opened 2023-08-11
docs: Release checklist

carver:
### What was wrong?

The section for releasing trin could use some love.

### How was it fixed?

Loved it.

Expanded the number of things to check during release. Split the release into three stages.

Bonus: improved the section on how to contribute to our mdbook docs.

### To-Do

- [ ] Clean up commit history


carver:
I'm making a judgement call to just merge. It's a docs-only change, and we can keep iterating forward to make it better.

PR #829, opened 2023-08-12
Upgrade utp-rs to v0.1.0-alpha.6 & remove mutex on discv5 transport layer for utp

### LINKED ISSUE #827: Upgrade utp-rs to latest
After v0.1.0-alpha.6 is released, upgrade to it in trin.

carver:
Handle new APIs:
- &mut self in socket trait methods
- shutdown() -> async close()

### What was wrong?

Fix #827 

utp library recently had a new release: https://github.com/ethereum/utp/releases/tag/v0.1.0-alpha.6

A (now) unnecessary mutex is wrapped around the discv5 talkreqs in the utp transport layer.

### How was it fixed?

Upgrade utp-rs, fix compile errors related to breaking changes. Also remove the mutex.

### To-Do

- [ ] Clean up commit history
- [x] Remove the mutex

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -279,18 +279,13 @@ impl Discovery {
 }
 
 pub struct Discv5UdpSocket {
-    // `Mutex` for interior mutability.
-    // TODO: Figure out a better mechanism here. The socket is the only holder of the lock.
-    talk_reqs: tokio::sync::Mutex<mpsc::UnboundedReceiver<TalkRequest>>,
+    talk_reqs: mpsc::UnboundedReceiver<TalkRequest>,
     discv5: Arc<Discovery>,
 }
 
 impl Discv5UdpSocket {
     pub fn new(discv5: Arc<Discovery>, talk_reqs: mpsc::UnboundedReceiver<TalkRequest>) -> Self {
-        Self {
-            discv5,
-            talk_reqs: tokio::sync::Mutex::new(talk_reqs),
-        }
+        Self { discv5, talk_reqs }
     }
 }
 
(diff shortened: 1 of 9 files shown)


portalnet/src/discovery.rs
```diff
@@ -281,16 +281,13 @@ impl Discovery {
 pub struct Discv5UdpSocket {
     // `Mutex` for interior mutability.
     // TODO: Figure out a better mechanism here. The socket is the only holder of the lock.
-    talk_reqs: tokio::sync::Mutex<mpsc::UnboundedReceiver<TalkRequest>>,
+    talk_reqs: mpsc::UnboundedReceiver<TalkRequest>,
```
njgheorghita, line 284: It seems to me like the above comments are now outdated.

portalnet/src/overlay_service.rs
```diff
@@ -1588,7 +1588,7 @@ where
             }
 
             // close uTP connection
-            if let Err(err) = stream.shutdown() {
+            if let Err(err) = stream.close().await {
```
njgheorghita, line 1591: IIUC, `close()` now confirms that all of the data was sent correctly? (Rather than just pushed into the send buffer which was the behavior for `shutdown()`). If so, it looks like we can remove some of the logic inside [`handle_find_content()`](https://github.com/ethereum/trin/blob/master/portalnet/src/overlay_service.rs#L1110-L1149) where we check the `write_size` to validate that a utp tx was successful.

njgheorghita (approved):
🚢

portalnet/src/overlay_service.rs
carver: Oh yeah, this called for a wider refactor: the duplicate code in handle_find_content had already drifted significantly. So I combined the two usages into a single function call. This also helped us get rid of the tracking boolean flag for whether we had recorded the utp outcome.

PR #831, opened 2023-08-14
chore: Upgrade Rust builds in CI to 1.71.1

carver:
### What was wrong?

New stable version of rust out that we don't compile on. Someone who downloads the latest stable rust and trin might experience failed trin builds.

### How was it fixed?

Upgrade rust versions everywhere, and fix related compile errors.

### To-Do

- [ ] Clean up commit history


njgheorghita (approved):
:ship:

PR #835, opened 2023-08-15
Add yaml support to portal-bridge

KolbyML:
### What was wrong?
json doesn't support comments I want to have comments in my test data so I know what the numbers are/mean
### How was it fixed?
by adding serde-yaml and and a match case on the file extension.


portal-bridge/src/bridge.rs
```diff
@@ -79,9 +79,18 @@ impl Bridge {
...
 
     async fn launch_test(&self, test_path: PathBuf) {
-        let test_asset = fs::read_to_string(test_path).expect("Error reading test asset.");
-        let assets: TestAssets =
-            serde_json::from_str(&test_asset).expect("Unable to parse test asset.");
+        let test_asset = fs::read_to_string(test_path.clone()).expect("Error reading test asset.");
+        let extension = test_path
+            .extension()
+            .and_then(|x| x.to_str())
+            .expect("Unable to parse test path extension");
```
carver, line 86: I believe the `.clone()` can be avoided by flipping the order of these two. The reason I think so is because `PathBuff::extension()` just takes a reference to the path, but the `read_to_string()` consumes the object.

```suggestion
        let extension = test_path
            .extension()
            .and_then(|x| x.to_str())
            .expect("Unable to parse test path extension");
        let test_asset = fs::read_to_string(test_path).expect("Error reading test asset.");
```

carver (commented):
It doesn't have to be extensive, but some basic test showing this working would be a big improvement. I assume you already tried it locally, but even if you're sure it works now, it's best to have a test in to prevent regressions in the future.

I see a bridge test in ethportal-peertest/src/scenarios/bridge. It looks like you could translate that data file to yaml and run the test against the new copy.

portal-bridge/src/bridge.rs
KolbyML:
```rust
error[E0505]: cannot move out of `test_path` because it is borrowed
  --> portal-bridge/src/bridge.rs:87:45
   |
81 |       async fn launch_test(&self, test_path: PathBuf) {
   |                                   --------- binding `test_path` declared here
82 |
83 |           let extension = test_path
   |  _________________________-
84 | |             .extension()
   | |________________________- borrow of `test_path` occurs here
...
87 |           let test_asset = fs::read_to_string(test_path).expect("Error reading test asset.");
   |                                               ^^^^^^^^^ move out of `test_path` occurs here
88 |
89 |           let assets: TestAssets = match extension {
   |                                          --------- borrow later used here

For more information about this error, try `rustc --explain E0505`.
```

I could do this if you want
```rust
        let extension = test_path
            .extension()
            .and_then(|x| x.to_str())
            .expect("Unable to parse test path extension").to_string();
        let test_asset = fs::read_to_string(test_path).expect("Error reading test asset.");

        let assets: TestAssets = match &extension[..] {
            "json" => serde_json::from_str(&test_asset).expect("Unable to parse json test asset."),
            "yaml" => serde_yaml::from_str(&test_asset).expect("Unable to parse yaml test asset."),
            _ => panic!("Invalid file extension"),
        };
        ```

Comment (KolbyML) [reply portal-bridge/src/bridge.rs:86]: I think I can use 2 expects instead of the and then I will try that

Comment (KolbyML) [reply portal-bridge/src/bridge.rs:86]: Nvm I just tried the idea I said and it still says a move happened

**Comment by KolbyML** [comment]
@carver Ok I added that test you wanted, and commented on the issue you had with the clone

File: portal-bridge/src/bridge.rs [hunk 2]
```diff
@@ -79,9 +79,19 @@ impl Bridge {
...
         let test_asset = fs::read_to_string(test_path).expect("Error reading test asset.");
-        let assets: TestAssets =
-            serde_json::from_str(&test_asset).expect("Unable to parse test asset.");
+
+        let assets: TestAssets = match &extension[..] {
+            "json" => serde_json::from_str(&test_asset).expect("Unable to parse json test asset."),
+            "yaml" => serde_yaml::from_str(&test_asset).expect("Unable to parse yaml test asset."),
+            _ => panic!("Invalid file extension"),
+        };
+
```
ogenev, line 94: I would like to extract this block into `utils` because It will be helpful in the beacon network bridge (which will be in a separate file) and maybe the state network bridge.

ogenev (approved):

ogenev:
I think `yaml` files are way better than `json` for test data (because they can include comments). At some point, we may want to switch all our `json` test files to `yaml` in` test_assets`.

portal-bridge/src/bridge.rs
```diff
@@ -79,9 +79,19 @@ impl Bridge {
     }
 
     async fn launch_test(&self, test_path: PathBuf) {
+        let extension = test_path
+            .extension()
+            .and_then(|x| x.to_str())
```
njgheorghita, line 84: Let's avoid `x` as a variable name, something like `path` would be fine and short

test_assets/portalnet/bridge_data.yaml
```diff
@@ -0,0 +1,2 @@
+- content_key: 0x006251d65b8a8668efabe2f89c96a5b6332d83b3bbe585089ea6b2ab9b6754f5e9
```
njgheorghita, line 1: Well, if we're going to switch to yaml to support comments, let's use some here! Something like at least the block number and also a blurb about where & how this test data is used would be a good start

ethportal-peertest/src/scenarios/bridge.rs
```diff
@@ -31,3 +31,26 @@ pub async fn test_bridge(peertest: &Peertest, target: &HttpClient) {
         "The received content {received_content_value:?}, must match the expected {content_value:?}",
     );
 }
+
+pub async fn test_bridge_yaml(peertest: &Peertest, target: &HttpClient) {
```
njgheorghita, line 35: I agree with @carver that this functionality should be tested, but I somewhat disagree with how. There's no need to actually run the entire bridge scenario again. All the test needs to do is validate that [`TestAssets`](https://github.com/ethereum/trin/blob/master/portal-bridge/src/bridge.rs#L83-L84) can be created from both a yaml and json file. After moving the decoding logic as @ogenev suggested, I would add a test to the utils file that simply validates that either a `yaml` or `json` file can be parsed into the `TestAssets` type.

njgheorghita:
Upon further thought, I think we should either choose yaml or json as our official standard for test data and use it comprehensively throughout our codebase. Supporting both is confusing and adds unnecessary complexity. If we don't feel like making the jump to exclusively using yaml, then maybe that jump is not worth it? (for the record, i'm on board with using yaml exclusively)

test_assets/portalnet/bridge_data.yaml
KolbyML: I don't know what block number you generated this data from

tests/self_peertest.rs
```diff
@@ -248,7 +249,24 @@ mod test {
...
+    async fn peertest_bridge_json() {
+        let (peertest, target, handle) = setup_peertest_bridge().await;
+        peertest::scenarios::bridge::test_bridge_json(&peertest, &target).await;
+        peertest.exit_all_nodes();
+        handle.stop().unwrap();
+    }
+
+    #[tokio::test(flavor = "multi_thread")]
+    #[serial]
+    async fn peertest_bridge_yaml() {
```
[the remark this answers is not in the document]
KolbyML: resolved

portal-bridge/src/bridge.rs
KolbyML: resolved

portal-bridge/src/bridge.rs
KolbyML: resolved

test_assets/portalnet/bridge_data.yaml
KolbyML: I found the block number so Sure I will add a comment on the blocks height

ethportal-peertest/src/scenarios/bridge.rs
KolbyML: resolved

KolbyML:
@njgheorghita I am going to merge this as I resolved the PR concerns. I think we should use YAML, but I don't think deciding that should block this or my other PR. I would argue supporting json/yaml is the same as us supporting http/ipc though and the complexity is going to be less then 20 lines so I think that is very managable.

We can create an issue or bring it up in meeting if we want to look into that.

PR #836, opened 2023-08-16
Add GenerateTestData to Portal-Bridge

KolbyML:
### What was wrong?
Portal-Bridge should be able to generate the test data it needs since it already supports 99% of the logic to create it. I was going to make a python script for this but then did a double take when I realized I would be reimplementing all of portal bridge and portal validation. As well as realizing if we support running bridge with test data it should be able to generate the files we need (manually making test files is no simple task)
### How was it fixed?
Refactoring ``constructor_and_gossip_*`` to only do construction and move the gossip part up stream so we can use constructor for more then just gossiping data to the network like in creating test data.

I also decided to make ``launch_generate_test_data`` handle creating the testfile  and collecting the data instead of using putting the logic through the serve function like backfill does. Serve was written with throughput as a goal where as I want the test data I am creating to be linear from block x to y. This is also good since then it isolates this code instead of leaching the test generation code in every function.

We can generate the test data in both json and yaml format's. When creating tests I believe yaml should be the preference since it allows comments

### DIFF
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -1,7 +1,7 @@
 use crate::full_header::FullHeader;
 use crate::mode::{BridgeMode, ModeType};
 use crate::pandaops::PandaOpsMiddleware;
-use crate::utils::{read_test_assets_from_file, TestAssets};
+use crate::utils::{read_test_assets_from_file, Asset, TestAssets};
 use anyhow::{anyhow, bail};
 use ethportal_api::jsonrpsee::http_client::HttpClient;
 use ethportal_api::types::execution::accumulator::EpochAccumulator;
@@ -14,14 +14,15 @@ use ethportal_api::types::execution::header::{
 };
 use ethportal_api::types::execution::receipts::Receipts;
 use ethportal_api::utils::bytes::hex_encode;
-use ethportal_api::HistoryContentValue;
 use ethportal_api::HistoryNetworkApiClient;
 use ethportal_api::{
     BlockBodyKey, BlockHeaderKey, BlockReceiptsKey, EpochAccumulatorKey, HistoryContentKey,
 };
+use ethportal_api::{ContentValue, HistoryContentValue, OverlayContentKey};
 use futures::stream::StreamExt;
 use ssz::Decode;
 use std::fs;
+use std::io::Write;
 use std::ops::Range;
 use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
@@ -72,6 +73,9 @@ impl Bridge {
         info!("Launching bridge mode: {:?}", self.mode);
         match self.mode.clone() {
             BridgeMode::Test(path) => self.launch_test(path).await,
+            BridgeMode::GenerateTestData(mode, path) => {
+                self.launch_generate_test_data(mode, path).await
+            }
             BridgeMode::Latest => self.launch_latest().await,
             _ => self.launch_backfill().await,
         }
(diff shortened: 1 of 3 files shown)


njgheorghita:
Not a complete review since this is currently blocked. But, I'd like to challenge the motivation for this pr. 

Portal-bridge should be the way to generate test data, I agree. I've used portal-bridge exclusively to generate all the test data that we use. But, I'm not convinced that we need to introduce any logic / complexity to achieve this. 

The way I generate the data is simple. I add a couple of print statements and then run portal-bridge in single block mode. Yes, it's a bit hacky, but it works very well. And I'm simply not convinced that we need to implement code to support this function.

KolbyML:
Currently portal-bridge is only usable by EF employees since it only supports panda ops. As I said above if I wrote a python script for this I would have to rewrite all of portal-bridge + a good chuck of validation utilities. I don't think there is worth writing that python script.

"print statements and then run portal-bridge in single block mode. Yes, it's a bit hacky, but it works very well". If I was to generate a testfile for a epoch I would need 8192 blocks. I would need to write 8192 comments, and copy paste 8192 * 3(content types) * 2 (key/value) = 49152 elements. I don't want to imagine how strained my hands would be after copy-pasting text from a log and spending most likely over 1 whole day of my time creating this potential test file. Active work. This PR is 99% reusing portal-bridge code so the only complexity is a custom yaml format. I think this is well worth the 15-40 lines of new code added by this PR (excluding the refactor) as it will save me over 16 hours of collecting data from single block mode and formatting myself and save my from getting arthritis from the strain of copy-paste-ing so much.

I wouldn't consider the refactor adding complexity. So then the only complexity added would be the YAML format I added which is around 20 lines. I would argue the readability this provides our test files will be well worth the 20 lines we have to maintain. Also the amount of time we will save instead of manually writing the test data files, as well as preventing health concerns from straining motions.


I think portal-bridge should support infura even though it is added complexity so I can use the bridge and for other non-EF employees (This will be my next PR.). I think if we accept a test data input we should be able to generate (since all the code is already here we just have to export it to a file). I think the hours we save manually creating test data will be well worth this investment.

njgheorghita:
Well, I do understand your frustration. But I'm still not convinced that this functionality is worth building out & maintaining. We've had a similar discussion before and the decision was made to build `portal-bridge` for internal use only, rather than user facing. I'd prefer to keep it that way, but I'm open to changing my mind, let's discuss this during sync tm.

carver:
I'm about to sign off today, so I can't get into a deeper review tonight, but I wanted to give the mile-high non-code review:

- based on the description, the separation of concerns described sounds like a valuable refactor all on its own (like separating construction and gossip)
- Kolby wants to make test data, and I want to support him in doing that. I didn't really hear an alternative besides patching in log statements, which isn't very compelling
- How is the change related to needing access to pandaops nodes? Don't we still need access to generate the test data?

---

This line is a nice chance to revisit the PR discussion from the Boulder summit:

> why I think this ... won't be as taxing as you say it will.

Using "you" here makes the comment focus less on the facts and more on just how wrong Nick is 😛 . It's more effective to drop the "you", for example:

> why I think this ... won't be too taxing

The wording change feels tiny, but it can really smooth out the conversation, and keep the focus on the right thing.

---

In short, I think this PR is on the right track, here, unless there's some unmentioned alternative that's even better. We'll talk about it in the morning meeting, and I'll dive into the code tomorrow to see the details.

KolbyML:
I changed this PR to draft since our direction is currently unclear. I did test it and patched the PR locally so I could use it though which was nice. (The patches on another branch for context)

PR #838, opened 2023-08-16
Responds to query if utp peer enr is unknown instead of dropping oneshot

mrferris:
### What was wrong?

When a utp connection ID is received at the end of a content query, trin uses `OverlayService::find_enr` to lookup the sender's ENR by its node ID. Currently if it can't find anything it just throws a warning and lets the oneshot get canceled so the query never completes.

### How was it fixed?

The best thing would be for `find_enr` to find the ENR, which I'm working on. In the meantime this PR just has trin send a response back so that the query completes and we can at least see the trace.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -818,6 +818,9 @@ where
                             Some(enr) => enr,
                             _ => {
                                 warn!("Received uTP payload from unknown {peer}");
+                                if let Some(responder) = callback {
+                                    let _ = responder.send((None, true, query_info.trace));
+                                };
                                 return;
                             }
                         };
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -818,6 +818,9 @@ where
                             Some(enr) => enr,
                             _ => {
                                 warn!("Received uTP payload from unknown {peer}");
+                                if let Some(responder) = callback {
+                                    let _ = responder.send((None, true, query_info.trace));
+                                };
                                 return;
```
njgheorghita, line 824: Nice catch! It looks like we will also want to do something similar on line 845-ish & 861-ish?

njgheorghita: And I checked inside the utp functionality over in `overlay.rs`, it looks like this case is handled there, but might be worth double-checking my checking.

PR #839, opened 2023-08-16
fix: bump rocksdb version to 0.21.0

danielrachi:
### What was wrong?

An [error](https://gist.github.com/danielrachi/376f43eac9fda2e8f07996a25302ddb0) appeared when building on systems with gcc 13.

### How was it fixed?

Bumped rocksdb version from 0.18.0 to 0.21.0, on all crates.

It was necessary to also update a for loop in `src/bin/purge_db.rs` for trin to compile with the updated rocksdb.

### To-Do

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -304,9 +304,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.3.12"
+version = "0.4.1"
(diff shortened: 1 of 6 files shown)


mrferris (approved):
Looks good, works good 👍

ogenev (approved):
LGTM :+1:

PR #841, opened 2023-08-17
Remove restriction on closeness of nodes returned on FindContent

KolbyML:
**blocked till https://github.com/ethereum/portal-network-specs/pull/222 is merged**

### What was wrong?
There is a update to the spec where we no longer ``list of ENR records of nodes that are closer than the recipient is to the requested content.`` and instead do ``list of ENR records of nodes that are closest to the requested content.``
### How was it fixed?
By removing the filter that compared if a nodes distance was closer then our distance or not

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2283,7 +2283,6 @@ where
         let closest_nodes = nodes_with_distance
             .into_iter()
             .take(FIND_CONTENT_MAX_NODES)
-            .filter(|node_record| node_record.0 < self_distance)
             .map(|node_record| SszEnr::new(node_record.1))
             .collect();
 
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -2269,8 +2269,6 @@ where
         content_key: impl OverlayContentKey,
```
njgheorghita, line 2269: The docstring for this method needs to be updated

njgheorghita (approved):
:ship::flamingo:⚡

njgheorghita:
It's not obvious to me that this change requires a test. But, it might be worth including one in `portal-hive` which tests that all clients respect this change in docs?

KolbyML:
https://github.com/ethereum/portal-hive/pull/88

I already have a PR open for it

PR #851, opened 2023-08-21
Bump to discv5 experimental branch for concurrent requests

njgheorghita:
### What was wrong?
Bump discv5 dependency to branch that contains support for concurrent requests.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ description = "A Rust implementation of the Ethereum Portal Network"
 [dependencies]
 anyhow = "1.0.68"
 clap = { version = "4.2.1", features = ["derive"] }
-discv5 = { version = "0.3.1", features = ["serde"] }
+discv5 = { git = "https://github.com/njgheorghita/discv5.git", rev = "09262ead2882e062de1a6a45df9ea61917eb8465", features = ["serde"] }
 eth2_ssz = "0.4.0"
 ethereum-types = "0.12.1"
 ethportal-api = { path = "ethportal-api" }
--- a/ethportal-api/Cargo.toml
+++ b/ethportal-api/Cargo.toml
@@ -15,7 +15,7 @@ anyhow = "1.0.68"
 base64 = "0.13.0"
 bytes = "1.3.0"
 clap = { version = "4.2.1", features = ["derive"] }
-discv5 = { version = "0.3.1", features = ["serde"]}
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -15,7 +15,7 @@ description = "A Rust implementation of the Ethereum Portal Network"
 [dependencies]
 anyhow = "1.0.68"
 clap = { version = "4.2.1", features = ["derive"] }
-discv5 = { version = "0.3.1", features = ["serde"] }
+discv5 = { git = "https://github.com/njgheorghita/discv5.git", rev = "6ab079e60f59b4124e024b1cee0f202efe3f0c87", features = ["serde"] }
```
carver, line 18: different revisions on different packages?

ethportal-api/Cargo.toml
```diff
@@ -15,7 +15,7 @@ anyhow = "1.0.68"
 base64 = "0.13.0"
 bytes = "1.3.0"
 clap = { version = "4.2.1", features = ["derive"] }
-discv5 = { version = "0.3.1", features = ["serde"]}
+discv5 = { git = "https://github.com/njgheorghita/discv5.git", rev = "09262ead2882e062de1a6a45df9ea61917eb8465", features = ["serde"] }
```
carver, line 18: other one starts 6ab

KolbyML (approved):
That is cool I didn't know you could pull by github commit.

njgheorghita:
@KolbyML Yeah, you can also just do it off the branch name, eg: `branch = "experimental"`, but when debugging it's a little more peace of mind to know that you're building off of the exact commit version you want, rather than worrying about whether or not the latest commit to "experimental" was fetched correctly.... Although, typically you'll see the rev hash in the compile logs when building off a branch, but still, the rev method is foolproof imo.

carver (approved):
LGTM! Someday we'll be more cautious about using unmerged/unapproved changes to complex libraries like discv5. But today is not that day!

Reliability went way up according to glados, and it's easy enough to revert if we need to quickly. Eventually, we'll switch back to their released version anyhow.

Thanks for hunting this down! We were not going to get Portal live without parallel discv5 packets...

ogenev (approved):
:rocket:

njgheorghita:
Big shoutout to @ackintosh who did most of the heavy lifting in squashing the final bugs, thanks!

PR #863, opened 2023-08-24
Add github stale issues action

njgheorghita:
### What was wrong?
As I'm combing through our issues list, I'm finding that we don't have the best track record with tracking issues after they've been opened. I've seen a fair number of duplicate issues, outdated bug reports, and issues that were solved but never close. 

The [Stale Issues Action](https://github.com/actions/stale), introduced here will...
- Run daily
- Add a label "Stale" on issues and pull requests after 60 days of inactivity and comment on them
- Close the stale issues and pull requests after 14 days of inactivity
- If an update/comment occur on stale issues or pull requests, the stale label will be removed and the timer will restart

The action itself is very customizable, but imo this default is sufficient to start with and we can adjust it as we see fit.

### How was it fixed?
Introduced github action

### To-Do

- [x] Clean up commit history

flamingo

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "30 1 * * *"
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v5
+        with:
+          days-before-issue-stale: 42
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 60 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          exempt-issue-labels: "shelf-stable"
+          repo-token: ${{ secrets.GITHUB_TOKEN }}


KolbyML:
Just a quick reply not an actual answer.

I think this is a good direction as I can definitly see what you mean about duplicate issues, outdated bug reports, etc. I do think there are "special" issues what should be treated different from this though well the cases are more rare, I do see a special issue tag which would prevent a issue from going stale.

I am fine with whatever happens but I do think this is an interesting precendent.

njgheorghita:
@KolbyML I'm inclined to mostly agree with you here, so I've added the `exempt-issue-labels: "shelf-stable"` rule, so any issue labeled `"shelf-stable"` will be exempt from this bot. (I'm not married to the name at all, just couldn't think of a better name).

I think it'll be a flamingo task to comb all of the recently "stale"d issues and decide whether they're..
- worth moving to a milestone
- worth pinging somebody to check on status (which would automatically reset the stale timer)
- worth adding the "shelf-stable" label

Though my best sense is that we'll really want to avoid using the `"shelf-stable"` label.

mrferris (approved):
Great idea

Though I'd be in favor of lowering the `days-before-issue-stale`, two months seems a bit past stale to me. What if we make it a multiple of seven too, like 42? Six weeks to staleness, eight weeks to closed automatically.

ogenev:
I think this PR needed more discussion across the team, it is quite a significant change in how we handle old issues. I'm not sure if the pros outweigh the cons here and we don't have a well-defined protocol for what to do with the issues marked as stale from the bot.

njgheorghita:
@ogenev Yeah no problem. It was ready for review for over a week, and I figured that was enough time that people would speak up if there was any opposition. But I do agree if we're not all on board with using the bot, or at least the current configuration, then that's something we should discuss and reach consensus on during a sync.

PR #867, opened 2023-08-25
Adds random method to NodeId type

mrferris:
### What was wrong?

`ethportal-api`'s `NodeId` implementation was missing `random()` method, this was causing problems for glados which was using this method in testing

### How was it fixed?

Implements `NodeId::random()`

### To-Do

- [ ] Clean up commit history


njgheorghita (approved):
Nice! I swear we've talked multiple times about switching out our `NodeId` implementation for the discv5 native type... Are we still blocked on something there?

mrferris:
iirc we at one point wanted to have our own implementation...perhaps for serialization purposes. We should revisit this and see if anybody has any reason not to use `discv5`'s

KolbyML:
Our internal NodeID is only still being used because recursiveFindContentTraced failed decoding using sigp/enr. I haven't tested it since, but that is currently why it isn't removed. I changed focuses to Portal-Hive as I seen it has higher priority which is why our NodeId isn't fully removed yet.

njgheorghita:
@KolbyML Nice, thanks for the info. Tracking in #869

Issue #873, opened 2023-08-26
Preferring variable ownership over locks in beacon bridge


carver:
I have some intuition that some refactors in the beacon bridge, `portal-bridge/src/beacon_bridge.rs`, can help us strip away some `Mutex`es, like this one:
```rust
let finalized_block_root = Arc::new(Mutex::new(String::new()));
```

Roughly, the approach would be to move the block root into the spawned async service task, and then return it from that task to re-assign the variable in the outer scope from the returned task value.

I can't do it in my head quite well enough to explain on a PR review, so I'll just code it myself locally to try it out, and show an example.

carver:
I linked an example. There might be more places, so we can leave the issue open until all similar cases are handled.

PR #874, opened 2023-08-26
fix(bridge): non-blocking sleep, to publish beacon every 12s && refactor(bridge): Mutex + Atomic to value-passing

carver:
We can skip variable locking by passing the variable in directly and reassigning it when the method is complete.

### What was wrong?

An example for #873

### DIFF
--- a/portal-bridge/src/beacon_bridge.rs
+++ b/portal-bridge/src/beacon_bridge.rs
@@ -27,9 +27,10 @@ use ssz_types::VariableList;
 use std::cmp::Ordering;
 use std::path::PathBuf;
 use std::sync::Arc;
-use std::thread::sleep;
 use std::time::SystemTime;
 use tracing::{info, warn};
+
+use tokio::time::{interval, sleep, Duration, MissedTickBehavior};
 
 pub struct BeaconBridge {
     pub api: ConsensusApi,
@@ -78,8 +79,27 @@ impl BeaconBridge {
     ///  Get and serve the latest beacon data.
     async fn launch_latest(&self) {
         // Current sync committee period known by the bridge
-        let mut current_period: usize = 0;
-        let mut finalized_block_root = String::new();
+        let (mut current_period, mut finalized_block_root) = Self::serve_latest(
+            self.api.clone(),
+            self.portal_clients.clone(),
+            0,
+            String::new(),
+        )
+        .await;
+
+        // Sleep until next update becomes available. This sets up the interval to update as soon as
+        // the following slot becomes available.
+        let now = SystemTime::now();
+        let next_update = duration_until_next_update(BEACON_GENESIS_TIME, now)
+            .to_std()
+            .expect("failed to convert chrono duration to std duration");
+        sleep(next_update).await;
+
+        // Run the beacon bridge update once every slot
+        let mut interval = interval(Duration::from_secs(12));
+        // If serving takes a little too long, then we want to serve the next one as soon as
+        // possible, but not serve any extras until the following slot. "Skip" gets this behavior.
+        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
         loop {
             let consensus_api = self.api.clone();
@@ -93,12 +113,7 @@ impl BeaconBridge {
             )
             .await;
 
-            let now = SystemTime::now();
-
-            let next_update = duration_until_next_update(BEACON_GENESIS_TIME, now)
-                .to_std()
-                .expect("failed to convert chrono duration to std duration");
-            sleep(next_update);
+            interval.tick().await;
         }
     }
 
--- a/portal-bridge/src/beacon_bridge.rs
+++ b/portal-bridge/src/beacon_bridge.rs
@@ -26,9 +26,7 @@ use serde_json::Value;
 use ssz_types::VariableList;
 use std::cmp::Ordering;
 use std::path::PathBuf;
-use std::sync::atomic::AtomicUsize;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


portal-bridge/src/beacon_bridge.rs
```diff
@@ -78,17 +75,20 @@ impl BeaconBridge {
...
-            let current_period = Arc::clone(&current_period);
 
-            tokio::spawn(async move {
-                Self::serve_latest(consensus_api, portal_clients, current_period).await;
+            let latest_result = tokio::spawn(async move {
+                Self::serve_latest(consensus_api, portal_clients, current_period).await
             });
 
+            current_period = latest_result
+                .await
```
ogenev, line 89: I think this `await` will block the loop until the tokio thread returns a result, which I'm trying to escape.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -78,17 +75,20 @@ impl BeaconBridge {
...
+                .expect("Serving latest data task should always succeed.");
```
ogenev, line 90: I'm not sure if this is a valid assumption. I guess we still want to request the data for the next slot even if the request for the current slot fails for some reason.

ogenev (commented):
My concern is that we are blocking the main loop and waiting for the `serve_latest` task to finish before estimating the time and serving the data for the next slot.

 I think a more structured approach would be to create a service that will manage all shared variables and modify them via commands from the tokio threads.

portal-bridge/src/beacon_bridge.rs
carver: Your concern is that serving this data will last longer than the slot time of 12s? If that's true, I'd argue the bridge is just not working anyway, since it's not keeping up with the chain.

portal-bridge/src/beacon_bridge.rs
carver: `serve_latest()` is infallible. This await will only fail if the task gets cancelled by tokio, which we're not doing explicitly anywhere. Since cancelling this task should only happen on shutdown, we shouldn't ever trigger this panic.

ogenev: You are right, this should not panic :yum:.

carver:
> My concern is that we are blocking the main loop and waiting for the `serve_latest` task to finish before estimating the time and serving the data for the next slot.

Yeah, totally valid that we need to carefully consider this case. Here's what I think happens if one data publish takes longer than a slot, in three cases: in the current master, the PR as you reviewed it, and the current PR.
- master: two publish tasks would be running at the same time. The retrieved values from the later task could be written first, then the earlier task could write, clobbering the value from the later one.
- earlier PR: it would skip attempting to publish during the "slots + 4" moment that it skipped.
- current PR: it will not attempt the next publish until the previous one finishes, but then it will attempt the next publish immediately, and resume publishing on the "slots + 4" schedule, assuming it can catch up.

If it broadly can't keep up with the chain, we have much bigger problems that need to be solved some other way.

> I think a more structured approach would be to create a service that will manage all shared variables and modify them from the tokio threads via commands.

I think this approach still leaves us in the "master" scenario, where earlier publishes can overwrite more recent ones. Between that, and wanting to avoid the extra lines of code, I think this PR is in merge-able shape.

---

Bonus fix: I noticed that there was a thread sleep in the async function! I assume it was a typo, but just in case here's [why not to thread::sleep in async](https://blog.hwc.io/posts/rust-futures-threadsleep-and-blocking-calls-inside-async-fn/#don-t-sleep-on-std-thread-sleep).

carver:
This might be easier to review as separate commits.

portal-bridge/src/beacon_bridge.rs
```diff
@@ -200,34 +215,26 @@ impl BeaconBridge {
...
-                    *finalized_block_root = latest_finalized_block_root;
-                }
-                Err(err) => return Err(err),
-            }
+            // Return the latest finalized block root if we successfully gossiped the latest bootstrap.
+            Self::gossip_beacon_content(portal_clients, content_key, content_value)
+                .await
+                .map(|_| latest_finalized_block_root)
+        } else {
+            Err(anyhow::anyhow!("Finalized block root has not changed"))
```
ogenev, line 223: We have to return Ok() here because it is expected behavior for the latest finalized block root to not change every slot:

```suggestion
            Ok(latest_finalized_block_root)
```

portal-bridge/src/beacon_bridge.rs
```diff
@@ -80,69 +79,83 @@ impl BeaconBridge {
...
     /// Serve latest beacon network data to the network
+    ///
+    /// Returns the new current period and finalized block root
     async fn serve_latest(
         api: ConsensusApi,
         portal_clients: Arc<Vec<HttpClient>>,
-        current_period: Arc<AtomicUsize>,
-        finalized_block_root: Arc<Mutex<String>>,
-    ) {
+        current_period: usize,
```
ogenev, line 126: nit: No need anymore to use `usize` for the period value, we can change it to `u64`.

ogenev (approved):
Cool, looks like a nice improvement to me :100:

carver:
Weird, got a test failure, but can't reproduce it locally.

carver:
Ok I tried reverting to the commit that passed yesterday, and now that same commit is failing in CI today. So... I have no idea what's going on. Any guesses @ogenev ?

carver:
Yup, it was just slow today. It succeeded on retry-attempt # 10 (the test was previously failing after 6 attempts).

PR #875, opened 2023-08-29
feat: define concrete type for RoutingTableInfo

### LINKED ISSUE #862: Create RoutingTableInfo type
Not sure why, but we [use a simple `Value` as the data type for the type alias `RoutingTableInfo`](https://github.com/ethereum/trin/blob/01b7d4ef03ed5297d0c72f42fd403d7aa6d7671f/ethportal-api/src/types/discv5.rs#L30). We should implement an actual `RoutingTableInfo` struct, that serializes & deserializes according to the [specified return type](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false).

0xcrust:
### What was wrong?
Loosely defined RoutingTableInfo type.

Closes #862

### How was it fixed?
Changed RoutingTableInfo to a struct according to the specification. The Routing table consists of a `localNodeId` field and a `buckets` field which is represented as a nested array of node-ids, with each inner array representing a single bucket in the k-buckets table. 

The serializable types represent node_id as a `String`(hex-encoded) so as to conform to the serialization and deserialization scheme proposed by the specs. This is a workaround for the fact that the `NodeId` type exported by the enr crate only supports default serialization and deserialization to and from a `bytes32`, which is possibly a reason why `RoutingTableInfo` was defined as a `Value` in the first place(#862).

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-api/src/types/discv5.rs
+++ b/ethportal-api/src/types/discv5.rs
@@ -28,6 +28,7 @@ pub struct NodeInfo {
     pub ip: Option<String>,
 }
 
+/// Information about a discv5/overlay network's routing table.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct RoutingTableInfo {
--- a/ethportal-api/src/types/discv5.rs
+++ b/ethportal-api/src/types/discv5.rs
@@ -1,20 +1,21 @@
 use super::enr::Enr;
+use crate::utils::bytes::hex_encode;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
 
 use discv5::enr::NodeId;
 
 /// Discv5 bucket
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct Bucket {
-    #[serde(flatten)]
-    pub node_ids: Vec<NodeId>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub node_ids: Vec<String>,
 }
 
 /// Represents a discv5 kbuckets table
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(transparent)]
 pub struct KBucketsTable {
-    #[serde(flatten)]
     pub buckets: Vec<Bucket>,
 }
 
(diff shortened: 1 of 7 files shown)


ethportal-peertest/src/scenarios/basic.rs
```diff
@@ -114,13 +111,8 @@ pub async fn test_history_store(target: &Client) {
...
     info!("Testing portal_historyRoutingTableInfo");
-    let result = HistoryNetworkApiClient::routing_table_info(target)
-        .await
-        .unwrap();
-    assert!(result.get("buckets").unwrap().is_object());
-    assert!(result.get("numBuckets").unwrap().is_u64());
-    assert!(result.get("numNodes").unwrap().is_u64());
-    assert!(result.get("numConnected").unwrap().is_u64());
+    let result = HistoryNetworkApiClient::routing_table_info(target).await;
+    assert!(result.is_ok());
```
ogenev, line 115: Here maybe we also want to check the `local_node_id` and `buckets` fields,

ogenev (approved):
Thanks, looks good to me! Makes Trin more rpc specs compliant :)

ethportal-api/src/types/discv5.rs
```diff
@@ -27,4 +28,25 @@ pub struct NodeInfo {
     pub ip: Option<String>,
 }
 
-pub type RoutingTableInfo = Value;
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct RoutingTableInfo {
```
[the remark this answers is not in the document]
0xcrust: Sure!

ethportal-peertest/src/scenarios/basic.rs
```diff
@@ -27,10 +27,7 @@ pub async fn test_discv5_node_info(peertest: &Peertest) {
 pub async fn test_discv5_routing_table_info(target: &Client) {
     info!("Testing discv5_routingTableInfo");
     let result = Discv5ApiClient::routing_table_info(target).await.unwrap();
-    let local_key = result.get("localNodeId").unwrap();
-    assert!(local_key.is_string());
-    assert!(local_key.as_str().unwrap().starts_with("0x"));
-    assert!(result.get("buckets").unwrap().is_array());
```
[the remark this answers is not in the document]
0xcrust: The assertion previously checked that the `buckets` field in the returned `Value` was an array. The rationale for its removal is that it's no longer needed now that `RoutingTableInfo` is a concrete type since successful deserialization proves that it fits the defined structure.

ethportal-peertest/src/scenarios/basic.rs
ogenev: I don't think an empty table is illegal if a node doesn't connect to a bootnode, the table should be empty.

0xcrust: Alright. I'll go ahead and resolve this then

ogenev:
@0xcrust could we rebase this and merge?

0xcrust:
Should be good to go now @ogenev

PR #879, opened 2023-08-31
Impose limit on batch request size

### LINKED ISSUE #830: Inspect bridge error when batching more than 100 requests.
It seems like (some or all of) the pandaops nodes are set to a limit of 100 for a batched request. This should be validated & if confirmed, we'll need some logic inside of the bridge to batch the batched requests.

njgheorghita:
### What was wrong?
Fixes #830 

Batch requests to a panda-ops provider of > 100 were failing as those clients are configured to handle a max of 100 batch requests at a time.

This bug was the reason why our bridge has not been pumping out receipts. Can confirm that bridge now gossips receipts as expected.

### How was it fixed?
Add a limit to haw many pandaops requests can be batched into a single request.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portal-bridge/src/pandaops.rs
+++ b/portal-bridge/src/pandaops.rs
@@ -3,7 +3,13 @@ use crate::constants::{BASE_CL_ENDPOINT, BASE_EL_ENDPOINT};
 use crate::{PANDAOPS_CLIENT_ID, PANDAOPS_CLIENT_SECRET};
 use anyhow::anyhow;
 use ethportal_api::types::jsonrpc::request::JsonRequest;
-use serde_json::json;
+use futures::future::join_all;
+use serde_json::{json, Value};
+use tracing::warn;
+
+/// Limit the number of requests in a single batch to avoid exceeding the
+/// provider's batch size limit configuration of 100.
+const BATCH_LIMIT: usize = 100;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct PandaOpsMiddleware {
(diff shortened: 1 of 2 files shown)


mrferris (approved):

portal-bridge/src/pandaops.rs
```diff
@@ -3,7 +3,9 @@ use crate::constants::{BASE_CL_ENDPOINT, BASE_EL_ENDPOINT};
 use crate::{PANDAOPS_CLIENT_ID, PANDAOPS_CLIENT_SECRET};
 use anyhow::anyhow;
 use ethportal_api::types::jsonrpc::request::JsonRequest;
-use serde_json::json;
+use serde_json::{json, Value};
+
+const BATCH_LIMIT: usize = 100;
```
ogenev, line 8: A short docstring will be useful here to describe this const.

portal-bridge/src/pandaops.rs
```diff
@@ -45,6 +47,23 @@ impl PandaOpsMiddleware {
...
+                acc.extend_from_slice(&next?);
+                Ok::<Vec<Value>, Box<dyn std::error::Error>>(acc)
+            },
+        ) {
+            Ok(val) => Ok(serde_json::to_string(&val)?),
+            Err(err) => Err(anyhow!("Unable to flatten batch request: {err:?}")),
+        }
+    }
+
+    async fn batch_request_100(&self, obj: Vec<JsonRequest>) -> anyhow::Result<Vec<Value>> {
```
ogenev, line 66: if we change the `BATCH_LIMIT` const this function name will not be relevant anymore.

ogenev (approved):

PR #890, opened 2023-09-05
Filter out source node from Nodes response

njgheorghita:
### What was wrong?
While I was debugging the flakiness in #851, I noticed that quite often we see the error ~ `Unable to insert node into routing table, invalid self update`. Indicating that a node is trying to add itself to its own routing table, since it is present in the `Nodes` response from a peer. 

This pr follows the spec update in https://github.com/ethereum/portal-network-specs/pull/230. It filters the "requesting" node out from a `FindNodes` response. We already filter out the "requesting" node from a `FindContent::Enrs` response in https://github.com/ethereum/trin/blob/master/portalnet/src/overlay_service.rs#L1148 & "responding" nodes are automatically omitted from that list

### How was it fixed?
Filter out source node from a `Nodes` response.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/ethportal-peertest/src/scenarios/find.rs
+++ b/ethportal-peertest/src/scenarios/find.rs
@@ -82,7 +82,7 @@ pub async fn test_recursive_find_nodes_random(peertest: &Peertest) {
         .recursive_find_nodes(target_node_id)
         .await
         .unwrap();
-    assert_eq!(result.len(), 3);
+    assert_eq!(result.len(), 2);
 }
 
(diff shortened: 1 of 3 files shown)


ogenev:
I think if this is not specified in the specs, we need to first bring the issue to https://github.com/ethereum/portal-network-specs for discussion and then implement it.

njgheorghita:
This pr is now ready for review after https://github.com/ethereum/portal-network-specs/pull/230 was merged into the specs.

carver (approved):
LGTM

PR #894, opened 2023-09-06
Not for Merging: testing CI with a very long beacon content wait time

carver:
### To-Do

- [ ] Clean up commit history


carver:
The `peertest_beacon_bridge` test is failing over on #874 for no apparent reason. Can we reproduce it here, and/or find out how long the content needs to propagate?

carver:
Hmmm, now it is passing... So maybe it is actually related to the PR somehow

PR #895, opened 2023-09-07
Adds preliminary Bridge-Infura integration

mrferris:
### What was wrong?

PandaOps down meant `portal-bridge` couldn't work.

### How was it fixed?

`--infura` option added to `portal-bridge`

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portal-bridge/src/cli.rs
+++ b/portal-bridge/src/cli.rs
@@ -50,6 +50,13 @@ pub struct BridgeConfig {
     )]
     pub network: Vec<NetworkKind>,
 
+    #[arg(
+        long = "infura",
+        help = "Use Infura as the execution layer provider",
+        default_value = "false"
+    )]
+    pub infura: bool,
+
     #[arg(long, help = "Url for metrics reporting")]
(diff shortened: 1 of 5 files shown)


carver:
Having a global flag for `--infura` seems okay. I think it's maybe even better to have each request hit pandaops once and then use Infura if it's down (or the pandaops key is missing). That lets us skip adding the CLI flag, and we won't even need a deploy to switch between them.

njgheorghita (commented):
I understand the motivation for this pr, but, I think there's some more logic that's needed here. Basically, (iirc) infura runs geth & pandaops does not. So it's not as simple just replacing the http url for requests. If we are going to introduce infura support, we need to make sure that the json responses are compatible & successfully decoded into our content types (aka tests against infura responses). We used to support infura, but removed "infura" support because of the complexities involved, which we didn't consider worthwhile to sort.

KolbyML:
Closing as this PR was superceded by https://github.com/ethereum/trin/pull/1089

PR #896, opened 2023-09-07
feat: Add slot number in light client finality and optimistic content keys

ogenev:
### What was wrong?
There is a portal specs [change](https://github.com/ethereum/portal-network-specs/pull/217) that adds slot number to the `LightClientOptimisticUpdateKey` and `LightClientFinalityUpdateKey`.

### How was it fixed?

- Remove the `ZeroKey` struct
- Add the slot number to the content keys
- Update tests

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/ethportal-api/src/types/content_key/beacon.rs
+++ b/ethportal-api/src/types/content_key/beacon.rs
@@ -13,48 +13,8 @@ use std::fmt;
 pub enum BeaconContentKey {
     LightClientBootstrap(LightClientBootstrapKey),
     LightClientUpdatesByRange(LightClientUpdatesByRangeKey),
-    LightClientFinalityUpdate(ZeroKey),
-    LightClientOptimisticUpdate(ZeroKey),
-}
-
-/// Represents a zero ssz bytes for a portal content key.
-#[derive(Clone, Debug, PartialEq, Eq)]
-pub struct ZeroKey;
-
-impl Encode for ZeroKey {
-    fn is_ssz_fixed_len() -> bool {
-        true
-    }
-
-    fn ssz_bytes_len(&self) -> usize {
-        8
-    }
-
-    fn ssz_append(&self, buf: &mut Vec<u8>) {
-        buf.extend_from_slice(&[0u8; 8]);
-    }
-}
-
-impl Decode for ZeroKey {
-    fn is_ssz_fixed_len() -> bool {
-        true
-    }
-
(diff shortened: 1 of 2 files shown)


carver:
> [11:59 AM]acolytec3: One thing that occurs to me is that with PR 217 to require the slot in the content key of both optimistic and finality updates, we should allow for that value to be zero in the finalityUpdate key and that imply "just send me your latest finality update" 
> [11:59 AM]acolytec3: since ideally we're only storing one at a time, you can just send the one you have in that case
> [11:59 AM]acolytec3: otherwise, we're left with clients having to spam the network guessing at the slot of the last finality update
> [12:00 PM]acolytec3: there isn't a good way to "guess" the last finality update other than to request the update from the last epoch boundary or the one before that and so on going backward
> [12:56 PM]Jason Carver: Potentially dumb idea: rebroadcast the finality update every slot at the new key, even if the value hasn't changed

Thoughts?

ethportal-api/src/types/content_key/beacon.rs
```diff
@@ -73,6 +33,32 @@ pub struct LightClientUpdatesByRangeKey {
     pub count: u64,
 }
 
+/// Key used to identify a light client finality update.
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct LightClientFInalityUpdateKey {
```
carver, line 38:
```suggestion
pub struct LightClientFinalityUpdateKey {
```

carver (approved):
Yup, seems good! Curious what you think about the issue of finding/requesting the finality update.

ogenev:
> Thoughts?

Copying my comment from discord:


_I don't see value in having zero content key for the finality update, it does not give us any useful information imo and will cause issues when gossiping. I think including the  optimistic_slot instead of the finality_slot in the key makes sense. This way, the node can request LightCLientFinalityUpdate by following its own internal clock as a supporting mechanism to the gossip. This was proposed by @deme  in https://github.com/ethereum/portal-network-specs/pull/217#discussion_r1259568544 and now seems reasonable to me._

We are waiting for an agreement to change the finality update slot to an optimistic slot and will update this PR before merging.

carver:
> We are waiting for an agreement to change the finality update slot to an optimistic slot and will update this PR before merging.

Yup, chat seems to have moved more clearly in that direction, since your comment. The relevant spec PR https://github.com/ethereum/portal-network-specs/pull/228

ogenev:
Updated the content keys to include the `signature_slot` as discussed in discord.

Issue #885, opened 2023-09-02
Return the basic header fields for eth_getBlockByHash request


carver:

carver:
One strategy would be to resurrect the code that was deleted (this was previously supported, if my skim of [old RPC PRs](https://github.com/ethereum/trin/pulls?q=is%3Apr+eth_getBlockByHash+is%3Aclosed) is correct).

Or, maybe it makes more sense to start pulling more code from reth here. Maybe even extracting more of our RPC-handling code to use the reth base code, and customize it to our needs.

carver:
Support was first added in #261, then a bugfix in #293, then removed (perhaps by accident?) in #508.

Be sure to start with an ethportal-peertest scenario (or perhaps portal-hive? ... maybe not until it's a little more stable locally, to enable us to quickly iterate on the tests)

PR #293, opened 2022-04-04
Properly serialize extra data bytes in header

njgheorghita:
### What was wrong?
the old web3's decoded response to `eth_getBlockByHash`...
```
AttributeDict({'author': '0x2a65aca4d5fc5b5c859090a6c34d164135398226',
 'base_fee_per_gas': None,
 'difficulty': 7767604946080,
 'extra_data': [215,
  131,
  1,
  3,
  2,
  132,
  71,
  101,
  116,
  104,
  135,
  103,
  111,
  49,
  46,
  53,
  46,
  49,
  133,
  108,
  105,
  110,
  117,
  120],
 'gas_limit': '0x2fefd8',
 'gas_used': '0x5208',
 'log_bloom': '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
 'mix_hash': '0xc5e99c6e90fbdee5650ff9b6dd41198655872ba32f810de58acb193a954e1589',
 'nonce': HexBytes('0x40f1ce50d18d7fdc'),
 'number': 633130,
 'parent_hash': '0x6add1c183f1194eb132ca8079197c7f2bc43f644f96bf5ab00a93aa4be499360',
 'receipts_root': '0x59cf53b2f956a914b8360ea6fe271ebe7b10461c736eb16eb1a4121ba3abbb85',
 'state_root': '0x5ae233f6377f0671c612ec2a8bd15c20e428094f2fafc79bead9c55a989294dd',
 'timestamp': 1449116606,
 'transactions_root': '0x64183d9f805f4aecbf532de75e6ad276dc281ba90947ff706beeaecc14eec6f5',
 'uncles_hash': '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347'})
```
### How was it fixed?
the new response ....
```
AttributeDict({'author': '0x2a65aca4d5fc5b5c859090a6c34d164135398226',
 'base_fee_per_gas': None,
 'difficulty': 7767604946080,
 'extra_data': '0xd783010302844765746887676f312e352e31856c696e7578',
 'gas_limit': '0x2fefd8',
[...]

PR #897, opened 2023-09-07
feat: add base eth_* rpc & eth_chainID is always 1

carver:
### What was wrong?

Need the `eth_` namespace to accomplish #885 

### How was it fixed?

I think this is the minimal set of changes to implement a new RPC namespace. I'm not totally impressed, because it seems to sprawl a bit. I don't have any ideas for how to improve it yet, besides leaving this PR as a guide.

### To-Do

- [x] Clean up commit history
- [x] extract chain id constant
- [x] move duplicate tracing logic to a shared utility module

### DIFF
--- a/rpc/src/eth_rpc.rs
+++ b/rpc/src/eth_rpc.rs
@@ -1,0 +1,31 @@
+use crate::jsonrpsee::core::{async_trait, RpcResult};
+use ethereum_types::U256;
+use ethportal_api::EthApiServer;
+use trin_validation::constants::CHAIN_ID;
+
+pub struct EthApi;
+
+impl EthApi {
+    pub fn new() -> Self {
+        Self
+    }
+}
+
+impl Default for EthApi {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+#[async_trait]
+impl EthApiServer for EthApi {
+    async fn chain_id(&self) -> RpcResult<U256> {
+        Ok(U256::from(CHAIN_ID))
+    }
+}
+
+impl std::fmt::Debug for EthApi {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EthApi").finish_non_exhaustive()
+    }
+}
--- a/tests/rpc_server.rs
+++ b/tests/rpc_server.rs
@@ -1,0 +1,53 @@
+/// Test that a 3rd-party web3 client can understand our JSON-RPC API
+use std::net::{IpAddr, Ipv4Addr};
+
+use ethers_core::types::U256;
+use ethers_providers::*;
+use serial_test::serial;
+
+use ethportal_api::types::cli::{TrinConfig, DEFAULT_WEB3_IPC_PATH};
+use rpc::RpcServerHandle;
+
+mod utils;
+
+async fn setup_web3_server() -> (RpcServerHandle, Provider<Ipc>) {
+    utils::init_tracing();
+
+    let test_ip_addr = IpAddr::V4(Ipv4Addr::new(127, 0, 0, 1));
+    // Use an uncommon port for the peertest to avoid clashes.
+    let test_discovery_port = 8999;
+    let external_addr = format!("{test_ip_addr}:{test_discovery_port}");
+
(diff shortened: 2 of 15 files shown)


carver:
Hm, so I don't like that this test missed a bug: the chain ID was not being hex-encoded. I'm looking at using a 3rd-party web3 library so that we don't have a "two wrongs make a right" situation of our server matching our client implementation, and failing to meet the actual RPC spec.

web3.py is usually strict (or was?) but interestingly, it accepts a chain ID that is not hex-encoded, without complaint:
```py
>>> from web3 import Web3
>>> w3 = Web3(Web3.IPCProvider('/tmp/trin-jsonrpc.ipc'))
>>> w3.client_version
'trin v0.1.1-alpha.1-da8572'
>>> w3.eth.chain_id
1
>>> w3.provider.make_request('eth_chainId', [])
{'jsonrpc': '2.0', 'result': 1, 'id': 2}
```

Note that `result` should have been `0x1` according to the spec.

I'm exploring using ethers-rs. Using a 3rd-party client to test our server will be more important as we start to return more complex types, and want to prevent regressions like: #293 (which we could easily miss by just running our own client against our own server).

tests/web3_server.rs
```diff
@@ -0,0 +1,64 @@
...
+mod test {
+    use std::net::{IpAddr, Ipv4Addr};
+
+    use ethers::prelude::*;
+    use serial_test::serial;
+
+    use ethportal_api::types::cli::{TrinConfig, DEFAULT_WEB3_IPC_PATH};
+    use rpc::RpcServerHandle;
+
+    // TODO move to shared test utils, this is duplicated in self_peertest
```
carver, line 13: Move this before merge

carver:
There we go, ethers-rs caught the formatting issue! I'm happy with that.

carver:
Both of those changes are pretty straightforward. I don't think they block getting a review.

I suspect #902 will be ready to merge before this, so I'll rebase on that, because I'm hoping it will reduce the bit of CircleCI flakiness that I've been seeing. (I guess we're starting to hit memory limits on the node. I opened a ticket to CircleCI to bump up to the next higher node, but there's also surely some low-hanging fruit to shrink the build a bit.)

tests/rpc_server.rs
```diff
@@ -0,0 +1,65 @@
...
+        let trin_config = TrinConfig::new_from(
+            [
+                "trin",
+                "--external-address",
+                external_addr.as_str(),
+                "--web3-ipc-path",
+                DEFAULT_WEB3_IPC_PATH,
+                "--ephemeral",
+                "--discovery-port",
+                "8999",
```
ogenev, line 43: nit:

```suggestion
                &test_discovery_port.to_sring(),
```

rpc/src/eth_rpc.rs
```diff
@@ -0,0 +1,30 @@
+use crate::jsonrpsee::core::{async_trait, RpcResult};
+use ethereum_types::U256;
+use ethportal_api::EthApiServer;
+
+pub struct EthApi;
```
ogenev, line 5: I think we will need to wrap the history send channel to have access to the history network for future `eth_*` endpoints, same as https://github.com/ethereum/trin/blob/8e97046679b32f419d398d58ac4dd8d60804ceaf/rpc/src/history_rpc.rs#L23

But I understand that this is not necessary the scope of this PR.

tests/rpc_server.rs
ogenev, line 1: nice :partying_face:

ogenev (approved):
Looks good :100:

rpc/src/eth_rpc.rs
njgheorghita: iiuc, whatever "constructs" `eth_*` responses will need access to every subnetwork. It's out of scope for this pr, but imo should be taken in consideration as we build out this functionality.

carver: 👍🏻  yeah I'll add access as we go.

I think it could be a good idea to have Options for the networks, so that if someone starts with just history, then we can still serve getBlock calls. Then if they try an `eth_call`, the rpc server will respond with an error saying that you need to re-launch `trin` with the state network enabled.

PR #899, opened 2023-09-08
light-client: Strip unnecessary endpoints from the light client node

ogenev:
### What was wrong?

We have a light client node which is a modified version of Helios. Integrating this node into the portal beacon will be easier if we strip all unnecessary consensus endpoints from the node and leave only the light client syncing mechanics and types.

Having a minimal version of the light client, without APIs to access some of the consensus data at this stage, will ease switching the crate types to `ethportal-api` types, migrate the ssz library to what we are using right now in Trin, and importing the sync mechanics to the beacon portal node.

### How was it fixed?

Remove the following consensus RPC from the light client: 

- get_block_transaction_count_by_hash
- get_block_transaction_count_by_number
- get_gas_price
- get_priority_fee
- get_block_number
- get_coinbase

If we want, It will be easier to add those endpoints back later when we switch all libraries and types to what Trin is using right now.

_“Sometimes, you have to destroy something in order to build something better.”_

### To-Do

- [ ] Clean up commit history


light-client/src/main.rs
```diff
@@ -20,7 +20,7 @@ async fn main() -> Result<()> {
 
     // Set the checkpoint to the last known checkpoint
     builder =
-        builder.checkpoint("0x760387ebbc6d7c48e421424def00cfd85d71dd8f4ea3a9c0bb2f01b684553a6d");
+        builder.checkpoint("0x31c982c923f839527f4be15308d5bcc41930dd4b2db514b8fbc52929509d126f");
```
ogenev, line 23: Updated to a recent checkpoint for faster sync.

carver: Are you thinking this is something we'll eventually want to pass in as a CLI argument? (not for this PR)

carver (approved):
Just took a quick glance. Definitely understand the desire to reduce scope. I've been looking at `eth_getBlockByHash` and it's kind of a pain to specify that single type in the response.

👍🏻 and agree we can add it back in later when it makes sense.

light-client/src/main.rs
ogenev: Definitely, this is the trusted block root mandatory for syncing.

PR #900, opened 2023-09-08
Add header saturation delay for history bridge test mode

KolbyML:
### What was wrong?
When using history bridge in test mode headers didn't have enough time to be propagated so block bodies and receipts would fail randomly.
### How was it fixed?
By adding the same delay used in the serve block code which was used for the same reason

### DIFF
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -86,9 +86,16 @@ impl Bridge {
             .expect("Error parsing history test assets.");
 
         for asset in assets.0.into_iter() {
-            Bridge::gossip_content(&self.portal_clients, asset.content_key, asset.content_value)
-                .await
-                .expect("Error serving block range in test mode.");
+            Bridge::gossip_content(
+                &self.portal_clients,
+                asset.content_key.clone(),
+                asset.content_value,
+            )
+            .await
+            .expect("Error serving block range in test mode.");
+            if let HistoryContentKey::BlockHeaderWithProof(_) = asset.content_key {
+                sleep(Duration::from_secs(HEADER_SATURATION_DELAY)).await;
+            }
         }
     }
 
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -94,7 +94,7 @@ impl Bridge {
             .await
             .expect("Error serving block range in test mode.");
             if let HistoryContentKey::BlockHeaderWithProof(_) = asset.content_key {
-                sleep(Duration::from_secs(HEADER_SATURATION_DELAY)).await;
+                sleep(Duration::from_secs(1)).await;
             }
         }
     }
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -94,7 +94,7 @@ impl Bridge {
             .await
             .expect("Error serving block range in test mode.");
             if let HistoryContentKey::BlockHeaderWithProof(_) = asset.content_key {
-                sleep(Duration::from_millis(100)).await;
+                sleep(Duration::from_millis(50)).await;
             }
         }
     }
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -94,7 +94,7 @@ impl Bridge {
             .await
             .expect("Error serving block range in test mode.");
             if let HistoryContentKey::BlockHeaderWithProof(_) = asset.content_key {
-                sleep(Duration::from_secs(1)).await;
+                sleep(Duration::from_millis(100)).await;
             }
         }
     }


carver:
> block bodies and receipts would fail randomly.

Can you link a failed test?

Obviously this is a problem that needs to be fixed, but I think we might be able to do it without adding so much wait time to every bridge test with a header. If I understand the failure correctly, maybe it would help to copy this change to the history version: https://github.com/ethereum/trin/pull/874/commits/520ec4db7597ebf5e5892ed70925dec6c0973b25

KolbyML:
> > block bodies and receipts would fail randomly.
> 
> Can you link a failed test?

https://github.com/ethereum/portal-hive/pull/87

> 
> Obviously this is a problem that needs to be fixed, but I think we might be able to do it without adding so much wait time to every bridge test with a header. If I understand the failure correctly, maybe it would help to copy this change to the history version: [520ec4d](https://github.com/ethereum/trin/commit/520ec4db7597ebf5e5892ed70925dec6c0973b25)

520ec4d isn't the same issue. This issue is block bodies/headers/receipts are gossiped at the same time. If the header isn't seeded by the time a block body is received validation will fail. Another solution could be packaging the header with block body's and receipts which is the only solution I can think of to not require a timeout. 

I am open to both. I think packaging the header with the bb/receipt would be better since then it prevents nodes from calling RFC every time they are gossiped a bb/receipt.

Hence adding a delay in the example above wouldn't help since it doesn't delay the validation attempt from occuring.

KolbyML:
https://github.com/ethereum/trin/blob/2a32224e3c2b0b80bc37c1b692c33016371f197a/portal-bridge/src/bridge.rs#L227

This function has the same sleep as the reason I am adding it here for context. So packaging the headers would prevent this sleep as well.

KolbyML:
We can shorten the delay if you want too up to you

carver:
Cool, yay for considering a couple options :)

> * package the headers with block bodies/ receipts

Yeah, this is interesting to me. I think there are going to be a lot of data offers in Portal that follow this model eventually. So exploring it seems worthwhile.

Can you flesh out the idea a little more? If code is easier than prose, that's fine by me. Whatever is easiest to communicate.

njgheorghita:
>  I think there are going to be a lot of data offers in Portal that follow this model eventually. 

Agreed. In fact, we already do (eg. the headers / bodies / receipts in (backfill & latest) modes). Since test mode & (backfill & latest) mode are handled differently, I would love to see a solution that all modes can share. 

A big unanswered question, is how long does it actually take for this content to propagate around the network? It will obviously be different in the testnet & in the mainnet.

- Does it make sense to hold off on propagating dependent content (eg bodies / receipts) until the bridge is able to perform a "RFC" lookup for the primary content (eg header)?
- Does it make sense to introduce a protocol-level `FindContent` message / variant where a node simply returns a bool indicating that it stores the target content? Probably not, but I could see this being useful for testing/network exploring rather than performing utp txs. 

> Two we package the headers with block bodies/ receipts.

This is an interesting idea, but imo it will require the quite a bit of heavy lifting / retooling to implement. Also it will mean that we're introducing a sizable amount of data-duplication across the network. Meaning the network will be using a lot of unnecessary storage capacity to store data multiple times over.

njgheorghita:
Something else to keep in mind is that soon, the primary data (eg beacon light client proofs) will be located on a separate subnetwork from the dependent data (history header). So it would be best if the solution we head towards is compatible for both...
- primary & dependent data that exist on the same network (eg history header & history bodies/receipts)
- primary & dependent data that exist on separate subnetworkrs (eg beacon proofs & history header)
Taking this into account, simply packaging up the history header with the receipts / bodies will only address one of these situations. A better solution would be one that can address both situations.

njgheorghita (approved):
I think the conversation we're having is important, and we should continue it (maybe in a tracking issue?), but is out of scope for this pr. Maybe, it turns out implementing a somewhat arbitrary sleep to the gossip mechanism is the best we can do to ensure that primary data is available before dependent data? But for the purposes of this pr, sleep is :+1:

carver:
> > I think there are going to be a lot of data offers in Portal that follow this model eventually.
> 
> Agreed. In fact, we already do (eg. the headers / bodies / receipts in (backfill & latest) modes)

We might be talking about different things. I was talking about having a single offered block body actually include the relevant receipt (or more likely, in the state network, having a single account data offer include it's merkle/verkle proof).

So to summarize the options again:
- Consider changing the spec to include the proof with any related data (obviously a longer conversation and project, and well outside scope)
- Doing a test for the presence of the required content to prove, before publishing (whether or not we do something smarter to avoid the actual content transfer, again outside scope. We could even do a relatively dumb thing and just never connect to the utp stream that the peer offers).
- Shrink the sleep

After looking at it a little longer, and the limited scope of where this comes into play for tests, I'm good just doing a shorter sleep. I wouldn't expect it to take more than a second or two on a tiny test run, which is where this is used, so maybe try it at 3 seconds to see if that's enough? Finishing tests faster by `(10 old seconds - 3 new seconds) * 2 affected tests = 14 seconds faster` in CI is a meaningful change, so that seems preferable to leaving it at 10 seconds each.

carver (approved):
👍🏻  with a new `HEADER_TESTING_SATURATION_DELAY` that's much shorter, and that's tested to make sure it works reliably in hive.

KolbyML:
I have done some testing and have decided to go with 50 milliseconds with is 200 times lower then the current timeout

PR #901, opened 2023-09-08
Better handle nodes in propagate gossip cross thread when none are connected

KolbyML:
### What was wrong?
If there was no connected nodes, we would go into the forloop then our code would fail since it had no connected nodes which was confusing when trying to debug.
### How was it fixed?
I gave this function the same treatment Mike gave for ``init_find_content_query``. We shouldn't go into the forloop if we don't have any nodes to gossip to in the first place.

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2513,7 +2513,6 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
         .collect();
 
     if all_nodes.is_empty() {
-        // If there are no connected nodes, use the nodes regardless of connectivity.
         warn!("No connected nodes, using disconnected nodes for gossip.");
         all_nodes = kbuckets
             .buckets_iter()
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2502,7 +2502,7 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
 ) -> usize {
     // Get all connected nodes from overlay routing table
     let kbuckets = kbuckets.read();
-    let all_nodes: Vec<&kbucket::Node<NodeId, Node>> = kbuckets
+    let mut all_nodes: Vec<&kbucket::Node<NodeId, Node>> = kbuckets
         .buckets_iter()
         .flat_map(|kbucket| {
             kbucket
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -2512,6 +2512,25 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
         })
         .collect();
 
+    if all_nodes.is_empty() {
```
njgheorghita, line 2515: I'm curious, can you justify why we want to retry all of the disconnected nodes here? I'm not sure it's worth it. We have other "bucket refresh" mechanisms, and imo at this layer in the protocol, if a node is disconnected I don't think it's worth reconnecting with every disconnected node in our routing table. That will likely result in many unproductive reqs. Instead, I'd suggest we just `return 0` if there's no connected nodes

KolbyML: https://github.com/ethereum/trin/pull/825

This PR is based off the same principle as this PR.

portalnet/src/overlay_service.rs
```diff
@@ -2512,6 +2512,25 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
...
+        // If there are no connected nodes, use the nodes regardless of connectivity.
+        warn!("No connected nodes, using disconnected nodes for gossip.");
```
carver, line 2517: I think it's okay to drop the comment, since the log explains the same situation adequately.

carver (approved):
🚢 LGTM

portalnet/src/overlay_service.rs
KolbyML: I think I will wait to merge this till I see what Nick says more. Cause I don't want to rush merge this.

KolbyML: If that is the case then all uses of add_enr should be considered "connected".

Cause if we don't have this and want to send gossip from A -> B

Wouldn't B have to send A a ping or another random request to become active.

So if we don't want to gossip to disconnected nodes. Then I would want add_enr to fully connected node B.

njgheorghita: > If that is the case then all uses of add_enr should be considered "connected".

I'm not sure what this means? (also, just a tip, but using \` to identify code mid-comment is very helpful when reading, eg. `add_enr`) 

> Since I don't want to wait for bucket refreshes.

Iiuc, this logic is mostly to help us in our test suites. It's typically not a good idea to introduce core logic to handle special edge cases in our test suites. 

> I think the routing table is handling this responsibility, and the overlay service is just checking the inputs received from the routing table and behaving accordingly.

If that was the case, then all the overlay service should do is get the "connected" nodes, and not even use any concept of "disconnected" nodes?

> I agree though that "No connected nodes" is a condition that should ultimately go away once portal interop gets closer to 100%.

Agreed, and iiuc this only justifies my side of the discussion?

Well, there are two approvals on this pr, I feel like I'm being unreasonably stubborn at this point, so let's go ahead with the merge.

njgheorghita (approved):

portalnet/src/overlay_service.rs
KolbyML: Oh I just mean I want to do standardized testing across clients so I don't want to do special steps for certain clients. Currently Fluffy only needs add_enr to use gossip, but we required before this sending a ping from B -> A if I recall correctly. So I just want to standardize "connecting" nodes.

Ok I will merge it, I really hope I am not being to quick.

PR #902, opened 2023-09-09
perf(light-client): faster & smaller builds by removing ethers-rs

carver:
After the latest slimming-down of light-client, we could remove ethers
completely. When we eventually add the functionality back in, we are
likely to use internal types instead of the ones from ethers anyway, so
this is probably a permanent-ish change.

I was noticing that a recent PR that added ethers 2.0 for testing the
RPC server was causing CI to crash while building. It seemed like it was
the Out of Memory Killer that sigkilled the build. Ethers was the only
big thing added, so I stripped it down, and that helped some, but while
I was getting familiar with ethers, I thought I'd also do this change to
make the build even a little bit smaller and faster.

The problem was multiplied because light-client was using a different
version of ethers, so two copies were built.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,20 +3,10 @@
 version = 3
 
 [[package]]
-name = "Inflector"
-version = "0.11.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
-dependencies = [
- "lazy_static",
- "regex",
-]
-
-[[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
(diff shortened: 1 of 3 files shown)


ogenev:
Meged https://github.com/ethereum/trin/pull/899 before seeing this, need to rebase.

carver:
Hah, procrastination would have helped me here, most of the work was not necessary after #899.

In fact, after that PR, it now seems to make the most sense to remove ethers-rs altogether.

carver:
Failure in `lint` looks like an unrelated infrastructure issue:
```
error: failed to get `syn` as a dependency of package `eth2_ssz_derive v0.3.0`
    ... which satisfies dependency `eth2_ssz_derive = "^0.3.0"` (locked to 0.3.0) of package `ethportal-api v0.2.2 (/home/circleci/project/ethportal-api)`

Caused by:
  failed to query replaced source registry `crates-io`

Caused by:
  download of 3/s/syn failed

Caused by:
  failed to download from `https://index.crates.io/3/s/syn`

Caused by:
  [28] Timeout was reached (Operation too slow. Less than 10 bytes/sec transferred the last 30 seconds)
```

carver:
I think this is pretty non-controversial, and I don't want to have to wait for another overnight review and rebase after #897, so I'll just pull the trigger now. If something seems objectionable, let's chat and I'll clean it up as soon as possible.

PR #904, opened 2023-09-10
feat: retries new blocks

### LINKED ISSUE #893: Portal history bridge should not panic on error
When we experience an error serving the latest history data, we immediately panic due to this line:

https://github.com/ethereum/trin/blob/413bfd47fb3edcfb01c99c785ac108976811ca8e/portal-bridge/src/bridge.rs#L120C53-L120C53

The history bridge should not panic on errors (including when the pandaops nodes are experiencing issues) and we should keep trying to request the new data every block.

samtvlabs:
### What was wrong?

This closes #893 

### How was it fixed?

Instead of panic, we loop over each new block , and emit a warning instead

                

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -115,9 +115,25 @@ impl Bridge {
                 info!("Discovered new blocks to gossip: {gossip_range:?}");
                 let gossip_stats = Arc::new(Mutex::new(GossipStats::new(gossip_range.clone())));
                 let epoch_acc = None;
-                self.serve(gossip_range.clone(), epoch_acc, gossip_stats.clone())
-                    .await
-                    .expect("Error serving block range in latest mode.");
+                loop {
+                    match self
+                        .serve(
+                            gossip_range.clone(),
+                            epoch_acc.clone(),
+                            gossip_stats.clone(),
+                        )
+                        .await
+                    {
+                        Ok(_) => break,
+                        Err(e) => {
+                            warn!(
+                                "Error serving block range in latest mode: {:?}. Retrying...",
+                                e
+                            );
+                            continue;
+                        }
+                    }
+                }
                 block_index = gossip_range.end;
             }
         }


njgheorghita (changes requested):
At quick glance, there's some things that need addressing here.
- I don't believe this is the correct location for the "retry" logic, since it will "retry" constructing & gossiping **all** of the header, body, & receipts, if any **one** of those steps fails. The step that we actually will want to "retry" would be constructing and gossiping a single content type (eg. header, block body, or receipts) if that individual step fails. 
- This only implements the "retry" logic for `launch_latest`, but we will want it in `launch_backfill` as well. 
- We will want some kind of limit on "retries", I'd start with `3`, so that we're not retrying to construct & gossip some content ad infinitum

samtvlabs:
Thanks for the review. I would like to clarify if it makes sense to move this logic to within the `serve`  function


i.e.
```rust
    async fn serve(
        &self,
        gossip_range: Range<u64>,
        epoch_acc: Option<Arc<EpochAccumulator>>,
        gossip_stats: Arc<Mutex<GossipStats>>,
    ) -> anyhow::Result<()> {
        let futures = futures::stream::iter(gossip_range.into_iter().map(|height| {
            let epoch_acc = epoch_acc.clone();
            let gossip_stats = gossip_stats.clone();
            async move {
                let mut retry_count = 0;
                while retry_count < 3 {
                    match self
                        .serve_full_block(height, epoch_acc, gossip_stats, self.portal_clients.clone())
                        .await
                    {
                        Ok(_) => break,
                        Err(e) => {
                            warn!(
                                "Error serving block in latest mode: {:?}. Retrying...",
                                e
                            );
                            retry_count += 1;
                            if retry_count >= 3 {
                                return Err(anyhow!("Failed to serve block after 3 retries")); // We dont panic here , just warn
                            }
                        }
                    }
                }
            }
        }))
        .buffer_unordered(FUTURES_BUFFER_SIZE)
        .collect::<Vec<()>>();
        futures.await;
        if let Ok(gossip_stats) = gossip_stats.lock() {
            gossip_stats.display_stats();
        } else {
            warn!("Error displaying gossip stats. Unable to acquire lock."); // We dont panic here , just warn
        }
        Ok(())
    }
```

that way , its passed on to both `launch_latest` and `launch_backfill`

njgheorghita:
Ahh, sorry, but I was a bit confused about the scope of this pr. It seems to me like I was over-complicating things. All the issue #893 is asking is to **not panic** in the case that we are unable to that we run into an error requesting/gossiping data. In which case, simply logging an warning in the error case is fine. We don't need to worry about retries. Something like...
```rust
if let Err(err) = self.serve(gossip_range.clone(), epoch_acc, gossip_stats.clone()).await {
    warn!("Error serving block range in latest mode: {err:?}");
}
```

KolbyML:
Closing as this PR is no longer applicable

PR #906, opened 2023-09-10
feat: pass fluffy bridge configs through bridge config

### LINKED ISSUE #880: portal-bridge: Pass bridge node arguments via `portal-bridge` CLI
When we run portal-bridge, we spin trin or fluffy bridge node alongside the service that queries the data from EL and CL nodes. Then we use this bridge node to gossip the data to the rest of the network.

Currently, this bridge node runs with a hard-coded arguments:
https://github.com/ethereum/trin/blob/633e53bd1464538ce984b14082b86f289ff9026e/portal-bridge/src/client_handles.rs#L18

What we would like is to be able to modify most of those arguments via the portal-bridge CLI here:
https://github.com/ethereum/trin/blob/633e53bd1464538ce984b14082b86f289ff9026e/portal-bridge/src/cli.rs#L20

samtvlabs:
### What was wrong?

This PR closes #880 

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portal-bridge/src/cli.rs
+++ b/portal-bridge/src/cli.rs
@@ -50,6 +50,14 @@ pub struct BridgeConfig {
     )]
     pub network: Vec<NetworkKind>,
 
+    #[arg(
+        long = "storage_size",
+        help = "Storage Size for Fluffy Bridge",
+        default_value = "0",
+        use_value_delimiter = true
+    )]
+    pub storage_size: u64,
+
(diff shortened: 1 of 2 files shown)


njgheorghita (commented):
@ogenev At this point it seems like we might want to include a fluffy executable & include a test that our bridge maintains compatibility?

KolbyML:
Closing this PR as it is stale and has been here for 7 months and it doesn't look like that will change

Issue #279, opened 2022-03-28
eth_getBlockByHash response should include uncles and txs


njgheorghita:
To be compliant with the eth json-rpc api spec, the [object returned from `eth_getBlockByHash` request](https://eth.wiki/json-rpc/API#eth_getblockbyhash) should include a list of uncle hashes and transaction hashes. Also, support for the `full_transactions` param would be nice to have.

njgheorghita:
closing since we no longer support this endpoint

carver:
> we no longer support this endpoint

Hm, but we should eventually. I would say support for `eth_getBlockByHash` is what it would mean to be ready for Devconnect.

carver:
Depends on #885

PR #908, opened 2023-09-12
feat: support eth_getBlockByHash over json-rpc (only with header fields, for now)

### LINKED ISSUE #885: Return the basic header fields for eth_getBlockByHash request

carver:
Will fix #885 

There are a few little tweaks I have left to do, but nothing standing in the way of a review, IMO. Everything planned is unchecked below. I may just skip the 944 rebase, since it is relatively small, and can probably be reviewed together in this PR. (and I don't want to add unnecessary delay, so I can move on to transactions & uncles)

- [x] Clean up commit history
- [x] What's up with? `2023-09-29T21:58:10.267211Z ERROR jsonrpsee_core::client::async_client: [backend]: Failed to read message: Networking or low-level protocol error: bytes remaining on stream` -- Wow, google is stumped. I guess I'll punt this for now...
- [x] uint conversion utilities, to test timestamp, gas_used, etc
- [x] Rebase on #942 
- [x] Rebase on #943
- [x] ~Rebase on #944~
- [x] Enumerate [full list of header fields](https://ethereum.org/en/developers/docs/apis/json-rpc/#eth_getblockbyhash) (any that don't require further requests, like transactions, uncles, or total difficulty), and test presence:
  - [x] bloom
  - [x] ~size~ - I couldn't find a reliable definition for size, so I am punting it for now
- [x] create issues for TODOs, and remove them
  - [x] https://github.com/ethereum/trin/issues/960
  - [x]  https://github.com/ethereum/trin/issues/279
  - [x]  #961

### DIFF
--- a/ethportal-api/src/eth.rs
+++ b/ethportal-api/src/eth.rs
@@ -1,9 +1,17 @@
-use ethereum_types::U256;
+use ethereum_types::{H256, U256};
 use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use reth_rpc_types::Block;
 
 /// Web3 JSON-RPC endpoints
 #[rpc(client, server, namespace = "eth")]
 pub trait EthApi {
     #[method(name = "chainId")]
     async fn chain_id(&self) -> RpcResult<U256>;
+
+    #[method(name = "getBlockByHash")]
+    async fn get_block_by_hash(
+        &self,
+        block_hash: H256,
+        hydrated_transactions: bool,
+    ) -> RpcResult<Block>;
 }
--- a/rpc/src/fetch.rs
+++ b/rpc/src/fetch.rs
@@ -1,9 +1,14 @@
 /// Fetch data from related Portal networks
+use ethereum_types::H256;
 use serde_json::Value;
 use tokio::sync::mpsc;
 
+use ethportal_api::types::constants::CONTENT_ABSENT;
+use ethportal_api::types::execution::header::Header;
 use ethportal_api::types::jsonrpc::endpoints::HistoryEndpoint;
 use ethportal_api::types::jsonrpc::request::HistoryJsonRpcRequest;
+use ethportal_api::utils::bytes::hex_decode;
+use ethportal_api::{ContentValue, HistoryContentKey, HistoryContentValue};
 
 use crate::errors::RpcServeError;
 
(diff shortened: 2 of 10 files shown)


carver:
I think it's ready for at least a first-pass review, although I obviously have a few cleanup tasks that are mentioned in the PR summary.

carver:
Ready for full review now.

ethportal-api/src/eth.rs
```diff
@@ -1,9 +1,16 @@
...
 /// Web3 JSON-RPC endpoints
 #[rpc(client, server, namespace = "eth")]
 pub trait EthApi {
     #[method(name = "chainId")]
     async fn chain_id(&self) -> RpcResult<U256>;
+    #[method(name = "getBlockByHash")]
+    async fn get_block_by_hash(
+        &self,
+        block_hash: H256,
+        full_transactions: bool,
```
njgheorghita, line 14: Personally, I'm fine with `full_transactions`, but the [spec](https://ethereum.github.io/execution-apis/api-documentation/) does say `Hydrated Transactions` - I guess it uses `"hydrated"` as the field title.

rpc/src/fetch.rs
```diff
@@ -28,3 +33,42 @@ pub async fn proxy_query_to_history_subnet(
...
+            let message =
+                format!("Invalid internal representation of block header; could not decode: {err}");
+            return Err(RpcServeError::Message(message));
+        }
+    };
+
+    match header {
+        HistoryContentValue::BlockHeaderWithProof(h) => Ok(h.header),
+        wrong_val => Err(RpcServeError::Message(format!(
+            "Found something other than block header at the key; got {:?}",
```
njgheorghita, line 70: I'm not sure I follow this error message. Looks like `at` is a typo for `as`. But also, we're dealing with a value here not the key.

tests/rpc_server.rs
```diff
@@ -39,16 +49,120 @@ async fn setup_web3_server() -> (RpcServerHandle, Provider<Ipc>) {
...
+
+    let block_id = ethers_core::types::H256::from(&block_hash.0);
+    let block = web3_client
+        .get_block(block_id)
+        .await
+        .expect("request to get block failed")
+        .expect("specified block not found");
+    web3_server.stop().unwrap();
+
+    tracing::debug!("retrieved block: {block:?}");
```
njgheorghita, line 138: I think we should remove this log from the test

njgheorghita (approved):
Nice!! Pumped to finally get a proper endpoint supported

rpc/src/fetch.rs
carver: I mean, it shouldn't ever happen but... what it means is that we asked HistoryEndpoint to give us the content value for a content key that specifies a header, but then the endpoint returned a value that is not a header.

So that's what the error message is trying to say. Given that the content key is definitely a header key, something went very wrong with the HistoryEndpoint, and it gave us some other type of content.

Does this error message convey that any more clearly?
"Internal trin error: got back a non-header from a key that must only point to headers; got ..."

ethportal-api/src/eth.rs
carver: Hm, yeah I haven't seen a lot of consensus on that field name, different copies of the spec call it different things. I can't remember know how I got to "full," because I also remember considering hydrated... I don't feel strongly about it. I will switch it.

PR #909, opened 2023-09-13
fix: improve the utp connection success rate, by using a default config

carver:
We were seeing failed connection attempts fairly regularly on the testnet. This is one step of resolving that issue.

utp-rs is mostly tested using the default configurations, which have changed a few times to improve results. As far as I can tell, the configuration choices here are just an old copy of the config.

So instead of copying in new values, I switched the config to use defaults whenever possible. (which requires using lazy_static) The only custom configuration that seems to matter is that discv5 enforces a custom maximum packet size.

Some reasons I think this will help:
- Number of connection attempts doubles from 3 -> 6
- Idle connection timeout ~doubles from 32->60
- Maximum packet timeout is cut in a quarter from 60->15, for more retries before triggering an idle timeout. (This will affect mid-stream timeouts, but won't affect the connection timeouts we're seeing)

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -20,6 +20,7 @@ use discv5::{
     rpc::RequestId,
 };
 use futures::{channel::oneshot, future::join_all, prelude::*};
+use lazy_static::lazy_static;
 use parking_lot::RwLock;
 use rand::seq::{IteratorRandom, SliceRandom};
 use smallvec::SmallVec;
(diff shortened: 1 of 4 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -82,16 +83,10 @@ const EXPECTED_NON_EMPTY_BUCKETS: usize = 17;
...
-    max_conn_attempts: 3,
-    max_idle_timeout: Duration::from_secs(32),
-    max_timeout: Duration::from_secs(60),
-    initial_timeout: Duration::from_millis(1500),
-    min_timeout: Duration::from_millis(500),
-    target_delay: Duration::from_millis(250),
-};
+lazy_static! {
+    /// The default configuration to use for uTP connections.
+    pub static ref UTP_CONN_CFG: ConnectionConfig = ConnectionConfig { max_packet_size: 1024, ..Default::default()};
```
njgheorghita, line 88: @carver You might want to read-through / follow-up on the conversation that happened [here](https://github.com/ethereum/trin/pull/767) about an accurate max packet size. It's been a while since I was thinking about this, but it seems like we may be able to fit a few more bytes in here. Don't consider this necessary for this pr, but tracking it in an issue/comment would be nice.

njgheorghita (approved):
🚢

portalnet/src/overlay_service.rs
carver: Tracked in #910

PR #920, opened 2023-09-15
Exempt gfi issues from stale issues bot

njgheorghita:
### What was wrong?
From discord: `I was just thinking that we should probably exempt GFI from the stale bot. It's normal for there to be no activity on those for long periods`

### To-Do

- [ ] Clean up commit history


njgheorghita:
@carver Seems like `flamingo` issues might also be worth exempting?

KolbyML (approved):

mrferris (approved):

PR #922, opened 2023-09-16
docs: add deep dive into kademlia

danielrachi:
### What was wrong?

The existing article about Kademlia in the book was a bit shallow.

### How was it fixed?

I wrote an article based on my understanding of the protocol. I believe it would complement the current information in the book.

(I wrote this several weeks ago as I was getting started with the portal network, but I realized that this might be a good thing to have in the docs.)

### To-Do

- [x] Clean up commit history
- [ ] Have someone check that the info in the document is accurate, and relevant to Trin.

### DIFF
--- a/book/src/developers/protocols/kademlia.md
+++ b/book/src/developers/protocols/kademlia.md
@@ -48,3 +48,80 @@ stored, and then other data is deleted until the storage is at the targed size.
 
 This involves repeatedly removing the content with the furthest ID from the node ID until
 the database is below the target.
+
+## Deep dive
+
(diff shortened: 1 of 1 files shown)


njgheorghita:
Thanks for the write-up! These kinds of explanations are very useful, especially to newcomers. However, this content feels more appropriately located inside of `www.ethportal.net`. It's general information that's relevant to all clients, and I'd like to start slimming down our docs to only containing trin-centric information. I'd recommend opening up a pr with this information over in https://github.com/ethereum/portal-website

danielrachi:
I agree, we should keep the book for trin-exclusive info and this is better placed at the general portal network docs.

Just opened ethereum/portal-website#7

PR #923, opened 2023-09-17
Make Trin compilable/(not crash) on Windows + Add check CI

KolbyML:
### What was wrong?
Certain parts of Trin either won't compile on Windows or crash because rust doesn't support windows Unix IPC
### How was it fixed?
By adding ``#![cfg(unix)]`` in files that should only be compile on windows and by adding an error if someone tries to use IPC on Windows.

Reth added ``#[cfg(unix)]`` a few months back so the reth-ipc library will compile it will just crash when we try to use ipc functions.

I also added a CI using ``check`` instead of ``build``
https://doc.rust-lang.org/cargo/commands/cargo-check.html

TL:DR it has all the CI benefits of building but without the codegen part. So it is a lot quicker for the result we want.

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -1140,6 +1140,8 @@ pub mod test {
         // The restarted store should have the same radius as the original
         assert_eq!(radius, new_storage.radius);
 
+        drop(storage);
+        drop(new_storage);
         temp_dir.close()?;
         Ok(())
     }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,7 @@ commands:
           command: sudo apt install clang
 orbs:
   rust: circleci/rust@1.6.0
+  win: circleci/windows@2.2.0
 executors:
   docker-publisher:
     environment:
@@ -159,6 +160,31 @@ jobs:
           name: Build Trin workspace
           command: cargo build --workspace
       - save-sccache-cache
+  check-windows:
+    description: |
+      Check the crate on Windows (Check will tell us if we can build on Windows without the need to codegen (which is the time consuming part)).
+    executor:
+      name: win/default
+      size: xlarge
+    environment:
+      RUSTFLAGS: '-D warnings'
+      RUST_LOG: 'debug'
+    steps:
+      - checkout
+      - run:
+          name: Install rustup and clang
+          # We are installing them at the same time because it is faster
+          # todo: Remove --ignore-checksums flag
+          command: choco install rustup.install llvm -y --ignore-checksums
+      - run:
+          name: Add target
+          command: rustup target add x86_64-pc-windows-msvc
+      - run:
+          name: Install target toolchain
+          command: rustup toolchain install stable-x86_64-pc-windows-msvc
+      - run:
+          name: Check Trin workspace
+          command: cargo check --workspace
   test:
     description: |
       Run tests.
@@ -259,4 +285,5 @@ workflows:
       - lint
       - build
       - test
+      - check-windows
       - utp-test
--- a/ethportal-peertest/src/lib.rs
+++ b/ethportal-peertest/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg(unix)]
 pub mod constants;
 pub mod scenarios;
 pub mod utils;
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -1520,7 +1520,7 @@ where
                 Request::PopulatedOffer(offer) => Ok(response_clone
                     .content_keys
                     .iter()
-                    .zip(offer.content_items.into_iter())
+                    .zip(offer.content_items)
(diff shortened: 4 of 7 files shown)


KolbyML:
I think in the future I will change the CI test from a pure windows one to one on linux doing a check for windows. I am not 100 sure if I can do it yet, but I think it would result in lower CI times so I am interested

portalnet/src/storage.rs
```diff
@@ -1140,6 +1140,8 @@ pub mod test {
         // The restarted store should have the same radius as the original
         assert_eq!(radius, new_storage.radius);
 
+        drop(storage);
+        drop(new_storage);
```
carver, line 1144: Why were the explicit drops necessary?

src/main.rs
```diff
@@ -10,6 +12,14 @@ use trin::run_trin;
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     init_tracing_logger();
     let trin_config = TrinConfig::from_cli();
+
+    // Placing this check here since TrinConfig has tests which this would fail.
+    // We also don't want to run unneeded code if we are going to panic for this anyways.
+    #[cfg(windows)]
```
carver, line 18: I had to read the comment a few times to understand it. Maybe something like?
```suggestion
    // Panic early on a windows build that is trying to use IPC, which is unsupported for now
    // Make sure not to panic on non-windows configurations.
    #[cfg(windows)]
```

carver (commented):
What do you think about moving the check to panic on Windows to the beginning of `run_trin`?

carver:
> I also added a CI using `check` instead of `build`
> https://doc.rust-lang.org/cargo/commands/cargo-check.html

Yeah, this makes a lot of sense. I'm even wondering if we should switch the `build` job to use check too. What do you think? 👍🏻 if you want to add it

portalnet/src/storage.rs
KolbyML: They are in the tests below and above. This change isn't related to the PR. Without these drops the test will panic on certain operating systems. If you read the doc for .close() on the line before there is a comment in the example about this.

KolbyML:
> What do you think about moving the check to panic on Windows to the beginning of run_trin?

resolved

> Yeah, this makes a lot of sense. I'm even wondering if we should switch the build job to use check too. What do you think? 👍🏻 if you want to add it

I am fine with that, would that belong in a different PR then?

src/main.rs
KolbyML: resolved

ogenev:
If I understand this correctly, the claim here is that windows does not support IPC? If this is the case, I'm curious what is this support for: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ ? Also saw this after googling: https://batsov.com/articles/2022/01/20/unix-sockets-are-now-supported-on-windows/

KolbyML:
> If I understand this correctly, the claim here is that windows does not support IPC? If this is the case, I'm curious what is this support for: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ ? Also saw this after googling: https://batsov.com/articles/2022/01/20/unix-sockets-are-now-supported-on-windows/

Rust doesn't support Windows UDS.

I am working on a PR to change this

KolbyML:
> If I understand this correctly, the claim here is that windows does not support IPC? If this is the case, I'm curious what is this support for: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/ ? Also saw this after googling: https://batsov.com/articles/2022/01/20/unix-sockets-are-now-supported-on-windows/

Rust doesn't support Windows UDS.

I am working on a PR to change this https://github.com/tokio-rs/mio/pull/1667

You asked the exact same question in my last support Windows PR so instead of answering it here I will link to the time I already answered this https://github.com/ethereum/trin/pull/739#discussion_r1196020473

ogenev (approved):
>> I never claimed Windows didn't support IPC I clearified Rust in my PR intro `Certain parts of Trin either won't compile on Windows or crash because rust doesn't support windows Unix IPC` --> `rust doesn't support windows Unix IPC`

I remember now the discussion about **Tokio** and windows uds. By claim, I was referring to the message in this panic: https://github.com/ethereum/trin/pull/923/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R34.

It was weird because it seems Rust [supports](https://crates.io/crates/uds_windows) windows Unix IPC, it is **Tokio** that has some issues with it.  So I guess fixing the PR description and the panic will make it less confusing?

KolbyML:
That is the library we used to use 2 years ago.

KolbyML:
Oh that panic error. Oh ok my bad I will fix that. I should really read panic errors I steal from my old PRs. Thank you for bringing it to light.

KolbyML:
Ok I resolved the panic error issue

carver (approved):
LGTM 🚢 

👍🏻 Yeah, switching from build to check can go in another PR.

PR #925, opened 2023-09-18
Add improvements to deployment.md

KolbyML:
### What was wrong?
When I was flamingo I ran into a few issues.
### How was it fixed?
By adding addition instructions for the things I missed.


carver (approved):
nice

PR #928, opened 2023-09-20
docs: remove kademlia article

danielrachi:
### What was wrong?

The article about Kademlia was useful for all the Portal Clients, not exclusive to Trin.

### How was it fixed?

Moved the info to the [portal network website](https://github.com/ethereum/portal-website) in ethereum/portal-website#7 and deleted it here.


njgheorghita (approved):
Looks like you'll also need to remove it here... https://github.com/ethereum/trin/blob/b43d206d0c7db984b47f15e64f8ddbcee612ba74/book/src/SUMMARY.md?plain=1#L36

PR #929, opened 2023-09-21
Portal Beacon Sync

ogenev:
### What was wrong?
It has been a while since we used this branch as a POC for beacon network sync. IMO, it is time to merge this and continue iterating over it.

### How was it fixed?
Implement sync mechanics to a beacon portal node.

### To-Do

- [ ] Clean up commit history


ethportal-api/src/types/cli.rs
```diff
@@ -255,6 +263,20 @@ fn check_private_key_length(private_key: &str) -> Result<H256, String> {
     ))
 }
 
+fn check_trusted_block_root(trusted_root: &str) -> Result<String, String> {
```
njgheorghita, line 266: This seems like an obvious candidate to be pulled out into `trin-utils` and consolidated w/ [this implementation](https://github.com/ethereum/trin/blob/092427a4f0babf6a8bc3423de84a89b4f244d0cd/ethportal-api/src/types/cli.rs#L248-L256)

light-client/src/consensus/rpc/portal_rpc.rs
```diff
@@ -0,0 +1,316 @@
...
+use ethportal_api::types::content_key::beacon::LightClientOptimisticUpdateKey;
+use ethportal_api::{
+    BeaconContentKey, BeaconContentValue, ContentValue, LightClientBootstrapKey,
+    LightClientUpdatesByRangeKey,
+};
+use eyre::eyre;
+use futures::channel::oneshot;
+use portalnet::overlay_service::OverlayCommand;
+use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::sync::mpsc::UnboundedSender;
```
njgheorghita, line 21: **nitpick** can you reorder the imports so we have a group of `std`, followed by a group of external libraries, followed by a group of internal imports (and in the other `light-client` libraries)

light-client/src/consensus/rpc/portal_rpc.rs
```diff
@@ -0,0 +1,316 @@
...
+use ethportal_api::types::consensus::light_client::finality_update::LightClientFinalityUpdate;
+use ethportal_api::types::consensus::light_client::optimistic_update::LightClientOptimisticUpdate;
+use ethportal_api::types::consensus::light_client::update::LightClientUpdate;
+use ethportal_api::types::content_key::beacon::LightClientFinalityUpdateKey;
+use ethportal_api::types::content_key::beacon::LightClientOptimisticUpdateKey;
+use ethportal_api::{
+    BeaconContentKey, BeaconContentValue, ContentValue, LightClientBootstrapKey,
+    LightClientUpdatesByRangeKey,
+};
+use eyre::eyre;
```
njgheorghita, line 17: I'm guessing the usage of this library is since it's grandfathered in from the original code we forked? Would it be a heavy lift (in a future pr) to convert over to `anyhow` so we have consistency amongst our codebases?

light-client/src/consensus/rpc/portal_rpc.rs
```diff
@@ -0,0 +1,316 @@
...
+
+    fn name(&self) -> String {
+        "portal".to_string()
+    }
+}
+
+pub fn expected_current_slot() -> u64 {
+    let now = SystemTime::now();
+    let now = now.duration_since(UNIX_EPOCH).expect("Time went backwards");
+    let since_genesis = now - std::time::Duration::from_secs(BEACON_GENESIS_TIME);
```
njgheorghita, line 313: Would like to see this import moved to the top. If it makes sense, I think you can even move the import to the const definition. eg. `const RETRY_AFTER: Duration = Duration::from_secs(15);`

light-client/src/main.rs
```diff
@@ -16,14 +17,14 @@ async fn main() -> Result<()> {
     builder = builder.network(networks::Network::Mainnet);
 
     // Set the consensus rpc url
-    builder = builder.consensus_rpc("https://www.lightclientdata.org");
+    builder = builder.consensus_rpc("http://testing.mainnet.beacon-api.nimbus.team");
```
njgheorghita, line 20: Let's move the updated values in this file to consts

light-client/src/node.rs
```diff
@@ -31,6 +31,19 @@ impl Node {
         })
     }
 
+    pub fn with_portal(config: Arc<Config>, portal_rpc: R) -> Result<Self, NodeError> {
+        let checkpoint_hash = &config.checkpoint.as_ref().unwrap();
```
njgheorghita, line 35: hmmmm... maybe in a future pr, but I'd like to see the clippy lints we use in our other workspaces added to this workspace
```rust
#![warn(clippy::unwrap_used)]
```

light-client/src/node.rs
```diff
@@ -31,6 +31,19 @@ impl Node {
...
+
+        let consensus =
+            ConsensusLightClient::with_custom_rpc(portal_rpc, checkpoint_hash, config.clone());
+
+        Ok(Node {
+            consensus,
+            config,
+            history_size: 64,
```
njgheorghita, line 43: define as const?

njgheorghita (approved):
:RUBBER STAMP: // 🚢

ethportal-api/src/types/cli.rs
ogenev: Not sure, I don't want to add `trin-utils` as a dependency of `ethportal-api` because ethportal-api should exist independently from trin.

light-client/src/consensus/rpc/portal_rpc.rs
ogenev: Yeah, I will create an issue after the merge.

light-client/src/consensus/rpc/portal_rpc.rs
ogenev: [rustfmt](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity) has such type of formats, it will be better if we start using them and keep this consistent through the repo, instead of manually changing those files.

light-client/src/node.rs
ogenev: Will create an issue to enable `#![warn(clippy::unwrap_used)]` and fix the unwraps.

light-client/src/node.rs
ogenev: Good catch, this is a leftover,  removed.

PR #935, opened 2023-09-27
fix(overlay-service): Check for peer `NodeId` in discovery ENR cache

ogenev:
### What was wrong?
Many `FindContent` queries fail when the result is an uTP stream.  This is because we cannot find the `ENR `of the source `NodeID` and reject the incoming uTP content.

This fix is expected to also improve `FindNode` queries.

### How was it fixed?
Check in the discovery ENR cache when trying to find an `ENR` for a known `NodeId` when processing `FIndContent` and `FindNode` queries.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -264,6 +264,14 @@ impl Discovery {
         self.node_addr_cache.write().get(node_id).cloned()
     }
 
+    /// Put a `NodeAddress` into cache. If the key already exists in the cache, then it updates the key's value and
+    /// returns the old value. Otherwise, `None` is returned.
+    pub fn put_cached_node_addr(&self, node_addr: NodeAddress) -> Option<NodeAddress> {
+        self.node_addr_cache
+            .write()
+            .put(node_addr.enr.node_id(), node_addr)
+    }
+
     /// Sends a TALKREQ message to `enr`.
     pub async fn send_talk_req(
         &self,
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2463,6 +2463,11 @@ where
             return Some(entry.value().clone().enr());
         }
 
+        // Check whether this node id is in our discovery ENR cache
+        if let Some(node_addr) = self.discovery.cached_node_addr(node_id) {
+            return Some(node_addr.enr);
+        }
+
         // Check the existing find node queries for the ENR.
         for (query_info, _) in self.find_node_query_pool.read().iter() {
             if let Some(enr) = query_info
@@ -2677,7 +2682,7 @@ mod tests {
     use rstest::rstest;
 
     use crate::{
-        discovery::Discovery,
+        discovery::{Discovery, NodeAddress},
         overlay::OverlayConfig,
         storage::{DistanceFunction, MemoryContentStore},
         types::messages::PortalnetConfig,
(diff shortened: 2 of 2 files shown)


portalnet/src/discovery.rs
```diff
@@ -264,6 +264,14 @@ impl Discovery {
         self.node_addr_cache.write().get(node_id).cloned()
     }
 
+    /// Put a `NodeAddress` into cache. If the key already exists in the cache, then it updates the key's value and
+    /// returns the old value. Otherwise, `None` is returned.
+    pub fn put_cached_node_addr(&self, node_addr: NodeAddress) -> Option<NodeAddress> {
```
njgheorghita, line 269: Looks like we're never calling this, besides in your test? Is there somewhere else in the production code that will update the `node_addr_cache`?

portalnet/src/overlay_service.rs
```diff
@@ -3632,6 +3638,20 @@ mod tests {
...
+
+        let node_addr = NodeAddress {
+            enr: enr3.clone(),
+            socket_addr: SocketAddr::V4(enr3.udp4_socket().unwrap()),
+        };
+
+        service.discovery.put_cached_node_addr(node_addr);
+
+        let found_enr3 = service.find_enr(&node_id_3).unwrap();
+        assert_eq!(found_enr3, enr3);
```
njgheorghita, line 3654: Should this test should include checking the case that the node id can still be found after the enr has been updated?

njgheorghita (commented):
Just a few questions so I can wrap my head around what's happening here, but it looks like a nice fix!

A thought I've had is... It seems like we're operating under the assumption that each `NodeId` correlates to a single network participant. Which is perfectly reasonable in theory. But in practice, we might have some (lazy) users using the same private keys (eg. `0x0000..`, `0x1111..`, `0x0101..`, etc). While I don't think this is something we need to account for, it might be worth keeping in mind as we build these caching mechanisms. Though, I also feel like there's very little we can do about this case, besides simply encouraging users (via docs & defaults) to use truly random private keys if they want their node to run cleanly.

portalnet/src/discovery.rs
ogenev: Not sure about a use case to add a node to the cache outside of discovery, besides in testing. The other way around is to expose the cache and make it public but I think it is more foolproof to add a discovery method for it.

portalnet/src/overlay_service.rs
ogenev: I don't feel so, the scope of this test is to check if `OverlayService` `find_enr` is looking into the discovery cache. I think updating Enr is out of scope here.

ogenev:
> A thought I've had is... It seems like we're operating under the assumption that each `NodeId` correlates to a single network participant. Which is perfectly reasonable in theory. But in practice, we might have some (lazy) users using the same private keys (eg. `0x0000..`, `0x1111..`, `0x0101..`, etc). While I don't think this is something we need to account for, it might be worth keeping in mind as we build these caching mechanisms. Though, I also feel like there's very little we can do about this case, besides simply encouraging users (via docs & defaults) to use truly random private keys if they want their node to run cleanly.

Well, if someone cares about privacy and security, their private key should be random and secure. Operating a staking node or crypto wallet uses the same concepts so I don't think we will need to do a lot of teaching here.

njgheorghita (approved):
:ship:

Issue #932, opened 2023-09-25
Install a TODO issue bot


carver:
Maybe later we want to leave the TODOs in the code and have a bot add a tracking issue in the repository? There's an example of this bot in uint, linked below. I was reminded of this by [Kolby's comment](https://github.com/ethereum/trin/pull/924#discussion_r1334929507).

See this github action that does the same thing (noting the related python in a neighbor folder):
https://github.com/recmo/uint/blob/64cd1bc929fa5737e724bd7db8d24bdcfbe2994f/.github/workflows/issue_tracker.yml

Plus an example issue created by the todo bot: https://github.com/recmo/uint/issues/301

PR #936, opened 2023-09-27
Refactor metrics to handle multiple subnetworks

njgheorghita:
### What was wrong?
Our current metrics registration setup is incompatible with running multiple subnetworks. Basically, each subnetwork tries to re-register an equivalently-named metric, which causes an error. 

The way we get around this in our current `StorageMetrics` is to append the subnetwork name to the metric name (eg `trin_radius_ratio_History`). But this isn't ideal, using a label is much better suited to distinguishing between the same metric for different subnetworks. Furthermore using a label makes creating interesting / useful dashboards in grafana much easier.

### How was it fixed?
- Created `PortalnetMetrics` which houses clonable references to both `StorageMetrics` & `OverlayMetrics`, which can be used by individual subnetworks
- moved `PortalnetConfig` from `portalnet/src/types/messages.rs` -> `portalnet/src/config.rs`

### To-Do

- [x] Clean up commit history
- [ ] Update dashboard json template (will address in following pr)


portalnet/src/overlay.rs
```diff
@@ -482,11 +479,13 @@ where
     ) -> anyhow::Result<()> {
         match self.validator.validate_content(content_key, content).await {
             Ok(_) => {
-                self.metrics.report_validation(true);
+                self.metrics
+                    .report_validation(&self.protocol.to_string(), true);
```
njgheorghita, line 483: My update to the metrics strategy requires the addition of a `protocol` argument to the metrics reporting functions. I don't love this. I tried a couple ideas to eliminate this design, but I couldn't find something satisfactorily simple. This pr is already somewhat complex, and that would have increased the complexity in exchange for simply adding a `protocol` argument. I opted to leave the simple design for now. But, I'd be curious to hear thoughts from others. Is it worth trying to refactor the `protocol` argument away? Any neat ideas on the best way to accomplish this?

carver:
Looks like this needs a rebase, so I'll just start with a high-level review

portalnet/src/config.rs
```diff
@@ -0,0 +1,56 @@
...
+        Self {
+            external_addr: None,
+            private_key: H256::random(),
+            listen_port: 4242,
+            bootnodes: Bootnodes::default(),
+            data_radius: Distance::MAX,
+            internal_ip: false,
+            no_stun: false,
+            node_addr_cache_capacity: NODE_ADDR_CACHE_CAPACITY,
+            metrics: PORTALNET_METRICS.overlay.clone(),
```
carver, line 53: I don't understand why the metrics are inside the config here. It doesn't really seem to belong with the other configured portal network values (simple entries from the CLI). I think the whole PR gets simpler, and the code makes more sense, if the config is left alone and the metrics are just handled separately.

carver (commented):
Since I suggested a high-level change, and there's a pending refactor, I'll just pause my review for now and resume when you're ready.

portalnet/src/overlay.rs
njgheorghita: Added `StorageMetricsReporter` & `OverlayMetricsReporter` to handle the per-subnetwork metrics reporting, avoiding the need to use an explicit `self.protocol` throughout `OverlayService` & `Storage`

portalnet/src/config.rs
njgheorghita: Yeah, good call. That design was cruft leftover from a previous design where I manually created the metrics registry and passed it through the configs to the respective services. Now using the `lazy_static` to globally initialize eliminates the need for this.

src/lib.rs
```diff
@@ -45,26 +45,19 @@ pub async fn run_trin(
...
-        listen_port: trin_config.discovery_port,
-        no_stun: trin_config.no_stun,
-        bootnodes: trin_config.bootnodes.clone(),
-        enr_file_location: Some(node_data_dir.clone()),
-        ..Default::default()
-    };
+    // Initialize prometheus metrics
+    if let Some(addr) = trin_config.enable_metrics_with_url {
+        prometheus_exporter::start(addr)?;
+    }
```
carver, line 51: The move of the exporter up here used to be required because we were passing it into the config, but we're not doing that anymore. So I think the code move is unnecessary, right?

If it is necessary, a comment explaining why is warranted. If not, *nit* to reduce code turnover a bit by moving the block back into place.

portalnet/src/discovery.rs
```diff
@@ -1,5 +1,8 @@
-use super::types::messages::{PortalnetConfig, ProtocolId};
-use crate::socket;
+use std::hash::{Hash, Hasher};
+use std::net::Ipv4Addr;
+use std::str::FromStr;
+use std::{convert::TryFrom, fmt, fs, io, net::SocketAddr, sync::Arc};
```
carver, line 4: I wonder if there's a clippy warning or something we can add to just have this done once in the code base and then not have to trickle it in over time. (I'm definitely happier with this python-like grouping of library imports at the top: system, 3rd-party, and local)

Are you doing this manually, or using some tool to reorder?

ethportal-api/src/dashboard/grafana.rs
```diff
@@ -12,6 +12,8 @@ pub struct GrafanaAPI {
     address: String,
 }
 
+// todo: automatically update datasource/dashboard via `create-dashboard` command
+// rather than deleting and recreating them
```
carver, line 16: To issue?

portalnet/src/metrics/overlay.rs
```diff
@@ -1,81 +1,84 @@
...
-            validation_count_options,
-            validation_count_labels,
-        );
-
-        Self {
-            protocol: protocol.into(),
+    pub fn new(registry: &Registry) -> anyhow::Result<Self> {
+        let message_count = register_int_counter_vec_with_registry!(
+            opts!(
+                "trin_message_count",
```
carver, line 30: `total` is the preferred suffix for a generic aggregating counter, according to this Prometheus naming "Best Practices" page: https://prometheus.io/docs/practices/naming/

portalnet/src/metrics/overlay.rs
```diff
@@ -1,81 +1,84 @@
...
+                "count all content validations successful and failed"
+            ),
+            &["protocol", "success"],
+            registry
+        )?;
+        Ok(Self {
             message_count,
             utp_outcome_count,
-            utp_active_count,
+            utp_active_gauge,
```
carver, line 63: 👍🏻 for the name improvement

portalnet/src/metrics/overlay.rs
```diff
@@ -1,81 +1,84 @@
...
+
+#[derive(Clone)]
+pub struct OverlayMetricsReporter {
+    pub protocol: String,
+    pub overlay_metrics: OverlayMetrics,
+}
+
+impl OverlayMetricsReporter {
+    pub fn message_total(&self, direction: &str, message_type: &str) {
+        let protocol: &str = &self.protocol.to_string();
```
carver, line 77: `protocol` is already a `String`, it seems like `to_string()` is doing an extra memory allocation with no benefit. (and repeated below)

carver (commented):
I didn't get through everything, but figured I would send the partial review for you to chew on. I'll wrap it up tomorrow.

portalnet/src/discovery.rs
njgheorghita: https://github.com/rust-lang/style-team/issues/131#issuecomment-995316504

https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports

I've just been doing this manually (following the python-like grouping) whenever I'm working on a file that I notice doesn't conform. Agree that this isn't ideal, but it also looks like the `rustfmt` lint is only available on `nightly`. 

What do you think? I could go through the codebase, and update locally using the `rustfmt` lint. But I'm not sure we have a solid way forward to prevent deterioration over time.

ethportal-api/src/dashboard/grafana.rs
njgheorghita: Well... Imo, this is a very low priority issue. Given the increased activity of our stale issue bot, I don't think it will be addressed anytime soon and will simply be marked as stale before anybody fixes this. 

Given that I updated the docs (in `book/src/users/monitoring.md`) I just went ahead and removed this `todo`. But lmk if you think it's worth an issue & I'll create one

portalnet/src/discovery.rs
carver: Yeah, I think the status quo is fine for now.

ethportal-api/src/dashboard/grafana.rs
carver: Eh, I guess I'm still not fully sold on the stale bot. The idea that we are hesitant to create issues that aren't going to be quickly resolved feels like an anti-pattern to me. I don't really see why TODOs peppered throughout the code is any better than a backlog of issues that have been around for a while (I'd argue it is worse, in fact. Partially because it's hard to have a conversation in a code TODO). But I guess the solution I'm leaning toward would handle the TODOs in code: https://github.com/ethereum/trin/issues/932 so I'm fine leaving it for now

carver (approved):
With the metric renames, presumably `./ethportal-api/src/dashboard/collected-metrics-dashboard.json.template` has gone stale. It's a little [clunky to update](https://github.com/ethereum/trin/pull/685#issuecomment-1518099365), but better than forcing everyone to figure out how to manually update their dashboard manually, especially since you should have the best idea what metrics changed to what names. Would you double-check that it's updated?

Ok, I feel a little uneasy that I might have missed something subtle in the refactor, but other than that and having the dashboard get created correctly, it LGTM!

👍🏻 for a little more Prometheus naming consistency too.

PR #937, opened 2023-09-28
feat: add increasing Enr sequence on updated Enr

KolbyML:
### What was wrong?
We didn't update the sequence number when we updated fields in our Enr.
### How was it fixed?
If in our ``folderpath/nodeid/trin.enr`` did exist we either update or set the enr sequence according to if we updated a field or not relative to the Enr saved to disk. If a file doesn't exist that means the Enr is new and we set it to 1.

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -19,7 +19,7 @@ use ethportal_api::NodeInfo;
 use std::hash::{Hash, Hasher};
 use std::net::Ipv4Addr;
 use std::str::FromStr;
-use std::{convert::TryFrom, fmt, io, net::SocketAddr, sync::Arc};
+use std::{convert::TryFrom, fmt, fs, io, net::SocketAddr, sync::Arc};
 use trin_utils::version::get_trin_version;
 
 /// Size of the buffer of the Discv5 TALKREQ channel.
@@ -89,7 +89,7 @@ impl Discovery {
             CombinedKey::secp256k1_from_bytes(portal_config.private_key.0.clone().as_mut_slice())
                 .map_err(|e| format!("Unable to create enr key: {:?}", e.to_string()))?;
 
-        let enr = {
+        let mut enr = {
             let mut builder = EnrBuilder::new("v4");
             if let Some(ip_address) = enr_address {
                 builder.ip(ip_address);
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):
Nice! Any opportunity you see to add a simple test for this functionality? I'm not sure it deserves it's own peertest scenario, but maybe a couple of unit tests in this module, building `Discovery` from a `PortalnetConfig::default()` to validate that the correct enr is being written to disk & updated.

src/lib.rs
```diff
@@ -51,6 +51,7 @@ pub async fn run_trin(
         listen_port: trin_config.discovery_port,
         no_stun: trin_config.no_stun,
         bootnodes: trin_config.bootnodes.clone(),
+        node_data_dir: Some(node_data_dir.clone()),
```
njgheorghita, line 54: Something feels a bit off about this. It doesn't make any sense that `node_data_dir` could be `None`. A trin node should never have a `None` data dir. It seems like you went with an option here to support `PortalnetConfig::default()`?

Imo, maybe the best way forward is just to rename this variable, maybe something like `enr_record_file`. Since `enr_record_file`, in theory, could be none. eg. every node/datadir that runs this code for the first time won't have an enr record file.

portalnet/src/discovery.rs
```diff
@@ -106,6 +106,27 @@ impl Discovery {
                 .map_err(|e| format!("When adding key to servers ENR: {e:?}"))?
         };
 
+        // Check if we have an old version of our Enr and if we do, increase our sequence number
```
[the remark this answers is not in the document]
KolbyML: Double checked and I couldn't find anywhere in our project where we update while running.

portalnet/src/discovery.rs
```diff
@@ -106,6 +106,27 @@ impl Discovery {
...
+        if let Some(node_data_dir) = portal_config.node_data_dir {
+            let trin_enr = node_data_dir.join("trin.enr");
```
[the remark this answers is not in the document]
KolbyML: resolved

src/lib.rs
KolbyML: Sounds good I will use enr_file_location, as enr (ethereum node record), already has record and just file could be it is loaded into memory or something, good catch

KolbyML:
Thank you for recommending I write a test for this, that was a great idea!!! I learnt a good bit about eliptic curves and figured out that you can have multiple signatures which prove the same thing, that is a very helpful edge case to be aware of !!!

src/lib.rs
```diff
@@ -51,6 +51,7 @@ pub async fn run_trin(
         listen_port: trin_config.discovery_port,
         no_stun: trin_config.no_stun,
         bootnodes: trin_config.bootnodes.clone(),
+        enr_file_location: Some(node_data_dir.clone()),
```
njgheorghita, line 54: Hey @KolbyML I think this is still somewhat confusing. The name `enr_file_location` implies that the variable is a path to the enr file itself.  But it's actually representing an `Option` for the data dir. A trin node data dir should never be an option, since we always have a data dir. We're also joining the file name to `enr_file_location` inside `discovery.rs`, which is also confusing (eg joining the enr file location to itself).

In this case, I think the simplest way forward is to replace `enr_file_location` with `trin_data_dir` in `PortalnetConfig`

KolbyML: Ok that is fine. I will need to append ''nodeid/trin.enr'' to the path I write to then in Discovery::new. Also that sounds about right on it not being an option.

KolbyML: Age Manning gaving me some insight on  my PR on sigp/enr so I wanted to do a few changes as well on how it worked!

PR #942, opened 2023-09-29
chore: add reth-rpc-types to serialize block to json-rpc

carver:
### What was wrong?

We didn't have a clean data structure for representing the combined header/block thing that's returned by `eth_getBlockByHash` and friends.

### How was it fixed?

Import the reth-rpc-types for `Block` and add a conversion utility from the trin `Header` into the reth `Header` which is a field in `Block`.

To understand more about how this is used, see #908 
 
### To-Do

- [ ] Clean up commit history

### DIFF
--- a/ethportal-api/src/types/execution/header.rs
+++ b/ethportal-api/src/types/execution/header.rs
@@ -1,5 +1,7 @@
 use ethereum_types::{Bloom, H160, H256, H64, U256};
+use reth_rpc_types::Header as RpcHeader;
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
+use ruint::Uint;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use ssz::{Encode, SszDecoderBuilder, SszEncoder};
 use ssz_derive::{Decode, Encode};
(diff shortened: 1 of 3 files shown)


carver:
Don't let the line count spook you. It's mostly in the Cargo.lock.

carver:
@KolbyML any idea what's going on here with Windows support?
```
  C:\Users\circleci\.cargo\git\checkouts\c-kzg-4844-87b9b7bbf5cb76f8\f5f6f86\src\c_kzg_4844.h:27:10: fatal error: 'stdio.h' file not found
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ClangDiagnostic("C:\\Users\\circleci\\.cargo\\git\\checkouts\\c-kzg-4844-87b9b7bbf5cb76f8\\f5f6f86\\src\\c_kzg_4844.h:27:10: fatal error: 'stdio.h' file not found\n")', C:\Users\circleci\.cargo\git\checkouts\c-kzg-4844-87b9b7bbf5cb76f8\f5f6f86\bindings\rust\build.rs:157:10
```

KolbyML:
We are currently using Windows server 2019 I will quickly see if I can get 2022 running and test if it works

KolbyML:
It seems like updating the CI to a newer windows server resolved the issue here is a PR with the CI fix https://github.com/ethereum/trin/pull/945

ethportal-api/src/types/execution/header.rs
```diff
@@ -195,6 +197,73 @@ impl PartialEq for Header {
...
+            withdrawals_root: withdrawals_root.map(|root| root.to_fixed_bytes().into()),
+            blob_gas_used: None,
+            excess_blob_gas: None,
+            hash,
+            parent_beacon_block_root: None,
+        }
+    }
+}
+
+fn u256_to_uint256(u256: U256) -> Uint<256, 4> {
```
njgheorghita, line 251: It seems like we could create our own custom wrapped type around `Uint<256, 4>` & then implement `From` rather than these conversion methods. Seems slightly cleaner, but also wrapped types can be annoying to manage, so I'm perfectly fine leaving this as is.

ethportal-api/src/types/execution/header.rs
```diff
@@ -195,6 +197,73 @@ impl PartialEq for Header {
     }
 }
 
+impl From<Header> for RpcHeader {
```
njgheorghita, line 200: Could you add a docstring comment here explaining why we need this `RpcHeader` type.

njgheorghita (approved):
:ship:

ethportal-api/src/types/execution/header.rs
carver: Ok, I'll take a look. My intuition is that wrapping it just to unwrap it immediately is probably not worth the coding overhead, but I'll explore it a bit.

carver:
```patch
diff --git a/ethportal-api/src/types/execution/header.rs b/ethportal-api/src/types/execution/header.rs
index e9407b8..91598ed 100644
--- a/ethportal-api/src/types/execution/header.rs
+++ b/ethportal-api/src/types/execution/header.rs
@@ -1,3 +1,5 @@
+use std::ops::Deref;
+
 use ethereum_types::{Bloom, H160, H256, H64, U256};
 use reth_rpc_types::Header as RpcHeader;
 use rlp::{Decodable, DecoderError, Encodable, Rlp, RlpStream};
@@ -232,7 +234,7 @@ impl From<Header> for RpcHeader {
             transactions_root: transactions_root.to_fixed_bytes().into(),
             receipts_root: receipts_root.to_fixed_bytes().into(),
             logs_bloom: logs_bloom.to_fixed_bytes().into(),
-            difficulty: u256_to_uint256(difficulty),
+            difficulty: *RethUint256::from(difficulty),
             number: Some(u64_to_uint256(number)),
             gas_limit: u256_to_uint256(gas_limit),
             gas_used: u256_to_uint256(gas_used),
@@ -252,6 +254,24 @@ impl From<Header> for RpcHeader {
     }
 }
 
+struct RethUint256(Uint<256, 4>);
+
+impl From<U256> for RethUint256 {
+    fn from(u256: U256) -> Self {
+        let mut bytes = [0u8; 32];
+        u256.to_big_endian(&mut bytes);
+        Self(Uint::from_be_bytes(bytes))
+    }
+}
+
+impl Deref for RethUint256 {
+    type Target = Uint<256, 4>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
 fn u256_to_uint256(u256: U256) -> Uint<256, 4> {
     let mut bytes = [0u8; 32];
     u256.to_big_endian(&mut bytes);
```

Not an obvious win to me. I think I'll leave it as-is unless you're seeing something I'm not about why it's better.

PR #943, opened 2023-09-29
refactor: reuse logic to fetch history content

carver:
Want to reuse the logic to make requests to the history network, in #908 

### To-Do

- [x] Clean up commit history


njgheorghita (approved):
:ship:

PR #944, opened 2023-09-29
chore: add channel from EthApi to history network

carver:
Need access to the history network to respond to `eth_` rpc requests

### To-Do

- [ ] Clean up commit history


carver:
Splitting this out to make #908 a little smaller

carver:
The main PR is ready for review now, might as well just fold it in.

PR #945, opened 2023-09-30
chore: update windows server ci version from 2019 to 2022

KolbyML:
### What was wrong?
CI broke for windows compiling c-kzg-4844 in pr https://github.com/ethereum/trin/pull/942
### How was it fixed?
By updating our windows server version from 2019 to 2022

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ commands:
           command: sudo apt install clang
 orbs:
   rust: circleci/rust@1.6.0
-  win: circleci/windows@2.2.0
+  win: circleci/windows@5.0
 executors:
   docker-publisher:
     environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -171,20 +171,13 @@ jobs:
...
-      - run:
-          name: Add target
-          command: rustup target add x86_64-pc-windows-msvc
-      - run:
-          name: Install target toolchain
-          command: rustup toolchain install stable-x86_64-pc-windows-msvc
       - run:
           name: Check Trin workspace
-          command: cargo check --workspace
+          # We are running all these together because this version of circleci windows has an environment variable bug
```
carver, line 176: Can you link to an explanation of the bug, or describe a bit more what's happening here if no reference is available? When we find this comment later, we might want to evaluate if the bug has been resolved and we can split the steps back out. We will appreciate if we leave ourselves more hints/direction on how to do that.

carver (approved):
Thanks for hunting this down!

carver:
Hm, I tried this patch on #942 and it still failed. Haven't looked at why, yet

.circleci/config.yml
KolbyML: resolved

PR #952, opened 2023-10-03
fix: enr signature changes on restart even if enr content is the same

KolbyML:
### What was wrong?
I made a PR https://github.com/sigp/enr/pull/53 and Age Manning gave me the insight that the signature should always be the same in the client if the content is the same.

We had an issue before my increase seq pr and after that even though our enr could be the same, the signature would change which isn't good.
### How was it fixed?

It was fixed by restructuring a bit of my code. I also cleaned it up a bit for how to compare them from suggestions from Age and Nick.

I also resolved a param name nitpik nick had in my last PR.

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -486,10 +486,7 @@ mod tests {
         // test if Enr changes the Enr sequence is increased and if it is written to disk
         portalnet_config.listen_port = 2424;
         let discovery = Discovery::new(portalnet_config.clone()).unwrap();
-        assert_ne!(
-            get_enr_rlp_content(&discovery.local_enr()),
-            get_enr_rlp_content(&old_enr)
-        );
+        assert_ne!(discovery.local_enr(), old_enr);
         let data = fs::read_to_string(trin_enr_file_location.clone()).unwrap();
         let old_enr = Enr::from_str(&data).unwrap();
         assert_eq!(discovery.local_enr().seq(), 2);
@@ -498,10 +495,7 @@ mod tests {
 
         // test if the enr isn't changed that it's sequence stays the same
         let discovery = Discovery::new(portalnet_config).unwrap();
-        assert_eq!(
-            get_enr_rlp_content(&discovery.local_enr()),
-            get_enr_rlp_content(&old_enr)
-        );
+        assert_eq!(discovery.local_enr(), old_enr);
         let data = fs::read_to_string(trin_enr_file_location).unwrap();
         let old_enr = Enr::from_str(&data).unwrap();
         assert_eq!(discovery.local_enr().seq(), 2);
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -116,7 +116,7 @@ impl Discovery {
         if trin_enr_path.is_file() {
             let data = fs::read_to_string(trin_enr_path.clone())
                 .expect("Unable to read Trin Enr from file");
-            let old_enr = Enr::from_str(&data).expect("Expected read trin.enr to be valid");
+            let old_enr = Enr::from_str(&data).expect("Expected to read valid Trin Enr from file");
             enr.set_seq(old_enr.seq(), &enr_key)
                 .expect("Unable to set Enr sequence number");
 
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -16,6 +16,7 @@ use rlp::RlpStream;
 use serde_json::{json, Value};
 use std::hash::{Hash, Hasher};
 use std::net::Ipv4Addr;
+use std::path::PathBuf;
 use std::str::FromStr;
 use std::{convert::TryFrom, fmt, fs, io, net::SocketAddr, sync::Arc};
 use tokio::sync::mpsc;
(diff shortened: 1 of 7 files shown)


portalnet/src/discovery.rs
```diff
@@ -425,7 +430,7 @@ impl AsyncUdpSocket<UtpEnr> for Discv5UdpSocket {
 
 // todo: remove this once sigp/enr implements this for enr's
 // we need this because signatures can be different for the same data but still valid
-fn verify_by_signature(enr: &Enr, signature: &[u8]) -> bool {
+fn get_enr_rlp_content(enr: &Enr) -> BytesMut {
```
divagant-martian, line 433: @KolbyML I don't think there is any harm in exposing `rlp_content` from `Enr`

KolbyML: When we get the update Enr version we can just use compare_content. If we want to expose rlp_content() that is fine too. It just looked like rlp_content() was intentionally privated so that is why I tried to get a compare* function added. Cause I didn't want to manually calculate rlp_content in multiple codebases.

Anyways up to you. I didn't know making rlp_content() was an option tbh.

KolbyML: I do think compare_content is cleaner though
``enr1.compare_content(&enr2)``
vs
``enr1.rlp_content() == enr2.rlp_content()``

divagant-martian: Neither is a problem. Nothing wrong with exposing it, this does not change the enr and it's not private info in any way

portalnet/src/types/messages.rs
```diff
@@ -182,7 +182,7 @@ impl Default for PortalnetConfig {
             data_radius: Distance::MAX,
             internal_ip: false,
             no_stun: false,
-            enr_file_location: None,
+            trin_data_dir: PathBuf::default(),
```
njgheorghita, line 185: Well, sorry to be a stickler, but this still feels off. Essentially we're still doing the same thing I was previously against, except this time with `PathBuf::default()` instead of an `Option`

Ultimately, it feels like `PortalnetConfig` is not an appropriate place for this argument, since there is no reasonable default value. 

I think we just remove this from `PortalnetConfig` and pass `node_data_dir` into `Discovery::new()` directly. Thoughts?

portalnet/src/discovery.rs
```diff
@@ -111,28 +111,33 @@ impl Discovery {
                 .map_err(|e| format!("When adding key to servers ENR: {e:?}"))?
         };
 
-        // Check if we have an old version of our Enr and if we do, increase our sequence number
-        if let Some(enr_file_location) = portal_config.enr_file_location {
-            let trin_enr_file_location = enr_file_location.join(ENR_FILE_NAME);
+        // This should only never run in test cases.
+        if portal_config.trin_data_dir != PathBuf::default() {
+            // Check if we have an old version of our Enr and if we do, increase our sequence number
+            let trin_enr_file_location = portal_config.trin_data_dir.join(ENR_FILE_NAME);
```
njgheorghita, line 117: I would change to `trin_enr_path` - basically, `_path` suffix to these types of variable names describing a file path is what we typically do in this codebase, and I think we should continue with that convention to maintain consistency
 - https://github.com/ethereum/trin/blob/040ad41f66b17733d2e7c2a07b6b07644dac8225/portal-bridge/src/cli.rs#L142
 - https://github.com/ethereum/trin/blob/040ad41f66b17733d2e7c2a07b6b07644dac8225/portal-bridge/src/bridge.rs#L50
 - https://github.com/ethereum/trin/blob/040ad41f66b17733d2e7c2a07b6b07644dac8225/rpc/src/lib.rs#L74

portalnet/src/discovery.rs
```diff
@@ -111,28 +111,33 @@ impl Discovery {
                 .map_err(|e| format!("When adding key to servers ENR: {e:?}"))?
         };
 
-        // Check if we have an old version of our Enr and if we do, increase our sequence number
-        if let Some(enr_file_location) = portal_config.enr_file_location {
-            let trin_enr_file_location = enr_file_location.join(ENR_FILE_NAME);
+        // This should only never run in test cases.
```
njgheorghita, line 114: `// This should never run in test cases.` - much less confusing in my opinion. But also the fact that this check needs to happen feels off. If we go with my suggestion about `node_data_dir` then we can just get rid of this check completely

KolbyML: The issue I have with this is some Discovery tests don't use a path, but I cause I can just generate a tmp path for those

KolbyML:
@njgheorghita I believe I resolved your concerns let me know what you think

portalnet/src/discovery.rs
```diff
@@ -111,33 +111,29 @@ impl Discovery {
...
-                    // the content is the same, we don't want to change signatures on restart
-                    // so set enr to old one to keep the same signature per sequence number
-                    enr = old_enr;
-                }
+        // Check if we have an old version of our Enr and if we do, increase our sequence number
+        let trin_enr_path = node_data_dir.join(ENR_FILE_NAME);
+        if trin_enr_path.is_file() {
+            let data = fs::read_to_string(trin_enr_path.clone())
+                .expect("Unable to read Trin Enr from file");
+            let old_enr = Enr::from_str(&data).expect("Expected read trin.enr to be valid");
```
njgheorghita, line 119: *nit* `"Expected to read valid Trin Enr from file"`

njgheorghita (approved):
Nice! Yeah, it's not ideal to create all those temp dirs in the tests but in the long run I think we're left off with better / easier to read production code this way

PR #959, opened 2023-10-05
Order imports by std, external, and internal

njgheorghita:
### What was wrong?
See conversation here, but the import statements throughout our codebase do not have a consistent grouping. It is unlikely that a lint for this will become available on `stable` channel anytime soon, so for the time being I just ran the [group_imports](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports) lint using the `StdExternalCrate` level.

### To-Do

- [ ] Clean up commit history

### DIFF
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -1,3 +1,6 @@
+use discv5::enr::NodeId;
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+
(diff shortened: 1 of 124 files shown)


njgheorghita:
In hindsight, this linter doesn't exactly treat all workspace members as "internal" libraries, which I think defeats the purpose of the orderings. I'm really not interested in going through and updating each module individually right now. But, if anybody feels like this could be useful, i'll keep that in mind next time i'm flamingo. Closing this for now.

PR #963, opened 2023-10-06
Add transaction hashes and uncles to the block that json-rpc returns

carver:
### What was wrong?

Fix #279 

### To-Do

- [x] Clean up commit history
- [x] Test that txn hashes match expected value
- [x] ~Add uncles~ Hah right, [there are no uncles post-merge](https://ethereum.stackexchange.com/questions/139390/are-uncle-ommer-blocks-still-possible-after-the-merge).
- [x] rebase all the things

PR #964, opened 2023-10-06
chore: pin reth dependencies to a release

carver:
Building against a moving target is hard. It's just good practice to have everyone working against a consistent version of the dependency anyway. The "version" attribute was basically ignored by cargo.

So those are the theoretical reasons. Here is a practical one: Basic things are changing about the RPC structs, like the type of the transaction hashes, even in just the last few weeks.

Split from #963 

- [ ] Clean up commit history

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1133,7 +1133,7 @@ dependencies = [
 [[package]]
 name = "codecs-derive"
 version = "0.1.0-alpha.10"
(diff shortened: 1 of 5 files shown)


njgheorghita (approved):
:ship: Agree about this change, but seems like we want to occasionally update the pinned dependency version, so we don't get too far out of date as we try to further integrate the library? No good thoughts on how often / when exactly to do that however

carver:
Yeah, I suppose one mechanism is we'll be forced to update as we push features upstream. (which we will likely end up doing for the state network)

But I agree that it would be nice to have more. I opened #969 to track

PR #965, opened 2023-10-06
feat(ethportal-api): add hash() to Transaction

carver:
### What was wrong?

It seems like transactions didn't have a hash method. Maybe I missed it?

### How was it fixed?

Added one.

This will end up being tested in integration, by the `eth_getBlockByHash` tests.

Split from #963 

### To-Do
- [ ] Clean up commit history


njgheorghita (commented):
Seems like the kind of change that's worth a test or two of at least an example for each tx type?

carver:
Yeah, I agree. Maybe I was hesitant because I thought there was a 50/50 chance that I was wrong and hash was implemented somewhere but I didn't know how to call it 😆

carver:
I added some hash checks against the post-Shanghai block already tested in the block body section. Checked both Legacy and EIP-1559 transactions.

njgheorghita (approved):
:ship:

PR #966, opened 2023-10-06
test: post-shanghai block for eth_getBlockByHash

carver:
Importantly, use a header that we have a block body for.

### What was wrong?

I didn't have easy access to the block body value for the header I was using for the json getBlock test. Also, our Devconnect target is for post-Shanghai, so that seems the more important era to test right now.

### How was it fixed?

Pulled in (plus restructured) part of this file from portal-hive:
`simulators/portal-interop/src/test-data/test_data_collection_of_forks_blocks.yaml`

Split from #963 

### To-Do

- [ ] Clean up commit history


tests/rpc_server.rs
```diff
@@ -156,12 +156,13 @@ async fn test_eth_get_block_by_hash() {
...
-    let json = json.as_object().unwrap();
-    let raw_header = json.get("1000001").unwrap().as_object().unwrap();
-    let raw_header = raw_header.get("value").unwrap().as_str().unwrap();
-    let hwp_ssz = hex_decode(raw_header).unwrap();
-    HeaderWithProof::from_ssz_bytes(&hwp_ssz).unwrap()
+    let file = fs::read_to_string("trin-validation/src/assets/hive/blocks.yaml").unwrap();
+    let value: Value = serde_yaml::from_str(&file).unwrap();
+    let all_blocks = value.as_sequence().unwrap();
+    let post_shanghai = all_blocks.last().unwrap();
+    assert_eq!(post_shanghai["number"], 1751000);
```
carver, line 163: The point of this assertion right now is that if someone appends a new block into the yaml file, then this function would start testing a different block. So this assertion is a way of adding some confidence that we're testing the block that we think we are.

It probably makes sense for me to add that comment inline into the code above.

njgheorghita: Now that we're using yaml and comments are supported, seems like it might also be worth a trailing comment at the bottom of the yaml file

njgheorghita (commented):
I understand why we're adding the post-shanghai block, but don't understand why we're removing the test for the original (`1000001`) block? Why aren't we leaving that test case in?

carver:
> I understand why we're adding the post-shanghai block, but don't understand why we're removing the test for the original (`1000001`) block? Why aren't we leaving that test case in?

The asset file doesn't have the block body for that block number. Without the body, the test will fail, because it can't retrieve the transactions. Note that this PR change *only* affects the `eth_getBlockByHash` tests of rpc, not all the other places that 1000001 is tested.

It should be straightforward to port over more of the older block assets from the same place I got this one (portal-hive), when support for older forks is a priority.

njgheorghita (approved):
:shipit:

Issue #752, opened 2023-05-24
Refactor duplication in overlay RPC


ogenev:
There is quite a duplication code when implementing the base RPC endpoints for every network.

For example, we can see that most of the code in [history rpc ](https://github.com/ethereum/trin/blob/master/trin-history/src/jsonrpc.rs) overlaps with [beacon rpc](https://github.com/ethereum/trin/blob/master/trin-beacon/src/jsonrpc.rs). With the implementation of the State network, this duplication will increase.

I think one piece of refactoring that we want here is to make the common methods generics, i.e. accepting generic `network` and `content_key`/`content_value` arguments. Then we can extract those methods in a separate common module in a way that every network crate has access to it.

KolbyML:
Maybe I could look at this since it would accomplish what I want https://github.com/ethereum/trin/issues/753 cleaner. Let me know if you are already working on this or planning (I might poke at it today) but it is alright either way.

rakshit087:
I am new to rust, so I might be completly wrong. As far as I understand, we can reduce the duplication code by making a trait for "new" function in Beacon Network and HistoryNetwork structs. 
By this we will be able to make the generics strictly typed ( structs that implement the traits ).

ogenev:
> I am new to rust, so I might be completly wrong. As far as I understand, we can reduce the duplication code by making a trait for "new" function in Beacon Network and HistoryNetwork structs. By this we will be able to make the generics strictly typed ( structs that implement the traits ).

Yes and then we can refactor the methods to accept a type that implements this trait and refactor the common code by adding new trait implementation for `HistoryRequestHandler` and `BeaconRequestHandler`.

PR #971, opened 2023-10-07
Refactor NetworkApiServer #752

muneeb-i-khan:
### What was wrong?
Code repetition in history.rs and beacon.rs.
### How was it fixed?
Made the code generic by defining common traits under network.rs

### DIFF
--- a/ethportal-api/src/lib.rs
+++ b/ethportal-api/src/lib.rs
@@ -11,14 +11,16 @@ mod dashboard;
 pub mod discv5;
 mod eth;
 mod history;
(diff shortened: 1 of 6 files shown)


KolbyML:
Currently ci lint and test are failing. If you would be able to resolve those issues that would be great. There is a details button you can click where you will see why they failed. There are instructions in the trin book https://ethereum.github.io/trin/developers/contributing/git/pull_requests.html?highlight=clipp#pull-requests on how you should be able to resolve this.

Issue #962, opened 2023-10-06
Portal bridge should guarantee that data was successfully bridged


mrferris:
Portal bridge cannot currently provide either information nor guarantees about how many nodes successfully received the bridged content. 

Trin's `propogate_gossip` function (which is used by portal bridge) is not aware of how many populated `ACCEPT` messages were received by trin in response to the `OFFER` messages that it sends, nor how many resulting uTP streams succeeded.

njgheorghita:
This was resolved in #1032

PR #975, opened 2023-10-10
Relocate gossip code to separate module

njgheorghita:
### What was wrong?
A bit of preparation for #962. Basically, in order to track the success rate & peers from gossip results, we'll need to introduce a new `propagate_*` utility function. In order to reduce the bloat to `overlay_service.rs` it seems like a good time to move this code to it's own module. 

### How was it fixed?
- Moved gossip related code to separate module.

### To-Do

- [x] Clean up commit history


KolbyML (approved):
:shipit:

mrferris (approved):

PR #953, opened 2023-10-04
Add convention commit lint to ci

njgheorghita:
### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [ ] Clean up commit history

PR #976, opened 2023-10-11
chore: Sunset towncrier release notes

ogenev:
### What was wrong?

Towncrier is not used anymore for generating release notes. We switched to conventional commits and GitHub releases.

### How was it fixed?

- remove `newsfragment` folder
- Update book commit message guide.
- Remove duplicated files from `docs` folder
- Update the PR template

### To-Do

- [ ] Clean up commit history


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,4 +5,4 @@
 ### To-Do
 
 [//]: # (Stay ahead of things, add list items here!)
-- [ ] Clean up commit history
+- [ ] Clean up commit history and use conventional commits
```
njgheorghita, line 8: I just closed #953 since it requires a 3rd-party orb, usage of which is not authorized (and will not be) on our circleci account. But I do think adding a lint step to our ci build which validates this is better than simply stating that commits should be "conventional" (tracked in #979). Either way, it's not a blocker for this pr and this is great for the time being. 

One thing I would change here is to just make this a hyperlink to "https://www.conventionalcommits.org/en/v1.0.0/" for easy reference

book/src/developers/contributing/git/commits.md
```diff
@@ -18,16 +18,22 @@ written.  Once things are complete, you can `git reset master` to unstage all
 of the changes you've made, and then re-commit them in small chunks using `git
 add -p`.
 
-## Commit Messages
+### Commit Messages
```
njgheorghita, line 21: Nice catch that our docs were getting out of date here!

njgheorghita (approved):
🚢

carver (approved):
💯

PR #977, opened 2023-10-11
chore: Rename `docs` folder to `etc` and move Grafana dashboard

ogenev:
### What was wrong?
Repo's `docs` folder is redundant because we use the book to manage our documentation.

### How was it fixed?
- Rename `docs` folder to `etc`. This folder will be used to store miscellaneous files like Grafana dashboards, Prometheus configurations, etc.
- Move Grafana dashboard config from `ethportal-api` to `etc`.

### To-Do

- [x] Clean up commit history


njgheorghita:
To me it looks like we can get rid of `docs/release_notes.md` & `docs/contributing.md`. Which only leaves the `prometheus.yml` behind. Imo it's not worth a root-level directory for a single config. I'd be in support of in-lining this config in the monitoring docs and then getting rid of this directory entirely

ogenev:
> Imo it's not worth a root-level directory for a single config. I'd be in support of in-lining this config in the monitoring docs and then getting rid of this directory entirely

Not sure about that, we also have a Grafana dashboard config that does not belong to the `ethportal-api` crate IMO.
This is why I moved it into this new folder. I expect to have more configs/dashboards in the future so it will be nice to have a place to store everything that does not belong to other crates.

 I got this idea from[ reth](https://github.com/paradigmxyz/reth/tree/18311976a1a8f517f52a268b9dbfae6d5b45f506/etc).

njgheorghita (approved):
I'm on board with this `etc/` ... just a final thought that it might be a good location for our root-level `test_assets/` as well?

ethportal-api/src/dashboard/grafana.rs
```diff
@@ -5,7 +5,7 @@ use std::fs;
 use ureq;
 
 pub const DASHBOARD_TEMPLATES: &[&str] =
-    &["./ethportal-api/src/dashboard/collected-metrics-dashboard.json.template"];
+    &["../etc/grafana/dashboards/collected-metrics-dashboard.json.template"];
```
carver, line 8: I'm wondering if this breaks `create_dashboard()` when running it from the released ethportal-api crate, which I suppose wouldn't include this `/etc` folder.

ogenev: Well, `create_dashboard()` is used by `ethportal-api/types/cli.rs` file but I don't think the trin cli file belongs to `ethportal-api` in the first place. I can't see a use case for using the cli without trin...

PR #980, opened 2023-10-11
fix: bug ping rpc returns our radius instead of target's

KolbyML:
### What was wrong?
When I would ping nodes I would recieve the wrong data_radius, turns out ping was giving me our data radius back
### How was it fixed?
By decoding the data radius from the pong instead of returning our nodes data_radius

### DIFF
--- a/trin-beacon/src/jsonrpc.rs
+++ b/trin-beacon/src/jsonrpc.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use discv5::enr::NodeId;
 use ethportal_api::types::content_value::ContentValue;
+use ethportal_api::types::distance::Distance;
 use ethportal_api::types::jsonrpc::endpoints::BeaconEndpoint;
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
Nice find! This certainly feels like something to me that our tests should have caught at some point. Though, I guess [it is](https://github.com/ethereum/trin/blob/master/ethportal-peertest/src/scenarios/basic.rs#L80) - and simply both peers were using the same data radius explains how it slipped through the cracks. 

I don't know if this case justifies a special peertest / hive scenario? Unless you can think of a low-effort way to test this case?

KolbyML:
@njgheorghita I want to make a hive test for this! This should definitely be a test. I need to wait for ultralight into implement a argument to set the data storage size for my idea. It is a little hard to test this till they add that.

KolbyML:
Here is an issue I made, https://github.com/ethereumjs/ultralight/issues/478. after that is done I believe I will be able to test this.

PR #886, opened 2023-09-02
Typo in stale issue message

njgheorghita:
### What was wrong?
Minor typo from #863 

### How was it fixed?
Fixed typo

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Add entry to the [release notes](https://github.com/ethereum/trin/blob/master/newsfragments/README.md) (may forgo for trivial changes)
- [x] Clean up commit history

PR #982, opened 2023-10-11
"Hydrate" the transactions returned by eth_getBlockByHash 

carver:
### What was wrong?

Fix #886

### How was it fixed?

Aggregate and convert information into the transaction representation. I didn't do a `From` implementation, because the jsonrpc representation of a transaction includes block-level information, so we need to compile information from the block and the transaction together first, before converting it into a reth-rpc style `Transaction`.
### To-Do

- [ ] Clean up commit history


KolbyML:
https://github.com/ethereum/trin/pull/886 isn't a issue. Maybe the wrong issue was linked? This PR has been open for a year are there any plans to get it past the finish line? If not maybe we should close it and make an issue of the problem (maybe there already is an issue but it isn't linked not sure)

KolbyML:
Closing as this PR has been stale for over 2 years and I sent a message asking if it would be update 5 months ago and there was no response

PR #985, opened 2023-10-12
chore: Automate release on new git tag push

ogenev:
### What was wrong?
Manually adding binaries for a release is a PITA.

### How was it fixed?
Added a git action workflow which should create a new draft release with binaries included on a new git tag push

### To-Do

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,219 @@
+# This workflow is borrowed from Lighthouse: https://github.com/sigp/lighthouse/blob/693886b94176faa4cb450f024696cb69cda2fe58/.github/workflows/release.yml
+
+name: release
+
+on:
+  push:
+    tags:
+      - v*
+
+env:
+  REPO_NAME: ${{ github.repository_owner }}/trin
+  IMAGE_NAME: ${{ github.repository_owner }}/trin
+  CARGO_TERM_COLOR: always
+
+jobs:
+  extract-version:
+    name: extract version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract version
+        run: echo "VERSION=$(echo ${GITHUB_REF#refs/tags/})" >> $GITHUB_OUTPUT
+        id: extract_version
+    outputs:
+      VERSION: ${{ steps.extract_version.outputs.VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,219 @@
...
+          mkdir artifacts
+          mv target/${{ matrix.arch }}/${{ matrix.profile }}/trin.exe ./artifacts
+
+      # ==============================
+      #       Signing
+      # ==============================
+
+      - name: Configure GPG and create artifacts
+        env:
+          GPG_SIGNING_KEY: ${{ secrets.GPG_SIGNING_KEY }}
```
ogenev, line 105: @carver do you have admin access to the repo, do you know if we have added `GPG_SIGNING_KEY` and `GPG_PASSPHRASE` as secrets?

ogenev:
I'm unsure how to test this other than merging it and pushing some test tags, so I'm okay with optimistically merging this and iterating if issues pop up.

.github/workflows/release.yml
carver: Looks like no secrets set:
image

carver: I'm not sure what the standards are for this, but I don't think the answer is for someone to put their private gpg key & passphrase into the repo secrets. I guess we would just generate a new one. If you want to do that and send it to me over keybase, I can save it into those secrets under Actions.

ogenev: >  If you want to do that and send it to me over keybase, I can save it into those secrets under Actions.

Sent the key over keybase.

carver (approved):
Frankly, this is kind of a rubber-stamp review. I scanned it but am not so familiar with github actions, so I probably missed it if something is buggy. :)

Testing in prod!

carver:
I'll merge so that we can test on the upcoming deployment.

.github/workflows/release.yml
carver: Saved to the repo 👍🏻

carver:
Hm, no draft release created when I pushed a new tag 🤷🏻‍♂️ -- I'll just build it manually following the previous release flow

PR #986, opened 2023-10-12
Fix broken bridge mode loop with epoch selector

njgheorghita:
### What was wrong?
Small bug (due to bad variable naming) resulted in a bug when using the epoch selector in bridge mode.  This should fix the problem we're seeing in our backfill bridge nodes that loop over to start at epoch #1 after completing their initial epoch.

Also fixed another small bug to correctly set `end_block` to `latest_block` in instances when it would be a value greater than the head of the chain.

Also fixed another small bug that skips all gossip if user sets start block in current epoch.

### How was it fixed?
🐛  🥾 

### To-Do

### DIFF
--- a/portal-bridge/src/bridge.rs
+++ b/portal-bridge/src/bridge.rs
@@ -141,10 +141,10 @@ impl Bridge {
         };
         let (mut start_block, mut end_block, mut epoch_index) = match mode_type {
             // end block will be same for an epoch in single & backfill modes
-            ModeType::Epoch(block) => (
-                block * EPOCH_SIZE,
-                ((block + 1) * EPOCH_SIZE),
-                block / EPOCH_SIZE,
+            ModeType::Epoch(epoch_number) => (
+                epoch_number * EPOCH_SIZE,
+                ((epoch_number + 1) * EPOCH_SIZE),
+                epoch_number,
             ),
             ModeType::Block(block) => {
                 let epoch_index = block / EPOCH_SIZE;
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

portal-bridge/src/bridge.rs
```diff
@@ -141,10 +141,10 @@ impl Bridge {
         };
         let (mut start_block, mut end_block, mut epoch_index) = match mode_type {
             // end block will be same for an epoch in single & backfill modes
-            ModeType::Epoch(block) => (
-                block * EPOCH_SIZE,
-                ((block + 1) * EPOCH_SIZE),
-                block / EPOCH_SIZE,
+            ModeType::Epoch(epoch_number) => (
+                epoch_number * EPOCH_SIZE,
```
ogenev, line 145: Not sure if I got this but If the `epoch_number` is `1`, and the `epoch_size` is `N` blocks, the start block of the epoch shouldn't be` (epoch_number - 1) * epoch_size` and the end block - `(epoch_number * epoch_size)` ?

njgheorghita: I think this is just a difference in indexing. The first epoch (eg blocks `0-8192`) is epoch 0. In the same way that we start the block count at 0 (eg block 0) it makes sense to me to also begin the epoch count at 0 rather than 1. 

To further back this up. The formula used in this code to get the epoch index is ...
```rust
let epoch_index = block / EPOCH_SIZE;
```
Following this logic, we do start at epoch `0`...
```
block 0 / EPOCH_SIZE = 0
block 1 / EPOCH_SIZE = 0
block 8192 / EPOCH _SIZE = 1
block 8193 / EPOCH_SIZE = 1
```

njgheorghita: @ogenev Going ahead with the merge on this one so I can run these backfill nodes over the weekend, but feel free to ping me if you still have any questions

Issue #764, opened 2023-06-06
Websocket port should be configurable via cli arguments


njgheorghita:
https://github.com/ethereum/trin/pull/757#discussion_r1219642958

ogenev:
HTTP and WS must run on a separate port.  Current defaults are 8545 for HTTP and 8546 for WS.

njgheorghita:
Following up ^ - it appears as though all of this logic should also be removed... https://github.com/ethereum/trin/blob/master/rpc/src/rpc_server.rs#L385-L425

tesol2y090:
Hi, i would like to take this issue but i seem like in cli arguments, there are --ws-port already.

morph-dev:
I think you are correct and this was already implemented in #992 , but issue was never closed.

PR #992, opened 2023-10-14
fix: prevent WS from using same port causing bridge failing to start on multiple nodes

KolbyML:
### What was wrong?
When I run bridge with multiple nodes I get an error on the lines of ``WS socker address already bound`` When bridge makes X connections it WS binds to 8456 everytime.
### How was it fixed?
By matching the modules (which is currently required to use the same port) and using the same port. If we don't want to do this we will just remove with_ws() and fully support WebSockets in the future


njgheorghita:
https://github.com/ethereum/trin/issues/764

KolbyML:
> #764

Ok I am just removing WS from starting with the HTTP rpc in that cause, as properly supporting WS isn't high priority right now

njgheorghita:
@kolbyml If I remember correctly, @ogenev  and I discussed this option before and decided that it was better to run both ws & http simultaneously... I'll see if I can dig up that conversation

njgheorghita:
Looks like it was probably to follow geth's behavior ... https://github.com/ethereum/trin/pull/757#discussion_r1218707885 unless I'm missing some other part of the conversation

njgheorghita:
I think the optimal solution in this case is to add a cli flag that allows for custom ws ports

KolbyML:
> I'd suggest we add a `--ws` flag which will...
> - enable the ws server if the flag is provided
> - take a port, `--ws 8546`

taken from https://github.com/ethereum/trin/pull/757#discussion_r1218707885

> I discussed this option before and decided that it was better to run both ws & http simultaneously

Sounds like the solution was to optionally run WS simulatneously if a port was provided. 

Which is different then 
> I think the optimal solution in this case is to add a cli flag that allows for custom ws ports, which we could use when running bridge with > 1 node to avoid collisions

^ which requires we run WS regardless of if we pass a port param or not

Optionally enabling WS is very easy to do and cleaner IMO and follows true to your original comment in #757. If both are mandatory well it isn't much more work I see WS in general as very low priority. When debugging I don't want to comment out master branch code because it is panicing my test runs. So preventing those panics is very high priority for me. So if we follow the optional WS suggested in the link you provided I am fine adding that as it is quick. If this PR is blocked by "manitory to run both ws & http" it will probably be on the back burner for me to get done.

So I am for the opt in WS, but against manitory WS because when I run trin instances I don't want to specify a port for something I don't use as that would be repetitive/tedius to write out ``--ws-transport-port=3321``. Why waste CPU cycles if the user doesn't use WS? I guess I am repeating myself a lot in this the main point being I don't want to run WS on my machine unless I specify I want it. It is a pain if I have to specify it when I only want to use http also this seems like an error prone edge case cause users will probably have to then google "WS port in use" then configure it if they are running 2 Trin instances which isn't intuitive.

Also shouldn't we able to to use IPC and WS

njgheorghita:
> Also why run WS if the user isn't using it.

Yup, I agree. And if I remember correctly, I also pushed for this argument. Unfortunately, I can't remember the justification, all I remember is @ogenev pushed for running WS by default. Maybe, I'm misremembering, but would like a final confirmation from Ognyan here that we're good to go ahead with making WS optional (and will only be enabled via an cli flag)

KolbyML:
> > Also why run WS if the user isn't using it.
> 
> Yup, I agree. And if I remember correctly, I also pushed for this argument. Unfortunately, I can't remember the justification, all I remember is @ogenev pushed for running WS by default. Maybe, I'm misremembering, but would like a final confirmation from Ognyan here that we're good to go ahead with making WS optional (and will only be enabled via an cli flag)

Okie dokie sounds like a plan

ogenev:
I'm ok with removing the WS by default but IMO this is not solving the real issue mentioned in the PR description and I don't think it justifies disabling WS.

If a bridge runs multiple nodes and all of those nodes bind to the same WS address, it is either a bridge configuration issue or Trin issue that doesn't allow using a custom WS address.

KolbyML:
> I'm ok with removing the WS by default but IMO this is not solving the real issue mentioned in the PR description and I don't think it justifies disabling WS.
> 
> If a bridge runs multiple nodes and all of those nodes bind to the same WS address, it is either a bridge configuration issue or Trin issue that doesn't allow using a custom WS address, or both :)

So are you ok with us only enabling ws if somebody specifies the port on running trin so ``trin -- --ws-port=1234``.
I.e the same way geth handles this situation?

njgheorghita:
In this situation, I feel like following what geth does is the easiest way to proceed, unless somebody can produce a strong use case for another setup...

[`geth --ws --ws.port 3334 `](https://geth.ethereum.org/docs/interacting-with-geth/rpc)

Using just the `--ws` flag, will enable WS connections using the default port, and users can otherwise specify their own custom port. Without the `--ws` flag, no ws is enabled. And without the `--ws` flag, providing a `--ws-port` flag should cause an error. This configuration should solve our bridge conflict. 

We could do something like .... `--ws-port default` / `--ws-port 3334` (aka remove the `--ws` flag) ... to help remove some redundancy, but imo I'd prefer to simply follow geth's behavior in this scenario.

KolbyML:
Ok PR is ready to review with the geth styled ws pattern nick posted above

njgheorghita (commented):
Looks good! My only question is w/ regards to [this comment](https://github.com/ethereum/trin/issues/764#issuecomment-1579339414) by @ogenev 

> HTTP and WS must run on a separate port. Current defaults are 8545 for HTTP and 8546 for WS.

Why do we require separate ports? From what I can tell, it seems technically possible to run both ws & http processes on the same port. Now, it's a different question about whether this is something we want to support or not? 

Right now, it looks like we support running ws & http on the same port, but I'm leaning towards requiring separate ports...

morph-dev:
I have few comments / suggestions

- might be worth documenting these changes in `book/src/developers/contributing/build_instructions/linux.md` (or somewhere else in the book)
- If I'm not mistaken, there is now difference in how we configure http and websocket (http using `web3_http_address` (address and port) and websocket just configuring port)
- there is also implied assumption that `web3_transport` needs to be set to http in order for websocket to work, we don't have any checks whether things are properly set (e.g. someone might set websocket without changing `web3_transport`)

If this PR is blocking some other work, you can merge it like this and I can look into making these changes.

njgheorghita:
Following up from the discussion in today's sync....

Let's shelve the `HTTP and WS must run on a separate port.` question for now. We currently support opening `HTTP` & `WS` connections on the same port.

Regarding @morph-dev 's suggestions...
- Agreed that we should make sure all relevant `book` updates are made to reflect the changes in this pr
- `there is now difference in how we configure http and websocket`.. personally, I'm ok with this. However, if you want to propose an updated cli arg design, I'm open to exploring how we can consolidate formats. Though this is out of scope for this pr
- This pr should handle the case where a user tries to use `ws` alongside `ipc`. Imo, simply throwing an error is sufficient for the meantime

KolbyML:
> * Agreed that we should make sure all relevant `book` updates are made to reflect the changes in this pr
> * This pr should handle the case where a user tries to use `ws` alongside `ipc`. Imo, simply throwing an error is sufficient for the meantime

@njgheorghita I resolved these 2 concerns. PR is ready for another review

morph-dev:
:shipit:

ethportal-api/src/types/cli.rs
```diff
@@ -154,6 +154,16 @@ pub struct TrinConfig {
     )]
     pub disable_poke: bool,
 
+    #[arg(long = "ws", help = "Used to enable WebSocket rpc.")]
+    pub ws: bool,
+
+    #[arg(
+        long = "ws-port", 
+        help = "The WebSocket port to listen on.", 
+        default_value_t = DEFAULT_WEB3_WS_PORT
```
njgheorghita, line 163: I think we should add [`requires`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires) the `ws` argument here

ethportal-api/src/types/cli.rs
```diff
@@ -207,13 +219,25 @@ impl TrinConfig {
...
+                            ErrorKind::ArgumentConflict,
+                            "Must not supply an ipc path when using http protocol for json-rpc",
+                        ))
+                    }
+                }
+            }
+            Web3TransportType::IPC => {
+                match config.web3_http_address.as_str() {
+                    DEFAULT_WEB3_HTTP_ADDRESS => {}
+                    p => return Err(Error::raw(ErrorKind::ArgumentConflict,format!("Must not supply an http address when using ipc protocol for json-rpc (received: {p})"))),
```
njgheorghita, line 236: I know it's not your change, but can you update this single letter variable to something more descriptive

njgheorghita (approved):
:ship:

PR #994, opened 2023-10-15
chore: add more detailed logs to handle offer code

KolbyML:
### What was wrong?
The errors which hard to grep through and organize. Also we used unwrap_or which was potentially dropping important information
### How was it fixed?
By changing the unwrap_or to a match case.

I added more detailed trace logs including enr, cids, which content keys a connection was using. I added content_keys to all the respective error logs so you can just grep by x key and it will all be there.

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -31,7 +31,7 @@ use tokio::{
     sync::mpsc::{self, UnboundedReceiver, UnboundedSender},
     task::JoinHandle,
 };
-use tracing::{debug, error, info, trace, warn};
+use tracing::{debug, enabled, error, info, trace, warn, Level};
 use utp_rs::{conn::ConnectionConfig, socket::UtpSocket, stream::UtpStream};
 
 use crate::events::EventEnvelope;
@@ -1232,6 +1232,12 @@ where
             OverlayRequestError::AcceptError("unable to find ENR for NodeId".to_string())
         })?;
         let enr = crate::discovery::UtpEnr(node_addr.enr);
+
+        // avoid memory allocation if trace isn't enabled
+        let mut enr_str = String::new();
+        if enabled!(Level::TRACE) {
+            enr_str = enr.0.to_base64();
+        }
         let cid = self.utp_socket.cid(enr, false);
         let cid_send = cid.send;
         let validator = Arc::clone(&self.validator);
@@ -1240,6 +1246,23 @@ where
         let command_tx = self.command_tx.clone();
         let utp = Arc::clone(&self.utp_socket);
         let metrics = self.metrics.clone();
+
+        let mut content_keys_string = vec![];
+        if enabled!(Level::ERROR) {
+            for content_key in content_keys.iter() {
+                content_keys_string.push(content_key.to_hex());
+            }
+        }
+
+        trace!(
+            protocol = %self.protocol,
+            request.source = %source,
+            cid.send = cid.send,
+            cid.recv = cid.recv,
+            enr = enr_str,
+            request.content_keys = ?content_keys_string,
+            "Content keys handled by offer",
+        );
 
         tokio::spawn(async move {
             // Wait for an incoming connection with the given CID. Then, read the data from the uTP
@@ -1252,7 +1275,7 @@ where
                         UtpDirectionLabel::Inbound,
                         UtpOutcomeLabel::FailedConnection,
                     );
-                    error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), "unable to accept uTP stream");
+                    error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), content_keys = ?content_keys_string, "unable to accept uTP stream");
                     return;
                 }
             };
@@ -1261,7 +1284,7 @@ where
             if let Err(err) = stream.read_to_eof(&mut data).await {
                 metrics
                     .report_utp_outcome(UtpDirectionLabel::Inbound, UtpOutcomeLabel::FailedDataTx);
-                error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), "error reading data from uTP stream, while handling an Offer request.");
+                error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), content_keys = ?content_keys_string, "error reading data from uTP stream, while handling an Offer request.");
                 return;
             }
 
@@ -1279,7 +1302,7 @@ where
             )
             .await
             {
-                error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), "unable to process uTP payload");
+                error!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), content_keys = ?content_keys_string, "unable to process uTP payload");
             }
         });
 
@@ -1592,6 +1615,18 @@ where
         content_keys: Vec<TContentKey>,
         payload: Vec<u8>,
     ) -> anyhow::Result<()> {
+        let mut content_keys_string = vec![];
+        if enabled!(Level::ERROR) {
+            for content_key in content_keys.iter() {
+                content_keys_string.push(content_key.to_hex());
+            }
+        }
+
+        trace!(
+            request.content_keys = ?content_keys_string,
+            "Processing accepted uTP payload",
+        );
+
         let content_values = portal_wire::decode_content_payload(payload)?;
 
         // Accepted content keys len should match content value len
@@ -1668,12 +1703,30 @@ where
                 })
             })
             .collect();
-        let validated_content: Vec<(TContentKey, Vec<u8>)> = join_all(handles)
-            .await
-            .into_iter()
-            // Whether the spawn fails or the content fails validation, we don't want it:
-            .filter_map(|content| content.unwrap_or(None))
-            .collect();
+        // Whether the spawn fails or the content fails validation, we don't want it:
+        let mut validated_content: Vec<(TContentKey, Vec<u8>)> = vec![];
+        for (index, content) in join_all(handles).await.into_iter().enumerate() {
+            match content {
+                Ok(content) => {
+                    if let Some(content) = content {
+                        validated_content.push(content)
+                    }
+                }
+                Err(err) => {
+                    // Extract the error from the thread handle
+                    let err = err.into_panic();
+                    let err = if let Some(err) = err.downcast_ref::<&'static str>() {
+                        err.to_string()
+                    } else if let Some(err) = err.downcast_ref::<String>() {
+                        err.clone()
+                    } else {
+                        format!("?{:?}", err)
+                    };
+                    error!(err, content_key = ?content_keys_string[index], "Process uTP payload tokio task failed to spawn:");
+                }
+            }
+        }
+
         // Propagate all validated content, whether or not it was stored.
         let validated_ids: Vec<String> = validated_content
             .iter()
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -1703,29 +1703,25 @@ where
                 })
             })
             .collect();
-        // Whether the spawn fails or the content fails validation, we don't want it:
-        let mut validated_content: Vec<(TContentKey, Vec<u8>)> = vec![];
-        for (index, content) in join_all(handles).await.into_iter().enumerate() {
-            match content {
-                Ok(content) => {
-                    if let Some(content) = content {
-                        validated_content.push(content)
-                    }
-                }
-                Err(err) => {
-                    // Extract the error from the thread handle
-                    let err = err.into_panic();
-                    let err = if let Some(err) = err.downcast_ref::<&'static str>() {
-                        err.to_string()
-                    } else if let Some(err) = err.downcast_ref::<String>() {
-                        err.clone()
-                    } else {
-                        format!("?{:?}", err)
-                    };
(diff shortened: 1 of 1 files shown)


KolbyML:
> It would be nice to see an example of what the new logs look like in the terminal

For most error logs we are just adding ``content_keys=["0xfsf7sd8f7sd78g87sd87afa"]``


```nim
        trace!(
            protocol = %self.protocol,
            request.source = %source,
            cid.send = cid.send,
            cid.recv = cid.recv,
            enr = enr_str,
            request.content_keys = ?content_keys_string,
            "Content keys handled by offer",
        );
```

This one will be like ``Content keys handled by offer protocol=history request.source=0x0131.1232 cid.send=123 ci.recv=124 enr="enr:efhasdiuhfsaiufidhfuasiuf-enr-stuff" request.content_keys=["content_key1", "content_key2"]``

```nim
 error!(err, content_key = ?content_keys_string[index], "Process uTP payload tokio task failed:");
 ```
 
 ^ this one is to handle if a thread panics or fails to start, before we weren't logging this and ignoring it the log should look like ``Process uTP payload tokio task failed err="some error xyz" content_key="contentkey1"``


@njgheorghita I hope that helps

File: portalnet/src/overlay_service.rs [hunk 1]
```diff
@@ -1232,6 +1232,12 @@ where
             OverlayRequestError::AcceptError("unable to find ENR for NodeId".to_string())
         })?;
         let enr = crate::discovery::UtpEnr(node_addr.enr);
+
+        // avoid memory allocation if trace isn't enabled
```
njgheorghita, line 1236: I don't love this pattern of adding complexity (if statements) to the core code to handle an edge case of a trace-level log. 
- Our code (especially `overlay_service.rs`) is already quite complex / convoluted. 
- I understand your personal vendetta against unnecessary memory allocation, but imo, this project is not nearly close to the point where we're concerned with optimization. These kinds of improvements are best saved for later down the road when we are optimizing, and can do so in a structured way. ( profiling / identifying / correcting bottlenecks, etc). Imo, at the current stage of this project we should prioritize code simplicity > performance.

portalnet/src/overlay_service.rs
```diff
@@ -1671,9 +1706,23 @@ where
         let validated_content: Vec<(TContentKey, Vec<u8>)> = join_all(handles)
             .await
             .into_iter()
+            .enumerate()
             // Whether the spawn fails or the content fails validation, we don't want it:
-            .filter_map(|content| content.unwrap_or(None))
+            .filter_map(|(index, content)| content.unwrap_or_else(|err| {
+                // Extract the error from the thread handle
+                let err = err.into_panic();
+                let err = if let Some(err) = err.downcast_ref::<&'static str>() {
```
njgheorghita, line 1714: Would a `match` statement be cleaner here?

portalnet/src/overlay_service.rs
```diff
@@ -1671,9 +1706,23 @@ where
...
+                    err.to_string()
+                } else if let Some(err) = err.downcast_ref::<String>() {
+                    err.clone()
+                } else {
+                    format!("?{:?}", err)
```
njgheorghita, line 1719: Is there an extra `?` in the message?

also... `format!("{err:?}")`

portalnet/src/overlay_service.rs
```diff
@@ -1232,6 +1232,12 @@ where
             OverlayRequestError::AcceptError("unable to find ENR for NodeId".to_string())
         })?;
         let enr = crate::discovery::UtpEnr(node_addr.enr);
+
+        // avoid memory allocation if trace isn't enabled
+        let mut enr_str = String::new();
```
njgheorghita, line 1237: I've been taught that this pattern of instantiating a mutable object and then mutating it later should be avoided when possible. Granted that was in python, which had fewer guarantees around type safety.

portalnet/src/overlay_service.rs
```diff
@@ -1671,9 +1706,23 @@ where
         let validated_content: Vec<(TContentKey, Vec<u8>)> = join_all(handles)
             .await
             .into_iter()
+            .enumerate()
             // Whether the spawn fails or the content fails validation, we don't want it:
```
njgheorghita, line 1710: Looks like this comment needs updating?

portalnet/src/overlay_service.rs
```diff
@@ -1671,9 +1706,23 @@ where
...
-            .filter_map(|content| content.unwrap_or(None))
+            .filter_map(|(index, content)| content.unwrap_or_else(|err| {
```
njgheorghita, line 1711: Looks like we already log when content fails validation & the reason on line 1664? What exactly are the errors you want to explicitly distinguish between here?

KolbyML: That isn't catching if the thread fails to be created. Or if a panic happens

KolbyML: Panics are just being ignored if they happen in this thread and aren't handled

portalnet/src/overlay_service.rs
KolbyML: I am not looking to optimize vaste amounts of memory usage. I am trying to be insightful of when I write new code that is is memory effiencent. If I can avoid tech debt when it comes to memory I think that is important. Little changes add up over time.

I don't want to do premature optimization either, but if we can achieve less memory usage without doing additional work now, it will greatly reduce the problem down the line. I am trying to be proactive about it so the debt doesn't pile up for now reason.

> Our code (especially overlay_service.rs) is already quite complex / convoluted.

If that is the case we need to refactor the code. If we are running offer 100 times avoiding calculating the base64 ENR is worth it in itself without the memory saves.

portalnet/src/overlay_service.rs
KolbyML: Nice catch

portalnet/src/overlay_service.rs
KolbyML: Ok I will just do an if else then which will remove the mut

portalnet/src/overlay_service.rs
KolbyML: Yep

portalnet/src/overlay_service.rs
KolbyML: The unwrap_or was unwarping the result of if the thread successfully ran or if it failed. So we weren't handling the errors it was returning.

portalnet/src/overlay_service.rs
carver: Yeah, there's some gray area here. In this case, I agree that if the only reason we're doing a calculation is to generate the base64 encoding for the log, then it is nice to skip it if the logging is disabled (especially for something as low as a trace that's rarely enabled, and a part of the code that's this regularly called). Skipping log work, especially in hot areas of the code, is a pattern that we ended up having to use pretty often in py-evm to improve performance.

njgheorghita: Basically, I'm surprised that we would even need to consider this (IMO) antipattern to get the best use out of a very popular rust library. Digging into it slightly - based on [this comment by a maintainer](https://www.reddit.com/r/rust/comments/x9nypb/comment/inusmd5/?utm_source=share&utm_medium=web2x&context=3) it appears to me like if the log level is not enabled, then we don't have to worry about the fields being evaluated. 

> a runtime-disabled span or event compiles down to two instructions: a branch and load of an atomic boolean, which is morally close to nothing on most processors.

But maybe I'm misunderstanding the comment? @KolbyML have you confirmed that we have a runtime performance hit? If it's the case that the runtime hit is "morally close to nothing" then I'd argue that it's not worth the extra lines of code.

portalnet/src/overlay_service.rs
njgheorghita: Yeah I think you're right here, I was misunderstanding what was going on

portalnet/src/overlay_service.rs
KolbyML: > Basically, I'm surprised that we would even need to consider this (IMO) antipattern to get the best use out of a very popular rust library. Digging into it slightly - based on [this comment by a maintainer](https://www.reddit.com/r/rust/comments/x9nypb/comment/inusmd5/?utm_source=share&utm_medium=web2x&context=3) it appears to me like if the log level is not enabled, then we don't have to worry about the fields being evaluated.
> 
> > a runtime-disabled span or event compiles down to two instructions: a branch and load of an atomic boolean, which is morally close to nothing on most processors.


I don't understand how what the maintainer said is relevant, I am not complaining about adding ``trace!()`` or ``error!()`` as a problem and my code doesn't signify that. I didn't want to allocate memory for a string and run ``enr.to_base64()`` which encodes to the rlp byte array then encodes to a hex string for a trace which only people debugging will have to run. Even a 0.04% proformance hit will add up over time if you keep making it. Little changes add up a lot over a big codebase.

> But maybe I'm misunderstanding the comment? @KolbyML have you confirmed that we have a runtime performance hit? If it's the case that the runtime hit is "morally close to nothing" then I'd argue that it's not worth the extra lines of code.

I made some changes so it is less intrusive
```rust
        // Generate a connection ID for the uTP connection if there is data we would like to
        // accept.
        let node_addr = self.discovery.cached_node_addr(source).ok_or_else(|| {
            OverlayRequestError::AcceptError("unable to find ENR for NodeId".to_string())
        })?;
        let enr_str = if enabled!(Level::TRACE) {
            node_addr.enr.to_base64()
        } else {
            String::with_capacity(0)
        };
        let enr = crate::discovery::UtpEnr(node_addr.enr);
        let cid = self.utp_socket.cid(enr, false);
        let cid_send = cid.send;
        let validator = Arc::clone(&self.validator);
        let store = Arc::clone(&self.store);
        let kbuckets = Arc::clone(&self.kbuckets);
        let command_tx = self.command_tx.clone();
        let utp = Arc::clone(&self.utp_socket);
        let metrics = self.metrics.clone();
```

^ I was able to make it look nicer which I think is the best solution

``let enr_str = node_addr.enr.to_base64();``

But I will just push this to the branch ^

portalnet/src/overlay_service.rs
```diff
@@ -1241,6 +1242,23 @@ where
         let utp = Arc::clone(&self.utp_socket);
         let metrics = self.metrics.clone();
 
+        let mut content_keys_string = vec![];
+        if enabled!(Level::ERROR) {
```
njgheorghita, line 1246: Are there really any situations when any user would not be running `ERROR` level logs?

KolbyML: I wasn't 100 percent sure if there isn't i will remove the if statements. I seen errors were enabled by default though

njgheorghita: I think it's safe to assume users will always be running at the bare minimum `ERROR` level logs.

KolbyML: Ok I will update the PE in 10 minutes with that change

portalnet/src/overlay_service.rs
```diff
@@ -1231,6 +1231,7 @@ where
         let node_addr = self.discovery.cached_node_addr(source).ok_or_else(|| {
             OverlayRequestError::AcceptError("unable to find ENR for NodeId".to_string())
         })?;
+        let enr_str = node_addr.enr.to_base64();
```
njgheorghita, line 1234: If you're not going with the `if enabled` check here, then why even create this variable? Looks like it's only used in the `tracing!` log, so let's just move it there? 
```rust
        trace!(
            protocol = %self.protocol,
            request.source = %source,
            cid.send = cid.send,
            cid.recv = cid.recv,
            enr = node_addr.enr.to_base64(),
            request.content_keys = ?content_keys_string,
            "Content keys handled by offer",
        );
```

That being said, I don't understand why you removed the `if` check? I don't want to strongarm you here if you feel strongly about leaving the check in. I appreciate you running the profiling. It's very helpful to have data when making these decisions. Could we do something like 

```rust 
    if enabled!(Level::TRACE) {
        trace!(
            protocol = %self.protocol,
            request.source = %source,
            cid.send = cid.send,
            cid.recv = cid.recv,
            enr = node_addr.enr.to_base64(),
            request.content_keys = ?content_keys_string,
            "Content keys handled by offer",
        );
    }
 ```

Though my interpretation of what the maintainer said is that this is essentially what is happening, so there's no need for the `if enabled!(Level::TRACE) {}` check

njgheorghita:
In conclusion, thanks for putting up with me through all this. I'll just leave some final thoughts here for you to ponder...

I do think optimization is important. But, personally I don't think we should care about it yet. Code simplicity & feature completeness is my priority. But, that's just my opinion. If performance is important to you, I would encourage you to take a more concentrated look at the codebase and isolate these types of inefficiences (via profiling) and correct them all together. And more importantly, add standards to our style guide that establish a standard in our codebase for how all contributors should handle these situations. Otherwise, our codebase ends up in an inconsistent state, where we have pieces of code over here that prioritize optimization, but other pieces of code that do not. That makes it confusing to read & to contribute to. Imo, optimization is better understood & implemented as a concentrated task across the whole codebase, rather than in a piecemeal fashion, a little bit here, a little bit there, as we develop.

njgheorghita (approved):
I'm happy with this, after you've removed the `if` checks for `ERROR` level logs, because that is the bare minimum log level we should be developing for.

KolbyML:
Me and Nick joined a call and talked about the PR

Issue #882, opened 2023-09-01
Switch from `eth2_ssz` to `ethereum_ssz` crate.


ogenev:
In our codebase, we depend on `ethereum_types v.0.12.1` which works fine with the current `eth2_ssz` library that we use.
However, newer `ethereum_types` versions are not compatible with this ssz library - for example, `ethereum_type::H256` type does not implement `eth2_ssz::Encode/Decode` traits.

Looking at Lighthouse, it seems that they use a newer [ssz library ](https://github.com/sigp/lighthouse/blob/dfcb3363c757671eb19d5f8e519b4b94ac74677a/consensus/types/Cargo.toml#L28-L30 )with similar API as the current and it is supported from the latest `ethereum_types v.0.14.1`. The [eth2_ssz](https://crates.io/crates/eth2_ssz) crate is also outdated compared to [ethereum_ssz](https://crates.io/crates/ethereum_ssz).

Investigate how feasible would be to switch to the newer [ethereum_ssz](https://crates.io/crates/ethereum_ssz) library, so we can use the latest version of [ethereum_types](https://crates.io/crates/ethereum-types).

PR #997, opened 2023-10-17
chore: Switch from `eth2_ssz` to `ethereum_ssz` crate

morph-dev:
### What was wrong?
We depend on `ethereum_types v.0.12.1` and we can't upgrade it to the newer versions, because they are no longer compatible with `eth2_ssz` library that we use. Solution is to switch to `ethereum_ssz` and upgrade dependencies together.

See #882 for details

### How was it fixed?
Some crates were replaced, while some were upgraded to newer version:

`eth2_ssz` v.0.4.0 -> `ethereum_ssz` v.0.5.3
`eth2_ssz_derive` v.0.3.0 -> `ethereum_ssz_derive` v.0.5.3
`eth2_ssz_types` v.0.2.1 -> `ssz_types` v.0.5.4

`ethereum-types`: v.0.12.1 -> v.0.14.1
`tree_hash`: v.0.4.0 -> v.0.5.2
`tree_hash_derive`: v.0.4.0 -> v.0.5.2
`keccak-hash`: v.0.8.0 -> v.0.10.0

Bonus: Some files were no longer needed: `ethportal-api/src/types/wrapped/h160.rs`

### To-Do

- [x] Wait until `eth_trie` is upgraded and released, and switch to new version as well. PR created: [carver/eth-trie.rs/#32](https://github.com/carver/eth-trie.rs/pull/32).


ethportal-api/src/types/execution/block_body.rs
```diff
@@ -138,6 +138,7 @@ impl BlockBody {
         }
 
         trie.root_hash()
+            .map(|root| H256::from_slice(root.as_bytes()))
```
carver, line 141: Both are ethereum_types::H256, can't we skip the mapping?

carver (approved):
Looks good! Yay for upgrading dependencies and removing code.

carver:
eth-trie v0.4 is published

ethportal-api/src/types/execution/block_body.rs
morph-dev: As discussed, it was needed before `eth_trie` was upgraded and new version published. No longer needed.

morph-dev:
Upgraded eth_trie to v0.4. Ready to be merged.

KolbyML (approved):
:shipit:

PR #998, opened 2023-10-17
fix: push changes requested in (#994) which weren't pushed

KolbyML:
### What was wrong?
I agreed with nick to make a change before merging and I thought I committed it waited 15 minutes and merged it assuming everything worked
### How was it fixed?
By committing the changes I forgot to :scream_cat:


njgheorghita (approved):
:shipit:

PR #1001, opened 2023-10-18
Update trace-recursive-find-content response to match spec changes

### LINKED ISSUE #987: shortcomings of current portal_historyTraceRecursiveFindContent implementation
I've been trying to implement `portal_historyTraceRecursiveFindContent` for fluffy and here's what I've found on my journey:
- fields mostly `snake_case` while everywhere else in the RPC they are `camelCase`
- `port` and `ip` in `metadata` are redundant since this info contained in the ENR
- `distance` in the `metadata` is obviously clipped, cause it can be up to 256 bits and JSON integer can only be 53 bits. It should probably be encoded into hex string.
- `distance_log2` is redundant, we already have distance
- `started_at` is never used in the glados.

Here's spec update that addresses this improvements: https://github.com/ethereum/portal-network-specs/pull/236
And here's glados PR that supports this change: https://github.com/ethereum/glados/pull/177

njgheorghita:
### What was wrong?
Changes to the `portal_*TraceRecursiveFindContent` endpoint in the [specs](https://github.com/ethereum/portal-network-specs/pull/236) needed implementation.

Fixes #987 

### How was it fixed?
- Implemented changes to endpoint
- Changed `ContentId` type from `NodeId` -> `[u8; 32]`

### To-Do

### DIFF
--- a/ethportal-api/src/types/query_trace.rs
+++ b/ethportal-api/src/types/query_trace.rs
@@ -1,36 +1,44 @@
-use super::enr::Enr;
-use super::node_id::NodeId;
 use std::collections::HashMap;
 use std::time::SystemTime;
 
+use ethereum_types::H256;
 use serde::{Deserialize, Serialize};
 
 use super::distance::{Metric, XorMetric};
+use super::enr::Enr;
+use super::node_id::NodeId;
 
-type ContentId = NodeId;
+type ContentId = [u8; 32];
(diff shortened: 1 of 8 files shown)


KolbyML (approved):
:shipit: nothing looks off

mrferris (approved):
🫡

Issue #487, opened 2022-12-03
No standard discovery port


perama-v:
## Problem description

Looking at ports used on the portal network I see two issues:
### 1. Clash with beacon chain
The port used by Trin clashes with the discovery port on beacon chain clients.
This means that by default there will be clashes if users have a beacon client running. This may become annoying if beacon light clients are coupled somehow with portal clients.

### 2. No standard portal network port
Users following generic setup instructions may have to configure port-related settings. If there is
a standard port, these instructions can be cross-client.

## Possible resolution
Decide on a discovery port to converge on for the Portal Network.

Example default ports.
| Client | Network | Port |
| - | - | - |
| Trin, Ultralight | portal | 9000 | 
| Fluffy | portal | ~~?~~ 9009 | 
| Teku, Nimbus, Lodestar, Lighthouse | consensus | 9000 |
| Prysm | consensus | 13000/12000 |
| Geth, Nethermind, Erigon, Besu | execution | 30303 |

ogenev:
Fluffy is on 9009. I agree that we should move away from using 9000 as the default.

pipermerriam:
relevant: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers

perama-v:
Nice, no clashes for 9009 on the wiki

PR #1002, opened 2023-10-19
chore: Set default discovery port to 9009

morph-dev:
### What was wrong?

The current default port (9000) is used by some beacon chain clients.
Per agreement on #487, I'm making 9009 the default one.

fix: #487

### How was it fixed?

Made 9009 the default and updated all references that I could find.

### To-Do


book/src/developers/contributing/build_instructions/linux.md
```diff
@@ -132,7 +132,7 @@ Open the file:
 ```sh
 $ sudo nano /etc/systemd/system/trin.service.d/override.conf
 ```
-Paste the following, replace the Infura ID with your own.
```
morph-dev, line 135: Side note: I noticed this line but it doesn't fit the rest of the text, so I assumed it's leftover from some previous versions.

ethportal-peertest/src/lib.rs
```diff
@@ -54,7 +54,7 @@ async fn launch_node(trin_config: TrinConfig) -> anyhow::Result<PeertestNode> {
 }
 
 fn generate_trin_config(id: u16, bootnode_enr: Option<&Enr>) -> TrinConfig {
-    let discovery_port: u16 = 9000 + id;
+    let discovery_port: u16 = 9009 + id;
```
njgheorghita, line 57: I would replace this value (`9009`) with an import of `DEFAULT_DISCOVERY_PORT`, it'll make life easier down the road if we ever need to change the default again. And similarly in `main.rs` / `overlay_service.rs`

portal-bridge/src/main.rs
```diff
@@ -24,7 +24,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
     let mut http_addresses = vec![];
     for (i, key) in private_keys.into_iter().enumerate() {
         let web3_http_port = 8545 + i;
```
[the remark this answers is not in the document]
morph-dev: done

ethportal-peertest/src/lib.rs
morph-dev: done

ethportal-api/src/types/cli.rs
```diff
@@ -78,7 +78,7 @@ pub struct TrinConfig {
     pub web3_ipc_path: PathBuf,
 
     #[arg(
-        default_value = DEFAULT_DISCOVERY_PORT,
+        default_value_t = DEFAULT_DISCOVERY_PORT,
```
njgheorghita, line 81: This looks like a typo? I don't see `default_value_t` listed under the available [magic methods](https://docs.rs/structopt/latest/structopt/#magical-methods)

morph-dev: I'm confused. Why did you reference `structopt` crate? I think we are using [`clap`](https://docs.rs/clap/latest/clap/). Am I missing something? (sorry, I started learning rust about a month ago)

Here is the documentation about attributes: https://docs.rs/clap/latest/clap/_derive/index.html#arg-attributes, but it might not be very clear what is the difference.

I found this [explanation](https://github.com/clap-rs/clap/discussions/3640) and in short, the `default_value` works on pre-parsed values (strings), while `default_value_t` works on post-parsed values (actual data type)

njgheorghita: Ahh, sorry my bad, I had it backwards. Yup, you're right we're using `clap` not `structopt` now.

book/src/developers/contributing/build_instructions/linux.md
```diff
@@ -71,12 +71,12 @@ ENR entry for IP blank. Some users report better connections over VPN."
...
-on the same machine.
+or other software on the same machine.
 
-`--discovery-port <port>`. If an Ethereum consensus client is already running, it may be using
-the default port 9000.
+`--discovery-port <port>`. The default port is 9009. Pick something else if in use.
 
-`--web3-http-address <ip_address>:<port>`. If an Ethereum execution client is already running, it may be using the default port 8545. The localhost IP address (127.0.0.1) is recommended here.
+`--web3-http-address <ip_address>:<port>`. If an Ethereum execution client is already running,
+it may be using the default port 8545. The localhost IP address (127.0.0.1) is recommended here.
```
njgheorghita, line 79: Hmmm... so if execution client's default port is `8545` (eg [geth](https://geth.ethereum.org/docs/interacting-with-geth/rpc)). Then doesn't it make sense to also change our default listening port so there isn't a conflict there? For basically the same reason we're changing the discovery port (though this change would definitely be out of scope for this pr, but if we think the change is worthwhile, let's open an issue to track it)

book/src/developers/contributing/build_instructions/linux.md
```diff
@@ -187,7 +187,7 @@ variables during testing.
 ```sh
 $ cargo test --workspace
 $ cargo run -- --discovery-port 9009 \
-    --web3-http-address 127.0.0.1:8547 \
+    --web3-http-address 127.0.0.1:8545 \
```
njgheorghita, line 190: This is the default listening address here. IIUC in this example, we're trying to show how to start a second trin instance alongside an already-running trin instance, so this would cause an address conflict. And since we're using `8546` as websocket, I'd say we can just leave this port as `8547`

portalnet/src/overlay_service.rs
```diff
@@ -2753,9 +2753,11 @@ mod tests {
         storage::{DistanceFunction, MemoryContentStore},
         utils::db::setup_temp_dir,
     };
-    use ethportal_api::types::content_key::overlay::IdentityContentKey;
     use ethportal_api::types::distance::XorMetric;
     use ethportal_api::types::enr::generate_random_remote_enr;
+    use ethportal_api::types::{
```
njgheorghita, line 2758: *nitpick* Our general import statement ordering standard is poorly defined, mostly due to there not being any solid linting options available on the `stable` rust channel. But, in this case I would prefer either...
```rust
use ethportal_api::types::{
    cli::DEFAULT_DISCOVERY_PORT,
    content_key::overlay::IdentityContentKey,
    distance::XorMetric,
    enr::generate_random_remote_enr,
};
```
or...
```rust
use ethportal_api::types::cli::DEFAULT_DISCOVERY_PORT;
use ethportal_api::types::content_key::overlay::IdentityContentKey;
use ethportal_api::types::distance::XorMetric;
use ethportal_api::types::enr::generate_random_remote_enr;
```

njgheorghita (approved):
:shipit:

pipermerriam:
Just to make things difficult... are we sure nobody else is using 9009 as their default discovery port?

book/src/developers/contributing/build_instructions/linux.md
morph-dev: As discussed, I added clarifications.

Note for others: All situation regarding ports is a bit confusing, especially considering we can't change the websocket port at the moment. Will create an issue to follow up on this.

portalnet/src/overlay_service.rs
morph-dev: Done. I tend not to look at imports and consider anything that automatic formatting / tooling does as good enough (especially if clippy doesn't complain).

morph-dev:
> Just to make things difficult... are we sure nobody else is using 9009 as their default discovery port?

Based on #487, this shouldn't be an issue. However, initial analysis was done some time ago...

I'm not even sure how I would look into this and find if there are some conflicts at the moment.

Issue #766, opened 2023-06-06
Run trin with POKE disabled


njgheorghita:
In order for glados to be able to collect an un-biased view of the network, we need to enable a mode for running trin with the POKE mechanism disabled. This mode should be toggled on/off via a cli flag.

carver:
Closed by the above linked PR

PR #1003, opened 2023-10-19
Adds disable poke CLI flag to trin

mrferris:
### What was wrong?

#766 

Disabling the poke mechanism is useful for keeping a clear, unaffected view of the network while doing network analysis eg in glados. 

Kind of like how when you're fishing, throwing a fish back in will disturb the water. 🐟 

### How was it fixed?

Adds `--disable-poke` flag to Trin, and passes it into the `OverlayService` where poking is handled.

Tested manually.

### To-Do


KolbyML (approved):
:rocket:

PR #1004, opened 2023-10-19
Get rid of error message for uninitialized networks

mrferris:
### What was wrong?
If you run trin with beacon ie `trin --networks history,beacon` and then later run with only history ie `trin --networks history`, you will still receive TALKREQ packets from nodes on the portal beacon network. These will cause `ERROR portalnet::events: Beacon event handler not initialized!` which is a little over the top for something that's not really a problem.

### How was it fixed?

Added some extra context and turned it down to `debug` level

### To-Do

### DIFF
--- a/portalnet/src/events.rs
+++ b/portalnet/src/events.rs
@@ -3,7 +3,7 @@ use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use discv5::TalkRequest;
 use tokio::sync::mpsc;
-use tracing::{error, warn, debug};
+use tracing::{debug, error, warn};
 
 use super::types::messages::ProtocolId;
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

mrferris:
I'm not sure why the changes made in the second commit were suddenly required, did our `cargo fmt` rules get updated?

KolbyML:
> I'm not sure why the changes made in the second commit were suddenly required, did our `cargo fmt` rules get updated?

I don't believe so haven't they always been ``cargo fmt --all -- --check``

mrferris:
Oh, turns out making the log lines longer made cargo's idea of proper formatting for a lot of other lines change. So all is well.

PR #1005, opened 2023-10-20
[WIP] Initial support for trace gossip jsonrpc endpoint

njgheorghita:
### To-Do


portalnet/src/gossip.rs
```diff
@@ -20,7 +22,15 @@ use ethportal_api::utils::bytes::hex_encode;
 use ethportal_api::OverlayContentKey;
 use ethportal_api::RawContentKey;
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize, Default)]
+pub struct GossipResult {
+    pub offered: Vec<NodeId>,
+    pub accepted: Vec<NodeId>,
+    pub transferred: Vec<NodeId>,
+}
+
```
KolbyML, line 31: Having the ENR string here would probably be a lot more helpful then NodeIds. If it is just nodeIds then we don't know there ip address and such. Also if it is an enr we can message them directly for futher debugging

KolbyML (approved):
:shipit:

portalnet/src/gossip.rs
```diff
@@ -135,6 +114,151 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
...
+        // mark peer as accepted if something went wrong during utp transfer
+        match result_rx.recv().await {
+            Some(result) => {
+                if result {
+                    gossip_result.transferred.push(enr);
+                } else {
+                    gossip_result.accepted.push(enr);
+                }
+            }
+            None => gossip_result.accepted.push(enr),
```
carver, line 214: I think a lot of this is simplified be having every peer that is offered the content get added to the list, etc.

carver (approved):
LGTM, with some thoughts

portalnet/src/gossip.rs
```diff
@@ -135,6 +114,151 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
...
+        }
+    }
+    gossip_result
+}
+
+/// Filter all nodes from overlay routing table where XOR_distance(content_id, nodeId) < node radius
+fn calculate_interested_enrs<TContentKey: OverlayContentKey>(
+    content_key: &TContentKey,
+    all_nodes: &Vec<&kbucket::Node<NodeId, Node>>,
```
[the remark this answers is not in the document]
njgheorghita: Agreed, and I did some searching but wasn't able to find similar code anywhere else in the repo. Though, the majority of this logic was extracted from the above `propagate_gossip_cross_thread()` function in this module, which (pre #975 ) used to live in `overlay_service.rs`... which might be what you're thinking about in this case?

PR #1007, opened 2023-10-24
Use `finalized slot` in `LightClientFinalityUpdate` content key and update beacon bridge to serve the data.

ogenev:
### What was wrong?
Beacon network [specs](https://github.com/ethereum/portal-network-specs/blob/master/beacon-chain/beacon-network.md#lightclientfinalityupdate) require a finalized slot to be included in `LightClientFinalityUpdate`'s content key.

### How was it fixed?
- Use the finalized instead of the signature slot in the `LightClientFinalityUpdate` content key.
- Update beacon bridge to serve `LightClientFinalityUpdate` content only if the finalized slot changes.
- Improve some beacon bridge logs. 

### To-Do

- [x] Record a screencast for Devconnect with the current beacon sync prototype before merging/deploying this PR.

### DIFF
--- a/portal-bridge/src/beacon_bridge.rs
+++ b/portal-bridge/src/beacon_bridge.rs
@@ -81,13 +81,15 @@ impl BeaconBridge {
     ///  Get and serve the latest beacon data.
     async fn launch_latest(&self) {
         // Current sync committee period known by the bridge
-        let (mut current_period, mut finalized_block_root) = Self::serve_latest(
-            self.api.clone(),
-            self.portal_clients.clone(),
-            0,
-            String::new(),
-        )
-        .await;
+        let (mut current_period, mut finalized_block_root, mut finalized_slot) =
+            Self::serve_latest(
+                self.api.clone(),
+                self.portal_clients.clone(),
+                0,
+                String::new(),
+                0,
+            )
+            .await;
 
         // Sleep until next update becomes available. This sets up the interval to update as soon as
         // the following slot becomes available.
@@ -107,11 +109,12 @@ impl BeaconBridge {
             let consensus_api = self.api.clone();
             let portal_clients = self.portal_clients.clone();
 
-            (current_period, finalized_block_root) = Self::serve_latest(
+            (current_period, finalized_block_root, finalized_slot) = Self::serve_latest(
                 consensus_api,
                 portal_clients,
                 current_period,
                 finalized_block_root,
+                finalized_slot,
             )
             .await;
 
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
:shipit:

portal-bridge/src/beacon_bridge.rs
```diff
@@ -298,21 +311,41 @@ impl BeaconBridge {
...
+    ) -> anyhow::Result<u64> {
         let data = api.get_lc_finality_update().await?;
         let update: Value = serde_json::from_str(&data)?;
         let update: LightClientFinalityUpdateCapella =
             serde_json::from_value(update["data"].clone())?;
         info!(
-            "Got lc finality update for slot {:?}",
-            update.attested_header.beacon.slot
+            finalized_slot = %update.finalized_header.beacon.slot,
+            "Got LightClientFinalityUpdate",
```
carver, line 322: If I read this log without context, I would have thought it meant: "Received LightClientFinalityUpdate from peer" rather than "Generated LightClientFinalityUpdate for bridging broadcast". Maybe `Generated LightClientFinalityUpdate`?

portal-bridge/src/beacon_bridge.rs
```diff
@@ -298,21 +311,41 @@ impl BeaconBridge {
...
+        let new_finalized_slot = update.finalized_header.beacon.slot;
+
+        match new_finalized_slot.cmp(&finalized_slot) {
+            Ordering::Equal => {
+                // We already gossiped the latest finality updated with the same finalized slot, no need to serve it again.
+                return Ok(finalized_slot);
+            }
+            Ordering::Less => {
+                // if we panic here, it is a bug
+                bail!("System clock went backwards: Expected finalized slot is less than known finalized slot");
```
carver, line 334: The clock isn't necessarily the culprit here, right? I'm not sure we can even really speculated what when wrong. Maybe as simple as stating what should never happen, which is that a finalized block should never get rewound and "unfinalized".

```suggestion
                bail!("Consensus client must not unwind finalized block, but: New finalized slot is less than known finalized slot");
```

portal-bridge/src/beacon_bridge.rs
```diff
@@ -298,21 +311,41 @@ impl BeaconBridge {
...
         let content_key = BeaconContentKey::LightClientFinalityUpdate(
-            LightClientFinalityUpdateKey::new(update.signature_slot),
+            LightClientFinalityUpdateKey::new(new_finalized_slot),
         );
         let content_value = BeaconContentValue::LightClientFinalityUpdate(update.into());
 
-        Self::gossip_beacon_content(portal_clients, content_key, content_value).await
+        Self::gossip_beacon_content(portal_clients, content_key, content_value)
+            .await
+            .map(|()| new_finalized_slot)
```
carver, line 348: Hm, now that I look at this pattern (of mapping that drops the old value) with fresh eyes it seems a little off. (I think I did it before).

This looks like a cleaner approach:
```suggestion
        Self::gossip_beacon_content(portal_clients, content_key, content_value).await?
        Ok(new_finalized_slot)
```

carver (approved):
LGTM!

PR #1008, opened 2023-10-25
Update local metrics dashboard

### LINKED ISSUE #871: Add trin-native metrics to grafana dashboard
Right now [trin-bench](https://trin-bench.ethdevops.io/) does *NOT* display our trin-native metrics for individual nodes. They should.

njgheorghita:
### What was wrong?
Local dashboard template needed updating to include newly available metrics.

Fixes #871 

### How was it fixed?
- updated metrics template
- fixed broken asset path
- added some notes to help avoid gotchas while testing updated template
- added dashboard template from `trin-bench`'s `Trin Metrics` dashboard for backup

### To-Do

### DIFF
--- a/ethportal-api/src/dashboard/grafana.rs
+++ b/ethportal-api/src/dashboard/grafana.rs
@@ -5,7 +5,7 @@ use std::fs;
 use ureq;
 
 pub const DASHBOARD_TEMPLATES: &[&str] =
-    &["../etc/grafana/dashboards/collected-metrics-dashboard.json.template"];
+    &["./etc/grafana/dashboards/collected-metrics-dashboard.json.template"];
 
 pub struct GrafanaAPI {
     basic_auth_string: String,
@@ -45,7 +45,6 @@ impl GrafanaAPI {
                 "url": url,
                 "basicAuth": false,
             }))?;
-
         let response: DatasourceCreationResponse = datasource_creation_response.into_json()?;
         Ok(response.datasource.uid)
     }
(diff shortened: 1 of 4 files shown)


ethportal-api/src/dashboard/grafana.rs
```diff
@@ -60,6 +59,13 @@ impl GrafanaAPI {
         let dashboard_json: serde_json::Value = serde_json::from_str(&filled_in_template[..])?;
 
         let dashboard_api_url = format!("{}/{}", self.address, "api/dashboards/db/");
+        // If this is failing with a 409 error, it means that there is likely a problem with the
+        // template, make sure that ...
+        // - All "uid" fields inside a prometheus datasource object contain `"uid": "{prometheus_id}"`
+        // - The root level `"id"` field is set to `null` (`"id": null`)
+        // - The root level `"title"` field is set to "Trin Metrics" (`"title": "Trin Metrics"`)
+        // - The root level `"uid"` field is set to "trin-metrics" (`"uid": "trin-metrics"`)
+
```
carver, line 68: Nit, the spacing makes it look like the comment is talking about the format! line. Maybe an empty line above, and no empty line below?

ethportal-api/src/dashboard/grafana.rs
```diff
@@ -5,7 +5,7 @@ use std::fs;
 use ureq;
 
 pub const DASHBOARD_TEMPLATES: &[&str] =
-    &["../etc/grafana/dashboards/collected-metrics-dashboard.json.template"];
+    &["./etc/grafana/dashboards/collected-metrics-dashboard.json.template"];
```
carver, line 8: Do we have to add the trin-bench metrics dashboard here? Seems like the template file was added, but is never created?

carver (approved):

ethportal-api/src/dashboard/grafana.rs
njgheorghita: The `trin-bench` template is really only meant as a backup for the `trin-bench.ethdevops.io` grafana process. Since it has a different configuration from the `collected-metrics-dashboard`, it's not useful to anybody running grafana locally, so I don't think there's any need to create it locally.

PR #1010, opened 2023-10-27
Log detailed gossip results from bridge processes.

njgheorghita:
### What was wrong?
Updated the bridge gossip stats to accurately reflect what was gossiped, using the new trace gossip result type. The idea is to make grepping through the bridge logs a feasible way to figure out whether content was successfully gossiped into the network.

### How was it fixed?
- Added improved metrics and logs to beacon and history bridge processes

### To-Do

### DIFF
--- a/portal-bridge/src/stats.rs
+++ b/portal-bridge/src/stats.rs
@@ -1,0 +1,332 @@
+use std::str::FromStr;
+
+use tracing::{info, trace};
+
+use ethportal_api::jsonrpsee::core::Error;
+use ethportal_api::types::enr::Enr;
+use ethportal_api::types::portal::TraceGossipInfo;
+use ethportal_api::{BeaconContentKey, HistoryContentKey};
+
+// Trait for tracking / reporting gossip stats
+pub trait StatsReporter<TContentKey> {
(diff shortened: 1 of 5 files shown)


portal-bridge/src/stats.rs
```diff
@@ -0,0 +1,329 @@
...
+                        let enr = Enr::from_str(enr).expect("");
+                        if !content_stats.accepted.contains(&enr) {
+                            content_stats.accepted.push(enr);
+                        }
+                    }
+
+                    for enr in info.transferred.iter() {
+                        let enr = Enr::from_str(enr).expect("");
+                        if !content_stats.transferred.contains(&enr) {
+                            content_stats.transferred.push(enr);
```
KolbyML, line 320: These ``expect("")`` should never fail. But currently they are being used as if they were ``unwrap()``. It would be nice to put a message saying ``x is impossible`` instead of an empty string. Just a nitpick so not a big deal if you don't want to, empty expect() just taste wrong idk.

KolbyML (approved):
:shipit: looks good. Having 3 trace logs for each seems like a lot, but I see were it is coming from.

PR #1016, opened 2023-10-31
Update README.md using PossibleHistoryContentValue to fix mismatch types

ltfschoen:
### What was wrong?

mismatched type

### How was it fixed?

use type that compiles

### DIFF
--- a/ethportal-api/README.md
+++ b/ethportal-api/README.md
@@ -41,7 +41,7 @@ async fn main() {
     assert!(result);
 
     // Call portal_historyLocalContent endpoint and deserialize to `HistoryContentValue::BlockHeaderWithProof` type
-    let result: HistoryContentValue = client.local_content(content_key).await.unwrap();
+    let result: PossibleHistoryContentValue = client.local_content(content_key).await.unwrap();
     assert_eq!(result, content_item);
 }
 ```


ethportal-api/README.md
```diff
@@ -41,7 +41,7 @@ async fn main() {
     assert!(result);
 
     // Call portal_historyLocalContent endpoint and deserialize to `HistoryContentValue::BlockHeaderWithProof` type
-    let result: HistoryContentValue = client.local_content(content_key).await.unwrap();
+    let result: PossibleHistoryContentValue = client.local_content(content_key).await.unwrap();
     assert_eq!(result, content_item);
```
ltfschoen, line 45:
```suggestion
    assert_eq!(result, PossibleHistoryContentValue::ContentPresent(content_item.clone()));
```

njgheorghita: Yup, this is also needed, though it doesn't seem like the clone is necessary

njgheorghita (commented):
Thanks! Looks like you'll also need to import the `PossibleHistoryContentValue` type at the top.

KolbyML:
Closing this PR as it is no longer relivent and the current doc is now correct

Issue #1015, opened 2023-10-31
test_unpopulated_offer incorrectly checks if the stored content value in the target DB matches that offered by the target instead of checking if the bootnode's DB matches that offered by the target like in test_populated_offer


ltfschoen:
In `test_populated_offer` it calls `wait_for_history_content(&peertest.bootnode.ipc_client, content_key).await`  [here](https://github.com/ethereum/trin/blob/master/ethportal-peertest/src/scenarios/offer_accept.rs#L65)

But `test_unpopulated_offer` it calls `let response = wait_for_history_content(target, content_key).await;` [here](https://github.com/ethereum/trin/blob/master/ethportal-peertest/src/scenarios/offer_accept.rs#L37).
Shouldn't `test_unpopulated_offer` also provide `&peertest.bootnode.ipc_client` in the first argument to `wait_for_history_content` instead of passing `target`?

PR #1017, opened 2023-10-31
Relates to #1015: Update offer_accept.rs to check stored content in peer DB matches of target

ltfschoen:
### What was wrong?

See #1015

### How was it fixed?

Pass peer similar to how it was done in `test_unpopulated_offer`


njgheorghita (approved):
Nice catch, thanks! If you can update the commit message to follow the [conventional commit format](https://www.conventionalcommits.org/en/v1.0.0/) I'll make the merge!

KolbyML:
@njgheorghita I think if a 3rd party contributor makes a PR without ``conventional commits`` we should just squash merge there PR and fit the commit title ourselves.

Requiring new people to rebase their commits is a barrier of entry for new people to contribute and waiting like this delays merging the PR's which feels unideal. Asking them to rebase with some custom format doesn't seem as productive as just fixing it ourselves.

PR #1019, opened 2023-11-01
fix(portal-bridge): push `LightClientUpdate` if finalized header is within the current expected period

ogenev:
### What was wrong?
Pushing `LightClientUpdate` data at the beginning of the current sync committee period prevents Trin from syncing when it is using those updates in future periods. Light client specs expect the `LightClientUpdate` data to have finality, i.e. the finalized header slot is within the current expected period.

### How was it fixed?
Update the beacon bridge to push `LightClientUpdate` content only when the finalized header slot is within the current expected period.

### To-Do

### DIFF
--- a/portal-bridge/src/beacon_bridge.rs
+++ b/portal-bridge/src/beacon_bridge.rs
@@ -36,6 +36,9 @@ use ethportal_api::{
     BeaconContentKey, BeaconContentValue, LightClientBootstrapKey, LightClientUpdatesByRangeKey,
 };
 
+/// The number of slots in a sync committee period.
+const SLOTS_PER_PERIOD: u64 = 32 * 256;
+
 pub struct BeaconBridge {
     pub api: ConsensusApi,
     mode: BridgeMode,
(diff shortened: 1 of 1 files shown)


portal-bridge/src/beacon_bridge.rs
```diff
@@ -292,6 +292,17 @@ impl BeaconBridge {
         let data = api.get_lc_updates(expected_current_period, 1).await?;
         let update: Value = serde_json::from_str(&data)?;
         let update: LightClientUpdateCapella = serde_json::from_value(update[0]["data"].clone())?;
+
+        // We don't serve a `LightClientUpdate` if its finalized header slot is not within the expected current period.
+        if update.finalized_header.beacon.slot / (32 * 256) != expected_current_period {
```
njgheorghita, line 297: *nitpick* I'd prefer to see `(32 * 256)` to be a const w/ a descriptive name. Seems like we can also set this value (`update.finalized_header.beacon.slot / (32 * 256)`) to a variable (eg `actual_current_period`) and then re-use it in the `warn!` log rather than recalculating it.

njgheorghita (approved):
🚢

Issue #22, opened 2021-05-26
Automatically forward port through NAT, using UPnP


carver:
Look at how the following project does it: https://github.com/pzmarzly/portforwarder-rs/blob/master/src/bin/pf.rs

Probably just use [`igd`](https://docs.rs/igd/0.12.0/igd/) directly.

---

On second thought: this is probably unnecessary, because the outbound requests will punch a hole in the NAT, and the PONGs provide the externally observed IP & port.

pipermerriam:
Is this a mistake?  Automated UPnP seems valuable for NAT'd users.

PR #1022, opened 2023-11-01
feat: add upnp support

qizhou:
### What was wrong?

Currently, Trin cannot support a private node running behind a firewall/gateway.  Inspired by libp2p, this PR adds UPnP support, which is the default NAT used in libp2p.

### How was it fixed?

UPnP protocol will ask the gateway to map a private IP/port to an external IP/port and return the external address to Trin.  By encoding the external address in ENR, Trin will allow any node to access a Trin node running behind the gateway.

Test:
- Run a Trin node behind a gateway with an empty bootnode.  The node will print the ENR with an external IP/port mapped by UPnP.
- Run another Trin node using the bootnode's ENR, which connects the bootnode.
- Call portal_historyGossip (using test vector) from one node and check that another node receives the data.

### To-Do


KolbyML:
The guy at nimbus who was pushing for 4444s told me why aren't you guys just using upnp, so I think this is a good first direction. 

As Qi said on discord in general channel we will probably end up supporting
- UPnP (default)
- rendezvous
- relay

In that order of priority.

This will also majorly support our effort to having a large network.

Thank you Qi for looking into this I believe this PR solves or at least work towards https://github.com/ethereum/trin/issues/22 Carver can correct me if I am wrong on that

I guess I am saying I support this change

portalnet/src/socket.rs
```diff
@@ -35,3 +45,85 @@ pub fn stun_for_external(local_socket_addr: &SocketAddr) -> Option<SocketAddr> {
...
+        Ok(external_ip) => external_ip,
+        Err(ref err) => {
+            warn!(error = %err, "Error getting external IP");
+            return None;
+        }
+    };
+
+    // Find a local ip to map.
+    // UPnP cannot use an unspecified IP (e.g., 0.0.0.0) to map and thus we need to give a specific IP connected to the gateway.
+    // TODO: Detect the local IP connected to the gateway.
```
carver, line 88: Yeah, as I recall, this has given us trouble before, that if you randomly select one of the IPs, you'll get the wrong one (for example, if a user is on ethernet or wifi). If you can add this into the current PR, that would be excellent.

qizhou: Thanks for the comment.  I added the feature to find a local IP connected to the gateway and use the IP as the external IP for UPnP NAT.

ethportal-api/src/types/cli.rs
```diff
@@ -170,6 +177,7 @@ impl Default for TrinConfig {
             bootnodes: Bootnodes::Default,
             external_addr: None,
             no_stun: false,
+            no_upnp: false,
```
njgheorghita, line 180: It seems like this means we're **enabling** upnp by default? Is that the desired behavior?

portalnet/src/discovery.rs
```diff
@@ -93,6 +93,35 @@ impl Discovery {
             }
         };
 
+        let (enr_address, enr_port) = if socket::is_local_addr(enr_address) || portal_config.no_upnp
```
njgheorghita, line 96: Personally, I'm not a fan of nested statements, especially when the nesting gets to be more than 2/3 levels deep. It just makes code a bit more difficult to read / reason about what's going on. Here's a suggestion for how to get rid of some of this nesting...

line 83:
```rust
        let (mut enr_address, mut enr_port) = if portal_config.no_stun {
```

```rust
        if socket::is_local_addr(enr_address) || portal_config.no_upnp {
            if let Some(socket) = socket::upnp_for_external(listen_all_ips) {
                if let Some(known_external) = enr_address {
                    if known_external != socket.ip() {
                        if portal_config.external_addr.is_some() {
                            return Err(format!(
                                "Mismatched known external address {} vs UPnP found address {}, consider disable --external-addr or --no-upnp",
                                known_external,
                                socket.ip()
                            ));
                        } else {
                            // STUN address is different from UPnP address. Use UPnP
                            warn!("overriding STUN address with known UPnP external address");
                        }
                    }
                };
                enr_address = Some(socket.ip());
                enr_port = socket.port();
            }
        };
```

portalnet/src/socket.rs
```diff
@@ -35,3 +45,86 @@ pub fn stun_for_external(local_socket_addr: &SocketAddr) -> Option<SocketAddr> {
...
+                    if *ip == external_ip {
+                        return true;
+                    }
+                }
+            }
+
+            false
+        }
+    }
+}
```
njgheorghita, line 66:
```suggestion
pub fn is_local_addr(external_ip: Option<IpAddr>) -> bool {
    if let Some(external_ip) = external_ip {
        let network_interfaces = local_ip_address::list_afinet_netifas();

        if let Ok(network_interfaces) = network_interfaces {
            for (_, ip) in network_interfaces.iter() {
                if *ip == external_ip {
                    return true;
                }
            }
        }
    }
    false
}
```

ethportal-api/src/types/cli.rs
qizhou: The behavior is borrowed from geth's default behavior, where UPnP is used by default (and there is no flag to disable it).  Feel free to let me know if you have better thoughts.

portalnet/src/discovery.rs
qizhou: Done.

portalnet/src/socket.rs
```diff
@@ -1,10 +1,20 @@
...
+    thread,
+};
 use tracing::{debug, info, warn};
 
 // This stun server is part of the testnet infrastructure.
 // If you are unable to connect, please create an issue.
 const STUN_SERVER: &str = "159.223.0.83:3478";
 
+/// The duration in seconds of an external port mapping by UPnP.
+const UPNP_MAPPING_DURATION: u32 = 3600;
```
ogenev, line 13: Why are we mapping UPnP for only 60 minutes? I think we want to use an infinite value here, which in this case should be 0.

ethportal-api/src/types/cli.rs
pipermerriam: I believe enable by default is the desired behavior

portalnet/src/socket.rs
pipermerriam: I am slightly in the camp of having a limit unless we're absolutely sure that these will expire after a period of disuse.  We don't want to be opening ports up on people's home networks indefinitely.  Seems worth the effort to implement periodic renewal of the mapping to make sure we aren't being bad citizens on user machines.

qizhou: > I am slightly in the camp of having a limit unless we're absolutely sure that these will expire after a period of disuse. We don't want to be opening ports up on people's home networks indefinitely. Seems worth the effort to implement periodic renewal of the mapping to make sure we aren't being bad citizens on user machines.

Agree.  The current design will initially map the port for 60 minutes, and then periodically (every 30 minutes) renew the port for another 60 minutes.  That said if Trin is down for no matter what reasons (crash or gracefully closed), the port reserved by Trin will be automatically released in < 60 minutes.

portalnet/src/socket.rs
```diff
@@ -35,3 +45,82 @@ pub fn stun_for_external(local_socket_addr: &SocketAddr) -> Option<SocketAddr> {
...
+
+    match gateway.add_port(
+        igd_next::PortMappingProtocol::UDP,
+        listen_addr.port(),
+        local_addr,
+        UPNP_MAPPING_DURATION,
+        "new_port",
+    ) {
+        Ok(()) => {
+            thread::spawn(move || loop {
```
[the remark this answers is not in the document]
qizhou: This thread is used to renew the port periodically so that Trin can own the port as long as it is alive, while the port will be released in < 60 mins if Trin quits in any matter (crash or gracefully shutdown).

portalnet/src/socket.rs
```diff
@@ -35,3 +45,82 @@ pub fn stun_for_external(local_socket_addr: &SocketAddr) -> Option<SocketAddr> {
...
+    info!("Connecting to UPnP gateway to map local address to external address");
+    let gateway = match igd_next::search_gateway(Default::default()) {
+        Ok(gateway) => gateway,
+        Err(ref err) => {
+            warn!(error = %err, "Error finding gateway");
+            return None;
+        }
+    };
+
+    let local_ip = match TcpStream::connect(gateway.addr) {
```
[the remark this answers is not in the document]
qizhou: Actually, the previous version of the PR uses local_ip and local_ip_v6 to find the local IP address as below:

```
local_addr = match local_ip_address::local_ip().or(local_ip_address::local_ipv6())local_ip_address::local_ip().or(local_ip_address::local_ipv6()) ...
```

However, as @carver points out that this may be problematic if multiple interfaces exist.  That is why I switch to a TCP way to detect the local IP.

Note that the UPnP protocol also uses an HTTP/TCP connection to the gateway.  Unfortunately, the UPnP library does not return which local IP address is used to talk to the gateway.  So, without modifying UPnP library, making a TCP connection to the gateway is an easy solution to determine the local IP address

ethportal-api/src/types/cli.rs
```diff
@@ -105,6 +105,13 @@ pub struct TrinConfig {
     )]
     pub no_stun: bool,
 
+    #[arg(
+        long = "no-upnp",
+        group = "external-ips",
+        help = "Do not use UPnP to determine an external port."
+    )]
+    pub no_upnp: bool,
```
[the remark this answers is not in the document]
qizhou: I tried to change it to `allow_upnp`, but looks like clap does not support setting a boolean flag to `false`.  Below is the error message I got

```
$ trin --allow-upnp=false

error: unexpected value 'false' for '--allow-upnp' found; no more were expected

Usage: trin --allow-upnp

For more information, try '--help'.
```

portalnet/src/socket.rs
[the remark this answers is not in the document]
ogenev: Ahh, I see now, it makes sense, Let's use tokio to spawn the thread with tokio sleep:

```suggestion
            tokio::spawn(async move { loop {
                tokio::time::sleep(time::Duration::from_secs(UPNP_MAPPING_TIMEOUT)).await;
```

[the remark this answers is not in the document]
qizhou: A comment is added.

I did some study on `tokio::spawn`, and I believe that it may not be a good way because `gateway.add_port` into the UPnP library is a blocking IO, which may block other async tasks.  I tried to use `tokio::tasks::spawn_blocking`, but I encountered an issue that SIGINT (ctrl + c) cannot exit Trin gracefully.  As a result, I have to revert to the dedicated thread method for this purpose.

Please let me know if you have better ideas.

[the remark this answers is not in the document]
ogenev: Hmm, I didn't know that `gateway.add_port` is a blocking IO. In this case we have two options tokio spawn_blocking or a dedicated thread but because we expect this thread to run forever, I think `std::thread::spawn` is the way to go :+1: .

PR #1029, opened 2023-11-09
Add poll_latest and sample_range scripts

njgheorghita:
### What was wrong?
Added two scripts to help with quick & easy network testing. These are meant to be "backups" to glados, but also will be very useful for running tests on kurtosis.

Later, I'd like to implement support for providers other than infura, but given that `audit_latest` requires a `ws` connection, which is really only "easily" available via infura (eg. pandaops doesn't support this), then I figured pandaops / generic provider support can wait

### How was it fixed?
- `audit_latest` script, which will basically poll how soon "latest" data appears on the network
- `sample_range` script, which samples X number of randomly selected blocks from any given range

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,16 @@
 version = 3
 
 [[package]]
+name = "Inflector"
+version = "0.11.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
+dependencies = [
+ "lazy_static",
+ "regex",
+]
+
+[[package]]
 name = "addr2line"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -57,6 +67,17 @@ dependencies = [
  "cpufeatures 0.2.12",
  "ctr 0.8.0",
  "opaque-debug",
+]
+
+[[package]]
+name = "aes"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+dependencies = [
+ "cfg-if 1.0.0",
+ "cipher 0.4.4",
+ "cpufeatures 0.2.12",
 ]
 
 [[package]]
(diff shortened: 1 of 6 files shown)


KolbyML (approved):
:shipit:

ogenev (commented):
I don't like coupling script dependencies with trin dependencies in main Cargo.toml.
It is also expected to add more scripts in the future (purging db and etc).

Having a `bin` directory inside `src` is also confusing and it is not very clear which file is the trin binary and which one is script without looking into `Cargo.toml` and the files.

I think a better design is to rename the `src` folder to `bin` and then add `trin`  crate inside `bin`, for example see reth: https://github.com/paradigmxyz/reth/tree/main/bin/reth

This will require updating `Cargo.toml` and adding the trin crate as a workspace member.

Тhen we can add any rust script as a separate crate inside `bin/`. This way, Trin and any script will have separate `Cargo.toml` and dependencies.

njgheorghita:
@ogenev

> I don't like coupling script dependencies with trin dependencies in main Cargo.toml.
It is also expected to add more scripts in the future (purging db and etc).

Agreed & agreed

> Having a bin directory inside src is also confusing 

That is the structure that is [officially recommended](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#binaries). It's also the structure we were using when we had the `purge_db.rs` script.

> It is not very clear which file is the trin binary and which one is script 

I don't really agree. `main.rs` is pretty obviously the trin binary, as defined by the cargo standards. 

Looking at reth, `reth/bin/reth` is a fairly large project, a separate workspace, with many contents. If our main trin process was a large project, then I would be inclined to follow the Reth convention. In our case `src/trin.rs` is basically just a simple script. Based on this, I'm not convinced that it's a convention worth adopting, when our use case seems to follow the standard use case defined in the cargo standards more closely.

Basically, I do agree it's not great coupling script dependencies with trin dependencies, but it appears to me like we're following cargo best practices. Seems like this is still a bit of an unsolved issue with [cargo](https://github.com/rust-lang/cargo/issues/1982)?

njgheorghita:
@ogenev looks like there's a newly failing test in `light-client` ... not sure what the cause might be, but i'm getting the same failure locally

ogenev (approved):
:boat:

PR #1030, opened 2023-11-10
feat: add cli option for root private key in bridge

njgheorghita:
### What was wrong?
In our bridge architecture, we spin up multiple bridge nodes, but their private keys are currently randomly generated. We want the option to provide custom private keys so that we can ensure the bridge nodes are equally distributed around the keyspace.

### How was it fixed?
- Added cli option for providing custom private keys to bridge nodes

### To-Do

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -245,9 +245,9 @@ impl TrinConfig {
     }
 }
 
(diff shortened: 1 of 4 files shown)


ogenev (approved):
:boat:

KolbyML (approved):

PR #1032, opened 2023-11-10
feat: add retries to bridge gossip, with rfc network checks

njgheorghita:
### What was wrong?
Bridges only made one gossip attempt, and gave up. This isn't ideal. The bridge gossip strategy has been updated to...
- attempt a gossip
- if no content was successfully **transferred**, make a rfc request
- if content is found on the network, abort the gossip
- if the bridge is unable to find the data on the network, wait 15 seconds and try another gossip / rfc cycle
- repeat for a maximum of 3 times before giving up

Testing locally, I've seen all kinds of cases, which is encouraging. eg.
- bridge successfully transfers data without retries
- bridge unsuccessfully transfers data, but finds it on the network and aborts
- bridge retries gossip multiple times, and makes a successful transfer on a retry
- bridge retries gossip multiple times, but finds it on the network after a retry and aborts
- bridge performs the gossip cycle 3 times w/ no success and aborts

### How was it fixed?
Updated beacon and history bridge with the above functionality.

### To-Do

### DIFF
--- a/portal-bridge/src/gossip.rs
+++ b/portal-bridge/src/gossip.rs
@@ -1,0 +1,146 @@
+use crate::stats::{BeaconSlotStats, HistoryBlockStats, StatsReporter};
+use ethportal_api::jsonrpsee::core::Error;
+use ethportal_api::types::portal::{ContentInfo, TraceGossipInfo};
+use ethportal_api::{
+    BeaconContentKey, BeaconContentValue, BeaconNetworkApiClient, HistoryContentKey,
+    HistoryContentValue, HistoryNetworkApiClient, OverlayContentKey, PossibleBeaconContentValue,
+    PossibleHistoryContentValue,
+};
+use jsonrpsee::http_client::HttpClient;
+use std::sync::{Arc, Mutex};
+use tokio::time::{sleep, Duration};
+use tracing::{debug, warn};
+
+const GOSSIP_RETRY_COUNT: u64 = 3;
+const RETRY_AFTER: Duration = Duration::from_secs(15);
+
+/// Gossip any given content key / value to the history network.
+pub async fn gossip_beacon_content(
+    portal_clients: Arc<Vec<HttpClient>>,
+    content_key: BeaconContentKey,
+    content_value: BeaconContentValue,
+    slot_stats: Arc<Mutex<BeaconSlotStats>>,
+) -> anyhow::Result<()> {
+    let mut results: Vec<Result<(Vec<TraceGossipInfo>, u64), Error>> = vec![];
+    for client in portal_clients.as_ref() {
+        let client = client.clone();
+        let content_key = content_key.clone();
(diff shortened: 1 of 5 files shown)


portal-bridge/src/stats.rs
```diff
@@ -272,55 +272,66 @@ impl StatsReporter<HistoryContentKey> for HistoryBlockStats {
...
+                Ok((traces, retries)) => {
+                    content_stats.retries += retries;
+                    for trace in traces {
+                        for enr in trace.offered.iter() {
+                            let enr = Enr::from_str(enr)
+                                .expect("ENR from trace gossip response to succesfully decode.");
+                            // don't double count an enr if multiple clients offered the same content
+                            // to a peer
+                            if !content_stats.offered.contains(&enr) {
+                                content_stats.offered.push(enr);
```
ogenev, line 317: nit: If we want the uniqueness of values, we can use [HashSet ](https://doc.rust-lang.org/std/collections/struct.HashSet.html)instead of Vec for offered, accepted and transferred.  This way, we can remove the `if` statement to check for existing Enr.

portal-bridge/src/gossip.rs
```diff
@@ -0,0 +1,146 @@
...
+        if let Ok(trace) = result {
+            traces.push(trace.clone());
+            if !trace.transferred.is_empty() {
+                return Ok((traces, retry_count));
+            }
+        }
+        // if not, make rfc request to see if data is available on network
+        let result =
+            BeaconNetworkApiClient::recursive_find_content(&client, content_key.clone()).await;
+        if let Ok(PossibleBeaconContentValue::ContentPresent(_)) = result {
```
ogenev, line 64: Beacon should return `ContentInfo::Content` (same as the history below). They have identical jsonr-rpc implementations:  [beacon](https://github.com/ethereum/trin/blob/092427a4f0babf6a8bc3423de84a89b4f244d0cd/trin-beacon/src/jsonrpc.rs#L125) | [history](https://github.com/ethereum/trin/blob/092427a4f0babf6a8bc3423de84a89b4f244d0cd/trin-history/src/jsonrpc.rs#L126).

ogenev (approved):
:motor_boat:

portal-bridge/src/gossip.rs
njgheorghita: Agreed that the subnetwork returns the same type, but the rpc deserialization of the response are different for the two endpoints: [beacon](https://github.com/ethereum/trin/blob/092427a4f0babf6a8bc3423de84a89b4f244d0cd/rpc/src/beacon_rpc.rs#L146C29-L146C29) | [history](https://github.com/ethereum/trin/blob/092427a4f0babf6a8bc3423de84a89b4f244d0cd/rpc/src/history_rpc.rs#L122C20-L122C31)

PR #1035, opened 2023-11-24
chore: adjust fluffy cli arguments for portal-bridge

kdeme:
### What was wrong?
There has been a breaking change from `--storage-size` to `--storage-capacity`.

### How was it fixed?

Arguments are adjusted.

Also the IP limits have been set (temporarily) to higher defaults.
And the network setting is default now the mainnet "testnet" since a while.

So those args are no longer needed and have been removed.


KolbyML (approved):
:shipit:

PR #1036, opened 2023-11-25
fix: filter out disconnected nodes on Node/(Content Enr) response

KolbyML:
### What was wrong?
during devconnect Mike brought up glados's cartographer was trying to ping 1800 enrs, because trin nodes were returning disconnect nodes.

I don't think we should remove the last resort of using disconnected nodes to initialize a RFC/RFN. Since the last resort will only be used if there are 0 connected nodes in the whole routing table which should only happen on startup.

in short I believe this is the right solution

This can also be in glados audits

image

In this picture the Trin node is returning Enr's of disconnected nodes which haven't run for over a month. The trin node returned these because it sorted by distance without filtering out disconnected first. This PR adds the filter so that should no longer happen.
### How was it fixed?
by filtering out disconnected nodes by status

I applied to filter to both the Nodes response and the Content Enr Response

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2360,11 +2360,15 @@ where
         }
     }
 
-    /// Returns a vector of all the ENRs of nodes currently contained in the routing table.
+    /// Returns a vector of all the ENRs of nodes currently contained in the routing table which are connected.
     fn table_entries_enr(&self) -> Vec<Enr> {
         self.kbuckets
             .write()
             .iter()
+            .filter(|entry| {
+                // Filter out disconnected nodes.
+                entry.status.is_connected()
+            })
             .map(|entry| entry.node.value.enr())
             .collect()
     }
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -2369,6 +2369,19 @@ where
             .collect()
     }
 
+    /// Returns a vector of all the ENRs of nodes currently contained in the routing table which are connected.
+    fn table_entries_enr_connected(&self) -> Vec<Enr> {
```
njgheorghita, line 2373: It seems like we could consolidate this method & `table_entries_enr()` with something like
```rust
    fn table_entries_enr(&self, filter_connected: bool) -> Vec<Enr> {
        self.kbuckets
            .write()
            .iter()
            .filter(|entry| {
                // Filter out disconnected nodes.
                if filter_connected {
                    entry.status.is_connected()
                }
            })
            .map(|entry| entry.node.value.enr())
            .collect()
    }
```
well.... I'm not sure if that's the best way to go about this, but if there's an easy way to reduce some of the duplicated lines here, that would be nice to help keep down the size of `overlay_service.rs`

mrferris (approved):

Issue #1039, opened 2023-11-27
Refactor portal storage to allow custom storage implementations per network


ogenev:
With the introduction of beacon and state networks, we need a way to use custom storage implementation based on the network specifications. 

For example, in beacon, we don't use the concept of data radius, we like to have some temporary data and we want to build a `FindContent` response from a range of values from our database. Also, SQLite is enough for beacon and we don't want to use RocksDB.

In our current architecture, we spawn an `OverlayService` implementation per network with a `PortalStorage` [instance](https://github.com/ethereum/trin/blob/44d1f020a4d852f7c0056dcd4bfd797f81bdefbf/trin-history/src/network.rs#L42).
The problem is that `PortalStorage` is specific to the history network, and we would like to swap this implementation with another for beacon and state networks.

To allow this, the most naive approach is to use the repository pattern and create a trait object that would serve as an API interface for the overlay service.  Then, we can have `HistoryStorage`, `BeaconStorage`, and `StateStorage` storage mechanics that would implement this trait. This will move all storage-specific responsibilities to those objects and will encapsulate the logic per network. This will not require refactoring of the overlay service code, because we will use the same API interface as the current `PortalStorage`. 

Because the storage code will grow and to make this implementation feasible and easily maintained, I think we have to extract the storage code from the `portalnet` crate and create a new workspace crate "storage" where all storage implementations will reside.

Here is a list of tasks that we will need to complete to achieve this architecture change:

- [x] 1. Move  `portalnet::types::messages` to `ethportal-api`. (this will allow us to do 2. because of circular dependencies) #1040 
- [x] 2. Extract `portalnet::metrics`  implementations to a separate workspace create. (It makes sense to have a separate crate for the metrics and this will allow us to extract the storage to a separate crate because of circular dependencies) #1043 
- [x] 3. Remove RocksDB #1058 
- [ ] 4. Extract `portalnet::storage` to a separate `storage` workspace crate. (As mentioned above, this is recommended for maintenance concerns and separation of responsibilities.) https://github.com/ethereum/trin/pull/1078
- [ ] 5. Add `HistoryStorage` and `BeaconStorage` implementations with a common portal storage trait. https://github.com/ethereum/trin/pull/1078

njgheorghita:
> I think we have to extract the storage code from the portalnet crate and create a new workspace crate "storage" where all storage implementations will reside.

[...]

PR #1043, opened 2023-11-28
refactor: create `trin-metrics` workspace crate

ogenev:
### What was wrong?

This is Step 2 of https://github.com/ethereum/trin/issues/1039. Having a separate crate for the metrics allows us to deal with circular dependencies and extract the storage code in a workspace crate.

### How was it fixed?

 1. Create a `trin-metrics` workspace crate.
 2. Моvе `portalnet::metric` module to the new crate.

### To-Do


njgheorghita (approved):
Looks good! My only concern is with the name `metrics` itself.... this is a somewhat generic name, that makes the workspace sound more like an external library than an internal workspace. Something like `trin-metrics` would be better imo but this is just a **nitpick**

PR #1048, opened 2023-11-29
fix: Uses ``rustls-tls`` for ``reqwest`` instead of requiring ``openssl``

mrferris:
### What was wrong?

`/usr/bin/trin: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory` when running a docker image built from trin's `docker/Dockerfile`.

I thought it was interesting that @njgheorghita saw this error but his instance kept running normally.

This was caused by this commit on master https://github.com/ethereum/trin/commit/00cbbd8e73ba6c1e92e4df609862e5401a20d543

Which started using openssl through our reqwest library
### How was it fixed?

By changing ``reqwest = { version = "0.11.13", features = ["json"] }``

to

``reqwest = { version = "0.11.13", default-features = false, features = ["json", "rustls-tls"] }``
``default-features = false`` disables openssl from being used
``features = ["rustls-tls"]`` enables using the rust native tls library

Kolby tested the docker build on portal-hive and it works

### To-Do

### DIFF
--- a/light-client/Cargo.toml
+++ b/light-client/Cargo.toml
@@ -23,7 +23,7 @@ tracing-subscriber = "0.3.15"
 hex = "0.4.3"
 parking_lot = "0.11.2"
 portalnet = { path = "../portalnet" }
-reqwest = { version = "0.11.13", features = ["json"] }
+reqwest = { version = "0.11.13", default-features = false, features = ["json", "rustls-tls"] }
 serde_json = "1.0.85"
 ssz-rs = { git = "https://github.com/ralexstokes/ssz-rs", rev = "d09f55b4f8554491e3431e01af1c32347a8781cd" }
 ssz_types = "0.5.4"


ogenev (commented):
Instead of downgrading Ubuntu, we can download and install the library from the openssl archive:

```sh
FROM ubuntu:22.04

RUN apt update && apt install wget -y
RUN wget http://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2_amd64.deb
RUN dpkg -i libssl1.1_1.1.1f-1ubuntu2_amd64.deb
```

njgheorghita (approved):

docker/Dockerfile.bridge
```diff
@@ -35,7 +35,9 @@ COPY ./rpc ./rpc
 RUN cargo build -p trin -p portal-bridge --release
 
 # final base
-FROM ubuntu:22.04
+FROM ubuntu:20.04
```
njgheorghita, line 38: I think it's also worth a comment mentioning why we're on an old ubuntu in both dockerfiles

KolbyML (changes requested):
I think we should revert these changes and add ``default-features = false, `` to the cargo.toml in trin-beacon/light-client/cargo.toml

This way we no longer rely on openssl for TLS. Our default JSON-RPC already doesn't support TLS, so why should trin-beacon/light-client rpc support it

``reqwest = { version = "0.11.13", features = ["json", "rustls-tls"] }`` this is another option which would keep TLS support, but not use openssl

KolbyML:
Yeah so it seems like our solution should be ``reqwest = { version = "0.11.13", default-features = false, features = ["json", "rustls-tls"] }`` after some investigation. This uses a rust native ssl library instead of openssl, which is a win win

KolbyML (approved):

KolbyML:
whoops didn't mean to switch your title on you xD

ogenev (approved):
Nice :+1:

Issue #1049, opened 2023-11-30
Remove all usage of disconnected nodes from the overlay service. 


njgheorghita:
Trin shouldn't have any understanding / usage of "disconnected" peers outside of the routing table. In the routing table an ongoing process to ping dead peers & resurrect them if they're alive should be the only place in our code that uses disconnected nodes. Currently, we still use disconnected nodes inside our `OverlayProtocol`. This should be removed along with improvements in our routing tables resurrection logic to fix the failing tests. 
For more context see: https://github.com/ethereum/trin/pull/1042

PR #1051, opened 2023-12-01
chore: dont gossip to disconnected nodes if there are no connected nodes available

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/issues/1049
### How was it fixed?
Certain tests where checking if certain functions were working by putting dummy values into our kbuckets. Since we have strict standards for how we handle connected vs disconnected this test failed as it was written before this connection standard was in place.

The solution? set it as connect, this doesn't matter as these tests aren't meant to test the status of the node functionality.


portalnet/src/overlay_service.rs
```diff
@@ -409,8 +409,12 @@ where
             // TODO: Decide default data radius, and define a constant. Or if there is an
             // associated database, then look for a radius value there.
             let node = Node::new(enr, Distance::MAX);
+            let state = match set_connected {
+                true => ConnectionState::Connected,
+                false => ConnectionState::Disconnected,
+            };
```
ogenev, line 415: It is maybe a personal preference, but I like more the following syntax:

```suggestion
            let state = if set_connected {
                    ConnectionState::Connected
                } else {
                    ConnectionState::Disconnected,
                }
            };
```

ogenev (approved):
:boat:

portalnet/src/overlay_service.rs
```diff
@@ -398,7 +398,11 @@ where
         command_tx
     }
 
-    fn add_bootnodes(&mut self, bootnode_enrs: Vec<Enr>) {
+    /// Insert a vector of enr's into the routing table
```
njgheorghita, line 401: `Insert a vector of enrs into the routing table`

portalnet/src/overlay_service.rs
```diff
@@ -398,7 +398,11 @@ where
...
+    /// set_connected: should only be true for tests, false for production code
+    /// Tests what use this function are testing if adding to queue's work not if our connection code
```
njgheorghita, line 403: `Tests that use this function are testing if adding to queues works, not if our connection code works.`

PR #1054, opened 2023-12-01
fix: add deployment step to confirm that docker images can be run

mrferris:
### What was wrong?
We had an issue where the docker image tagged for testnet couldn't be run, it was missing a library. 

### How was it fixed?
Adds a step to the CI that attempts to run the docker image (with the `-h` flag for immediate but successful exit).

Adds a step to release steps to check that the master CI was successful for what we're about to deploy. This both applies to the docker image, as well as to the testing iteself, which I realized was also failing for the latest commit on master after having deployed a broken docker image.

### To-Do

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,9 @@ jobs:
           no_output_timeout: 30m
           command: docker build -f ./docker/Dockerfile -t $IMAGE_NAME:latest --build-arg GIT_HASH=$CIRCLE_SHA1 .
       - run:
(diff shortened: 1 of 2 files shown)


KolbyML:
@mike.ferris <https://github.com/ethereum/trin/pull/1054>

Are you sure -h does an immediate exit if successful? <https://docs.docker.com/engine/reference/commandline/run/> says it is for specifying host name

KolbyML (approved):
Not sure how helpful this will be but if it does what it says it does it could save us

PR #1058, opened 2023-12-04
Remove RocksDB

ogenev:
### What was wrong?
Remove RocksDB and switch to SQLite for storing content values.

### How was it fixed?
- Remove RocksDB code and library.
- Add `content_value` column to SQLite main db table and use this column to store and lookup data.
- Add a "network" column to the main db table
- Remove `purge_db` binary.
- Remove state network `TrieDB` because it is based on RocksDB.

Tested this PR locally with `portal-hive`.

### To-Do


.circleci/config.yml
```diff
@@ -38,7 +38,7 @@ commands:
           name: Install libssl-dev for openssl-sys
           command: sudo apt install -y libssl-dev
       - run:
-          name: Install libclang for rocksdb
+          name: Install libclang
```
KolbyML, line 41: Do we need clang anymore? maybe we can remove this step now

njgheorghita, line 41: Can't we just remove these installs entirely if they're no longer needed for `rocksdb`

Cargo.toml
```diff
@@ -27,7 +27,6 @@ prometheus_exporter = "0.8.4"
 rand = "0.8.4"
 reth-ipc = { tag = "v0.1.0-alpha.10", git = "https://github.com/paradigmxyz/reth.git"}
 rlp = "0.5.0"
-rocksdb = "0.21.0"
```
njgheorghita, line 30: :heart:

book/src/developers/contributing/build_instructions/windows.md
```diff
@@ -9,12 +9,6 @@ If you don't already have Rust install it
 $ winget install Rustlang.Rustup
 ```
 
-Install clang/llvm as it is required to compile rocksdb
-If you don't already have Rust install it
-```sh
-$ winget install LLVM.LLVM
-```
-
```
KolbyML, line 17: If we are removing this in our build instructions why do we still install it in our CI?

trin-state/Cargo.toml
```diff
@@ -19,7 +19,6 @@ ethportal-api = { path = "../ethportal-api" }
 eth_trie = "0.4.0"
```
njgheorghita, line 19: Seems like we could also get rid of this import?

src/bin/purge_db.rs
```diff
@@ -1,181 +0,0 @@
-use anyhow::Result;
```
njgheorghita, line 1: It's not obvious to me that this is worth removing rather than transforming it to work with `sqlite`? I guess ultimately, we're not currently using this script in devops. But at some point I imagine we will? But maybe that's a big effort, and not really worth tackling at this point.

portalnet/src/storage.rs
```diff
@@ -890,15 +869,16 @@ pub mod test {
         let storage_config =
             PortalStorageConfig::new(CAPACITY_MB, temp_dir.path().to_path_buf(), node_id).unwrap();
         let mut storage = PortalStorage::new(storage_config, ProtocolId::History)?;
-        let content_key = generate_random_content_key();
+        use ethportal_api::BlockHeaderKey;
```
njgheorghita, line 872: Could you relocate this to be with the other imports

portalnet/src/storage.rs
```diff
@@ -766,40 +741,44 @@ impl PortalStorage {
...
-                            CREATE INDEX content_id_short_idx ON content_metadata(content_id_short);
-                            CREATE INDEX content_id_long_idx ON content_metadata(content_id_long);";
+                            CREATE INDEX content_size_idx ON content_data(content_size);
+                            CREATE INDEX content_id_short_idx ON content_data(content_id_short);
+                            CREATE INDEX content_id_long_idx ON content_data(content_id_long);";
 
 const INSERT_QUERY: &str =
-    "INSERT OR IGNORE INTO content_metadata (content_id_long, content_id_short, content_key, content_size)
-                            VALUES (?1, ?2, ?3, ?4)";
+    "INSERT OR IGNORE INTO content_data (content_id_long, content_id_short, content_key, content_value, content_size)
```
KolbyML, line 756: Before we deploy this change we should add a field to tag which network the content is for. We can probably extract it from the content_key with our new changes, but this might not always be true. Also we will have to use some sql magic if we don't add a tag field, which I would like this avoid.

KolbyML (approved):
Looks good a few minor concerns but nothing blocking.

portalnet/src/storage.rs
```diff
@@ -399,18 +395,12 @@ impl PortalStorage {
...
         let content_key = hex_encode(content_key).trim_start_matches("0x").to_string();
-        // Revert rocks db action if there's an error with writing to metadata db
-        if let Err(err) = self.meta_db_insert(&content_id, &content_key, value) {
-            debug!(
-                "Error writing content ID {:?} to meta db. Reverting: {:?}",
-                content_id, err
-            );
-            self.db.delete(content_id)?;
+        if let Err(err) = self.db_insert(&content_id, &content_key, value) {
+            debug!("Error writing content ID {:?} to db: {:?}", content_id, err);
```
njgheorghita, line 403: These args can be inlined, so I'm curious why our [lint](https://github.com/ethereum/trin/blob/master/portalnet/src/lib.rs#L2) isn't picking up on it?

.circleci/config.yml
ogenev: `c-kzg` which is a dependency of `reth-primitives` requires it

src/bin/purge_db.rs
ogenev: I thought about this but we still don't know what exactly our database would look like when we refactor it to support beacon and state. I think it is more productive to remove it now and rewrite it if needed at a later stage.

book/src/developers/contributing/build_instructions/windows.md
ogenev: Good catch, reverted this because LLVM is required to compile c-kzg.

ogenev:
Added a "network" column to the main db table as discussed in Trin's sync.

ogenev:
Going to merge this to the upstream `remove-rocksdb` branch.

PR #1060, opened 2023-12-06
Update content key selector bytes for Beacon network 

### LINKED ISSUE #1044: Update content key selector bytes for Beacon network
Update content key selector bytes for Beacon network according to this spec change: https://github.com/ethereum/portal-network-specs/pull/233

ogenev:
### What was wrong?

Closes #1044

### How was it fixed?

- Implemented a custom SSZ encode/decode for beacon content keys because the new selector bytes don't match the SZZ union selector bytes.
- Added an SSZ encode/decode test for beacon content keys

### To-Do


ogenev:
> Looks good to me, but it also seems like we should address the changes to the state network content keys before considering the issue fixed?

Sure, but this can happen in another PR.

ethportal-api/src/types/content_key/beacon.rs
```diff
@@ -119,20 +195,20 @@ impl OverlayContentKey for BeaconContentKey {
 
         match self {
             BeaconContentKey::LightClientBootstrap(key) => {
-                bytes.push(0x00);
+                bytes.push(0x10);
```
mrferris, line 198:
```suggestion
                bytes.push(LIGHT_CLIENT_BOOTSTRAP_KEY_PREFIX);
```
I think we should define these as `const`s near the top of the file

mrferris (approved):

PR #1067, opened 2023-12-12
Bump discv5 version dependency

njgheorghita:
### What was wrong?
Bump `discv5` library to latest `0.4.0` version.
This includes native concurrent discv5 requests, and the latest `enr` library dependency.

### To-Do


KolbyML (approved):
:shipit:

PR #1068, opened 2023-12-12
fix: update base el endpoint for bridge provider

njgheorghita:
### What was wrong?
Investigating some of the issues with our bridge provider.
1. The assumption that we were simply not able to deserialize the requests from different clients seems to have been "mostly" disproven.
  - I don't even think this is a valid concern, I would be shocked if production execution clients have jsonrpc responses formatted differently from each other. I mean that's the entire point of the spec. Of course... I haven't thoroughly audited all jsonrpc responses from different clients, but I think this concern has been a bit of a red herring.
2. It seems like the cause of most of our issues is dshackle. A new load-balancing service that routes "most" of the requests to the pandaops nodes. All pandaops urls use dshackle (as of today) except for those that have an explicitly defined `el-cl` pair. eg. `https://erigon-lighthouse.mainnet.eu1.ethpandaops.io`
3. Dshackle seemed to be particularly finicky with the reth archive node endpoint. There were many instances where reth failed to return some receipts from a batched `eth_getTransactionReceipt` request, which led to invalid receipts roots. That's why I decided to go with erigon in this pr. 

This is the simplest fix. There are some improvements that I'd like to implement in future prs. 
- [ ] Use different urls for `latest` and `backfill` bridge modes.
- [ ] Implement a fallback url, if there's a problem getting responses from the primary uri.
- [ ] Write failed jsonrpc responses to disk, for easy debugging & improve general error messages / logs for failed jsonrpc requests
- [ ] Explore different concurrent rate limits & batch size limits to see if those have an effect on request failure rates.
- [ ] add support for infura / non-pandaops nodes
- [ ] set provider url via env var?

### To-Do

### DIFF
--- a/portal-bridge/src/constants.rs
+++ b/portal-bridge/src/constants.rs
@@ -7,8 +7,13 @@ pub const BATCH_SIZE: u64 = 128;
 // These are only intended to be used by core team members who have access to the nodes.
 // If you don't have access to the PANDAOPS nodes, but still want to use the bridge feature, let us
 // know on Discord or Github and we'll prioritize support for any provider.
+//
 /// Execution layer PandaOps endpoint
-pub const BASE_EL_ENDPOINT: &str = "https://archive.mainnet.ethpandaops.io";
+// This endpoint points towards an archive node (erigon) and skips dshackle (by using el-cl url
+// format), shackle is known to be somewhat buggy has caused some invalid responses.
+// Reth's archive node, has also exhibited some problems with the concurrent requests rate we
+// currently use.
+pub const BASE_EL_ENDPOINT: &str = "https://erigon-lighthouse.mainnet.eu1.ethpandaops.io";
 /// Consensus layer PandaOps endpoint
 /// We use Nimbus as the CL client, because it supports light client data by default.
 pub const BASE_CL_ENDPOINT: &str = "https://nimbus.mainnet.ethpandaops.io";


KolbyML:
There def is a problem with infura though, when I was trying to dump 50 block ranges I was having a hard time finding one where the receipt wouldn't fail or maybe that was the block body. Anyways that shouldn't be too hard to validate

njgheorghita:
@KolbyML what were the block ranges you were working with?

KolbyML (approved):
:shipit:

KolbyML:
I will try and find a block number for you

KolbyML:
Ok so just a quick check there seems to be 2 issues with infura 1 with block bodies which is probably infura specific and 1 with receipts which is probably not infura specific
- block bodies on block 15505052 I get ``Error gossiping block body #15505052: unable to find uncle header`` with json ``[{"id":0,"jsonrpc":"2.0","result":null}]``.
- receipt on block 3005052 I get ``Error gossiping receipt #3005052: missing field 'status'`` with json ``[{"id":0,"jsonrpc":"2.0","result":{"blockHash":"0xc2afb855395ac663f74e69c73473ea27df5228da1a7a161735583f204286030b","blockNumber":"0x2dda7c","contractAddress":null,"cumulativeGasUsed":"0x5208","effectiveGasPrice":"0x4a817c800","from":"0x1e9939daaad6924ad004c2560e90804164900341","gasUsed":"0x5208","logs":[],"logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","root":"0xb53fbf66ff90c9d0ff0931eec73fb1df5352f081bcf5bd3d025fc118df3cf2fb","to":"0x2aa054be8415407e09c7962ce998becdfbb6f7dc","transactionHash":"0x1559d5309d21333846a6499e35de70b54d138a7ca68abec1b618e2bc0c0d74bc","transactionIndex":"0x0","type":"0x0"}}]``

ogenev (approved):
:boat:

njgheorghita:
@KolbyML Thanks for the block details! I'll be sure to test those as I'm wrapping up infura support

Issue #1006, opened 2023-10-23
Beacon Sync progress


ogenev:
This is a tracking issue of the current progress of the beacon sync implementation.

### Tasks
- [ ] Update `LightClientFinalityUpdate` to include `finalized_slot` instead of `signature_slot` in the content key.
- [ ] Update and deploy Trin beacon bridge to serve the new lc finality data.
- [ ] Modify light client crate to not require LC finality update for the current slot on sync startup.
- [ ] Modify light client portal interface strategy for requesting LC finality update every slot (with `finalized_slot` as content key) when tracking the head of the chain.
- [ ] Support building a batch of `LightClientUpdate` records as a response to a period range request.
  - May require a heavy refactoring in overlay service?
  - Special database for the beacon network?
- [ ] Re-evaluate the bridge strategy for pushing `LightClientUpdate` data into the network in real-time?

pipermerriam:
Reguarding the "special database for beacon network"...  sqlite is quite powerful and will allow a schema'd database with complex queries...  I believe that our use of rocksdb is a bit of cargo-cult leftover from execution clients storing all of the state and needing blazing fast read-write access.  My intuition is that in our usage pattern the database read/write speeds are unlikely to be the bottleneck.

PR #1069, opened 2023-12-13
Add Beacon Storage

ogenev:
_This PR is open against the upstream `remove-rocksdb`  branch._

### What was wrong?

We need different storage mechanics for the Beacon portal network

### How was it fixed?

- create a `BeaconStorage` struct which implements the `ContentStore` trait used by the overlay service.
- create a new `lc_update` SQL table for storing and querying `LigthClientUpdate` content (see #1006).
- Add beacon specific implementations to all `ContentStore` methods.
- some common utils for `PortalStorage` and `BeaconStorage` were extracted

It may be useful to review this PR commit by commit.

All storage code and SQL queries are in a single `storage.rs` file, which is a mess. We will refactor this at a later stage.

Ideally, we want to rename `PortalStorage` to `HistoryStorage` but I prefer to do this in a separate PR because I don't want to introduce many diffs here.

In a following PR, we will resolve the TODOs and implement a storage cache for storing lc finality and optimistic updates.

### To-Do

### DIFF
--- a/portalnet/src/storage.rs
+++ b/portalnet/src/storage.rs
@@ -759,21 +759,64 @@ impl ContentStore for BeaconStorage {
         self.store(&key, &value.as_ref().to_vec())
     }
 
+    /// The "radius: concept is not applicable for Beacon network
     fn is_key_within_radius_and_unavailable<K: OverlayContentKey>(
         &self,
         key: &K,
     ) -> Result<ShouldWeStoreContent, ContentStoreError> {
-        let key = key.content_id();
-        let is_key_available = self
-            .lookup_content_key(key)
-            .map_err(|err| {
-                ContentStoreError::Database(format!("Error looking up content key: {err:?}"))
-            })?
-            .is_some();
-        if is_key_available {
-            return Ok(ShouldWeStoreContent::AlreadyStored);
+        let content_key: Vec<u8> = key.clone().into();
+        let beacon_content_key =
+            BeaconContentKey::from_ssz_bytes(content_key.as_slice()).map_err(|err| {
+                ContentStoreError::InvalidData {
+                    message: format!("Error deserializing BeaconContentKey value: {err:?}"),
+                }
+            })?;
+
+        match beacon_content_key {
+            BeaconContentKey::LightClientBootstrap(_) => {
+                let key = key.content_id();
+                let is_key_available = self
+                    .lookup_content_key(key)
+                    .map_err(|err| {
+                        ContentStoreError::Database(format!(
+                            "Error looking up content key: {err:?}"
+                        ))
+                    })?
+                    .is_some();
+                if is_key_available {
+                    return Ok(ShouldWeStoreContent::AlreadyStored);
+                }
+                Ok(ShouldWeStoreContent::Store)
+            }
+            BeaconContentKey::LightClientUpdatesByRange(content_key) => {
+                // Check if any of the periods are available, return AlreadyStored if so otherwise Store
+                let periods =
+                    content_key.start_period..(content_key.start_period + content_key.count);
+
+                for period in periods {
+                    let is_period_available = self
+                        .lookup_lc_update_period(period)
+                        .map_err(|err| {
+                            ContentStoreError::Database(format!(
+                                "Error looking up lc update period: {err:?}"
+                            ))
+                        })?
+                        .is_some();
+                    if is_period_available {
+                        return Ok(ShouldWeStoreContent::AlreadyStored);
+                    }
+                }
+                Ok(ShouldWeStoreContent::Store)
+            }
+            BeaconContentKey::LightClientFinalityUpdate(_) => {
+                // TODO: Check content key availability in cache
+                Ok(ShouldWeStoreContent::AlreadyStored)
+            }
+            BeaconContentKey::LightClientOptimisticUpdate(_) => {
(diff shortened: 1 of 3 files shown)


portalnet/src/storage.rs
```diff
@@ -744,6 +677,496 @@ impl PortalStorage {
...
+            value_size
+        ],
+    ) {
+        Ok(_) => Ok(()),
+        Err(err) => Err(err.into()),
+    }
+}
+
 // SQLite Statements
 const CREATE_QUERY: &str = "CREATE TABLE IF NOT EXISTS content_data (
```
njgheorghita, line 1171: A lot of these `const` sql query variable names feel insufficiently descriptive now that you've added the more-descriptive LC table names. But, this is just a nitpick to keep in mind in upcoming refactor prs. It probably won't matter if the `BeaconStorage` and `HistoryStorage` are separated into different modules.

portalnet/src/storage.rs
```diff
@@ -744,6 +677,496 @@ impl PortalStorage {
...
+        match conn.execute(
+            INSERT_LC_UPDATE_QUERY,
+            params![period, value, 0, value_size],
+        ) {
+            Ok(_) => Ok(()),
+            Err(err) => Err(err.into()),
+        }
+    }
+
+    pub fn store(
```
njgheorghita, line 885: It feels like this (& any other shared methods inside this `impl BeaconStorage {}`) should be moved to the `ContentStore` trait... or maybe a new shared trait? I'm not sure, since IIRC `ContentStore` was primarily created to enable `MemoryContentStore` and use that inside tests. Maybe, we should create a new trait that will be shared amongst all subnetwork storage types. This could also include shared metrics reporting methods. Just thinking out loud.

portalnet/src/storage.rs
```diff
@@ -744,6 +677,496 @@ impl PortalStorage {
...
+            }
+        }?
+        .num_bytes;
+
+        self.metrics.report_content_data_storage_bytes(sum);
+
+        Ok(sum as u64)
+    }
+
+    pub fn lookup_content_key(&self, id: [u8; 32]) -> anyhow::Result<Option<Vec<u8>>> {
```
njgheorghita, line 1003: Am I missing something? Is this not essentially the same as `lookup_content_value()`?

portalnet/src/storage.rs
```diff
@@ -744,6 +677,496 @@ impl PortalStorage {
...
+    }
+
+    /// Public method for looking up a content value by its content id
+    pub fn lookup_content_value(&self, id: [u8; 32]) -> anyhow::Result<Option<Vec<u8>>> {
+        let conn = self.sql_connection_pool.get()?;
+        lookup_content_value(id, conn)?
+    }
+
+    /// Public method for looking up a  light client update value by period number
+    pub fn lookup_lc_update_value(&self, period: u64) -> anyhow::Result<Option<Vec<u8>>> {
```
njgheorghita, line 1050: Also, isn't this exactly the same as `lookup_lc_update_period()`?

portalnet/src/storage.rs
ogenev: Yes, the plan is to separate those per network but the goal now is to add the functionality asap and refactor afret.

portalnet/src/storage.rs
ogenev: "store" is not a required method because it is wrapped by `put` which is in the `ContentStore` trait and this is what is used by tests and the overlay service. I'm using it in `BeaconStorage` because I try to follow a similar structure as `PortalStorage`, but this shouldn't be a hard requirement.

Anyway, I see some benefits of having additional traits but imo we will have more visibility here when we start to refactor the code.

portalnet/src/storage.rs
ogenev: `lookup_content_value` accepts content id and give us a content value result, while `lookup_content_key` accepts content id and returns a content key.

portalnet/src/storage.rs
ogenev: Similar logic as above, `lookup_lc_update_period` returns light client update period number while `lookup_lc_update_value` returns the light client update ssz bytes (content value).

portalnet/src/storage.rs
```diff
@@ -744,6 +677,496 @@ impl PortalStorage {
...
+
+    fn put<K: OverlayContentKey, V: AsRef<[u8]>>(
+        &mut self,
+        key: K,
+        value: V,
+    ) -> Result<(), ContentStoreError> {
+        self.store(&key, &value.as_ref().to_vec())
+    }
+
+    /// The "radius: concept is not applicable for Beacon network
```
njgheorghita, line 762: Seems like this function name/purpose should be updated in upcoming refactoring so that it can be shared amongst storages while also being logically consistent.

portalnet/src/storage.rs
njgheorghita: Ahhh, yeah that makes sense. Imo it's worth adding a docstring here (and to `lookup_content_value()`), even though the fn names are quite descriptive, it can be useful (at least I would've found it useful in this case) to have the functionality spelled out in plain english

njgheorghita (approved):
Looks good! Thanks for answering my questions. Agreed with all the todo refactoring ideas. 🚢

PR #1074, opened 2023-12-18
feat: add utp transfer limit for requests not initiated by rpc

KolbyML:
#### What was wrong?
Currently Trin can open up limitless uTP streams. 

X = base X amount of memory trin uses
Y = arbatrary statical average each new uTP connection incurs

So currently Trins memory usage is =  X + (Y * (# of connections))
This is unbounded growth, this PR proposes to set a configurable limit of how many inbound and outbound uTP connections can be made to the node.

I am proposing this limit for connections that are incurred from the node running itself, not requestions started from JSON-RPC requests.

#### why seperate inbound/outbound limits
It is very common for ISP's to provide high download speeds but low upload speeds. So being able to set these seperately can help people who face those kinds of constraints.

#### What will this limit not be applied to
Requests initiated by rpc, we can add a seperate limit if we want but this is a user experience kind of thing. If a user is paying for bandwidth they shouldn't be rate limited, what should be rate limited is its autonomous interations with other nodes.
#### What will this limit be applied to
handle_offer
handle_find_content
poke_content
^ calls which are initiated from the overlay service, not by the rpc
### future work
- optimize uTP more however that would look
- ad hoc protection added to utp controller
- protection from bad uTP connections (implement forced connection shutdowns) currently uTP only supports graceforce connection shutdowns.

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -184,7 +184,7 @@ pub struct TrinConfig {
 
     #[arg(
         long = "utp-transfer-limit", 
-        help = "The limit of max inbound and max outbound uTP transfers", 
+        help = "The limit of max background uTP transfers for any given channel (inbound or outbound) for each subnetwork", 
         default_value_t = DEFAULT_UTP_TRANSFER_LIMIT,
     )]
     pub utp_transfer_limit: usize,
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -15,6 +15,7 @@ pub const DEFAULT_WEB3_HTTP_ADDRESS: &str = "http://127.0.0.1:8545/";
 pub const DEFAULT_WEB3_HTTP_PORT: u16 = 8545;
 pub const DEFAULT_WEB3_WS_PORT: u16 = 8546;
 pub const DEFAULT_DISCOVERY_PORT: u16 = 9009;
+pub const DEFAULT_UTP_TRANSFER_LIMIT: usize = 50;
 pub const BEACON_NETWORK: &str = "beacon";
 pub const HISTORY_NETWORK: &str = "history";
 pub const STATE_NETWORK: &str = "state";
@@ -181,6 +182,13 @@ pub struct TrinConfig {
     )]
     pub ws_port: u16,
 
+    #[arg(
+        long = "utp-transfer-limit", 
+        help = "The limit of max inbound and max outbound uTP transfers", 
+        default_value_t = DEFAULT_UTP_TRANSFER_LIMIT,
+    )]
+    pub utp_transfer_limit: usize,
+
     #[command(subcommand)]
     pub command: Option<TrinConfigCommands>,
 }
@@ -214,6 +222,7 @@ impl Default for TrinConfig {
             ws: false,
             ws_port: DEFAULT_WEB3_WS_PORT,
             command: None,
+            utp_transfer_limit: DEFAULT_UTP_TRANSFER_LIMIT,
         }
     }
 }
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -2777,7 +2777,7 @@ mod tests {
         let discv5_utp =
             crate::discovery::Discv5UdpSocket::new(Arc::clone(&discovery), utp_talk_req_rx);
         let utp_socket = utp_rs::socket::UtpSocket::with_socket(discv5_utp);
-        let utp_controller = UtpController::new(50, utp_socket);
+        let utp_controller = UtpController::new(50, Arc::new(utp_socket));
         let utp_controller = Arc::new(utp_controller);
 
         let node_id = discovery.local_enr().node_id();
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -30,14 +30,15 @@ use tokio::{
     sync::{
         broadcast,
         mpsc::{self, UnboundedReceiver, UnboundedSender},
+        OwnedSemaphorePermit,
     },
     task::JoinHandle,
 };
 use tracing::{debug, enabled, error, info, trace, warn, Level};
-use utp_rs::{conn::ConnectionConfig, socket::UtpSocket, stream::UtpStream};
+use utp_rs::{cid::ConnectionId, conn::ConnectionConfig, stream::UtpStream};
 
 use crate::{
-    discovery::Discovery,
+    discovery::{Discovery, UtpEnr},
     events::{EventEnvelope, OverlayEvent},
     find::{
         iterators::{
@@ -51,6 +52,7 @@ use crate::{
     gossip::propagate_gossip_cross_thread,
     types::node::Node,
     utils::portal_wire,
+    utp_controller::UtpController,
 };
 use ethportal_api::{
     generate_random_node_id,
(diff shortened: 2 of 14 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -1258,13 +1305,14 @@ where
         } else {
             String::with_capacity(0)
         };
-        let cid = self.utp_socket.cid(enr, false);
+        let cid: utp_rs::cid::ConnectionId<crate::discovery::UtpEnr> =
```
njgheorghita, line 1308: Imports should happen at the top of the file unless there's a good reason not to

portalnet/src/overlay_service.rs
```diff
@@ -1322,6 +1370,8 @@ where
             {
                 debug!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), content_keys = ?content_keys_string, "unable to process uTP payload");
             }
+            // explictically drop semaphore permit in thread so it gets moved into it
```
njgheorghita, line 1373: `explicitly` ... but also i'm not sure what you mean by `so it gets moved into it`, so what gets moved into what?

portalnet/src/overlay_service.rs
```diff
@@ -1651,6 +1708,10 @@ where
                     }
                 }
             }
+            // explicitally drop permit in new thread so it it is included in the move
```
njgheorghita, line 1711: `so it it is included in the move` is also confusing. Typically, it's much easier to understand comments if you avoid vague nouns like `it`

njgheorghita (commented):
Nice, I think it's a great start. This is also a promising location for how we handle multiple, concurrent utp streams for the same content id. 

I'm also curious if you've tested this locally, and observed that the utp metrics never go above 50?

portalnet/src/overlay_service.rs
KolbyML: so that permit gets moved into the new thread.

njgheorghita: the question is not so much for me, but more to update the comment with the answer for whoever is reading the code in the future and might have the same question

KolbyML: oh I get what you mean now `it` isn't a descriptive word of the object in this case the permit fair enough :+1:

portalnet/src/overlay_service.rs
```diff
@@ -1116,8 +1141,14 @@ where
                 ))
             }
         };
-        match self.store.read().get(&content_key) {
-            Ok(Some(content)) => {
+        match (
+            self.store.read().get(&content_key),
+            self.utp_controller
```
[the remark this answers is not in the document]
KolbyML: Good catch

portalnet/src/overlay_service.rs
```diff
@@ -1170,7 +1201,7 @@ where
                     Ok(Content::ConnectionId(cid_send.to_be()))
                 }
             }
-            Ok(None) => {
+            (Ok(Some(_)), _) | (Ok(None), _) => {
```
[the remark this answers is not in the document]
KolbyML: If we don't have a permit avaliable we want to drop it and send the other client a list of Enr's. By using _ for the permit, if one is obtained and doesn't fit the conditions it is instantly dropped.

- (Ok(Some(_)), _) we didn't obtain a semaphore for some reason
- (Ok(None), _) we didn't have a utp transfer to generate so we just drop whatever it returned.

KolbyML:
resolved most if not all the pr concerns. ready for another look

portalnet/src/overlay_service.rs
```diff
@@ -944,13 +958,23 @@ where
                 let content_items = vec![(content_key.clone().into(), content.clone())];
                 let offer_request = Request::PopulatedOffer(PopulatedOffer { content_items });
 
+                let permit = match utp_controller
```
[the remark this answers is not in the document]
njgheorghita: typo: `continue`

portalnet/src/overlay_service.rs
```diff
@@ -1651,6 +1709,10 @@ where
                     }
                 }
             }
+            // explicitally drop permit in the thread so the permit is included in the thread
```
njgheorghita, line 1712: `explicitly`

portalnet/src/utp_controller.rs
```diff
@@ -0,0 +1,48 @@
...
+use utp_rs::{cid::ConnectionId, conn::ConnectionConfig, socket::UtpSocket, stream::UtpStream};
+
+use crate::discovery::UtpEnr;
+
+/// UtpController is meant to be a container which contains all code related to/for managing uTP streams
+/// We are implementing this because we want the utils of controlling uTP connection to be as contained as it can, instead of
+/// extending overlay_service even more.
+/// Currently we are implementing this to control the max utp_transfer_limit
+/// But in the future this will be where we implement
+/// - thunder hurding protection
```
njgheorghita, line 13: `thundering herd protection`

portalnet/src/overlay_service.rs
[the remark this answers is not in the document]
njgheorghita: My point is that it's very difficult to read that line of code and understand what's going on. At the very least it deserves a comment to help readers parse the logic

njgheorghita (commented):
Could you test this locally, let your node run for a while and confirm that you never observe the `active_utp_streams` metrics rise above 50

KolbyML:
I ran it locally. The issue is there currently isn't enough uTP requests on the network to hit the limit 
image

So it currently isn't possible to see this limit reached on mainnet, we will be able to when we reset the nodes though when we do the db change

njgheorghita:
@kolby If you just decrease the max limit via the cli flag to something like 5 then I'd imagine you should be able to see the desired behavior

KolbyML:
:laughing:

njgheorghita:
:sweat_smile: You are correct, sorry @kolby !

KolbyML:
image
Inbound seems to be working, but something is wrong with outbound so I will debug that and try to see what is wrong

KolbyML:
image
Nice I figured out why the limit wasn't working. I added it to poke, but not propagate_gossip_cross_thread. Silly me

Now I just to do figure out why accept_with_cid is deadlocking I will make a seperate PR for that as that isn't related to this PR, hopefully I can find it quickly

ethportal-api/src/types/cli.rs
```diff
@@ -181,6 +182,13 @@ pub struct TrinConfig {
     )]
     pub ws_port: u16,
 
+    #[arg(
+        long = "utp-transfer-limit", 
+        help = "The limit of max inbound and max outbound uTP transfers",
```
njgheorghita, line 187: *nitpick* This is just a bit confusing to understand how the limit is actually applied. Eg is 50 the total for both inbound/outbound? Or is it 50 for inbound, 50 for outbound? What about on a per/subnetwork basis?

Maybe something like: `The limit of max uTP transfers for any given channel (inbound or outbound) for each subnetwork` ? But if you can think of a more concise way to phrase the info then 👍

njgheorghita (approved):
Nice! Thanks for doing the extra testing :ship:

KolbyML:
@njgheorghita here is the uTP PR https://github.com/ethereum/utp/pull/122 I was talking about with the fixes so we can merge this PR (uTP Transfer limit).

ethportal-api/src/types/cli.rs
KolbyML: That is a great idea :+1: I will add that quick to this pr

njgheorghita (approved):
:ship:

PR #1076, opened 2023-12-24
fix: remove outdated trin-core

qiweiii:
### What was wrong?

When reading the trin book, found references to `trin-core` crate which do not exist anymore

### How was it fixed?

Updated the book and related diagrams, also updated outdated `repository` link in some cargo toml


book/src/developers/protocols/discovery.md
qiweiii, line 1: these files are auto created when run mdbook locally, do I need to remove them?

ogenev: I think those are created because the headers are included in `src/SUMMARY.md`.

IMO removing the headers for the empty files from `SUMMARY.md`  is better than including empty files.

ogenev (approved):
Thanks!

PR #1077, opened 2023-12-24
feat: add rustfmt imports_granularity

### LINKED ISSUE #1046: Enforce imports granularity config option in rustfmt
It would be nice if we style all imports in a systematic way across the repo.

Rstfmt has a [config option ](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#imports_granularity)that requires a `rustfmt.toml` config file and it will auto-format those for us.

However, this option requires a rust nightly build to run, so we should update CI with the nightly rustfmt command and have the nightly toolchain installed locally (for running `rustfmt` locally before push):

```sh
cargo +nightly fmt --all -- --check
```

qiweiii:
### What was wrong?

resolve #1046

### How was it fixed?

- add rustfmt.toml
- run cargo +nightly fmt --all
- update ci to use nightly toolchain

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,7 +50,7 @@ commands:
           command: apt install -y libssl-dev
       - run:
           name: Install libclang for rocksdb
-          command: apt install clang
+          command: apt install -y clang
 orbs:
   rust: circleci/rust@1.6.0
   win: circleci/windows@5.0
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -132,9 +132,9 @@ jobs:
           command: |
             echo "$DOCKERHUB_PASS" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
             docker push $IMAGE_NAME:latest-bridge
-  lint:
+  cargo-fmt:
     description: |
-      Check linting with Clippy and rustfmt.
+      Check linting with rustfmt.
     resource_class: xlarge
     executor: docker-nightly
     environment:
@@ -146,6 +146,19 @@ jobs:
       - run:
           name: Run rustfmt
           command: cargo +nightly fmt --all -- --check
+  cargo-clippy:
+    description: |
+      Check linting with Clippy.
+    resource_class: xlarge
+    executor:
+      name: rust/default
+      tag: 1.71.1
+    environment:
+      RUSTFLAGS: "-D warnings"
+      RUST_LOG: "debug"
+    steps:
+      - checkout
+      - install-depends
       - run:
           name: Install Clippy
           command: rustup component add clippy
@@ -288,7 +301,8 @@ workflows:
           filters:
             branches:
               only: master
-      - lint
+      - cargo-fmt
+      - cargo-clippy
       - build
       - test
       - check-windows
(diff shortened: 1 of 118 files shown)


qiweiii:
the lint ci failed, not very familiar with circleci, please help😁

ogenev (commented):
The new rustlang/rust:nightly-slim image seems to break the CircleCI lint job because the sudo command is not included.

I'm not sure what the slim version contains compared to the regular nightly but I guess we can get away with the regular image.

ogenev (commented):
We need also to update the fmt command in [book/src/developers/contributing/git/pull_requests.md](https://github.com/ethereum/trin/blob/4645fef9979d358f3c73a440147c8f076fb7aebe/book/src/developers/contributing/git/pull_requests.md?plain=1#L12).

ogenev:
The lint is still failing, it looks like we don't need the `sudo` command for the nightly image used inside the `install-depends` step. But this step is also used in other jobs. 

Creating an `install-depends-nighly` step without `sudo` and adding this step for the lint (instead of the default `install-depends`) is a reasonable solution.

qiweiii:
Nice, lint works now, I also added a rust-toolchain file so developers don't need to manually switch toolchain

ogenev:
> Nice, lint works now, I also added a rust-toolchain file so developers don't need to manually switch toolchain

Great but we don't want to pin Trin to the latest nightly version (at least we need to have a discussion about that with the rest of the team). 

The current workflow is to use `nightly`  only for clippy, so let's remove the `rust-toolchain` file for now. In the future, if we decide, we can pin Trin to a nightly specific version (for example `nightly-2023-01-04`) and this will not require to manually set nighly clippy with `+nighlty`.

ogenev:
By the way, while we are still here, could you also add  those configs to the `rustfmt.toml` file:

```
comment_width = 100
wrap_comments = true
format_code_in_doc_comments = true
doc_comment_code_block_width = 100
```

.circleci/config.yml
```diff
@@ -119,29 +133,27 @@ jobs:
...
+      - run:
+          name: Run rustfmt
+          command: cargo +nightly fmt --all -- --check
+      - run:
+          name: Install Clippy
+          command: rustup component add clippy
+      - setup-and-restore-sccache-cache
+      - run:
+          name: Run Clippy
+          command: cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
```
ogenev, line 155: Looking at this now, I assume here `clippy` is running with the nightly version, but because we don't pin Trin to nightly there may be some discrepancies between nightly clippy and stable clippy in the future.

It would be better if we split the `lint` job to `cargo-fmt` and `cargo-clippy` and run clippy with the default stable toolchain.

 If you don't feel like doing it in this PR, we can leave it for a future fix (this will not be a blocker for merging this PR).

qiweiii: I could try it in this PR, seems like a simple change

qiweiii: all done!

ogenev (approved):
Looks good to me :100: !. Thank you for this contribution, really happy to see some structure to our imports. :heart:

PR #1079, opened 2024-01-03
fix: use beacon types for beacon rpc instead of history types

KolbyML:
### What was wrong?
I was looking into getting a foundation built out so we can have beacon tests on portal hive and I ran into a issue oh no.

Beacon RFC was out of date with the spec, then I found for FC, RFC, TRFC, PLCK we were returning history Key/Value types :scream: 
### How was it fixed?
By making variants of ContentInfo, PaginateLocalContentInfo, TraceContentInfo for beacon and using them

### DIFF
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -1,11 +1,9 @@
 use crate::{
     types::{
+        beacon::{ContentInfo, PaginateLocalContentInfo, TraceContentInfo},
         content_key::beacon::BeaconContentKey,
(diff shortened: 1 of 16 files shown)


ogenev (approved):
Nice, looks good! :100: 

Please rebase to reformat the imports.

KolbyML:
utp-test failing has nothing to do with us I am looking to resolve the main issue, but the will take a few days to resolve. I am going to merge this as it has been approved and it is clear the CI issue is external

PR #1056, opened 2023-12-04
POC test to show bug DOS vector exists Show 1 node accepting multiple streams for the same content

KolbyML:
### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #1080, opened 2024-01-03
feat: implement basic thundering herd protection

KolbyML:
### What was wrong?
If we started 100 brand new nodes we end up in a situation where nodes endlessly regossip to each other eating resources and driving the network to a crawl. A big issue was a node would accept multiple streams from multiple peers as well.

A proof of concept recreating this issue https://github.com/ethereum/trin/pull/1056
A deeper write up of this kind of problem https://github.com/ethereum/trin/pull/1056

This PR is meant to be a basic implementation of idea B, but written to allow for C and D to be built upon it. A part of this is the idea of creating a utp_connection controller a struct where we manage utp connections instead of doing it in overlay_service, which should we want to avoid adding more and more logic to. this utp_controller struct will have the soul job of managing uTP connections we have with other nodes, or at least that is the long term idea.
### To-Do


KolbyML:
Closing this as nick is working on the fix for this

PR #1081, opened 2024-01-06
feat: add an instrument to portal-bridge for more clear logs

KolbyML:
### What was wrong?
I was trying to read through bridge logs and I couldn't tell if they were coming from the history or beacon bridge. This makes grepping and debugging hard.
### How was it fixed?

with this change all the logs generated by the history bridge are marked by history and all the logs generated by the beacon bridge are tagged with beacon

```nim
2024-01-06T15:47:10.139336Z  INFO history: surf::middleware::logger::native: request completed    
2024-01-06T15:47:10.143257Z DEBUG history:send_async{method=POST uri=https://erigon-lighthouse.mainnet.eu1.ethpandaops.io/}:handler{id=8}: isahc::handler: Connection #0 to host erigon-lighthouse.mainnet.eu1.ethpandaops.io left intact
2024-01-06T15:47:10.290635Z DEBUG history: portal_bridge::bridge::history: Gossip: Block #18949112 Receipts
2024-01-06T15:47:10.321927Z DEBUG history: hyper::client::connect::http: connecting to 127.0.0.1:8545
2024-01-06T15:47:10.322390Z DEBUG history: hyper::client::connect::http: connecting to 127.0.0.1:8545
2024-01-06T15:47:10.322571Z DEBUG history: portal_bridge::gossip: Unable to locate content on network, after failing to gossip, retrying in 15s seconds. content key="0x028ae09f4a4939679fe55e0481dc24fc4a69ca9a87ba8c22dd8fe52b6597949352"
```

previously it was unclear for example which bridge generated ``DEBUG: portal_bridge::gossip``, but now it is marked ``DEBUG history: portal_bridge::gossip``

This also improves grepibility since you can just grep logs by network which is awesome.

PR #1082, opened 2024-01-06
fix(bridge): prevent await from blocking new gossips

KolbyML:
### What was wrong?
we need to merge this PR first https://github.com/ethereum/trin/pull/1081

serve() can block new blocks from being gossipped for a very long time currently before waiting a worse case right now is
10s wait for letting head propagate
15s * (3 (header, body, receipt)
60s * 3 if trin was the only bridge and we gossiped to a fleet of empty fluffy nodes. The uTP transfer would be successful technically but trin uTP would return a close failure. RFC would find it and return successful though

that is a wait of 3m 55s.

maybe there are other combinations which are worse, but the point is we don't want this to block latest blocks from starting the process of being gossiped.
### How was it fixed?
Instead of awaiting a vector of handles, we just spawn new tokio threads for each block we are gossiping. now serve() can't block gossip from happening.

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -6,10 +6,9 @@ use std::{
 };
 
 use anyhow::{anyhow, bail};
-use futures::stream::StreamExt;
 use ssz::Decode;
 use tokio::time::{sleep, Duration};
-use tracing::{debug, info, warn};
+use tracing::{debug, info, warn, Instrument};
 
 use crate::{
     api::execution::ExecutionApi,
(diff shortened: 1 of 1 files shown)


KolbyML:
If that is the case I can create a PR which rate limits the production of new threads where we can set a max rate. I believe this is the right direction and as shouldn't buil,d potential blocks in our code, but instead controlled flow rate

portal-bridge/src/bridge/history.rs
```diff
@@ -193,9 +201,20 @@ impl HistoryBridge {
                 None
             };
             info!("fetching headers in range: {gossip_range:?}");
-            self.serve(gossip_range, epoch_acc)
-                .await
-                .expect("Error serving headers in backfill mode.");
+            for height in gossip_range.clone() {
+                let cloned_self = self.clone();
```
njgheorghita, line 205: Is this a good idea to simply clone the entire bridge in order to enable spawning threads for each new block to be gossiped?

KolbyML:
Ok PR 1088 and 1087 should resolve you concerns :+1:

PR #1084, opened 2024-01-08
feat: add beacon storage cache for light client updates

ogenev:
### What was wrong?
Light client finality and optimistic updates are transient data for the beacon network and we don't need to store those values in the main DB.

### How was it fixed?
Implemented a cache for the beacon storage and updated the put and get methods to retrieve those updates from the cache.

### To-Do


KolbyML (approved):
:shipit:

PR #1086, opened 2024-01-08
style: remove check flag from makefile lint

njgheorghita:
### What was wrong?
I'd rather remove the `--check` flag from the newly updated makefile `lint` commands. It makes sense to leverage the automatic correction of these lints rather than going through manually and fixing each error.

### How was it fixed?
Removed the flag from the makefile `lint` commands.

### To-Do

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,12 +7,12 @@ help:
 		@echo "push-docker-image - push docker image"
 
 lint:
+		cargo +nightly fmt --all
 		cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
-		cargo +nightly fmt --all -- --check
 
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
Assuming fmt can't make mistakes this seems good

ogenev (approved):

njgheorghita:
Made one small update, to run `fmt` before clippy. I use the `make lint` command fairly regularly, and it's a much cleaner user experience to run the auto formatting before moving on to clean up the clippy errors. Imo, this is a fairly minor change, so gonna go ahead with the merge, but happy to revisit if anyone feels otherwise.

PR #1087, opened 2024-01-09
feat: add rate limit for history backfill. Will be opening a new PR with a different approach. Wrote write up here on why.

KolbyML:
### What was wrong?
Here https://github.com/ethereum/trin/pull/1082#issuecomment-1883285454 Nick expressed concerns about backfill not being ratelimited to prevent pandaops from being overloaded.
### How was it fixed?
By using tokio_utils ratelimiter.  So now we get the combined benefits of the tokio:spawn solution and also a ratelimiter to not overwelm pandaops!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,6 +612,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-stdin"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1ff8b5d9b5ec29e0f49583ba71847b8c8888b67a8510133048a380903aa6822"
+dependencies = [
+ "tokio",
+]
+
+[[package]]
 name = "async-stream"
 version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


njgheorghita:
I'm having a hard time understanding how exactly the `rate_limiter` mechanism is working here... It doesn't seem to me like the docs are too descriptive. Or maybe I'm just confused. What exactly is the rate it's limiting? Like the clock rate used by the cpu?

What I prefer about the `buffer_unordered` solution is that it's fairly simple. eg. "There will never be more than X simultaneous threads". If we adopt your `rate_limiter` solution, I'd be interested in understanding better how does the "rate limit" correlate to # of simultaneous threads. eg. In practice, how many threads are active at a given point in time based on different rate limits that are set. I also don't think this should be externally configurable via cli flag. It's too confusing / abstract for the average user. We should find a reasonable rate limit that is stable (eg. can run backfill on an entire epoch without any errors), based off of local experiments, and set that as a constant.

KolbyML:
Closing this as me and nick disscused we will both propose our own solutions and pick the best one here is mine https://github.com/ethereum/trin/pull/1091

PR #1088, opened 2024-01-09
refactor: history bridge remove self from calls

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/pull/1082#discussion_r1446375882
### How was it fixed?
clones what is needed instead of self itself


njgheorghita (approved):
:ship:

PR #1089, opened 2024-01-09
feat(bridge): add support for 3rd party and local providers

njgheorghita:
### What was wrong?
It's about time that we add support to the bridge for non-pandaops providers. A primary use case being that the kurtosis tooling is functionally useless for anybody who doesn't have pandaops access.

### How was it fixed?
Added support for 3rd party providers / local nodes via a configurable url. Removed the `Middleware` concept, since it was somewhat useless. Rather than having a `Middleware` type containing the cl & el providers, the `HistoryApi` & `BeaconApi` should just house their own clients, since `HistoryApi` never needs cl provider data and `BeaconApi` never needs el provider data.

### To-Do


njgheorghita:
@KolbyML Whenever you have some time, could you pull down this branch and verify that you can get it working against a local geth node? At least for only gossiping history network data. The guide in the `README` should be sufficient, but lmk if you have any questions

KolbyML:
> @KolbyML Whenever you have some time, could you pull down this branch and verify that you can get it working against a local geth node? At least for only gossiping history network data. The guide in the `README` should be sufficient, but lmk if you have any questions

Yeah I can do that no problem, I will probably be able to do it tomorrow morning.

KolbyML:
I tested it on latest, block 10 million, block 10k and it work on geth. So looks like we are good to go :+1:

KolbyML:
After this PR is merged we will add a rate limiter on surf requests to EL/CL https://docs.rs/surf-governor/latest/surf_governor/struct.GovernorMiddleware.html

This PR is nice because it is already going into the right direction to use surf_governor for this!!!!!!!

njgheorghita:
@KolbyML Interesting find... so then is my understanding correct that the implication of this "governer" is that we could use it instead of the semaphore / `buffer_unordered` solutions? After all, the rate we're mostly concerned about is the http request rate to the providers. (though there is a possible problem with overloading the trin node with too many concurrent rfc/gossip requests, so we might still want some rate limiting outside of the http level).

KolbyML:
I think we should have both cause they are rate limiting 2 different constraints (EL/CL infastructure usage) and (Trin usage)

KolbyML (approved):
:shipit:

portal-bridge/src/api/execution.rs
```diff
@@ -85,11 +123,115 @@ impl ExecutionApi {
...
+                Ok::<Vec<Value>, Box<dyn std::error::Error>>(acc)
+            }) {
+            Ok(val) => Ok(serde_json::to_string(&val)?),
+            Err(err) => Err(anyhow!("Unable to flatten batch request: {err:?}")),
+        }
+    }
+
+    async fn send_batch_request(&self, requests: Vec<JsonRequest>) -> anyhow::Result<Vec<Value>> {
+        if requests.len() > BATCH_LIMIT {
+            warn!("Attempting to send requests outnumbering pandaops limit")
```
ogenev, line 156: This warning should be changed to match agnostic providers.

portal-bridge/src/pandaops.rs
ogenev, line 1: The reason for the `Middleware` was that we may want to make the batch request function available also to the beacon network API. Indeed, at the moment we don't have a use case for beacon batch requests but if this need arises, we will want to refactor again this and add another layer of abstraction to manage all common functions between EL and CL endpoints.

My feeling is that it would be better if we keep the Middleware abstraction because it keeps the code more flexible and any common CL and EL function can go there, even if we don't need this at the moment.

I'll leave it to you to decide what to do here, I don't want this to be a blocker for the PR.

ogenev (approved):

portal-bridge/src/pandaops.rs
njgheorghita: Ah, ok thanks for the explanation. Just to clarify, you could see a use case where the beacon network api would want to send batch requests to the CL provider, correct? And there's no use case where the beacon network api would want to request data from the el provider?

ogenev: > Just to clarify, you could see a use case where the beacon network api would want to send batch requests to the CL provider, correct? And there's no use case where the beacon network api would want to request data from the el provider?

Correct :+1: .

njgheorghita: Thanks for clarifying. So, the initial motivation for removing `Middleware` was to align the EL provider with the `ExecutionApi` and the CL provider with the `BeaconApi`. If there is no reason that the `ExecutionApi` needs CL data (and v.v.) then I figured, let's align the `Api`s with only the providers they need. 

Tbh, I hadn't considered the use case you've brought up (eg. `BeaconApi` using batched requests and sharing that mechanism with `ExecutionApi`). But, thinking about it a bit more...

Requests to the EL provider & CL provider are fundamentally different. EL providers are json inserted into the http request body. CL providers use the url (aka rest api style?) to interpret the request. I don't think there is any way for these two, fundamentally different, types of requests to share the same `"batch_requests"` functionality. In fact, I'm not sure it's even possible to batch requests for the cl provider. For the EL, we "batch" the json requests that are inserted into the http request body. This isn't possible for the CL. 

There are some other considerations, but honestly the one I mentioned above I believe is strong enough to move ahead with the merge on this pr as-is. Maybe, I'm missing something? More than happy to revisit this if I am.

PR #1090, opened 2024-01-10
Refactor storage code

ogenev:
### What was wrong?
All storage code was in a single file: `portalnet/storage.rs`. This makes it very hard to maintain and read because we have implementations and logic for multiple networks plus common utils and traits.

### How was it fixed?
- rename `PortalStorage` to `HistoryStorage` (todo left from previous PRs).
- create a `trin-storage` workspace crate containing common storage utils, traits, and types.
- add  `storage.rs` file to `train-history` crate containing storage implementations for the history network.
- add `storage.rs` file in the `trin-beacon` crate containing storage implementations for the beacon network.
- add `storage.rs` file in the `trin-state` crate containing a template for future state storage implementation.
-  remove `portalnet/storage` module.

### To-Do


trin-history/src/storage.rs
```diff
@@ -0,0 +1,818 @@
...
+        Ok(ShouldWeStoreContent::Store)
+    }
+
+    fn radius(&self) -> Distance {
+        self.radius
+    }
+}
+
+impl HistoryStorage {
+    /// Public constructor for building a `PortalStorage` object.
```
njgheorghita, line 87: `/// Public constructor for building a HistoryStorage object.`

trin-history/src/storage.rs
```diff
@@ -0,0 +1,818 @@
...
+
+    /// Sets the radius of the store to `radius`.
+    pub fn set_radius(&mut self, radius: Distance) {
+        self.radius = radius;
+        self.metrics.report_radius(radius);
+    }
+
+    /// Returns a paginated list of all available content keys from local storage (from any
+    /// subnetwork) according to the provided offset and limit.
+    pub fn paginate(
```
njgheorghita, line 157: For example, my proposed `StorageManager` would house functions like this

njgheorghita (commented):
This is great work. I really see the advantage of the refactor, and the flexibility it affords us now that we can use per-subnetwork caches. 

In general, I have some questions about the structure. 
1. I'm still in favor of having federated sqlite dbs for each subnetwork. Though, iirc there are some good reasons not to go this route w/ regards to r/w bottlenecks?
2. What feels a bit off to me is that this structure (`BeaconStorage`/`HistoryStorage`/`StateStorage`/etc) implies that we do have federated sqlite dbs for each subnetwork. And it implies that each `XStorage` type only operates on data relevant to that subnetwork. When in fact, each `XStorage` type is operating on the same, single sqlite db, containing all data from every subnetwork (eg pruning data in `HistoryStorage` might remove a piece of `Beacon` content). This makes it tricky to reason about what each fn is actually doing. Is it operating only on subnetwork data, or on all data in the db together?
3. Edge cases. User runs trin w/ `history,beacon,state` activated. Then restarts trin with only `beacon` activated. Now we have a lot of state data in the db that's occupying space. If we used federated dbs, we could just trash the history db. 

I'm not going to push too hard for federated dbs, though I still do think it will help a lot here, even at the expense of performance. Maybe something like a `StorageManager` object could be a useful compromise here? (though, I'm happy if it comes in a future pr.) This would be responsible for the overall state of the entire storage. And could help remove some of the subnetwork-agnostic fns out from the subnetwork `Storage` implementations. Really, my main complaint is that it's mentally tasking to track the difference b/w subnetwork-specific operations and subnetwork-agnostic operations. 

I'm also happy to jump on a zoom to discuss this, since I could be missing exactly what's going on here.

trin-beacon/src/storage.rs
```diff
@@ -0,0 +1,532 @@
...
+        let storage = Self {
+            node_data_dir: config.node_data_dir,
+            sql_connection_pool: config.sql_connection_pool,
+            storage_capacity_in_bytes: config.storage_capacity_mb * BYTES_IN_MB_U64,
+            metrics,
+            network: ProtocolId::Beacon,
+            cache: BeaconStorageCache::new(),
+        };
+
+        // Report current storage capacity.
```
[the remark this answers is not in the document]
ogenev: Yes, you are right, those metrics report the whole storage because we will need also to refactor  `StorageMetricsReporter` to support multiple networks.

I'm happy to leave this for now and create a new issue that will deal with supporting metrics for multiple networks, what do you think?

trin-beacon/src/storage.rs
```diff
@@ -0,0 +1,532 @@
...
+        value: &Vec<u8>,
+    ) -> Result<(), ContentStoreError> {
+        let content_id = key.content_id();
+        let content_key: Vec<u8> = key.clone().into();
+
+        match content_key.first() {
+            Some(&LIGHT_CLIENT_BOOTSTRAP_KEY_PREFIX) => {
+                // store content key w/o the 0x prefix
+                let content_key = hex_encode(content_key).trim_start_matches("0x").to_string();
+                if let Err(err) = self.db_insert(&content_id, &content_key, value) {
```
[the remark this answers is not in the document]
ogenev: No, because we don't have a "capacity" for the beacon network. Every node should store all of the beacon data (it is not too much) and we don't use radius and we don't leave an option for the user to decide how much mbs she wants to allocate to this network.

trin-beacon/src/storage.rs
```diff
@@ -0,0 +1,532 @@
...
+            BeaconContentKey::LightClientOptimisticUpdate(content_key) => {
+                match self.cache.get_optimistic_update(content_key.signature_slot) {
+                    Some(_) => Ok(ShouldWeStoreContent::AlreadyStored),
+                    None => Ok(ShouldWeStoreContent::Store),
+                }
+            }
+        }
+    }
+
+    /// The "radius: concept is not applicable for Beacon network, this is why we always return the
```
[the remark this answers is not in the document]
ogenev: This is a CoPilot typo, it wasn't me :smiley:

ogenev:
> 1. I'm still in favor of having federated SQLite dbs for each subnetwork. Though, if there are some good reasons not to go this route w/regard to r/w bottlenecks?

Usually, it is a good practice for one application to have only one database. This is because managing multiple databases can be hard and adds one more layer of complexity. The good thing about SQL is that we can have one database but many db tables inside and every `*Storage` structure can manage only the data and the tables that are related to the specific network. 

In our case, one benefit of a single DB is that we can calculate faster the total DB storage (because most of the data is in the same table).

> 2. What feels a bit off to me is that this structure (`BeaconStorage`/`HistoryStorage`/`StateStorage`/etc) implies that we do have federated SQLite DBS for each subnetwork. And it implies that each `XStorage` type only operates on data relevant to that subnetwork. When in fact, each `XStorage` type is operating on the same, single SQLite db, containing all data from every subnetwork (eg pruning data in `HistoryStorage` might remove a piece of `Beacon` content). This makes it tricky to reason about what each fn is doing. Is it operating only on subnetwork data or all data in the db together?

I'm not sure why having (`BeaconStorage`/`HistoryStorage`/`StateStorage`/etc) leaves the impression that we should have multiple databases. Those storage structs act as an abstraction and operate only on data that is related to the specific network.

For example, `BeaconStorage` is responsible for creating the `lc_updates` table and the beacon cache. When `BeaconStorage` operates on the main table, it only touches the beacon data by checking the network column.
`BeaconStorage` never touches data from other networks and `HistoryStorage` also doesn't operate on beacon data and the `lc_updates` table.

So in short, all storage structs operate on one database (because we have one) but they have different responsibilities and they manage different DB tables inside the same DB.

> 3. Edge cases. User runs trin w/ `history,beacon,state` activated. Then restarts trin with only `beacon` activated. Now we have a lot of state data in the db that's occupying space. If we used federated DBs, we could just trash the history DB.

This is a viable use case but we can still trash the history db with the current structure. If we have a DB per network, we will need to trash the whole history DB, in the current case we will purge only the history data from the main table inside the DB with a simple query like:

```sql
 "DELETE FROM content_data
  WHERE network = (?1)";
```

I understand that this may not be the perfect structure and we can improve it a lot on demand but for now, we need some direction to be able to iterate.

trin-beacon/src/storage.rs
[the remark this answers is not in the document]
njgheorghita: Well... isn't it possible that the pruning inside the `HistoryStorage` will prune beacon data? Which is not something we want to happen, if I'm understanding the situation correctly

trin-storage/src/sql.rs
```diff
@@ -0,0 +1,62 @@
...
+
+pub const CONTENT_KEY_LOOKUP_QUERY: &str =
+    "SELECT content_key FROM content_data WHERE content_id_long = (?1) LIMIT 1";
+
+pub const CONTENT_VALUE_LOOKUP_QUERY: &str =
+    "SELECT content_value FROM content_data WHERE content_id_long = (?1) LIMIT 1";
+
+pub const TOTAL_DATA_SIZE_QUERY: &str = "SELECT TOTAL(content_size) FROM content_data";
+
+pub const TOTAL_ENTRY_COUNT_QUERY: &str =
```
njgheorghita, line 41: Here's an example where I feel like naming could be improved. We have `TOTAL_DATA_SIZE_QUERY` which operates on all data in the db, and `TOTAL_ENTRY_COUNT_QUERY` which only operates on a subsection of data. From the `XStorage` level, it's impossible to distinguish b/w what is a network-specific/network-agnostic sql command without looking up the actual sql query. I think it would be beneficial if we introduce some kind of naming convention that easily distinguishes between those two types of sql queries....

Though, I'm struggling to come up with a good example :sweat_smile: - maybe just simple appending `DB` / `NETWORK` to the end of these would be sufficient? (at least for all the `content_data` table queries`)
eg...
- `TOTAL_ENTRY_COUNT_QUERY_NETWORK`
- `TOTAL_DATA_SIZE_QUERY_DB`
- `CONTENT_KEY_LOOKUP_QUERY_DB`
- `XOR_FIND_FARTHEST_QUERY_NETWORK`
- `PAGINATE_QUERY_DB` 
- etc...

Idk, if you feel like I'm being too nitpicky with this, I'm happy to back down. My sense is that if we're operating on a single db at the `XStorage` then it's pretty crucial to implicitly understand the effects of the sql queries when writing/reviewing/debugging `XStorage` level logic.

trin-storage/src/sql.rs
```diff
@@ -0,0 +1,62 @@
...
+    "SELECT COUNT(content_id_long) FROM content_data WHERE network = (?1)";
+
+pub const PAGINATE_QUERY: &str =
+    "SELECT content_key FROM content_data ORDER BY content_key LIMIT :limit OFFSET :offset";
+
+pub const CONTENT_SIZE_LOOKUP_QUERY: &str =
+    "SELECT content_size FROM content_data WHERE content_id_long = (?1)";
+
+pub const CREATE_LC_UPDATE_TABLE: &str = "CREATE TABLE IF NOT EXISTS lc_update (
```
njgheorghita, line 50: *nitpick* Shouldn't this be `LC_UPDATE_CREATE_TABLE` to be more consistent with the other `lc_update` sql query names?

trin-storage/src/utils.rs
```diff
@@ -0,0 +1,123 @@
...
+use ethportal_api::utils::bytes::{hex_decode, hex_encode};
+use r2d2::{Pool, PooledConnection};
+use r2d2_sqlite::SqliteConnectionManager;
+use rusqlite::params;
+use std::{fs, path::Path};
+use tracing::{debug, info};
+
+/// Helper function for opening a SQLite connection.
+pub fn setup_sql(node_data_dir: &Path) -> Result<Pool<SqliteConnectionManager>, ContentStoreError> {
+    let sql_path = node_data_dir.join("trin.sqlite");
```
njgheorghita, line 15: Feels like this db name should probably be a constant

trin-history/src/storage.rs
```diff
@@ -0,0 +1,818 @@
...
+    },
+    utils::{
+        byte_vector_to_u32, get_total_size_of_directory_in_bytes, insert_value,
+        lookup_content_value,
+    },
+    ContentId, ContentStore, DataSize, DistanceFunction, EntryCount, PortalStorageConfig,
+    ShouldWeStoreContent, BYTES_IN_MB_U64,
+};
+
+/// Struct whose public methods abstract away Kademlia-based store behavior.
```
njgheorghita, line 30: Also think this comment is out of date?

njgheorghita (commented):
> I'm not sure why having (BeaconStorage/HistoryStorage/StateStorage/etc) leaves the impression that we should have multiple databases. Those storage structs act as an abstraction and operate only on data that is related to the specific network.

I pointed out some comments that lead to this impression. Also, just the general structure that we have 3x subnetwork-specific `XStorage` leads to this impression. Without digging into the code, there's no signal that they operate on the same db. 

If we created some sort of parent, `StorageManager` struct, responsible for the db-wide operations (eg paginate, metrics, instantiation, etc.), and this housed `XStorage` level structs for dealing with per-network data, then that would help readers understand the structure more naturally imo. Though, this `StorageManager` concept I'm proposing isn't a blocker for this pr.

> When BeaconStorage operates on the main table, it only touches the beacon data by checking the network column.

Sure, but this is something you know b/c you're deeply involved in writing the code. My point is that it's not so easy to reach this conclusion without methodically looking up each sql query to see if they operate on networks specifically/generically. I think the naming / comments could be improved to help communicate this.

trin-storage/src/sql.rs
ogenev: We can indeed improve those, I will add the suffix DB/NETWORK for now.

Another way to abstract this, as you suggested, is to create some structs with adequate naming that will be responsible for the following:

- operates on the storage as a whole  - `StorageManager`
- operates on the network level of the storage `NetworkStorageManager`

Then we can add methods to those structs which will call the SQL queries. This way, any `*Storage` instances will not directly touch the SQL queries but will operate via the structs above and it will be more clear what is happening.

trin-beacon/src/storage.rs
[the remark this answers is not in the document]
ogenev: `HistoryStorage` prunes only history data because `find_farthest_content_id` uses `XOR_FIND_FARTHEST_QUERY` which requires a `network` param to look into the main db.

trin-beacon/src/storage.rs
```diff
@@ -0,0 +1,530 @@
...
+        self.optimistic_update = Some(optimistic_update);
+    }
+
+    /// Sets the light client finality update
+    pub fn set_finality_update(&mut self, finality_update: ForkVersionedLightClientFinalityUpdate) {
+        self.finality_update = Some(finality_update);
+    }
+}
+
+/// A data store for Beacon Network content.
```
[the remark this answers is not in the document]
ogenev: Updated some of those comments.

ogenev:
@njgheorghita It seems that the biggest confusion is understanding which queries operate on the DB as a whole and which queries operate only on specific network data.

I think if we introduce an abstraction over the SQL queries with good comments and documentation, we will make this clearer and easier to maintain. (in a future PR).

For example we can have something like: `SqlManager` struct  with `DbSqlManager` and `NetworkSqlManager` members which will implement methods that touches the relevant SQL queries. Then we can include the `SqlManager` as part of every `xStorage` implementation.

What do you think?

PR #1091, opened 2024-01-11
feat(bridge): add (semaphore to limit gossips) + (timeout on gossips) for history backfill

KolbyML:
### What was wrong?
this is a new solution for https://github.com/ethereum/trin/pull/1087 me and nick jumped on a call and disscussed it and decided for each of us to take stabs at what we think the best solution is and compare them side to side and merge the best one.
### How was it fixed?

- By adding semaphore's to limit the amount of gossiping at a time.
- adding a timeout on serve_full_block() in case a timeout bug happens

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -49,7 +49,7 @@ const HEADER_SATURATION_DELAY: u64 = 10; // seconds
 const LATEST_BLOCK_POLL_RATE: u64 = 5; // seconds
 const EPOCH_SIZE: u64 = EPOCH_SIZE_USIZE as u64;
 const GOSSIP_LIMIT: usize = 32;
-const GOSSIP_TIMEOUT: u64 = 10 * 60; // seconds
+const SERVE_BLOCK_TIMEOUT: u64 = 120; // seconds
 
 pub struct HistoryBridge {
     pub mode: BridgeMode,
@@ -135,9 +135,18 @@ impl HistoryBridge {
                     let portal_clients = self.portal_clients.clone();
                     let execution_api = self.execution_api.clone();
                     tokio::spawn(async move {
-                        let _ = Self::serve_full_block(height, None, portal_clients, execution_api)
-                            .in_current_span()
-                            .await;
+                        if (timeout(
+                            Duration::from_secs(SERVE_BLOCK_TIMEOUT),
+                            Self::serve_full_block(height, None, portal_clients, execution_api)
+                                .in_current_span(),
+                        )
+                        .await)
+                            .is_err()
+                        {
+                            error!(
+                                "serve_full_block() timed out: this is an indication a bug is present"
+                            )
+                        };
                     });
                 }
                 block_index = gossip_range.end;
@@ -215,7 +224,7 @@ impl HistoryBridge {
                 };
                 tokio::spawn(async move {
                     if (timeout(
-                        Duration::from_secs(GOSSIP_TIMEOUT),
+                        Duration::from_secs(SERVE_BLOCK_TIMEOUT),
                         Self::serve_full_block(height, epoch_acc, portal_clients, execution_api)
                             .in_current_span(),
                     )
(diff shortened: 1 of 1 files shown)


njgheorghita (commented):
But we still have an issue here of overwhelming the pandaops providers? This is happening to me locally (in backfill mode), since currently on `master` we spawn an entire epoch's worth of pandaops requests instantly.

KolbyML:
@njgheorghita This PR would overwelm Pandaops just as must as the future's solution.

To not overwelm Pandaops we would need to ratelimit the surf client.

This can only be done after this https://github.com/ethereum/trin/pull/1089 is merged and is also not related to this PR at all so I am not understanding something here. That would have to come after https://github.com/ethereum/trin/pull/1089 is merged in a new PR

KolbyML:
@njgheorghita this should be ready to review now

portal-bridge/src/bridge/history.rs
```diff
@@ -200,11 +215,27 @@ impl HistoryBridge {
                 let epoch_acc = epoch_acc.clone();
                 let portal_clients = self.portal_clients.clone();
                 let execution_api = self.execution_api.clone();
+                let permit = match gossip_send_semaphore.clone().acquire_owned().await {
```
njgheorghita, line 218: If `"this should be impossible"` as you state in the error message, then I'd probably go with an `.expect("..")`

portal-bridge/src/bridge/history.rs
```diff
@@ -130,9 +135,18 @@ impl HistoryBridge {
                     let portal_clients = self.portal_clients.clone();
                     let execution_api = self.execution_api.clone();
                     tokio::spawn(async move {
-                        let _ = Self::serve_full_block(height, None, portal_clients, execution_api)
-                            .in_current_span()
-                            .await;
+                        if (timeout(
+                            Duration::from_secs(SERVE_BLOCK_TIMEOUT),
```
njgheorghita, line 139: You should be able to update the constant to
```rust
const SERVE_BLOCK_TIMEOUT: Duration = Duration::from_secs(120);
```

portal-bridge/src/bridge/history.rs
```diff
@@ -180,6 +194,7 @@ impl HistoryBridge {
             start: start_block,
             end: end_block,
         };
+        let gossip_send_semaphore = Arc::new(Semaphore::new(GOSSIP_LIMIT));
```
njgheorghita, line 197: I'd leave a comment here explaining why we're using a semaphore for backfill and the purpose of the `GOSSIP_LIMIT`

portal-bridge/src/bridge/history.rs
```diff
@@ -200,11 +215,27 @@ impl HistoryBridge {
...
+                let permit = match gossip_send_semaphore.clone().acquire_owned().await {
+                    Ok(permit) => permit,
+                    Err(_) => {
+                        error!("acquire_owned() can only error on semaphore close, this should be impossible");
+                        continue;
+                    }
+                };
                 tokio::spawn(async move {
-                    let _ =
+                    if (timeout(
```
njgheorghita, line 226: I'm not sure if I'm wrong here, but it seems like there is some duplicated logic here & in the other location where we spawn the thread in `"latest"`. Seems like we should be able to move some of this logic to `Self::serve_full_block()` since we're doing the same thing in each location. It looks like `Self::serve_full_block()` could take care of everything from spawning the thread and everything inside the spawn (eg error handling & dropping the permit)

KolbyML: We can't move the logic to ``Self::serve_full_block()`` because timeout() requires we pass in a future. So I created a non-async function called ``spawn_serve_full_block`` which I believe accomplishes the refactor you were looking for.

KolbyML:
@njgheorghita I believe I resolve the PR concerns so it should be ready for another look

portal-bridge/src/bridge/history.rs
njgheorghita: small *nitpick* .. `to limit the amount of active gossip transfers to make sure we don't overwhelm the trin client`

njgheorghita (approved):
:ship:

Issue #1047, opened 2023-11-29
Migrate from `eyre` to `anyhow` in `light-client` crate


ogenev:
To keep consistency amongst our code base, we want to use the same error handling library which in our case is [anyhow](https://crates.io/crates/anyhow). 

However, in the `light-client` crate [we use](https://github.com/ethereum/trin/blob/5eadae656ae3b18c5bff7bdf366088d1af6c347e/light-client/Cargo.toml#L17) the [eyre](https://crates.io/crates/eyre) library. It will be nice if we migrate from `eyre` to `anyhow` to keep consistency.

qiweiii:
I could try to take this

ogenev:
> I could try to take this

sure!

PR #1092, opened 2024-01-11
feat: migrate from eyre to anyhow

qiweiii:
### What was wrong?

fix: https://github.com/ethereum/trin/issues/1047

### How was it fixed?

migrate eyre to anyhow equivalents

### To-Do

### DIFF
--- a/light-client/src/client.rs
+++ b/light-client/src/client.rs
@@ -317,7 +317,7 @@ impl<DB: Database, R: ConsensusRpc + 'static> Client<DB, R> {
                 NodeError::ConsensusSyncError(err) => match err
                     .downcast_ref()
                     .ok_or(err.to_string())
-                    .map_err(|err| anyhow::anyhow!(err))?
+                    .map_err(|err| anyhow!(err))?
                 {
                     ConsensusError::CheckpointTooOld => {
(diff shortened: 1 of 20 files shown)


light-client/src/client.rs
```diff
@@ -317,7 +317,7 @@ impl<DB: Database, R: ConsensusRpc + 'static> Client<DB, R> {
                 NodeError::ConsensusSyncError(err) => match err
                     .downcast_ref()
                     .ok_or(err.to_string())
-                    .map_err(|err| eyre!(err))?
+                    .map_err(|err| anyhow::anyhow!(err))?
```
ogenev, line 320: nitpick: for consistency, I would prefer to use directly `anyhow!`, instead of `anyhow::anyhow!`  in this file and in [light-client/src/config/checkpoints.rs](https://github.com/ethereum/trin/pull/1092/files#diff-ca43eb11df384433c0830c3335fa7b3ddca950da5a8067f9ac20d771d9a0bc46):

```suggestion
                    .map_err(|err| anyhow!(err))?
```

ogenev (approved):
Thanks, looks good :+1: . 

Could you rebase before merging this?

PR #1093, opened 2024-01-11
feat: add support for configuring network protocol_id's

KolbyML:
### What was wrong?
We aren't able to configure protocol_id in the case we want to support different networks like a testnet
### How was it fixed?
This is basically the same thing we already did except we use a bidirectional map instead of trait conversions acting as a bidirectional map.

For storing mainnet/future network data we will use a folder structure I seen L1 clients use ``trin/mainnet/...`` ``trin/futurenetworkname/...``

Now we can easily configure future networks

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -315,7 +315,9 @@ fn network_parser(network_string: &str) -> Result<Arc<NetworkSpec>, String> {
     match network_string {
         "mainnet" => Ok(MAINNET.clone()),
         "testnet" => Ok(TESTNET.clone()),
-        _ => Err(format!("Not a valid network: {network_string}")),
+        _ => Err(format!(
+            "Not a valid network: {network_string}, must be 'testnet' or 'mainnet'"
+        )),
     }
 }
 
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -4,7 +4,7 @@ use clap::{
     error::{Error, ErrorKind},
     Args, Parser, Subcommand,
 };
-use std::{env, ffi::OsString, fmt, net::SocketAddr, path::PathBuf, str::FromStr};
+use std::{env, ffi::OsString, fmt, net::SocketAddr, path::PathBuf, str::FromStr, sync::Arc};
 use trin_utils::build_info;
 use url::Url;
 
@@ -21,10 +21,13 @@ pub const BEACON_NETWORK: &str = "beacon";
 pub const HISTORY_NETWORK: &str = "history";
 pub const STATE_NETWORK: &str = "state";
 const DEFAULT_SUBNETWORKS: &str = "history";
+pub const DEFAULT_NETWORK: &str = "mainnet";
 pub const DEFAULT_STORAGE_CAPACITY_MB: &str = "100";
 pub const DEFAULT_WEB3_TRANSPORT: &str = "ipc";
 
 use crate::dashboard::grafana::{GrafanaAPI, DASHBOARD_TEMPLATES};
+
+use super::portal_wire::{NetworkSpec, MAINNET, TESTNET};
 
 #[derive(Debug, PartialEq, Clone)]
 pub enum Web3TransportType {
@@ -147,12 +150,20 @@ pub struct TrinConfig {
     pub trusted_block_root: Option<String>,
 
     #[arg(
-    long = "networks",
+    long = "portal-subnetworks",
         help = "Comma-separated list of which portal subnetworks to activate",
         default_value = DEFAULT_SUBNETWORKS,
         use_value_delimiter = true
     )]
-    pub networks: Vec<String>,
+    pub portal_subnetworks: Vec<String>,
+
+    #[arg(
+        long = "network",
+            help = "Choose mainnet or testnet",
+            default_value = DEFAULT_NETWORK,
+            value_parser = network_parser
+        )]
+    pub network: Arc<NetworkSpec>,
 
     /// Storage capacity specified in megabytes.
     #[arg(
@@ -225,7 +236,7 @@ impl Default for TrinConfig {
             no_upnp: false,
             private_key: None,
             trusted_block_root: None,
-            networks: DEFAULT_SUBNETWORKS
+            portal_subnetworks: DEFAULT_SUBNETWORKS
                 .split(',')
                 .map(|n| n.to_string())
                 .collect(),
@@ -240,6 +251,7 @@ impl Default for TrinConfig {
             ws_port: DEFAULT_WEB3_WS_PORT,
             command: None,
             utp_transfer_limit: DEFAULT_UTP_TRANSFER_LIMIT,
+            network: MAINNET.clone(),
         }
     }
 }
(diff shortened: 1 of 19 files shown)


njgheorghita (commented):
I appreciate the pr & work towards this direction. But imo we shouldn't merge any testnet implementation until https://github.com/ethereum/portal-network-specs/pull/258 is approved by all client teams & merged

KolbyML:
@njgheorghita I removed the 10 lines relating to a testnet you can review my code now

KolbyML:
By what is said between you and piper this is in a deadlock now. This PR more has to do with configuring protocol ID's and I don't see how that a PR formalizing protocol_ids for a testnet blocks a pr allowing the option to configure protocol ID's.

Being able to configure network parameters is a standard feature in all L1 clients. I can comprehend how we wouldn't want to formalize the testnet protocol_ids in our code, but that is 15 lines of 200. This code could have been reviewed and we could have just removed the 15 line testnet part instead of a blanket ignore because the word testnet was mentioned.

Shouldn't by the same logic we should have blocked TraceGossip till it was reviewed into the spec and reviewed by all client teams? I don't see why 95% (configure protocol id's) of this PR is being blocked for 5% (15 lines of testnet parameters)

pipermerriam:
I wouldn't categorize my opinion as "strong".  Only that since we haven't done "testnet" yet my intuition is that actually doing the implementation will likely uncovers somes "unknown unknown" types of learning... but feel free to proceed however you guys feel is best.

I think the only thing I have a "strong" opinion on is that we **should not** write the spec and then blindly implement the clients to follow the spec.  Lets make sure that client implementation learnings are actively used to influence the specification because things don't always unfold the way we expect them to.

njgheorghita (commented):
I think kurtosis could be useful in testing this.
Something like...
- Spin up 3 trin nodes running testnet
- Spin up 3 trin nodes running mainnet
- Send a ping between the testnet & mainnet clients
- Spin up a bridge gossiping testnet data (eg epoch # 1)
- Spin up a bridge gossiping mainnet data (eg epoch # 2)

Test that the mainnet nodes only store epoch # 2 data, and testnet nodes only have epoch # 1 data.

KolbyML:
@njgheorghita @ogenev ready for a review

I added the small peertest I talked about in the call and comment above

ethportal-api/src/types/cli.rs
```diff
@@ -299,6 +311,14 @@ pub fn check_private_key_length(private_key: &str) -> Result<B256, String> {
     ))
 }
 
+fn network_parser(network_string: &str) -> Result<Arc<NetworkSpec>, String> {
+    match network_string {
+        "mainnet" => Ok(MAINNET.clone()),
+        "testnet" => Ok(TESTNET.clone()),
+        _ => Err(format!("Not a valid network: {network_string}")),
```
njgheorghita, line 318: `"Not a valid network: {network_string}, must be 'testnet' or 'mainnet'"`

ethportal-api/src/types/portal_wire.rs
```diff
@@ -174,22 +176,41 @@ pub enum ProtocolId {
...
+        network_name: "mainnet".to_string(),
+    }
+    .into()
+});
+
+pub static TESTNET: Lazy<Arc<NetworkSpec>> = Lazy::new(|| {
+    let mut portal_networks = BiHashMap::new();
+    portal_networks.insert(ProtocolId::State, "0x501A".to_string());
+    portal_networks.insert(ProtocolId::History, "0x501B".to_string());
+    portal_networks.insert(ProtocolId::TransactionGossip, "0x501C".to_string());
```
njgheorghita, line 204: I don't see any reason to include `TransactionGossip` & `CanonicalIndices`?

ethportal-api/src/types/portal_wire.rs
```diff
@@ -590,21 +611,20 @@ impl From<Accept> for Value {
...
-        ProtocolId::from_str(hex).unwrap();
+        assert!(!MAINNET.portal_networks.contains_right(hex));
     }
 
     #[test]
     fn protocol_id_encoding() {
         let hex = "0x500A";
-        let protocol_id = ProtocolId::from_str(hex).unwrap();
-        let expected_hex = hex_encode_upper(Vec::try_from(protocol_id).unwrap());
+        let protocol_id = MAINNET.portal_networks.get_by_right(hex).unwrap();
```
njgheorghita, line 626: I'm not a huge fan of using the native `BiHashMap` api. It's not obvious (without looking up the map &/or `BiHashMap` source code), what these functions do (aka left vs. right)... Could you `impl` something more intuitive, like `get_protocol_id` / `get_protocol_id_hex` (or whatever fn name makes the most sense here)?

portalnet/src/events.rs
```diff
@@ -118,10 +122,13 @@ impl PortalnetEvents {
 
     /// Dispatch Discv5 TalkRequest event to overlay networks or uTP socket
     fn dispatch_discv5_talk_req(&self, request: TalkRequest) {
-        let protocol_id = ProtocolId::from_str(&hex_encode_upper(request.protocol()));
+        let protocol_id = self
+            .network_spec
+            .portal_networks
+            .get_by_right(&hex_encode_upper(request.protocol()));
```
njgheorghita, line 128: Same here. Some custom `impl` that has a more intuitive name would be better

src/lib.rs
```diff
@@ -100,7 +111,11 @@ pub async fn run_trin(
         beacon_event_tx,
         beacon_jsonrpc_tx,
         beacon_event_stream,
-    ) = if trin_config.networks.iter().any(|val| val == BEACON_NETWORK) {
+    ) = if trin_config
+        .portal_subnetworks
```
njgheorghita, line 115: Think we could also do a custom `trin_config.portal_subnetwork.contains()` or something similar

tests/self_peertest.rs
```diff
@@ -206,13 +206,22 @@ async fn peertest_state_offer_contract_bytecode() {
 #[tokio::test(flavor = "multi_thread")]
 #[serial]
 async fn peertest_state_recursive_gossip() {
-    let (peertest, target, handle) = setup_peertest().await;
+    let (peertest, target, handle) = setup_peertest(None).await;
```
njgheorghita, line 209: I think it would be a bit better to get rid of the `Option` and just use `"mainnet"` as the arg here

njgheorghita (approved):
Looks pretty good, just a few comments on fn naming. But, I'd also like to hold off on merging this until after the db update is deployed, since this will require some changes to the cluster repo, and it'd be better to do the 2 somewhat significant changes in separate goes to isolate their effects/bugs from each other

KolbyML:
> since this will require some changes to the cluster repo,

This PR requires 0 changes to the cluster repo, to my knowledge? Would it be possible to provide an explanation on how that conclusion was achieved? Maybe I am misunderstanding something.

ethportal-api/src/types/portal_wire.rs
KolbyML: I mean it is included in our mainnet code

KolbyML:
ready for another look

njgheorghita:
Well, it's not a major change, but we'll need to update the `--networks` flag to `--portal-subnetworks`. Anyways, I guess it's not a huge blocker for this pr, just something that we'll make sure we want to do in parallel. And if we forget to do it, the nodes will crash, so the problem will be pretty obvious. So, I guess nvm let's go ahead with the merge here.

PR #1095, opened 2024-01-14
fix: remove `0x` prefix from trin node data directory

### LINKED ISSUE #941: Trin's data directory hash includes `0x`
A Trin should store data in a directory formatted as `.../trin_aabbccdd` where `aabbccdd` is the prefix of the node's public key. Instead, Trin is currently storing in the directory `.../trin_0xaabbcc`. There should be no `0x` prefix.

[Here is where this happens](https://github.com/ethereum/trin/blob/5b385c9bdd8bc1325d81ac2eb4da26a208846267/portalnet/src/utils/db.rs#L64)

sseyren:
### What was wrong?

Fixes #941 

### How was it fixed?

Used `hex::encode` directly to encode node id.

### To-Do

### DIFF
--- a/portalnet/src/utils/db.rs
+++ b/portalnet/src/utils/db.rs
@@ -66,7 +66,7 @@ pub fn configure_node_data_dir(
 fn get_node_data_dir(trin_data_dir: PathBuf, node_id: NodeId) -> PathBuf {
     // Append first 8 characters of Node ID
     let mut application_string = "trin_".to_owned();
-    let node_id_string = hex_encode(node_id.raw());
+    let node_id_string = hex::encode(node_id.raw());
     let suffix = &node_id_string[..8];
     application_string.push_str(suffix);
     trin_data_dir.join(application_string)
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4429,6 +4429,7 @@ dependencies = [
  "ethportal-api",
  "fnv",
  "futures 0.3.30",
+ "hex",
  "igd-next",
  "lazy_static",
  "leb128",
--- a/portalnet/Cargo.toml
+++ b/portalnet/Cargo.toml
@@ -24,6 +24,7 @@ ethereum_ssz_derive = "0.5.3"
 ethportal-api = { path = "../ethportal-api" }
 fnv = "1.0.7"
 futures = "0.3.21"
+hex = "0.4.3"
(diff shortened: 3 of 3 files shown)


KolbyML:
We are going to be resetting all of the nodes on the network on the next deployment so we don't need the legacy folder conversion code. But that is pretty clever. Other then that the PR looks good to merge

portalnet/src/utils/db.rs
```diff
@@ -57,15 +57,38 @@ pub fn configure_node_data_dir(
         None => get_application_private_key(&trin_data_dir)?,
     };
     let node_id = Enr::empty(&pk)?.node_id();
-    let node_data_dir = get_node_data_dir(trin_data_dir, node_id);
-    fs::create_dir_all(&node_data_dir)?;
+
+    let node_data_dir = get_node_data_dir(&trin_data_dir, node_id);
+    let node_legacy_data_dir = get_legacy_node_data_dir(&trin_data_dir, node_id);
+    if node_legacy_data_dir.is_dir() && !node_data_dir.is_dir() {
```
ogenev, line 63: Ас @KolbyML added we will trash the whole database on the next deployment so there is no need for this legacy migration.

sseyren:
I updated my PR.

ogenev (approved):
Thanks!

portalnet/src/utils/db.rs
```diff
@@ -66,7 +66,7 @@ pub fn configure_node_data_dir(
 fn get_node_data_dir(trin_data_dir: PathBuf, node_id: NodeId) -> PathBuf {
     // Append first 8 characters of Node ID
     let mut application_string = "trin_".to_owned();
-    let node_id_string = hex_encode(node_id.raw());
+    let node_id_string = hex::encode(node_id.raw());
```
njgheorghita, line 69: Sorry, I'm late to this review. But it seems like we can avoid an unnecessary library import.. `hex_encode` is an internal utility fn that always returns a `0x` prefixed string. I believe there are multiple possible solutions to removing the `0x` prefix that don't require importing the `hex` library.
eg. `node_id_string.trim_start_matches("0x")` or even something like `let suffix = &node_id_string[2..8]` I believe would work. This is just a suggestion, but imo, if we can avoid unnecessary library imports then it's worth considering.

PR #1096, opened 2024-01-15
feat(history bridge): add backfill by block range

KolbyML:
### What was wrong?
Nick said he wanted backfill by block range for making his kurtosis tests. I already wrote most of the code for it in a previous PR so  here is the PR with that.

If this PR doesn't suit what you imagined Nick just close it and open a new one no biggy. I am not too picky about this, just thought it could save you some time which is why I mentioned it. If this PR is 90% percent the way there of course I would be happy to get this reviewed and merged.
### How was it fixed?
This includes the block range change + a quick refactor I did to make it so backfill used block ranges as a whole. This simplifes the code a lot and removes a whole while loop which is code. Functionally both of these will perform the same. Without this refactor adding the backfill via block range code felt awkward.

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -150,8 +150,8 @@ impl HistoryBridge {
             "Error launching bridge in backfill mode. Unable to get latest block from provider.",
         );
         let (is_single_mode, mode_type) = match self.mode.clone() {
-            BridgeMode::Backfill(val) => (true, val),
-            BridgeMode::Single(val) => (false, val),
+            BridgeMode::Backfill(val) => (false, val),
+            BridgeMode::Single(val) => (true, val),
             _ => panic!("Invalid backfill mode"),
         };
         let (start_block, mut end_block) = match mode_type {
@@ -240,7 +240,7 @@ impl HistoryBridge {
             .await)
                 .is_err()
             {
-                error!("serve_full_block() timed out: this is an indication a bug is present")
+                error!("serve_full_block() timed out on height {height}: this is an indication a bug is present")
             };
             if let Some(permit) = permit {
                 drop(permit);
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -149,23 +149,23 @@ impl HistoryBridge {
         let latest_block = self.execution_api.get_latest_block_number().await.expect(
             "Error launching bridge in backfill mode. Unable to get latest block from provider.",
         );
-        let (looped, mode_type) = match self.mode.clone() {
+        let (is_single_mode, mode_type) = match self.mode.clone() {
             BridgeMode::Backfill(val) => (true, val),
             BridgeMode::Single(val) => (false, val),
             _ => panic!("Invalid backfill mode"),
         };
         let (start_block, mut end_block) = match mode_type {
             ModeType::Epoch(epoch_number) => {
-                let end_block = match looped {
-                    true => latest_block,
-                    false => (epoch_number + 1) * EPOCH_SIZE,
+                let end_block = match is_single_mode {
+                    true => (epoch_number + 1) * EPOCH_SIZE,
+                    false => latest_block,
                 };
                 (epoch_number * EPOCH_SIZE, end_block)
             }
             ModeType::Block(block) => {
-                let end_block = match looped {
-                    true => latest_block,
-                    false => block + 1,
+                let end_block = match is_single_mode {
+                    true => block + 1,
+                    false => latest_block,
                 };
                 (block, end_block)
             }
@@ -181,7 +181,9 @@ impl HistoryBridge {
                         Please specify a starting block/epoch that begins before the current block."
             );
         }
-        let mut current_epoch_index = start_block / EPOCH_SIZE;
+        // initialize current_epoch_index as an impossible value u64::MAX so that
+        // epoch_acc gets set on the first iteration of the loop
+        let mut current_epoch_index = u64::MAX;
         let gossip_range = Range {
             start: start_block,
             end: end_block,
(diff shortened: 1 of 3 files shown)


KolbyML:
@njgheorghita I fixed the bug we talked about in the call and it works

portal-bridge/src/types/mode.rs
```diff
@@ -82,6 +86,28 @@ impl FromStr for ModeType {
                     .map_err(|_| "Invalid bridge mode arg: block number")?;
                 Ok(ModeType::Block(block))
             }
+            "r" => {
+                let range_vec = s[1..]
+                    .split('-')
+                    .map(|x| {
+                        x.parse()
+                            .expect("Invalid bridge mode arg: range format invalid expected (x:y)")
```
njgheorghita, line 94: I don't understand what you mean by `expected (x:y)` ... The correct format for range is `backfill:r10-100`. So user seeing this message might be confused and think that `10:100` is a valid range selector.
If this error message is displaying, then the user probably already knows about the `backfill:r` prefix, and just messed up formatting the range selector, but in this case I think it's better to be explicit and just restate the entire format, aka `range format invalid, expected backfill:rX-Y (eg. backfill:r10-100)`

portal-bridge/src/types/mode.rs
```diff
@@ -55,13 +57,15 @@ impl FromStr for BridgeMode {
 pub enum ModeType {
     Epoch(u64),
     Block(u64),
+    BlockRange(u64, u64),
 }
 
 impl ModeType {
```
njgheorghita, line 63: I think we can get rid of this function entirely

portal-bridge/src/types/mode.rs
```diff
@@ -8,6 +8,8 @@ use trin_validation::constants::EPOCH_SIZE;
 ///   - ex: "e123" starts at epoch 123
 /// - Single: executes a single block
 ///   - ex: "b123" executes block 123
+/// - BlockRange: generates test data from block x to y
+///   - ex: "r10-12" starts from block 10 and goes till 12
```
njgheorghita, line 12: Same nitpick about `till`, and I do think it's useful to be explicit about whether or not the range is inclusive in these docstrings

portal-bridge/src/bridge/history.rs
```diff
@@ -149,21 +149,22 @@ impl HistoryBridge {
...
+            ModeType::Epoch(epoch_number) => {
+                let end_block = match looped {
+                    true => latest_block,
+                    false => (epoch_number + 1) * EPOCH_SIZE,
+                };
+                (epoch_number * EPOCH_SIZE, end_block)
+            }
             ModeType::Block(block) => {
-                let epoch_index = block / EPOCH_SIZE;
                 let end_block = match looped {
```
njgheorghita, line 161: We're not using `looped` anymore. At least in the context of looping over gossip ranges. We should get rid of the variable, or rename it if it's still useful for setting other variable values

portal-bridge/src/bridge/history.rs
```diff
@@ -175,49 +176,38 @@ impl HistoryBridge {
...
+
+        info!("fetching headers in range: {gossip_range:?}");
+        let mut epoch_acc = None;
+        for height in gossip_range {
             // Using epoch_size chunks & epoch boundaries ensures that every
             // "chunk" shares an epoch accumulator avoiding the need to
             // look up the epoch acc on a header by header basis
-            let epoch_acc = if gossip_range.end <= MERGE_BLOCK_NUMBER {
-                match self.get_epoch_acc(epoch_index).await {
+            if height <= MERGE_BLOCK_NUMBER && current_epoch_index != height / EPOCH_SIZE {
```
njgheorghita, line 191: I don't think this is correct. At least on the first pass through. Let's say `current_epoch_index = 4`. On the first loop, we won't set the `epoch_acc` correctly, since the second clause here is not `true`. I don't think the `epoch_acc` will be set correctly for any blocks in the first epoch, but once we hit the next epoch, and the second clause evaluates to `true` then it should work

KolbyML: Good catch!

A solution for this I just pushed was to
```
        // initialize current_epoch_index as an impossible value u64::MAX so that
        // epoch_acc gets set on the first iteration of the loop
        let mut current_epoch_index = u64::MAX;
```

Set current_epoch_index to MAX as that should be an impossible value and now ``epoch_acc`` will be properly set on the first iteration of the loop

KolbyML:
@njgheorghita rebased ready for another review

portal-bridge/src/bridge/history.rs
```diff
@@ -149,26 +149,27 @@ impl HistoryBridge {
...
-                    true => (epoch_index + 1) * EPOCH_SIZE,
-                    false => block + 1,
+                let end_block = match is_single_mode {
+                    true => block + 1,
+                    false => latest_block,
                 };
-                (block, end_block, epoch_index)
+                (block, end_block)
             }
+            ModeType::BlockRange(start_block, end_block) => (start_block, end_block + 1),
```
njgheorghita, line 172: Tbh, I think it would be a worthwhile improvement to extract this logic into a method on `BridgeMode`. Looks like the only external variable needed is `latest_block` which could be easily passed in. It would also be good for sanity's sake to add some unit testing to this logic. If you want to address it in a following pr, that's fine by me.

njgheorghita (approved):
Nice! Just want to confirm that you've tested this? aka replace `Self::spawn_full_block()` with something like a `println!("{height}")` statement. And run it from the cli to confirm that the expected blocks are being gossiped for all of the various `backfill` / `single` options. Also one thing to keep an eye on is to add some `println!` statements to confirm that the epoch acc is being correctly set in each scenario. After confirming that those work, then let's go ahead with the :ship:

portal-bridge/src/bridge/history.rs
KolbyML: I think it should go in a follow up PR yeah I made a issue for it to track it https://github.com/ethereum/trin/issues/1100

KolbyML:
I tested it quick and found I mixed up a true/false so it is a good thing I verified that. I made a issue for the follow up PR going to merge this

PR #1098, opened 2024-01-16
fix(bridge): use buffer unordered api for history futures

njgheorghita:
### What was wrong?
My solution for the problem described in https://github.com/ethereum/trin/pull/1091

### How was it fixed?
- Reverted to using the `buffer_unordered` api to manage futures as opposed to spawning tokio threads
- Added a timeout to futures, so a hanging future doesn't block

### To-Do

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -6,9 +6,10 @@ use std::{
 };
 
 use anyhow::{anyhow, bail};
+use futures::stream::StreamExt;
 use ssz::Decode;
-use tokio::time::{sleep, Duration};
-use tracing::{debug, info, warn, Instrument};
+use tokio::time::{sleep, timeout, Duration};
+use tracing::{debug, info, warn};
 
 use crate::{
     api::execution::ExecutionApi,
@@ -45,6 +46,8 @@ use trin_validation::{
 const HEADER_SATURATION_DELAY: u64 = 10; // seconds
 const LATEST_BLOCK_POLL_RATE: u64 = 5; // seconds
 const EPOCH_SIZE: u64 = EPOCH_SIZE_USIZE as u64;
+const FUTURES_BUFFER_SIZE: usize = 32;
+const SERVE_BLOCK_TIMEOUT: u64 = 60; // seconds
 
 pub struct HistoryBridge {
     pub mode: BridgeMode,
@@ -126,15 +129,9 @@ impl HistoryBridge {
                     end: latest_block + 1,
                 };
                 info!("Discovered new blocks to gossip: {gossip_range:?}");
-                for height in gossip_range.clone() {
-                    let portal_clients = self.portal_clients.clone();
-                    let execution_api = self.execution_api.clone();
-                    tokio::spawn(async move {
-                        let _ = Self::serve_full_block(height, None, portal_clients, execution_api)
-                            .in_current_span()
-                            .await;
-                    });
-                }
+                if let Err(msg) = self.serve(gossip_range.clone(), None).await {
+                    warn!("Error serving latest blocks: {msg:?}");
+                };
                 block_index = gossip_range.end;
             }
         }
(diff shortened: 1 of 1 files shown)


njgheorghita:
@KolbyML Here's what I had in mind as a solution to the "hanging futures blocking the gossip loop" problem. 

Basically, we have two outcomes we want to prevent.
1. Overwhelming pandaops providers with too many concurrent http requests
2. Overwhelming trin with too many concurrent jsonrpc requests

If I understand the situation correctly, your intention is to solve `#1` by adding a governer to surf to limit http requests. And solve `#2` with the semaphore solution.

**copying over some quotes from https://github.com/ethereum/trin/pull/1091#issuecomment-1889834742 **
> The only solution to prevent overwelming pandaops providers is by directly rate limiting the requests to pandaops

I'm not sure I agree with this claim. I think it's very much possible to avoid overwhelming pandaops providers by putting a cap on the number of concurrent "serve" tasks. eg. in the way that I use the `buffer_unordered` api in this pr. This approach also has the added effect of not overwhelming the trin node with too many concurrent jsonrpc requests. 

We have two different approaches here. My main concern with your approach is that it uses two configurable values (eg. the rate limit on the surf governer & the semaphore limit). If we have two configurable values, then it will be harder to identify the cause and effect of each value, as they'll influence each other with each change. My approach uses a single configurable value (`FUTURES_BUFFER_SIZE`). It's much simpler to deal with a single configurable value than two. 

Ultimately, imo a single configurable value is better than two. But lmk what your thoughts are, if you agree, disagree, etc.

KolbyML:
> I'm not sure I agree with this claim. I think it's very much possible to avoid overwhelming pandaops providers by putting a cap on the number of concurrent "serve" tasks. eg. in the way that I use the buffer_unordered api in this pr. This approach also has the added effect of not overwhelming the trin node with too many concurrent jsonrpc requests.

https://github.com/ethereum/trin/pull/1091 the semaphore based PR does exactly what this PR claims to do. So this approach #1098 doesn't have an "added effect"

It does have the added downside of if a timeout occurs on latest we block new latest gossips for over a minute which isn't acceptable.

So since #1098 and #1091 are the same except with #1098 having the issues described above with latest.

I am fine if you want to do your ``futures`` approach for backfill, but it isn't acceptable for latest. So we will need to do the ``thread:spawn`` approach for that either way

I don't think you should conflate 2 different PR's (surf_govener and semaphore's) as one approach when #1098 and #1091 almost function identically with the downside listed above about #1098 

>  If we have two configurable values, then it will be harder to identify the cause and effect of each value, as they'll influence each other with each change.

Lets say someone is using an infura free account. They have a limited amount of requests they can make a day. surf_govener is the only solution which resolves this problem.

If you only have 1 constraining factor well then you would only have to set configuration option. Hobbists are the only ones going to be running bridges not your average Trin user. We are not marketing non-technical people to run bridges and if we are they wouldn't be using commandline options anyways they would be using safe defaults. If you are overloading el/cl then you know  you have to lower surf_governer. If trin is crashing then you know you have to lower active gossips.

> Ultimately, imo a single configurable value is better than two. But lmk what your thoughts are, if you agree, disagree, etc.
Who is going to be running a bridge?

It is important to keep in mind who would use these command line arguements. Providing more control isn't a bad thing. Simple doesn't mean something is better. Hobbists have there own constraints. And #1098 and #1091 isn't really a rate limit against EL/CL. If users are only able to make 1000 EL requests a day they should be able to configure that.

KolbyML:
latest should use a tokio::spawn approach without semaphores just a timeout to specifically address this.

njgheorghita:
Closing this after a zoom call with @KolbyML - ultimately, when it comes to `"latest"` the "thread-spawning" approach is better. When it comes to `"backfill"` there isn't a significant advantage to either approach. We agreed that remaining consistent in the mechanism used by each approach is worthwhile. Additionally, using the `surf_governer`, which directly rate limits http requests provides a better ux than an arbitrary `"MAX_FUTURES_BUFFER"` - which can only be used to indirectly rate limit http requests.

PR #1099, opened 2024-01-17
fix: light-client goerli source data broken

njgheorghita:
### What was wrong?
A `light-client` test recently started failing breaking our ci. I'm not sure I 100% grasp the problem, but believe I have all the important parts covered.

There seems to have been some issues with a fork on goerli yesterday (from ACD discord channel). It seems like the checkpoint fallback data providers we were using "probably" haven't made whatever fixes were necessary to track the head of the chain. Therefore they were not able to find a "finalized block_root". 

### How was it fixed?
So... the data providers we were using could not track the head of the chain, but some of the other checkpoint providers did implement the changes necessary to track the head of the chain. The problem is that we are unable to deserialize the json from these providers. So the json responses (containing a finalized block root) were just being turned into errors. These providers return the slot number as a `String` and not `u64`, which was why were unable to deserialize these responses. The fix was to add the deser rule `str_or_u64`.

There was also a case where some of the endpoints had a trailing `/` - so when we constructed the url, we would have endpoints containing `//`. 

Typically, it's not a good idea to have unit tests that rely on external sources of data. I would definitely be in favor of removing this dependency if it's at all possible?

### To-Do

### DIFF
--- a/light-client/src/config/checkpoints.rs
+++ b/light-client/src/config/checkpoints.rs
@@ -2,7 +2,7 @@ use std::collections::HashMap;
 
 use anyhow::anyhow;
 use ethereum_types::H256;
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 
 use crate::config::networks;
 
@@ -21,11 +21,29 @@ pub struct RawSlotResponseData {
 
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Slot {
+    #[serde(deserialize_with = "str_or_u64")]
     pub slot: u64,
     pub block_root: Option<H256>,
     pub state_root: Option<H256>,
     pub epoch: u64,
     pub time: StartEndTime,
+}
+
+fn str_or_u64<'de, D>(deserializer: D) -> Result<u64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum StrOrU64<'a> {
+        Str(&'a str),
+        U64(u64),
+    }
+
+    Ok(match StrOrU64::deserialize(deserializer)? {
+        StrOrU64::Str(v) => v.parse().unwrap_or(0), // Ignoring parsing errors
+        StrOrU64::U64(v) => v,
+    })
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Eq, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

KolbyML:
> Typically, it's not a good idea to have unit tests that rely on external sources of data. I would definitely be in favor of removing this dependency if it's at all possible?

I agree

njgheorghita:
Going to go ahead with this merge, since it has one approval and it's breaking the build on master. However, I'd like to hear from @ogenev if there's any possible way to remove these external dependencies?

KolbyML:
Yeah if there is a cleaner solution we can do a follow up PR :+1:

njgheorghita:
Typically, the solution for avoiding external dependencies on data providers is to implement some kind of "MockServer". Though, tbh, in this situation, we're dealing with like 8/9 different providers for `Goerli` and just as many for `Mainnet`... so my guess is that it'll be quite a pain in the *** to implement a mock server for all of these endpoints

light-client/src/config/checkpoints.rs
```diff
@@ -21,13 +21,31 @@ pub struct RawSlotResponseData {
...
 pub struct Slot {
+    #[serde(deserialize_with = "str_or_u64")]
     pub slot: u64,
     pub block_root: Option<H256>,
     pub state_root: Option<H256>,
     pub epoch: u64,
     pub time: StartEndTime,
 }
 
+fn str_or_u64<'de, D>(deserializer: D) -> Result<u64, D::Error>
```
ogenev, line 32: We don't need to implement this, `ethportal-api` already [uses](https://github.com/ethereum/trin/blob/01f9a5e4e0d28e5c2f0649eca78426a07f2d2447/ethportal-api/Cargo.toml#L38) a crate [serde_this_or_that](https://docs.rs/serde-this-or-that/latest/serde_this_or_that/index.html) where [as_u64](https://docs.rs/serde-this-or-that/latest/serde_this_or_that/fn.as_u64.html) is implemented for us.

ogenev:
> However, I'd like to hear from @ogenev if there's any possible way to remove these external dependencies? (of course, I'm also interested in feedback on the actual solution implemented here)

Looking at the bigger picture, this test is testing the checkpoint fallback service.

 For a light client to start syncing, we must provide a finalized block root for a checkpoint within the subjectivity period.

If the consensus node or in our case Beacon portal network doesn't have the requested checkpoint available, we may redirect this request to trusted checkpoint providers. I'm not sure if we have this use case for Portal and if we want to use a fallback service. If the answer is negative, we can remove this service and the test.

I agree that using external dependencies is not a good idea but with the current fix, we may get away with it. I will give this test one more chance to start falling before taking more drastic actions :smile_cat: .

PR #1101, opened 2024-01-17
refactor: extract gossip range calculation into BridgeMode

### LINKED ISSUE #1100: Simplify bridge + add tests
> Tbh, I think it would be a worthwhile improvement to extract this logic into a method on BridgeMode. Looks like the only external variable needed is latest_block which could be easily passed in. It would also be good for sanity's sake to add some unit testing to this logic. If you want to address it in a following pr, that's fine by me.

https://github.com/ethereum/trin/pull/1096#discussion_r1455731349

njgheorghita:
### What was wrong?
Fixes #1100

### How was it fixed?
Extracted the logic and wrote unit tests

### To-Do


KolbyML (approved):
:shipit: cool tests

PR #1102, opened 2024-01-18
refactor: replace native implementation of this_or_that

njgheorghita:
### What was wrong?
Ahhh, I knew I was forgetting an easier way to accomplish #1099 - this is the fix suggested in https://github.com/ethereum/trin/pull/1099#discussion_r1457142840

### How was it fixed?
- Replace native implementation of "this or that" logic for library version.

### To-Do


KolbyML (approved):
:shipit:

PR #1103, opened 2024-01-19
refactor(bridge): extract logic from history bridge into execution api

njgheorghita:
### What was wrong?
Noticed an opportunity to refactor while working on some scripts to test providers. Decided to split it out into its own pr to minimize the diff in the other pr. 

### How was it fixed?
Some logic was located in the `HistoryBridge` to construct a block body. In fact, this logic is much more "relationally adjacent" to the logic inside the `ExecutionApi` so it belongs there.

### To-Do


KolbyML (approved):
:shipit:

PR #1104, opened 2024-01-19
fix(bridge): use archive endpoint for backfill

njgheorghita:
### What was wrong?
After some initial testing, the `archive.mainnet.ethpandaops.io` endpoint is clearly superior to the `geth` / `erigon` endpoints w/ regards to successfully creating receipts (aka it has access to what I assume are receipts being pruned on `geth` / `erigon`).

### How was it fixed?
This is sort of a hotfix. While it will improve the amount of valid receipts making it into the network (which is a good thing), it's still not 100% from my experiments. Imo, we should go ahead and merge this update. But I'm happy to hold off on this change until we figure out why it's not quite hitting 100%.

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/bridge.rs
+++ b/ethportal-peertest/src/scenarios/bridge.rs
@@ -24,7 +24,9 @@ pub async fn test_history_bridge(peertest: &Peertest, target: &HttpClient) {
     let portal_clients = vec![target.clone()];
     let epoch_acc_path = "validation_assets/epoch_acc.bin".into();
     let mode = BridgeMode::Test("./test_assets/portalnet/bridge_data.json".into());
-    let execution_api = ExecutionApi::new(Provider::Test).await.unwrap();
+    let execution_api = ExecutionApi::new(Provider::Test, mode.clone())
+        .await
+        .unwrap();
     // Wait for bootnode to start
     sleep(Duration::from_secs(1)).await;
     let bridge = HistoryBridge::new(
--- a/portal-bridge/src/api/execution.rs
+++ b/portal-bridge/src/api/execution.rs
@@ -12,8 +12,11 @@ use url::Url;
 
 use crate::{
     cli::Provider,
-    constants::BASE_EL_ENDPOINT,
-    types::full_header::{FullHeader, FullHeaderBatch},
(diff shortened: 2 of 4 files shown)


KolbyML:
If you noticed more success I assume you verified that the old providers are not performing up to spec. And are testing using the new scripts and not using glados as an indicator
image

This amount of active uTP transfers is unbelievable. And a clear indication something is wrong, wrong with trin not the bridge. Glados is also only reporting 30ish trin nodes because all the find_nodes requests it is running are timing out, probably because the Trin nodes are dying from unrealistic conditions once we merge uTP connection limits and fix the bugs relating to deadlocking accept_with_cid(). (I believe also with connect_with_cid(), but it is a lot harder to replicate) There is probably other bugs as well, and because  of these bugs using the network when it is still syncing (100% radius accepting 100% of offers) right now isn't a realistic metric, soon this kind of overloading won't be possible (at least in this way)

Receipts are bigger then headers/bodie's so they are most likely to fail right now they take the longest to transfer hence have the highest odds of failing because of the various bugs in Trin. 

I think going forward from X to 100% will mostly be done by fixes in Trin

KolbyML (approved):
Looks good :+1:

ogenev (approved):
:boat:

njgheorghita:
> Are we in a place to test if there is a difference in provider performance when the network is mainly suffering from problems not related to provider performance, but bugs in trin itself

I wrote a script (which I'll clean up and submit in a pr shortly) - which tests the ability of various providers to build valid content values. Basically it randomly samples X blocks from each "fork period" (eg, frontier, homestead, london, etc.). And then tests the ability of the bridge infrastructure to build valid content, so there's no gossiping / network involved (isolating the bugs in provider architecture from network/trin bugs). This pr is based on that script. Using the `"archive"` endpoint I saw a drastically improved ability in constructing valid data when compared to `el-cl` endpoints. 

Honestly, there is still a lot more testing to be done w/ regards to provider/bridge performance. Imo, the bulk of that will be with the aid of kurtosis. I'm already running rudimentary tests on kurtosis (eg. run the bridge over an epoch and sample the network for expected data). But there's still some little bugs to figure out / cleaning up to do before it's presentable.

PR #1106, opened 2024-01-23
feat: add test_providers script to test ability of various providers …

njgheorghita:
### What was wrong?
! There is a fair amount of re-org included in this pr !
If any reviewer thinks it's too much, then just lmk and I'll happily split it up into multiple prs.

It's hard to test whether our bridge is successfully creating valid data based on the responses from various providers. This script makes it easy to tell which providers are compatible, and which providers are not (and the reasons why they are not compatible). This could be converted into a unit test, but due to the reliance on live endpoints (pandaops/infura), which we want to avoid in our unit tests, it seemed suitable to just create a script that can be run whenever needed to test various providers. 

### How was it fixed?
- added `test_providers` script
- extracted validation logic from `portal-bridge/bridge/history.rs` into `portal-bridge/api/execution.rs`. This makes it easier to test the validation logic in the `test_provider.rs` script
- Improved error handling in receipts deserialization
- Added `range` functionality to `sample_range` script

### To-Do


pipermerriam:
A suggestion on how to maybe make this better.

Have your script be used to generate static test vectors that represent the responses from a given provider...  save those... have unit tests that use the saved responses to validate we can indeed handle the responses from all of the providers that run in CI.

This way you guys don't have to manually spin up each type of provider each time something changes to be able to test those changes.

KolbyML:
When Portal-Hive code is moved to hive. I am planning to write Bridge tests where we spin up a EL node fill it with the block data we are testing, set the EL as the data source for the bridge and gossip the content to something like trin. Then we can easily verify if all the EL clients are compatible.

portal-bridge/src/types/full_header.rs
```diff
@@ -77,6 +80,40 @@ impl TryFrom<Value> for FullHeader {
...
+                .as_ref()
+                .ok_or_else(|| anyhow!("epoch_acc is missing for pre-merge block"))?;
+
+            // Fetch HeaderRecord from EpochAccumulator for validation
+            let header_index = self.header.number % EPOCH_SIZE;
+            let header_record = &epoch_acc[header_index as usize];
+
+            // Validate Header
+            let actual_header_hash = self.header.hash();
+            if header_record.block_hash != actual_header_hash {
```
ogenev, line 98: nitpick: Can we leverage here and on line 106 [anyhow::ensure](https://docs.rs/anyhow/latest/anyhow/macro.ensure.html#)?

src/bin/test_providers.rs
```diff
@@ -0,0 +1,444 @@
...
+    fn iterator() -> Vec<Providers> {
+        vec![
+            Providers::PandaGeth(SuccessRate::default()),
+            Providers::PandaErigon(SuccessRate::default()),
+            Providers::PandaArchive(SuccessRate::default()),
+            Providers::Infura(SuccessRate::default()),
+        ]
+    }
+
+    fn client(&self) -> Client {
```
ogenev, line 399:
```suggestion
    fn get_client(&self) -> Client {
```

src/bin/test_providers.rs
```diff
@@ -0,0 +1,444 @@
...
+            | Ranges::London((_, success_rate))
+            | Ranges::Merge((_, success_rate))
+            | Ranges::Shanghai((_, success_rate)) => {
+                success_rate.failures += failures;
+                success_rate.total += total;
+            }
+        }
+    }
+
+    pub fn iterator(sample_size: usize, latest_block: u64) -> Vec<Ranges> {
```
ogenev, line 242:
```suggestion
    pub fn into_vec(sample_size: usize, latest_block: u64) -> Vec<Ranges> {
```

src/bin/test_providers.rs
```diff
@@ -0,0 +1,444 @@
...
+            | Providers::PandaErigon(success_rate)
+            | Providers::PandaArchive(success_rate)
+            | Providers::Infura(success_rate) => {
+                success_rate.failures += failures;
+                success_rate.total += total;
+            }
+        }
+    }
+
+    fn iterator() -> Vec<Providers> {
```
ogenev, line 390:
```suggestion
    into_vec() -> Vec<Providers> {
```
We can also leverage here https://docs.rs/strum_macros/0.25.3/strum_macros/derive.EnumIter.html here but I'm not sure if it is worth importing `strum` only for this use case.

ogenev (approved):
⛵

njgheorghita:
@pipermerriam 

In hindsight, this script accomplishes two things at once
1. Provider liveness
2. Provider compatibility

Testing liveness is a useful thing imo (ex. I used it multiple times this week to check if our `erigon` provider was live, which it was not) and not something that can be extracted into a unit test. However, this script is somewhat overkill for simply checking provider liveness.

As you suggested, provider compatibility is something that is better suited for a static, unit test: tracking it in #1117 . 

I'm going to go ahead with this merge, because this pr also includes some useful refactoring that makes generating `era1` files easier. Given that...
1. Provider compatibility will be addressed in an upcoming pr
2. Removing the `test_providers.rs` script is very simple and can be done anytime

PR #1108, opened 2024-01-23
refactor(bridge): improve error handling in bridge

njgheorghita:
### What was wrong?
The error handling in the bridge isn't great. Lack of clearly outputting errors led to a change (#1104) which turns out to not be as performant as previously expected.

### How was it fixed?
- When launching a thread w/ a timeout to gossip a full block, we should still handle the error case and output the error message in the situation when the thread does not timeout but an error occurs. If there is a timeout, then a timeout error message will be displayed.
- There's no reason not to attempt gossiping receipts when there is a failure gossiping bodies. Instead of returning an error, simply displaying the error should be sufficient to communicate that something's wrong.

### To-Do

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -193,15 +193,17 @@ impl HistoryBridge {
         permit: Option<OwnedSemaphorePermit>,
     ) {
         tokio::spawn(async move {
-            if (timeout(
+            match timeout(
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
Looks good in makes sense. Crazy we would just give up on trying to gossip a receipt if a body failed to be gossiped. Good find :+1:

PR #1110, opened 2024-01-24
build: update uTP depend to version with fix for uTP transfer limit

KolbyML:
### What was wrong?
Me and nick talked and he said it was best we pulled the updated fixes like this. The same way we did it when waiting for the concurrent discv5 changes to be merged.
### How was it fixed?
Updated utp depend to the branch with my fixes required for (uTP transfer Limit) PR


ogenev (approved):

njgheorghita (approved):
:ship:

PR #1111, opened 2024-01-25
feat(state network): align StateContentKey with spec

morph-dev:
### What was wrong?

The StateContentKey was not aligned with the current spec.

### How was it fixed?

I did the following:

- aligned content key with the spec
- added `Nibbles` type, in a separate folder because it will be reused with other content value types
- added tests, using test vectors from https://gist.github.com/mynameisdaniil/9cb64fa3206b07e8c20cfa23178f5030 and plus more
- small error handling refactoring

### To-Do

### DIFF
--- a/ethportal-api/Cargo.toml
+++ b/ethportal-api/Cargo.toml
@@ -54,7 +54,6 @@ validator = { version = "0.13.0", features = ["derive"] }
 env_logger = "0.9.0"
 quickcheck = "1.0.3"
 rstest = "0.16.0"
-test-case = "3.3.1"
 test-log = { version = "0.2.11", features = ["trace"] }
 tracing = "0.1.36"
 tracing-subscriber = "0.3.15"
--- a/ethportal-api/Cargo.toml
+++ b/ethportal-api/Cargo.toml
@@ -54,6 +54,7 @@ validator = { version = "0.13.0", features = ["derive"] }
 env_logger = "0.9.0"
 quickcheck = "1.0.3"
 rstest = "0.16.0"
+test-case = "3.3.1"
 test-log = { version = "0.2.11", features = ["trace"] }
 tracing = "0.1.36"
 tracing-subscriber = "0.3.15"
--- a/ethportal-api/src/types/content_key/error.rs
+++ b/ethportal-api/src/types/content_key/error.rs
@@ -1,9 +1,11 @@
 use thiserror::Error;
 
+use crate::utils::bytes::hex_encode;
+
 /// An error decoding a portal network content key.
 #[derive(Clone, Debug, Error, PartialEq)]
 pub enum ContentKeyError {
-    #[error("unable to decode key SSZ bytes {input} due to {decode_error:?}")]
+    #[error("Unable to decode key SSZ bytes {input} due to {decode_error:?}")]
     DecodeSsz {
         decode_error: ssz::DecodeError,
         input: String,
@@ -11,4 +13,13 @@ pub enum ContentKeyError {
 
     #[error("Input Vec has length {received}, expected {expected})")]
     InvalidLength { received: usize, expected: usize },
+}
+
+impl ContentKeyError {
+    pub fn from_decode_error<T: AsRef<[u8]>>(decode_error: ssz::DecodeError, input: T) -> Self {
+        Self::DecodeSsz {
+            decode_error,
+            input: hex_encode(input),
+        }
+    }
 }
--- a/ethportal-api/src/types/content_key/state.rs
+++ b/ethportal-api/src/types/content_key/state.rs
@@ -1,100 +1,129 @@
-use crate::{types::content_key::overlay::OverlayContentKey, utils::bytes::hex_encode_compact};
-use ethereum_types::{U256, U512};
+use crate::{
+    types::{content_key::overlay::OverlayContentKey, state_trie::nibbles::Nibbles},
+    utils::bytes::hex_encode_compact,
+    ContentKeyError,
+};
+use ethereum_types::{Address, H256};
 use sha2::{Digest as Sha2Digest, Sha256};
-use sha3::{Digest, Keccak256};
-use ssz::{Decode, Encode};
+use ssz::{Decode, DecodeError, Encode};
 use ssz_derive::{Decode, Encode};
-use ssz_types::{typenum, FixedVector, VariableList};
 use std::fmt;
 
+// TODO(morph): replace with correct link once PR is merged.
+// Prefixes for the different types of state content keys:
+// https://github.com/ethereum/portal-network-specs/blob/655f3e1cac7450888023aec1fc6f339cd679ffdd/state-network.md
+pub const STATE_ACCOUNT_TRIE_NODE_KEY_PREFIX: u8 = 0x20;
+pub const STATE_STORAGE_TRIE_NODE_KEY_PREFIX: u8 = 0x21;
+pub const STATE_CONTRACT_BYTECODE_KEY_PREFIX: u8 = 0x22;
+
 /// A content key in the state overlay network.
-#[derive(Clone, Debug, Decode, Encode)]
-#[ssz(enum_behaviour = "union")]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum StateContentKey {
     /// A trie node from the state trie.
-    AccountTrieNode(AccountTrieNode),
+    AccountTrieNode(AccountTrieNodeKey),
     /// A trie node from some account's contract storage.
-    ContractStorageTrieNode(ContractStorageTrieNode),
-    /// A leaf node from the state trie and the associated Merkle proof against a particular state
-    /// root.
-    AccountTrieProof(AccountTrieProof),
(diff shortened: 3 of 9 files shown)


ogenev (approved):
Looks great! 💯

ethportal-api/Cargo.toml
```diff
@@ -54,6 +54,7 @@ validator = { version = "0.13.0", features = ["derive"] }
 env_logger = "0.9.0"
 quickcheck = "1.0.3"
 rstest = "0.16.0"
+test-case = "3.3.1"
```
ogenev, line 57: We use [rstest](https://docs.rs/rstest/latest/rstest/index.html) for test fixtures but I like `test-case` so no blocker on my side for adding this crate.

njgheorghita: This is somewhat of a *nitpick*, but I do think there's a strong argument to remain consistent throughout the codebase w/ regards to what libraries we're using. The one exception I can think of is when we use `surf` instead of `ureq` for http requests, b/c `ureq` has poor async support. It does seem to me like `rstest` and `test-case` are doing essentially the same function...

ogenev: > This is somewhat of a _nitpick_, but I do think there's a strong argument to remain consistent throughout the codebase w/ regards to what libraries we're using. The one exception I can think of is when we use `surf` instead of `ureq` for http requests, b/c `ureq` has poor async support. It does seem to me like `rstest` and `test-case` are doing essentially the same function...

I agree but it is a development dependency, we don't require this on a release so i'm not strongly against using both libraries.

morph-dev: I'm with @njgheorghita on this one. It is just dev dependency but there are advantages in having consistent codebase.
I just wasn't aware of rstest. Done

ethportal-api/src/types/content_key/state.rs
```diff
@@ -1,140 +1,154 @@
...
+use ethereum_types::{Address, H256};
 use sha2::{Digest as Sha2Digest, Sha256};
-use sha3::{Digest, Keccak256};
-use ssz::{Decode, Encode};
+use ssz::{Decode, DecodeError, Encode};
 use ssz_derive::{Decode, Encode};
-use ssz_types::{typenum, FixedVector, VariableList};
 use std::fmt;
 
+// TODO(morph): replace with correct link once PR is merged.
```
njgheorghita, line 12: Typically we prefer opening issues & assigning to self, trying to avoid todo's scattered around the codebase

ethportal-api/src/types/content_key/state.rs
```diff
@@ -143,224 +157,117 @@ impl fmt::Display for StateContentKey {
...
-        0x29, 0x33, 0x33, 0xa8, 0x30,
+        0x00, 0x0d, 0x83, 0x62, 0x01, 0x31, 0x8e, 0xc6, 0x89, 0x9a, 0x67, 0x54, 0x06, 0x90, 0x38,
+        0x27, 0x80, 0x74, 0x32, 0x80,
     ];
 
-    const CODE_HASH: [u8; 32] = [
-        0xd1, 0xc3, 0x90, 0x62, 0x4d, 0x3b, 0xd4, 0xe4, 0x09, 0xa6, 0x1a, 0x85, 0x8e, 0x5d, 0xcc,
-        0x55, 0x17, 0x72, 0x9a, 0x91, 0x70, 0xd0, 0x14, 0xa6, 0xc9, 0x65, 0x30, 0xd6, 0x4d, 0xd8,
-        0x62, 0x1d,
+    const HODE_HASH: [u8; 32] = [
```
njgheorghita, line 179: `CODE_HASH`

ethportal-api/src/types/content_key/state.rs
```diff
@@ -143,224 +157,117 @@ impl fmt::Display for StateContentKey {
...
-            path: path.into(),
-        };
-
-        let key = StateContentKey::AccountTrieNode(node);
-        let encoded: Vec<u8> = key.clone().into();
-
-        assert_eq!(hex_decode(expected_content_key).unwrap(), encoded);
-        assert_eq!(expected_content_id, key.content_id());
+    fn account_trie_node_key() {
+        let expected_content_key = "0x2024000000c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a4700005000000123456789abc";
```
njgheorghita, line 187: It seems appropriate to add these state test vectors to the [specs](https://github.com/ethereum/portal-network-specs/tree/master) now, in the same way that we do for history & beacon networks.

ethportal-api/src/types/content_key/state.rs
```diff
@@ -143,224 +157,117 @@ impl fmt::Display for StateContentKey {
...
-            address: ADDRESS.to_vec().into(),
-            code_hash: CODE_HASH,
-        };
+    fn decode_key_with_invalid_selector_should_fail() {
+        let invalid_selector_content_key = "0x0024000000c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a4700005000000";
+        assert_eq!(
+            StateContentKey::try_from(hex_decode(invalid_selector_content_key).unwrap())
+                .unwrap_err()
+                .to_string(),
+            "Unable to decode key SSZ bytes 0x0024000000c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a4700005000000 due to UnionSelectorInvalid(0)",
```
njgheorghita, line 249: I would probably use `format!()`/string interpolation here so that you can re-use the `invalid_selector_content_key` variable, without having to duplicate the actual value.

ethportal-api/src/types/state_trie/nibbles.rs
```diff
@@ -0,0 +1,189 @@
...
+        if self.is_odd_length {
+            write!(f, "Nibbles {{ path: {} }}", &path[1..])
+        } else {
+            write!(f, "Nibbles {{ path: {} }}", &path)
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
njgheorghita, line 72: Nice tests! 🚀 & thanks for updating to `rstest`

ethportal-api/src/types/state_trie/nibbles.rs
```diff
@@ -0,0 +1,189 @@
...
+        self.packed_nibbles
+            .iter()
+            .flat_map(Self::unpack_nibble_pair)
+            .skip(if self.is_odd_length { 1 } else { 0 })
+            .collect()
+    }
+
+    fn try_pack_nibbles(nibbles: &[u8]) -> Result<u8> {
+        if let [a, b] = nibbles {
+            ensure!(*a <= 0xF, "Invalid nibble: {}", a);
```
njgheorghita, line 44: Oooh, I wasn't aware of `ensure!`, nice

ethportal-api/src/types/content_key/error.rs
```diff
@@ -12,3 +14,12 @@ pub enum ContentKeyError {
     #[error("Input Vec has length {received}, expected {expected})")]
     InvalidLength { received: usize, expected: usize },
 }
+
+impl ContentKeyError {
```
njgheorghita, line 18: I'm not great at rust, so this is just a bit of speculation and definitely a nitpick. Your instinct to create this fn is on point. But, I suspect that it's possible to do it even a bit more cleanly... I think if you `impl From<DecodeError> for ContentKeyError` then you might be able to do away with the `.map_err(..)` completely in `history.rs`. Maybe something like [this](https://stackoverflow.com/questions/57942901/how-to-implement-from-trait-for-custom-error-types)?

njgheorghita (approved):
Looking good! Thanks! Just a few comments/nitpicks

ethportal-api/src/types/content_key/state.rs
morph-dev: Done. I don't have permission to assign it to myself or set labels. Can you please do it?

ethportal-api/src/types/content_key/state.rs
morph-dev: There is already [github gist](https://gist.github.com/mynameisdaniil/9cb64fa3206b07e8c20cfa23178f5030) with test vectors. I took values from there.

The spec PR is in more or less final stage, but not yet merged. Once that one is in, I will create a PR to add test vectors as well.

ethportal-api/src/types/content_key/state.rs
morph-dev: Oops. Thanks. Done.

ethportal-api/src/types/content_key/error.rs
morph-dev: I'm not sure I can do it exactly like that, because I'm not passing only `DecodeError`, I'm also passing `input` (`AsRef<[u8]`).

I guess I could still use `From` trait, but it would be from a tuple. Not sure if that is clean enough... What do you think?

njgheorghita: I wouldn't worry about it, it was just a thought, doesn't seem like it's worth more effort

PR #1112, opened 2024-01-25
fix: update receipts type to work with pre-byzantium receipts

njgheorghita:
### What was wrong?
Pre-byzantium receipts were actually broken. Since pre-byzantium receipts have a different name for the `"root"` field, as opposed to `"status"` for post-byzantium, we need a way to decode either type.

### How was it fixed?
- added an alias to the `"outcome"` field

### To-Do


KolbyML (approved):
:shipit: very cool of RyanSchneider helping us fix this by letting us know they do serve roots

PR #1113, opened 2024-01-25
build: upgrade rust version to 1.75.0

morph-dev:
### What was wrong?

We were using a bit stale version of rust in our tools (1.71.1).

I noticed this because on my machine I was using the latest stable version (1.75.0) and clippy was complaining about some issue that weren't caused by me. After a bit of digging I realized that those were checks introduced in recent versions.

### How was it fixed?

- Updated the rust version in all places that I could find
- Fixed clippy warnings

### To-Do


ogenev (approved):
Looks good, I was going to propose to update to 1.75.0 too 👍.

KolbyML (approved):

njgheorghita:
@ogenev Does that mean we can get rid of using the `async_trait` library?

njgheorghita (approved):

ogenev:
> @ogenev Does that mean we can get rid of using the `async_trait` library?

I think so

PR #1115, opened 2024-01-26
feat: configure bridge provider via env variable

njgheorghita:
### What was wrong?
We need a better way to switch between pandaops provider endpoints. For example, the `erigon` nodes went down the other day, breaking our bridge. The only way to change this would be to update the hardcoded `BASE_*_ENDPOINT` variables, then cut a new release, and then deploy the new release. With this change, we can simply change the env variable in the deployment scripts and re-run the deployments. 

*Note* - These env variables will override the default values for pandaops endpoints. If the user selects a non-pandaops provider via the cli flag, the env variables will be ignored. 

### How was it fixed?
- Added the feature to select the pandaops provider endpoint via env variable

### To-Do

### DIFF
--- a/portal-bridge/src/api/consensus.rs
+++ b/portal-bridge/src/api/consensus.rs
@@ -4,9 +4,7 @@ use anyhow::anyhow;
 use surf::{Client, Config};
 use url::Url;
 
-use crate::{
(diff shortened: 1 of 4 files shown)


mrferris (approved):

KolbyML (approved):
We should add fallback provider support as well

njgheorghita:
@KolbyML tracking in #1123

PR #1118, opened 2024-01-26
feat(state network): create StateContentValue type

morph-dev:
### What was wrong?

We didn't have any content-value types for the state network.

### How was it fixed?

Created minimal version of all required types. More functionality will be added as they are needed.

### To-Do

- [x] Add tests


morph-dev:
I still have to create test vectors and add tests, but would like some initial feedback.

ethportal-api/src/types/state_trie/witness.rs
```diff
@@ -0,0 +1,31 @@
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum, VariableList};
+
+use super::nibbles::Nibbles;
+
+/// The RLP encoding of a trie node.
+pub type EncodedTrieNode = VariableList<u8, typenum::U1024>;
+
+/// The ordered list of trie nodes. Together the make the path in a trie, first node being the
```
ogenev, line 9:
```suggestion
/// The ordered list of trie nodes. Together they make the path in a trie, first node being the
```

ogenev (approved):
Looks solid to me 🚀 .

ethportal-api/src/types/content_value/state.rs
```diff
@@ -0,0 +1,285 @@
...
+pub struct ContractStorageTrieNodeWithProof {
+    /// A contract storage trie node with a proof.
+    pub proof: ContractStorageTrieWitness,
+    /// A block at which the proof is anchored.
+    pub block_hash: H256,
+}
+
+pub type ByteCode = VariableList<u8, typenum::U65536>;
+
+/// A content value type, used when retriving contract's bytecode.
```
njgheorghita, line 97: `retrieving`

ethportal-api/src/types/content_value/state.rs
```diff
@@ -0,0 +1,285 @@
...
+/// A content value type, used when offering a trie node from the contract storage trie.
+#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
+pub struct ContractStorageTrieNodeWithProof {
+    /// A contract storage trie node with a proof.
+    pub proof: ContractStorageTrieWitness,
+    /// A block at which the proof is anchored.
+    pub block_hash: H256,
+}
+
+pub type ByteCode = VariableList<u8, typenum::U65536>;
```
njgheorghita, line 95: maybe just a small comment explaining why this max size was chosen

ethportal-api/src/types/content_value/state.rs
```diff
@@ -0,0 +1,285 @@
...
+    use rstest::rstest;
+    use serde_json::Value;
+
+    use crate::{
+        types::state_trie::{nibbles::Nibbles, witness::TrieWitness},
+        utils::bytes::hex_decode,
+    };
+
+    use super::*;
+
```
njgheorghita, line 128: This seems like a good place to explain where the test values were sourced from - usually a github link (that contains some commit hash) is ideal

njgheorghita (approved):
Lookin' good!

ethportal-api/src/types/content_value/state.rs
morph-dev: I just updated tests with new test vectors that are being discussed in new [PR](https://github.com/ethereum/portal-network-specs/pull/271). I will wait with merging this PR until the other one is merged and then I will add a link here.

pipermerriam (commented):
No need to block this because of this comment, but it's something probably worth doing soon.

Our test vectors need to be importable via code.  Preferably something like embedding a git submodule at a specific commit and reading the files from that submodule.  Without this, these test vectors will inevitably become outdated because nobody copied in the new values...

And I suspect that may require some basic work standardizing how the test vectors are written and stored.

KolbyML:
Standardized tests go in this https://github.com/ethereum/portal-spec-tests repo

KolbyML (approved):
:shipit:

morph-dev:
I converted test vectors to .yaml files and added them to the https://github.com/ethereum/portal-spec-tests repo.

However, I'm not sure how I can access them from this repo? I know how to add repo as a dependency, but not how to read files from there. Any advice?

I'm fine with merging this as is, and followup later

pipermerriam:
> However, I'm not sure how I can access them from this repo? I know how to add repo as a dependency, but not how to read files from there. Any advice?

I don't know how the team will want to do this, but in my past lives as a developer, we added the tests repo as a git submodule, which requires that CI be setup to pull down the submodules so that the files are available during test runs.

KolbyML:
I looked into it a little bit and I think the best options are
- what piper suggested using a git submoduals
- or using a cargo make task https://github.com/sagiegurari/cargo-make?tab=readme-ov-file#tasks-dependencies-and-aliases which we could make when cargo test is run that portal-spec-tests is pulled

I haven't looked too much into git submoduals but ideally these tests are only pulled in if someone is using them aka running tests. So maybe this looks like a combination of both of the things I listed or just one of them.

I am :+1: for handling this in a follow up PR

morph-dev:
I believe that git submodules are the best way forward, but I would definitely leave that work for a separate PR.

ogenev:
Going to merge this, let's iterate on it in future PRs.

PR #1119, opened 2024-01-26
fix: add deployment steps for updating glados' trin node

mrferris:
### What was wrong?

Trin deployment steps leave glados's trin node un-updated

### To-Do


KolbyML (changes requested):
Trin on glados hasen't been updated in 6 weeks. These instructions don't work to pull the latest image of Trin. I found the fix and will make a PR on the cluster repo.

I think instead of sshing in we should just use anisble like we do for trin. 
For
- consistency
- reproducibility
- etc

KolbyML:
Here is a PR to supercede this one https://github.com/ethereum/trin/pull/1120

For consistence we should ansible for deployments instead of sshing and running commands. Excluding the issue in the ansible and in this PR's instructions here which won't update the trin node. Which is a seperate issue introduced when we added fluffy support to glados's docker compose file

PR #1120, opened 2024-01-28
feat: add instructions for glados deployment

KolbyML:
### What was wrong?
supercedes https://github.com/ethereum/trin/pull/1119 We didn't have instructions to do glados's deployment

Also glados's ansible playbook was broken glados Trin instance hasn't been updated in 6 weeks https://github.com/ethereum/cluster/pull/923 here is the fix
image
^ before ansible fix
image
^ after ansible fix
### How was it fixed?
Wrote the instructions and fixed the playbook

### DIFF
--- a/book/src/developers/contributing/releases/deployment.md
+++ b/book/src/developers/contributing/releases/deployment.md
@@ -53,6 +53,9 @@ This step directs Ansible to use the current master version of trin. Read [about
 - Make sure you've pulled the latest master branch of the deployment scripts, to include any recent changes: `git pull origin master`
 - Go into Portal section of Ansible: `cd portal-network/trin/ansible/`
 - Run the deployment: `ansible-playbook playbook.yml --tags trin`
+- Run Glados deployment
+  - `cd ../../glados/ansible`
+  - `ansible-playbook playbook.yml --tags trin`
 - Wait for completion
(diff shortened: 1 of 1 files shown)


book/src/developers/contributing/releases/deployment.md
```diff
@@ -53,6 +53,9 @@ This step directs Ansible to use the current master version of trin. Read [about
 - Make sure you've pulled the latest master branch of the deployment scripts, to include any recent changes: `git pull origin master`
 - Go into Portal section of Ansible: `cd portal-network/trin/ansible/`
 - Run the deployment: `ansible-playbook playbook.yml --tags trin`
+- Run Glados deployment
```
njgheorghita, line 56: Could you add a little more detail, what exactly will this restart? I'm guessing just the `trin` instance that glados uses, and not also the glados processes themselves? If I'm right, then something like ...
`Glados: To update the Trin instance that Glados uses:` would be great

KolbyML: This is to run a glados deployment. Glados can run both Trin and Fluffy. So this command would update the glados crates and update whichever Portal client is configured

njgheorghita (approved):

PR #1122, opened 2024-01-29
feat: initial utils for reading/writing e2s files

njgheorghita:
### What was wrong?
In `portal-bridge` we'll soon need the ability to read/write era1 files. These are the initial set of utils for reading/writing `e2s` files, which are the standard format that `era1` files are based on

### How was it fixed?
- Implemented basic encode/decode functionality for `e2s` files
- Added standard test vectors

### To-Do


KolbyML (approved):
:shipit: nothing stands out as wrong. I want to get this in so we (nick/ogy) can start iterating on it

Issue #919, opened 2023-09-15
--ephemeral should use a non-default port


carver:
If a port isn't explicitly specified, and the `--ephemeral` flag is set, then we should set a different port.

The main annoyance of using the default port is if you use `--ephemeral` twice in a row, or run ephemeral and then a non-ephemeral run. Because a bunch of former peers are contacting you on that port, expecting a different public key, your logs get flooded with warnings about incoming packets that can't be decoded. (This happened on Joseph Cook's first contact with trin)

Maybe chose the port based on the generated node-id, to help spread it out across runs? Maybe randomize the port?

1010adigupta:
I can take up this, you can assign me

KaydenML:
Fixed this issue in this PR
https://github.com/ethereum/trin/pull/1124

KolbyML:
There was a PR open to fix this issue for 10 months. It wasn't reviewed due to lack of interest so I am closing this issue. If a member of the team is interested in this issue, https://github.com/ethereum/trin/pull/1124 should be reviewed and merged

PR #1124, opened 2024-01-30
Feat: --ephemeral should use a non-default port solution

KaydenML:
### What was wrong?
Fixes: https://github.com/ethereum/trin/issues/919
If someone was using a default port and used --ephemeral twice in a row, or run ephemeral and then a non-ephemeral run. Because a bunch of former peers are contacting you on that port, expecting a different public key, your logs get flooded with warnings about incoming packets that can't be decoded
### How was it fixed?
This code sets the discovery port. If in ephemeral mode with the default port, it generates a random port between 30000 and 60000 otherwise it uses the specified discovery port to avoid log flooding issues during consecutive ephemeral runs or mode switches.

### DIFF
--- a/portalnet/src/config.rs
+++ b/portalnet/src/config.rs
@@ -43,11 +43,11 @@ impl Default for PortalnetConfig {
 }
 
 impl PortalnetConfig {
-    pub fn new(trin_config: &TrinConfig, private_key: H256) -> Self {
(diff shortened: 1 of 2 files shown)


KolbyML:
Do we want this or not? If not I am going to close this pr and the issue for it, I will give it a month or some amount of time which feels right.

KolbyML:
I didn't make this issue and since my brother made this PR I felt like I shouldn't review this PR. I am going to close this PR and the issue due to lack of interest

PR #1125, opened 2024-01-31
fix: prevent history backfill bridge from closing before block is gossiped

KolbyML:
### What was wrong?
With the non-blocking changes to backfill, we weren't waiting till all blocks finished gossiping before closing the bridge. I noticed this running kurtosis in single mode, but this will happen for in all modes when it is about to stop.
### How was it fixed?
Solution after we are done creating all the gossipping threads. Await the thread handles. This won't deadlock, because the tokio:spawn has a timeout.

Another solution could be sleeping the tokio::spawn timeout, I don't like that solution as much because then we can keep the application running longer then it has to.

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -6,9 +6,11 @@ use std::{
 };
 
 use anyhow::anyhow;
+use futures::future::join_all;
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
:ship:

PR #1126, opened 2024-01-31
feat(bridge): add EL api ratelimit

KolbyML:
### What was wrong?
Depending on the EL provider we can overwhelm it with requests. Or some providers have a request limit per a day. This is a problem because we don't want to overwhelm/(go over our api limit on something like infura)
### How was it fixed?
By adding surf_govener api rate limiter

### DIFF
--- a/portal-bridge/src/api/consensus.rs
+++ b/portal-bridge/src/api/consensus.rs
@@ -1,10 +1,14 @@
-use std::fmt::Display;
+use std::{fmt::Display, time::Duration};
 
 use anyhow::anyhow;
 use surf::{Client, Config};
+use surf_governor::GovernorMiddleware;
 use url::Url;
 
-use crate::{cli::Provider, BASE_CL_ENDPOINT, PANDAOPS_CLIENT_ID, PANDAOPS_CLIENT_SECRET};
+use crate::{
+    cli::Provider, constants::SECONDS_IN_A_DAY, BASE_CL_ENDPOINT, PANDAOPS_CLIENT_ID,
+    PANDAOPS_CLIENT_SECRET,
+};
 
 /// Implements endpoints from the Beacon API to access data from the consensus layer.
 #[derive(Clone, Debug, Default)]
@@ -13,8 +17,8 @@ pub struct ConsensusApi {
 }
 
 impl ConsensusApi {
-    pub async fn new(provider: Provider) -> Result<Self, surf::Error> {
-        let client = match provider {
+    pub async fn new(provider: Provider, daily_request_limit: f64) -> Result<Self, surf::Error> {
+        let client: Client = match provider {
             Provider::PandaOps => {
                 let base_cl_endpoint = Url::parse(&BASE_CL_ENDPOINT)
                     .expect("to be able to parse static base cl endpoint url");
@@ -38,6 +42,11 @@ impl ConsensusApi {
                 )))
             }
         };
+        // Limits requests to no more then daily_request_limit / SECONDS_IN_A_DAY per a second
+        let period = Duration::from_secs_f64(daily_request_limit / SECONDS_IN_A_DAY);
+        let rate_limit = GovernorMiddleware::with_period(period)
+            .expect("Expect GovernerMiddleware should have received a valid Duration");
+        let client = client.with(rate_limit);
         check_provider(&client).await?;
         Ok(Self { client })
     }
--- a/portal-bridge/src/api/consensus.rs
+++ b/portal-bridge/src/api/consensus.rs
@@ -42,7 +42,7 @@ impl ConsensusApi {
                 )))
             }
         };
-        // Limits requests to no more then daily_request_limit / SECONDS_IN_A_DAY per a second
+        // Limits the number of requests sent to the CL provider in a day
         let period = Duration::from_secs_f64(daily_request_limit / SECONDS_IN_A_DAY);
         let rate_limit = GovernorMiddleware::with_period(period)
             .expect("Expect GovernerMiddleware should have received a valid Duration");
--- a/portal-bridge/src/api/execution.rs
+++ b/portal-bridge/src/api/execution.rs
@@ -8,6 +8,7 @@ use surf::{
     middleware::{Middleware, Next},
     Body, Client, Config, Request, Response,
 };
+use surf_governor::GovernorMiddleware;
 use tokio::time::{sleep, Duration};
 use tracing::warn;
 use url::Url;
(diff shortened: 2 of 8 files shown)


portal-bridge/src/api/execution.rs
```diff
@@ -57,21 +66,28 @@ impl ExecutionApi {
...
+            Provider::Url(url) => {
+                let client: Client = Config::new()
+                    .add_header("Content-Type", "application/json")?
+                    .set_base_url(url.clone())
+                    .try_into()?;
+                client.with(rate_limit).with(Retry::default())
+            }
+            Provider::Test => {
+                let client: Client = Config::new().try_into()?;
+                client.with(rate_limit).with(Retry::default())
```
njgheorghita, line 89: I'm not sure we need a rate limit for reading data from a test file.... but if so, you can move the..
```
client.with(rate_limit).with(Retry::default())
```
line below, since it applies in each case, and then you don't have to repeat it 3 times inside each clause

portal-bridge/src/api/execution.rs
```diff
@@ -48,7 +49,15 @@ pub struct ExecutionApi {
 }
 
 impl ExecutionApi {
-    pub async fn new(provider: Provider, mode: BridgeMode) -> Result<Self, surf::Error> {
+    pub async fn new(
+        provider: Provider,
+        mode: BridgeMode,
+        requests_allowed_per_day: f64,
+    ) -> Result<Self, surf::Error> {
+        // Limits requests to no more then requests_allowed_per_day / 86400.0 per a second
```
njgheorghita, line 57: This comment is just a bit confusing...
*typo* `then` -> `than`
... 
Is it no more than `requests_allowed_per_day` or `86400.0 / second`?
Or is it no more than `(requests_allowed_per_day / 86400.0)` / `second`?

Imo changing `requests_allowed_per_day` to something like `daily_request_limit` helps with the understanding. 
It keeps it simple (in the comments and the cli help string), with just a "daily limit", that's the only variable users need to know about. Users can do the conversion on their own if they need to know the "per second" rate. I'd also do something like..
```rust
const SECONDS_IN_A_DAY: f64 = 86400.0;

let period = Duration::from_secs_f64(daily_request_limit / SECONDS_IN_A_DAY);
```

portal-bridge/src/api/execution.rs
```diff
@@ -48,7 +49,15 @@ pub struct ExecutionApi {
...
+        let period = Duration::from_secs_f64(requests_allowed_per_day / 86400.0);
+        let rate_limit = GovernorMiddleware::with_period(period)
+            .expect("Expect GovnernorMiddleware should have received a valid Duration");
```
njgheorghita, line 60: *typo* `Expect GovernerMiddleware`

portal-bridge/src/cli.rs
```diff
@@ -92,6 +93,13 @@ pub struct BridgeConfig {
         help = "Data provider for consensus layer data. (\"pandaops\" / local node url)"
     )]
     pub cl_provider: Provider,
+
+    #[arg(
+        long = "el-provider-request-limit",
```
njgheorghita, line 98: Why not just make it a "provider-daily-request-limit" and apply it to both el & cl providers?

portal-bridge/src/cli.rs
```diff
@@ -92,6 +93,13 @@ pub struct BridgeConfig {
...
+        help = "Limits the amount of EL provider requests request day_limit / 86400 seconds. Ex. if set to 86400 we will only send EL 1 request a second",
```
njgheorghita, line 99: Again, I think here it's just a bit simpler if we condense the variable/help msg to simply the concept of a "daily limit", this is typically the limit people are concerned about (eg. the infura daily limit) .. and if they need to set the limit to on a "per second" basis, they can calculate the appropriate daily limit themselves
```rust
    long = "provider-daily-request-limit",
    help = "Maximum number of requests sent to the provider in a day."
    default_value_t = PROVIDER_DAILY_REQUEST_LIMIT,
)]
pub provider_daily_request_limit: f64,
```

portal-bridge/src/constants.rs
```diff
@@ -11,3 +11,5 @@ pub const HEADER_WITH_PROOF_CONTENT_VALUE: &str =
 
 // Beacon chain mainnet genesis time: Tue Dec 01 2020 12:00:23 GMT+0000
 pub const BEACON_GENESIS_TIME: u64 = 1606824023;
+
+pub const EL_PROVIDER_REQUEST_LIMIT: f64 = 1_000_000.0;
```
njgheorghita, line 15: `PROVIDER_DAILY_REQUEST_LIMIT`

portal-bridge/src/cli.rs
KolbyML: I guess beacon only uses CL and history only uses EL so this wouldn't be an issue, but it could become an issue when we support state. I guess we can just do it naively now and fix it later.

I would prefer specifing them seperately through for that reason.

KolbyML: I am going to make the change suggested with the assumption bridges only use CL or EL, but not both. If they use both I would want to have the parameters seperate, but we can do that in a future PR if needed.

KolbyML:
All PR concerns are resolved it should be ready for another review

ogenev (approved):
⛵

portal-bridge/src/cli.rs
njgheorghita: I'm not sure that's a safe assumption, for example our `"latest"` bridges run both `history` & `state` network bridges, meaning they send requests to both the el & cl provider

KolbyML: Ok I made it so you can specify a limit for el and a limit for cl pr should be ready for another review.

portal-bridge/src/api/consensus.rs
```diff
@@ -38,6 +42,11 @@ impl ConsensusApi {
                 )))
             }
         };
+        // Limits requests to no more then daily_request_limit / SECONDS_IN_A_DAY per a second
```
njgheorghita, line 45: This comment is still very confusing

portal-bridge/src/api/execution.rs
```diff
@@ -73,6 +79,11 @@ impl ExecutionApi {
                 .try_into()?,
             Provider::Test => Config::new().try_into()?,
         };
+        // Limits requests to no more then daily_request_limit / SECONDS_IN_A_DAY per a second
```
njgheorghita, line 82: And here

njgheorghita (approved):

PR #1127, opened 2024-01-31
feat: add support for reading/writing era1 files

njgheorghita:
### What was wrong?
Add utils for reading/writing from `era1` files.

### How was it fixed?
- wrote utils in `portal-bridge` that support reading / writing from `era1` files

### To-Do


ethportal-api/src/types/execution/transaction.rs
```diff
@@ -54,6 +54,32 @@ impl Transaction {
...
+    fn rlp_append(&self, s: &mut RlpStream) {
+        match self {
+            Self::Legacy(tx) => tx.rlp_append(s),
+            Self::AccessList(tx) => tx.rlp_append(s),
+            Self::EIP1559(tx) => tx.rlp_append(s),
+        }
+    }
+}
+
+impl Decodable for Transaction {
```
njgheorghita, line 67: It really isn't useful / ideal that we don't use the native `Decodable`/`Encodable` traits for rlp stuff on `Transaction`. I'm working on replacing our custom rlp methods on `Transaction` in another pr, but decided to leave those changes out of this pr to avoid bloat.

portal-bridge/src/types/e2s.rs
```diff
@@ -108,16 +124,16 @@ impl Entry {
...
 }
 
-#[derive(Decode, Encode)]
-struct Header {
-    entry_type: u16,
-    length: u32,
-    reserved: u16,
+#[derive(Clone, Debug, Decode, Encode, Default, Eq, PartialEq)]
+pub struct Header {
+    pub type_: u16,
```
KolbyML, line 129: There there a reason for the _ here? `` pub type_: u16,``

KolbyML (approved):
:shipit:

Once milos makes a PR git sub-modualing test data, I would want to move these era1 files there as they are a bit bulky to be in the main repo. But of course that isn't a blocker for this PR. I am happy with us merging this now.

morph-dev:
My PR (#1131) is ready.
Up to you if you want to wait for it, but I agree that test files should most likely be there.

:shipit:

portal-bridge/src/types/e2s.rs
njgheorghita: `type` is a reserved keyword in rust, so the trailing `_` is required to distinguish it from `type`

njgheorghita:
I'm on board with relocating the files to the test vectors repo, but I'm going to go ahead with the merge on this pr now, so I can work off this code a bit more easily, and then I'll address relocating the test files in a future pr

Issue #866, opened 2023-08-25
ethportal-peertest: Move all constant hexstr content to load from yaml files


ogenev:
We store history content hexstr data in the `ethportal-peertest/src/constants.rs` file as strings:

https://github.com/ethereum/trin/blob/01b7d4ef03ed5297d0c72f42fd403d7aa6d7671f/ethportal-peertest/src/constants.rs#L64

 We want to load those strings from yaml files instead.

KaydenML:
I fixed this issue in this PR
PR: https://github.com/ethereum/trin/pull/1130

PR #1130, opened 2024-02-01
refactor: ethportal-peertest: Move all constant hexstr content to load from yaml files

KaydenML:
### What was wrong?
Fixes: https://github.com/ethereum/trin/issues/866
Content keys and values needed to be loaded from yaml files instead of being loaded directly from the constants.rs file
### How was it fixed?
I loaded the strings from a yaml file into the constants.rs instead of loading them directly from the constants.rs file


morph-dev:
If I'm not mistaken, some (if not all) of these files are already in [portal-spec-tests](https://github.com/ethereum/portal-spec-tests) repo.
Maybe wait for my PR (#1131) to be merged and then use those files instead.

KolbyML:
@KaydenML you should make a PR on the portal-spec-tests repo to update those test files to yaml.

ethportal-peertest/src/constants.rs
```diff
@@ -1,64 +1,65 @@
 use ethportal_api::{HistoryContentKey, HistoryContentValue};
-use serde_json::json;
+use serde_yaml::Value;
+use std::fs;
 
 /// History HeaderWithProof content key & value
 /// Block #1000010
 pub fn fixture_header_with_proof_1000010() -> (HistoryContentKey, HistoryContentValue) {
+    let yaml_content = fs::read_to_string(
```
morph-dev, line 8: Bodies of all functions are identical except for the file name.

Create separate function (e.g. `readHistoryContentKeyValue`) that accepts file name as an input and returns `(HistoryContentKey, HistoryContentValue)`.

It can even return `Result<...>` or `Option<...>` but that's up to you.

KaydenML: @morph-dev should be good to go now

morph-dev:
In my opinion, maybe this is too much, especially because this will only be used in tests.

I think you could have used `anyhow::Result<...>` as a return type and you wouldn't need to create `io::Error`. Something like this:

```rust
fn read_history_content_key_value(
    file_name: &str,
) -> anyhow::Result<(HistoryContentKey, HistoryContentValue)> {
    let yaml_content = fs::read_to_string(file_name)?;
    let value: Value = serde_yaml::from_str(&yaml_content)?;

    let content_key: HistoryContentKey =
        serde_yaml::from_value(value.get("content_key").unwrap().clone())?;
    let content_value: HistoryContentValue =
        serde_yaml::from_value(value.get("content_value").unwrap().clone())?;

    Ok((content_key, content_value))
}
```

But, it's up to you. If you prefer it your way, I'm fine with it.
:rocket:

morph-dev (approved):

ethportal-peertest/src/constants.rs
ogenev, line 1: Let's move those functions into `utils.rs`, the `constants.rs` file name implies that we are going to store only constants.

KaydenML: Ok @ogenev  should be good to go now, moved the functions to utils and updated the functions via milos advice

ogenev, line 1: This is an empty file, please delete it.

KaydenML: @ogenev ok now it should be good to go

KaydenML: @ogenev alright now it should be good to go

ogenev (approved):
LGTM 👍

Issue #1129, opened 2024-01-31
Use files from ethereum/portal-spec-tests directly in tests


morph-dev:
Find a way to integrate [ethereum/portal-spec-tests](https://github.com/ethereum/portal-spec-tests) and use them directly in tests.

Discussion on how it can be done can be found here: #1118
Leading idea is to use [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules).

Some test that should be easy to refactor to use files from there are:
- [ethportal-api/src/types/content_value/state.rs](https://github.com/ethereum/trin/blob/master/ethportal-api/src/types/content_value/state.rs)

PR #1131, opened 2024-02-01
chor: add portal-spec-tests repo as submodule

morph-dev:
### What was wrong?

We couldn't use files from [ethereum/portal-spec-tests](https://github.com/ethereum/portal-spec-tests) directly in our tests. See: #1129

### How was it fixed?

- Added `portal-spec-tests` as submodule
- Refactored tests in `ethportal-api/src/types/content_value/state.rs` to use those files instead
- Updated `.circleci/config.yml` to add support for submodules ([instructions](https://circleci.com/docs/configuration-reference/#checkout)

fix: #1129

### To-Do


njgheorghita (commented):
It would be useful to include the basic instructions for using submodules (eg. init / update) in the `README` file as a reference. Basically what you posted in the trin discord channel

morph-dev:
Created dedicated page in the book. Added short info and link to the README.

ethportal-api/src/types/content_value/state.rs
```diff
@@ -121,9 +121,11 @@ mod test {
 
     use super::*;
 
+    const TEST_DATA_DIRECTORY: &str = "../portal-spec-tests/tests/mainnet/state/serialization";
```
njgheorghita, line 124: I'm sure this value will eventually be shared w/ other locations in the code. I'd recommend creating a `ethportal-api/src/constants.rs` module to locate this value. Also, maybe change the var name to `PORTAL_SPEC_TESTS_SUBMODULE_PATH` to be a bit more explicit in what it's pointing towards?

But... maybe it makes more sense to do something like ...
in `ethportal-api/src/constants.rs`
```
pub const PORTAL_SPEC_TESTS_SUBMODULE: &str = "../portal-spec-tests";
```
and then here, import `PORTAL_SPEC_TESTS_SUBMODULE` and ...
```
const TEST_DATA_DIRECTORY = "/tests/mainnet/state/serialization";
```
and then join the values here? Since, it's likely other locations will want the path to the specs submodule, but don't necessarily want the same folder

njgheorghita (approved):

ethportal-api/src/types/content_value/state.rs
morph-dev: I created `src/test_utils/*` and put it there, together with functions that reads file as string (less boilerplate everywhere).

I decided against `ethportal-api/src/constants.rs` because that feels like a file that maybe we want to export publicly (as a crate), while this constant wouldn't make sense or fit there.

njgheorghita: The question I'd ask is, where are we likely to use this constant? According to #1130 we'll need it inside `ethportal-peertest`, and I know we'll eventually need it inside `portal-bridge`. So, maybe it makes sense that it's a constant we should export? But also, each workspace that needs assets from `portal-spec-tests` can define it's own constant inside it's own workspace...

I'm undecided honestly, and it doesn't feel like a make or break decision here. So, let's just move ahead as is

book/src/developers/contributing/git/submodules.md
```diff
@@ -0,0 +1,34 @@
...
+
+```console
+$ git submodule update --init
+```
+
+This page provides short overview of most common use cases.
+
+## Pulling in Upstream Changes from the Submodule Remote
+
+You want to do this when remote version of submodule is updated. The simplest way to resolve this
```
njgheorghita, line 14: `You want to do this when the remote version ...`

njgheorghita (approved):

ethportal-api/src/types/content_value/state.rs
morph-dev: My worry is about its value to the crates that import it outside of this repo, like `glados` and `portal-hive`. It would just be meaningless in that scenario.

But considering that most frequent usage comes from this repo, I understand your point as well. And same as you, I'm undecided.

With that being said, I still prefer it in a test related path (not `ethportal-api/src/constants.rs`). And if we decide to export it, we can just export it from there.

morph-dev:
Rebased to head and cleaned up commit history.

ogenev (approved):
LGTM 👍

PR #1134, opened 2024-02-01
refactor: remove duplicate transaction rlp implementations

njgheorghita:
### What was wrong?
We had a custom implementation of the `Encodable`/`Decodable` rlp traits for `Transaction`. These were removed in favor of actually implementing those traits

### How was it fixed?
- Removed redundant implementations

### To-Do


KolbyML (approved):
:shipit:

PR #1136, opened 2024-02-02
refactor: replace VariableList<u8, X> with ByteListX types

morph-dev:
### What was wrong?

I noticed that we have a type called `ByteList`, which was just:
```
type ByteList = VariableList<u8, typenum::U2048>;
```

I liked the idea considering that `VariableList<u8, ...>` is somewhat common, but I didn't like that the length is somehow assumed.

### How was it fixed?

I renamed the type to `ByteList2048` and I introduced similar types (for lengths that appear in the codebase).

### To-Do


njgheorghita (approved):
💯

ogenev (approved):
⛵

PR #1137, opened 2024-02-02
fix: add timeout to checkpoint requests

### LINKED ISSUE #1132: Fetching latest checkpoint can get stuck
Many tests on circleci were failing yesterday, and in most cases because there was no output for 10 minutes.
This doesn't happen frequently, but it's not the first time it happens.
It also doesn't happen completely randomly, meaning when it starts happening it affects most tests running at that time (so only solution is to wait and try few hours later).

Luckily, it happened on my machine as well and I noticed that one test was completely stuck: [test_fetch_latest_checkpoints](https://github.com/ethereum/trin/blob/a314eec3d31ff26ab6f745b0e32af9567fc10568/light-client/src/config/checkpoints.rs#L320).

After a bit of digging, it seems that we are fetching checkpoints from multiple sources and if any of them is stuck, entire function (`fetch_latest_checkpoints`) is. I'm guessing that one of the sources has some issues every now and then and we don't recover properly.

Another issue is that we are using actual data from the internet and spamming real servers for no reason (we should use mock data in tests).

I propose two improvements that we can do here, and I think we should do both:

1. add timeout to our request. I don't know why they don't fail on their own (maybe library doesn't support that?).
  - improves app functionality 
  - probably fixes tests (because we have multiple sources), at least they will not get stuck, but we will still spam real servers when we run tests

2. don't fetch data from the internet during tests (use mock server)
  - fixes our tests 
  - prevents us from unnecessarily spamming real servers
  - doesn't improve our app (actual app can still get stuck). It might not be that bad as it happens only once and only in fallback case (at least to my understanding)

morph-dev:
### What was wrong?

Fix #1132

Fetching checkpoint could get stuck indefinitely. See #1132 for more details.

### How was it fixed?

I added timeout to our requests.

I'm not sure if this should fix and close #1132, because ideally we should use mock servers for testing (not to spam real ones), but that might not be so easy to do.
I'm fine with either way.

### To-Do

### DIFF
--- a/light-client/src/config/checkpoints.rs
+++ b/light-client/src/config/checkpoints.rs
@@ -1,7 +1,8 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, time::Duration};
 
 use anyhow::anyhow;
 use ethereum_types::H256;
+use reqwest::Response;
 use serde::{Deserialize, Serialize};
 use serde_this_or_that::as_u64;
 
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
Nice! I'm fine with closing the issue for now. I think we're all familiar with the problematic test, so if we see it failing again in the future we can re-open the issue

ogenev (approved):

PR #1138, opened 2024-02-02
feat(state network): improve StateContentKey tests

morph-dev:
### What was wrong?

The StateContentKey tests weren't using test vectors from `portal-spec-tests` and they weren't testing serialization.

### How was it fixed?

I rewrote the tests and added new one.

### To-Do

### DIFF
--- a/ethportal-api/src/types/content_key/state.rs
+++ b/ethportal-api/src/types/content_key/state.rs
@@ -1,14 +1,16 @@
-use crate::{
-    types::{content_key::overlay::OverlayContentKey, state_trie::nibbles::Nibbles},
-    utils::bytes::hex_encode_compact,
-    ContentKeyError,
-};
+use std::fmt;
+
 use ethereum_types::{Address, H256};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use sha2::{Digest as Sha2Digest, Sha256};
 use ssz::{Decode, DecodeError, Encode};
 use ssz_derive::{Decode, Encode};
-use std::fmt;
+
+use crate::{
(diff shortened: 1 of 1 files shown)


ethportal-api/src/types/content_key/state.rs
```diff
@@ -268,24 +252,95 @@ mod test {
...
-    /// Asserts that encoding and decoding returns the same key.
-    fn assert_encode_decode(key: &StateContentKey) {
-        let bytes = key.to_bytes();
-        let decoded_key = StateContentKey::try_from(bytes).unwrap();
-        assert_eq!(key, &decoded_key);
+    #[rstest]
+    #[case::account_trie_node_key("account_trie_node_key.yaml")]
+    #[case::contract_storage_trie_node_key("contract_storage_trie_node_key.yaml")]
+    #[case::contract_bytecode_key("contract_bytecode_key.yaml")]
+    fn encode_decode(#[case] filename: &str) -> Result<()> {
```
ogenev, line 259: I don't think that we need to return `Result` here and below., because we are not doing anything with the error. In tests we usually want to panic or return `( )`.

morph-dev: I personally like to return `Result` in tests because I can use `?` operator instead of `.unwrap()`, which in my mind makes code cleaner (returning Err makes the code fail).

I also don't think it's something that is generally considered bad practice. See:
Rust Book: https://doc.rust-lang.org/book/ch11-01-writing-tests.html#using-resultt-e-in-tests
Rust by Example: https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html#tests-and-

ogenev (approved):
⛵

PR #1139, opened 2024-02-02
feat(state_network): wire up state network json-rpc components

morph-dev:
### What was wrong?

The state network had many components implemented, but it was missing some in order to be able to receive and respond to queries.

### How was it fixed?

- added missing components
    - mostly by coping equivalent ones from history and beacon
-  wired everything together
- tested locally by running trin with state network, querying manually (using json rpc) and observing expected results
    - I had some mocked results that are not part of the PR

### To-Do


ogenev (approved):
Looks good! 👍

PR #1140, opened 2024-02-06
chore: fix github release workflow

ogenev:
### What was wrong?
Github release workflow is [failing](https://github.com/ethereum/trin/actions/runs/7796150696/job/21265228082)

### How was it fixed?
Update gpg pub key and release command

### To-Do


ogenev:
Merging this without approval because it is a hotfix for the github release workflow.

PR #1141, opened 2024-02-06
chore: fix github release workflow

ogenev:
### What was wrong?
Another fix of github release workflow

### To-Do


ogenev:
FInal rage merge :rage4: !

PR #1142, opened 2024-02-06
feat(bridge): add support for gossiping from era1 files

njgheorghita:
### What was wrong?
We want to use era1 files as our data source for gossiping pre-merge content, to relieve some of our dependency on live providers. 

It may be possible to achieve this feature more elegantly, but for some reasons, I decided against trying to integrate the `HistoryBridge` with `era1` file compatibility.
- we still want to support live providers in the `HistoryBridge`
- data sources are different (eg http requests to era1 file sources & live providers)
- validation model is different for data from live providers & era1 files

For these reasons, it seems like the best way forward is to implement a new "kind" of bridge, `Era1Bridge`. 

### How was it fixed?
The `Era1Bridge` takes a directory of `era1` files, processes them all and then gossips out the data.

A few to-dos remain, which will be handled in upcoming prs to avoid bloat in this pr.
- fetch `era1` files from the `era1.ethportal.net` domain. though, this is blocked until the files have been moved onto cloudflare
- validate `era1` files via their checksum,and replace the current validation on a per-content-value basis

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4853,6 +4853,7 @@ dependencies = [
  "jsonrpsee",
  "lazy_static",
  "portalnet",
+ "rand 0.8.5",
  "rlp",
  "rstest 0.18.2",
  "serde",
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
:shipit: looks good to iternate on.

PR #1143, opened 2024-02-06
feat(state_network): implement enr and node_id endpoints for state network

morph-dev:
### What was wrong?

State network doesn't have any endpoint implemented.

### How was it fixed?

I added implementation for the enr and node_id related queries.

Next steps are:

- Implement `StateStorage` (at least minimal support for now)
- Implement the rest of the endpoints

### To-Do

### DIFF
--- a/ethportal-api/src/types/jsonrpc/endpoints.rs
+++ b/ethportal-api/src/types/jsonrpc/endpoints.rs
@@ -14,42 +14,42 @@ pub enum Discv5Endpoint {
 /// State network JSON-RPC endpoints. Start with "portal_state" prefix
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum StateEndpoint {
+    /// params: None
+    RoutingTableInfo,
+    /// params: [enr]
+    Ping(Enr),
     /// params: [enr]
     AddEnr(Enr),
-    /// params: None
-    DataRadius,
     /// params: [node_id]
     DeleteEnr(NodeId),
-    /// params: [enr, content_key]
-    FindContent(Enr, StateContentKey),
+    /// params: [node_id]
+    GetEnr(NodeId),
+    /// params: [node_id]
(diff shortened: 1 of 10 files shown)


ethportal-api/src/types/jsonrpc/endpoints.rs
```diff
@@ -14,42 +14,42 @@ pub enum Discv5Endpoint {
 /// State network JSON-RPC endpoints. Start with "portal_state" prefix
 #[derive(Debug, PartialEq, Eq, Clone)]
 pub enum StateEndpoint {
```
morph-dev, line 16: I reordered these so they are grouped by similar logic: enr/node related, content related

njgheorghita (commented):
@morph-dev Any thoughts on adding some tests to `ethportal-peertest`? Seems like simply copying some history endpoints in [basic.rs](https://github.com/ethereum/trin/blob/master/ethportal-peertest/src/scenarios/basic.rs) is a good place to start. There's certainly some opportunity for refactoring, since I don't believe `ethportal-peertest` was updated throughout the beacon network implementation. However, I don't feel too strongly about refactoring `ethportal-peertest` but these seem like endpoints that should be tested. I'm also happy if you want to do it in a following pr

ogenev (approved):
LGTM 👍

ogenev:
> However, I don't feel too strongly about refactoring `ethportal-peertest` but these seem like endpoints that should be tested. I'm also happy if you want to do it in a following pr.

I think we definitely need to revisit and refactor `ethportal-peertest` to support multiple networks and add beacon and state 
tests.

morph-dev:
I added some tests to `ethportal-peertest` and did a bit of refactoring.

More refactoring would be needed for future tests. And probably with a different approach (because content key/value types are different).

njgheorghita (approved):
Nice! Thanks for the improvements to `peertest`!

PR #1147, opened 2024-02-08
feat(bridge): add http request timeout to prevent requests over 6 minutes

KolbyML:
### What was wrong?
image
Yesterday when testing the bridge we found that a surf http  request took 6 minutes to respond. We assume this is becase the provider took forever to respond. Either way we don't want to wait 6 minutes + to timeout if a provider can't respond in a responsible about of time we should give up and log it, so we aren't confused why we aren't seeing logs.
### How was it fixed?
by setting a 20 second timeout in surf.

I choose 20 seconds to be conservative. I don't believe any provider request should take over 1s and if that does happens the provider is overloaded.

### DIFF
--- a/portal-bridge/src/constants.rs
+++ b/portal-bridge/src/constants.rs
@@ -1,3 +1,5 @@
+use tokio::time::Duration;
+
 // This number was chosen after some experimentation with different batch sizes.
 // It was found that a batch size of 128 was the best compromise between speed and
 // successful response rate. This number may change in the future.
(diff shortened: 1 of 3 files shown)


ogenev (approved):
:shipit:

portal-bridge/src/constants.rs
```diff
@@ -14,3 +16,7 @@ pub const BEACON_GENESIS_TIME: u64 = 1606824023;
 
 pub const PROVIDER_DAILY_REQUEST_LIMIT: f64 = 1_000_000.0;
 pub const SECONDS_IN_A_DAY: f64 = 86400.0;
+
+// This is a very conservative timeout if a provider takes longer then even 1 second it is probably
```
njgheorghita, line 20: `than even 1 second`

njgheorghita (approved):

PR #1149, opened 2024-02-08
feat(bridge): use ethportal.net as source for era1 files

njgheorghita:
### What was wrong?
Source `era1` files from `https://era1.ethportal.net` instead of the local test folder. Randomly shuffle the `era1` files and then gossip out their content.

### How was it fixed?
- added support for fetching `era1` files & randomly shuffling them

### To-Do


KolbyML (approved):
:shipit:

PR #1150, opened 2024-02-08
feat(ethportal-api): add `BeaconState` type

ogenev:
### What was wrong?
To add support for [Era](https://github.com/status-im/nimbus-eth2/blob/stable/docs/e2store.md#era-files) files, we need to deserialize the beacon state. No `BeaconState` type is available in `ethportal-api`.

### How was it fixed?
Implement `BeaconState` type

### To-Do


KolbyML:
I think these test files should go in portal-test-specs.

KolbyML:
> > I think these test files should go in portal-test-specs.
> 
> Hmm, not sure, those are test assets for ethereum beacon network, not related to portal network specs.

The repo is just has spec in the name because of convention (`execution-spec-tests` and `consensus-spec-tests`) the repo is just for test data. Expessially with how big these test data files are. This is exactly why the `portal-spec-test` repo was made so we didn't need to have big test files in `ethereum/trin` and so other portal implementations could reuse the same test data. I am assuming since we are using these test data file other portal beacon implementations would also find value in it as well. Those reasons don't have to be mutually exlusive though to put test data there.

So the general direction should be a migration from `test_assets` -> `portal-spec-tests` for where we host our test data.

ogenev:
> > > I think these test files should go in portal-test-specs.
> > 
> > 
> > Hmm, not sure, those are test assets for ethereum beacon network, not related to portal network specs.
> 
> The repo is just has spec in the name because of convention (`execution-spec-tests` and `consensus-spec-tests`) the repo is just for test data. Expessially with how big these test data files are. This is exactly why the `portal-spec-test` repo was made so we didn't need to have big test files in `ethereum/trin` and so other portal implementations could reuse the same test data. I am assuming since we are using these test data file other portal beacon implementations would also find value in it as well. Those reasons don't have to be mutually exlusive though to put test data there.
> 
> So the general direction should be a migration from `test_assets` -> `portal-spec-tests` for where we host our test data.

Those test files are extracted from `consensus-spec-tests` repo because I don't want to add this repo as a submodule  - it is very huge, we want to use a few folders and use Git Large File Storage which will complicate our workflow.

Those files here are used for testing serialization/deserialization for the `BeaconState` which is part of the consensus specs and is used for deserializing Era files. I don't think that we want to migrate all `test_assets` to `portal-spec-tests` because most of those files are used to test Trin-specific implementations and not all of them are part of portal specs. I think `portal-test-specs` should be kept only for test assets for the portal network implementation which should not include consensus and execution layer test files.

> The repo is just has spec in the name because of convention (`execution-spec-tests` and `consensus-spec-tests`) the repo is just for test data.

I don't think this is true. If we just want to have a repo for storing large files for testing in trin, it should be called `trin-tests`. Using the same convention as `execution-spec-tests` and `consensus-spec-tests` is not a coincidence, it is because we want to accomplish the same thing for the portal - store test assets and test vectors that are used for testing `portal network specs`.

@njgheorghita, @morph-dev  what do you think?

morph-dev:
I'm not sure I have enough knowledge of beacon network and these files to provide opinion on this specific case. I'm also missing broader context of what `portal-spec-tests` should be (and if it should be follow the same principles as `execution-spec-tests` and `consensus-spec-tests`). With that being said, I will still give my opinion.

My feeling is that `portal-spec-tests` should contain test files that can/should be used by all clients in order to test their implementation of the Portal Network. Are all clients expected to deserialize Era files?!
Even if it is not required by portal network spec, but if it is expected or useful in general context (e.g. in order to verify data, build bridge, etc.), I would say they can go there (e.g. under special directory, with README file as explanation).

I also believe that we probably have some files in `test_assets` that we don't want to move to `portal-spec-tests` because they are trin specific.

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -0,0 +1,312 @@
...
+use serde::Deserialize;
+use serde_this_or_that::as_u64;
+use serde_utils;
+use ssz::{Decode, DecodeError, Encode};
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum::U16777216, BitVector, FixedVector, VariableList};
+use std::sync::Arc;
+use superstruct::superstruct;
+use tree_hash::{Hash256, TreeHash};
+use tree_hash_derive::TreeHash;
```
njgheorghita, line 24: Don't we have a clippy lint for ordering imports into groups? eg. `std` first, exterior second, internal third

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -0,0 +1,312 @@
...
+    #[superstruct(only(Capella), partial_getter(copy))]
+    #[serde(deserialize_with = "as_u64")]
+    pub next_withdrawal_validator_index: u64,
+    // Deep history valid from Capella onwards.
+    #[superstruct(only(Capella))]
+    pub historical_summaries: VariableList<HistoricalSummary, HistoricalRootsLimit>,
+}
+
+impl BeaconState {
+    #[cfg(test)]
```
njgheorghita, line 132: Really? Why do we only want to ssz decode the state inside our tests? Is this to avoid the `dead_code` lint? If so, then I'd prefer to use `allow[clippy::dead_code]`

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -0,0 +1,312 @@
...
+    #[case("case_0")]
+    #[case("case_1")]
+    fn serde_beacon_state_bellatrix(#[case] case: &str) {
+        let value = std::fs::read_to_string(format!(
+            "../test_assets/beacon/bellatrix/BeaconState/ssz_random/{case}/value.yaml"
+        ))
+        .expect("cannot find test asset");
+        let value: Value = serde_yaml::from_str(&value).unwrap();
+        let content: BeaconStateBellatrix = serde_json::from_value(value.clone()).unwrap();
+        let serialized = serde_json::to_value(content).unwrap();
```
njgheorghita, line 253: Uhhh, I guess this works, but should we just use `serde_yaml::to_value` for consistency?

njgheorghita (approved):
A few questions about automatically deriving traits, just to double check that there's a good reason for manually implementing them, rather than automatically deriving them.

With regards to the test assets location...
I also don't have a clear understanding of the spec repo's **defined** purpose.
- Is it strictly for client-agnostic json test vectors for clients to test against?

or...

- Is it a dump for all large test files b/c we don't want to increase the size of `trin`?

This is probably something that needs defining, and inscribing in the readme's, so we don't have to repeat this debate for every future test vector. For the purpose of this pr, I think it's ok to move ahead as-is (eg. test files inside `trin`). Let's discuss this in the next sync, figure out how we want to define what test assets go where. And then if we do decide that these test vectors are better suited in the specs repo, then we can make that change later.

ethportal-api/src/types/consensus/beacon_state.rs
ogenev: yes, it is to avoid the `dead_code` lint. I prefer this because that way we will not forget to remove the dead code attribute if it is not used.

Anyways, in #1167 I'm removing this.

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -0,0 +1,312 @@
+#[cfg(test)]
```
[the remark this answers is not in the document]
ogenev: Same reasoning as [this comment ](https://github.com/ethereum/trin/pull/1150#discussion_r1495425807).

ethportal-api/src/types/consensus/beacon_state.rs
ogenev: Not sure what ordering we use but this can be changed in the config. I think now it is internal first, std next and external latest.

ethportal-api/src/types/consensus/participation_flags.rs
```diff
@@ -0,0 +1,65 @@
...
+use tree_hash::{Hash256, PackedEncoding, TreeHash, TreeHashType};
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct ParticipationFlags {
+    bits: u8,
+}
+
+/// Decode implementation that transparently behaves like the inner `u8`.
+impl Decode for ParticipationFlags {
```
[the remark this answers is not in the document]
ogenev: yes, we need custom implementation.

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -0,0 +1,312 @@
...
+    fn ssz_fixed_len() -> usize {
+        <u64 as Encode>::ssz_fixed_len()
+    }
+
+    fn ssz_bytes_len(&self) -> usize {
+        0_u64.ssz_bytes_len()
+    }
+}
+
+impl Decode for Epoch {
```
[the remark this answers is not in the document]
ogenev: Because the `Epoch` struct is a wrapper over `u64`, the derived trait will recognize it as an ssz container.

ethportal-api/src/types/consensus/beacon_state.rs
ogenev: No, because we can't compare `serde_yaml::Value` to `JsonValue` and we want to test json ser/de.

ethportal-api/src/types/consensus/beacon_state.rs
morph-dev: Based on the `rustfmt` [documentation](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports), the default is (what we use) to preserve the source file's import groups.

Alternative, which we could enable, is closer to what @njgheorghita said:
1. `std`, `core` and `alloc`,
2. external crates,
3. `self`, `super` and `crate` imports.

PR #1071, opened 2023-12-13
refactor: re-organize portal-bridge

njgheorghita:
### What was wrong?
Some pretty light re-organization inside `portal-bridge`. Thought it would be better to get this pr through, before some of the more feature-oriented bridge prs. The flat structure of `portal-bridge`'s modules wasn't great and there were some opportunities to nest modules inside related submodules.

### How was it fixed?
- Re-organized `portal-bridge`
- Re-named `Bridge` -> `HistoryBridge`
- Ordered import statements

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #1151, opened 2024-02-09
feat(bridge): make provider rate limit opt in

KolbyML:
### What was wrong?
I did some thinking and I don't like the idea of a rate limit by default. I think it should be an intentional action.
### How was it fixed?
By only setting a rate limit if it was manually asked for. This may change in the future if we have some smart rate limit mechanism, but currently that doesn't exist.

### DIFF
--- a/ethportal-peertest/src/scenarios/bridge.rs
+++ b/ethportal-peertest/src/scenarios/bridge.rs
@@ -10,7 +10,6 @@ use portal_bridge::{
     api::{consensus::ConsensusApi, execution::ExecutionApi},
     bridge::{beacon::BeaconBridge, history::HistoryBridge},
     cli::Provider,
-    constants::PROVIDER_DAILY_REQUEST_LIMIT,
     types::mode::BridgeMode,
 };
 use serde_json::Value;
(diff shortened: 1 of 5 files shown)


njgheorghita:
Why don't you like the idea of a rate limit anymore?

KolbyML:
I do I just don't want it set by default to a random number unless it manually configures itself. For example is 1 million enough calls a day to gossip all new blocks produced a day I haven't checked.

So i would rather people see of this is hitting my node too hard. Where if the api rate limit is too low by default they might think the bridge is broken.

I would be fine if we did the math and set the default to the maximum possible requests for gossiping latest a day

njgheorghita:
- There are approx. [7120](https://ycharts.com/indicators/ethereum_blocks_per_day) blocks / day. 
- For each block we make 3 requests to the provider.
- Our loop for polling the `"latest"` block per day runs on a [5 sec](https://github.com/ethereum/trin/blob/master/portal-bridge/src/bridge/history.rs#L34C7-L34C29) loop. 
  - So at max it will poll `get_latest_block` 17,280 times in one day
  
If we assume that the loops works perfectly well (eg. it will either find a single `"latest"` block or no new block), then at minimum we expect: 
- `17280` - `eth_getBlock('latest')` requests
- `7120` - `eth_getBlockByNumber()` requests
- 0 requests for bodies (since post-merge there are no uncles)
- `7120` batched requests for `getTransactionReceipt` ... there are approx `1million` [transactions per day](https://ycharts.com/indicators/ethereum_transactions_per_day)... since we batch the requests for a transaction receipt, then my understanding is that infura treats each transaction as a separate request...

So.. if my assumptions are correct here, on a "normal" day, in `latest` mode, we will exceed `1million` provider requests....

But, this analysis also exposes a flaw in our architecture....

The rate limiter is for # of http requests. In the case of transactions, we will exceed the infura daily limit while still far below the daily limit for http requests, since a single http request can contain hundreds of batched infura requests.

Infura's free tier allows for [100,000](https://www.infura.io/pricing) requests / day.

There are also other factors to consider... my instinct here is to simplify rather than solve this situation...

In my experience with the bridge-provider connection, it's too difficult to understand whether a provider request failed and for what reason...

For example, in #1152 it took a while to figure out that the rate limiter was the cause of failed http requests to the provider, not the provider itself.

Given that this is the case, I'm sort of leaning towards removing the rate limiter entirely, and simply making sure that we handle & log all cases when provider requests fail. 

When we handle the bridge refactor in #1071 - then I think we can approach the rate limit question a bit more sensibly. Eg. different providers and different modes will have their own rate limits that make sense for those use cases.

What do you think?

PR #1152, opened 2024-02-09
fix(bridge): change rate limiter period calculation

njgheorghita:
### What was wrong?
When the rate limiter was added, it seems to me like this broke our bridge. Locally (running just trin & inside kurtosis), I'm consistently unable to request any data from any provider, latest / backfill mode. 

The culprit appears to be `GovernorMiddleware::with_period(period)`.. it just doesn't appear to be performing as expected. I've updated it to use the `GovernerMiddleware::per_hour()` limit api, which solves the problem locally. 

### How was it fixed?
- changed default el provider to `geth`, since `erigon` is down
- added some debug logs
- removed the rate limiter (aka governer) and the concept of a daily rate limit

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3015,24 +3015,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "governor"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "821239e5672ff23e2a7060901fa622950bbd80b649cdaadd78d1c1767ed14eb4"
-dependencies = [
- "cfg-if 1.0.0",
- "dashmap",
- "futures 0.3.30",
- "futures-timer",
- "no-std-compat",
- "nonzero_ext",
- "parking_lot 0.12.1",
- "quanta",
- "rand 0.8.5",
- "smallvec 1.12.0",
-]
-
-[[package]]
 name = "group"
 version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


njgheorghita:
I'm not sure I like this rate limiter. The problem is that the http responses when the limit was exceeded was simply an `""`. This makes it difficult to propagate a useful error to the user informing them that the rate limit is exceeded. But, I'm looking into a way to see if this is possible

njgheorghita:
On another note, this kind of bug should really have been caught in our testing. Though, testing the bridge isn't straightforward in unit tests....

KolbyML:
image
I think the solution is to handle this error in the retry middleware. Which we weren't handling before

KolbyML:
> On another note, this kind of bug should really have been caught in our testing. Though, testing the bridge isn't straightforward in unit tests....
> 
> Adding a kurtosis run to our ci would be very, very nice for catching these types of bugs.

I think "unit testing" the bridge should be done in hive

KolbyML:
I don't think `GovernerMiddleware::per_hour()` would give a desired response it would just make it so the bridge only works for bursts at the start of every hour. Then a consisent stream through the day

njgheorghita:
@KolbyML tbh, I think there's some other issue going on here with the middleware. When I use `GovernorMiddleware::with_period(period)` - it's not that I'm receiving a 429 error. But rather the response from the provider is simply an empty string - eg `Ok("")` is the response. But when I use `GovernorMiddleware::per_hour()` then the expected response is visible.

njgheorghita:
> I think "unit testing" the bridge should be done in hive

Sure, and iirc it will be possible to integrate this into our ci?

KolbyML:
I am going to test this https://github.com/ethereum/trin/pull/1153 on kurtosis right now and see what happens

KolbyML:
> > I think "unit testing" the bridge should be done in hive
> 
> Sure, and iirc it will be possible to integrate this into our ci?

That is what I heard as well in the testing meeting

njgheorghita:
After talking offline, we decided that the rate limiter is not actually very useful for our use case, especially after introducting era1 file support. So we've decided to remove it entirely

njgheorghita:
@KolbyML ready for another :eyes:

KolbyML (approved):
You forgot to remove the surf_governer library from the cargo.toml

PR #1153, opened 2024-02-09
fix(bridge): handle too many requests error in retry middleware

KolbyML:
### What was wrong?
https://github.com/ethereum/trin/pull/1152
### How was it fixed?

When api rate limit was added we weren't handling the errors it sent in our retry middleware

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4863,6 +4863,7 @@ dependencies = [
  "ssz_types",
  "surf",
  "surf-governor",
+ "surf-retry",
  "test-log",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


njgheorghita:
re https://github.com/ethereum/trin/pull/1152#issuecomment-1936465515 - Also just checking that running the bridge in `"latest"` mode would also be worthwhile

KolbyML:
image
oh well looks like I have to change to the geth provider quick

njgheorghita:
@KolbyML the pandaops geth provider works well for me, you can just do [this](https://github.com/ethereum/trin/pull/1152/files#diff-fe5ce2b51a8fecff7a78b9af38cf2558da21af12f32769eaf56907450f39d160R25) or change it via env variable

njgheorghita:
I pulled this down to do some testing, and from what I can tell, using the surf retry middleware doesn't solve the situation. I'm leaning towards merging #1152 so that we can deploy that fix to the network today, and then continue to investigate the best way to handle & test rate-limit/retry middleware in this pr

KolbyML:
Ok I think I figured it out. Apparently middleware order matters

njgheorghita:
Ahh, nice yeah that would make sense. I just tried with `retry` before `rate_limit` and it's working also

KolbyML:
I was reading surf_retry and it always had surf_govner inserted after so yeah

njgheorghita:
The only argument I can think of against using `surf-retry` is that we've talked about using a "fallback" url to use on retry attempts, and it doesn't appear to me like this will be possible in `surf-retry` - whereas when we have our own "retry" implementation it will be possible

PR #1154, opened 2024-02-09
refactor(bridge): print EL provider response on error

KolbyML:
### What was wrong?
we didn't print the EL providers response in the error messages
### How was it fixed?
by adding them

### DIFF
--- a/portal-bridge/src/api/execution.rs
+++ b/portal-bridge/src/api/execution.rs
@@ -83,7 +83,9 @@ impl ExecutionApi {
             client.config().base_url
         );
         if provider != Provider::Test {
-            check_provider(&client).await?;
+            check_provider(&client).await.map_err(|err| {
+                anyhow!("Check EL provider failed. Provider may be offline: {err:?}")
+            })?;
         }
         let master_acc = MasterAccumulator::default();
         Ok(Self { client, master_acc })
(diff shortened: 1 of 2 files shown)


portal-bridge/src/api/execution.rs
```diff
@@ -303,25 +305,31 @@ impl ExecutionApi {
             .client
             .post("")
             .body_json(&json!(requests))
-            .map_err(|e| anyhow!("Unable to construct json post request: {e:?}"))?
+            .map_err(|e| anyhow!("Unable to construct json post request: {e:?}"))?;
```
njgheorghita, line 308: It's typically a good idea to not have multiple error messages that are the same (though ofc I know you didn't write these in this pr).
maybe something like `Unable to construct json post for batched requests: ...`

portal-bridge/src/api/execution.rs
```diff
@@ -303,25 +305,31 @@ impl ExecutionApi {
...
+        })
     }
 
     async fn send_request(&self, request: JsonRequest) -> anyhow::Result<Value> {
         let result = self
             .client
             .post("")
             .body_json(&request)
-            .map_err(|e| anyhow!("Unable to construct json post request: {e:?}"))?
+            .map_err(|e| anyhow!("Unable to construct json post request: {e:?}"))?;
```
njgheorghita, line 323: `Unable to construct json post for single request: ...`

njgheorghita (approved):

PR #1155, opened 2024-02-09
feat(bridge): add support for gossiping a single era1 file

njgheorghita:
### What was wrong?
It's useful in kurtosis to be able to gossip a single epoch. Need this ability w/ era1 files

### How was it fixed?
- added an option to gossip a single era1 file

### To-Do


njgheorghita:
closing in favor of #1159

PR #1158, opened 2024-02-11
feat(history latest bridge): add timeout to get_latest_block_number()

KolbyML:
### What was wrong?
Setting the timeout on the http client to 20 seconds wasn't the play apparently the default is 60 secconds, so if there was some kind of deadlock that wouldn't avoid it.

On the latest bridge there has been no history logs for around 16 hours. What is interesting is ``History Latest: Haven't received a new block in over 60 seconds. EL provider could be out of sync.`` showed up quite a few times.

So I wanted to avoid adding a future::timeout() so we could see what would happen. But now it is  clear it is required to test with it.
### How was it fixed?
I add a future::timeout() with a somewhat clear error after a minute. The goal with this is to just see some kind of error instead of nothing. Clearly something is blocking this loop or killing it, but what is it?

If we get an error: then somehow the http client is deadlocking somehow
If we get no error: maybe the http is panicing?

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -15,6 +15,7 @@ use tracing::{debug, error, info, warn, Instrument};
 use crate::{
     api::execution::ExecutionApi,
     bridge::utils::lookup_epoch_acc,
+    constants::HTTP_REQUEST_TIMEOUT,
     gossip::gossip_history_content,
     stats::{HistoryBlockStats, StatsReporter},
     types::{full_header::FullHeader, mode::BridgeMode},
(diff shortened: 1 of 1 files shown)


njgheorghita:
> Setting the timeout on the http client to 20 seconds  

Can you explain why? I think I'm just forgetting, but where did we set this timeout?

KolbyML:
We set it in the http_config which had a default timeout of 60. Which I learnt yesterday. So oh course it won't stop a deadlocked future if that is the problem. I assume there wasn't a default timeout. But now that I learnt about that this seems like that right approach.

Our constant is defined in constants.rs

njgheorghita:
Ok right, thanks. So the `http_client` inside `ExecutionApi` has a custom timeout of 20 seconds which we set. This will [override the default timeout of 60 seconds](https://docs.rs/http-client/6.5.1/src/http_client/config.rs.html#107-110)

iiuc, this will stop a deadlocked future (eg provider request)

KolbyML:
If that was the case then why have all the history latest bridge logs stopped again then? This PR should answer question that in a more definitive way. Also if we set the http_client timeout to 20 instead of 60 it wouldn't change the outcome of it deadlocking.

KolbyML:
This was a while ago but I believe me and nick talked on soon a while ago and discussed our plans for what we want. I am going to close this as we ended up finding the issue was a provider issue

PR #1159, opened 2024-02-11
feat(era1 bridge): add gossip epoch mode for reliable testing

KolbyML:
### What was wrong?
Using a provider is unreliable for testing using kurtosis

Solution? Use era1 files as they remove the provider failure point out of the equation
### How was it fixed?
Adding a optional specifier for epoch.

### DIFF
--- a/portal-bridge/src/bridge/era1.rs
+++ b/portal-bridge/src/bridge/era1.rs
@@ -101,12 +101,11 @@ impl Era1Bridge {
     }
 
     pub async fn launch_single(&self, epoch: u64) {
-        let epoch_string = format!("{epoch}-");
-        let era1_path = self
-            .era1_files
-            .clone()
-            .into_iter()
-            .find(|file| file.contains(&epoch_string));
+        let era1_path = self.era1_files.clone().into_iter().find(|file| {
+            file.contains("mainnet-")
+                && file.contains(&format!("{epoch}-"))
+                && file.contains(".era1")
+        });
         match era1_path {
             Some(path) => self.gossip_era1(path).await,
             None => panic!("4444s bridge couldn't find request epoch on era1 file server"),
--- a/portal-bridge/src/bridge/era1.rs
+++ b/portal-bridge/src/bridge/era1.rs
@@ -24,7 +24,7 @@ use crate::{
     stats::{HistoryBlockStats, StatsReporter},
     types::{
         era1::{BlockTuple, Era1},
-        mode::{BridgeMode, FourFoursModeType},
+        mode::{BridgeMode, FourFoursMode},
     },
 };
 use ethportal_api::{
@@ -87,10 +87,8 @@ impl Era1Bridge {
     pub async fn launch(&self) {
         info!("Launching era1 bridge: {:?}", self.mode);
         match self.mode.clone() {
-            BridgeMode::FourFours(FourFoursModeType::Random) => self.launch_random().await,
-            BridgeMode::FourFours(FourFoursModeType::Epoch(epoch)) => {
-                self.launch_epoch(epoch).await
-            }
+            BridgeMode::FourFours(FourFoursMode::Random) => self.launch_random().await,
+            BridgeMode::FourFours(FourFoursMode::Single(epoch)) => self.launch_single(epoch).await,
             _ => panic!("4444s bridge only supports 4444s modes."),
         }
         info!("Bridge mode: {:?} complete.", self.mode);
@@ -102,19 +100,21 @@ impl Era1Bridge {
         }
     }
 
-    pub async fn launch_epoch(&self, epoch: u64) {
-        let epoch_string = epoch.to_string() + "-";
-        for era1_path in self.era1_files.clone().into_iter() {
-            if era1_path.contains(&epoch_string) {
-                self.gossip_era1(era1_path).await;
-                return;
-            }
+    pub async fn launch_single(&self, epoch: u64) {
+        let epoch_string = format!("{epoch}-");
+        let era1_path = self
+            .era1_files
+            .clone()
+            .into_iter()
+            .find(|file| file.contains(&epoch_string));
+        match era1_path {
+            Some(path) => self.gossip_era1(path).await,
+            None => panic!("4444s bridge couldn't find request epoch on era1 file server"),
         }
-        panic!("4444s bridge couldn't find request epoch on era1 file server")
     }
 
     pub async fn gossip_era1(&self, era1_path: String) {
-        info!("Processing era1 file at path: {:?}", era1_path);
+        info!("Processing era1 file at path: {era1_path:?}");
         // We are using a semaphore to limit the amount of active gossip transfers to make sure
         // we don't overwhelm the trin client
         let gossip_send_semaphore = Arc::new(Semaphore::new(GOSSIP_LIMIT));
--- a/portal-bridge/src/bridge/era1.rs
+++ b/portal-bridge/src/bridge/era1.rs
@@ -101,11 +101,10 @@ impl Era1Bridge {
     }
 
     pub async fn launch_single(&self, epoch: u64) {
-        let era1_path = self.era1_files.clone().into_iter().find(|file| {
-            file.contains("mainnet-")
-                && file.contains(&format!("{epoch}-"))
-                && file.contains(".era1")
-        });
+        let era1_path =
+            self.era1_files.clone().into_iter().find(|file| {
+                file.contains(&format!("mainnet-{epoch:05}-")) && file.contains(".era1")
+            });
         match era1_path {
             Some(path) => self.gossip_era1(path).await,
             None => panic!("4444s bridge couldn't find request epoch on era1 file server"),
--- a/portal-bridge/src/bridge/era1.rs
+++ b/portal-bridge/src/bridge/era1.rs
@@ -22,7 +22,10 @@ use crate::{
     },
     gossip::gossip_history_content,
     stats::{HistoryBlockStats, StatsReporter},
-    types::era1::{BlockTuple, Era1},
+    types::{
+        era1::{BlockTuple, Era1},
+        mode::{BridgeMode, FourFoursModeType},
+    },
 };
 use ethportal_api::{
     jsonrpsee::http_client::HttpClient, types::execution::accumulator::EpochAccumulator,
@@ -36,6 +39,7 @@ use trin_validation::{
 const ERA1_DIR_URL: &str = "https://era1.ethportal.net/";
 
 pub struct Era1Bridge {
+    pub mode: BridgeMode,
     pub portal_clients: Vec<HttpClient>,
     pub header_oracle: HeaderOracle,
     pub epoch_acc_path: PathBuf,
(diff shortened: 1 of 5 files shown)


njgheorghita:
ooh, looks like we've done the same thing... #1155

KolbyML:
Oh I didn't see that. I like this version as you only have to enter a number from 0 to 1967* instead of a full file name

njgheorghita:
Agreed that only entering in `e600` is better than the entire file path. But, iiuc your version isn't functional until we migrate to an `index.html` directory? Since our xml directory only has 1000 keys, any epoch from `1000` -> `1867` won't work

njgheorghita:
Ok, well anyways we don't need every epoch right now so I'll close #1155

portal-bridge/src/bridge/era1.rs
```diff
@@ -79,50 +85,76 @@ impl Era1Bridge {
...
-                let serve_block_tuple_handle = Self::spawn_serve_block_tuple(
-                    self.portal_clients.clone(),
-                    block_tuple,
-                    epoch_acc.clone(),
-                    master_acc.clone(),
-                    permit,
-                );
-                serve_block_tuple_handles.push(serve_block_tuple_handle);
+    pub async fn launch_epoch(&self, epoch: u64) {
+        let epoch_string = epoch.to_string() + "-";
```
njgheorghita, line 106: *nitpick* imo using `format!` is a little easier to read / understand

portal-bridge/src/bridge/era1.rs
```diff
@@ -79,50 +85,76 @@ impl Era1Bridge {
...
+            if era1_path.contains(&epoch_string) {
+                self.gossip_era1(era1_path).await;
+                return;
+            }
+        }
+        panic!("4444s bridge couldn't find request epoch on era1 file server")
+    }
+
+    pub async fn gossip_era1(&self, era1_path: String) {
+        info!("Processing era1 file at path: {:?}", era1_path);
```
njgheorghita, line 117: I think this should be `info!("Processing era1 file at path: {era1_path:?}");` ... but also clippy should be catching this?

portal-bridge/src/types/mode.rs
```diff
@@ -155,6 +159,30 @@ impl FromStr for ModeType {
...
+}
+
+impl FromStr for FourFoursModeType {
+    type Err = ParseError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match &s[..1] {
+            "e" => {
+                let epoch = s[1..]
+                    .parse()
+                    .map_err(|_| "Invalid bridge mode arg: epoch number")?;
```
njgheorghita, line 175: `"Invalid 4444s bridge mode arg: era1 epoch number"`

portal-bridge/src/types/mode.rs
```diff
@@ -155,6 +159,30 @@ impl FromStr for ModeType {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FourFoursModeType {
```
njgheorghita, line 163: I'm not a fan of this type name. I think `FourFoursMode` is enough.

portal-bridge/src/types/mode.rs
```diff
@@ -155,6 +159,30 @@ impl FromStr for ModeType {
...
+            "e" => {
+                let epoch = s[1..]
+                    .parse()
+                    .map_err(|_| "Invalid bridge mode arg: epoch number")?;
+                if epoch > 1896 {
+                    return Err("Invalid bridge mode arg: Four Fours only contains up to epoch 1896 was given: {epoch}");
+                }
+                Ok(FourFoursModeType::Epoch(epoch))
+            }
+            _ => Err("Invalid bridge mode arg: type prefix"),
```
njgheorghita, line 181: `"Invalid 4444s bridge mode arg: type prefix"`

portal-bridge/src/types/mode.rs
```diff
@@ -155,6 +159,30 @@ impl FromStr for ModeType {
...
+impl FromStr for FourFoursModeType {
+    type Err = ParseError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match &s[..1] {
+            "e" => {
+                let epoch = s[1..]
+                    .parse()
+                    .map_err(|_| "Invalid bridge mode arg: epoch number")?;
+                if epoch > 1896 {
+                    return Err("Invalid bridge mode arg: Four Fours only contains up to epoch 1896 was given: {epoch}");
```
njgheorghita, line 177: `Invalid 4444s bridge mode arg: era1 epoch greater than 1896 was given: {epoch}"`

portal-bridge/src/types/mode.rs
```diff
@@ -155,6 +159,30 @@ impl FromStr for ModeType {
     }
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum FourFoursModeType {
+    Random,
+    Epoch(u64),
```
njgheorghita, line 165: *nitpick* imo, `Single(u64)` here is a bit more intuitive. `Single` when compared to `Random` is fairly simple to understand. I think it's safe to assume that readers will understand that the `u64` is in reference to the epoch number, though it's worth leaving a short comment... `// Gossips a single epoch`

We also use `"single"` in the history bridge, so it's a bit more consistent

portal-bridge/src/types/mode.rs
```diff
@@ -77,7 +77,7 @@ impl FromStr for BridgeMode {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "latest" => Ok(BridgeMode::Latest),
-            "fourfours" => Ok(BridgeMode::FourFours),
+            "fourfours" => Ok(BridgeMode::FourFours(FourFoursModeType::Random)),
```
njgheorghita, line 80: It's worth adding a couple test cases to the test below to test this & the `fourfours:e*` use case

portal-bridge/src/bridge/era1.rs
```diff
@@ -79,50 +85,76 @@ impl Era1Bridge {
...
-                    .expect("to be able to acquire semaphore");
-                let serve_block_tuple_handle = Self::spawn_serve_block_tuple(
-                    self.portal_clients.clone(),
-                    block_tuple,
-                    epoch_acc.clone(),
-                    master_acc.clone(),
-                    permit,
-                );
-                serve_block_tuple_handles.push(serve_block_tuple_handle);
+    pub async fn launch_epoch(&self, epoch: u64) {
```
njgheorghita, line 105: `launch_single()`

portal-bridge/src/bridge/era1.rs
```diff
@@ -79,50 +85,76 @@ impl Era1Bridge {
...
+        let epoch_string = epoch.to_string() + "-";
+        for era1_path in self.era1_files.clone().into_iter() {
```
njgheorghita, line 107: I think something like this might be a bit neater. It's odd to `return` in the middle of a function while the last line of a function is a `panic!`. Also, anytime we can avoid nesting, I tend to think it's worthwhile. Imo nested statements are typically confusing to read

```rust
let era1_path = self.era1_files.clone().into_iter().find(|file| file == &epoch_string);
match era1_path {
    Some(path) => self.gossip_era1(path).await,
    None => panic!("...."),
}
```

njgheorghita (commented):
While I was working on #1155 I ran into some weird edge cases using the surf http client... So just want to make sure that you've tested the features locally & that they're all working?

KolbyML:
I was using this during my kurtosis test runs yesterday. As I wanted to remove the provider out of my equation when testing trin. I am interested what weird behavior dod you experience?

njgheorghita:
Imo, the weird behavior was proabably due to a slight difference in setup b/w your approach and mine. I don't think it's worth explaining here b/c it's somewhat complex / unrelated. Basically, just a final double check to make sure that gossiping `fourfours` & `fourfours:e*` works, since these paths are tough to test, is what I'm asking.

portal-bridge/src/bridge/era1.rs
```diff
@@ -79,50 +85,76 @@ impl Era1Bridge {
...
-                let serve_block_tuple_handle = Self::spawn_serve_block_tuple(
-                    self.portal_clients.clone(),
-                    block_tuple,
-                    epoch_acc.clone(),
-                    master_acc.clone(),
-                    permit,
-                );
-                serve_block_tuple_handles.push(serve_block_tuple_handle);
+    pub async fn launch_single(&self, epoch: u64) {
+        let epoch_string = format!("{epoch}-");
```
njgheorghita, line 104: Are you sure this is correct? I think it depends on how `Display` is implemented for `u64`. For example, we want to make sure epoch `mainnet-00001-*.era1` and `mainnet-00011-*.era` will be correctly found for an epoch value of `1`. I think you'll need to pad out this value with `0` with something like [this](https://stackoverflow.com/questions/50458144/what-is-the-easiest-way-to-pad-a-string-with-0-to-the-left)

KolbyML: display for u64 just converts the base 10 representation to a string.

I just pushed a chain which will make sure it is mainnet with the era1 file extension

portal-bridge/src/bridge/era1.rs
```diff
@@ -101,12 +101,11 @@ impl Era1Bridge {
...
-        let epoch_string = format!("{epoch}-");
-        let era1_path = self
-            .era1_files
-            .clone()
-            .into_iter()
-            .find(|file| file.contains(&epoch_string));
+        let era1_path = self.era1_files.clone().into_iter().find(|file| {
+            file.contains("mainnet-")
+                && file.contains(&format!("{epoch}-"))
+                && file.contains(".era1")
```
njgheorghita: This still doesn't fix the problem. For an epoch value of `1`, this case will be true for all files
- `mainnet-00001-hash.era1`
- `mainnet-00011-hash.era1`
- `mainnet-00101-hash.era`
- etc.

You need to left-pad the epoch value with up to 5 zeroes.

portal-bridge/src/bridge/era1.rs
njgheorghita: The problem isn't the `mainnet` or `.era1` file extension. It's the middle value of the path

portal-bridge/src/bridge/era1.rs
KolbyML: I just fixed that. but the example you gave isn't true
It would only return true for ``mainnet-00001-hash.era1``

KolbyML: `{epoch}-` this dash ties it to check the right

njgheorghita: yeah, but it's the left side of the epoch value that's the problem

KolbyML: image

here is proof

KolbyML: I just pushed a change which will make it more readable. But the code I had worked.
image

njgheorghita: - `mainnet-00001-hash.era1`
- `mainnet-00011-hash.era1`
- `mainnet-00101-hash.era1`

All of those strings "contain" the following substrings: `"mainnet"`, `"1-"`, and `".era1"`

KolbyML: Oh I see what you mean now. I fixed it which is good. My bad for the confusion xD.

KolbyML: My code did work assuming the list was sorted, but yeah the issue is the list wouldn't be sorted because it is shuffle. Sorry about that

njgheorghita (approved):

PR #1160, opened 2024-02-12
feat(state_network): update Nibbles encoding according to spec

morph-dev:
### What was wrong?

Nibbles encoding/decoding was no longer up to spec. Spec was updated in ethereum/portal-network-specs/pull/275

### How was it fixed?

Wrote encoding/decoding and updated the `portal-spec-tests` submodule (it has updated test vectors).

### To-Do


njgheorghita (approved):
:ship:

PR #1161, opened 2024-02-12
fix: use execute_batch for multi sql statement

### LINKED ISSUE #1156: Indexes are not created for SQL tables
It looks like [`CREATE_QUERY_DB`](https://github.com/ethereum/trin/blob/234b08945227f285ae95132e4e558e6a43ba9c62/trin-storage/src/sql.rs#L3C11-L3C26) creates table and indexes (also `LC_UPDATE_CREATE_TABLE`), but `rusqlite`'s `execute(...)` function only executes a single SQL statement ([source](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.execute)).

We should either create indexes in separate calls, or use `execute_batch`.

qiweiii:
### What was wrong?

Fix #1156

### DIFF
--- a/trin-storage/src/sql.rs
+++ b/trin-storage/src/sql.rs
@@ -8,10 +8,10 @@ pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                                 network INTEGER NOT NULL DEFAULT 0,
                                 content_size INTEGER
                             );
-                            CREATE INDEX content_size_idx ON content_data(content_size);
-                            CREATE INDEX content_id_short_idx ON content_data(content_id_short);
-                            CREATE INDEX content_id_long_idx ON content_data(content_id_long);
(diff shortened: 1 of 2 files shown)


morph-dev:
The documentation for the `execute_batch` ([link](https://docs.rs/rusqlite/latest/rusqlite/struct.Connection.html#method.execute_batch)) has an example where statement starts with `BEGIN;` and ends with `COMMIT;`.

Are you sure this actually fixes the issue?

qiweiii:
it works, after tesing, I can see the indices that does not exist before. BEGIN and COMMIT are for transactions, we can use them if we want to ensure both the table creation and index addition are treated as a single atomic operation, do we need them in this case?

morph-dev:
I don't think we need them as atomic operation. So this looks ok to me.

:rocket:

morph-dev (approved):

ogenev (approved):
Looks good 👍!

PR #1169, opened 2024-02-14
fix(bridge): include fluffy executable in docker image

njgheorghita:
### What was wrong?
In our bridge, we support using fluffy as the bridge client. This is done by passing an "executable path" via cli. This is possible when you run the bridge locally, but it's not possible when running a fluffy bridge via docker (eg. in kurtosis). 

Also, our method for adding bootstrap enrs to the fluffy executable was broken.

### How was it fixed?
- Included the fluffy executable inside the bridge docker image
- Fixed the bootstrap enrs cli arg formatting for fluffy

### To-Do

### DIFF
--- a/docker/Dockerfile.bridge
+++ b/docker/Dockerfile.bridge
@@ -45,6 +45,14 @@ COPY ./rpc ./rpc
 # build for release
 RUN cargo build -p trin -p portal-bridge --release
 
+# Get the fluffy executable.
+# If you're building on a different architecture,
+# you'll need to include the `--platform linux/amd64` flag
+# to the `docker build` command.
+# This isn't an issue for our ci pipeline, as it's already
+# building on an amd64 machine.
+FROM statusim/nimbus-fluffy:amd64-master-latest AS fluffy
+
 # final base
 FROM final_base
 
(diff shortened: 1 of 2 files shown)


KolbyML:
Our bridge requires clients to support trace_gossip to be used

Fluffy doesn't support trace gossip

KolbyML:
So if fluffy can't be used why support it in the bridge at all?

njgheorghita:
Ahhh, ok good to know. These were just the initial steps to getting the fluffy executable to run as a bridge client. I hadn't quite reached the part to validate that the client was successfully gossiping content yet. 

Ummm.. 
1. We could update the bridge to support simple `gossip` endpoints instead of `trace_gossip`.
2. I've heard @kdeme mention they're working on their own bridge. 
  - Is it likely that fluffy will support a `trace_gossip` endpoint down the road?
  - What's the status of your bridge? Would you be :+1: / :-1: on removing support for fluffy from this bridge implementation?
  
IIRC, we've decided for leaving fluffy support inside this bridge implementation in the past. Maybe this was before fluffy was making progress on their own bridge implementation? Obviously, this bridge's fluffy integration is broken / has been for some time. Now seems like a good time to revisit that decision.

KolbyML:
That is to say if we made the backfill bridge use gossip there would be no way to limit how many utp connections are happening at a time other then allocating time allocations you think it will be done in, but that would be fairly inaccurate prone to error and you are likely to have either too many or not enough open at a time. Where using trace_gossip we have guarantees of knowing when something is done or not.

njgheorghita:
Yeah, I think that's a good point. Imo, this depends on fluffy and what their plans are for their bridge / support of `trace_gossip`

njgheorghita:
closing this since the bridge no longer supports fluffy

PR #1171, opened 2024-02-14
refactor: replace PossibleContentValue with error according to spec u…

njgheorghita:
### What was wrong?
We are unable to distinguish between empty receipts & content not found values in our jsonrpc responses. It was decided to return an `Error` in the case of an unsuccessful content lookup. 

### How was it fixed?
Replaced `PossibleContentValue` strategy with `Error` based strategy for jsonrpc responses with failed content lookups. 

This pr is "sort of" blocked by an upcoming pr to the specs, in which we'll officially define the actual error message, code, and data fields. However, since this pr is very useful w/ regards to 4444s data availability measurement, imo, we are good to push it through, and then revisit any disparities b/w the spec & this implementation after the spec is finalized.

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/find.rs
+++ b/ethportal-peertest/src/scenarios/find.rs
@@ -8,8 +8,7 @@ use crate::{utils::fixture_header_with_proof, Peertest};
 use ethportal_api::{
     types::{history::ContentInfo, portal_wire::ProtocolId},
     utils::bytes::hex_decode,
-    BeaconNetworkApiClient, Enr, HistoryNetworkApiClient, OverlayContentKey,
-    PossibleHistoryContentValue, StateNetworkApiClient,
+    BeaconNetworkApiClient, Enr, HistoryNetworkApiClient, OverlayContentKey, StateNetworkApiClient,
 };
 
 pub async fn test_recursive_find_nodes_self(protocol: ProtocolId, peertest: &Peertest) {
@@ -106,10 +105,7 @@ pub async fn test_trace_recursive_find_content(peertest: &Peertest) {
     let content = trace_content_info.content;
     let trace = trace_content_info.trace;
 
-    assert_eq!(
-        content,
-        PossibleHistoryContentValue::ContentPresent(content_value)
-    );
+    assert_eq!(content, content_value);
 
     let query_origin_node: NodeId = peertest.nodes[0].enr.node_id();
     let node_with_content: NodeId = peertest.bootnode.enr.node_id();
@@ -143,14 +139,14 @@ pub async fn test_trace_recursive_find_content_for_absent_content(peertest: &Pee
     let client = &peertest.nodes[0].ipc_client;
     let (content_key, _) = fixture_header_with_proof();
 
-    let result = HistoryNetworkApiClient::trace_recursive_find_content(client, content_key)
+    let error = HistoryNetworkApiClient::trace_recursive_find_content(client, content_key)
         .await
-        .unwrap();
-
-    assert!(!result.utp_transfer);
-    assert_eq!(result.content, PossibleHistoryContentValue::ContentAbsent);
-    // Check that at least one route was involved.
-    assert!(result.trace.responses.len() > 1);
+        .unwrap_err()
+        .to_string();
+    assert!(error.contains("Content not found"));
+    // test that trace is present
+    assert!(error.contains("respondedWith"));
+    assert!(error.contains("-32001"));
 }
 
 pub async fn test_trace_recursive_find_content_local_db(peertest: &Peertest) {
@@ -173,10 +169,7 @@ pub async fn test_trace_recursive_find_content_local_db(peertest: &Peertest) {
     .await
     .unwrap();
     assert!(!trace_content_info.utp_transfer);
-    assert_eq!(
-        trace_content_info.content,
-        PossibleHistoryContentValue::ContentPresent(content_value)
-    );
+    assert_eq!(trace_content_info.content, content_value);
 
     let origin = trace_content_info.trace.origin;
     assert_eq!(trace_content_info.trace.received_from.unwrap(), origin);
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -134,6 +134,8 @@ pub enum OverlayCommand<TContentKey> {
 
 /// An overlay request error.
 #[derive(Clone, Error, Debug)]
+// required for the ContentNotFound error response
+#[allow(clippy::large_enum_variant)]
 pub enum OverlayRequestError {
     /// A failure to transmit or receive a message on a channel.
     #[error("Channel failure: {0}")]
@@ -184,6 +186,13 @@ pub enum OverlayRequestError {
 
     #[error("Received invalid remote discv5 packet")]
     InvalidRemoteDiscv5Packet,
+
+    #[error("Content wasn't found on the network: {message}")]
+    ContentNotFound {
+        message: String,
+        utp: bool,
+        trace: Option<QueryTrace>,
+    },
 }
 
 impl From<discv5::RequestError> for OverlayRequestError {
@@ -803,7 +812,11 @@ where
                 match query.into_result() {
                     FindContentQueryResult::ClosestNodes(_closest_nodes) => {
                         if let Some(responder) = callback {
-                            let _ = responder.send((None, false, query_info.trace));
+                            let _ = responder.send(Err(OverlayRequestError::ContentNotFound {
+                                message: "Content not found".to_string(),
+                                utp: false,
+                                trace: query_info.trace,
+                            }));
                         }
                     }
                     FindContentQueryResult::Content {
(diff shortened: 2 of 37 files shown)


trin-history/src/jsonrpc.rs
```diff
@@ -107,21 +107,40 @@ async fn recursive_find_content(
...
+        None => match overlay
+            .lookup_content(content_key.clone(), is_trace)
+            .await
+            .expect("XXXX fix this")
+        {
+            Ok((content_bytes, utp_transfer, trace)) => (content_bytes, utp_transfer, trace),
+            Err(err) => match err.clone() {
+                OverlayRequestError::ContentNotFound((_msg, _utp, _trace)) => {
+                    // this should be formatted...
+                    return Err(err.to_string());
```
njgheorghita, line 127: @ogenev I'm not sure how to handle this. If content is not found on the network, the overlay now returns this `OverlayRequestError::ContentNotFound()`. However, this method (`recursive_find_content()`) returns `Result<Value, String>`. But, inside `rpc/src/fetch.rs` we want to distinguish between `ContentNotFound` errors & regular errors (eg that only contain an error message), so that we can return the appropriate `RpcServeError` with its own code / message / data fields in the case that data is not available on the network.

It seems to me like the only option here is to serialize `OverlayRequestError::ContentNotFound` to a string here, and then **try** to deserialize the string to `RpcServeError::ContentNotFound` inside `rpc/src/fetch.rs` & if that fails, then handle it as a regular `RpcServeError::Message`? 

Any thoughts? Or suggestions on how to better approach this?

ogenev: Yeah, I think this would be the simplest approach - returning `String` here and then converting it to the appropriate error message in `fetch.rs`.

portalnet/src/overlay_service.rs
```diff
@@ -184,6 +186,9 @@ pub enum OverlayRequestError {
 
     #[error("Received invalid remote discv5 packet")]
     InvalidRemoteDiscv5Packet,
+
+    #[error("Content wasn't found on the network")]
+    ContentNotFound((String, bool, Option<QueryTrace>)),
```
morph-dev, line 191: Can you document what parameters are? Maybe doing something like this:
```
ContentNotFound {
  message: String,
  utp_transfer: bool,
  trace: Option<QueryTrace>,
}
```
You can also add `message` to the error string.

rpc/src/errors.rs
```diff
@@ -51,11 +52,14 @@ impl RpcError {
...
 }
 
+#[allow(clippy::large_enum_variant)]
 pub enum RpcServeError {
     /// A generic error with no data
     Message(String),
     /// Method not available
     MethodNotFound(String),
+    /// ContentNotFound
+    ContentNotFound((String, bool, QueryTrace)),
```
morph-dev, line 62: Same here.

rpc/src/serde.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::result_large_err)] // todo remove
```
morph-dev, line 1: comment

src/bin/poll_latest.rs
```diff
@@ -216,11 +213,13 @@ async fn audit_content_key(
 ) -> anyhow::Result<Instant> {
     let mut attempts = 0;
     while Instant::now() - timestamp < timeout {
-        let content = client.recursive_find_content(content_key.clone()).await?;
-        if let ContentInfo::Content { content, .. } = content {
-            if let PossibleHistoryContentValue::ContentPresent(_) = content {
+        match client.recursive_find_content(content_key.clone()).await? {
+            ContentInfo::Content {
+                content: _content, ..
```
morph-dev, line 218: Also here, I think we can remove: `content: _content`

src/bin/sample_range.rs
```diff
@@ -145,38 +144,32 @@ async fn audit_block(
...
-    match client.recursive_find_content(header_ck).await? {
-        ContentInfo::Content { content, .. } => {
-            if let PossibleHistoryContentValue::ContentPresent(_) = content {
-                metrics.lock().unwrap().header.success_count += 1;
-            } else {
-                warn!("Header not found for block {hash}");
-                metrics.lock().unwrap().header.failure_count += 1;
-            }
+    match client.recursive_find_content(header_ck).await {
+        Ok(_) => {
```
morph-dev, line 148: Doesn't this change the logic?
It could be some other kind of `ContentInfo` and we would count it as success, while previously we wouldn't.
Or am I missing something :thinking:

trin-beacon/src/jsonrpc.rs
```diff
@@ -106,21 +106,40 @@ async fn recursive_find_content(
...
             trace.node_responded_with_content(&local_enr);
-            (Some(val), false, if is_trace { Some(trace) } else { None })
+            (val, false, if is_trace { Some(trace) } else { None })
         }
-        None => overlay.lookup_content(content_key.clone(), is_trace).await,
+        // data is not available locally, make network request
+        None => match overlay
+            .lookup_content(content_key.clone(), is_trace)
+            .await
+            .expect("XXXX fix this")
```
morph-dev, line 120: reminder

morph-dev (commented):
I know that it's still in the draft phase, but still did a review.
Mostly nitpicks and reminders to address some comments that you left.

ethportal-peertest/src/scenarios/find.rs
```diff
@@ -143,14 +139,13 @@ pub async fn test_trace_recursive_find_content_for_absent_content(peertest: &Pee
...
         .await
-        .unwrap();
-
-    assert!(!result.utp_transfer);
-    assert_eq!(result.content, PossibleHistoryContentValue::ContentAbsent);
-    // Check that at least one route was involved.
-    assert!(result.trace.responses.len() > 1);
+        .unwrap_err()
+        .to_string();
+    assert!(error.contains("Content not found"));
```
njgheorghita, line 146: I spent a long time trying to figure out a way to access the underlying `ErrorObject` here, so we don't have to test via `string.contains("..")`. I wasn't able to figure it out, but if anybody has any ideas then this test can be improved significantly

src/bin/sample_range.rs
njgheorghita: The update in this pr is to return an `Error` in the case where a content lookup failed. In this script, all we really care about is whether the content lookup succeeded/failed, eg. we don't really care about the actual piece of content. Previously, we had to unpack the `ContentInfo` to make sure that it wasn't the "possible content value" variant. Now, all `ContentInfo`s are expected to be valid pieces of content, and all failed content lookups result in an `Error`, which we handle below.

ethportal-peertest/src/scenarios/find.rs
morph-dev: If I'm not wrong, the error type is `jsonrpsee_types::ErrorObjectOwned`, which means you can access code, which should be -32001.
I think checking for both code and `string.contains` is good enough.

Maybe I'm missing something.

rpc/src/fetch.rs
```diff
@@ -29,7 +28,27 @@ pub async fn proxy_query_to_history_subnet(
     match resp_rx.recv().await {
         Some(val) => match val {
             Ok(result) => Ok(result),
-            Err(msg) => Err(RpcServeError::Message(msg)),
+            Err(msg) => {
```
ogenev, line 31: I think, we also need to update this for [state](https://github.com/ethereum/trin/blob/ebc372affd7bc13d254f955d2517ad7adde1fdd8/rpc/src/state_rpc.rs#L31) network.

ogenev (approved):
⛵

rpc/src/fetch.rs
njgheorghita: None of the endpoints that would use this kind of error are actually supported inside the [state jsonpc handler](https://github.com/ethereum/trin/blob/master/trin-state/src/jsonrpc.rs#L32-L47) so for the sake of moving this pr ahead, I'm going to leave state alone. But this is definitely something that @morph-dev should keep in mind as he develops the state network's capabilities

morph-dev: Yes, that's correct. I finished PR #1175 that implements those endpoints for state network, but it's easier if you merge this one first, and I integrate changes there afterwards.

ethportal-peertest/src/scenarios/find.rs
njgheorghita: Yeah, to be honest, I think we're both missing something. My best guess is that there is some strange type-casting going on inside the automatic `rpc` scaffolding, but I haven't been able to figure it out. I also agree that it should be `ErrorObjectOwned` but all I see is `jsonrpsee_core::Error` which has no available methods ( except `to_string()`). I haven't been able to find a way to get at the underlying `ErrorObjectOwned`. But, I don't think it's worth any more effort since it only tidies up this code a bit. I added a check for `string.contains("-32001")` though, so all of our bases should be covered.

Issue #1157, opened 2024-02-10
New database design


morph-dev:
We currently store `content-key` and `content-value` as hex text, which occupies twice as much space as needed (each byte uses 2 characters).
We also observe the need to change/update scheme from time to time, and current design doesn't support that.

The goal of this issue is to discuss and design:
- new scheme that uses blobs (instead of text)
- migration from one scheme to another

morph-dev:
## Proposal [2]: Improve Pruning

Pruning currently works like this:

1. After every new content that is inserted, check if we are above storage capacity (+1 query)
2. Lookup for the furthest content id (+1 query)
3. If removing that content brings us below storage capacity (+1 query), don't remove it, update radius and finish pruning
4. Otherwise remove that content (+1 query) and go back to step 2

This leaves us always slightly above storage capacity all the time, resulting in steps 1-3 happening after every insert, and on average it would repeat steps 2-4. So on average ~6 queries after every insert. Query 1 is the most expensive (it has to process every row).

I propose to change pruning logic, by introducing new concepts:
- target capacity - aimed capacity when pruning, e.g. 90% of storage capacity
- pruning capacity - threshold when to start pruning, e.g. 95% of storage capacity
- `pruning_delay` - how many new inserts are needed in order to check if we should prune
- `average_content_size` - `total_content_size / content_count`
  - it will be inaccurate early on (which won't matter much), but it will be very accurate as we get closer to storage capacity

Logic would roughly look like this:
1. After every insert, reduce `pruning_delay` by one and finish if it is not yet zero (otherwise go to 2) (no queries)
2. Check if we are above pruning capacity (1 query). Pruning would look like this:
    1. estimate how many entries we have to remove to reach target capacity and remove that many farthest ones
        - together with query from step 2, this can take only one query (using `DELETE FROM ... ORDER BY ... LIMIT ...`)
        - repeat until we are at least under pruning capacity (we can be above target capacity)
    2. update radius to match current farthest entry that is stored (1 query)
3. Set `pruning_delay` to estimated number of entries that we need to insert in order to get to the storage capacity (100%) (1 query)
    1. to compensate for errors when database is small, don't make it more than current number of stored entries

This would result in not doing any additional queries most of the time, only periodically checking if we should prune (with pruning being most frequent case) and prune larger amount less frequently and more efficiently.

Also, when we report metrics (i.e. on heartbeat), we can update average_content_size and pruning delay for free.

ogenev:
[...]

PR #1172, opened 2024-02-15
feat(storage): add core enums, traits and functions for versioned storage

morph-dev:
### What was wrong?

Our storage system needs update. See #1157 for details.

### How was it fixed?

Created core types for the new versioned storage system:
- `ContentType` - the type of content that is stored
- `StoreVersion` - the version of the store used
- `VersionedContentStore` - the trait for the actual content store
  - there should be at most one `VersionedContentStore` for each `StoreVersion` (ideally it should be exactly one, but we are fine missing some implementations until we migrate all existing implementations)
- `MemoryContentStore` - in memory implementation of the `VersionedContentStore`
- `create_store` function - creates instance of the `VersionedContentStore` and migrates from previous version (if needed and migration is implemented)

It's worth noting that one network might want to use multiple `VersionedContentStore` (for example, this might be the case for beacon), in which case it will need some custom logic on top of it.

New schema (different table schema and different prune logic) will be done in the followup PR.

### To-Do


trin-storage/src/versioned/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+use rusqlite::types::{FromSql, FromSqlError, ValueRef};
+use strum::{AsRefStr, Display, EnumIter, EnumString};
+
+pub use store::VersionedContentStore;
+pub use utils::create_store;
+
+#[derive(Copy, Clone, Debug, Display, Eq, PartialEq, EnumString, AsRefStr, EnumIter)]
+#[strum(serialize_all = "snake_case")]
+pub enum ContentType {
+    Beacon,
```
morph-dev, line 15: These are mostly placeholders for now. Their purpose is to define the type of content that is stored in one _store_, and for store to differentiate between types (e.g. different column in a table, or completely different table name).
@ogenev Maybe Beacon should have multiple types?

trin-storage/src/versioned/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+    History,
+    State,
+}
+
+#[derive(Copy, Clone, Debug, Display, Eq, PartialEq, EnumString, AsRefStr, EnumIter)]
+#[strum(serialize_all = "snake_case")]
+pub enum StoreVersion {
+    InMemory,
+    LegacyContentData,
```
morph-dev, line 24: `Legacy*` is also placeholder for now. We should have implementation of `VersionedContentStore` that uses current logic. I'm fine with removing them until we have implementations in place (not any time soon).

trin-storage/src/versioned/store.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Migrates content from previous version to the new version.
+    fn migrate_from(
+        content_type: ContentType,
+        old_version: StoreVersion,
+        config: &PortalStorageConfig,
+    ) -> Result<(), ContentStoreError>;
+
+    /// Creates the instance of the store. This shouldn't be used directly. Store should be
+    /// created using `create_store` function.
+    fn create(
```
morph-dev, line 25: I don't like that anybody can just call this (as comments says, `create_store` function should be used). But I don't know how to enforce that.

I'm open for suggestions

trin-storage/src/versioned/utils.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Ensures that the correct version of the content store is used (by migrating the content if
+/// that's not the case).
+pub fn create_store<S: VersionedContentStore>(
+    content_type: ContentType,
+    network: ProtocolId,
+    config: PortalStorageConfig,
+) -> Result<S, ContentStoreError> {
+    let conn = config.sql_connection_pool.get()?;
+
+    let old_version = lookup_store_version(content_type, &conn)?;
```
morph-dev, line 20: We should probably have some logic to detect legacy version (e.g. `Legacy*`).

It's not important until we have implementations.

Once this PR is merged, I will create an Issue to track this.

morph-dev (commented):
I added few comments.
Let me know if you think I should add some comments in the code as well.

trin-storage/src/versioned/store.rs
```diff
@@ -0,0 +1,107 @@
...
+        content_type: ContentType,
+        config: PortalStorageConfig,
+        metrics: StorageMetricsReporter,
+    ) -> Result<Self, ContentStoreError>;
+
```
morph-dev, line 30: I'm now thinking that `VersionedContentStore` doesn't have to have any of the functions below except maybe `get_summary_info`. That will allow more flexibility on what store can store and do (and work better with `LC_UPDATE` from beacon network)

trin-storage/src/versioned/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    Beacon,
+    History,
+    State,
+}
+
+#[derive(Copy, Clone, Debug, Display, Eq, PartialEq, EnumString, AsRefStr, EnumIter)]
+#[strum(serialize_all = "snake_case")]
+pub enum StoreVersion {
+    LegacyContentData,
+    IdIndexed,
```
njgheorghita, line 23: I'm not sure I follow the vision here, if you can help me understand. For each new "version" we would add an entry to the `StoreVersion` enum? Is this better than just using numbers or semver(might be overkill)? If we just use these "strings" to identify each version...
- it is helpful to understand the changes included in the update (though, this can be accomplished with comments)
- it's not intuitive to understand the order of the upgrades (though, this can be accomplished with comments)
I'm not sure, it's just seems like using something like `v0`, `v1` .. `v100` is a bit more intuitive

trin-storage/src/versioned/sql.rs
```diff
@@ -0,0 +1,14 @@
+pub const STORE_INFO_CREATE_TABLE: &str = "
+    CREATE TABLE IF NOT EXISTS store_info (
+        content_type TEXT PRIMARY KEY,
+        version TEXT NOT NULL
+    )";
```
njgheorghita, line 5: Is it worth adding a timestamp here, so that we can identify when the db was last updated?

morph-dev (commented):
I left more detailed explanation about my idea on one comment in the `trin-storage/src/versioned/mod.rs` file.

I'm happy to have a video call and discuss things in order to move faster with this.

ogenev:
I don't understand how this migration system would work in practice. Here is how I imagine this:

```rust
use rusqlite::{params, Connection, Result};

struct Migration {
    version: i64,
    up: &'static str,
}

const MIGRATIONS: &[Migration] = &[
    Migration {
        version: 1,
        up: "CREATE TABLE IF NOT EXISTS  content_store (version INTEGER PRIMARY KEY)",
    },
    Migration {
        version: 2,
        up: "CREATE TABLE IF NOT EXISTS table_name1 (...)",
    },
    Migration {
        version: 3,
        up: "modify content_store table (...)",
    },
    Migration {
        version: 4,
        up: "CREATE TABLE IF NOT EXISTS table_name2(...)",
    },
    // Add more migrations here as needed.
];

pub fn migrate(conn: &Connection) -> Result<()> {
    let tx = conn.transaction()?;

    tx.execute(
        "CREATE TABLE IF NOT EXISTS content_store (version INTEGER PRIMARY KEY)",
        params![],
    )?;

    let current_version: i64 = tx
        .query_row("SELECT version FROM content_store ORDER BY version DESC LIMIT 1", params![], |row| row.get(0))
        .optional()?
        .unwrap_or(0);

    for migration in MIGRATIONS {
        if migration.version > current_version {
            tx.execute(migration.up, params![])?;
            tx.execute(
                "INSERT INTO content_store (version) VALUES (?1)",
                params![migration.version],
            )?;
        }
    }

    tx.commit()
}
```

Maybe we are trying to accomplish a similar thing here but I'm not sure if I understand it.

morph-dev:
I don't think we are trying to accomplish similar things.

In my model, you have to manually write transition from `VersionA` to `VersionB` (but you can reuse and chain them if that makes sense).

The way that I envisioned is that migration from one version to another can be quite complicated and not always possible just via SQL. For example, to migrate current `content_data` table to something described in #1157, I would say we would have to:
- create new table for the new version
- read every entry from the old table
- convert `content-value` from hex TEXT to BLOB
- calculate distance from `NodeId`
- write entry to the new database
- delete old table
- update version in the `store_info` table
(in this particular case, we might be able to do everything via SQL but that might not always be the case).

This process can take a long time, but I don't think that's the issue as it would happen on start. The only issue is that if program is stopped mid transition (but I think we can deal with that if we make sure new table is empty when we start).

Also, as I said in the other comment, Version doesn't mean that it's used only for one type (regular content key/value data). Storing LC_Updates can/should be one store Version as well.
It also doesn't imply that it has to use SQLite (e.g. we decide to go back to RocksDB, or use Redis, or even simple file (i.e. MasterAccumulator)).

trin-storage/src/versioned/mod.rs
ogenev: We currently store four types for beacon:  
- `LightCllentBootstrap` (in `content_data` table, but we will need to purge any values older than the subjectivity period (a few months))
- `LightClientUpdate` (in `lc_update`  table)
- `LightClientOptimisticUpdate/LightClientFIialityUpdate` (those are stored  in a cache, because we keep only the latest values).

trin-storage/src/versioned/utils.rs
```diff
@@ -0,0 +1,235 @@
...
+    config: PortalStorageConfig,
+) -> Result<S, ContentStoreError> {
+    let conn = config.sql_connection_pool.get()?;
+
+    let old_version = lookup_store_version(&content_type, &conn)?;
+
+    match old_version {
+        Some(old_version) => {
+            // Migrate if version doesn't match
+            if S::version() != old_version {
```
ogenev, line 25: Do we want to migrate only from an older version to the new version, i.e. 
```suggestion
            if S::version() > old_version {
```

ogenev (approved):
:shipit:

trin-storage/src/versioned/utils.rs
morph-dev: The versions are not strictly ordered. In this context, the `old_version` should be interpreted as "previous version".

If someone tries to migrate from newer to the older version, the `migrate_from` function should fail.

trin-storage/src/versioned/mod.rs
morph-dev: So for the `LightCllentBootstrap`, there is no concept of distance (in deciding whether we store) or storage limit?

Meaning, we should store all fresh content and discard/purge all older content, right?

trin-storage/src/versioned/sql.rs
njgheorghita: I don't have a great reason... Maybe just to check when the db was last migrated? But I'm not sure that's a strong reason, since the version to which it was last migrated is already available, and that's the important information. So feel free to disregard

njgheorghita (approved):
Cool! Excited to see how this works out in state network-land

PR #1173, opened 2024-02-16
perf(history storage): decrease `prune_db()` loop iteration time by 40-50s on 8gb database

KolbyML:
### What was wrong?
The main performance bottleneck in `prune_db()` was calling `get_total_storage_usage_in_bytes_from_network()` twice per iteration of the loop in `prune_db()` (once in `does_eviction_cause_under_capacity()` and once in `capacity_reached()`). Which was taking 40-50 seconds in total (2 x 20-25s). We should only be calling `get_total_storage_usage_in_bytes_from_network()` at the initialization of `HistoryStorage` and maintaining the count from their in memory.

This 40-50 second result was taken from one of our testnet nodes with a ~8GB sqlite database doing the inital prune_db() at startup initalization.
### How was it fixed?
We only insert data or evict data in 2 spots in the code. So we can just keep a count of how much storage we use in memory and increment it or decrement it from there. We will only call `get_total_storage_usage_in_bytes_from_network()` once during initialization of the database.

### DIFF
--- a/trin-history/src/storage.rs
+++ b/trin-history/src/storage.rs
@@ -29,7 +29,7 @@ pub struct HistoryStorage {
     node_id: NodeId,
     node_data_dir: PathBuf,
     storage_capacity_in_bytes: u64,
-    network_content_storage_used: u64,
+    storage_occupied_in_bytes: u64,
     radius: Distance,
     sql_connection_pool: Pool<SqliteConnectionManager>,
     distance_fn: DistanceFunction,
@@ -101,14 +101,14 @@ impl HistoryStorage {
             distance_fn: config.distance_fn,
             metrics,
             network: protocol,
-            network_content_storage_used: 0,
+            storage_occupied_in_bytes: 0,
         };
 
         // Set the metrics to the default radius, to start
         storage.metrics.report_radius(storage.radius);
 
         // Set the network content storage used at start
-        storage.network_content_storage_used =
+        storage.storage_occupied_in_bytes =
             storage.get_total_storage_usage_in_bytes_from_network()?;
 
         // Check whether we already have data, and use it to set radius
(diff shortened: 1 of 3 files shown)


trin-history/src/storage.rs
```diff
@@ -218,10 +217,15 @@ impl HistoryStorage {
         let content_key: Vec<u8> = key.clone().into();
         // store content key w/o the 0x prefix
         let content_key = hex_encode(content_key).trim_start_matches("0x").to_string();
+        let value_size = value.len() as u64;
         if let Err(err) = self.db_insert(&content_id, &content_key, value) {
             debug!("Error writing content ID {content_id:?} to db: {err:?}");
             return Err(err);
         } else {
+            // we are inserting into the database increase total network storage count
```
mrferris, line 225:
```suggestion
            // Insertion successful, increase total network storage count
```

trin-history/src/storage.rs
```diff
@@ -261,6 +267,10 @@ impl HistoryStorage {
             if let Err(err) = self.evict(id_to_remove) {
                 debug!("Error writing content ID {id_to_remove:?} to db: {err:?}",);
             } else {
+                // we are evicting content from the database decrease total network storage count
```
mrferris, line 270:
```suggestion
                // Eviction successful, decrease total network storage count
```

mrferris (approved):

morph-dev (approved):

trin-history/src/storage.rs
```diff
@@ -642,11 +641,11 @@ pub mod test {
             let value: Vec<u8> = vec![0; 32000];
```
njgheorghita, line 641: Can we add a check for `network_content_storage_used` attribute here? Just check that if in each iteration, it is the expected size. If we can also add a check inside a test to make sure that it's decremented appropriately, that would be great

trin-history/src/storage.rs
```diff
@@ -29,6 +29,7 @@ pub struct HistoryStorage {
     node_id: NodeId,
     node_data_dir: PathBuf,
     storage_capacity_in_bytes: u64,
+    network_content_storage_used: u64,
```
njgheorghita, line 32: I don't love this attribute name... Imo, `network` is somewhat redundant, since we're already inside `HistoryStorage` & imo that implies that all content that we're dealing with is already from the history network. I also think there's a better word than `_used` that we can "use" (lol the irony).

njgheorghita: Maybe... if you agree.... `storage_capacity_in_bytes` also feels somewhat redundant / probably made more sense when we had a single `Storage` struct. What about...
`storage_capacity_in_bytes` -> `bytes_capacity`
`network_content_storage_used` -> `bytes_occupied`

trin-history/src/storage.rs
```diff
@@ -218,10 +217,15 @@ impl HistoryStorage {
         let content_key: Vec<u8> = key.clone().into();
         // store content key w/o the 0x prefix
         let content_key = hex_encode(content_key).trim_start_matches("0x").to_string();
+        let value_size = value.len() as u64;
         if let Err(err) = self.db_insert(&content_id, &content_key, value) {
```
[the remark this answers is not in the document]
KolbyML: We do check upstream if the key is already inserted. But you are right it is free to check the status of the result as well here good catch!.

trin-history/src/storage.rs
KolbyML: I like `storage_capacity_in_bytes`/`storage_allocated_in_bytes` as it makes it clear what the variable is without relying on knowing what struct you are in. 

`bytes_occupied` of what? having `storage_capacity` prefix makes it more readable and less ambigious. Expessially for people less familar with the codebase.

PR #1174, opened 2024-02-16
fix: add network column to indices

morph-dev:
### What was wrong?

Some of our queries were very slow because our db indices weren't setup correctly.
For example, the `TOTAL_DATA_SIZE_QUERY_DB` query took 20-25 seconds on 4GB database.

### How was it fixed?

I added `network` column to the indices. Based on Kolby's measurements on the same db, we noticed following improvements:

| query | before (sec) | after (sec) |
|---|---|---|
| TOTAL_DATA_SIZE_QUERY_DB | 23.087 | 0.008 |
| XOR_FIND_FARTHEST_QUERY_NETWORK | 6.154 | 0.017 |

### To-Do

### DIFF
--- a/trin-storage/src/sql.rs
+++ b/trin-storage/src/sql.rs
@@ -1,7 +1,7 @@
 // SQLite Statements
 
 // NOTE: The indices `content_size_idx`, `content_id_short_idx` and `content_id_long_idx` weren't
-// felpful mostly because they didn't have `network` column in them.
+// helpful mostly because they didn't have `network` column in them.
 pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                                 content_id_long TEXT PRIMARY KEY,
                                 content_id_short INTEGER NOT NULL,
--- a/trin-storage/src/sql.rs
+++ b/trin-storage/src/sql.rs
@@ -1,5 +1,7 @@
 // SQLite Statements
 
+// NOTE: The indices `content_size_idx`, `content_id_short_idx` and `content_id_long_idx` didn't
+// have `network` column in them and weren't very helpful with the queries.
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
Amazing find 🚀.

mrferris (approved):

KolbyML:
Looking at our sql queries should we keep the index for content_id_long?

morph-dev:
I looked more online and again into our queries and made few changes.

> Looking at our sql queries should we keep the index for content_id_long?

I confirmed both by manual inspection and by [this post](https://sqlite.org/forum/forumpost/9b7ed07018) that index is not needed for the primary key or unique column (and it might hurt) because they are automatically created.

So I also dropped the index for primary key from `lc_update` table.

I also figured out that:
- `content_id_short_idx_2` should include `content_id_long`
  - alternative is to put `LIMIT 1` to the `XOR_FIND_FARTHEST_QUERY_NETWORK` query. Currently we only read first item, but potentially we could read more and delete more in a batch in which case the index is useful
- the `(network, content_id_long_idx)` was not needed. Instead I simplified the `TOTAL_ENTRY_COUNT_QUERY_NETWORK` query so it can rely on `network_idx`
- the we needed index for `content_key` because of the `PAGINATE_QUERY_DB` query (this one shouldn't have network column)

Please take another look

trin-storage/src/sql.rs
```diff
@@ -1,5 +1,7 @@
 // SQLite Statements
 
+// NOTE: The indices `content_size_idx`, `content_id_short_idx` and `content_id_long_idx` weren't
+// felpful mostly because they didn't have `network` column in them.
```
njgheorghita, line 4: `helpful`

njgheorghita (approved):
🚀

KolbyML (approved):

PR #1175, opened 2024-02-17
feat(state): implement jsonrpc for the state network

morph-dev:
### What was wrong?

A lot of json_rpc methods weren't implemented for the state network.

### How was it fixed?

Implemented missing methods.
Most of them were implemented in a similar/same way as the history network.

NOTE: Currently, none of these methods would work because storage for the state network is not implemented. However, I do have some version of the storage implemented in the local branch and methods that I tested work fine.

### To-Do

### DIFF
--- a/trin-state/src/jsonrpc.rs
+++ b/trin-state/src/jsonrpc.rs
@@ -4,6 +4,8 @@ use discv5::{enr::NodeId, Enr};
 use portalnet::overlay_service::OverlayRequestError;
 use serde_json::{json, Value};
 use tokio::sync::mpsc;
+use tracing::error;
+use trin_storage::ContentStore;
 
 use crate::network::StateNetwork;
 use ethportal_api::{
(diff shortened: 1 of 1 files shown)


trin-state/src/jsonrpc.rs
```diff
@@ -42,7 +50,43 @@ impl StateRequestHandler {
                 recursive_find_nodes(network, node_id).await
             }
             StateEndpoint::DataRadius => radius(network),
-            _ => Err("Not implemented".to_string()),
+            StateEndpoint::LocalContent(content_key) => local_content(network, content_key).await,
+            StateEndpoint::FindContent(enr, content_key) => {
+                find_content(network, enr, content_key).await
+            }
+            StateEndpoint::RecursiveFindContent(content_key) => {
+                recursive_find_content(network, content_key, /* is_trace= */ false).await
```
ogenev, line 58: Nice, didn't know we could inline comments this way 💡

ogenev (approved):
⛵

trin-state/src/jsonrpc.rs
njgheorghita: As I've mentioned in a previous pr, I don't love it. But, I guess in this case especially I see the usefulness. And creating a `let is_trace=false;` variable does seem overkill. So, maybe I'll have to change my mind about this

njgheorghita (commented):
These feel like the kind of changes that should be accompanied with tests in `ethportal-peertest`. Even if the tests aren't really testing anything other than the endpoint is supported

njgheorghita (approved):
Oh, ok, just saw your comment in the pr description about the local testing. Fine to move ahead with this, though as soon as some basic state storage is supported, we should remember to also add tests for these endpoints

trin-state/src/jsonrpc.rs
morph-dev: They might look weird at first, but they grow on you.

They shouldn't be overused, but they have their use cases.

KolbyML (approved):

PR #1176, opened 2024-02-18
feat: content valid for storing and recursive gossip

morph-dev:
### What was wrong?

Implementation of `Validator` and `OverlayService` weren't flexible enough for the state network.

For the state network, content-value that is sent via `Offer/Accept` and `Find/Found Content` is not the same, and doesn't have the same validation properties. The `Offer/Accept` should always be provable that is part of a canonical chain, while that's not the case for the content from `Find/Found Content` (we can prove that content corresponds to the content-key).

Another issue is that Recursive Gossip that is part of the state network was not possible.

### How was it fixed?

Created `ValidationResult` object that is returned by the validator that contains all important information. Overlay Service now knows if content is canonical and it's safe for storing locally, and if there is additional content to be propagated (to be used for Recursive Gossip).

There should be no changes for current implementations (they all return the same result and everything should behave as before). Added logic will be used once validator for the state network is implemented.

NOTE: I would also like to add more validation metrics to grafana, but I would prefer to do that in a separate PR (maybe only once state network is implemented so I can test more functionality).

### To-Do

### DIFF
--- a/portalnet/src/overlay.rs
+++ b/portalnet/src/overlay.rs
@@ -51,7 +51,7 @@ use ethportal_api::{
 };
 use trin_metrics::{overlay::OverlayMetricsReporter, portalnet::PORTALNET_METRICS};
 use trin_storage::ContentStore;
-use trin_validation::validator::Validator;
+use trin_validation::validator::{ValidationResult, Validator};
 
 use crate::events::EventEnvelope;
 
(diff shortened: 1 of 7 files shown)


ogenev (approved):
Looks good 👍

njgheorghita (approved):
Nice comments! They were very useful

PR #1177, opened 2024-02-19
feat: add storage processing time metrics

KolbyML:
### What was wrong?
We didn't have a clear way to see if storage calls were performing well or not. This caused us to make a major regression in performance.
### How was it fixed?

By adding grafana metrics on how long different storage calls take. I think this is the first step in getting well rounded performance metrics throughout our codebase.

image
In a future PR I would want to make a graph which says which network is using the most I/O and by how much.


trin-history/src/storage.rs
```diff
@@ -182,11 +182,15 @@ impl HistoryStorage {
     }
 
     fn total_entry_count(&self) -> Result<u64, ContentStoreError> {
+        let timer = self
+            .metrics
+            .start_storage_process_timer("total_entry_count");
         let conn = self.sql_connection_pool.get()?;
```
morph-dev, line 188: The metric would capture the 'self.sql_connection_pool.get()' call.
My understanding is that here we are waiting to get a connection to the db. Is there a limit on number of connection and sometimes we wait here or is it usually instant?

I'm not sure this should qualify as time spent as part of this storage operation (in case db is blocked by other queries).

morph-dev (commented):
Few concerns, but I'm fine with going with this for now (since it's better than not having anything).

Also, I didn't look at grafana config since I know nothing about it.

trin-history/src/storage.rs
KolbyML: There is a read write lock over storage instance. If `self.sql_connection_pool.get()`. This is important information if it is blocking then that is important information.

We want to know if these calls are slow `self.sql_connection_pool.get()` is apart of that

morph-dev: Doesn't read-write lock allow multiple (unlimited?) reads at the same time? If that's the case, and `sql_connection_pool` has a limit, then we could be blocked on one request due to other requests taking long time.

And if that's desired depends on how you interpret the graphs. If there is congestion, all query metrics might go up while in fact problem can be with only one type of query.

If you time the requests without getting connection from the pool, you know how long each request takes and if there is congestion you will see exactly what is causing it. It's true that in this case you don't know how affected other queries are, so there is a trade off.

Either way, I'm fine with whatever you decide.

morph-dev (approved):

trin-history/src/storage.rs
morph-dev: The `r2d2` crate that we use [has a max number of connections](https://docs.rs/r2d2/0.8.10/r2d2/struct.Builder.html#method.max_size).

trin-metrics/src/storage.rs
```diff
@@ -75,6 +86,17 @@ pub struct StorageMetricsReporter {
 }
 
 impl StorageMetricsReporter {
+    pub fn start_storage_process_timer(&self, storage_function: &str) -> HistogramTimer {
```
njgheorghita, line 89: `start_process_timer()` & `stop_process_timer` below

trin-metrics/src/storage.rs
```diff
@@ -21,6 +23,14 @@ const BYTES_IN_MB_F64: f64 = 1000.0 * 1000.0;
 
 impl StorageMetrics {
     pub fn new(registry: &Registry) -> anyhow::Result<Self> {
+        let storage_process_time = register_histogram_vec_with_registry!(
```
njgheorghita, line 26: `let process_timer = ...`

trin-history/src/storage.rs
```diff
@@ -342,30 +347,38 @@ impl HistoryStorage {
...
         match result?.first() {
             Some(val) => Ok(Some(val.into())),
             None => Ok(None),
         }
     }
 
-    /// Public method for looking up a content value by its content id
-    pub fn lookup_content_value(&self, id: [u8; 32]) -> anyhow::Result<Option<Vec<u8>>> {
+    /// Internal method for looking up a content value by its content id
+    fn lookup_content_value(&self, id: [u8; 32]) -> anyhow::Result<Option<Vec<u8>>> {
```
njgheorghita, line 358: > If something fails nothing is recorded. We have logs for that.

Can you elaborate on that here? If `lookup_content_value()` fails, then it's not obvious to me where we're emitting these logs. Also, just to clarify, if the process timer is **never** stopped, then nothing is recorded?

njgheorghita (commented):
We need to think about how we can best solidify our grafana json template workflow. Here are the following places (that I can think of right now) where we use the `trin-metrics` grafana template.
- trin-bench
- kurtosis
- running metrics locally
If we're adding a graph / making changes to the dashboard, we want to make sure that the graph is being updated / deployed in all 3 situations, while maintaining the appropriate data source ids.

trin-history/src/storage.rs
KolbyML: image
> Also, just to clarify, if the process timer is never stopped, then nothing is recorded?

To my knowledge yes

trin-metrics/src/storage.rs
```diff
@@ -21,6 +23,14 @@ const BYTES_IN_MB_F64: f64 = 1000.0 * 1000.0;
...
+            histogram_opts!(
+                "trin_storage_process_time",
```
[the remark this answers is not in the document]
KolbyML: In grafana we calculate seconds from 2 of the histograms metrics `rate(trin_storage_process_time_sum[1m])/rate(trin_storage_process_time_count[1m])` so I don't think it is appropiate to add `_secs` as it isn't that simple

KolbyML:
> We need to think about how we can best solidify our grafana json template workflow. Here are the following places (that I can think of right now) where we use the `trin-metrics` grafana template.
> 
> * trin-bench
> * kurtosis
> * running metrics locally
>   If we're adding a graph / making changes to the dashboard, we want to make sure that the graph is being updated / deployed in all 3 situations, while maintaining the appropriate data source ids.

I am not exactly answering this question, but I think this point should be made clear.

Currently for a single node view and a multi node view we use the same dashboard. I think we should instead support 2
A single node view dashboard: more in depth specific stats
And a multi node view dashboard: An arrgrated overview

We already have a multi node dashboard, but we currently don't have a dashboard optimized to get an indepth look at 1 node

KolbyML:
Ready for  another look @njgheorghita

trin-history/src/storage.rs
KolbyML: This error then gets forwarded to the overlay which then reports the error

trin-history/src/storage.rs
KolbyML: Maybe we can report a new metric on how many connections is being used then in a future PR. If we are hitting the max number of connections and that is causing delays I think there is a problem with our code. My goal with these stats is to see how long it takes to execute the functions storage. Getting a connection to the database is included in that as that is what the code calling storage functions has to deal with.

trin-history/src/storage.rs
njgheorghita: > This timer can be stopped and observed at most once, either automatically (when it goes out of scope) or manually. Alternatively, it can be manually stopped and discarded in order to not record its value.

My understanding from the [docs](https://docs.rs/prometheus/latest/prometheus/struct.HistogramTimer.html) is that if it goes out of scope (eg. what happens in this case when `lookup_content_value()?` fail) then it will record the time when it is dropped and count it as an observation. Seems like we probably need to handle these cases and use `stop_and_discard()` on failures? Otherwise we'll end up with false observations skewing the histogram?

KolbyML: Good catch I should I should have read the documentation a bit more :laughing:.

I just pushed a fix for this issue by implementing a CustomHistogramTimer which on drop does `stop_and_discard()` instead of `stop_and_record()`.

The reason I choose to fix the problem in this way, is because I did a lot of thinking and came to the conclusion this was the cleanest way to achieve our goal well staying within the structural designs of our project in relation to how we handle metrics

KolbyML:
@njgheorghita I am ready for another review cargo-fmt ci broke because of an upstream change to the nightly docker excutor we we in ci here is a pr for that https://github.com/ethereum/trin/pull/1179

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl Drop for CustomHistogramTimer {
+    fn drop(&mut self) {
+        if !self.observed {
+            self.observe(false);
+        }
+    }
+}
```
njgheorghita, line 74: Now that we're implementing custom logic, this seems worth adding some unit tests to validate the new functionality

KolbyML: Ok I added a test. I took the pre-existing test promethus has for this and modified it to work for our `CustomHistogramTimer` 
https://docs.rs/prometheus/0.13.3/src/prometheus/histogram.rs.html#1217-1260

I also added 3 comments which describe the functionality which is special aka not counting on drop

trin-history/src/storage.rs
```diff
@@ -268,7 +273,7 @@ impl HistoryStorage {
                 "Capacity reached, deleting farthest: {}",
                 hex_encode(id_to_remove)
             );
-            if let Err(err) = self.evict(id_to_remove) {
+            if let Err(err) = self.db_remove(&id_to_remove) {
```
njgheorghita, line 276: It looks like the `debug` log below is inaccurate.
```
debug!("Error removing content ID {id_to_remove:?} from db: {err:?}");
```

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
+use std::time::Instant;
```
njgheorghita, line 1: *nitpick* this feels like more than a simple `util` imo, and deserves its own named module, something like `trin-metrics/src/timer.rs`

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+
+use prometheus_exporter::prometheus::Histogram;
+
+/// Timer to measure and record the duration of an event.
+///
+/// This timer can be stopped and observed at most once manually.
+/// Alternatively, if it isn't manually stopped it will be discarded in order to not record its
+/// value.
+#[must_use = "Timer should be kept in a variable otherwise it cannot observe duration"]
```
njgheorghita, line 10: Nice! Didn't know about this

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+#[derive(Debug)]
+pub struct CustomHistogramTimer {
```
njgheorghita, line 12: This name isn't great imo. The `Custom` prefix doesn't really give any information about the actual type. aka. All it really communicates is that this is a "custom" implementation. I think a better name would answer why are we using a "custom" implementation. But... this is a tricky one. I'm not coming up with any better names tbh. `DropableHistogramTimer` (well `dropable` isn't a word, and the regular histogram timer is also "dropable") / `UnobservedDropHistogramTimer` / `HistogramTimerWithDiscardedDrop` / `HistogramTimerWithUnobservedDrops`....

Idk, the last one is probably my favorite, but I'll leave this one up to you, at the very least the docstring above is very useful for understanding the purpose of the custom timer. Ideally, the name would also communicate the purpose of the custom timer, but there is a tradeoff here against the wordiness of the name.

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+    ///
+    /// It returns a floating-point number of seconds elapsed since the timer started,
+    /// without recording to any histogram.
+    pub fn stop_and_discard(self) -> f64 {
+        let mut timer = self;
+        timer.observe(false)
+    }
+
+    fn observe(&mut self, record: bool) -> f64 {
+        let v = Instant::now().saturating_duration_since(self.start);
```
njgheorghita, line 57: We should avoid single letter variables

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+
+        // In this thread the timer isn't manually closed so its result will be dropped
+        let timer = CustomHistogramTimer::new(histogram.clone());
+        let handler = thread::spawn(move || {
+            let _timer = timer;
+            thread::sleep(Duration::from_millis(400));
+        });
+        assert!(handler.join().is_ok());
+
+        let mut mfs = histogram.collect();
```
njgheorghita, line 108: Hahah, all I can think of here is `motherf***ers`... Maybe that's just me? Imo it's worthwhile to update it to `metric_families` and the following `mf` -> `metric_family`, `m` -> `metric`... It makes the code more readable and avoids single letter variables

trin-metrics/src/utils.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    use prometheus_exporter::prometheus::{
+        core::Collector, Histogram, HistogramOpts, DEFAULT_BUCKETS,
+    };
+
+    use crate::utils::CustomHistogramTimer;
+
+    /// Test taken from https://docs.rs/prometheus/0.13.3/src/prometheus/histogram.rs.html#1217-1260
+    /// Modified to work with CustomHistogramTimer
+    #[test]
```
njgheorghita, line 88: I'm not sure about this. The point of testing our implementation of `CustomHistogramTimer` is not to make sure that the wrapped `HistogramTimer` functions properly (which is what this test is testing iiuc). But to make sure that our custom api is working as expected. I'd like to see a test that tests `CustomHistogramTimer::observe_duration`, `CustomHistogramTimer::stop_and_record`, `CustomHistogramTimer::stop_and_discard`, and `CustomHistogramTimer::observe` (although this last will be tested indirectly through the other fns).

trin-metrics/src/utils.rs
morph-dev: suggestion: `DiscardOnDropHistogramTimer`

KolbyML: Dropable is totally a word. If it isn't how did you say it xD

trin-metrics/src/utils.rs
KolbyML: I think I did it?

KolbyML:
@njgheorghita ready for another look

trin-metrics/src/timer.rs
```diff
@@ -0,0 +1,266 @@
...
+        let metric_family = metric_families.pop().unwrap();
+        let metric = metric_family.get_metric().first().unwrap();
+        assert_eq!(metric.get_label().len(), 1);
+        let proto_histogram = metric.get_histogram();
+        assert_eq!(proto_histogram.get_sample_count(), 0);
+        assert!(proto_histogram.get_sample_sum() >= 0.0);
+    }
+
+    #[test]
+    fn test_discard_through_explict_drop() {
```
njgheorghita, line 201: `explicit` & below in the histogram name. and `implicit` for the same cases in the tests below

njgheorghita (approved):
🚀

PR #1178, opened 2024-02-19
feat: add history storage benchmark

KolbyML:
### To-Do


KolbyML:
Closing this as I don't think this kind of benchmark holds any value. A benchmark which is more wholistic gives us a better view. But the metrics we have I think do a much better job at what this tries to accomplish. Of course the metrics are hard to read for an outsider to a degree, but I would make the argument the same could be said about a storage benchment. A lot of assumptions have to be made it both cases.

PR #1180, opened 2024-02-23
fix: accepting offer with multiple keys

morph-dev:
### What was wrong?

Partially accepting `Offer` request doesn't work.
The `process_accept_utp_payload` function expects to receive keys and values that were accepted, while we send all requested keys (even the ones that are not accepted).

I noticed this while implementing the state network, as this logic is used during Recursive gossip.

### How was it fixed?
Propagated correct keys.

**NOTE: Testing**

- I tried for at least 8 hours to make a test for this but I failed. I'm open for suggestions on how to do it.
- I was mostly trying to make a test inside `overlay_service.rs`
  -  I still have the branch and I can push it to my github if someone wants to see and try to make it work (I think I got close)
- I thought about doing with via `ethportal-peertest`, but none of the public facing apis exposes this functionality.

I verified that it works on the local implementation of the state network with recursive gossip

### To-Do

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -1308,6 +1308,8 @@ where
                 )
             })?;
 
+        let mut accepted_keys: Vec<TContentKey> = Vec::default();
+
         for (i, key) in content_keys.iter().enumerate() {
             // Accept content if within radius and not already present in the data store.
             let accept = self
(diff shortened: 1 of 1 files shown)


ogenev (approved):
Nice find, I think this will solve some of our  "Accepted content already stored" warnings in the logs.

Overlay service definitely need more love and i'm not sure if this change is currently testable.

morph-dev:
> * Offer content `0x123` & `0x456` from Node B -> Node A

How do you offer this? There is no public API (as far as I know) that you can call for this functionality

njgheorghita:
@morph-dev Ahh, yeah I think you're right.

PR #1181, opened 2024-02-23
fix(storage): uses milos find of blobs + proposal 1 for history/beacon

KolbyML:
### What was wrong?
Most of the improvements were found/discovered by milos and discussed here https://github.com/ethereum/trin/issues/1157

image
This was caused because of `XOR_FIND_FARTHEST_QUERY_NETWORK`
- we were calculating distance on every value in the database
image
- We are doing `hex_encode(value)` which is making us store double the values
- `hex_encode`/`hex_decode` also makes inserts/reads slower as well
### How was it fixed?
I implemented 2 new tables `content_data_history` and `content_data_beacon` which now uses blobs instead of text

The biggest improvement though in terms of eff is `XOR_FIND_FARTHEST_QUERY_NETWORK` now we precalculate the distance values

### DIFF
--- a/ethportal-api/src/types/distance.rs
+++ b/ethportal-api/src/types/distance.rs
@@ -41,10 +41,8 @@ impl Distance {
 
     /// Returns the top 4 bytes representation of `self`.
     pub fn big_endian_u32(&self) -> u32 {
-        let mut be: [u8; 32] = [0; 32];
-        self.0.to_big_endian(&mut be);
         let mut be_bytes = [0u8; 4];
-        be_bytes.copy_from_slice(&be[..4]);
+        be_bytes.copy_from_slice(&self.big_endian()[..4]);
         u32::from_be_bytes(be_bytes)
     }
 }
--- a/ethportal-api/src/types/distance.rs
+++ b/ethportal-api/src/types/distance.rs
@@ -38,6 +38,15 @@ impl Distance {
         self.0.to_big_endian(&mut be);
         be
     }
+
+    /// Returns the top 4 bytes representation of `self`.
+    pub fn big_endian_u32(&self) -> u32 {
+        let mut be: [u8; 32] = [0; 32];
+        self.0.to_big_endian(&mut be);
+        let mut be_bytes = [0u8; 4];
+        be_bytes.copy_from_slice(&be[..4]);
+        u32::from_be_bytes(be_bytes)
+    }
 }
 
 impl From<U256> for Distance {
--- a/trin-beacon/src/storage.rs
+++ b/trin-beacon/src/storage.rs
@@ -303,13 +303,15 @@ impl BeaconStorage {
         value: &Vec<u8>,
     ) -> Result<usize, ContentStoreError> {
         let conn = self.sql_connection_pool.get()?;
-        match conn.execute(
+        Ok(conn.execute(
             INSERT_QUERY_BEACON,
-            params![content_id.to_vec(), content_key, value, value.len()],
-        ) {
-            Ok(result) => Ok(result),
-            Err(err) => Err(err.into()),
-        }
+            params![
+                content_id.to_vec(),
+                content_key,
+                value,
+                32 + content_key.len() + value.len()
+            ],
+        )?)
     }
 
     fn db_insert_lc_update(&self, period: &u64, value: &Vec<u8>) -> Result<(), ContentStoreError> {
--- a/trin-beacon/src/storage.rs
+++ b/trin-beacon/src/storage.rs
@@ -11,7 +11,6 @@ use ethportal_api::{
         distance::Distance,
         portal_wire::ProtocolId,
     },
-    utils::bytes::{hex_decode, hex_encode},
     BeaconContentKey, OverlayContentKey,
 };
 use r2d2::Pool;
@@ -25,10 +24,11 @@ use trin_metrics::{portalnet::PORTALNET_METRICS, storage::StorageMetricsReporter
 use trin_storage::{
     error::ContentStoreError,
     sql::{
-        CONTENT_KEY_LOOKUP_QUERY_DB, INSERT_LC_UPDATE_QUERY, LC_UPDATE_LOOKUP_QUERY,
-        LC_UPDATE_PERIOD_LOOKUP_QUERY, LC_UPDATE_TOTAL_SIZE_QUERY, TOTAL_DATA_SIZE_QUERY_DB,
+        CONTENT_KEY_LOOKUP_QUERY_BEACON, CONTENT_VALUE_LOOKUP_QUERY_BEACON, INSERT_LC_UPDATE_QUERY,
+        INSERT_QUERY_BEACON, LC_UPDATE_LOOKUP_QUERY, LC_UPDATE_PERIOD_LOOKUP_QUERY,
+        LC_UPDATE_TOTAL_SIZE_QUERY, TOTAL_DATA_SIZE_QUERY_BEACON,
     },
-    utils::{get_total_size_of_directory_in_bytes, insert_value, lookup_content_value},
+    utils::get_total_size_of_directory_in_bytes,
     ContentStore, DataSize, PortalStorageConfig, ShouldWeStoreContent, BYTES_IN_MB_U64,
 };
 
@@ -111,7 +111,6 @@ pub struct BeaconStorage {
     sql_connection_pool: Pool<SqliteConnectionManager>,
     storage_capacity_in_bytes: u64,
     metrics: StorageMetricsReporter,
-    network: ProtocolId,
     cache: BeaconStorageCache,
 }
 
@@ -273,7 +272,6 @@ impl BeaconStorage {
             sql_connection_pool: config.sql_connection_pool,
             storage_capacity_in_bytes: config.storage_capacity_mb * BYTES_IN_MB_U64,
             metrics,
-            network: ProtocolId::Beacon,
             cache: BeaconStorageCache::new(),
         };
 
@@ -301,11 +299,17 @@ impl BeaconStorage {
     fn db_insert(
         &self,
         content_id: &[u8; 32],
-        content_key: &String,
+        content_key: &Vec<u8>,
         value: &Vec<u8>,
     ) -> Result<usize, ContentStoreError> {
         let conn = self.sql_connection_pool.get()?;
-        insert_value(conn, content_id, content_key, value, u8::from(self.network))
+        match conn.execute(
+            INSERT_QUERY_BEACON,
+            params![content_id.to_vec(), content_key, value, value.len()],
+        ) {
+            Ok(result) => Ok(result),
+            Err(err) => Err(err.into()),
+        }
     }
 
     fn db_insert_lc_update(&self, period: &u64, value: &Vec<u8>) -> Result<(), ContentStoreError> {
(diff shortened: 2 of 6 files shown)


KolbyML:
I have been talking with milos and with the find of https://github.com/ethereum/trin/pull/1182 making database performance 10x better I am comfortable waiting till Milos is done his big database refactor PR.

KolbyML:
This PR is very small plus includes the database transition code to the same database format milos uses https://github.com/ethereum/trin/pull/1183

After lookihg at #1183 I think it will take a good bit longer to get in for the performance improvements in history that I want. That being said #1181 contains 90% of the improvements with a lot less code to review and it is fully compatible with Milos's changes.

morph-dev (commented):
While we are at refactoring storage, I want to raise one question. What is the point of `metrics.report_total_storage_usage_bytes`?
Currently it just reports the size of the folder, so it's not network related at all.

Is that how it should be?
If not, should we remove it? We can use `dbstat` table as a replacement, but I think the query is not very fast.

trin-storage/src/sql.rs
```diff
@@ -1,71 +1,108 @@
...
-                            DROP INDEX IF EXISTS content_id_long_idx;
-                            CREATE INDEX IF NOT EXISTS network_idx ON content_data(network);
-                            CREATE INDEX IF NOT EXISTS content_key_idx ON content_data(content_key);";
-
-pub const INSERT_QUERY_NETWORK: &str =
-    "INSERT OR IGNORE INTO content_data (content_id_long, content_id_short, content_key, content_value, network, content_size)
-                            VALUES (?1, ?2, ?3, ?4, ?5, ?6)";
+pub const INSERT_QUERY_HISTORY: &str =
+    "INSERT OR IGNORE INTO history (content_id, content_key, content_value, distance_short, content_size)
+                            VALUES (?1, ?2, ?3, ?4, ?5)";
```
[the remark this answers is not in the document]
KolbyML: I want to move over to `sqlx` which doesn't support named params to my knowledge. I looked into using ``tokio_spawnblocking`` but it wasn't that clean and looked harder to maintain

trin-storage/src/sql.rs
```diff
@@ -1,71 +1,108 @@
...
-                            CREATE INDEX IF NOT EXISTS content_id_short_idx_3 ON content_data(network, content_id_short);
-                            DROP INDEX IF EXISTS content_id_long_idx;
-                            CREATE INDEX IF NOT EXISTS network_idx ON content_data(network);
-                            CREATE INDEX IF NOT EXISTS content_key_idx ON content_data(content_key);";
-
-pub const INSERT_QUERY_NETWORK: &str =
-    "INSERT OR IGNORE INTO content_data (content_id_long, content_id_short, content_key, content_value, network, content_size)
-                            VALUES (?1, ?2, ?3, ?4, ?5, ?6)";
+pub const INSERT_QUERY_HISTORY: &str =
+    "INSERT OR IGNORE INTO history (content_id, content_key, content_value, distance_short, content_size)
```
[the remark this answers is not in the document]
KolbyML: Yeah good call. I think or ignore is fine for the create table statements, but not insert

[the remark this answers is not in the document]
KolbyML: Good catch I didn't realize or ignore was alrea;dy here before we got here

trin-storage/src/utils.rs
```diff
@@ -1,34 +1,99 @@
...
+use ethportal_api::utils::bytes::hex_decode;
+use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 use rusqlite::params;
 use std::{fs, path::Path};
 use tracing::info;
 
+// todo: remove this once non_versioned to versioned migration is over
+struct ContentData {
+    content_id_long: String,
```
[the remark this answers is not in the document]
KolbyML: Yeah that is how we read and decode all the old `TEXT` values I will double check right now though just to be safe

trin-storage/src/utils.rs
```diff
@@ -1,34 +1,99 @@
...
+                        Err(err) => unreachable!("{}", err),
+                    };
+                },
+                _ => unreachable!("only history or beacon should be stored here")
+            }
+        }
+
+        conn.execute_batch(DROP_CONTENT_DATA_QUERY_DB)?;
+    }
+    update_store_info(&ContentType::History, StoreVersion::IdIndexedV1, &conn)?;
```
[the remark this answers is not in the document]
KolbyML: Ok

KolbyML:
After talking with Milos a bit and how nick also thinks it is fine to reset I removed the migration code as then we are less prone to possible errors on our next deploy

KolbyML:
> While we are at refactoring storage, I want to raise one question. What is the point of metrics.report_total_storage_usage_bytes?

I think this is just a metric of overall storage usage as each network should be isolated in terms of storage so it is nice to be able to get 1 distinct number

trin-storage/src/sql.rs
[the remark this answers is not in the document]
KolbyML: not related to this PR, but I just found this https://github.com/bikeshedder/deadpool/tree/master/r2d2 if anyone is looking. It looks like it would make it pretty drop in which is good

ethportal-api/src/types/distance.rs
```diff
@@ -38,6 +38,15 @@ impl Distance {
         self.0.to_big_endian(&mut be);
         be
     }
+
+    /// Returns the top 4 bytes representation of `self`.
+    pub fn big_endian_u32(&self) -> u32 {
```
morph-dev, line 43: I think this can be simplified:


    pub fn big_endian_u32(&self) -> u32 {
        let be_bytes: [u8; 4] = self.big_endian()[..4]
            .try_into()
            .expect("to get top 4 bytes from distance");
        u32::from_be_bytes(be_bytes)
    }

trin-beacon/src/storage.rs
```diff
@@ -515,7 +516,20 @@ impl BeaconStorage {
...
+        let id = id.to_vec();
+        let result: Result<Vec<Vec<u8>>, ContentStoreError> = query
+            .query_map([id], |row| {
+                let row: Vec<u8> = row.get(0)?;
+                Ok(row)
+            })?
+            .map(|row| row.map_err(ContentStoreError::Rusqlite))
+            .collect();
+
+        match result?.first() {
```
morph-dev, line 529: Wouldn't simple `result?.first()` work?

If not, I'm sure something like this should:
```rust
result?.first().map(|val| val.to_vec())
```

trin-history/src/storage.rs
```diff
@@ -218,14 +205,13 @@ impl HistoryStorage {
...
-        // store content key w/o the 0x prefix
-        let content_key = hex_encode(content_key).trim_start_matches("0x").to_string();
         let value_size = value.len() as u64;
         match self.db_insert(&content_id, &content_key, value) {
             Ok(result) => {
                 // Insertion successful, increase total network storage count
                 if result == 1 {
-                    self.storage_occupied_in_bytes += value_size;
+                    // adding 32 bytes for content_id and 32 for content_key
+                    self.storage_occupied_in_bytes += value_size + 64;
```
morph-dev, line 214: Maybe create a function in `trin-storage/src/utils.rs` for calculating the content_size. That way it can be documented once and we don't have to add constants all over the code.

It might be overkill from function functionality, but I think it improves readability.

trin-beacon/src/storage.rs
```diff
@@ -301,11 +299,19 @@ impl BeaconStorage {
...
-        content_key: &String,
+        content_key: &Vec<u8>,
         value: &Vec<u8>,
     ) -> Result<usize, ContentStoreError> {
         let conn = self.sql_connection_pool.get()?;
-        insert_value(conn, content_id, content_key, value, u8::from(self.network))
+        Ok(conn.execute(
+            INSERT_QUERY_BEACON,
+            params![
+                content_id.to_vec(),
```
morph-dev, line 309: The `content_id.to_vec()` will make a copy and I don't think it's necessary.

If you can't pass `content_id` directly, you should be able to pass: `content_id.as_slice()`.

Same in many other places as well.

trin-history/src/storage.rs
```diff
@@ -354,7 +336,20 @@ impl HistoryStorage {
...
+        let id = id.to_vec();
+        let result: Result<Vec<Vec<u8>>, ContentStoreError> = query
+            .query_map([id], |row| {
+                let row: Vec<u8> = row.get(0)?;
+                Ok(row)
+            })?
+            .map(|row| row.map_err(ContentStoreError::Rusqlite))
+            .collect();
+
+        let result: Result<Option<Vec<u8>>, _> = match result?.first() {
```
morph-dev, line 349: Can't this be:
```rust
let result = result?.first();
self.metrics.stop_process_timer(timer);
Ok(result)
```

trin-storage/src/utils.rs
```diff
@@ -26,9 +20,11 @@ pub fn setup_sql(node_data_dir: &Path) -> Result<Pool<SqliteConnectionManager>,
     let manager = SqliteConnectionManager::file(sql_path);
     let pool = Pool::new(manager)?;
     let conn = pool.get()?;
-    conn.execute_batch(CREATE_QUERY_DB)?;
+    conn.execute_batch(CREATE_QUERY_DB_HISTORY)?;
```
morph-dev, line 23: I think it's safe to move `CREATE_QUERY_DB_HISTORY` to `HistoryStorage` and run it only when it's created.
Similarly, maybe we can move Beacon and LcUpdate to `BeaconStorage` as well.

I'm not sure about `STORE_INFO_CREATE_TABLE`, as that migh have to run multiple times and there isn't a constructor than we can easily use. Maybe there should be...

morph-dev (commented):
Mostly it looks ok. I'm waiting to approve after today's discussion

morph-dev (approved):

trin-history/src/storage.rs
KolbyML: We are most likely going to replace this with your code in a week or 2 so I think that util function would be overkill as we will be removing it soon enough. I will make a constant for it for the time being though

trin-beacon/src/storage.rs
KolbyML: Good find

trin-beacon/src/storage.rs
```diff
@@ -332,7 +338,6 @@ impl BeaconStorage {
         match content_key.first() {
             Some(&LIGHT_CLIENT_BOOTSTRAP_KEY_PREFIX) => {
                 // store content key w/o the 0x prefix
```
njgheorghita, line 340: Seems like we can remove this comment now

trin-history/src/storage.rs
```diff
@@ -380,22 +373,31 @@ impl HistoryStorage {
...
         let conn = self.sql_connection_pool.get()?;
-        let result = insert_value(conn, content_id, content_key, value, u8::from(self.network));
+        let result = conn.execute(
+            INSERT_QUERY_HISTORY,
+            params![
+                content_id.as_slice(),
+                content_key,
+                value,
+                self.distance_to_content_id(content_id).big_endian_u32(),
+                32 + content_key.len() + value.len()
```
njgheorghita, line 388: Why not do `CONTENT_ID_AND_KEY_LENGTH + value.len()` here?

trin-history/src/storage.rs
```diff
@@ -616,7 +615,7 @@ pub mod test {
         }
 
         let bytes = storage.get_total_storage_usage_in_bytes_from_network()?;
-        assert_eq!(1600000, bytes); // 32kb * 50
+        assert_eq!(1603200, bytes); // 32kb * 50 + 64 * 50
```
njgheorghita, line 618: `32kb * 50 + CONTENT_ID_AND_KEY_LENGTH * 50` or `(32kb + CONTENT_ID_AND_KEY_LENGTH) * 50` ... and below on `787`

njgheorghita (approved):

PR #1182, opened 2024-02-24
refactor(perf): fix indexes with 10x increase in performance for find_farthest_content_id

KolbyML:
### What was wrong?
image

### How was it fixed?
image
image
image

Non optimal sql indexes which were made in the last PR.

With this fix we are getting 10x faster find_farthest_content_id and up to 20-40x more database calls per minute

### DIFF
--- a/trin-storage/src/sql.rs
+++ b/trin-storage/src/sql.rs
@@ -14,7 +14,7 @@ pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                             CREATE INDEX IF NOT EXISTS content_size_idx_2 ON content_data(network, content_size);
                             DROP INDEX IF EXISTS content_id_short_idx;
                             DROP INDEX IF EXISTS content_id_short_idx_2;
-                            CREATE INDEX IF NOT EXISTS content_id_short_idxf_2 ON content_data(network, content_id_short);
+                            CREATE INDEX IF NOT EXISTS content_id_short_idx_3 ON content_data(network, content_id_short);
                             DROP INDEX IF EXISTS content_id_long_idx;
                             CREATE INDEX IF NOT EXISTS network_idx ON content_data(network);
                             CREATE INDEX IF NOT EXISTS content_key_idx ON content_data(content_key);";
--- a/trin-storage/src/sql.rs
+++ b/trin-storage/src/sql.rs
@@ -13,7 +13,8 @@ pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                             DROP INDEX IF EXISTS content_size_idx;
                             CREATE INDEX IF NOT EXISTS content_size_idx_2 ON content_data(network, content_size);
                             DROP INDEX IF EXISTS content_id_short_idx;
-                            CREATE INDEX IF NOT EXISTS content_id_short_idx_2 ON content_data(network, content_id_short, content_id_long);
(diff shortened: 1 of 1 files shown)


trin-storage/src/sql.rs
```diff
@@ -13,7 +13,8 @@ pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                             DROP INDEX IF EXISTS content_size_idx;
                             CREATE INDEX IF NOT EXISTS content_size_idx_2 ON content_data(network, content_size);
                             DROP INDEX IF EXISTS content_id_short_idx;
-                            CREATE INDEX IF NOT EXISTS content_id_short_idx_2 ON content_data(network, content_id_short, content_id_long);
+                            DROP INDEX IF EXISTS content_id_short_idx_2;
+                            CREATE INDEX IF NOT EXISTS content_id_short_idxf_2 ON content_data(network, content_id_short);
```
morph-dev, line 17:
```suggestion
                            CREATE INDEX IF NOT EXISTS content_id_short_idx_3 ON content_data(network, content_id_short);
```

morph-dev (approved):

trin-storage/src/sql.rs
```diff
@@ -33,7 +34,7 @@ pub const XOR_FIND_FARTHEST_QUERY_NETWORK: &str = "SELECT
                                     content_id_long
                                     FROM content_data
                                     WHERE network = (?2)
-                                    ORDER BY ((?1 | content_id_short) - (?1 & content_id_short)) DESC";
+                                    ORDER BY ((?1 | content_id_short) - (?1 & content_id_short)) DESC LIMIT 1";
```
njgheorghita, line 37: Just to confirm my understanding, but this change doesn't really do much to help with performance, correct? The query takes just as long as it did before, the only difference is that we're returning the first list entry from the query rather than every list entry?

trin-storage/src/sql.rs
```diff
@@ -13,7 +13,8 @@ pub const CREATE_QUERY_DB: &str = "CREATE TABLE IF NOT EXISTS content_data (
                             DROP INDEX IF EXISTS content_size_idx;
                             CREATE INDEX IF NOT EXISTS content_size_idx_2 ON content_data(network, content_size);
                             DROP INDEX IF EXISTS content_id_short_idx;
-                            CREATE INDEX IF NOT EXISTS content_id_short_idx_2 ON content_data(network, content_id_short, content_id_long);
+                            DROP INDEX IF EXISTS content_id_short_idx_2;
+                            CREATE INDEX IF NOT EXISTS content_id_short_idx_3 ON content_data(network, content_id_short);
```
njgheorghita, line 17: Can you explain why this change improved performance?

trin-storage/src/sql.rs
KolbyML: Pretty much this should be negligible in most cases but it is best practice to if that is what your returns

trin-storage/src/sql.rs
KolbyML: image
It improved performance because the index is indexing the 2 fields this query has to manipulate to get the result.

containing all 3 decreases performance because content_id_long isn't need during the commutation it is just returned at the end. So it is just taking up time as the query goes through.

Lets say you have a query X which is selecting A ordering B and only performing the option on C. In this query we aren't performing any commutations on A. But because it is there we will still check at every step this check takes time. this will also make inserts longer cause the index takes more work has more variables to commute

PR #1184, opened 2024-02-27
feat(bridge): add single random epoch mode to fourfours bridge

njgheorghita:
### What was wrong?
We know we have some issues with our bridge when it's run as a long-running service. When our "latest" audit was switched to Infura, it still witnessed some stability issues. However, these were always resolved by rebooting the bridge, since the Infura provider is reliable. But debugging long-running processes is difficult. 

This solution is a bit of a hack, but effectively it turns our "backfill" bridges into short-term (aka the time it takes to gossip a single epoch) processes rather than long-running processes. With this new mode, after gossiping a **single** randomly selected epoch, the bridge process will naturally terminate. In which case the `docker_container` process used in our Ansible deployment scripts, will automatically restart the container, due to the `restart_policy: always` setting. Then the bridge will select another random epoch, gossip it, terminate, etc. 

### How was it fixed?
- Added a mode to the four fours bridge to gossip a single, random epoch and then terminate.

### To-Do

### DIFF
--- a/portal-bridge/README.md
+++ b/portal-bridge/README.md
@@ -33,6 +33,7 @@ Current options include `"trin"` / `"fluffy"`.
 - `"--mode single:b100"`: gossip a single block #100
 - `"--mode single:e100"`: gossip a single epoch #100
 - `"--mode fourfours`: will randomly select era1 files from `era1.ethportal.net` and gossip them
+- `"--mode fourfours:random_epoch"`: will randomly select a single era1 file from `era1.ethportal.net` and then gossip it
 - `"--mode fourfours:e600`: will select era1 file 600 from `era1.ethportal.net` and gossip it
 - `"--mode fourfours:r100-200`: will gossip a block range from an era1 file, range must be from the same epoch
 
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
We need to update the era1 bridge to parse the index.html instead of the xml as the xml only contains the first 1000

morph-dev:
What is the cost of restarting the container? Would it make sense to gossip more (e.g. 10, or configurable with a flag) instead of just one?

KolbyML:
The cost is negligible so basically nothing

KolbyML:
What metric are we using for cost in this context though?

morph-dev:
I guess time it takes to shutdown and restart. But I don't have enough context, so maybe it's stupid.
It is just an a idea to consider

KolbyML:
I mean I think we need to add metrics to the bridge if we want to debug it for me that is currently super low priority because Trin still has a ton of bugs. If the foundation of the pyramid doesn't work aka trin then of course you would see failures in the bridge. 

That reminds me we should probably get my 2 uTP changes in now.

njgheorghita:
Yeah, I agree with @KolbyML that the restart time really is negligible here, and not worth introducing another cli flag for the time being

njgheorghita:
> We need to update the era1 bridge to parse the index.html instead of the xml as the xml only contains the first 1000

@KolbyML I'll take care of that in a follow up pr

PR #1188, opened 2024-02-29
chore: bump uTP version to v0.1.0-alpha.10

KolbyML:
### What was wrong?
old utp version
### How was it fixed?
bumping it up :sunglasses:


njgheorghita (approved):
:ship:

PR #1189, opened 2024-02-29
refactor: wait for semaphore permit only if accepting content

njgheorghita:
### To-Do


KolbyML (approved):
:shipit:

PR #1190, opened 2024-03-01
feat: make our storage async from blocking

KolbyML:
### What was wrong?
Are application is written in Rust using the Tokio runtime. A green threads runtime like you can find in languages like golang.

Our 2 main sources of IO are
- network io
- storage io

But we were doing all our storage io calls non-async.
https://dtantsur.github.io/rust-openstack/tokio/task/fn.spawn_blocking.html
> In general, issuing a blocking call or performing a lot of compute in a future without yielding is not okay, as it may prevent the executor from driving other futures forward

So all of our storage calls to sqlite are issuing blocking calls which by Tokio's own documentation is `not okay`.

You can see the affects of us using a blocking library by looking for io wait on the graphs from what I see it is 5-30%. With this PR iowait will/should become 0%

### How was it fixed?
At first I posted about the find we don't handle this to the discord and Ogenev let me know about `tokio::spawn_blocking`
#### investigation 1: `tokio::spawn_blocking`
The issue with this is because of the spread nature of SQL queries it would be very hard to maintain code using it as `tokio::spawn_blocking` need to go in a lot of places.

The other issue with this is `tokio::spawn_blocking` wouldn't stop our sql connection thread from blocking if it is hitting the limit which is a real concern once we get more networks online as they will all share the same pool of connections.

From the maintainability and amount of work to properly handle the issue this way this route was a no go

#### investigation 2: deadpool-r2d2/deadpool-sqlite a async wrapper for r2d2
I implemented a small proof of concept but once I got to more complex queries this solution had a lot of the same issues as `tokio::spawn_blocking`. Because deadpool is basically a pool which then wraps blocking commands with calls like `tokio::spawn_blocking` its usability wasn't that good and required moving variables to deadpools calls which was a big no no.

#### note on another option: async-sqlite
async-sqlite is basically the same as deadpool but with basically 1 dev behind it. It has all the same problems I had with deadpool

#### investigation 3: sqlx the option I choose
Well sqlx is a little different then what we were used it provides a lot:
- native async interface to the deveopler making it a lot easier to maintain
  -  much better code readability over the other options
- has the biggest community out of the options I could find so the project is very active.

#### final notes
Sqlx is the best option I could find and I believe it is the right path for the future of our codebase

The change is in the title I migrated us from a blocking sqlite library to an async one which will reduce the amount of wasted cpu cycles we are currently getting on storage calls

- [x] passed all portal hive tests

From my testing on machines with faster storage speeds and hence less less iowait times performance is the same.

Noticable improvements in performance are found on machines which have slower storage like hard drives or sd cards end up with much higher iowait times where a performance increases are much more noticable

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -675,6 +675,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "atoi"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f28d99ec8bfea296261ca1af174f24225171fea9664ba9003cbebee704810528"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
 name = "atomic"
 version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -688,6 +697,16 @@ name = "atomic-waker"
 version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
+name = "atomic-write-file"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "edcdbedc2236483ab103a53415653d6b4442ea6141baf1ffa85df29635e88436"
+dependencies = [
+ "nix",
+ "rand 0.8.5",
+]
 
 [[package]]
 name = "attohttpc"
@@ -1760,6 +1779,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fffa369a668c8af7dbf8b5e56c9f744fbd399949ed171606040001947de40b1c"
 dependencies = [
  "const-oid",
+ "pem-rfc7468",
  "zeroize",
 ]
 
@@ -1905,7 +1925,7 @@ dependencies = [
  "enr 0.10.0",
  "fnv",
  "futures 0.3.30",
- "hashlink 0.8.4",
+ "hashlink",
  "hex",
  "hkdf 0.12.4",
  "lazy_static",
@@ -1922,6 +1942,12 @@ dependencies = [
  "uint",
  "zeroize",
 ]
+
+[[package]]
+name = "dotenvy"
+version = "0.15.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1aaf95b3e5c8f23aa320147307562d361db0ae0d51242340f558153b4eb2439b"
 
 [[package]]
 name = "dtoa"
@@ -1994,6 +2020,9 @@ name = "either"
 version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a26ae43d7bcc3b814de94796a5e736d4029efb0ee900c12e2d54c993ad1a1e07"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "elliptic-curve"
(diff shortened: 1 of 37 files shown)


morph-dev (commented):
Personally, I would prefer if this was 2 PR-s, one for the sqlite library and the other for the transition to the `tokio::sync::RwLock` in OverlayService. It would have been much easier to review changes.

With that being said, I did quick overview of the entire PR. I did find one thing that worries me in the overlay service, but I would leave that par to be reviewed by someone who has more knowledge of the overlay service (perhaps Nick).

I focused mostly on the storage part but mostly had high-level view and it mostly looks ok. One thing that I would improve, but I'm not sure how much it makes sense:

We use `query(..).bind(..).map(..)` syntax a lot, and I don't really like it, especially when there are a lot of `.bind(..)` together, like during insert and when `.map(..)` has trivial logic.

I think we should use [`query_with`](https://docs.rs/sqlx/latest/sqlx/fn.query_with.html), [`query_scalar`](https://docs.rs/sqlx/latest/sqlx/fn.query_scalar.html), [`query_as`](https://docs.rs/sqlx/latest/sqlx/fn.query_as.html) and their combinations wherever possible. There are even [macros](https://docs.rs/sqlx/latest/sqlx/index.html#macros) that can help with this.

KolbyML:
We can't separate this pr. Packing_lot doesn't implement async traits and hence was erroring out on me

njgheorghita:
Based on your explanation during sync this morning, iiuc this is more of an enhancement for clients using standard hardware. Our network nodes would not experience a significant increase in performance from this change. Therefore, I'd categorize this as more of an optimization than a bugfix and would recommend closing it for now and revisiting it once we're more focused on optimization rather than core bug squashing

KolbyML:
I don't think we should not do a overlay service refactor then as it doesn't accomplish the goals stated and should be blocked for the same reasons.

I am not against delaying this PR I said I am fine with it in the call. But if we are stating   is our focus then I don't see the argument for an overlay service refactor as it doesn't fix anything.

KolbyML:
> We can't separate this pr. Packing_lot doesn't implement async traits and hence was erroring out on me

Milos dmed me when I posted this and let me know I could do the RwLock PR first I totally didn't think about that so I will do that.

KolbyML:
@njgheorghita also this does help our deployment the ssd's on our ams nodes are bad.

morph-dev:
> I think we should not do a overlay service refactor then as it doesn't accomplishes "bug fixes" which is the stated goal.

Just to add to this discussion.

I would say that maybe refactoring overlay service doesn't fixes bugs, but I would say it's very important:
- the overlay service is hard to maintain, test and upgrade, and the longer we wait the harder it will be
- in the long run, it will help with catching and fixing bugs
- refactoring like this can create bugs as well, but that's another reason to do it sooner

While it might not be as important as bug fixing, I would say it's probably more important than performance optimization. Of course, depending on the magnitude of performance improvements.

njgheorghita:
> also this does help our deployment the ssd's on our ams nodes are bad.

I was under the impression that we're using the same kind of drives amongst all of our network nodes? (eg ams, sgp, nyc)

> While it might not be as important as bug fixing, I would say it's probably more important than performance optimization. Of course, depending on the magnitude of performance improvements.

Yeah, I pretty much agree with this. My understanding of this pr is that it falls in the "optimization" category. I don't consider the overlay service work an "optimization" but a "refactor". And not necessarily a purely aesthetic refactor, but a refactor that's aimed towards improving functionality / ease of implementing new features (eg inbound utp tx content key queues).

But, these are all loosely defined categories, and there's shades of grey between them all. Ultimately, you're free to work on whatever you think is the most important thing. This is how the team has operated in the past. If you think this feature is the most important thing you can work on to improve the performance of the network then go for it. I wasn't trying to tell you not to work on this, I was just giving my opinion that I don't think this is the most important thing you could be working on. It's up to you to interpret that feedback (and Milos' feedback) and make your own decision.

KolbyML:
Closing this as there is no need for it for the foresee able future. I learnt a lot from it which will be valuable down the line which is pretty cool

PR #1192, opened 2024-03-01
fix: remove duplicate ids when propagating

morph-dev:
### What was wrong?

It's possible to propagate the same content id multiple times within the same offer.

The #1176 added functionality to propagate additional content key/value, which made this more likely to occur (originally it was possible only if offer contained duplicates).

### How was it fixed?

Now we are keeping only unique content ids.

NOTE: Similar to other functionality that involves offer with multiple entries, I don't know how to test this.
I tested it using state network and recursive gossip on my local machine (that's how I noticed issue to begin with).
Once, #1187 is in, I will create peer-to-peer tests that will test this and other similar functionality.

### To-Do

### DIFF
--- a/portalnet/Cargo.toml
+++ b/portalnet/Cargo.toml
@@ -26,6 +26,7 @@ fnv = "1.0.7"
 futures = "0.3.21"
 hex = "0.4.3"
 igd-next = "0.14.2"
+itertools = "0.12.1"
 lazy_static = "1.4.0"
 leb128 = "0.2.1"
 local-ip-address = "0.5.6"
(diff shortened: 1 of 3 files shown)


njgheorghita (approved):

portalnet/Cargo.toml
```diff
@@ -50,6 +50,7 @@ validator = { version = "0.13.0", features = ["derive"] }
 url = "2.3.1"
 utp-rs = { git = "https://github.com/ethereum/utp", tag = "v0.1.0-alpha.10" }
 tokio-stream = { version = "0.1.14", features = ["sync"] }
+itertools = "0.12.1"
```
njgheorghita, line 53: can you move this so the list is sorted alphabetically

morph-dev: I used `$cargo add ...` which keeps dependencies sorted, so I wasn't paying attention.
The thing is that it keeps dependencies sorted only if they are already sorted.
Fixed everything now.

PR #1194, opened 2024-03-02
refactor: remove async_trait from validators

morph-dev:
### What was wrong?

We were using `async_trait` in our validators, which I was told should no longer be needed.

### How was it fixed?

Removed the `async_trait` from validators.

I didn't test this in any way, I'm counting on our tests catching if there is any issue.

### To-Do


ogenev (approved):
LGTM 👍

PR #1195, opened 2024-03-02
feat: create usage_stats table and triggers to update it

morph-dev:
### What was wrong?

As described in #1157, we could use sql triggers to track count and size of content that is stored, instead of doing:
- it manually (what history does) 
- making expensive queries (what history used to do and what `id_indexed_v1` does now)

### How was it fixed?

Created `usage_stats` table and provided utility functions for interaction.
We already had `struct UsageStats` as part of the `id_indexed_v1` store, that I repurposed. As a consequence I had to refactor some of it's logic (mostly move it inside `IdIndexedV1StoreConfig`).

I created dedicated tests for it, but it's indirectly tested by tests in `id_indexed_v1`.

### To-Do

### DIFF
--- a/trin-storage/src/versioned/id_indexed_v1/store.rs
+++ b/trin-storage/src/versioned/id_indexed_v1/store.rs
@@ -81,9 +81,11 @@ impl IdIndexedV1Store {
         self.metrics
             .report_storage_capacity_bytes(self.config.storage_capacity_bytes as f64);
 
-        // query real usage stats and update "usage_stats" table to match.
         let conn = self.config.sql_connection_pool.get()?;
-        let usage_stats = conn.query_row(
+        let mut usage_stats = get_usage_stats(&conn, &self.config.content_type)?;
+
+        // query real usage stats and update "usage_stats" table to match.
+        let real_usage_stats = conn.query_row(
             &sql::entry_count_and_size(&self.config.content_type),
             [],
             |row| {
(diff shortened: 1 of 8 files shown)


ogenev (approved):
Haven't gone into too much details but looks solid to me and happy to move it forward ⛵.

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -76,26 +81,41 @@ impl IdIndexedV1Store {
...
+            |row| {
+                let used_capacity: f64 = row.get("used_capacity")?;
+                Ok(UsageStats {
+                    entry_count: row.get("count")?,
+                    total_entry_size_bytes: used_capacity.round() as u64,
+                })
+            },
+        )?;
+        usage_stats.report_metrics(&self.metrics);
+        update_usage_stats(&conn, &self.config.content_type, &usage_stats)?;
```
KolbyML, line 98: If `usage_stats` is being tracked by triggers why do we recalculate the values for the tables on each reboot? I am not saying this is bad, but wouldn't this not be needed unless there was a bug?

Souldn't we just be able to get the stats from the table on boot?

This isn't really a performance concern as this is only on boot so not a blocker just an interesting thing to think about

KolbyML (approved):
:rocket:

trin-storage/src/versioned/id_indexed_v1/store.rs
morph-dev: It's more of a safety check, making sure it has correct value. I added `error!` if they are different.

But you are right, it shouldn't be needed in most cases.

But there might be some cases when it might be needed, e.g. after transitioning from one version to another (of course depending on how transition is done, it might not be needed in this case either).

KolbyML: That is cool you added an error because now we will be aware if a bug was introduced :rocket:

PR #1200, opened 2024-03-05
fix: updates ContentNotFound error code to match spec

mrferris:
### What was wrong?

[Spec proposal](https://github.com/ethereum/portal-network-specs/pull/276) proposes different error code than trin currently uses

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/find.rs
+++ b/ethportal-peertest/src/scenarios/find.rs
@@ -146,7 +146,7 @@ pub async fn test_trace_recursive_find_content_for_absent_content(peertest: &Pee
     assert!(error.contains("Content not found"));
     // test that trace is present
     assert!(error.contains("respondedWith"));
-    assert!(error.contains("-32001"));
+    assert!(error.contains("-39001"));
 }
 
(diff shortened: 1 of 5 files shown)


ethportal-peertest/src/scenarios/find.rs
```diff
@@ -146,7 +146,7 @@ pub async fn test_trace_recursive_find_content_for_absent_content(peertest: &Pee
     assert!(error.contains("Content not found"));
```
njgheorghita, line 146: I think this error message is also out of date, and should be "Unable to locate content on the network." [source](https://github.com/ethereum/portal-network-specs/pull/276/files#diff-121f57a5231273b26c33d1d5ef2881a0b6f15f570c9f5ff181d02cdd260bf71dR4)

mrferris: Ok, but note that both don't seem consistent with the way trin uses them, because trin uses this error when utp fails in addition to when the query doesn't find the data

njgheorghita: Yeah, at some point we'll probably have to update the spec to specify between content not found and utp failures

njgheorghita (approved):

PR #1201, opened 2024-03-05
Revert "refactor: wait for semaphore permit only if accepting content"

KolbyML:
This reverts commit 3094fbb2a9b937eb1e5e57e543910610da080570.

### What was wrong?
Tons of uTP requests are "failing" reporting trying to connect to connection id 0 

This bug was introduced in https://github.com/ethereum/trin/pull/1189

The bug basically was there wouldn't be a semaphore avaliable but the bitlist would already be filled with data
### How was it fixed?
by reverting https://github.com/ethereum/trin/pull/1189

As stated https://github.com/ethereum/trin/pull/1189#issuecomment-1971881083 `try_acquire_owned` is none blocking it instantly fails if a semaphore isn't avaliable which is the intended behavior.

We should check if a semaphore is avaliable first because the call `try_acquire_owned` is instant and free, we don't have to wait because if there are no semaphores avaliable it instantly returns with `Some(TryAcquireError::NoPermits)`. Where as filling the bitlist requires storage calls which are much more expensive. So if we can't obtain a semaphore we shouldn't waste cpu reading from disk

I would like to give a big thanks to @mrferris for practically telling me the problem.

### DIFF
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -1283,6 +1283,15 @@ where
             })?;
 
         // Attempt to get semaphore permit if fails we return an empty accept
+        // `try_acquire_owned()` isn't blocking and will instantly return with
+        // `Some(TryAcquireError::NoPermits)` error if there isn't a permit avaliable
+        // The reason we get the permit before checking if we can store it is because
+        // * checking if a semaphore is avaliable is basically free it doesn't block and will return
+        //   instantly
+        // * filling the `requested_keys` is expensive because it requires calls to disk which
+        //   should be avoided.
+        // so by trying to acquire the semaphore before the storage call we avoid unnecessary work
+        // **Note:** if we are not accepting any content `requested_keys` should be empty
         let permit = match self
             .utp_controller
             .inbound_utp_transfer_semaphore
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -1282,6 +1282,22 @@ where
                 )
             })?;
 
+        // Attempt to get semaphore permit if fails we return an empty accept
+        let permit = match self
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay_service.rs
```diff
@@ -1282,6 +1282,22 @@ where
...
+            .utp_controller
+            .inbound_utp_transfer_semaphore
+            .clone()
+            .try_acquire_owned()
+        {
+            Ok(permit) => permit,
+            Err(_) => {
+                return Ok(Accept {
+                    connection_id: 0,
+                    content_keys: requested_keys,
```
mrferris, line 1296:
```suggestion
                    content_keys: vec![],
```

Without this change won't the remote peer still try to connect to connection id 0?

mrferris: > it is empty already.

I see, thanks 👍

portalnet/src/overlay_service.rs
```diff
@@ -1282,6 +1282,22 @@ where
...
+        let permit = match self
+            .utp_controller
+            .inbound_utp_transfer_semaphore
+            .clone()
+            .try_acquire_owned()
+        {
+            Ok(permit) => permit,
+            Err(_) => {
+                return Ok(Accept {
+                    connection_id: 0,
```
njgheorghita, line 1295: Well I guess my understanding was something along the lines that a connection id of `0` was an "empty accept" (according to the comment above). My understanding was that a peer wouldn't try to connect with a connection id of `0`. But, I guess that was incorrect. 

Good find, I just think it's definitely worth adding some comments here explaining why this should happen before we populate the bitlist & that the semaphore check is effectively free

njgheorghita (approved):

PR #1202, opened 2024-03-05
refactor(portalnet): extract basic utp stream handling into utp controller

njgheorghita:
### What was wrong?
Initial refactor of `OverlayService`. Fairly straightforward extracting of all utp stream handling logic into the `UtpController`.

Coming up in a following pr will be the "offer" queue to eliminate concurrent utp streams for the same content key.

### How was it fixed?
- extracted logic from `OverlayService` -> `UtpController`
- tested locally via kurtosis & resolved a bug in metrics reporting

### To-Do

### DIFF
--- a/portalnet/src/overlay.rs
+++ b/portalnet/src/overlay.rs
@@ -541,8 +541,13 @@ where
         conn_id: u16,
     ) -> Result<Vec<u8>, OverlayRequestError> {
         self.utp_controller
-            .connect_inbound_stream(conn_id, enr, /* query_trace */ None)
+            .connect_inbound_stream(conn_id, enr)
             .await
+            .map_err(|err| OverlayRequestError::ContentNotFound {
+                message: format!("Unable to locate content on the network: {err:?}"),
+                utp: true,
+                trace: None,
+            })
     }
 
     /// Offer is sent in order to store content to k nodes with radii that contain content-id
--- a/portalnet/src/overlay.rs
+++ b/portalnet/src/overlay.rs
@@ -540,8 +540,13 @@ where
         enr: Enr,
         conn_id: u16,
     ) -> Result<Vec<u8>, OverlayRequestError> {
+        let cid = utp_rs::cid::ConnectionId {
+            recv: conn_id,
+            send: conn_id.wrapping_add(1),
+            peer: UtpEnr(enr),
+        };
         self.utp_controller
-            .connect_inbound_stream(conn_id, enr)
+            .connect_inbound_stream(cid)
             .await
             .map_err(|err| OverlayRequestError::ContentNotFound {
                 message: format!("Unable to locate content on the network: {err:?}"),
--- a/portalnet/src/overlay.rs
+++ b/portalnet/src/overlay.rs
@@ -23,14 +23,14 @@ use parking_lot::RwLock;
 use ssz::Encode;
 use tokio::sync::{broadcast, mpsc::UnboundedSender};
 use tracing::{debug, error, info, warn};
+use utp_rs::socket::UtpSocket;
 
 use crate::{
-    discovery::Discovery,
+    discovery::{Discovery, UtpEnr},
     find::query_info::{FindContentResult, RecursiveFindContentResult},
     gossip::{propagate_gossip_cross_thread, trace_propagate_gossip_cross_thread, GossipResult},
     overlay_service::{
         OverlayCommand, OverlayRequest, OverlayRequestError, OverlayService, RequestDirection,
-        UTP_CONN_CFG,
     },
     types::node::Node,
     utp_controller::UtpController,
@@ -38,6 +38,7 @@ use crate::{
 use ethportal_api::{
     types::{
         bootnodes::Bootnode,
+        cli::DEFAULT_UTP_TRANSFER_LIMIT,
         discv5::RoutingTableInfo,
         distance::{Distance, Metric},
         enr::Enr,
@@ -70,6 +71,7 @@ pub struct OverlayConfig {
     pub query_num_results: usize,
     pub findnodes_query_distances_per_peer: usize,
     pub disable_poke: bool,
+    pub utp_transfer_limit: usize,
 }
 
 impl Default for OverlayConfig {
@@ -87,6 +89,7 @@ impl Default for OverlayConfig {
             query_num_results: MAX_NODES_PER_BUCKET,
             findnodes_query_distances_per_peer: 3,
             disable_poke: false,
+            utp_transfer_limit: DEFAULT_UTP_TRANSFER_LIMIT,
         }
     }
 }
@@ -135,7 +138,7 @@ where
     pub async fn new(
         config: OverlayConfig,
         discovery: Arc<Discovery>,
-        utp_controller: Arc<UtpController>,
+        utp_socket: Arc<UtpSocket<UtpEnr>>,
         store: Arc<RwLock<TStore>>,
         protocol: ProtocolId,
         validator: Arc<TValidator>,
@@ -152,6 +155,11 @@ where
             overlay_metrics: PORTALNET_METRICS.overlay(),
             protocol: protocol.to_string(),
         };
+        let utp_controller = Arc::new(UtpController::new(
+            config.utp_transfer_limit,
+            utp_socket,
+            metrics.clone(),
+        ));
         let command_tx = OverlayService::<TContentKey, TMetric, TValidator, TStore>::spawn(
             Arc::clone(&discovery),
             Arc::clone(&store),
@@ -532,23 +540,9 @@ where
         enr: Enr,
         conn_id: u16,
     ) -> Result<Vec<u8>, OverlayRequestError> {
-        let cid = utp_rs::cid::ConnectionId {
-            recv: conn_id,
-            send: conn_id.wrapping_add(1),
-            peer: crate::discovery::UtpEnr(enr),
-        };
-        let mut stream = self
-            .utp_controller
-            .connect_with_cid(cid, *UTP_CONN_CFG)
+        self.utp_controller
+            .connect_inbound_stream(conn_id, enr, /* query_trace */ None)
             .await
-            .map_err(|err| OverlayRequestError::UtpError(format!("{err:?}")))?;
-        let mut data = vec![];
-        stream
-            .read_to_eof(&mut data)
-            .await
-            .map_err(|err| OverlayRequestError::UtpError(format!("{err:?}")))?;
-
-        Ok(data)
     }
 
     /// Offer is sent in order to store content to k nodes with radii that contain content-id
--- a/portalnet/src/overlay_service.rs
+++ b/portalnet/src/overlay_service.rs
@@ -872,13 +872,21 @@ where
                         tokio::spawn(async move {
                             let trace = query_info.trace;
                             let data = match utp_controller
-                                .connect_inbound_stream(connection_id, source, trace.clone())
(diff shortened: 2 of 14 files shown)


portalnet/src/utp_controller.rs
```diff
@@ -15,33 +25,253 @@ use crate::discovery::UtpEnr;
...
+            metrics,
+        }
+    }
+
+    /// Connect with a peer given the connection id, and return the data received from the peer.
+    pub async fn connect_inbound_stream(
+        &self,
+        connection_id: u16,
+        peer: Enr,
+        trace: Option<QueryTrace>,
```
mrferris, line 56: nit: Passing the trace into this function is a bit off putting to me because it's not immediately clear why connecting an inbound stream would require a trace. After looking, I see that it's to pass it back out in an error in case the utp fails, but I wonder if we want to instead just add the trace to the error received in the overlay service so that this function's signature remains as simple as possible. 

That is....unless we want to add utp info to the trace. 🤔

mrferris (approved):

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +25,253 @@ use crate::discovery::UtpEnr;
...
+                false
+            }
+        }
+    }
+
+    /// Accept an outbound stream given the connection id, and transfer the data to the peer.
+    pub async fn accept_outbound_stream(
+        &self,
+        cid: ConnectionId<UtpEnr>,
+        content_id: [u8; 32],
```
KolbyML, line 152:
```
        content_id: [u8; 32],
```
is only used in the debug log. I think we should just remove it in that case. As I think think a debug log justifies passing a variable

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +25,253 @@ use crate::discovery::UtpEnr;
...
+            );
+        };
+    }
+
+    /// Accept an inbound stream given the connection id, and return the data received from the
+    /// peer.
+    pub async fn accept_inbound_stream(
+        &self,
+        cid: ConnectionId<UtpEnr>,
+        content_keys_string: Vec<String>,
```
KolbyML, line 192: same I think we should just remove this log

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +25,253 @@ use crate::discovery::UtpEnr;
...
+
+        // report utp tx as successful, even if we go on to fail to process the
+        // payload
+        self.metrics
+            .report_utp_outcome(UtpDirectionLabel::Inbound, UtpOutcomeLabel::Success);
+        Ok(data)
+    }
+
+    /// Connect with a peer given the connection id, and transfer the data to the peer.
+    pub async fn connect_outbound_stream(&self, cid: ConnectionId<UtpEnr>, data: Vec<u8>) -> bool {
```
KolbyML, line 111: Can't the logic in `connect_outbound_stream` and `accept_outbound_stream` be condensed? there is a ton of duplicate code

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +25,253 @@ use crate::discovery::UtpEnr;
...
         Self {
             utp_socket,
             inbound_utp_transfer_semaphore: Arc::new(Semaphore::new(utp_transfer_limit)),
             outbound_utp_transfer_semaphore: Arc::new(Semaphore::new(utp_transfer_limit)),
+            metrics,
+        }
+    }
+
+    /// Connect with a peer given the connection id, and return the data received from the peer.
+    pub async fn connect_inbound_stream(
```
KolbyML, line 52: Can't the logic in `connect_inbound_stream` and `accept_inbound_stream` be condensed? there is a ton of duplicate code

KolbyML (commented):
It looks pretty good. `utp_controller.rs` has a lot of code duplication so we can cut down on that a lot then I think this is good to go

portalnet/src/utp_controller.rs
njgheorghita: I don't think so... `connect_outbound_stream` establishes the stream via `connect_with_cid` unlike `accept_outbound_stream` which establishes the stream via `accept_with_cid`. Plus they have different `debug!` logs while doing so, and they return different values. The common logic between `connect_outbound_stream` & `accept_outbound_stream` has already been encapsulated into `send_utp_content`, which they both already use..

portalnet/src/utp_controller.rs
njgheorghita: > As I think think a debug log justifies passing a variable

? you do think or you don't think?

KolbyML: Doesn't justify

portalnet/src/utp_controller.rs
njgheorghita: you want to remove both debug logs that use this var?

portalnet/src/utp_controller.rs
njgheorghita: I don't think there's a ton of duplicate code. But I did extract the reading of the utp stream data into `receive_utp_content`, though I'm not too sure we won too many lines with that change

portalnet/src/utp_controller.rs
njgheorghita: i'm leaning towards leaving (all the debug logs) in, but just removing the `content_keys_string` and `content_id` vars from those logs

portalnet/src/utp_controller.rs
KolbyML: I will take a shot at it then

portalnet/src/utp_controller.rs
njgheorghita: nvm, we can actually just remove them since we emit a `debug!` log inside `OverlayService` if an error is returned

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +22,235 @@ use crate::discovery::UtpEnr;
...
+        peer: Enr,
+    ) -> anyhow::Result<Vec<u8>> {
+        self.metrics
+            .report_utp_active_inc(UtpDirectionLabel::Inbound);
+        let cid = utp_rs::cid::ConnectionId {
+            recv: connection_id,
+            send: connection_id.wrapping_add(1),
+            peer: UtpEnr(peer),
+        };
+        let stream = match self.connect_with_cid(cid.clone(), *UTP_CONN_CFG).await {
```
morph-dev, line 61: This is my personal preference. If you don't like, feel free to ignore it.

I don't like the use of `match` for `Option` or `Result` when one case is trivial. I find the following code more readable.

```rust
let stream = self.connect_with_cid(...)
    .await
    .map_err(|err| {
        ...
        anyhow!(...)
    })?;
```

This applies to many places in this file.

KolbyML:
I dmed nick on discord but I will make a PR on this PR here is a POC of what I am thinking of https://github.com/njgheorghita/trin/pull/10 I am almost done

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +22,235 @@ use crate::discovery::UtpEnr;
...
+                    cid.recv,
+                    peer = ?cid.peer.client(),
+                    "Unable to establish uTP conn based on Content response",
+                );
+                return Err(anyhow!("unable to establish utp conn"));
+            }
+        };
+
+        // receive_utp_content handles metrics reporting of successful & failed rx
+        match Self::receive_utp_content(stream, self.metrics.clone()).await {
```
morph-dev, line 80: And even more here:

```rust
Self::receive_utp_content(...)
    .await
    .map_err(|err| {
        debug!(...);
        anyhow!(...)
    })
```

portalnet/src/utp_controller.rs
```diff
@@ -15,33 +22,235 @@ use crate::discovery::UtpEnr;
...
+            Ok(data) => Ok(data),
+            Err(err) => {
+                debug!(%err, cid.send, cid.recv, peer = ?cid.peer.client(), "error reading data from uTP stream, while handling a FindContent request.");
+                Err(anyhow!("error reading data from uTP stream"))
+            }
+        }
+    }
+
+    /// Connect with a peer given the connection id, and transfer the data to the peer.
```
morph-dev, line 89: nitpick: I would say it makes more sense and it's consistent if `connect_outbound_stream` returns `anyhow::Result<()>`. You can also do proper error logging one level up.

portalnet/src/utp_controller.rs
morph-dev: If I'm not mistaken, the caller already logs if there is failure.
I think this function can just return `Err` with message as a reason, and caller can log all details, including the reason.

morph-dev (commented):
I started reviewing but just noticed that @KolbyML is adding changes on top of it. 
So I will just share my comments so far.

In general, it looks good from my side.

KolbyML:
This PR is ready https://github.com/njgheorghita/trin/pull/10.
me and nick talked on dms and the plan is `aka merge it into my branch? and then we'll merge it into trin master w/ the same pr?` when he gets home

portalnet/src/overlay.rs
```diff
@@ -535,20 +543,16 @@ where
...
+            .connect_inbound_stream(cid)
             .await
-            .map_err(|err| OverlayRequestError::UtpError(format!("{err:?}")))?;
-
-        Ok(data)
+            .map_err(|err| OverlayRequestError::ContentNotFound {
+                message: format!("Unable to locate content on the network: {err:?}"),
+                utp: true,
+                trace: None,
+            })
```
KolbyML, line 555: we already through this error in the utp_controller no? so we can probably remove this `map_err`

portalnet/src/overlay_service.rs
```diff
@@ -873,78 +862,37 @@ where
...
+                                        let _ = responder.send(Err(
+                                            OverlayRequestError::ContentNotFound {
+                                                message: format!(
+                                                    "Unable to locate content on the network: {e}"
+                                                ),
+                                                utp: true,
+                                                trace,
+                                            },
+                                        ));
+                                    }
```
KolbyML, line 892: `connect_inbound_stream` returns this error I would be a fan to pass trace through `connect_inbound_stream` just use the error returned directly instead of wrapping it twice

KolbyML (approved):
Looks pretty much good to go. I think we should do my 2 recommendations as then the error creation is contained inside utp_controller. Less stuff to remember across the codebase.

So I do think we should pass the trace in or at least figure out a way to avoid rewrapping the errors through by `connect_inbound_stream`

portalnet/src/overlay_service.rs
KolbyML: We create `OverlayRequestError::ContentNotFound {}` in `connect_inbound_stream` and here I think we should remove the one here as doing it in both places is redundent.

portalnet/src/utp_controller.rs
njgheorghita: Yup, I'd agree with that style guideline, updated the code. Except for in places where we need to return a non-error value from within the closure (eg. a bool inside `outbound_stream`), which I don't think is possible to return early from inside a closure in Rust.

njgheorghita:
@KolbyML Ahhh, I took the opposite approach where we should remove all concepts of `OverlayServiceError` from the `UtpController`. It seems a bit weird to handle some of those errors inside `UtpController` and some inside `OverlayService` and imo it's much more consistent to handle `OverlayServiceError`s inside `OverlayService`

KolbyML (approved):
:shipit:

PR #1204, opened 2024-03-07
fix(bridge): fix bridge cli args to be compatible with client metrics

njgheorghita:
### What was wrong?
`portal-bridge` was using `url::Url` for the metrics url, when core `trin` uses `SocketAddr`. 

### How was it fixed?
To avoid some extra string parsing, I just updated the bridge's metrics url to be a `SocketAddr` as well

### To-Do


njgheorghita:
@KolbyML this along with https://github.com/ethereum/cluster/pull/955 should get us bridge metrics reporting on `trin-bench`

KolbyML (approved):

PR #1205, opened 2024-03-07
refactor(portalnet): split out overlay_service into submodules

njgheorghita:
### What was wrong?
`overlay_service.rs` is too big.

### How was it fixed?
Split out the components inside `overlay_service.rs` into their own separate modules.

### To-Do


morph-dev (approved):
:rocket:

PR #1207, opened 2024-03-08
Refact/portalnet

AmirH-A:
### What was wrong?

- check size zero for find node distance 
- handle panic for distance between 1 and 127

### To-Do


KolbyML:
Are you able to recreate a panic without this check in place? Would you be able to post instructions for the non-intended behavior this PR is meant to resolve

KolbyML:
closing as there has been no response for almost a year

PR #1209, opened 2024-03-11
feat(portalnet): add accept_queue to handle incoming content keys

### LINKED ISSUE #856: Queue accepted content
# Problem
Currently, when `trin` handles an `OFFER` message, it cycles through every offered `content_key` and checks with storage if [the key is within radius and not already present in the data store](https://github.com/ethereum/trin/blob/master/portalnet/src/overlay_service.rs#L1194-L1209). This strategy ignores content keys that have already been "accepted" but are still being transported over utp. This can lead to a node initiating multiple utp streams with different peers for the same content key.

Given the relatively small size of our network, it seems quite reasonable that our bridge nodes will be offering the same content key (eg. `latest`) to the same node in a relatively short timeframe. Due to the large size of some of these content values (eg. bodies/receipts), a node might accept the same content key from each bridge node before it has a chance to complete a single utp transfer & store the content value. 

Running multiple utp streams for the same content key adds bloat to the network. 

### Sidenote
While running trin, it's fairly common that I've seen the [Accepted content outside radius or already stored](https://github.com/ethereum/trin/blob/master/portalnet/src/overlay_service.rs#L1637C34-L1637C83) log, which is a solid indicator to me that we are duplicating our accepted content keys. 

# Solution
For the sake of this issue, a content key is considered in "processing" from the instant it is `Accepted` until the moment it is saved into local storage. This covers the timeframe that it takes to transport the actual content and validate it. This can take some time depending on how large the content value is or how many network requests are needed to fetch the relevant validation data. 

Introduce some kind of queue to the `OverlayService` that maintains a record of every content key which is currently "processing".

The order of operations for deciding whether or not to `Accept` an `Offer`ed content key looks something like this...

1. Check with storage if key is within radius & not already present in the data store (same as before).
2. Check with the queue if the content key is currently "processing"
  A. If yes, respond to the `Offer` that we do not want to `Accept` this content key.
  B. If no, save it as "processing", and respond to the `Offer` that we want to `Accept` this content key. 
3. After successful "processing" or if any step of "processing" fails we want to be sure we clear the content key from the queue.

This is the simplest solution I can think of, and certainly not optimal. But it seems to me like starting with the simplest solution is best. 

### Possible Improvements
- Trigger a recursive find content lookup if any step of "processing" fails. 
- Store the peer's enr in the queue for all content keys that go to step 2A, so that if "processing" fails we can send that peer a `FindContent` request, with a relatively high degree of confidence that they will have that content

njgheorghita:
### What was wrong?

Fixes #856

Created an `AcceptQueue` to help combat the "Thundering Herd" problem.

### How was it fixed?
- `AcceptQueue` which tracks the content keys that are currently inbound via `OFFER`/`ACCEPT`
- Some minor refactoring to `UtpController`

### To-Do


portalnet/src/offer_queue.rs
```diff
@@ -0,0 +1,313 @@
...
+                    err
+                );
+            }
+        } else {
+            warn!(
+                "Failed to process content key: {content_key}, but a corresponding OfferQueue record was not found"
+            );
+        }
+    }
+}
```
KolbyML, line 145: Shouldn't this have
```rust
        self.content_key_map.insert(
            content_key.clone(),
            SeenPeers {
                origin: fallback_peer.clone(),
                fallback: node_ids.fallback
            },
        );
```

Or is it intentional to only after attempt 1 more time after a failure occured

portalnet/src/offer_queue.rs
```diff
@@ -0,0 +1,313 @@
...
+        let _ = future::poll_fn(|cx| self.content_key_map.poll_expired(cx)).now_or_never();
+        if let Some(mut node_ids) = self.content_key_map.remove(content_key) {
+            if node_ids.origin == *peer || node_ids.fallback.contains(peer) {
+                warn!(
+                    "Received multiple offers containing the same content key: {} from peer: {}",
+                    content_key,
+                    peer.node_id(),
+                );
+                return false;
+            }
```
KolbyML, line 71:
```
self.content_key_map.remove(content_key) {
```
We are removing the key shouldn't we have `self.content_key_map.insert(content_key.clone(), node_ids);` before the return

njgheorghita: Ahh yeah, good catch. I also added a test case to cover this

portalnet/src/offer_queue.rs
KolbyML: Oh nvm I see how my statement is wrong. But shouldn't we try more then 1 fallback node if the first one fails

njgheorghita: My thinking is that it's just simpler to erase the record from the queue. The queue will submit a `FINDCONTENT` request to the fallback node. This request won't use the `OfferQueue` anyways, and I'm not sure it's worth the complexity to track a content key beyond submitting the `FINDCONTENT` request to the overlay.

I guess the one case that this doesn't handle, is that the `OfferQueue` will then mark incoming offers for the content key as ones we should accept, during the execution of the `FINDCONTENT` request, when technically we don't want to mark these as valid.

It wouldn't be too hard to just send a `callback` channel along with the `FINDCONTENT` request, in which case we could repeat this process, with all the fallback peers, until we've guaranteed a successful `FINDCONTENT` request.

But, I just don't think it's worth the complexity right now. At the very worst, we'll have 2 concurrent incoming processes for the same content key, but this is only for a somewhat unique edge case. It's still an improvement over the current process. And, if we continue to see a lot of those `Unable to store accepted content, content already stored` errors, then we can improve the `OfferQueue` to handle this / any other edge case.

KolbyML: I am having trouble finding where we handle the accepted content from the FC you are sending in your pr
Yeah currently if the first one fails and a FC happens it isn't handled.

Currently this is sending a FC and not handling the response so even if the fallback has the data we won't get it

njgheorghita: iiuc, inside `OverlayService` we handle `FindContent` requests via `OverlayService::process_received_content()` which will validate & store the received content. Imo, we don't really care about the actual data at this level, as long as it was validated & stored locally. So there's no need to return it via a callback

KolbyML: But that is only if you create the overlay request using `OverlayCommand::FindContentQuery` where as here you are using `OverlayCommand::Request`

KolbyML: A FC Request command takes a channels for a responder and calls `OverlayService::process_content()` which only marks a query if provided as complete

This isn't a query and we don't pass a channel for a response hence we aren't handling the result

njgheorghita: I'm not positive... But if what you're saying is correct, then that means we aren't doing any validation on content received from a simple `FindContent` query initiated over jsonrpc? iiuc, that's essentially what I'm doing here by using the `OverlayCommand::Request` wrapper rather than the `OverlayCommand::FindContentQuery` wrapper. I'll dig more into this tomorrow and confirm/deny

njgheorghita: Ahhh, ok yeah, looks like I had some invalid assumptions going on here. 

Well, the problem with submitting a `OverlayCommand::FindContentQuery` is that it doesn't accept a target `enr`. eg. it just starts the recursive find content lookup process. I felt like it was better to submit a `FindContent` request to a peer who has offered us that content key rather than starting a RFC lookup. 

I'd also like to avoid using `OverlayCommand::Request` and processing the content received via a callback here inside `OfferQueue` (eg doing the validation / storing). But maybe that's not as thorny as it seems...

portalnet/src/utp_controller.rs
```diff
@@ -51,6 +51,60 @@ impl UtpController {
         }
     }
 
+    pub fn cid(&self, peer: UtpEnr, is_initiator: bool) -> ConnectionId<UtpEnr> {
+        self.utp_socket.cid(peer, is_initiator)
+    }
+
+    // `try_acquire_owned()` isn't blocking and will instantly return with
+    // `Some(TryAcquireError::NoPermits)` error if there isn't a permit avaliable
```
KolbyML, line 59:
```suggestion
    /// `try_acquire_owned()` isn't blocking and will instantly return with
    /// `Some(TryAcquireError::NoPermits)` error if there isn't a permit avaliable
```

These are doc strings hence should use /// right?

portalnet/src/utp_controller.rs
```diff
@@ -51,6 +51,60 @@ impl UtpController {
...
+            .clone()
+            .try_acquire_owned()
+        {
+            Ok(permit) => Some(permit),
+            Err(_) => None,
+        }
+    }
+
+    // `try_acquire_owned()` isn't blocking and will instantly return with
+    // `Some(TryAcquireError::NoPermits)` error if there isn't a permit avaliable
```
KolbyML, line 72:
```suggestion
    /// `try_acquire_owned()` isn't blocking and will instantly return with
    /// `Some(TryAcquireError::NoPermits)` error if there isn't a permit avaliable
```
These are doc strings hence should use /// right?

portalnet/src/offer_queue.rs
```diff
@@ -0,0 +1,350 @@
...
+            content_key_map: HashMapDelay::new(OFFER_QUEUE_TIMEOUT),
+            command_tx,
+        }
+    }
+
+    // Checks if we should accept the given content key based on the queue state.
+    // - Polls for expired items, which will remove them from the queue.
+    // - If the key is not in the queue, it is added to the queue and returns true.
+    // - If the key is in the queue, the seen peer is stored, and it returns false.
+    pub fn should_accept(&mut self, content_key: &TContentKey, peer: &Enr) -> bool {
```
KolbyML, line 60:
```suggestion
    /// Checks if we should accept the given content key based on the queue state.
    /// - Polls for expired items, which will remove them from the queue.
    /// - If the key is not in the queue, it is added to the queue and returns true.
    /// - If the key is in the queue, the seen peer is stored, and it returns false.
    pub fn should_accept(&mut self, content_key: &TContentKey, peer: &Enr) -> bool {
```

These are doc strings hence should use /// right?

portalnet/src/offer_queue.rs
KolbyML: Maybe the best case forword is to refactor `OverlayService::process_content()` and `Overlay::send_find_content()`

> I'd also like to avoid using OverlayCommand::Request and processing the content received via a callback here inside OfferQueue (eg doing the validation / storing). But maybe that's not as thorny as it seems...

I think we should use `OverlayCommand::Request` now that you mentioned

> Well, the problem with submitting a OverlayCommand::FindContentQuery is that it doesn't accept a target enr. eg. it just starts the recursive find content lookup process

We could use tokio::spawn to make it non blocking so `OverlayCommand::Request` with a reponse channel would work then

KolbyML: Maybe one route would be if no responder is provider handle content in  `OverlayService::process_content()`

portalnet/src/offer_queue.rs
njgheorghita: Imo, these are more implementation details, and don't necessarily deserve "docstring" level status. But I don't mind upgrading them at all

portalnet/src/offer_queue.rs
KolbyML: I think providing `OverlayCommand::Request` a response channels calling `tokio::spawn` then handling might be the play though cause then we  could configure how many fallbacks we want to try.

I guess there are quite a few ways we can fix this issue

portalnet/src/overlay/service.rs
```diff
@@ -1574,19 +1584,26 @@ where
...
-                debug!(err, content_key = ?content_keys_string[index], "Process uTP payload tokio task failed:");
-                None
-            }))
+            .filter_map(|(index, content)| match content {
+                Ok(content) => {
+                    offer_queue.write().successful_processing(&content_keys[index]);
+                    content
+                }
+                Err(err) => {
+                    offer_queue.write().failed_to_process(&content_keys[index]);
```
ogenev, line 1593: Reading this doesn't imply that we are sending overlay requests inside `failed_to_process` which is a hidden side effect of this function. Why do we not send the `FIndContent` request from here?

njgheorghita: Well, to be honest, it was primarily to reduce the size of `OverlayService` but I see your point and agree that it makes more sense to handle this inside the service, regardless of the additional code

portalnet/src/offer_queue.rs
njgheorghita: Actually, yeah these do qualify as "docstrings" imo, it was your other comments that I considered more "implementation details"

portalnet/src/overlay/service.rs
njgheorghita: Though, handling this inside the `OverlayService` does make it significantly more difficult to test... I do still agree that it makes more sense

njgheorghita:
Not quite ready for another round of reviews here. Still have some testing that I want to take care of. I'm leaning towards a single fallback attempt for every failed content key just to keep things simple for now.

portalnet/src/overlay/service.rs
```diff
@@ -2461,6 +2589,82 @@ fn pop_while_ssz_bytes_len_gt(enrs: &mut Vec<SszEnr>, max_size: usize) {
     }
 }
 
+/// References to `OverlayService` components required for processing
+/// a utp stream. This is basically a utility struct to avoid passing
+/// around a large number of individual references.
+struct UtpProcessing<TValidator, TStore, TContentKey>
```
njgheorghita, line 2595: I'm not sure about this name, but it's quite useful to have this structure to pass around vs. individually handling its components. It seems like there are opportunities to further refactor the `OverlayService` using this datatype, but I felt best to leave that for a future pr to avoid irrelevant bloat to this pr.

portalnet/src/offer_queue.rs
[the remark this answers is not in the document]
njgheorghita: Well, I agree that `should_accept` might be misleading. But, I'm not sure it's worth splitting the logic out into separate methods, since they're only used together. I updated the name to `add_key_to_queue` which should communicate a bit better the actual functionality of the method

portalnet/src/offer_queue.rs
```diff
@@ -0,0 +1,350 @@
...
+        // poll for expired items, which will remove them from the queue
+        let _ = future::poll_fn(|cx| self.content_key_map.poll_expired(cx)).now_or_never();
+        if let Some(mut node_ids) = self.content_key_map.remove(content_key) {
```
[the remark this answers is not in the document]
njgheorghita: Ehhh, I'm not sure how valuable it is to use `get` over `remove`. I made some tweaks to the logic so that it's a bit smoother, but I'm happy to switch to `get` if you still feel like it's worthwhile

portalnet/src/offer_queue.rs
```diff
@@ -0,0 +1,350 @@
...
+            if node_ids.fallback.is_empty() {
+                debug!(
+                    "Failed to process content key: {content_key}, but no fallback peers found."
+                );
+                return;
+            }
+            // choose a random seen peer as the fallback peer
+            // and send a FINDCONTENT request to them
+            node_ids.fallback.shuffle(&mut thread_rng());
```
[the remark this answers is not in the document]
njgheorghita: I think `remove` is the right choice here rather than `choose` since `choose` returns a reference

njgheorghita:
Ok, ready for another round of :eyes: here. Had to do some more tinkering to the `OverlayService` than expected, to allow for processing content keys individually, rather than in a batch (like we were doing inside `process_accept_utp_stream`).

portalnet/src/accept_queue.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// If the key is not in the queue, it is added to the queue and returns true.
+    /// If the key is in the queue, the seen peer is stored as a fallback, and it returns false.
+    /// Also polls for expired items, which will remove them from the queue.
+    pub fn add_key_to_queue(&mut self, content_key: &TContentKey, peer: &Enr) -> bool {
+        // poll for expired items, which will remove them from the queue
+        let _ = future::poll_fn(|cx| self.content_key_map.poll_expired(cx)).now_or_never();
+        if let Some(mut seen_peers) = self.content_key_map.remove(content_key) {
+            if seen_peers.origin == *peer || seen_peers.fallback.contains(peer) {
+                warn!(
+                    "Received multiple offers containing the same content key: {content_key} from peer: {}",
```
njgheorghita, line 62: One interesting thing to note is that I'm seeing this log quite often when running this locally. I was expecting to hardly run into the case when the same peer sends multiple offers for the same content key. I'm very skeptical that it's due to a bug in this pr, since `add_key_to_queue` is only used once inside `OverlayService`. But, this might indicate a bug elsewhere in our logic where a node would be sending multiple offers for the same content key?

KolbyML: Post the enr in the log

portalnet/src/overlay/service.rs
```diff
@@ -2461,6 +2588,82 @@ fn pop_while_ssz_bytes_len_gt(enrs: &mut Vec<SszEnr>, max_size: usize) {
...
+            validator: Arc::clone(&self.validator),
+            store: Arc::clone(&self.store),
+            metrics: self.metrics.clone(),
+            kbuckets: Arc::clone(&self.kbuckets),
+            command_tx: self.command_tx.clone(),
+            utp_controller: Arc::clone(&self.utp_controller),
+            accept_queue: Arc::clone(&self.accept_queue),
+        }
+    }
+}
```
KolbyML, line 2647: Should these be using `'static` for the reason I mentioned before, maybe arc's are a better option?

portalnet/src/overlay/service.rs
```diff
@@ -1127,31 +1131,122 @@ where
...
+                    // Spawn a fallback FINDCONTENT task for each content key
+                    // in a payload that failed to be received.
+                    //
+                    // We spawn these additional fallback FINDCONTENT tasks using
+                    // the same semaphore permit that was initially acquired for
+                    // the ACCEPT utp stream. This might cause the utp active count
+                    // to temporarily exceed the max, but that seems preferable to
+                    // requiring additional permits for each individual content key,
+                    // which could drop the fallback request entirely if there are no
+                    // more permits available.
```
KolbyML, line 1153:
```suggestion
                    // Spawn a fallback FINDCONTENT task for each content key
                    // in a payload that failed to be received.
                    //
                    // We spawn these additional fallback FINDCONTENT tasks using
                    // the same semaphore permit that was initially acquired for
                    // the ACCEPT utp stream. 
```

There is no "max uTP count" There is a max uTP count initated by the overlay service. RPC calls are exempt from this requirement. So we can remove that extra part of the comment. As this code is functioning as intented

KolbyML (commented):
I am concerned with the use of `'static`. Maybe it would be better to use generic lifetimes or arc's.

From what I understand `'static` should only be used for const's. If we need to define lifetimes for non-consts  `generic life times` or `arc's` should be used instead

The reason for this is `'static` marks an object that it lives as long as the program hence should never be deallocated. In other words `'static` marks objects as static like in C/C++.


I could be wrong or miss-understanding something, if so I would be very pleased with understanding why this isn't an issue so I can improve. It is just when I see this tons of alarm bells go off in my brain, so if there is nothing wrong here I would be fasinated with why that isn't the case so I can grow to become more competent.

KolbyML (approved):
:shipit:

portalnet/src/accept_queue.rs
```diff
@@ -0,0 +1,206 @@
...
+pub struct AcceptQueue<TContentKey>
+where
+    TContentKey: OverlayContentKey,
+{
+    // a map of content keys actively being transferred / processed
+    // pointing to seen peers that have offered them the content
+    content_key_map: HashMapDelay<TContentKey, SeenPeers>,
+}
+
+const OFFER_QUEUE_TIMEOUT: Duration = Duration::from_secs(120);
```
ogenev, line 35: I would move this to the beginning of the file.

portalnet/src/accept_queue.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// If the key is in the queue, the seen peer is stored as a fallback, and it returns false.
+    /// Also polls for expired items, which will remove them from the queue.
+    pub fn add_key_to_queue(&mut self, content_key: &TContentKey, peer: &Enr) -> bool {
+        // poll for expired items, which will remove them from the queue
+        let _ = future::poll_fn(|cx| self.content_key_map.poll_expired(cx)).now_or_never();
+        if let Some(mut seen_peers) = self.content_key_map.remove(content_key) {
+            if seen_peers.origin == *peer || seen_peers.fallback.contains(peer) {
+                warn!(
+                    "Received multiple offers containing the same content key: {content_key} from peer: {peer}"
+                );
```
ogenev, line 63: If the key is in the queue, we will get inside this block and then jump to line 73 and return `true`. I think in this case we want to return `false`, because the key is already included in the queue?

ogenev (approved):
⛵

portalnet/src/accept_queue.rs
njgheorghita: I don't think so, we jump to line `70` not `73` - and from line `70` we re-insert the "removed" `seen_peers` and then return a `false`. This case is also covered by the `test_queue_keeps_record_after_duplicate_offers_from_same_peer` test

PR #1210, opened 2024-03-12
feat: add state peertests

morph-dev:
### What was wrong?

The state network didn't have any peer tests.

### How was it fixed?

Added peer tests for the state network

### To-Do


KolbyML (approved):
:shipit:

PR #1212, opened 2024-03-14
feat(ethportal-api): add methods to build beacon chain proofs

ogenev:
### What was wrong?
No implementation to generate beacon chain block proofs

### How was it fixed?
- Add methods to build beacon chain block proofs for Bellatrix execution block headers.
- Add some additional serde and ssz tests.

The [rs_merkle](https://crates.io/crates/rs_merkle) library is used to build the ssz merkle trees. We also have merkle tree implementation in [trin_validation
](https://github.com/ethereum/trin/tree/master/trin-validation/src/merkle) but it requires some refactoring to be used inside `ethportal-api` and this can be done in additional PR or create a community issue if we decide to keep our implementation.

### To-Do


KolbyML (approved):
:shipit:

PR #1213, opened 2024-03-14
fix: add support for blob transaction types

njgheorghita:
### What was wrong?
Initial, basic support for new blob transaction type. Still have some testing to do, but this fix gets the "latest" bridge operational again. Mostly creating this pr just to signal that I'm working on this.

### To-Do


njgheorghita:
closing in favor of #1214

PR #1214, opened 2024-03-15
feat: add EIP-4788 parent_beacon_block_root to execution header

KolbyML:
### What was wrong?
We didn't include the field added by EIP-4788
### How was it fixed?
by adding the field


ethportal-api/src/types/execution/header.rs
```diff
@@ -58,6 +55,18 @@ pub struct Header {
     pub base_fee_per_gas: Option<U256>,
     /// Withdrawals root from execution payload. Introduced by EIP-4895.
     pub withdrawals_root: Option<H256>,
+    /// Blob gas used. Introduced by EIP-4844
+    pub blob_gas_used: Option<U64>,
+    /// Excess blob gas. Introduced by EIP-4844
+    pub excess_blob_gas: Option<U64>,
+    /// The hash of the parent beacon block's root is included in execution blocks, as proposed by
```
njgheorghita, line 62: Imo, just simply `The parent beacon block's root hash. Introduced by EIP-4788` is fine. If users want to understand more, they can go read the eip themselves

ethportal-api/src/types/execution/header.rs
```diff
@@ -157,16 +183,31 @@ impl Decodable for Header {
             nonce: Some(rlp.val_at(14)?),
             base_fee_per_gas: None,
             withdrawals_root: None,
+            blob_gas_used: None,
+            excess_blob_gas: None,
+            parent_beacon_block_root: None,
         };
 
-        if header.number >= LONDON_BLOCK_NUMBER {
+        if rlp.at(15).is_ok() {
```
njgheorghita, line 191:
```rust
if let Ok(val) = rlp.at(15) {
    header.base_fee_per_gas = Some(val)
}
```

same goes for the following statements

njgheorghita (approved):
Nice find! Can't believe I missed a whole EIP 🙈 A few nitpicks. And there's still a lot more testing to do with the `BlobTransaction` type to make sure that our `Bodies` type is still valid. But, I'm happy to tackle that in a follow-up pr if we want to go ahead and merge this. At the very least, this will fix "latest" headers.

ethportal-api/src/types/execution/header.rs
njgheorghita: I guess this does assume that the `Header` json values are correctly constructed. Aka, no pre-london headers will contain `base_fee_per_gas`, which I think is a safe assumption. And if a source header is incorrectly constructed, we'll know when we try to hash it

ethportal-api/src/types/execution/header.rs
```diff
@@ -557,6 +605,109 @@ mod tests {
             &hex_decode("0x17cf53189035bbae5bce5c844355badd701aa9d2dd4b4f5ab1f9f0e8dd9fea5b")
                 .unwrap(),
         );
+        assert_eq!(header.number, 17034871);
+        assert_eq!(header.hash(), expected_hash);
+    }
+
+    #[test]
+    fn dencun_rlp() {
```
njgheorghita, line 613: I think we can remove this test, it was mostly a helper test while debugging, and the following test covers the same functionality

ethportal-api/src/types/execution/header.rs
```diff
@@ -557,6 +605,109 @@ mod tests {
...
+            blob_gas_used: Some(U64::from_str("020000").unwrap()),
+            excess_blob_gas: Some(U64::from_str("0").unwrap()),
+            parent_beacon_block_root: None,
+        };
+        assert_eq!(decoded, expected);
+
+        let expected_hash =
+            H256::from_str("0x10aca3ebb4cf6ddd9e945a5db19385f9c105ede7374380c50d56384c3d233785")
+                .unwrap();
+        assert_eq!(decoded.hash(), expected_hash);
```
njgheorghita, line 687: Though, I would re-encode the `Header` here, and assert that it matches the original value

ethportal-api/src/types/execution/header.rs
```diff
@@ -557,6 +605,109 @@ mod tests {
...
+            gas_limit: U256::from_str("0x7fffffffffffffff").unwrap(),
+            gas_used: U256::from_str("0x02a865").unwrap(),
+            timestamp: 0x079e,
+            extra_data: vec![0x42],
+            mix_hash: Some(
+                H256::from_str("56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421")
+                    .unwrap(),
+            ),
+            nonce: Some(H64::zero()),
+            base_fee_per_gas: Some(U256::from_str("9").unwrap()),
```
njgheorghita, line 673: Why are some of these `0x` prefixed (eg. `gas_used`) and some not? I think we should be consistent

ethportal-api/src/types/execution/header.rs
```diff
@@ -557,6 +605,109 @@ mod tests {
...
+            std::fs::read_to_string("../test_assets/mainnet/block_19433902_value.json").unwrap();
+        let response: Value = serde_json::from_str(&body).unwrap();
+        let header: Header = serde_json::from_value(response["result"].clone()).unwrap();
+        let expected_hash =
+            H256::from_slice(&hex_decode(response["result"]["hash"].as_str().unwrap()).unwrap());
+        assert_eq!(header.number, 19433902);
+        assert_eq!(header.hash(), expected_hash);
+    }
+
+    #[test]
```
njgheorghita, line 702: I think you can also just use `rstest` here for the two `post_dencun_header_*` tests

ethportal-api/src/types/execution/header.rs
```diff
@@ -202,6 +202,10 @@ impl Decodable for Header {
             header.parent_beacon_block_root = Some(val)
         }
 
+        if rlp.at(19).is_ok() {
```
njgheorghita, line 205: @KolbyML I think this was added incorrectly

ethportal-api/src/types/execution/block_body.rs
```diff
@@ -470,6 +470,7 @@ mod tests {
     #[case(TX17, 1544975)]
     // EIP1559 w/ populated access list
     #[case(TX6, 41942)]
+    // todo blob
```
njgheorghita, line 473: This should also be removed since it was handled in the last commit

portal-bridge/src/types/full_header.rs
```diff
@@ -162,6 +162,23 @@ mod tests {
         block_body.validate_against_header(&header).unwrap();
     }
 
+    #[rstest::rstest]
+    // without blob txs
+    #[case::legacy("block_19433902_value.json")]
+    // with blob txs
+    #[case::legacy("block_19433903_value.json")]
```
njgheorghita, line 169: Not sure what happened, but it seems like these should be `"../test_assets/mainnet/block_17034873_value.json"` ...
or you could do `format!(""../test_assets/mainnet/block_{}_value.json", block_number)` below

PR #1215, opened 2024-03-16
chore: remove non-needed mutex from rpc handler

KolbyML:
### What was wrong?
we had a non needed mutex
### How was it fixed?
removed it

### DIFF
--- a/trin-beacon/src/jsonrpc.rs
+++ b/trin-beacon/src/jsonrpc.rs
@@ -16,7 +16,7 @@ use ethportal_api::{
 };
 use portalnet::overlay::errors::OverlayRequestError;
 use serde_json::{json, Value};
-use tokio::sync::{mpsc, Mutex, RwLock};
+use tokio::sync::{mpsc, RwLock};
 use tracing::error;
 use trin_storage::ContentStore;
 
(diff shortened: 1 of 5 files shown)


trin-beacon/src/jsonrpc.rs
```diff
@@ -25,14 +25,13 @@ use crate::network::BeaconNetwork;
 /// Handles Beacon network JSON-RPC requests
 pub struct BeaconRequestHandler {
     pub network: Arc<RwLock<BeaconNetwork>>,
```
morph-dev, line 27: Since you are at this topic, is there a reason why network should be wrapped around RwLock?
If I'm not mistaken, every possible usage would need `read` access anyway and there is no functionality that needs `write` access.

Similar in few other places.

morph-dev (commented):
Was this causing any issues, or was it just not needed?

In general, looks good to me, but somebody else should take a look as well.

:shipit:

KolbyML:
> Was this causing any issues, or was it just not needed?
> 
> In general, looks good to me, but somebody else should take a look as well.
> 
> :shipit:

To my knowledge this just isn't needed so this is more just housekeeping removing redundent code

trin-beacon/src/jsonrpc.rs
KolbyML: I didn't look into it yet, but you are right it would probably be good to look into this

ogenev (approved):

njgheorghita (approved):

morph-dev (approved):

mrferris (approved):

trin-beacon/src/jsonrpc.rs
KolbyML: If we can remove it I will handle it in a follow up PR as this PR has already been reviewed/approved with the current changes

PR #1219, opened 2024-03-19
feat(bridge): launch bridge with custom discovery and http ports

njgheorghita:
### What was wrong?
Our bridge is currently stuck inbetween two paradigms. 
- a single bridge gossiping via a single node
- a single bridge gossiping via more than one nodes

Personally, I'd like to only support the first use case. Especially, if you want to run two bridges in different modes (as I do). But, that's a different discussion. This pr basically allows for a user to run two bridges from the same machine, which isn't possible unless you can configure the discovery port.

### How was it fixed?
- added ability to customize discovery port & http port via cli
- custom http port is not necessarily a feature I need, but it's so closely related to the discovery port, that I figured it was better to update the functionality for both ports to maintain consistency

### To-Do

### DIFF
--- a/portal-bridge/src/cli.rs
+++ b/portal-bridge/src/cli.rs
@@ -2,13 +2,15 @@ use std::{net::SocketAddr, path::PathBuf, str::FromStr};
 
 use clap::{Parser, Subcommand};
 use ethereum_types::H256;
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
:shipit:

PR #1220, opened 2024-03-19
refactor: reduce noisy log level

njgheorghita:
### What was wrong?
This log is actually very noisy when it's being run. Turning down the log level to reduce log bloat, created issue to track investigation of the cause log noisiness (#1221)

### How was it fixed?
- reduced log level

### To-Do

### DIFF
--- a/portalnet/src/accept_queue.rs
+++ b/portalnet/src/accept_queue.rs
@@ -58,7 +58,7 @@ where
         let _ = future::poll_fn(|cx| self.content_key_map.poll_expired(cx)).now_or_never();
         if let Some(mut seen_peers) = self.content_key_map.remove(content_key) {
             if seen_peers.origin == *peer || seen_peers.fallback.contains(peer) {
-                warn!(
+                debug!(
                     "Received multiple offers containing the same content key: {content_key} from peer: {peer}"
                 );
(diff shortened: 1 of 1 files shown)


KolbyML:
I don't think we should reduce this logs level. It is noisy because we have a problem no? so it is doing its job.

If we fix the problem it will stop being noisy so I think solving the root problem is the solution in this case

mrferris:
I agree that this indicates something wrong with some client on the network that we don't want to ignore. Have you checked the client string in the ENR? I'm unable to reproduce locally.

njgheorghita:
Looking back into this, I'm still seeing a fairly significant number of duplicate offers. However, they all appear to be from fluffy clients, judging by the client info in the enr. Given that fluffy doesn't yet support any kind of offer queue to prevent duplicate offers, I think it's reasonable to reduce the log level to `debug` since it's not necessarily indicative of an error. Though, I'm surprised that @mrferris wasn't able to reproduce this locally? Given the amount that I'm seeing it, this should be a fairly common experience.

ogenev (approved):
I can replicate the issue.

I think a warning log indicates that the user can do something about this, but in this case it is something that can't be prevented.

PR #1222, opened 2024-03-19
feat: add bridge metrics for history bridge

KolbyML:
### What was wrong?
We didn't have an easy visualization to see how our bridges were performing. The goal being we can look at a graph and in a few seconds be able to form an assumptions on the data. A goal of this is to get a better understanding of how our bridge is performing. 
### How was it fixed?
By adding metrics I add a few

- info this is so we can see what bridge and mode we are running from grafana
- histogram as you can calculate so much from this data as seen when I added it to storage. A few examples (time it takes to x), (how many are we performing a second), to name a few.

The goal is to start off basic, but with something useful and I think this strikes a healthy balance of that. This PR lays the ground work for future metrics etc etc


mrferris (approved):
Looks good!

(be sure to squash & merge to get rid of that first commit)

PR #1223, opened 2024-03-19
refactor: remove RwLock from History/Beacon rpc handler

KolbyML:
### What was wrong?
As said here https://github.com/ethereum/trin/pull/1215#discussion_r1527267279 we also don't need the RwLock held on the rpc handler for the beacon and history rpc handler's
### How was it fixed?

By removing the rwlock


ogenev (approved):
⛵

morph-dev (approved):

njgheorghita (approved):

PR #1224, opened 2024-03-21
refactor: extract history storage into separate file

morph-dev:
### What was wrong?

Current implementation of history storage is within `trin-history` crate.
In order to migrate it to the new storage (`id_indexed_v1`), we first have to move current logic into a separate module.

### How was it fixed?

Created `legacy_history` module and moved all related logic into it. This was done with very little code changes (mostly regarding initialization).

Plan for the followup PR is to write a code for migration from `legacy_history` to `id_indexed_v1` (which should be just table rename). After that PR we should be able to remove `legacy_history` module, and cleanup code a bit.

### To-Do

### DIFF
--- a/trin-storage/src/versioned/legacy_history/store.rs
+++ b/trin-storage/src/versioned/legacy_history/store.rs
@@ -68,7 +68,7 @@ impl VersionedContentStore for LegacyHistoryStore {
 }
 
 impl LegacyHistoryStore {
-    /// Public constructor for building a `HistoryStorage` object.
+    /// Public constructor for building a `LegacyHistoryStore` object.
     /// Checks whether a populated database already exists vs a fresh instance.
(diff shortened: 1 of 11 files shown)


trin-storage/src/versioned/legacy_history/store.rs
```diff
@@ -0,0 +1,843 @@
...
+    fn create(content_type: ContentType, config: Self::Config) -> Result<Self, ContentStoreError> {
+        if content_type != ContentType::History {
+            panic!("LegacyHistoryStore only supports History content type");
+        }
+        Self::new(config)
+    }
+}
+
+impl LegacyHistoryStore {
+    /// Public constructor for building a `HistoryStorage` object.
```
ogenev, line 71:
```suggestion
    /// Public constructor for building a `LegacyHistoryStore` object.
```

ogenev (approved):
⛵

KolbyML (approved):

PR #1225, opened 2024-03-21
[WIP] Rlp block bodies bug

njgheorghita:
### What was wrong?
Seems like we have a discrepancy between how we encode our typed transactions & receipts in the portal wire spec, and then how they're encoded in eth1 devp2p messages.

### To-Do


njgheorghita:
closing in favor of #1234

PR #1226, opened 2024-03-22
feat: add pagination to the id_indexed_v1 store

morph-dev:
### What was wrong?

The `id_indexed_v1` store didn't support pagination.
This is a requirement before transitioning history storage to this store as well.

### How was it fixed?

Implemented pagination and using it from state network.
I added tests and tested JSON api locally.

### To-Do


ogenev (approved):
⛵

KolbyML (approved):
:shipit:

PR #1229, opened 2024-03-27
feature: add legacy_history to id_indexed_v1 store migration

morph-dev:
### What was wrong?

We need custom logic in order to migrate History Storage to `id_indexed_v1` Store.

### How was it fixed?

Added custom logic for detecting if legacy history storage is used and migration to the new store.

### To-Do


ogenev (approved):
Looks good, could you rebase to make sure all tests pass?

PR #1233, opened 2024-03-31
fix: ci by removing Goerli which is no longer supported

### LINKED ISSUE #1230: Light Client still supports Goerli
The [light-client](https://github.com/ethereum/trin/tree/bd3a411f79fe7e1c13a40cd0eab71b86b0dc6179/light-client) still supports/uses goerli that is no longer producing blocks:

- https://blog.ethereum.org/2023/11/30/goerli-lts-update
- https://grafana.observability.ethpandaops.io/dashboard/snapshot/fwALK9PEe7CLEOO81lma5bT6k7jq7Gcs?orgId=0&from=1707927490905&to=1710519490905

This is causing some tests to fail: `config::checkpoints::test::test_fetch_latest_checkpoints`

To proper solution is to replace Goerli with Sepolia or Holesky (or both).

KolbyML:
### What was wrong?

FIxes #1230

Goerli is no longer supported and it was breaking our ci
### How was it fixed?
by removing it


ogenev (changes requested):
We also need to do some cleanup in `client_config.rs`.

ogenev (approved):

PR #1235, opened 2024-04-02
feat: migrate history to id_indexed_v1 store

morph-dev:
### What was wrong?

History storage is still using the legacy store.

### How was it fixed?

This PR migrates it to the new storage scheme.

Remaining work afterwards will be to remove legacy store and cleanup code where applicable.

### To-Do

### DIFF
--- a/trin-history/src/jsonrpc.rs
+++ b/trin-history/src/jsonrpc.rs
@@ -196,7 +196,7 @@ async fn paginate_local_content_keys(
     offset: u64,
     limit: u64,
 ) -> Result<Value, String> {
-    let response = match network.overlay.store.read().paginate(&offset, &limit)
+    let response = match network.overlay.store.read().paginate(offset, limit)
         {
             Ok(val) => Ok(json!(val)),
             Err(err) => Err(format!(
--- a/trin-history/src/storage.rs
+++ b/trin-history/src/storage.rs
@@ -1,25 +1,22 @@
 use ethportal_api::{
-    types::{distance::Distance, history::PaginateLocalContentInfo},
+    types::{distance::Distance, history::PaginateLocalContentInfo, portal_wire::ProtocolId},
     OverlayContentKey,
 };
 use trin_storage::{
     error::ContentStoreError,
-    versioned::{create_store, ContentType, LegacyHistoryStore},
-    ContentStore, PortalStorageConfig, ShouldWeStoreContent,
+    versioned::{create_store, ContentType, IdIndexedV1Store, IdIndexedV1StoreConfig},
+    ContentId, ContentStore, PortalStorageConfig, ShouldWeStoreContent,
 };
 
 /// Storage layer for the history network. Encapsulates history network specific data and logic.
 #[derive(Debug)]
 pub struct HistoryStorage {
-    store: LegacyHistoryStore,
+    store: IdIndexedV1Store,
 }
 
 impl ContentStore for HistoryStorage {
     fn get<K: OverlayContentKey>(&self, key: &K) -> Result<Option<Vec<u8>>, ContentStoreError> {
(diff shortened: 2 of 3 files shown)


KolbyML (approved):
:shipit:

ogenev (approved):
LGTM 👍

KolbyML:
I ran the database migration of 5 of my nodes. Nothing seems to be wrong, I did have to rebase the PR for the fix for blobs. Other then that I didn't see any issues. It took 50 minutes for the migration to run though.

morph-dev:
> It took 50 minutes for the migration to run though.

That's definitely way more than I expected. I'm wondering why it takes so long. How big are the databases, both in GB and how many entries?

I checked the migration code, and this is what it does (sadly, SQLite doesn't support renaming indices, otherwise this would be much simpler):

1. Rename table:
    ```sql
    ALTER TABLE history RENAME TO ii1_history
    ```
2. Deleting old indices:
    ```sql
    DROP INDEX history_distance_short_idx;
    DROP INDEX history_content_size_idx;
    ```
3. Updating usage stats:
    ```sql
    SELECT COUNT(*) as count, TOTAL(content_size) as used_capacity FROM ii1_history
    ```
4. Creating new indices:
    ```sql
    CREATE INDEX IF NOT EXISTS ii1_history_distance_short_idx ON ii1_history (distance_short);
    CREATE INDEX IF NOT EXISTS ii1_history_content_size_idx ON ii1_history (content_size);
    ```



I think renaming the table and dropping indices should be fast (maybe few seconds).
Creating new indices might be slow, but I believe that 50 minutes is too much. You were playing with creating and deleting indices in the past in order to improve performance. How long would it usually take?

So the issue might be calculating usage stats. Now that I wrote everything down, that one is definitely slower then in needs to be because it runs without any content_size index being present. @KolbyML , would you be able to test migration again if I modify the migration code? I can also add some debug statements and we can see how long each operation takes.

morph-dev:
I rebased this PR.
I also created a commit (not part of this pr, yet) that hopefully improves migration: https://github.com/morph-dev/trin/commit/176f64a13cbe51bfce6a26c77527c5116c28befd

I will try to test it on my machine, but it will be a small db. I hope at least to be able to figure out expected outcome.

KolbyML:
The nodes are only configured to hold like 100mb or something of logs so they got deleted before I could see them
image
image
image

It looks like the time is now 7-8 minutes to migrate instead of 50 minutes on a 10gb database with ~1 million keys.

morph-dev:
> It looks like the time is now 7-8 minutes to migrate instead of 50 minutes on a 10gb database with ~1 million keys.

I would say that's reasonable. If everyone agrees, I would proceed with merging this PR.

@KolbyML Do you have the logs so we can see how long each part takes? It would be useful to know for potential future migration

KolbyML:
> @KolbyML Do you have the logs so we can see how long each part takes? It would be useful to know for potential future migration

>> The nodes are only configured to hold like 100mb or something of logs so they got deleted before I could see them

KolbyML:
> I would say that's reasonable. If everyone agrees, I would proceed with merging this PR.

I think it is fine, I just think we need status info logs which are like

info!("Trin Database Migration from X to Y in progress: Z% done") or something along those lines just so people know something is happening

morph-dev:
> I think it is fine, I just think we need status info logs which are like
> 
> info!("Trin Database Migration from X to Y in progress: Z% done") or something along those lines just so people know something is happening

We can't have "%done part", because we only run simple SQL commands. I changed "Migration started" and "Migration finished" logs to info level. I would say that's enough, considering that this will happen only once.

KolbyML:
I would just make it up if we have 6 steps 1/6= how much percent is done

PR #1236, opened 2024-04-03
bugfix: broken blob transaction ssz decoding

njgheorghita:
### What was wrong?
Bug in blob transaction ssz decoding was preventing glados from hitting 100% on block bodies.

### How was it fixed?
Fixed broken decoding and added test case.

### To-Do


KolbyML (approved):
Nice catch :shipit:

PR #1237, opened 2024-04-04
fix: era1 typed-(transaction/receipt) encode/decode (open-ethereum/rlp version)

KolbyML:
| open-ethereum/rlp version  | reth's alloy/rlp version  |
|---|---|
| image Runtime 220.35s : peak ram usage 3.9GB | image Runtime 60.52s : peak ram usage 3.2GB |

open-ethereum/rlp is 220.35s/60.52s = 3.64 times lower then reth's alloy/rlp
It also uses much more ram.

I don't think it makes sense to use this when `reth's alloy/rlp version` is already implemented and switching to alloy provides many more advantages.

The work is done already, it just doesn't make sense to keep using open-ethereum rlp. It being dog slow just displays this even more.

`reth's alloy/rlp` is based off fast-rlp developed by the now defunct Rust Ethereum Execution Client named Akula. It is indeed fast.

### DIFF
--- a/ethportal-api/src/types/execution/block_body.rs
+++ b/ethportal-api/src/types/execution/block_body.rs
@@ -199,14 +199,21 @@ pub struct BlockBodyLegacy {
 impl Encodable for BlockBodyLegacy {
     fn rlp_append(&self, s: &mut RlpStream) {
         s.begin_list(2);
-        s.append_list(&self.txs);
+        s.begin_list(self.txs.len());
+        for tx in &self.txs {
+            tx.encode_with_envelope(s, true);
+        }
         s.append_list(&self.uncles);
     }
 }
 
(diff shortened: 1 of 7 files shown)


pipermerriam (commented):
Are there test vectors for this stuff that are being run in CI?  If not, it seems that we should have these somewhere in the client testing suite to ensure that clients are able to decode and serialize all of these formats.  This is the kind of thing that makes for "good first issue" type things if you can be specific about what data should go into the suite.

ogenev:
I understand that `alloy-rlp` is a more performant version than what we are currently using. 

I would ask the same question again here:

Could we migrate to `alloy-rlp` without migrating to `alloy-primitives` until sgp updates all core libraries that we are depending on?

PR #1242, opened 2024-04-07
fix: use correct column name in "on update" usage stats trigger

morph-dev:
### What was wrong?

While looking into issue regarding #1235, I discovered the bug regarding usage stats "on update" trigger.
There was a test that was supposed to test this functionality but it was passing by chance.

### How was it fixed?

Updated the test (verified that it fails), fixed the code and verified that test passes.

The more correct approach would be to drop the trigger if it exists, and create new one with different name.
I think that this is overkill as this code is not yet used in public. The only exception are 5 machines that @KolbyML 's used while testing the #1235 (this [comment](https://github.com/ethereum/trin/pull/1235#issuecomment-2040483220) for context), which can he can fix by manually executing following query:

```sql
DROP TRIGGER ii1_history_on_update_update_usage_stats_trigger;
```

This PR should be merged before: #1235.

### To-Do

### DIFF
--- a/trin-storage/src/versioned/sql.rs
+++ b/trin-storage/src/versioned/sql.rs
@@ -67,7 +67,7 @@ pub fn create_usage_stats_triggers(
         FOR EACH ROW
         BEGIN
             UPDATE usage_stats
-            SET size = size - OLD.size + NEW.size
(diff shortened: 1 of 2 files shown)


ogenev (approved):

KolbyML (approved):
:shipit:

PR #1243, opened 2024-04-08
feat: add support for Deneb consensus types and fix beacon sync

ogenev:
### What was wrong?

1. We had a recently introduced bug in beacon sync which prevented a node from syncing with the tip of the chain.
2. To fully support post-merge Era files we need to support the Deneb fork changes introduced in consensus types.

### How was it fixed?
1. Fix the beacon sync bug
2. Update beacon block, execution payload header, and beacon state to support Deneb changes.
3. Update light client types to Deneb.

### To-Do


ethportal-api/src/types/consensus/kzg_commitment.rs
```diff
@@ -0,0 +1,138 @@
...
+        Self::from_str(&string).map_err(serde::de::Error::custom)
+    }
+}
+
+impl FromStr for KzgCommitment {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some(stripped) = s.strip_prefix("0x") {
+            let bytes = hex::decode(stripped).map_err(|e| e.to_string())?;
```
njgheorghita, line 95: why not use `ethportal_api::utils::bytes::hex_decode`?

ethportal-api/src/types/consensus/kzg_commitment.rs
```diff
@@ -0,0 +1,138 @@
...
+            }
+        } else {
+            Err("must start with 0x".to_string())
+        }
+    }
+}
+
+impl Debug for KzgCommitment {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", serde_utils::hex::encode(self.0))
```
njgheorghita, line 115: same with `ethportal_api::utils::bytes::hex_encode()`

ethportal-api/src/types/consensus/kzg_commitment.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+#[test]
```
njgheorghita, line 119: I think we should continue the convention of using `mod tests {}` as a wrapper around the tests, which is how we typically do it across the code base.

ethportal-api/src/types/consensus/kzg_commitment.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+        write!(f, "…")?;
+        for i in &self.0[BYTES_PER_COMMITMENT - 2..BYTES_PER_COMMITMENT] {
+            write!(f, "{i:02x}")?;
+        }
+        Ok(())
+    }
+}
+
+impl TreeHash for KzgCommitment {
```
njgheorghita, line 53: It seems worth adding tests specifically for `tree_hash_root()` if we're custom implementing this trait

light-client/src/consensus/consensus_client.rs
```diff
@@ -384,6 +389,9 @@ impl<R: ConsensusRpc> ConsensusLightClient<R> {
 
         let should_apply_update = {
             let has_majority = committee_bits * 3 >= 512 * 2;
+            if !has_majority {
```
njgheorghita, line 392: is this something you meant to leave in?

njgheorghita (approved):

KolbyML (approved):
:shipit:

light-client/src/consensus/consensus_client.rs
ogenev: yes, updated the log message to be more clear.

PR #1245, opened 2024-04-08
feat(bridge): cli flag for random era1 file with floor

njgheorghita:
### What was wrong?
launches era1 bridge mode with random single, but only for epochs greater than the provided floor

### How was it fixed?
added cli flag

### To-Do


KolbyML (approved):
:shipit:

PR #1247, opened 2024-04-09
feat(bridge): make gossip limit a cli option

KolbyML:
### What was wrong?
we have headroom on our era1 bridges which we can use to gossip more blocks simultaneously

### How was it fixed?
by adding a cli option to configure the constant

### DIFF
--- a/ethportal-peertest/src/scenarios/bridge.rs
+++ b/ethportal-peertest/src/scenarios/bridge.rs
@@ -7,6 +7,7 @@ use portal_bridge::{
     api::{consensus::ConsensusApi, execution::ExecutionApi},
     bridge::{beacon::BeaconBridge, history::HistoryBridge},
     cli::Provider,
+    constants::DEFAULT_GOSSIP_LIMIT,
     types::mode::BridgeMode,
 };
 use serde_json::Value;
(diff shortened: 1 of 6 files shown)


ogenev (approved):
⛵

njgheorghita (approved):
Nice. Just so I'm up to date. I remember we had some kind of governer on http requests to providers, but it looks like that was removed? So, moving forward, this cli arg can be considered the de facto "limit" (aka the only factor to consider) for the quantity of ongoing concurrent requests (an all bridges: history, era1, beacon, etc) ?

njgheorghita:
fyi, i've been running my personal bridge on era1 gossip with a limit of 256 (:rocket:) but there appear to be quite a significant number of errors happening when running with such a high limit. anyways, just a note that moar != better gossip. using stats though should help to find the ideal combination of concurrency limit & gossip success rate

KolbyML:
> fyi, i've been running my personal bridge on era1 gossip with a limit of 256 🚀 but there appear to be quite a significant number of errors happening when running with such a high limit. anyways, just a note that moar != better gossip. using stats though should help to find the ideal combination of concurrency limit & gossip success rate

Sounds like we need to eventually find 1 change that improves performance by 90%

PR #1248, opened 2024-04-09
feat(bridge): gossip latest block-body/receipt at the same time

KolbyML:
### What was wrong?
For latest content getting on the network as fast as possible is critical.

Currently we use a sequential model where we gossip: header -> block body -> receipt. But this means gossiping block body's blocks receipts from getting gossiped, which could be (not always) the difference in glados reporting a success or not
### How was it fixed?
Instead of gossiping receipts after block bodies we can gossip them at the same time

**old model**
header -> block body -> receipt
**new model**
header -> block body
||||||||||||||-> receipt

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -36,6 +36,7 @@ const LATEST_BLOCK_POLL_RATE: u64 = 5; // seconds
 pub const EPOCH_SIZE: u64 = EPOCH_SIZE_USIZE as u64;
 pub const GOSSIP_LIMIT: usize = 32;
 pub const SERVE_BLOCK_TIMEOUT: Duration = Duration::from_secs(120);
+pub const SERVE_CONTENT_TIMEOUT: Duration = Duration::from_secs(30);
 
 pub struct HistoryBridge {
     pub mode: BridgeMode,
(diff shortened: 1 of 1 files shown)


ogenev (approved):
🛥️

portal-bridge/src/bridge/history.rs
```diff
@@ -277,30 +278,54 @@ impl HistoryBridge {
...
+                    receipt_block_stats,
+                    metrics,
+                ).in_current_span(),
+            )
+            .await {
+                Ok(result) => match result {
+                    Ok(_) => (),
+                    Err(msg) => warn!("Error gossiping block receipt #{height:?}: {msg:?}"),
+                },
+                Err(_) => error!("construct_and_gossip_receipt() timed out on height {height}: this is an indication a bug is present")
```
njgheorghita, line 323: Not a blocker for this pr, but this error message is very frustrating. Usually, I'll see it on my personal node after it's been running for a while. My understanding is that there's really no way to extract a relevant error message here, since the request could be blocked at any one of the many given points throughout the gossip process. But, it would be really really really cool if we could figure out some way to extract some kind of information to help understand why the timeout happened...

njgheorghita (approved):

portal-bridge/src/bridge/history.rs
KolbyML: Well we already know why a timeout would happen
- http client vs pandops deadlock bug
- unable to find a node to gossip 2

PR #1249, opened 2024-04-10
fix: add sleep to allow for header saturation in era1 gossip

njgheorghita:
### What was wrong?
In our `history` gossip, we have a 10 second sleep after gossiping the header to allow for network saturation, since this header is necessary to validate the following receipts & body values. We are missing this sleep in `era1` gossip. Now, the effectiveness of this sleep & it's value are up for debate, but it seems to me like it's worthwhile to match this behavior inside `era1` gossip, at least until it's proven unnecessary.

### How was it fixed?
added sleep to `era1` gossip

### To-Do


KolbyML (approved):
:shipit:

ogenev (approved):

PR #1250, opened 2024-04-11
feat: limit beacon sync attempts

ogenev:
### What was wrong?
Beacon was trying to sync forever in an infinite loop.

### How was it fixed?
Add a counter and limit beacon sync attempts to 10 with 10-sec intervals.

### To-Do

### DIFF
--- a/trin-beacon/src/network.rs
+++ b/trin-beacon/src/network.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 
 use parking_lot::RwLock as PLRwLock;
 use tokio::sync::RwLock;
+use tracing::error;
 use utp_rs::socket::UtpSocket;
 
 use crate::{storage::BeaconStorage, sync::BeaconSync, validation::BeaconValidator};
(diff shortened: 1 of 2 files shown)


trin-beacon/src/sync.rs
```diff
@@ -39,15 +39,25 @@ impl BeaconSync {
...
+        // Try sync the client 10 times with 10 seconds interval. Return error if failed after 10
+        // attempts
+        let mut counter = 10;
+        while counter > 0 {
             info!(trusted_block_root = %trusted_block_root, "Starting syncing portal light client ...");
             match client.start().await {
                 Ok(_) => break,
                 Err(e) => {
-                    error!("Error syncing portal light client: {e} Retrying...");
+                    error!("Error syncing portal light client: {e} Retrying in 10 secs...");
```
KolbyML, line 50:
```suggestion
                    error!("Error syncing portal light client attempt {}: {e} Retrying in 10 secs...", 11 - counter);
```

Just a nit, but it would be to see which attempt it failed on, as we already hold the count, and it would be easiest then counting logs.

KolbyML (approved):
:shipit:

trin-beacon/src/network.rs
```diff
@@ -56,14 +57,16 @@ impl BeaconNetwork {
...
                     .start(
                         portal_config
                             .trusted_block_root
                             .expect("Trusted block root should be available"),
                     )
-                    .await
-                    .expect("Beacon sync failed to start");
+                    .await;
+                if let Err(e) = beacon_sync {
+                    error!("Failed to start beacon sync: {:?}", e);
```
njgheorghita, line 68: *nitpick* interpolate `e` into the string... though clippy lint should be picking up on this?

njgheorghita (approved):

PR #1251, opened 2024-04-11
fix: latest bridge waiting for 2 new blocks before gossiping

KolbyML:
### What was wrong?
image
The latest bridge was waiting for 2 new blocks before it started gossiping them. If you notice in the picture groups of 2 form due to this bug
### How was it fixed?
I choose to just use a loop because it was easier for me to read and understand, a Range was a little overkill so this simplification helped me fix the issue.

I tested it by copy pasting it into a test and replacing any function call with logs, then fixing it till I got the expected log output I was looking for.

It no longer has a delay in gossiping latest blocks.

### Before this is how our current code works
image

### After
image

I also tested in interops of 2,3 new blocks per loop {} and it is working as expected

### DIFF
--- a/portal-bridge/src/bridge/history.rs
+++ b/portal-bridge/src/bridge/history.rs
@@ -140,6 +140,12 @@ impl HistoryBridge {
                 last_seen_block_counter = 0;
                 info!("Discovered new blocks to gossip: {block_index}-{latest_block}");
             }
+            // `..=` is the inclusive range operator, it includes the right side of the range.
+            // Example: 1..=3 is equivalent to 1, 2, 3
+            // Where as: 1..3 is equivalent to 1, 2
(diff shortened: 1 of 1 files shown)


portal-bridge/src/bridge/history.rs
```diff
@@ -137,25 +136,21 @@ impl HistoryBridge {
...
-                        self.portal_clients.clone(),
-                        self.execution_api.clone(),
-                        None,
-                        self.metrics.clone(),
-                    );
-                }
-                block_index = gossip_range.end;
+                info!("Discovered new blocks to gossip: {block_index}-{latest_block}");
+            }
+            for height in block_index..=latest_block {
```
njgheorghita, line 143: This syntax is just a bit unfamiliar to me. Maybe just add a comment explaining what's going on?

njgheorghita (approved):

PR #1253, opened 2024-04-12
refactor: rename master accumulator and accumulator proof

ogenev:
### What was wrong?
We will introduce two more accumulators for verifying pre-capella and post-capella headers and using "master" for the pre-merge accumulator is confusing and doesn't make sense anymore.

### How was it fixed?
- Rename `master accumulator ` to `pre-merge accumulator`.
- Rename `AccumulatorProof` to `PreMergeAccumulatorProof`.

### To-Do

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -10,7 +10,7 @@ use url::Url;
 
 use crate::types::bootnodes::Bootnodes;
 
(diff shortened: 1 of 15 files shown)


njgheorghita (commented):
Any thoughts on updating the specs also/first? If it's confusing enough to change in this repo, then it's probably worthwhile updating the specs to reflect this. Ultimately, with this kind of type that's explicitly defined in the specs, I'd like to maintain 1:1 consistency with the definition

PR #1256, opened 2024-04-15
feat(bridge): add found to gossip reports

njgheorghita:
### What was wrong?
Update gossip report logs to include a `found` field, indicating whether or not the gossip attempt aborted b/c it located the content on the network.

### How was it fixed?
added `found` field to gossip report

### To-Do


KolbyML (approved):
:shipit:

PR #1257, opened 2024-04-16
fix: revert history storage to use LegacyHistoryStore

morph-dev:
### What was wrong?

Pruning 5-10% of the database can take too much time that, which also blocks other tasks.

### How was it fixed?

Revert code to use `LegacyHistoryStore` again (partial revert of #1235 ).

### To-Do


KolbyML (approved):
:shipit:

PR #1258, opened 2024-04-16
fix: check for pruning more frequently

morph-dev:
### What was wrong?

The `IdIndexedV1Store` store was checking for pruning only once it estimated the it reached full capacity.
In practice, this could lead to using way more storage than desired.

### How was it fixed?

I added `PRUNING_CHECKING_FREQUENCY` that forces us to check for pruning more frequently.
Checking for pruning is no longer as expensive as it was originally (because we are now using triggers to update usage stats), so this shouldn't have big performance impact.

I also removed one test that was hard to test with new logic. It's worth pointing out that most parts of the tests were already covered by other tests.

### To-Do


KolbyML (approved):
:shipit: I am happy to see where this goes

PR #1260, opened 2024-04-17
feat(bridge): add fallback providers

njgheorghita:
### What was wrong?
Added the ability to set a fallback provider, in case a request to the primary provider fails.

### How was it fixed?
- Removed `Retry` middleware
- Added fallback & primary providers for both el & cl bridge apis
- Removed ability to set provider urls via env variable, since these can now just be set via cli flag

### To-Do


KolbyML (approved):
:shipit:

PR #1264, opened 2024-04-22
feat(ethportal-api): add and refactor post-merge header proofs

ogenev:
### What was wrong?
We need types to contain post-merge execution header proofs

### How was it fixed?
- move all header proofs to a new `header_proofs.rs` file inside the `execution` folder.
- add and update post-merge header proof types
- rename the old `header_proofs.rs` file in the consensus folder to `historical_summaries.rs` because now It contains only those types.

Post-merge proof types are based  on the following portal specs PRs: 

- https://github.com/ethereum/portal-network-specs/pull/287
- https://github.com/ethereum/portal-network-specs/pull/291

### To-Do

### DIFF
--- a/ethportal-api/src/types/consensus/beacon_state.rs
+++ b/ethportal-api/src/types/consensus/beacon_state.rs
@@ -5,7 +5,7 @@ use crate::consensus::{
     },
     fork::ForkName,
     header::BeaconBlockHeader,
-    header_proof::HistoricalSummary,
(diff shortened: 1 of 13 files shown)


njgheorghita (approved):
*nitpick* I'd probably prefer `execution/header_with_proof.rs` over `execution/header_proofs.rs`. It's slightly longer, but the latter makes it sound like the file only contains the proofs rather than the `HeaderWithProof` type itself along with the proofs

PR #1266, opened 2024-04-23
fix(bridge): add sleep before request retry to fallback

njgheorghita:
### What was wrong?
When we removed the `Retry` middleware, we lost a built-in sleep mechanism. Some of the current bridge errors seems to be a result of "DOS protection". Re-introducing a manual sleep "should" help prevent this error from happening.

### To-Do


KolbyML (approved):
:shipit: 15 seconds seems a little too long but if it proves an issue we can always shorten it in the future

PR #1267, opened 2024-04-23
feat(portalnet): add rfn lookup to gossip

njgheorghita:
### To-Do

### DIFF
--- a/portalnet/src/gossip.rs
+++ b/portalnet/src/gossip.rs
@@ -6,7 +6,7 @@ use discv5::{
 };
 use futures::channel::oneshot;
 use parking_lot::RwLock;
-use rand::seq::IteratorRandom;
+use rand::{seq::SliceRandom, thread_rng};
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc;
 use tracing::{debug, error, warn};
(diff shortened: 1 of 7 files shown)


portalnet/src/gossip.rs
```diff
@@ -40,38 +40,51 @@ pub struct GossipResult {
...
-                .filter(|node| node.status.is_connected())
-                .collect::<Vec<&kbucket::Node<NodeId, Node>>>()
-        })
-        .collect();
-
-    if all_nodes.is_empty() {
-        // If there are no nodes whatsoever in the routing table the gossip cannot proceed.
-        warn!("No nodes in routing table, gossip cannot proceed.");
-        return 0;
-    }
```
KolbyML, line 67: why was this removed?

KolbyML, line 67: Shouldn't this still be here?

KolbyML:
I think the same result would have happened if you ran master on your home node running latest. So I think both should cases should have been tested as a control. 

Debugging and fixing why we are having provider issues is the highest issue then this. I do think this is a smarter way of gossipping.

Without the logs or seeing if there was provider errors I don't think I can comment anything of value on the results you told me of the tested you reported to me earlier today

njgheorghita:
closing since this feature is no longer a priority, now that we've added direct offer to the state bridge

PR #1270, opened 2024-04-24
refactor(bridge): add default value for cli flags

njgheorghita:
### What was wrong?
This is a bit of a **nitpick**, but imo requiring these cli flags is unnecessary, and while running the bridge locally, it clutters up the command line. These seem like reasonable values for local testing, and ofc in production / edge cases you can still use the cli flag to specify a different value.

### How was it fixed?
Added default values for two cli flags in the bridge

### To-Do


KolbyML (approved):
:shipit:

PR #1273, opened 2024-04-25
bridge(fix): reintroduce retry middleware

njgheorghita:
### What was wrong?
I'm having a hard time explaining this, but our bridges are experiencing a persistent error.
```
2024-04-25T13:39:05.565172Z  WARN portal_bridge::api::execution: Failed to send request to primary provider, retrying with fallback provider: Unable to request execution payload from provider: Broken pipe (os error 32)
```

However, it's mostly on one "latest" bridge, periodically on another "latest" bridge, never on the third "latest" bridge. I've been unable to reproduce this locally, making this problem difficult to debug.

What's strange is that our bridges had never experienced this bug until #1260, when we removed the `Retry` middleware and introduced a `fallback_provider` instead. 

Yesterday, I deployed this pr to the network bridges, and didn't see any of "Broken pipe" errors for at least a period of 5 hours. I can't explain why, but given that this problem is fairly difficult to debug, I'm in favor of reintroducing the `Retry` middleware.

The main reason we removed the `Retry` middleware was due to insufficient logging (but imo, this is something we can fix with a separate [logger](https://docs.rs/surf/latest/surf/middleware/index.html) middleware. 

The `fallback_provider` we introduced will still be used if the initial request to the primary provider (after 3 retries) fails.

### How was it fixed?
Reintroduced the `Retry` middleware.

### To-Do

### DIFF
--- a/portal-bridge/src/cli.rs
+++ b/portal-bridge/src/cli.rs
@@ -2,9 +2,15 @@ use std::{env, net::SocketAddr, path::PathBuf, str::FromStr};
 
 use alloy_primitives::B256;
 use clap::{Parser, Subcommand};
-use surf::{Client, Config};
-use tokio::process::Child;
-use tracing::error;
+use surf::{
+    middleware::{Middleware, Next},
+    Body, Client, Config, Request, Response,
+};
+use tokio::{
+    process::Child,
+    time::{sleep, Duration},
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

Issue #1228, opened 2024-03-23
Have the bridge use HD ephemeral identities


pipermerriam:
### What is wrong

IIUC the bridge generates a new ephemeral key each time it boots up leaving us with a lot of *single use* entries in our census data.

### How can it be fixed

https://github.com/ethereum/devp2p/issues/254

I propose we default the bridge to operate using ephemeral HD identities.  THere would be a single persistent "parent key".  Each time the bridge runs it would randomly generate a new "path" in order to generate a child key that it would use as it's private key for that run.   The ENR record would be updated to include the "parent key" and "path" so that glados would be able to link these ephemeral identities together.

njgheorghita:
Is there another motivation for implementing this? Aka, testing out this scheme for the case where a single user with 100gb available storage spins up 10 x 10gb nodes derived from a parent key... 

There have been some concerns in the past about letting the ENR grow too large, even adding a single character for the client type wasn't quickly adopted. 

The simplest fix in this scenario is to simply use a consistent private key for the bridge, which we already support, it's just not enabled in our deployment scripts.

pipermerriam:
Yes, this would also be part of supporting https://github.com/ethereum/portal-network-specs/issues/283

PR #1274, opened 2024-04-26
refactor(bridge): remove multiple client functionality

njgheorghita:
### What was wrong?
The bridge was initially architected to support gossip via multiple clients at once. This introduced some edge cases and unwieldy code which resulted in a feature that's not actually used / nor properly supported & tested. 

Taking into account that this general structure is not suitable for state network gossip (since a bridge should only gossip content that it's "near" to), I think it's a good idea to go ahead and remove this feature. 

Furthermore, it's likely that we'll introduce an alternative approach to "covering the keyspace" with multiple clients. via #1228 

### How was it fixed?
Removed support for `--node-count` argument

### To-Do

- [ ] Update cluster scripts to accommodate this change

### DIFF
--- a/ethportal-peertest/src/scenarios/bridge.rs
+++ b/ethportal-peertest/src/scenarios/bridge.rs
@@ -10,16 +10,14 @@ use portal_bridge::{
     types::mode::BridgeMode,
 };
 use serde_json::Value;
-use std::sync::Arc;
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
:shipit:

ogenev (approved):
⛵

PR #1276, opened 2024-04-29
refactor: disable loading pre-merge accumulator from custom path

ogenev:
### What was wrong?
There is no real use case for loading a pre-merge accumulator from a user-provided path, because we embed the accumulator in Trin binary.

### How was it fixed?
Remove the `pre_merge_acc_path` argument from Trin CLI.

### To-Do

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,12 +77,8 @@ pub async fn run_trin(
     )?;
 
     // Initialize validation oracle
-    let pre_merge_accumulator =
-        PreMergeAccumulator::try_from_file(trin_config.pre_merge_acc_path.clone())?;
-    info!(
-        "Loaded pre-merge accumulator from: {:?}",
-        trin_config.pre_merge_acc_path
-    );
+    let pre_merge_accumulator = PreMergeAccumulator::default();
+    info!("Loaded pre-merge accumulator.",);
     let header_oracle = HeaderOracle::new(pre_merge_accumulator);
     let header_oracle = Arc::new(RwLock::new(header_oracle));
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -77,12 +77,8 @@ pub async fn run_trin(
...
 
     // Initialize validation oracle
-    let pre_merge_accumulator =
-        PreMergeAccumulator::try_from_file(trin_config.pre_merge_acc_path.clone())?;
-    info!(
-        "Loaded pre-merge accumulator from: {:?}",
-        trin_config.pre_merge_acc_path
-    );
+    let pre_merge_accumulator = PreMergeAccumulator::default();
+    info!("Loaded pre-merge accumulator.",);
```
njgheorghita, line 81: It seems worthwhile putting the accumulator hash in this message, then users can confirm the hash matches against an external / 3rd party source if they want.

pipermerriam:
I'm onboard with you guys removing this, however, I'll point out that by baking the pre-merge accumulators into the binary **and** not providing users with an alternative way to load there own, we are effectively making ourselves the authority to say "this is the only canonical chain"...

Is that a responsibility that you guys want?

njgheorghita:
Not necessarily a responsibility we want, more like a responsibility we'll temporarily assume until users start questioning our "source of truth" and want the option to use their own. (which tbf, even if they generate their own, should result in the exact same hash, so I'm not too sure we'll run into this problem down the road)

ogenev:
I think we can easily revert this if/when people start to complain.

src/lib.rs
ogenev: Forgot to address this before merging, will add it to my next PR.

PR #1278, opened 2024-04-29
fix(bridge): broken compilation issue with bridge

njgheorghita:
### What was wrong?
Not sure how this snuck through, but `master` was broken due to conflicting pr changes

### To-Do


njgheorghita:
rage merging to fix master, but @KolbyML just fyi that it looks like the results from state gossip were not being accurately reported / logged

PR #1279, opened 2024-04-29
fix(bridge): use client object per request in bridge

njgheorghita:
### What was wrong?
This is a bit of an experimental pr. The "latest" bridge issues we're experiencing only occur after a significant time of bridge operation. It's possible that there's an error with our request construction / response handling for long-running `Client` objects, handling many requests. Imo, it's worth trying out this change to see if constructing a `Client` object for each request at least resolves the issue, then we can see about changing this back / fixing the long-running `Client` object if we feel like it's worthwhile.

### How was it fixed?
- replaced a single-use `Client` object for a per-request `Client` object

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/bridge.rs
+++ b/ethportal-peertest/src/scenarios/bridge.rs
@@ -10,7 +10,6 @@ use portal_bridge::{
     types::mode::BridgeMode,
 };
 use serde_json::Value;
-use surf::{Client, Config};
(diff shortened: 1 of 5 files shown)


KolbyML (approved):
:shipit: https://github.com/ethereum/trin/pull/1260 in this PR we did remove this yeah. So reverting that change does seem enticing. Maybe it caused us to run out of file descriptors or something

PR #1281, opened 2024-04-29
feat(bridge): throw error if invalid executable path was provided

KolbyML:
### What was wrong?
When I was working on the state bridge. An incorrect executable-path took a lot of my time, cause their was no error, but the bridge kept trying to gossip content anyways
### How was it fixed?

throw an error if the executable path doesn't lead to a file. This is a naive solution which doesn't handle all cases, but it handles most which would save people time if they entered the incorrect path.

### DIFF
--- a/portal-bridge/src/cli.rs
+++ b/portal-bridge/src/cli.rs
@@ -252,6 +252,11 @@ impl FromStr for ClientType {
 
 impl ClientType {
     pub fn build_handle(&self, bridge_config: &BridgeConfig) -> anyhow::Result<Child> {
+        if !bridge_config.executable_path.is_file() {
+            return Err(anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


portal-bridge/src/cli.rs
```diff
@@ -252,6 +252,11 @@ impl FromStr for ClientType {
 
 impl ClientType {
     pub fn build_handle(&self, bridge_config: &BridgeConfig) -> anyhow::Result<Child> {
+        if !bridge_config.executable_path.is_file() {
+            return Err(anyhow::anyhow!(
+                "Invalid executable path: exectuable doesn't exist"
```
njgheorghita, line 257: *nitpick* import `anyhow!` at top of file. Also, I think it'd be useful to add the path that failed the lookup to the error message

njgheorghita (approved):

PR #1284, opened 2024-04-30
feat(trin-validation): add validation logic for post-merge/pre-Capella headers

ogenev:
### What was wrong?
No logic was implemented for validating post-merge/pre-Capella execution headers.

### How was it fixed?
- Implement the chain of proofs validation using the historical roots accumulator.
- move the build of the pre-merge accumulator inside `HeaderValidator`.

### To-Do

### DIFF
--- a/trin-validation/src/header_validator.rs
+++ b/trin-validation/src/header_validator.rs
@@ -1,12 +1,15 @@
 use crate::{
     accumulator::PreMergeAccumulator,
-    constants::{EPOCH_SIZE, MERGE_BLOCK_NUMBER},
+    constants::{EPOCH_SIZE, MERGE_BLOCK_NUMBER, SHANGHAI_BLOCK_NUMBER},
     historical_roots_acc::HistoricalRootsAccumulator,
     merkle::proof::verify_merkle_proof,
 };
+use alloy_primitives::B256;
 use anyhow::anyhow;
 use ethportal_api::{
-    types::execution::header_with_proof::{BlockHeaderProof, HeaderWithProof},
+    types::execution::header_with_proof::{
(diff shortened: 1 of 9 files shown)


pipermerriam:
This seems like a thing that we should probably at least track in our hive test suite.  At some point here we want to have both positive and negative tests for all forks that verify that clients will accept valid headers and also that they will reject invalid headers so that we can see that all clients are properly validating each piece of chain data that they store.

ogenev:
> This seems like a thing that we should probably at least track in our hive test suite. At some point here we want to have both positive and negative tests for all forks that verify that clients will accept valid headers and also that they will reject invalid headers so that we can see that all clients are properly validating each piece of chain data that they store.

Yeah, we should add tests for this in hive.

trin-validation/src/header_validator.rs
```diff
@@ -62,8 +67,82 @@ impl HeaderValidator {
...
+            BlockHeaderProof::HistoricalRootsBlockProof(proof) => self
+                .verify_post_merge_pre_capella_header(hwp.header.number, hwp.header.hash(), proof),
+            BlockHeaderProof::HistoricalSummariesBlockProof(_) => {
+                if hwp.header.number <= SHANGHAI_BLOCK_NUMBER {
+                    return Err(anyhow!(
+                        "Invalid HistoricalSummariesBlockProof found for pre-Shanghai header."
+                    ));
+                }
+                // TODO: Validation for post-Capella headers is not implemented
+                Ok(())
```
KolbyML, line 79:
```suggestion
               unimplemented!("Unsupported proof type found."),
```

njgheorghita (approved):

PR #1285, opened 2024-04-30
feat(bridge): add four fours hunter mode

njgheorghita:
### What was wrong?
Added a four fours mode that accepts a sample size & threshold, will iter through all the epochs until it finds an epoch that's insufficiently available on the network according to the provided levels. Then it will gossip this epoch. For each individual piece of content within the epoch, it will perform an RFC lookup and will only gossip that piece of content if the RFC lookup fails.

### How was it fixed?
added "hunter" mode to 4444s bridge

### To-Do


KolbyML (approved):
:shipit:

PR #1288, opened 2024-05-07
refactor: update pruning strategy

morph-dev:
### What was wrong?

The pruning strategy of `IdIndexedV1Store` (deleting ~5% of db) was taking too long on large databases (example: 40sec on 10GB db).

### How was it fixed?

After trying several approaches and measuring performance, I refactored the pruning logic:

1. Don't use triggers and update usage stats manually
    - Made **delete** queries return `content_size` to save few sql queries
2. Because of 1. no need for estimating how many entries to insert before pruning
    - This also allowed removing `pruning_threshold` from the config
4. Prune up to 5%, but no more than 100 items at the time

I run this version on one of Kolby's machines and metrics look ok to me.

### For future PR-s:

- Use `IdIndexedV1Store` for history network
- Delete `LegacyHistoryStore`
- Delete triggers (might be useful in the future, but don't like to have unused code seating around)

### To-Do

### DIFF
--- a/trin-storage/src/versioned/id_indexed_v1/config.rs
+++ b/trin-storage/src/versioned/id_indexed_v1/config.rs
@@ -11,10 +11,7 @@ use crate::{
 };
 
 /// The fraction of the storage capacity that we should aim for when pruning.
-const TARGET_CAPACITY_FRACTION: f64 = 0.9;
-
-/// The fraction of the storage capacity that we need to pass in order to start pruning.
-const PRUNING_CAPACITY_THRESHOLD_FRACTION: f64 = 0.95;
+const TARGET_CAPACITY_FRACTION: f64 = 0.95;
 
 /// The config for the IdIndexedV1Store
 #[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


trin-storage/src/versioned/id_indexed_v1/config.rs
```diff
@@ -11,10 +11,7 @@ use crate::{
 };
 
 /// The fraction of the storage capacity that we should aim for when pruning.
-const TARGET_CAPACITY_FRACTION: f64 = 0.9;
-
-/// The fraction of the storage capacity that we need to pass in order to start pruning.
-const PRUNING_CAPACITY_THRESHOLD_FRACTION: f64 = 0.95;
+const TARGET_CAPACITY_FRACTION: f64 = 0.95;
```
njgheorghita, line 14: Maybe add a note here explaining why it's set to `95%` & what side effects to expect if someone wants to change it in the future

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -10,16 +10,16 @@ use crate::{
...
+        VersionedContentStore,
     },
     ContentId,
 };
 
-/// How frequently to check for pruning. Higher number means we will check more frequently while
-/// lower number (minimum 1) means we will check less frequently.
-const PRUNING_CHECKING_FREQUENCY: u64 = 100;
+/// The maximum number of entries to prune with a single query.
+const MAX_TO_PRUNE_PER_QUERY: u64 = 100;
```
njgheorghita, line 20: Same here w/ why this number was chosen & side effects to expect if it's changed down the road

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -80,8 +80,11 @@ impl VersionedContentStore for IdIndexedV1Store {
 
         let mut store = Self {
             config,
-            to_insert_until_pruning: 0,
             radius: Distance::MAX,
+            usage_stats: UsageStats {
```
njgheorghita, line 84: *nitpick* seems like this could be replaced with a `UsageStats::new()` implementation

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -481,10 +503,9 @@ fn maybe_create_table_and_indexes(
 ) -> Result<(), ContentStoreError> {
     let conn = pool.get()?;
     conn.execute_batch(&sql::create_table(content_type))?;
-    conn.execute_batch(&create_usage_stats_triggers(
+    conn.execute_batch(&delete_usage_stats_triggers(
```
njgheorghita, line 506: Not sure I understand this change?

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -494,17 +515,20 @@ mod tests {
...
     use ethportal_api::{types::portal_wire::ProtocolId, IdentityContentKey, OverlayContentKey};
+    use rand::Rng;
     use tempfile::TempDir;
 
     use crate::{test_utils::generate_random_bytes, utils::setup_sql, DistanceFunction};
 
     use super::*;
 
-    // Storage configuration that stores 10_000 items
+    const CONTENT_DEFAULT_SIZE: u64 = 100;
```
njgheorghita, line 525: `CONTENT_DEFAULT_SIZE_BYTES`?

njgheorghita (approved):

PR #1295, opened 2024-05-10
feat: implement dynamic pruning strategy

morph-dev:
### What was wrong?

Current pruning logic for `IdIndexedV1Store` is to delete up to 5% of storage, but no more than 100 elements. However, even 100 elements is too slow on big databases (e.g. 35GB that are currently used). It can take up to a second and sometimes even longer. This wouldn't be such a big problem if this wouldn't block other read/writes.

### How was it fixed?

The idea is to have dynamic number of items that we delete. The `PruningConfig` allows us to set parameters, for this dynamic strategy:
- what is the optimal pruning duration - default 100-300 milliseconds
- by how much to increase/decrease number of items to prune when pruning is faster/slower than optimal - default 20%

Also added logic for handling the zero storage capacity to prune the db (if non-empty) and set radius to zero.

### To-Do

### DIFF
--- a/trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
+++ b/trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
@@ -13,6 +13,10 @@ pub struct PruningConfig {
     pub target_capacity_fraction: f64,
     /// The fraction by which we increase/decrease the `max_pruning_count` when pruning duration is
     /// outside `optimal_pruning_duration_range`.
+    ///
+    /// For example, let's assume that value is `0.1`. If pruning is too slow, the
+    /// `max_pruning_count` will decrease by 10%, while if pruning is too fast, the
+    /// `max_pruning_count` will increase by 10%. Note that increase and decrease don't cancel out.
     pub max_pruning_count_change_fraction: f64,
     /// The range of pruning durations that we consider optimal.
     pub optimal_pruning_duration_range: Range<Duration>,
@@ -32,7 +36,7 @@ impl PruningConfig {
         max_pruning_count_change_fraction: f64,
         optimal_pruning_duration_range: Range<Duration>,
     ) -> Self {
-        if !(0.0..1.0).contains(&target_capacity_fraction) {
+        if !(0.0..=1.0).contains(&target_capacity_fraction) {
             panic!(
                 "Invalid pruning strategy parameters: target_capacity_fraction={}",
                 target_capacity_fraction
--- a/trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
+++ b/trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
@@ -1,0 +1,360 @@
+use std::{fmt::Debug, ops::Range, time::Duration};
+
+use tracing::debug;
+
+use crate::versioned::usage_stats::UsageStats;
+
+use super::IdIndexedV1StoreConfig;
+
+/// The configuration parameters used by [PruningStrategy].
+#[derive(Clone, Debug)]
+pub struct PruningConfig {
+    /// The fraction of storage capacity that we aim for when pruning.
+    pub target_capacity_fraction: f64,
+    /// The fraction by which we increase/decrease the `max_pruning_count` when pruning duration is
+    /// outside `optimal_pruning_duration_range`.
+    pub max_pruning_count_change_fraction: f64,
+    /// The range of pruning durations that we consider optimal.
+    pub optimal_pruning_duration_range: Range<Duration>,
+}
+
+impl PruningConfig {
+    /// By default, we aim to prune down to 95% of storage capacity.
+    pub const DEFAULT_TARGET_CAPACITY_FRACTION: f64 = 0.95;
+    /// By default, we increase/decrease `max_pruning_count` by 20%.
+    pub const DEFAULT_CHANGE_FRACTION: f64 = 0.2;
+    /// By default, we consider optimal pruning duration between 0.1 and 0.3 seconds.
+    pub const DEFAULT_OPTIMAL_PRUNING_DURATION_RANGE: Range<Duration> =
+        Duration::from_millis(100)..Duration::from_millis(300);
+
+    pub fn new(
+        target_capacity_fraction: f64,
+        max_pruning_count_change_fraction: f64,
+        optimal_pruning_duration_range: Range<Duration>,
+    ) -> Self {
+        if !(0.0..1.0).contains(&target_capacity_fraction) {
+            panic!(
+                "Invalid pruning strategy parameters: target_capacity_fraction={}",
+                target_capacity_fraction
+            )
+        }
+        if !(0.0..1.0).contains(&max_pruning_count_change_fraction) {
+            panic!(
+                "Invalid pruning strategy parameters: change_fraction={}",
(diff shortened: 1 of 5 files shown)


njgheorghita:
It seems like this might be relevant... Taking into account #1228 and that (IMO) we will sooner rather than later want to support this, along with a "max db" size (aka a user has 100 gbs of disk space they want to contribute, we'll actually spin up 10 nodes of 10gbs under the hood). Once that's implemented, we'll have a consistent max db size, which means we really don't have to worry about improving the pruning mechanism to support large dbs and can optimize db performance against a fixed target (eg a 10gb storage). Assuming that my description is the direction we're heading, does this pr still make sense?

pipermerriam:
> we'll actually spin up 10 nodes of 10gbs under the hood)

I don't think this is how the architecture will actually work out.  I think that under the hood it will be a single 100gb database that is represented in the network across 10 different node-ids.

See: https://github.com/ethereum/portal-network-specs/issues/283#issuecomment-2037869161

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -441,26 +455,27 @@ impl IdIndexedV1Store {
...
-            let params = named_params! { ":limit": to_delete };
             let deleted_content_sizes = delete_query
-                .query_map(params, |row| row.get("content_size"))?
+                .query_map(named_params! { ":limit": to_delete }, |row| {
+                    row.get("content_size")
+                })?
                 .collect::<Result<Vec<u64>, rusqlite::Error>>()?;
             self.metrics.stop_process_timer(delete_timer);
+            self.pruning_strategy
+                .observe_pruning_duration(pruning_start_time.elapsed());
```
carver, line 478: Seems like a bummer to have prune run its own `Instant` timer, when we're already calculating the time in the metrics. I suppose if `stop_process_timer` returned the duration it observed (looks like it's an f64 in seconds, but we could convert to a `Duration`), then `observe_duration` could reuse it. Do you agree that would be preferable? If you don't feel like doing that, I might do it after this merges, just so it's ready the next time we want to re-use the timing.

trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
```diff
@@ -0,0 +1,360 @@
...
+    /// By default, we consider optimal pruning duration between 0.1 and 0.3 seconds.
+    pub const DEFAULT_OPTIMAL_PRUNING_DURATION_RANGE: Range<Duration> =
+        Duration::from_millis(100)..Duration::from_millis(300);
+
+    pub fn new(
+        target_capacity_fraction: f64,
+        max_pruning_count_change_fraction: f64,
+        optimal_pruning_duration_range: Range<Duration>,
+    ) -> Self {
+        if !(0.0..1.0).contains(&target_capacity_fraction) {
```
carver, line 35: Nit: I suppose 1.0 is a valid choice here, which Range seems to exclude. Due to database compression, there was even a time when >1.0 choices might be valid. I haven't observed how SQL performs since the switchover. Anyway it all seems mostly moot since it's hard-coded for now, but just wanted to make a note that I think it would be fine to limit the input up to 2.0 or something, since it would just be about letting people play with configuration values eventually, I don't think it's a massive footgun that we need to panic to avoid letting them play with larger values.

trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
```diff
@@ -0,0 +1,360 @@
...
+            panic!(
+                "Invalid pruning strategy parameters: target_capacity_fraction={}",
+                target_capacity_fraction
+            )
+        }
+        if !(0.0..1.0).contains(&max_pruning_count_change_fraction) {
```
carver, line 41: Similarly, it's not totally obvious to me that 1 is a hard upper limit here. If someone wanted to play with 2x the pruning count change, my intuition isn't to panic to prevent them from playing with it.

_Edit: I see where the number came from now, but stand by the above comment when paired with the suggestion below._

trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
```diff
@@ -0,0 +1,360 @@
...
+                Db = %self.config.content_type,
+                "Pruning was too fast. Increasing max_pruning_count",
+            );
+            1. + pruning_config.max_pruning_count_change_fraction
+        } else if pruning_duration > optimal_pruning_duration.end {
+            debug!(
+                Db = %self.config.content_type,
+                "Pruning was too slow. Decreasing max_pruning_count",
+            );
+            1. - pruning_config.max_pruning_count_change_fraction
```
carver, line 154: Ah, I see now where the upper limit of 1 came from. My intuition before getting to this code is that a change fraction of 1.0 would double on the upside and cut in half on the downside. Those seem like the more symmetric options to me. (ie~ one increase and one decrease takes you back to where you started)

In this approach, choosing a value of .9 nearly doubles on the upside, but is 1/10th on the downside, which is aggressively dropping but not that aggressively climbing (it takes more than 3 upside adjustments to make up for 1 downside adjustment). So if we want one number to govern both directions, my intuition says that this line should be:
```suggestion
            1. / (1. + pruning_config.max_pruning_count_change_fraction)
```

carver (approved):
LGTM! Mostly just nit comments that you're welcome to ignore.

I apparently have a strong opinion about the ideal way to calculate the percentage update of the pruning number (more below). But in the end, it's not a blocker because even if the number isn't ideal, I don't foresee any catastrophe. So don't let me stop you from merging it if you disagree about the percentage update 😆

ogenev (approved):
⛵

njgheorghita (approved):

trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
morph-dev: Definitely while coding I had a mental model that `target_capacity` should we lower than `storage_capacity`.

Maybe code would work even if that's not the case, but I would say it doesn't makes much sense, because both storage capacity and target capacity represent the raw content size (`content_id.len() + content_key.len() + content_value.len()`), or at least approximation (there is some extra storage used for indexes and other smaller columns). Irrelevant, but I don't think that there is any compression happening at the moment.

With that being said, I will allow value 1.0. And once we have real use case for using values higher than 1, we can come back to it (in which case we would probably have to rewrite some of the logic).

trin-storage/src/versioned/id_indexed_v1/pruning_strategy.rs
morph-dev: I actually wanted increasing the `max_pruning_count` to be harder than decreasing, because downside of being too low is not so bad (we prune less items more frequently) while downside of being too high is that db is blocked for longer.

I added better explanation to the `max_pruning_count_change_fraction` field.

However, if the behavior that you expected is more common way of doing it, I'm willing to change it in the followup PR.

morph-dev (commented):
@carver I do disagree with the proposed changes about pruning number changes, but not strongly :). Willing to discuss it more before merging.

morph-dev:
> @carver I do disagree with the proposed changes about pruning number changes, but not strongly :). Willing to discuss it more before merging.

I'm going to merge this make a followup PR if needed.

PR #1297, opened 2024-05-12
chore: remove unused PortalnetConfig::data_radius field

morph-dev:
### What was wrong?

The `data_radius` of `PortalnetConfig` is not used.

### How was it fixed?

I removed the field

### To-Do


KolbyML (approved):
:shipit:

PR #1299, opened 2024-05-16
feat(beacon-storage): store `HistoricalSummariesWithProof` content and add beacon storage tests

ogenev:
### What was wrong?

We don't store `HistoricalSummariesWithProof` in db.

### How was it fixed?
- store the latest "HistoricalSummariesWithProof" version in the beacon cache storage
- add tests for beacon storage

### To-Do


KolbyML (approved):
:shipit:

PR #1300, opened 2024-05-17
fix(storage): skip error log if storage config is 0

njgheorghita:
### What was wrong?
If storage capacity is set to 0, you get a whole lot of these `ERROR` warnings, which also sets the `data_radius` to `MAX` incorrectly.

### How was it fixed?
Skip pruning if storage config is set to 0

### To-Do

### DIFF
--- a/trin-storage/src/versioned/id_indexed_v1/store.rs
+++ b/trin-storage/src/versioned/id_indexed_v1/store.rs
@@ -436,6 +436,12 @@ impl IdIndexedV1Store {
             return Ok(());
         }
 
+        if self.config.storage_capacity_bytes == 0 {
+            warn!(Db = %self.config.content_type,
+                "Pruning requested but storage capacity is 0. Skipping");
+            return Ok(());
+        }
+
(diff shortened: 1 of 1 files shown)


njgheorghita:
Actually, not sure yet if this is the best way to solve this...

morph-dev:
Probably the correct way would be to set `radius` to zero during init. I can add that logic to my PR #1295 that is already refactoring pruning.

Btw, what is the error that you are seeing?
"Farthest not found!" or "Pruning requested but we are below target capacity" or both or something else?

morph-dev:
Added logic to #1295. It will also make sure to empty the db (prune everything) at the startup.

njgheorghita:
@morph-dev it's the `Farthest not found!` error, also if you check the scatter plot on glados, you can see that the bridge nodes are being incorrectly set to a radius of 100. But, imo it seems that `probably the correct way would be to set radius to zero during init.` is the way to go and should solve these issues. Closing this pr in favor of #1295

PR #1301, opened 2024-05-18
chore: cleanup no longer used code from trin-storage

morph-dev:
### What was wrong?

After migrating history to `IdIndexedV1Store`, we have quite a bit no longer used code.

### How was it fixed?

Removed most of the unused code (legacy history store and usage stats triggers).

Left for future PR:

- migration from legacy_history
- dropping `usage_stats` table

### To-Do


KolbyML (approved):
:shipit:

PR #1303, opened 2024-05-19
chore: update git submodules and make tests pass

morph-dev:
### What was wrong?

The `portal-spec-tests` submodule was updated.

### How was it fixed?

Updated the submodules and make sure that tests pass.

The state validator doesn't verify block's state root at the moment but it should. Once that is implemented, tests should be updated and use data in updated files.

### To-Do

### DIFF
--- a/ethportal-peertest/src/utils.rs
+++ b/ethportal-peertest/src/utils.rs
@@ -141,7 +141,7 @@ fn read_state_fixture_from_file(file_name: &str) -> Result<Vec<StateFixture>> {
     let yaml_content = fs::read_to_string(file_name)?;
     let value: Value = serde_yaml::from_str(&yaml_content)?;
 
-    let result = <Vec<StateFixture>>::deserialize(value)?;
+    let result = Vec::<StateFixture>::deserialize(value)?;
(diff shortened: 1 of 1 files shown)


ethportal-peertest/src/utils.rs
```diff
@@ -138,20 +141,7 @@ fn read_state_fixture_from_file(file_name: &str) -> Result<Vec<StateFixture>> {
...
-            recursive_gossip: fixture.get("recursive_gossip").and_then(|value| {
-                if value.is_null() {
-                    None
-                } else {
-                    Some(StateContentData::deserialize(value).unwrap())
-                }
-            }),
-        });
-    }
+    let result = <Vec<StateFixture>>::deserialize(value)?;
```
KolbyML, line 144:
```suggestion
    let result = <Vec<StateFixture>>::deserialize(value)?;
```
Do we need the extra <>, maybe we do it is just my first time seeing something like this :monocle_face:

KolbyML (approved):
:shipit:

ogenev (approved):

PR #1304, opened 2024-05-20
fix(bridge): add floor to hunter mode

njgheorghita:
### What was wrong?
Block range prior to `4.200.000` is very well saturated already. I've noticed on my personal "hunter" nodes that they still spend a significant amount of time trying to find un-saturated epochs from this range, so to expedite the hunting process it seems worthwhile to implement a floor to the hunter. This can easily be reversed / undone down the road once 4444s is thoroughly saturated.

### How was it fixed?
Skip epochs < 512 in hunter mode.

### To-Do


ogenev (approved):

njgheorghita:
closing this for now since i'm no longer convinced this is necessary to merge into master, and found a more effective strategy that i've just deployed to my personal bridges

PR #1306, opened 2024-05-22
refactor: make `stop_process_timer` return duration

morph-dev:
### What was wrong?

The `StorageMetricsReporter` doesn't return elapsed duration when `stop_process_timer` is called, which can be useful in some situations.

This is a followup from #1295.

### How was it fixed?

Make `stop_process_timer` return elapsed duration and use it to simplify the logic for updating pruning strategy.

### To-Do


njgheorghita (approved):
🚢

PR #1310, opened 2024-05-24
fix: improve latest reliability by using eth_getBlockReceipts

KolbyML:
### What was wrong?
We were getting null for a lot of receipts via ``eth_getTransactionReceipt`` using batch requests, felix from geth said that using this like we are is very bad practice
### How was it fixed?

by using eth_getBlockReceipts this command returns all receipts for the block avoiding the need to make multiple requests or batch requests for receipts 

I also added a little retry loop which should increase reliability as well


portal-bridge/src/api/execution.rs
```diff
@@ -217,20 +218,26 @@ impl ExecutionApi {
...
-                let method = "eth_getTransactionReceipt".to_string();
-                JsonRequest::new(method, params, id as u32)
-            })
-            .collect();
-        let response = self.batch_requests(request).await?;
-        serde_json::from_str(&response)
-            .map_err(|err| anyhow!("Unable to parse receipts from provider response: {err:?}"))
+    async fn get_trusted_receipts(&self, height: u64) -> anyhow::Result<Receipts> {
+        let block_param = format!("0x{height:01X}");
+        for _ in 0..3 {
```
njgheorghita, line 223: I'm not sure if this is the right place to implement a "retry" mechanism. If we do want this 3x retry mechanism, then it makes more sense to me to implement this inside `try_request()` - so that all functions that use `try_request()` will use it. But inside `try_request()` we already have a "fallback" attempt to the fallback provider, so I'm not even sure that this 3x retry mechanism is necessary?

njgheorghita:
@KolbyML any updates here? iiuc, with this change we might be able to ditch our paid infura account, which would be nice

portal-bridge/src/api/execution.rs
KolbyML: I don't think this retry method should be used in `try_request()`. I added it because of the specific constraints of `eth_getBlockReceipts`. At the moment I don't have the bandwidth or think this is high enough priority to think of the implications of implementing this suggestion.

I don't know if this retry mechanism would work for `get_block_number` or `get_header` and I don't think it should be as well.

The reason for this retry mechanism, is because we only call it when we *know* receipts should be available. In the case of fetching a latest header, this assumption isn't valid. I am also not aware if `get_block_number` etc use the same response pattern.

KolbyML: Sometimes a receipt should be a avaliable but is still process, which can cause `getReceipts to return none`, which we seen in our method used before this PR

njgheorghita: Ok gotcha, then if you don't mind just adding a quick comment explaining why we have a dedicated retry for receipts only, since it's a special case where sometimes they're not returned when we expect them to be available

njgheorghita (approved):
:ship:

PR #1311, opened 2024-05-25
chore: add trin docker compose file

KolbyML:
### What was wrong?
People want to run trin in docker images, but doing it with raw docker is complex for someone new to Trin
### How was it fixed?
By making a good starter docker compose file which people can expand easily with what they want

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,14 +1,13 @@
 version: '3'
 services:
   trin:
+    container_name: trin
     hostname: trin
     image: portalnetwork/trin:latest
     environment:
-      RUST_LOG: info
+      - RUST_LOG: info
     command: "--web3-transport http --web3-http-address http://0.0.0.0:8545/"
     network_mode: "host"
     volumes: 
-      - $HOME/.local/share:$HOME/.local/share
-    ports:
-      - "8545:8545"
+      - ${HOME}/.local/share:${HOME}/.local/share
     restart: always
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,0 +1,14 @@
+version: '3'
+services:
+  trin:
+    hostname: trin
+    image: portalnetwork/trin:latest
+    environment:
+      RUST_LOG: info
+    command: "--web3-transport http --web3-http-address http://0.0.0.0:8545/"
+    network_mode: "host"
+    volumes: 
+      - $HOME/.local/share:$HOME/.local/share
+    ports:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    hostname: trin
+    image: portalnetwork/trin:latest
+    environment:
+      RUST_LOG: info
+    command: "--web3-transport http --web3-http-address http://0.0.0.0:8545/"
+    network_mode: "host"
+    volumes: 
+      - $HOME/.local/share:$HOME/.local/share
+    ports:
+      - "8545:8545"
```
carver, line 13: Wouldn't this also need to expose the discovery port, default 9009?

KolbyML: Since host mode is enabled I should remove the ports entry as in this case we use the host ports so this entry doesn't do anything

ogenev: Yeah, we don't need the port mapping if "host" is enabled.

docker/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
+services:
+  trin:
+    hostname: trin
```
ogenev, line 4: I would add a `container_name` for more clarity.

```suggestion
  trin:
    container_name: trin
    hostname: trin
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: portalnetwork/trin:latest
+    environment:
+      RUST_LOG: info
+    command: "--web3-transport http --web3-http-address http://0.0.0.0:8545/"
+    network_mode: "host"
+    volumes: 
+      - $HOME/.local/share:$HOME/.local/share
```
ogenev, line 11: nit for a more explicit path:

```suggestion
      - ${HOME}/.local/share:${HOME}/.local/share
```

docker/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3'
+services:
+  trin:
+    hostname: trin
+    image: portalnetwork/trin:latest
+    environment:
+      RUST_LOG: info
```
ogenev, line 7: nit:

```suggestion
    environment:
      -  RUST_LOG: info
```

ogenev (approved):
⛵

carver (approved):
:shipit: with the ports removed. Plus 1 to Ognyan's feedback.

PR #1314, opened 2024-05-30
feat: update testnet name to angelfood

KolbyML:
### What was wrong?
we were using the wrong name (aka made one up during portal summit
### How was it fixed?

by using it


ogenev (approved):
⛵

Issue #1292, opened 2024-05-09
Bridge(State): don't miss gossipping intermediate state roots


KolbyML:
> It is still possible to have some trie reorgs that you wouldn't gossip this way, meaning it's possible that trie node is modified that is not in the path of any modified account
I can elaborate more on this if needed.

On call me and Milos talked about it. I am likely to work on this after adding auditing state on glados

PR #1316, opened 2024-06-07
fix: gossip state from trie diff and gossip every intermediate node

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/issues/1292

Calling get proof on all changed accounts missing changed intermediate trie nodes

recursive gossip doesn't work so we have to gossip proofs for every intermediate try node

### How was it fixed?

This PR gossips every intermediate node (as from the summit we discussed recursive gossip doesn't work)

This PR also makes it so we don't miss a changed node, by utilizing a Trie Diff

By getting the Trie diff from eth-trie.rs

Then by walking the diff building back traces then iterating over all the intermediate trie keys and building proofs to gossip

### DIFF
--- a/portal-bridge/Cargo.toml
+++ b/portal-bridge/Cargo.toml
@@ -23,8 +23,9 @@ discv5 = { version = "0.4.1", features = ["serde"] }
 ethereum_ssz = "0.5.3"
 ethereum_ssz_derive = "0.5.3"
 ethportal-api = { path = "../ethportal-api" }
-eth_trie = { git = "https://github.com/kolbyml/eth-trie.rs.git", rev = "11ec003e3276e1413f06328ab746af5d99f112bb" }
+eth_trie = { git = "https://github.com/kolbyml/eth-trie.rs.git", rev = "a69f99cdee5ea37edddd1147058f1990a5d943eb" }
 futures = "0.3.21"
+hashbrown = "0.14.0"
 jsonrpsee = { version = "0.20.0", features = [
   "async-client",
   "client",
--- a/portal-bridge/src/bridge/state.rs
+++ b/portal-bridge/src/bridge/state.rs
@@ -2,6 +2,7 @@ use std::{path::PathBuf, sync::Arc};
 
 use super::era1::get_shuffled_era1_files;
 use anyhow::anyhow;
+use eth_trie::RootWithTrieDiff;
 use ethportal_api::{jsonrpsee::http_client::HttpClient, StateContentKey, StateContentValue};
 use revm_primitives::SpecId;
 use surf::{Client, Config};
@@ -24,6 +25,7 @@ use crate::{
             content::{create_content_key, create_content_value},
             execution::State,
             spec_id::get_spec_block_number,
+            trie_walker::TrieWalker,
         },
     },
 };
@@ -91,7 +93,7 @@ impl StateBridge {
         let gossip_send_semaphore = Arc::new(Semaphore::new(self.gossip_limit));
         let mut current_epoch_index = u64::MAX;
         let mut current_raw_era1 = vec![];
-        let mut state = State::new().map_err(|e| anyhow!("unable to create genesis state: {e}"))?;
+        let mut state = State::new();
         for block_index in 0..=last_block {
             info!("Gossipping state for block at height: {block_index}");
             let epoch_index = block_index / EPOCH_SIZE;
@@ -114,19 +116,26 @@ impl StateBridge {
 
             // process block
             let block_tuple = Era1::get_tuple_by_index(&current_raw_era1, block_index % EPOCH_SIZE);
-            let updated_accounts = match block_index == 0 {
-                true => state.database.accounts.keys().copied().collect(),
+            let RootWithTrieDiff {
+                root: root_hash,
+                trie_diff: changed_nodes,
+            } = match block_index == 0 {
+                true => state
+                    .initialize_genesis()
+                    .map_err(|e| anyhow!("unable to create genesis state: {e}"))?,
                 false => state.process_block(&block_tuple)?,
             };
 
+            let walk_diff = TrieWalker::new(root_hash, changed_nodes);
+
             // gossip block's new state transitions
-            for updated_account in updated_accounts {
+            for node in walk_diff.nodes.keys() {
                 let permit = gossip_send_semaphore
                     .clone()
                     .acquire_owned()
                     .await
                     .expect("to be able to acquire semaphore");
-                let account_proof = state.get_proof(&updated_account)?;
+                let account_proof = walk_diff.get_proof(*node);
                 let content_key = create_content_key(&account_proof)?;
                 let content_value =
                     create_content_value(block_tuple.header.header.hash(), &account_proof)?;
--- a/portal-bridge/src/types/state/execution.rs
+++ b/portal-bridge/src/types/state/execution.rs
@@ -1,7 +1,7 @@
 use alloy_primitives::{keccak256, Address, Bytes, B256, U256};
 use alloy_rlp::EMPTY_STRING_CODE;
-use anyhow::{ensure, Error};
-use eth_trie::Trie;
+use anyhow::{anyhow, bail, ensure, Error};
+use eth_trie::{RootWithTrieDiff, Trie};
 use ethportal_api::types::{
     execution::transaction::Transaction,
     state_trie::account_state::AccountState as AccountStateInfo,
@@ -9,7 +9,7 @@ use ethportal_api::types::{
 use revm::{
     db::EmptyDB, inspector_handle_register, inspectors::NoOpInspector, DatabaseCommit, Evm,
 };
-use revm_primitives::{Env, ResultAndState, SpecId};
+use revm_primitives::{Env, SpecId};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{BTreeSet, HashMap},
@@ -28,6 +28,8 @@ use crate::types::{
     },
 };
 
+use super::trie_walker::AccountProof;
+
 #[derive(Debug, Serialize, Deserialize)]
 struct AllocBalance {
     balance: U256,
@@ -40,13 +42,6 @@ struct GenesisConfig {
     state_root: B256,
 }
 
-#[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct AccountProof {
-    pub address: Address,
-    pub state: AccountStateInfo,
-    pub proof: Vec<Bytes>,
-}
-
 pub struct State {
     pub database: CacheDB<EmptyDB>,
     next_block_number: u64,
@@ -54,21 +49,36 @@ pub struct State {
 
 const GENESIS_STATE_FILE: &str = "resources/genesis/mainnet.json";
 
+impl Default for State {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl State {
-    pub fn new() -> anyhow::Result<Self> {
-        let mut cache_db = CacheDB::new(EmptyDB::default());
-        let mut accounts = BTreeSet::new();
+    pub fn new() -> Self {
+        State {
+            next_block_number: 0,
+            database: CacheDB::new(EmptyDB::default()),
+        }
+    }
+
+    pub fn initialize_genesis(&mut self) -> anyhow::Result<RootWithTrieDiff> {
+        ensure!(
+            self.next_block_number == 0,
+            "Trying to initialize genesis but received block {}",
+            self.next_block_number,
+        );
 
         let genesis_file = File::open(GENESIS_STATE_FILE)?;
         let genesis: GenesisConfig = serde_json::from_reader(BufReader::new(genesis_file))?;
 
         for (address, alloc_balance) in genesis.alloc {
-            accounts.insert(address);
             let mut account = DbAccount::default();
             account.info.balance += alloc_balance.balance;
             let account_info = account.info.clone();
-            cache_db.accounts.insert(address, account);
-            cache_db.trie.lock().insert(
+            self.database.accounts.insert(address, account);
+            self.database.trie.lock().insert(
                 keccak256(address).as_ref(),
                 &alloy_rlp::encode(AccountStateInfo {
                     nonce: account_info.nonce,
@@ -79,18 +89,18 @@ impl State {
             )?;
         }
 
-        let root = B256::from_slice(cache_db.trie.lock().root_hash()?.as_slice());
+        let root_with_trie_diff = self.get_root_with_trie_diff()?;
         ensure!(
-            root == genesis.state_root,
+            root_with_trie_diff.root == genesis.state_root,
             "Root doesn't match state root from genesis file"
         );
-        Ok(State {
-            next_block_number: 1,
-            database: cache_db,
-        })
+
+        self.next_block_number += 1;
+
+        Ok(root_with_trie_diff)
     }
 
-    pub fn process_block(&mut self, block_tuple: &BlockTuple) -> anyhow::Result<BTreeSet<Address>> {
+    pub fn process_block(&mut self, block_tuple: &BlockTuple) -> anyhow::Result<RootWithTrieDiff> {
         info!(
             "State EVM processing block {}",
             block_tuple.header.header.number
@@ -102,9 +112,7 @@ impl State {
             block_tuple.header.header.number
         );
 
-        let mut updated_accounts: BTreeSet<Address> = BTreeSet::new();
-
-        // initalize evm environment
+        // initialize evm environment
         let mut env = Env::default();
         env.block.number = U256::from(block_tuple.header.header.number);
         env.block.coinbase = block_tuple.header.header.author;
@@ -142,7 +150,7 @@ impl State {
             .transactions()
             .map_err(|err| Error::msg(format!("Error getting transactions: {err:?}")))?;
         for transaction in transactions {
-            updated_accounts.extend(self.execute_transaction(transaction, &env)?);
+            self.execute_transaction(transaction, &env)?;
         }
 
         // update beneficiary
@@ -164,10 +172,13 @@ impl State {
                     self.database.accounts.insert(beneficiary, account);
                 }
             }
-            updated_accounts.insert(beneficiary);
         }
 
-        if self.get_root()? != block_tuple.header.header.state_root {
+        let RootWithTrieDiff {
+            root,
+            trie_diff: changed_nodes,
+        } = self.get_root_with_trie_diff()?;
+        if root != block_tuple.header.header.state_root {
             panic!(
                 "State root doesn't match! Irreversible! Block number: {}",
                 self.next_block_number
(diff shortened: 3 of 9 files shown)


portal-bridge/src/types/state/execution.rs
```diff
@@ -164,10 +174,13 @@ impl State {
...
                 }
             }
-            updated_accounts.insert(beneficiary);
         }
 
-        if self.get_root()? != block_tuple.header.header.state_root {
+        let RootWithTrieDiff {
+            root,
+            trie_diff: changed_nodes,
+        } = self.get_root_with_changed_nodes()?;
```
morph-dev, line 182: There is one thing that crossed my mind that worries me because I'm not sure how `eth_trie` works.

Are we sure that `get_root_with_changed_nodes` doesn't return nodes that didn't change? (e.g. they were just read during transaction but not modified)

morph-dev (changes requested):
Some important, some less important topics. Happy to jump into a call to discuss things.

portal-bridge/src/bridge/state.rs
```diff
@@ -91,7 +93,12 @@ impl StateBridge {
         let gossip_send_semaphore = Arc::new(Semaphore::new(self.gossip_limit));
         let mut current_epoch_index = u64::MAX;
         let mut current_raw_era1 = vec![];
-        let mut state = State::new().map_err(|e| anyhow!("unable to create genesis state: {e}"))?;
+        let StateWithTrieDiff {
+            mut state,
+            trie_diff: changed_nodes,
```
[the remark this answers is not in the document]
KolbyML: Why would we include deleted nodes? I am not aware how that information would be useful. We can update the name if you want

portal-bridge/src/bridge/state.rs
```diff
@@ -114,19 +121,27 @@ impl StateBridge {
...
+                    trie_diff: changed_nodes.clone(),
+                },
                 false => state.process_block(&block_tuple)?,
             };
 
+            let walk_diff = WalkTrieDiff::new(root_hash, changed_nodes);
+
             // gossip block's new state transitions
-            for updated_account in updated_accounts {
+            for node in walk_diff.changed_nodes.keys() {
```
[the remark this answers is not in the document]
KolbyML: Yeah we can make an issue, as that isn't valuable until we need to run multiple state bridges but we aren't there yet

portal-bridge/src/types/state/execution.rs
```diff
@@ -176,14 +189,17 @@ impl State {
...
     }
 
     fn execute_transaction(
         &mut self,
         tx: Transaction,
         evm_evnironment: &Env,
-    ) -> anyhow::Result<BTreeSet<Address>> {
+    ) -> anyhow::Result<()> {
         let block_number = evm_evnironment.block.number.to::<u64>();
         let ResultAndState { state: changes, .. } = {
```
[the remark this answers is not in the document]
KolbyML: transact_commit() doesn't appear to be a thing. I update it to inline the transact.

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
...
+                                    .insert(hash_node.hash.to_vec(), (node_key.to_vec(), i as u8));
+                            }
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        path_backwards
```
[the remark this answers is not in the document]
KolbyML: > I'm not sure if it is possible in the real scenario (definitely it is if there is a bug), but what would happen if there is a node in changed_nodes that is not reachable from provided root?

This wouldn't be possible. changed_nodes includes all the changed trie nodes. So all nodes must be reachable by the root, unless someone did something wrong.

The assumption is we aren't inserting invalid `changed_nodes` into the TrieWalker.

If someone is inserting invalid data, well then of course it would output invalid data as aswell

Garbage in Garbage out

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
...
+
+            match decoded_node {
+                Node::Extension(extension) => {
+                    let extension = extension.read().expect("Reading an extension should work");
+                    if let Node::Hash(hash_node) = &extension.node {
+                        if changed_nodes.contains_key(&hash_node.hash.to_vec()) {
+                            stack.push(hash_node.hash);
+                            path_backwards.insert(
+                                hash_node.hash.to_vec(),
+                                (node_key.to_vec(), extension.prefix.get_data()[0]),
```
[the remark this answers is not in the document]
KolbyML: Shouldn't the extension only be 1 or 0?

portal-bridge/src/types/state/execution.rs
KolbyML: I don't read from the trie, I only update/delete accounts. For reading during evm execution I interface with the accounts in a flat key value store. 

> they were just read during transaction

This interaction isn't done at least in how I implemented the code, so in this case it would only return updated nodes. Maybe someone could do X, but I didn't read or implement using X. How I implemented things we should be good.

portal-bridge/Cargo.toml
```diff
@@ -23,8 +23,9 @@ discv5 = { version = "0.4.1", features = ["serde"] }
 ethereum_ssz = "0.5.3"
 ethereum_ssz_derive = "0.5.3"
 ethportal-api = { path = "../ethportal-api" }
-eth_trie = { git = "https://github.com/kolbyml/eth-trie.rs.git", rev = "11ec003e3276e1413f06328ab746af5d99f112bb" }
+eth_trie = { git = "https://github.com/kolbyml/eth-trie.rs.git", rev = "4d72f83a1dcb12750038e0d376576e2198541251" }
```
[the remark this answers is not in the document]
KolbyML: I will update it

portal-bridge/src/types/state/execution.rs
```diff
@@ -281,4 +323,25 @@ mod tests {
             );
         }
     }
+
+    #[test_log::test]
+    fn test_walk_trie_diff_builds_valid_proof() {
```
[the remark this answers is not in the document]
KolbyML: It is testing `walk_diff.get_proof`

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
+use alloy_primitives::{Bytes, B256};
+use eth_trie::{decode_node, node::Node};
+use hashbrown::HashMap as BrownHashMap;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AccountProof {
```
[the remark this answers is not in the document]
KolbyML: You already use it for something else. We can worry about ContractStorage tries when we add support to use them, but currently that is out of scope of this PR.

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub path: Vec<u8>,
+    pub proof: Vec<Bytes>,
+}
+
+/// This struct takes in a root hash and a hashmap of changed nodes, then you can call an iterator
```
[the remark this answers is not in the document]
KolbyML: > parent_info

`path_backwards` says it how it is. Not a fan of `parent_info` as it has nothing to do this the parents info, it is a the path to the parent, so I could comprise on `paths_to_parent`

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
...
+/// which will return every proof to gossip
+pub struct WalkTrieDiff {
+    pub changed_nodes: BrownHashMap<Vec<u8>, Vec<u8>>,
+    path_backwards: BrownHashMap<Vec<u8>, (Vec<u8>, u8)>,
```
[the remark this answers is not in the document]
KolbyML: > if you want more optimized version

I am currently deleting all the in memory database code in my next PR. I don't think it is the time to optimize stuff when it can and will be all rewritten over and over none of this code is stable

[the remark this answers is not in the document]
KolbyML: I will call it PathToParent, as it isn't ParentInfo

portal-bridge/src/types/state/walk_trie_diff.rs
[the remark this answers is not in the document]
KolbyML: Oh I see now

portal-bridge/src/types/state/walk_trie_diff.rs
```diff
@@ -0,0 +1,101 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AccountProof {
+    pub path: Vec<u8>,
+    pub proof: Vec<Bytes>,
+}
+
+/// This struct takes in a root hash and a hashmap of changed nodes, then you can call an iterator
+/// which will return every proof to gossip
+pub struct WalkTrieDiff {
+    pub changed_nodes: BrownHashMap<Vec<u8>, Vec<u8>>,
```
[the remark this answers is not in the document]
KolbyML: It looks like eth_trie should just be using `B256` for keys as it has no right using Vec<u8>'s for keys.

Good catch

portal-bridge/src/types/state/execution.rs
```diff
@@ -40,35 +42,43 @@ struct GenesisConfig {
...
+        State {
+            next_block_number: 0,
+            database: CacheDB::new(EmptyDB::default()),
+        }
+    }
+
+    pub fn initialize_genesis(&mut self) -> anyhow::Result<RootWithTrieDiff> {
+        ensure!(
+            self.next_block_number == 0,
+            "Expected block {}, received 0",
```
morph-dev, line 69: nit: This error is a bit confusing. Maybe something like: `Trying to initialize genesis but expecting block {}`.

portal-bridge/src/types/state/execution.rs
```diff
@@ -176,17 +187,20 @@ impl State {
...
     fn execute_transaction(
         &mut self,
         tx: Transaction,
         evm_evnironment: &Env,
-    ) -> anyhow::Result<BTreeSet<Address>> {
+    ) -> anyhow::Result<()> {
         let block_number = evm_evnironment.block.number.to::<u64>();
         let ResultAndState { state: changes, .. } = {
-            let mut evm = Evm::builder()
+            Evm::builder()
```
morph-dev, line 203: nit: I find this slightly harder to read. Maybe something like:

```rust
let tx_result = Evm::builder()
    ...
    .build()
    .transact()?;
self.database.commit(tx_result.state);
```

portal-bridge/src/types/state/execution.rs
```diff
@@ -207,18 +221,21 @@ impl State {
...
         self.database.commit(changes);
-        Ok(updated_accounts)
+        Ok(())
     }
 
     pub fn get_root(&mut self) -> anyhow::Result<B256> {
         Ok(self.database.trie.lock().root_hash()?)
     }
 
+    pub fn get_root_with_changed_nodes(&mut self) -> anyhow::Result<RootWithTrieDiff> {
```
morph-dev, line 235: nit: maybe `get_root_with_trie_diff`?

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,155 @@
...
+                .get(&node_key)
+                .expect("The stack should only contain nodes that are in the changed nodes");
+
+            let decoded_node = decode_node(&mut encoded_node.as_slice())
+                .expect("Should should only be passing valid encoded nodes");
+
+            match decoded_node {
+                Node::Extension(extension) => {
+                    let extension = extension.read().expect("Reading an extension should work");
+                    if let Node::Hash(hash_node) = &extension.node {
```
morph-dev, line 59: I think you misunderstood my comment from previous review. I wanted you to comment why we only care when `extension.node` is `Node::Hash`, and not other cases (and I believe you understood that I wanted you to comment the `if nodes.contains_key(&hash_node.hash) {` below).

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,155 @@
...
+                                    PathToParent::new(node_key, vec![i as u8]),
+                                );
+                            }
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
+
```
morph-dev, line 90: Regarding previous discussion, I would add a check here:
```rust
// Ensure we discovered all nodes (root is not present in the paths_to_parent)
ensure!(nodes.len() == paths_to_parent.len() + 1, "...");
```

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,155 @@
...
+                .expect("The stack should only contain nodes that are in the changed nodes");
+
+            let decoded_node = decode_node(&mut encoded_node.as_slice())
+                .expect("Should should only be passing valid encoded nodes");
+
+            match decoded_node {
+                Node::Extension(extension) => {
+                    let extension = extension.read().expect("Reading an extension should work");
+                    if let Node::Hash(hash_node) = &extension.node {
+                        // We could be walking a partial trie, so we need to check if the node is in
```
morph-dev, line 60: nit: While I understand what you wanted to say, I find this comment a bit confusing and I imagine somebody not fully familiar with functionality might as well.

I would add something like:
_Only process provided nodes (they belong to the partial trie that we care about)_

portal-bridge/src/types/state/walk_trie_diff.rs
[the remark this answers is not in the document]
morph-dev: It's more path **from** parent.

To me, parent_info makes sense because it's a map where key is a node and value are information about it's parent in the trie: hash+path_from_parent.

I'm fine with just `parent` to simplify and avoid confusion.

portal-bridge/src/types/state/trie_walker.rs
KolbyML: I gave it a shot let me know what you think

portal-bridge/src/types/state/trie_walker.rs
KolbyML: This is so nice good catch

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,155 @@
...
+use hashbrown::HashMap as BrownHashMap;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AccountProof {
+    pub path: Vec<u8>,
+    pub proof: Vec<Bytes>,
+}
+
+pub struct PathToParent {
```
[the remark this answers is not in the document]
KolbyML: Sounds good 🚀

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,155 @@
...
+        let mut proof = vec![];
+        let mut current_node = node_hash;
+        proof.push(Bytes(
+            self.nodes
+                .get(&current_node)
+                .expect("Changed nodes should contain this intermediate node")
+                .clone()
+                .into(),
+        ));
+        while let Some(PathToParent {
```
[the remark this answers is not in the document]
KolbyML: Clean 👍

portal-bridge/src/types/state/walk_trie_diff.rs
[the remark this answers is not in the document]
KolbyML: `parent` is too simple. People should be able to understand what the hashmap stores just from reading the name. They shouldn't need to read a comment.

So I would want it to indicate it is a reverse map from the child to the path from the parent

KolbyML:
@morph-dev ready for another look, let me know if I missed anything

portal-bridge/src/types/state/execution.rs
```diff
@@ -79,18 +89,18 @@ impl State {
             )?;
         }
 
-        let root = B256::from_slice(cache_db.trie.lock().root_hash()?.as_slice());
+        let root_with_trie_diff = self.database.trie.lock().root_hash_with_changed_nodes()?;
```
morph-dev, line 92: nit: would it make sense to do `self.get_root_with_trie_diff()?` instead?

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
...
+            path_nibbles,
+        }
+    }
+}
+
+/// This struct takes in a root hash and a hashmap of changed nodes, then you can call an iterator
+/// which will return every proof to gossip
+pub struct TrieWalker {
+    pub nodes: BrownHashMap<B256, Vec<u8>>,
+    paths_to_parent: BrownHashMap<B256, ParentWithPath>,
```
morph-dev, line 31: nit: Since we renamed the type, I would also rename this field (and everywhere below, including `get_paths_backwards` function).

I would say just `parents` is fine, but `parent_with_path_map` or something similar is also ok.

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
...
+            let decoded_node = decode_node(&mut encoded_node.as_slice())
+                .expect("Should should only be passing valid encoded nodes");
+
+            match decoded_node {
+                Node::Extension(extension) => {
+                    let extension = extension.read().expect("Reading an extension should work");
+                    // We only look for hash nodes, as commit the changes to the patrica merkle trie
+                    // it converts all the nodes to hash nodes, so they can be stored in the db
+                    if let Node::Hash(hash_node) = &extension.node {
+                        //Only process provided nodes (they belong to the partial trie that we care
```
morph-dev, line 64: nit: space after `//`

portal-bridge/src/types/state/trie_walker.rs
morph-dev: It's actually not true. Nodes are converted to hash nodes only if their encoded version is 32 bytes or longer, which is not always the case for the contract storage trie.

I would put something like:

_We look for hash nodes in order to connect them to the root. If this node is not a hash node, then neither is any of its children._
_We know this because any node that has a hash node as it's descendant would also become hash node, because its encoding would be longer than 32 bytes._

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
...
+                                hash_node.hash,
+                                ParentWithPath::new(node_key, extension.prefix.get_data().to_vec()),
+                            );
+                        }
+                    }
+                }
+                Node::Branch(branch) => {
+                    let branch = branch.read().expect("Reading a branch should work");
+                    for (i, child) in branch.children.iter().enumerate() {
+                        if let Node::Hash(hash_node) = child {
```
morph-dev, line 78: nit: Also add comment here

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
...
+                            }
+                        }
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        // Ensure we discovered all nodes (root is not present in the paths_to_parent)
+        ensure!(nodes.len() == paths_to_parent.len() + 1, "We expect the node length and paths_to_parent length to be the same, as we should have discovered all nodes in the trie. If this fails it implies an invalid trie was trie was provided with an unnavigable path.");
```
morph-dev, line 96:
```suggestion
        ensure!(nodes.len() == paths_to_parent.len() + 1, "We expect the node length and paths_to_parent length to be the same, as we should have discovered all nodes in the trie. If this fails it implies an invalid trie was provided with an unnavigable path.");
```

morph-dev (approved):

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
...
+        Self {
+            parent_hash,
+            path_nibbles,
+        }
+    }
+}
+
+/// This struct takes in a root hash and a hashmap of changed nodes, then you can call an iterator
+/// which will return every proof to gossip
+pub struct TrieWalker {
```
morph-dev, line 29: One thing crossed my mind right after submitting previous review.

I think it would make sense to change this structure to something like:

```rust

struct TrieWalkerNode {
  /// The encoded version of the trie node.
  encoded_node: Vec<u8>,
  /// The hash of the parent node. It is `None` only for root of the trie.
  parent: Option<B256>,
  /// Path from parent node to this node.
  parent_path: Vec<u8>,
}

pub struct TrieWalker {
  nodes: BrownHashMap<B256, TrieWalkerNode>,
}
```

This would remove the need for `ParentWithPath` and I think this would also remove ambiguity about naming.

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,159 @@
+use alloy_primitives::{Bytes, B256};
+use anyhow::ensure;
+use eth_trie::{decode_node, node::Node};
+use hashbrown::HashMap as BrownHashMap;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct AccountProof {
```
morph-dev, line 8: nit: Might be more correct to call this `TrieProof`.

morph-dev (approved):

portal-bridge/src/types/state/trie_walker.rs
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    pub fn get_proof(&self, node_hash: B256) -> AccountProof {
+        // Build path and proof from the node to the root and reverse it at the end.
+        let mut reverse_path = vec![];
+        let mut reverse_proof = vec![];
+        let mut next_node: Option<B256> = Some(node_hash);
+        while let Some(current_node) = next_node {
+            let Some(parent) = self.nodes.get(&current_node) else {
+                break;
```
morph-dev, line 145: nit: I believe that this should be consider an error now (because even root is stored in the `self.nodes`. I would `bail!`/`panic!` here, but if you think it's too much, we should at least print error.

morph-dev (commented):
Found few very small readability improvements. If you plan to refactor this in the future, you can leave it until them.

KolbyML:
Sounds good I will look into the feedback when I clean up my database PR and prepare it for review

PR #1320, opened 2024-06-17
feat(beacon-network): add content validation

ogenev:
### What was wrong?
We don't validate the beacon content type before storing and gossiping.

### How was it fixed?
- Add validations for the beacon content types.
- Extract a `test_utils` module for common usage in tests.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8384,6 +8384,7 @@ name = "trin-beacon"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "discv5",
  "ethereum_ssz",
  "ethportal-api",
--- a/ethportal-api/src/types/content_value/beacon.rs
+++ b/ethportal-api/src/types/content_value/beacon.rs
@@ -92,6 +92,15 @@ impl ForkVersionedLightClientBootstrap {
             bootstrap: light_client_bootstrap,
         })
     }
+
+    /// Get the slot of the `LightClientBootstrap`
+    pub fn get_slot(&self) -> u64 {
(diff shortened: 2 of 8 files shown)


KolbyML (approved):
:shipit:

pipermerriam:
Reading through the validation code here, I wanted to ask what the status of our hive tests is for validating these objects.  It seems like with as many validation checks as there are it would be easy for a client to miss or incorrectly implement the validation logic.  If we don't have comprehensive tests vectors in place for both valid and invalid payloads, that seems like a place where effort should go.  At minimum, getting issues written in the specs repo that detail the tests that need to be written.

KolbyML:
> Reading through the validation code here, I wanted to ask what the status of our hive tests is for validating these objects. It seems like with as many validation checks as there are it would be easy for a client to miss or incorrectly implement the validation logic. If we don't have comprehensive tests vectors in place for both valid and invalid payloads, that seems like a place where effort should go. At minimum, getting issues written in the specs repo that detail the tests that need to be written.

Currently there is no negatives, and we don't really test for positives either

ogenev:
> Reading through the validation code here, I wanted to ask what the status of our hive tests is for validating these objects. It seems like with as many validation checks as there are it would be easy for a client to miss or incorrectly implement the validation logic. If we don't have comprehensive tests vectors in place for both valid and invalid payloads, that seems like a place where effort should go. At minimum, getting issues written in the specs repo that detail the tests that need to be written.

Yeah, we will work on adding more tests for testing content validations for history and beacon. We are currently trying to fix the failing beacon hive tests for Trin.

PR #1324, opened 2024-06-19
feat(trin execution): added support for dao fork + evm tracer

KolbyML:
### What was wrong?

Our state execution didn't support the DAO fork. Which caused us to generate an incorrect state root and the client to crash

Our code wasn't setup to easily provide traces to debug with

### How was it fixed?

By draining all the specified accounts and moving the balances to the specified account as specified in EIP-779

also by adding a cli config for tracing blocks

### DIFF
--- a/trin-execution/src/config.rs
+++ b/trin-execution/src/config.rs
@@ -1,8 +1,11 @@
+use crate::{cli::TrinExecutionConfig, types::block_to_trace::BlockToTrace};
+
 #[derive(Debug, Clone)]
 pub struct StateConfig {
     /// This flag when enabled will storage all the trie keys from block execution in a cache, this
     /// is needed for gossiping the storage trie's changes.
     pub cache_contract_storage_changes: bool,
+    pub block_to_trace: BlockToTrace,
 }
 
 #[allow(clippy::derivable_impls)]
@@ -10,6 +13,16 @@ impl Default for StateConfig {
     fn default() -> Self {
         Self {
             cache_contract_storage_changes: false,
+            block_to_trace: BlockToTrace::None,
+        }
+    }
+}
+
+impl From<TrinExecutionConfig> for StateConfig {
+    fn from(cli_config: TrinExecutionConfig) -> Self {
+        Self {
+            cache_contract_storage_changes: false,
+            block_to_trace: cli_config.block_to_trace,
         }
     }
 }
--- a/trin-execution/src/execution.rs
+++ b/trin-execution/src/execution.rs
@@ -1,5 +1,5 @@
 use alloy_primitives::{keccak256, Address, Bytes, B256, U256};
-use alloy_rlp::{Decodable, EMPTY_STRING_CODE};
+use alloy_rlp::Decodable;
 use anyhow::{anyhow, bail, ensure, Error};
 use e2store::era1::BlockTuple;
 use eth_trie::{RootWithTrieDiff, Trie};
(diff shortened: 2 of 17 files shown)


trin-execution/src/execution.rs
```diff
@@ -213,6 +221,61 @@ impl State {
             }
         }
 
+        // check if dao fork, if it is drain accounts and transfer it to beneficiary
+        if block_tuple.header.header.number == get_spec_block_number(SpecId::DAO_FORK) {
+            let mut drained_balance_sum = U256::ZERO;
```
morph-dev, line 226: This should be in `dao_fork.rs`. I think you will just have to pass `&mut self.database`, which is fine.

trin-execution/src/execution.rs
```diff
@@ -255,11 +319,32 @@ impl State {
...
                     Transaction::Blob(tx) => tx.modify(block_number, tx_env),
                 }
-            })
-            .with_external_context(NoOpInspector)
-            .append_handler_register(inspector_handle_register)
-            .build()
-            .transact()?)
+            });
+
+        Ok(
```
morph-dev, line 324: I think something like this is much nicer:
```rust
let mut evm_builder = Evm::builder()
    ...; // part from above

if self.config.block_to_trace.should_trace(block_number) {
  ...
  evm_builder = evm_builder.with_external_context(
      TracerEip3155::new(Box::new(output_file)));
} else {
  evm_builder = evm_builder.with_external_context(NoOpInspector);
}

Ok(evm_builder
    .append_handler_register(inspector_handle_register)
    .build()
    .transact()?)
```

trin-execution/src/types/block_to_trace.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl FromStr for BlockToTrace {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "none" => Ok(BlockToTrace::None),
+            "all" => Ok(BlockToTrace::All),
+            val => {
+                let index = val
```
morph-dev, line 29: Use [`str.split_once()`](https://doc.rust-lang.org/std/primitive.str.html#method.split_once). For example:

```rust
match s.split_once(':') {
  Some(("block", value)) => {
      let block_number = value
          .parse()
          .map_err(|err| format!("Invalid block number: {err}"))?;
      Ok(BlockToTrace::Block(block_number))
  },
  _ => Err(format!("Invalid block trace argument: {s}"))
}
```

morph-dev (approved):
Only few small comments
:rocket:

trin-execution/src/execution.rs
KolbyML: I tried earlier, but I get this 
image

KolbyML: I tried your example quite a few things and I am not sure this example or idea is possible. I originally tried your example when I first tried implementing this a few days ago, but I couldn't get it to compile

trin-execution/src/config.rs
```diff
@@ -1,15 +1,28 @@
...
 impl Default for StateConfig {
     fn default() -> Self {
         Self {
             cache_contract_storage_changes: false,
+            block_to_trace: BlockToTrace::None,
+        }
+    }
+}
+
+impl From<TrinExecutionConfig> for StateConfig {
```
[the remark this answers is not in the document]
KolbyML: We can change it in the future if it makes sense, but for now we don't need a reference so I will leave it for now

PR #1325, opened 2024-06-20
chore: update rust to 1.79.0 and add rust-toolchain file

ogenev:
### What was wrong?
Trin was unable to run on the latest stable 1.79.0 Rust version

### How was it fixed?
- Update the `stunclient` library which caused Trin to crash.
- Fix all clippy warnings
- Add `rust-toolchain.toml` file

### To-Do


KolbyML (approved):
:shipit:

.circleci/config.yml
```diff
@@ -143,7 +143,7 @@ jobs:
           command: apt install -y clang
       - run:
           name: Install rustfmt
-          command: rustup component add rustfmt     
+          command: rustup component add --toolchain nightly rustfmt
```
carver, line 146: Won't this change mean that circle will start failing at every new nightly change to rustfmt? I prefer to fix the formatting at the rust version bumps, like this PR does, rather than do it on the nightly release schedule.

carver (approved):
LGTM!

I have a moderate preference for the stable version of rustfmt.

.circleci/config.yml
ogenev: We use nightly channel only for rustfmt because of the [imports_granularity](https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=group#imports_granularity) formatter via our [config file](https://github.com/ethereum/trin/blob/a998c8500bebee7402d0da31aeba4ba4ac835f24/rustfmt.toml).

Unfortunately, this formatter is available only for nightly (probably soon on stable).  

We need to specify the toolchain now, because when we add the `rust-toolchain`  file, rustup implicitly assumes that we want to add rustfmt on the stable channel.

PR #1331, opened 2024-06-27
feat: implement saving execution position + era_manager + gas used validation

KolbyML:
### What was wrong?
- when we stopped or trin-execution crashed we would need to restart from 0
- If a download of a era file failed, it would just panic instead of retrying getting the era file
- we didn't verify gas used, which means we could commit invalid data to the database, which means we would corrupt our database on invalid state execution
- we were running trin-execution from a test which wasn't sustainable
### How was it fixed?
- I implemented `ExecutionPosition` which saves our position on every commitment, and is used to initial where to start on starting trin-execution
- I implemented `EraManager` a "Smart" object which will fetch us any block we request by number, it also has log_2 sleep retries which is very nice.
- Normally gas used verification is done at the end of block execution, but because we execute from era1 files which contain receipts. We can use the receipts which contain the incremental cumulative_gas_used to verify exactly which transaction failed executing so our database doesn't get corrupted with bad data
- I made trin-execution so it could be ran as a standalone application, this will allow us to better build out trin-execution, as running it from a test was just not cutting it anymore.

I am so hyped about this PR because now I can focus on fixing bugs, instead of waiting forever to resync every time


trin-execution/src/constants.rs
```diff
@@ -0,0 +1,2 @@
+// The location in the database which describes the current execution position.
+pub const CURRENT_EXECUTION_POSITION: &[u8; 26] = b"current_execution_position";
```
morph-dev, line 2: nit: I think this should be better suited in `trin-execution/src/storage/db_key.rs` or something like that. I also don't think that `CURRENT_` is needed in the name, as we are storing only one `EXECUTION_POSITION`.

trin-execution/src/execution.rs
```diff
@@ -144,13 +149,34 @@ impl State {
...
+        let mut cumulative_gas_used = U256::ZERO;
+
         let transactions = block_tuple
             .body
             .body
             .transactions()
             .map_err(|err| Error::msg(format!("Error getting transactions: {err:?}")))?;
-        for transaction in transactions {
-            self.execute_transaction(transaction, &env)?;
+        for (index, transaction) in transactions.into_iter().enumerate() {
```
morph-dev, line 159: If I'm not mistaken, we are updating and committing all data after every transaction, but not the main account trie.
This is done later at line 212 when we call `self.get_root_with_trie_diff()`. Or am I missing something?

Now, It might be easier to commit all trie changes after every transaction as well, but I don't think we should do that (I expect that it will slow things down a lot).
Not sure how easy it is, but I believe that we should execute and verify entire block (keep all changes in memory) and then commit all changes (also implying that `ExecutionPosition` shouldn't track `transaction_index`, or at least not store it).

morph-dev (changes requested):
Either I'm missing something, or this is a bug regarding storing the state (account trie is not stored after every transaction). See relevant comment.

Others are minor things.

trin-execution/src/execution.rs
KolbyML: > keep all changes in memory 

this would be implementing a block execution cache and is out of scope of this PR.

this is more of a performance improvement then anything else

e2store/src/era1.rs
```diff
@@ -490,4 +490,13 @@ mod tests {
         let era1_raw_bytes = fs::read(path).unwrap();
         let _block_tuples: Vec<BlockTuple> = Era1::iter_tuples(era1_raw_bytes).collect();
     }
+
+    #[rstest::rstest]
+    #[case("../test_assets/era1/mainnet-00000-5ec1ffb8.era1", 0)]
+    #[case("../test_assets/era1/mainnet-00001-a5364e9a.era1", 8192)]
+    #[case("../test_assets/era1/mainnet-00010-5f5d4516.era1", 81920)]
+    fn test_era1_index(#[case] path: &str, #[case] index: u64) {
```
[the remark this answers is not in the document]
KolbyML: the last pre-merge era1 file is too big to be included in these tests

trin-execution/src/era_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+        let era1_files = get_shuffled_era1_files(&http_client).await?;
+        Ok(Self {
+            current_era: None,
+            http_client,
+            era1_files,
+        })
+    }
+
+    pub async fn get_block_by_number(&mut self, block_number: u64) -> anyhow::Result<BlockTuple> {
+        if self.current_era.is_none() {
```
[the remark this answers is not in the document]
KolbyML: I did everything but `era1_epoch_number`

trin-execution/src/execution.rs
KolbyML: I do think this is or something like it is required eventually, but were are not there yet

KolbyML:
@morph-dev ready for another look, let me know as always if I missed anything

trin-execution/src/execution.rs
```diff
@@ -42,34 +42,38 @@ struct GenesisConfig {
...
 impl State {
-    pub fn new(path: Option<PathBuf>, config: StateConfig) -> Self {
-        State {
-            next_block_number: 0,
-            config: config.clone(),
-            database: EvmDB::new(path, config).expect("Failed to create EVM database"),
-        }
+    pub fn new(path: Option<PathBuf>, config: StateConfig) -> anyhow::Result<Self> {
+        let (database, execution_position) =
+            EvmDB::new(path, config.clone()).expect("Failed to create EVM database");
```
morph-dev, line 54: It's weird to have `EvmDB::new` return `execution_position` as well.
I would initialize / load `execution_position` here and pass it as an argument to `EvmDB::new`.

You will probably have to  do the same for the DB as well, but I would say it's better that way (why would EvmDB initialize db connection).

trin-execution/src/era_manager.rs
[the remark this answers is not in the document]
KolbyML: I don't think so, e2store is meant to be a simple encode/decode crate for e2store formats, era_manager is meant to  manage era1/era files for executing them in trin-execution.

era_manager is trin-execution specific, e2store can be used by other projects

trin-execution/src/storage/execution_position.rs
```diff
@@ -0,0 +1,83 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, RlpDecodable, RlpEncodable)]
+pub struct ExecutionPosition {
+    /// Version of the current execution state struct
+    version: u8,
+
+    /// The block number we are currently executing
+    block_execution_number: u64,
+
+    account_trie_root: B256,
```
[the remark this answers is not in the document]
KolbyML: When we are syncing from a consensus client at HEAD - 8192, I am not sure if we will have block number to hash look ups available so I will keep it for now

e2store/src/era1.rs
```diff
@@ -118,6 +118,10 @@ impl Era1 {
         file.write(&mut buf)?;
         Ok(buf)
     }
+
+    pub fn epoch_number(&self) -> u64 {
+        self.block_index.block_index.starting_number / 8192
```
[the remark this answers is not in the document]
KolbyML: I will do 1 and 2, but leave `self.block_tuples[block_number % EPOCH_SIZE]` for now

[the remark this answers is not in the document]
KolbyML: I don't think these make sense in Era1, eventually we will handle era files as well, but we can do this for now then see what happens when we get there

trin-execution/src/execution.rs
KolbyML: https://github.com/ethereum/trin/issues/1333

Me and milos talked on a call what was outlined is basically required in the future, after the call we hit performance bottle necks so I will probably look into this after a metrics PR

KolbyML:
@morph-dev ready for another look. I say this every time but if I missed something don't be afraid to let me know

trin-execution/src/era_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+            current_era: None,
+            http_client,
+            era1_files,
+        })
+    }
+
+    pub async fn get_block_by_number(&mut self, block_number: u64) -> anyhow::Result<&BlockTuple> {
+        let era1 = match &self.current_era {
+            Some(era1)
+                if era1.epoch_number() == Era1::epoch_number_from_block_number(block_number) =>
```
morph-dev, line 34: nit: you can extract `Era1::epoch_number_from_block_number(block_number)` into separate variable at the beginning of function so you don't have to repeat it here and below, and it will be easier to read.

trin-execution/src/era_manager.rs
```diff
@@ -0,0 +1,69 @@
...
+            {
+                self.current_era.as_ref().expect("current_era to be some")
+            }
+            _ => self.current_era.insert(
+                self.download_era1_file(Era1::epoch_number_from_block_number(block_number))
+                    .await?,
+            ),
+        };
+        Ok(&era1.block_tuples[(block_number % 8192) as usize])
```
morph-dev, line 43: nit: If you don't want to move `block_number % 8192` into `era1.rs` as function, at least use constant (8192) from there.

e2store/src/era1.rs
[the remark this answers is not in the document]
morph-dev: I think it makes sense because that's the place that **should** know how many blocks are in each era1 file, and in which file and at what index the specific block is. Other crates shouldn't have to implement that calculation (even tho it's simple).

Maybe not that specific file, but definitely somewhere in that crate.

morph-dev (approved):

PR #1335, opened 2024-06-30
feat: add metrics for Trin Execution

KolbyML:
### What was wrong?
I didn't know what was causing performance issues
### How was it fixed?
by adding basic metrics

I also added a `metrics` folder which contains a `docker-compose` which will automatically setup the grafana dashboard up

This PR lays the foundation we can build off as we need into the future

image
here is a picture of it, we can improve it as we go and as we find we need more


KolbyML:
I thought about adding more metrics in this PR, but I think this lays the foundation I have lots of idea's though
- client version
- block height
- transaction processing (more in depth then block processing provides)

etc

trin-execution/README.md
morph-dev, line 1: nit: Maybe add link to the `trin-execution/metrics/README.md`

trin-execution/src/execution.rs
```diff
@@ -192,9 +201,12 @@ impl State {
             self.execution_position
                 .increase_transaction_index(self.database.db.clone())?;
             self.database.commit(evm_result.state);
```
morph-dev, line 203: nit: I think it would be useful to have separate timer just for this step (`self.database.commit(_)`) as I think most of the time is spend there.

morph-dev (approved):
Looks good. I didn't really check any `.yaml` files as I'm not familiar with prometheus/grafana configurations (and to my understanding they are automatically exported).

trin-execution/src/execution.rs
KolbyML: This will be addressed in the next PR which will include a graph specific to transaction execution performance

PR #1336, opened 2024-07-02
refactor: make storage generic over content key trait

njgheorghita:
### What was wrong?
Preliminary work for gossiping dropped content. Spoke with @morph-dev offline and decided this was a good step, since I will soon update `prune()` to return content keys. Rather than having 4 separate fns on `IdIndexedStoreV1` using custom generics, better to make the entire struct generic over `OverlayContentKey`. This also helps simplify the upcoming implementation of `prune()` returning dropped content.

### To-Do

### DIFF
--- a/trin-history/src/storage.rs
+++ b/trin-history/src/storage.rs
@@ -1,6 +1,6 @@
 use ethportal_api::{
     types::{distance::Distance, history::PaginateLocalContentInfo, portal_wire::ProtocolId},
-    OverlayContentKey,
+    HistoryContentKey, OverlayContentKey,
 };
 use trin_storage::{
     error::ContentStoreError,
@@ -11,25 +11,22 @@ use trin_storage::{
 /// Storage layer for the history network. Encapsulates history network specific data and logic.
 #[derive(Debug)]
 pub struct HistoryStorage {
-    store: IdIndexedV1Store,
+    store: IdIndexedV1Store<HistoryContentKey>,
 }
 
-impl ContentStore for HistoryStorage {
-    fn get<K: OverlayContentKey>(&self, key: &K) -> Result<Option<Vec<u8>>, ContentStoreError> {
+impl<TContentKey: OverlayContentKey> ContentStore<TContentKey> for HistoryStorage {
+    fn get(&self, key: &TContentKey) -> Result<Option<Vec<u8>>, ContentStoreError> {
         self.store.lookup_content_value(&key.content_id().into())
     }
 
-    fn put<K: OverlayContentKey, V: AsRef<[u8]>>(
-        &mut self,
-        key: K,
-        value: V,
-    ) -> Result<(), ContentStoreError> {
+    fn put<V: AsRef<[u8]>>(&mut self, key: TContentKey, value: V) -> Result<(), ContentStoreError> {
+        let key = HistoryContentKey::try_from(key.to_bytes())?;
         self.store.insert(&key, value.as_ref().to_vec())
     }
 
-    fn is_key_within_radius_and_unavailable<K: OverlayContentKey>(
+    fn is_key_within_radius_and_unavailable(
         &self,
-        key: &K,
+        key: &TContentKey,
     ) -> Result<ShouldWeStoreContent, ContentStoreError> {
         let content_id = ContentId::from(key.content_id());
         if self.store.distance_to_content_id(&content_id) > self.store.radius() {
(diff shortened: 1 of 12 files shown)


trin-storage/src/lib.rs
```diff
@@ -45,22 +45,18 @@ pub enum ShouldWeStoreContent {
 }
 
 /// A data store for Portal Network content (data).
-pub trait ContentStore {
+pub trait ContentStore<TContentKey: OverlayContentKey> {
```
morph-dev, line 48: Something that I was thinking about before, and I think it would make sense to do it now (since you are refactoring anyway).

I believe that in the case of `ContentStore`, it might be better to use [Associated Type](https://doc.rust-lang.org/rust-by-example/generics/assoc_items/types.html), not the generic type. That way, the actual implementations (`HistoryStorage`/`BeaconStorage`/`StateStorage`) can specify the actual type and maybe cleanup code a bit as well. What do you think?

In the end, I don't think it makes big difference either way, so whatever you decide. But I think Associated type is better in this case?

trin-history/src/storage.rs
```diff
@@ -107,7 +104,10 @@ pub mod test {
                 PortalStorageConfig::new(CAPACITY_MB, temp_dir.path().to_path_buf(), node_id)
                     .unwrap();
             let mut storage = HistoryStorage::new(storage_config).unwrap();
-            let content_key = IdentityContentKey::random();
+            let prefixes = [0u8, 1u8, 2u8, 3u8];
+            let mut raw_content_key: Vec<u8> = IdentityContentKey::random().to_bytes();
+            raw_content_key.insert(0, *prefixes.choose(&mut rand::thread_rng()).unwrap());
+            let content_key = HistoryContentKey::try_from(raw_content_key).unwrap();
```
morph-dev, line 110: nit: It is weird use bytes from random `IdentityContentKey` to create random `HistoryContentKey`.

I would say just just pick random prefix (as you do now) and then random 32 bytes. Maybe even implement this directly in `ethportal-api/src/types/content_key/history.rs`.

morph-dev (approved):
I would say the only important comment is the usage of associated type for the `ContentStore`. Ultimately, whatever you decide, but I think it would be better that way. Happy to discuss if you want.

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -85,13 +89,14 @@ impl VersionedContentStore for IdIndexedV1Store {
             pruning_strategy,
             usage_stats: UsageStats::default(),
             metrics: StorageMetricsReporter::new(protocol_id),
+            phantom_content_key: PhantomData,
```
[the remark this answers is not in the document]
njgheorghita: Do you mean you've seen that in many places other than this codebase? I saw that `_` prefix being used a fair amount in other places, but inside the codebase we've dropped the `_` prefix, so I just followed suit for consistency. But now it seems (based off your comment and what I've seen online) that the `_` prefix is convention, so I'll go ahead and make all the changes here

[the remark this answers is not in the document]
morph-dev: Yeah, in other casebases, as a general convention. I wasn't aware that we dropped it in our codebase.
I'm fine with whatever you decide (using `_` because of the general convention, or leaving it as it is for consistency within codebase).

trin-storage/src/lib.rs
njgheorghita: I gave this a go since it does seem cleaner, but I'm not sure if it's possible. Since `OverlayService` & `OverlayProtocol` are generic over the type of storage (`TStore`) and they use the `ContentStore` trait functions directly, then they need knowledge of this associated type. There's no way (from what I could find) to define the associated type on `ContentStore` as `impl OverlayContentKey`. So it seems to me like we're only able to define it with generics. Maybe I was doing something wrong/misunderstood how to implement this?

morph-dev: Obviously, I'm not sure until I try it, but I believe that it would work the same way as with generics.
Meaning, if you named associated type `Key`, then instead of having: `ContentStore<TContentKey>`, you would have something like `ContentStore<Key=TContentKey>`. But I'm not sure if this is allowed because `TContentKey` is another generic...

Either way, feel free to leave it your way, and I will give it a try afterwards (if you agree that associated type is better).

njgheorghita: Ahh nice, yeah cool that syntax did the trick!

PR #1341, opened 2024-07-15
Gossip pruned

### LINKED ISSUE #981: Trin should gossip content before it drops it from storage
A possible enhancement came to mind during a recent chat with @carver that it would be worthwhile to gossip content as it gets dropped from storage. This won't help in situations when the network is over capacity, but in all other situations it should help maintain data availability in the network, as new nodes might have recently come online that would be interested in the data.

njgheorghita:
### What was wrong?
fixes #981 

### How was it fixed?
- add content that's dropped from local storage to propagation content vector
- limit offer message in propagation to 64 content keys (according to spec)

### To-Do

### DIFF
--- a/portalnet/src/gossip.rs
+++ b/portalnet/src/gossip.rs
@@ -27,7 +27,7 @@ use ethportal_api::{
         portal_wire::{PopulatedOffer, PopulatedOfferWithResult, Request, Response},
     },
     utils::bytes::{hex_encode, hex_encode_compact},
-    OverlayContentKey, RawContentKey,
+    OverlayContentKey,
 };
 
 /// Datatype to store the result of a gossip request.
@@ -76,19 +76,18 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
 
     // HashMap to temporarily store all interested ENRs and the content.
     // Key is base64 string of node's ENR.
-    let mut enrs_and_content: HashMap<String, Vec<(RawContentKey, Vec<u8>)>> = HashMap::new();
+    let mut enrs_and_content: HashMap<String, Vec<(TContentKey, Vec<u8>)>> = HashMap::new();
 
     for (content_key, content_value) in content {
         let interested_enrs = calculate_interested_enrs(&content_key, &all_nodes);
 
         // Temporarily store all randomly selected nodes with the content of interest.
         // We want this so we can offer all the content to an interested node in one request.
-        let raw_item = (content_key.into(), content_value);
         for enr in interested_enrs {
             enrs_and_content
                 .entry(enr.to_base64())
                 .or_default()
-                .push(raw_item.clone());
+                .push((content_key.clone(), content_value.clone()));
         }
     }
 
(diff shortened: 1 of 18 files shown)


KolbyML (approved):
:shipit:

trin-state/src/storage.rs
```diff
@@ -124,16 +128,22 @@ impl StateStorage {
         let trie_node = TrieNode {
             node: last_trie_node.clone(),
         };
-        self.store
+        match self
```
morph-dev, line 131: I think this can be simplified to:
`self.store.insert(...).and(Ok(vec![]))`

Optionally and in addition, you can let these functions return dropped content and handle it once in a `put` function above.

trin-beacon/src/storage.rs
```diff
@@ -223,8 +223,16 @@ impl ContentStore for BeaconStorage {
...
     }
 
-    fn put<V: AsRef<[u8]>>(&mut self, key: Self::Key, value: V) -> Result<(), ContentStoreError> {
-        self.store(&key, &value.as_ref().to_vec())
+    fn put<V: AsRef<[u8]>>(
+        &mut self,
+        key: Self::Key,
+        value: V,
+    ) -> Result<Vec<(Self::Key, Vec<u8>)>, ContentStoreError> {
+        match self.store(&key, &value.as_ref().to_vec()) {
```
morph-dev, line 231: Same as state store, this could be:
`self.store(...).and(Ok(vec![]))`.

Or if you prefer this:
`self.store(...).map(|_| vec![])`.

trin-storage/src/lib.rs
```diff
@@ -52,7 +52,15 @@ pub trait ContentStore {
     fn get(&self, key: &Self::Key) -> Result<Option<Vec<u8>>, ContentStoreError>;
 
     /// Puts a piece of content into the store.
-    fn put<V: AsRef<[u8]>>(&mut self, key: Self::Key, value: V) -> Result<(), ContentStoreError>;
+    /// Returns a list of keys that were evicted from the store, which are gossiped into the
```
morph-dev, line 55: nit: maybe this
```suggestion
    /// Returns a list of keys that were evicted from the store, which should be gossiped into the
```

trin-storage/src/versioned/id_indexed_v1/sql.rs
```diff
@@ -75,7 +75,7 @@ pub fn delete_farthest(content_type: &ContentType) -> String {
             ORDER BY distance_short DESC
             LIMIT :limit
         )
-        RETURNING content_size",
+        RETURNING content_key, content_value",
```
morph-dev, line 78: nit: I would still return `content_size` as well. This should save us some calculations and make sure we are using the right value.

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -241,7 +242,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
         &mut self,
         content_key: &TContentKey,
         content_value: Vec<u8>,
-    ) -> Result<(), ContentStoreError> {
+    ) -> Result<Vec<(TContentKey, Vec<u8>)>, ContentStoreError> {
```
morph-dev, line 245: nit: Update function documentation.

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -275,12 +276,13 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
         self.usage_stats.total_entry_size_bytes += content_size as u64;
         self.usage_stats.report_metrics(&self.metrics);
 
+        let mut dropped_content: Vec<(TContentKey, Vec<u8>)> = Vec::new();
```
morph-dev, line 279: nit: I don't think that you need to declare type here, so simple `let mut dropped_content = vec![];` should be enough.

Also, you can do something like this:
```rust
let mut dropped_content = if self.pruning_strategy.should_prune(&self.usage_stats) {
    self.prune()?
} else {
    vec![]
}
```

morph-dev (commented):
Nicely done. Couple of suggestion, nothing major (except maybe calculating distance correctly).

morph-dev (approved):

portalnet/src/gossip.rs
```diff
@@ -103,6 +111,31 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
...
+        // offer messages are limited to 64 content keys
+        if interested_content.len() > 64 {
+            warn!(
+                enr = %enr,
+                content.len = interested_content.len(),
+                "Too many content items to offer to a single peer, dropping {}.",
+                interested_content.len() - 64
+            );
+            // sort content keys by distance to the node
+            interested_content.sort_by(|a, b| {
```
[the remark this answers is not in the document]
njgheorghita: Ahh, yup my mistake, we should be calculating on the id.

PR #1342, opened 2024-07-17
Test gossip propagation

njgheorghita:
### What was wrong?
Did some digging while debugging #1341 - but looks like our code is working as expected 🥳 Though I feel like it's useful to keep this test in for sanity.

Also added a check to our `OFFER` functionality to make sure that we can only offer content that's present in the local datastore for unpopulated offers. 

### How was it fixed?
- added tests
- added check to make sure that offered data is available locally 

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/offer_accept.rs
+++ b/ethportal-peertest/src/scenarios/offer_accept.rs
@@ -7,7 +7,7 @@ use tokio::time::{sleep, Duration};
 use tracing::info;
 
 use crate::{
-    utils::{fixture_block_body, fixture_header_with_proof, wait_for_history_content},
+    utils::{fixture_header_with_proof, wait_for_history_content},
     Peertest,
 };
 use ethportal_api::{
@@ -51,6 +51,32 @@ pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
     );
 }
 
+pub async fn test_unpopulated_offer_fails_with_missing_content(
+    peertest: &Peertest,
+    target: &Client,
+) {
+    info!("Testing Unpopulated OFFER/ACCEPT flow with missing content");
+
+    let (content_key, _content_value) = fixture_header_with_proof();
+
+    // validate that unpopulated offer fails if content not available locally
+    match target
+        .offer(
+            Enr::from_str(&peertest.bootnode.enr.to_base64()).unwrap(),
+            content_key.clone(),
+            None,
+        )
+        .await
+    {
+        Ok(_) => panic!("Unpopulated offer should have failed"),
+        Err(e) => {
+            assert!(e
+                .to_string()
+                .contains("Content key not found in local store"));
+        }
+    }
+}
+
 pub async fn test_populated_offer(peertest: &Peertest, target: &Client) {
     info!("Testing Populated Offer/ACCEPT flow");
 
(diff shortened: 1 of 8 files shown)


rpc/src/history_rpc.rs
```diff
@@ -170,6 +170,9 @@ impl HistoryNetworkApiServer for HistoryNetworkApi {
     }
 
     /// Send an OFFER request with given ContentKey, to the designated peer and wait for a response.
+    /// If the content value is provided, a "populated" offer is used, which will not store the
```
morph-dev, line 173: Should we add this to other `*_rpc.rs` files?

Related to this, maybe state shouldn't supported "unpopulated" version.

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -45,6 +51,25 @@ pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
...
+
+    // validate that unpopulated offer fails if content not available locally
+    assert!(target
+        .offer(
+            Enr::from_str(&peertest.bootnode.enr.to_base64()).unwrap(),
+            content_key.clone(),
+            None,
+        )
+        .await
+        .is_err());
```
morph-dev, line 70: Would be worth checking the error itself, because it could fail for some other reason, e.g. wrong test configuration, not being connected/able to talk to the target node, etc.

portalnet/src/overlay/protocol.rs
```diff
@@ -531,6 +531,32 @@ where
...
+            let content_key = TContentKey::try_from(content_key.clone()).map_err(|err| {
+                OverlayRequestError::ContentNotFound {
+                    message: format!(
+                        "Error decoding content key for content key: {content_key:02X?} - {err}"
+                    ),
+                    utp: false,
+                    trace: None,
+                }
+            })?;
+            match self.store.read().get(&content_key) {
```
morph-dev, line 546: Question more for my understanding: considering that now we are getting the value from the DB before the OFFER request anyway, why not use `Request::PopulatedOffer` instead?

And if we switch to `Request::PopulatedOffer`, do we ever use `Request::Offer`? And if not, can we remove it?

morph-dev (approved):
LGTM.

Few non-blocking comments, more for discussion purposes.

rpc/src/history_rpc.rs
njgheorghita: Maybe this is true for beacon network as well?

portalnet/src/overlay/protocol.rs
morph-dev: I wasn't talking about JSON rpc endpoint, as much as I was talking about `Request` object and how we handle it internally.

For example, the json rpc handler can retrieve the content value from the local storage, and call `network.overlay.send_populated_offer(..)` instead. In that case, I believe we don't need `Request::Offer` type at all (but still support json offer endpoint).

If you are still not sure what I mean, we can have a look at the code together. It's very possible that I'm not understanding how some things work.

njgheorghita: Ahh, ok yeah that's interesting. Off the top of my head that seems feasible, but will probably need to do a little tinkering to confirm and make sure there aren't any edge cases that wouldn't be supported. Not super high-priority imo but I put it on my flamingo list of todos

PR #1344, opened 2024-07-22
feat: implement .era2 (state snapshots) import/export from Trin Execution

KolbyML:
### What was wrong?
we have `.era2` types in `e2store`, but we don't support importing or exporting these files using `trin-execution`
### How was it fixed?
By implementing a minimal trie walking implementing which provides an api where you can loop over the leafs of a trie.

Then we implement
- exporting to era2 
- importing from era2


morph-dev:
I know this is still in progress and maybe you already had this in mind, but this pr is huge! Please split it up in smaller parts before sending for review.

I assume some work is already done, even if not final, and can already be reviewed. For example new structure/type definitions and helping structures/functions around them (without introducing much new logic with the existing code, if not needed).

KolbyML:
> I know this is still in progress and maybe you already had this in mind, but this pr is huge! Please split it up in smaller parts before sending for review.
> 
> I assume some work is already done, even if not final, and can already be reviewed. For example new structure/type definitions and helping structures/functions around them (without introducing much new logic with the existing code, if not needed).

It is hard to take this comment seriously https://github.com/ethereum/trin/pull/1344#issuecomment-2267365857 when if it should be said at all, it should be said on this PR https://github.com/ethereum/trin/pull/1337

Realistically this work can be broken up into 3 PR's, and it is already in 2, the 3rd one would be a PR for fixing sender recovery

I am not going to open a PR for review until I make a dump of state at block 15 million
- I have fixed a few bugs well executing the code
- It doesn't make sense to review any of this code yet, for one it is currently being tested, 2 it risks what I am executing.

I am sure not everyone was aware this PR was already split up into 2 PRs, these PR's are drafts for a reason, they are not ready to be reviewed.

morph-dev:
My comment wasn't specifically tied to this PR, and definitely not to this PR at the current stage (still in progress), but in general.
It was a recommendation / suggestion to try to split big PRs into smaller once before sending for review.

I apologize if you were planning on doing this anyway and my feedback was coming on not desired time (not being send for review), but I wanted to provide early feedback and avoid delays later on (I find it easier to split work while I'm working on it, not when everything is done, but that's just my personal workflow).
Also, if certain work is more or less complete (e.g. defining era2 types), then it can we extracted into separate PR and be reviewed already (if you want with a note that it's not a final version), and reduce time for future reviews. But it's up to you to estimate how complete each part is and whether they can be extracted.

I commented on this PR because you updated it recently. And I acknowledge that you already split it up before, but I believe that they should be split into even smaller PRs. And again, I apologize if this is something that you were planning on doing anyway

KolbyML:
> Also, if certain work is more or less complete (e.g. defining era2 types), then it can we extracted into separate PR and be reviewed already (if you want with a note that it's not a final version), 

I don't think this is logical to try to merge .era2 without import or export, how do we know if .era2 is even useful without that. Making a PR for unused types is redundant, I don't find value in that. I think there is a lot of value in being able to export an .era2 file and import it from the PR adding .era2.

> reduce time for future reviews. But it's up to you to estimate how complete each part is and whether they can be extracted.

The main delay for me "opening these PR's for review", is I am currently executing blocks using this code. I am not really concerned with `reduce time for future reviews`. I am concerned with not accidentally running a change suggested in a PR which corrupts my state database for xyz reason. It is easier and less risky to get the state snapshot then open the PR's for review.

I am not really able to test .era2 until I am done executing, I have brought this up in meetings.

pipermerriam:
I believe what @morph-dev has a reasonable point here.

There is an often cited [study](https://static1.smartbear.co/support/media/resources/cc/book/code-review-cisco-case-study.pdf) that showed that code review effectiveness drops off significantly after 400 lines of code and is most effective when kept under 200 lines of code.

These shouldn't be strict rules of course.

Which is probably why the suggestion was made to extract any pieces of this (like the types) into a standalone PR.  You are correct that the types aren't much use on their own.  However, the act of splitting them into their own PR has value on it's own because that PR is easy to review AND it reduces the size of the final PR which means that the other members of the team will both have an easier time reviewing this code and the review they give will be higher quality.

KolbyML:
@morph-dev 
Ok I made these PR's which are ready for review based off the 2 I already have
- more metrics https://github.com/ethereum/trin/pull/1355
- fixing sender recovery https://github.com/ethereum/trin/pull/1356
- add era pre-(fetching/processing) https://github.com/ethereum/trin/pull/1357

I am fine to split .era2 types and import/export into 2 different PR's, but I can't do that in good faith till it is battle tested around 4-5 days from now.

The 3 PR's I made above don't make any changes to the database code so they are fine to be reviewed right now

carver:
I will send this back from "in review" to "in progress" since it's got some conflicts to resolve

KolbyML:
@morph-dev this PR is ready for review

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,346 @@
...
+        partial_key_path.pop();
+        let full_key_path = [&path, partial_key_path.as_slice()].concat();
+        let address_hash = full_nibble_path_to_address_hash(&full_key_path);
+        self.leafs_to_process.push(LeafNodeWithKeyHash {
+            key_hash: address_hash,
+            value: leaf.value.clone(),
+        });
+    }
+
+    fn process_node(&mut self, node: Node, path: Vec<u8>) -> anyhow::Result<()> {
```
morph-dev, line 72: I think that era2 files must be canonical (meaning, there is only one valid version of it). To achieve that, all accounts should be sorted by account_hash (and similarly, storage entries should be sorted by key_hash).

This code doesn't do that. For example, Let's say this is the trie (let's assume with degree of 4):

```
B-Branch, L-Leaf, N-NULL
                           B_root
      /                /             \              \
     B1               L2              B3             N
  / / \ \                          / / \  \
L11 N  N N                        N N  L33 N
```
Desired order of leafs is `L11, L2, L33`.

If I'm not mistaken, after processing `B_root`, we would have these values:

```
stack = [B1, B3]
leafs_to_process = [L2]
```

And that would result in order of leafs being process: `L2, L33, L11`

morph-dev (changes requested):
I think that leaf iteration would not work on storage tries. Did you test this code and does it work? (e.g. on block 1_000_000)?

I also think that you can utilize some functionality from `ethportal-api/src/types/state_trie/trie_traversal.rs`. It would be even better if we can move this traversal there as well (or at least partially). I have an idea on how it can be done nicely, but I have to try it out to see if it works.

morph-dev:
After spending quite some time implementing trie traversal myself ([pr/1344 branch in my trin fork](https://github.com/ethereum/trin/commit/76bdd3598a8e762b13a391922126a2a494909d9c)), I realized that `EthTrie` already has trie iterator: https://docs.rs/eth_trie/0.4.0/eth_trie/struct.EthTrie.html#method.iter. I think we should use it (but you should check and make sure that it traverses the trie in-order)

KolbyML:
> I think that leaf iteration would not work on storage tries. Did you test this code and does it work? (e.g. on block 1_000_000)?

I both exported and imported this files at 1 million and 15 million so I can factually say this code properly handles storage tries

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,346 @@
...
+                let path_with_extension_prefix =
+                    [path, extension.prefix.get_data().to_vec()].concat();
+                match &extension.node {
+                    Node::Hash(hash) => {
+                        self.stack
+                            .push(TrieNode::new(hash.hash, path_with_extension_prefix));
+                    }
+                    Node::Leaf(leaf) => self.process_leaf(leaf.clone(), path_with_extension_prefix),
+                    _ => {
+                        panic!("Invalid extension node, must be either a leaf or a hash");
```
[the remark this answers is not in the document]
KolbyML: If this approximation was true, then why does this code work for block 15,000,000

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,346 @@
...
+            Node::Branch(branch) => {
+                let branch = branch.read().expect("Branch node must be readable");
+                for (i, child) in branch.children.iter().enumerate() {
+                    let branch_path = [path.clone(), vec![i as u8]].concat();
+                    match child {
+                        Node::Leaf(leaf) => self.process_leaf(leaf.clone(), branch_path),
+                        Node::Hash(hash) => self.stack.push(TrieNode::new(hash.hash, branch_path)),
+                        Node::Empty => {} // Do nothing
+                        _ => {
+                            panic!("Invalid branch node, must be either a leaf or a hash")
```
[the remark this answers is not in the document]
KolbyML: If this approximation was true, then why does this code work for block 15,000,000

trin-execution/src/subcommands/era2.rs
KolbyML: https://github.com/ethereum/trin/pull/1344#issuecomment-2338336517

Replying to both. I do agree this is an oversight in my implementation well it is small and can be fixed. I do agree the EthTrie iter should be used here.

At first look it traverses in order, one thing I am not sure of is if it traverses branch values properly which could be possible in the case of the storage trie.

A major complaint I have of the iter, is it hides error's

https://docs.rs/eth_trie/0.4.0/src/eth_trie/trie.rs.html#180 I think here an error should be returned. Because if this was to happen clearly a critical error has happened and the .era2 export would be corrupt so I will make a pr on ethereum/eth-trie.rs and ping you for review once that problem is resolved I will update this PR and we can continue the review from there

KolbyML: If the branch's partial nimble of the path is 0x10 which is 16, shouldn't that be read after all the branch's children? But this iter reads the value first. 

Overall this iter seems todo anything correct, so if you have an idea if branch.value should be read first or last, I think everything else is good

morph-dev: > If the branch's partial nimble of the path is 0x10 which is 16, shouldn't that be read after all the branch's children? But this iter reads the value first.

Now sure what you mean by this. Can you point me to the code line?

> Overall this iter seems todo anything correct, so if you have an idea if branch.value should be read first or last, I think everything else is good

While MPT allows for values to be stored in the branch trie node, Ethereum state trie doesn't use that feature, so I think it doesn't matter much. But I would say it makes sense to go first (which is what I think EthTrie does already, but I didn't look into details).

> https://docs.rs/eth_trie/0.4.0/src/eth_trie/trie.rs.html#180 I think here an error should be returned. Because if this was to happen clearly a critical error has happened and the .era2 export would be corrupt so I will make a pr on ethereum/eth-trie.rs and ping you for review once that problem is resolved I will update this PR and we can continue the review from there

Yes, I agree that this critical error shouldn't be ignored. I think it's better to make it iterate over `Result<_, TrieError>`, but panicking might be good enough as well.

morph-dev:
> I both exported and imported .era2 files I generated at block 1 million and 15 million so I can factually say this code properly handles storage tries

I believe that it works for your cases. After a bit of thinking, I now think that it's likely going to work on all blocks. But there is a possible case when it might not. And while chances of such case happening are very low, but it's theoretically possible.

Now that we are switching to the iteration done by EthTrie, this becomes irrelevant. I still want to document the problem, as it's important concept to understand regarding how MPT works in general.

---

The issue is the following. If encoding of a node is less than 32 bytes, its parent would inline that node directly (instead of using hash of it).

For account tries, this can't happen because:
- each leaf is rlp encoding of the account state (which includes storage_root and code_hash, making it at least 64 bytes)
- all other nodes include at least one child, meaning they are longer than 32 bytes as well

For storage trie, this is different, as values that are stored can be very short (only 1 byte). But since leaf nodes include path from parent node, in majority of cases, encoding of the entire leaf node would be at least 32 bytes (there are some extra bytes used by rlp encoding itself).

But if the tree is deep, this path can be small enough, resulting in leaf node being inlined directly into parent (instead of as hash), and if encoding of that parent is small, it can be included in its own parent, etc.

Your code handles the case when child of a branch node is inlined leaf node (so I assume you encountered this is practice), but it's technically possible for child of a branch or extension node to be inlined another branch or extension node. Chances for this happening are very-very small (because each additional node that you are inlining adds some, potentially quite a bit, extra bytes for encoding), and we might not encounter it in practice. But I would say it's important corner case that should be covered so that we are resilient regardless of the size (depth) of the trie.

trin-execution/src/subcommands/era2.rs
KolbyML: > > If the branch's partial nimble of the path is 0x10 which is 16, shouldn't that be read after all the branch's children? But this iter reads the value first.
> 
> Now sure what you mean by this. Can you point me to the code line?

```
impl TraceNode {
    fn advance(&mut self) {
        self.status = match &self.status {
            TraceStatus::Start => TraceStatus::Doing,
            TraceStatus::Doing => match self.node {
                Node::Branch(_) => TraceStatus::Child(0),
                _ => TraceStatus::End,
            },
            TraceStatus::Child(i) if *i < 15 => TraceStatus::Child(i + 1),
            _ => TraceStatus::End,
        }
    }
}
```

Anyways so for a branch node as I was saying before it will read the `value` first

morph-dev: Yes, I think this is correct.

But this is irrelevant for Ethereum State because keys are always 32 bytes (64 nibbles) long, so they are always stored in leaf nodes (values in branch nodes will be empty).

KolbyML:
> > I both exported and imported .era2 files I generated at block 1 million and 15 million so I can factually say this code properly handles storage tries
> 
> I believe that it works for your cases. After a bit of thinking, I now think that it's likely going to work on all blocks. But there is a possible case when it might not. And while chances of such case happening are very low, it's theoretically possible.
> 
> Now that we are switching to the iteration done by EthTrie, this becomes irrelevant. I still want to document the problem, as it's important concept to understand regarding how MPT works in general.
> 
> The issue is the following. If encoding of a node is less than 32 bytes, its parent would inline that node directly (instead of using hash of it).
> 
> For account tries, this can't happen because:
> 
> * each leaf is rlp encoding of the account state (which includes storage_root and code_hash, making it at least 64 bytes)
> * all other nodes include at least one child, meaning they are longer than 32 bytes as well
> 
> For storage trie, this is different, as values that are stored can be very short (only 1 byte). But since leaf nodes include path from parent node, in majority of cases, encoding of the entire leaf node would be at least 32 bytes (there are some extra bytes used by rlp encoding itself).
> 
> But if the tree is deep, this path can be small enough, resulting in leaf node being inlined directly into parent (instead of as hash), and if encoding of that parent is small, it can be included in its own parent, etc.
> 
> Your code handles the case when child of a branch node is inlined leaf node (so I assume you encountered this is practice), but it's technically possible for child of a branch or extension node to be inlined another branch or extension node. Chances for this happening are very-very small (because each additional node that you are inlining adds some, potentially quite a bit, extra bytes for encoding), and we might not encounter it in practice. But I would say it's important corner case that should be covered so that we are resilient regardless of the size (depth) of the trie.

The original question was if when I ran my code it worked in the first place which I said it did, I didn't comment on 1 byte data being compacted, because it wasn't asked or seemed probable to our usecase, but as you said `Now that we are switching to the iteration done by EthTrie, this becomes irrelevant`

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,346 @@
...
+                        .insert(storage_index_hash.as_slice(), &alloy_rlp::encode(value))?;
+                }
+            }
+
+            if account_trie.root_hash()? != account_state.storage_root {
+                return Err(Error::msg("Failed importing account storage trie: storage roots don't match expect value, .era2 import failed"));
+            }
+
+            // Insert contract if available
+            if !bytecode.is_empty() && account_state.code_hash != KECCAK_EMPTY {
```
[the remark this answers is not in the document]
KolbyML: For lines
- 297  we already check it
- 333 that wouldn't generate a valid state  root, we are just calling `get_root()` to commit the changes to the database as if we didn't ram usage would go through the roof

KolbyML:
@morph-dev ready for another review, I believe I resolved all the PR's concerns.

I also tested the changes by
- executing to 1 million
- exporting it to a .era2 file
- deleting my data dir
- importing the .era2 file I just generated
- executing to 2 million plus blocks

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,243 @@
...
+                self.trin_execution
+                    .database
+                    .db
+                    .get(account_state.code_hash)?
+                    .expect("If code hash is not empty, code must be present")
+            } else {
+                vec![]
+            };
+
+            let mut storage: Vec<StorageItem> = vec![];
```
morph-dev, line 64: nit: I don't know if this is big problem, but I don't like that we are storing all storage items in memory, instead of exporting them to file. Don't know how big storage can get and if this is an issue (at least on weaker machines).

I know that this is needed in order to write `storage_count` in the `AccountEntry` first. The possible solutions would be:

1. iterate entire trie once and count number of items, write `AccountEntry`, iterate second time and write as you go
    - considering that reading from db is most likely the slowest part, this would make export twice as slow
2. Start processing as we do now.
    1. if we finish processing entire storage trie before getting to `MAX_STORAGE_ITEMS`, do as we do now (in which case, there will be only one StorageEntry)
    2. If we process `MAX_STORAGE_ITEMS`, stop, and do some variation of 1. from above.
    - this approach gives us best performance in majority of cases (single StorageEntry), but still handles big storage tries. but it's not elegant, and questionable how much it would actually slow things down

Overall, I think current implementation is good for now, and if we notice issue with later (bigger) blocks, we can evaluate if improvement is needed. So no action is expected here (except if you want to create an issue to document this and provide your opinion).

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,243 @@
...
+                account_state,
+                bytecode,
+                storage_count,
+            } = account;
+
+            let account_state: Account = account_state.into();
+
+            // Build storage trie
+            let account_db = AccountDB::new(address_hash, self.trin_execution.database.db.clone());
+            let mut account_trie = EthTrie::new(Arc::new(account_db));
```
morph-dev, line 151: nit: this should be `storage_trie`

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,243 @@
...
+                    return Err(Error::msg("Expected storage, got account entry"));
+                };
+                for StorageItem {
+                    storage_index_hash,
+                    value,
+                } in storage_entry.0
+                {
+                    account_trie
+                        .insert(storage_index_hash.as_slice(), &alloy_rlp::encode(value))?;
+                }
```
morph-dev, line 163: Should we call `account_trie.root_hash()?` once per `StorageEntry`, in order to commit change to db and lower RAM usage?

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,243 @@
...
+            }
+
+            if account_trie.root_hash()? != account_state.storage_root {
+                return Err(Error::msg("Failed importing account storage trie: storage roots don't match expect value, .era2 import failed"));
+            }
+
+            // Insert contract if available
+            if !bytecode.is_empty() && account_state.code_hash != KECCAK_EMPTY {
+                ensure!(
```
morph-dev, line 172: nit: With this check being inside `if`, it's possible that we will accept invalid era2 file, for example: bytecode provided with this account is empty but code_hash is not (and serious attack vector if the same code_hash is hash of bytecode of some other account).

Therefore, I think it's better to move this check before `if` statement. And with this check present, we don't need both conditions in the `if` statement (`!bytes.is_empty()` is enough).

trin-execution/src/subcommands/era2.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// insert the last 256 block hashes into the database
+    pub async fn import_last_256_block_hashes(&mut self) -> anyhow::Result<()> {
+        let first_block_hash_to_add = self
+            .trin_execution
+            .next_block_number()
+            .saturating_sub(BLOCKHASH_SERVE_WINDOW);
+        let mut era_manager = EraManager::new(first_block_hash_to_add).await?;
+        for block_number in first_block_hash_to_add..self.trin_execution.next_block_number() {
+            let block = era_manager.get_next_block().await?;
+            ensure!(
```
morph-dev, line 229: nit: I'm not sure how I feel about this check. On one side, it's good to double check, on the other side, we rely on `EraManager` returning blocks in consecutive order in other places.

If it was me, I would have trusted that `EraManager` does the correct job, and just do something like this:

```rust
while era_manager.next_block_number() < self.trin_execution.next_block_number() {
    let block = era_manager.get_next_block().await?;
    self.trin_execution.database.db.put(...);
}
```

Up to you.

morph-dev (approved):
Good work :shipit:

trin-execution/src/main.rs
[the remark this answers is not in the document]
KolbyML: The team has discussed this in the past and the conclusion has came to including utility sub-commands in the main binary as then you always know the utility's work with your binary version. It is also less tedious to use them. Also most Execution Clients I read follow this practice. Multiple binaries feels like a lot of work both from a users perspective and developer's perspective

trin-execution/src/subcommands/era2.rs
KolbyML: If slower machines can't even run an execution client, I am not sure why they should be able to generate a `.era2` file.

They should minimum have 16GB of ram, and if that is the case there should be no problem, and if they need more they can add a swap file.

This isn't an issue, creating `.era2` files is already
- expensive
- slow
- niche

99% of people won't be generating these files, they will be using the files we generated.

So I don't think we should optimize this for the lowest case, they can use a .era2 file we generated or use a functional machine.

trin-execution/src/subcommands/era2.rs
KolbyML: sure

PR #1347, opened 2024-07-26
refactor: offer rpc endpoints

njgheorghita:
### What was wrong?
- To debug #1345 we need a jsonrpc endpoint that allows gossiping multiple pieces of data simultaneously (aka exactly how it's specified in the wire protocol)
- our current jsonrpc api is not [spec](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/ethereum/portal-network-specs/assembled-spec/jsonrpc/openrpc.json&uiSchema%5BappBar%5D%5Bui:splitView%5D=false&uiSchema%5BappBar%5D%5Bui:input%5D=false&uiSchema%5BappBar%5D%5Bui:examplesDropdown%5D=false) compliant
  - spec defines: `portal_historyOffer(enr, content_key, content_value)`
  - we have `portal_historyOffer(enr, content_key, Option<content_value>)`
    - this is not a great way of handling our offer endpoint...
      - doesn't allow offering multiple content keys simultaneously
      - the `Option` is awkward and can be confusing to the user about what's actually happening (aka which offer kind is being used) 

### How was it fixed?
- Updated our `portal_historyOffer()` endpoint to be spec compliant
- Added a `portal_historyWireOffer()` endpoint that allows direct access to the offer wire message type.
  - this endpoint requires all content keys to be available locally

### To-Do

### DIFF
--- a/ethportal-api/src/state.rs
+++ b/ethportal-api/src/state.rs
@@ -92,6 +92,7 @@ pub trait StateNetworkApi {
     ) -> RpcResult<TraceGossipInfo>;
 
     /// Send an OFFER request with given ContentKey, to the designated peer and wait for a response.
+    /// Does not store the content locally.
     /// Returns the content keys bitlist upon successful content transmission or empty bitlist
     /// receive.
     #[method(name = "stateOffer")]
@@ -99,7 +100,7 @@ pub trait StateNetworkApi {
         &self,
         enr: Enr,
         content_key: StateContentKey,
-        content_value: Option<StateContentValue>,
+        content_value: StateContentValue,
     ) -> RpcResult<AcceptInfo>;
 
     /// Store content key with a content data to the local database.
--- a/portalnet/src/overlay/protocol.rs
+++ b/portalnet/src/overlay/protocol.rs
@@ -42,8 +42,8 @@ use ethportal_api::{
         distance::{Distance, Metric},
         enr::Enr,
         portal_wire::{
-            Accept, Content, CustomPayload, FindContent, FindNodes, Message, Nodes, Offer, Ping,
-            Pong, PopulatedOffer, ProtocolId, Request, Response,
+            Accept, Content, CustomPayload, FindContent, FindNodes, Message, Nodes, Ping, Pong,
+            PopulatedOffer, ProtocolId, Request, Response,
         },
     },
     utils::bytes::hex_encode,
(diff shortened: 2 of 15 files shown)


njgheorghita:
@morph-dev I looked into implementing [your suggestion](https://github.com/ethereum/trin/pull/1342#discussion_r1684634841) but I'm not sure it's possible. IIUC you suggested that we can remove the `Populated` paradigm (since for all offers we now require a db lookup to validate that the content is available locally, so why not just use a "populated" offer to handle all internal offer requests). Well, the problem is that we still need to handle "unpopulated" offer requests incoming from other peers on the network. So the "unpopulated" request types are still very much needed. This seems enough to me to leave the internal offer request handling as-is. 

However, it does leave us with a slightly awkward disparity in naming between internal/external offer messages. (aka "wire" for external-facing and "populated" for internal usage). Ok, this is not the cleanest, but I'm not convinced it's worthwhile to consolidate all the naming since..
- this discrepancy basically has always existed (we never used "populated" in external api)
- a "wire offer" (external) is a non-populated offer (internal) ... and vv. If we take the foundation here to be the jsonrpc api specification (defining `portal_historyOffer(enr, content_key, content_value`), then I'm not sure there's a satisfactorily clean way to consolidate the naming tension between internal/external apis

Open to feedback for alternative approaches ofc

KolbyML (approved):
:shipit: this is nice offer now matches spec

ethportal-api/src/state.rs
```diff
@@ -92,14 +92,26 @@ pub trait StateNetworkApi {
...
-        content_value: Option<StateContentValue>,
+        content_value: StateContentValue,
+    ) -> RpcResult<AcceptInfo>;
+
+    /// Send an OFFER request with given ContentKeys, to the designated peer and wait for a
+    /// response. Requires the content keys to be stored locally.
+    /// Returns the content keys bitlist upon successful content transmission or empty bitlist
+    /// receive.
+    #[method(name = "stateWireOffer")]
+    async fn wire_offer(
```
morph-dev, line 111: This doesn't make sense for the state network, because content value that is stored can't be validated by receiving side (it's missing proof data). So please remove it.

Alternatively, API can be something like:
`(&self, end: Enr, content: Vec<(StateContentKey, StateContentValue)>)`
which is vectorized version of `offer` above.

portalnet/src/overlay/protocol.rs
```diff
@@ -518,48 +518,42 @@ where
 
     /// Offer is sent in order to store content to k nodes with radii that contain content-id
     /// Offer is also sent to nodes after FindContent (POKE)
-    pub async fn send_offer(
+    pub async fn send_wire_offer(
         &self,
-        content_keys: Vec<RawContentKey>,
         enr: Enr,
+        content_keys: Vec<RawContentKey>,
```
morph-dev, line 524: I would make this type `Vec<TContentKey>` because:
1. all calls to this function convert `Vec<TContentKey>` to `Vec<RawContentKey>` right before the call
2. we immediately convert it back to `Vec<TContentKey>`

Seems like we are doing some useless work

morph-dev (commented):
In general, LGTM. Few small comments.

Regarding our discussion, thanks for the explanation. I can't say I fully understand all details but it seems that it's not as straight forward as it appears on the first look. I trust your judgement, so let's leave it this way.

PR #1348, opened 2024-07-27
fix: unable to find ENR for NodeId killing network perf

KolbyML:
### What was wrong?
image

image

### How was it fixed?

image
image
image
image
image
image

In short this bug was caused by us only checking for `Enr`'s from our `node_addr_cache` which only contains new temporary connections. The fix was to also check our `kbuckets` and pending query caches

## Other issue
image

^ So I did implement a solution for this which checks each active sub-network as a as a fallback method.

GetEnr() json-rpc checks the kbuckets of the sub-network, but it doesn't check

image
This is only available in the OverlayService, not Overlay, so if in the future we still find a issue we can add it to the json-rpc GetEnr, but for now I think the solution is good enough.

I wish we wouldn't need to query every active sub-network as a backup option, but without a much bigger code change I think this is the best solution.

`enr_cache` should help reduce any locks on discv5 kbuckets or doing Json-rpc requests, uTP recv_from is called a lot and needing to lock kbucket or do a json-rpc request isn't the viable.

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -17,8 +17,9 @@ use discv5::{
 };
 use lru::LruCache;
 use parking_lot::RwLock;
-use tokio::sync::mpsc;
+use tokio::sync::{mpsc, RwLock as TokioRwLock};
 use tracing::{debug, info, warn};
+use trin_validation::oracle::HeaderOracle;
 use utp_rs::{cid::ConnectionPeer, udp::AsyncUdpSocket};
 
 use super::config::PortalnetConfig;
@@ -42,6 +43,9 @@ const ENR_PORTAL_CLIENT_KEY: &str = "c";
 
 /// ENR file name saving enr history to disk.
 const ENR_FILE_NAME: &str = "trin.enr";
+
+/// Amount of Enr to cache for Discv5UdpSocket
+const ENR_CACHE_CAPACITY: usize = 150;
 
 pub type ProtocolRequest = Vec<u8>;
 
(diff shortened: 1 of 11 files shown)


portalnet/src/discovery.rs
```diff
@@ -418,25 +481,10 @@ impl AsyncUdpSocket<UtpEnr> for Discv5UdpSocket {
...
-                                Err(io::Error::new(
-                                    io::ErrorKind::Other,
-                                    "ENR not found for talk req destination",
-                                ))
-                            }
-                        }?;
-                        UtpEnr(enr)
-                    }
-                };
+                let enr = self.find_enr(src_node_id).await?;
```
carver, line 487: Can we determine the subprotocol making the request here, and pass that information into `find_enr`? Then `find_enr` can look more directly for the relevant ENR (if it needs the oracle), rather than hitting all of beacon, history, state, etc.

carver (approved):
Looks good, thanks for hunting this down and fixing!

There are a couple suggested tweaks. Feel free to make best effort and merge, unless you have more questions.

I'm not sure I love the patterns in place, but this PR at least matches what exists. Especially for a bugfix that's affecting "mainnet," I wouldn't want to hold it up for an architecture/style issue.

One example: the name HeaderOracle doesn't seem to match its functionality anymore, it sort of crept away from that into a cross-network proxy. Anyway, that's not to fix on this PR, just an observation.

portalnet/src/discovery.rs
KolbyML: There isn't any clear way to determine which Portal sub-protocol the uTP message is under, without a major refactor as currently all uTP connections are opaque. Other then all uTP connections currently being opaque on what Portal sub-network they are using.

One possible solution could be to share a Arc<RwLock<uTPEnrCache>> between all networks and this class the uTPEnrCache size would be determined by the max possible uTP connections that can be opened at a time. We put add the Enr whenever we would open a uTP connection. This way we wouldn't need to store which sub-network initiated the connection and would also not have to rely on calling the JSON-RPC. It would be weird that this Arc<> would basically be tangled everywhere though, but maybe that is fine.

If we think this solution is cleaner we can implement it in a follow up PR. Thinking more about it I do kind of like the solution I just suggested, but I think this would better be discussed on a meeting or something.

I am a little confused on what to do, but I do want to get something working out for the time being

KolbyML: A issue with that approach though is we are possibly using more memory then we need to, so it is a jungling act of trade offs should we use less memory and worry less about being in-efficient or should we prioritize something in the middle I am not sure.

I think without proper performance testing (which wouldn't be the easiest to do or be a priority at this time) I am not sure if what I just proposed would even be worth it, because if we didn't do `utp_limit * 2 * active_networks` then we would still need to make JSON_RPC requests.

So on second thought I am not sure if it is that easy to come up with a best solution, only a group of solutions with trade offs

portalnet/src/discovery.rs
```diff
@@ -43,6 +44,9 @@ const ENR_PORTAL_CLIENT_KEY: &str = "c";
 /// ENR file name saving enr history to disk.
 const ENR_FILE_NAME: &str = "trin.enr";
 
+/// Amount of Enr to cache for Discv5UdpSocket
+const ENR_CACHE_CAPACITY: usize = 150;
```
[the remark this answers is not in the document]
KolbyML: Sounds good I did a quick refactor 

```
    // Set the enr_cache_capacity to half the maximum uTP limit between all active networks. This is
    // a trade off between memory usage and increased searches from the networks for each Enr.
    let enr_cache_capacity =
        portalnet_config.utp_transfer_limit * 2 * trin_config.portal_subnetworks.len() / 2;
```

KolbyML:
Currently `node_addr_cache` is fairly high so maybe we could reduce its size. Do this idea and have https://github.com/ethereum/trin/pull/1348#discussion_r1694321077

``` 
struct EnrCache {
      node_addr_cache (for temp connections): LRU<100>
      uTP_Enr_cache: max potential uTP connections inbound and outbound>
}

impl EnrCache {
    pub fn getEnr(nodeId) { try to fetch from both caches}
}

Then we could do the solution without relying on JSON-RPC requests or knowing which network it is well, well having a well rounded cache.

**Review by njgheorghita — APPROVED** [review]
Sick! Yeah, agreed that this architecture isn't entirely satisfactory but that question is probably better addressed during a re-evaluation of `HeaderOracle` and how we want to name / architect this module moving forward

File: src/lib.rs [hunk 3]
```diff
@@ -68,9 +67,25 @@ pub async fn run_trin(
...
+    let header_oracle = Arc::new(RwLock::new(header_oracle));
+
     // Initialize and spawn uTP socket
     let (utp_talk_reqs_tx, utp_talk_reqs_rx) = mpsc::unbounded_channel();
-    let discv5_utp_socket = Discv5UdpSocket::new(Arc::clone(&discovery), utp_talk_reqs_rx);
+
+    // Set the enr_cache_capacity to half the maximum uTP limit between all active networks. This is
+    // a trade off between memory usage and increased searches from the networks for each Enr.
+    let enr_cache_capacity =
+        portalnet_config.utp_transfer_limit * 2 * trin_config.portal_subnetworks.len() / 2;
```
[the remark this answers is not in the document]
KolbyML: `portalnet_config.utp_transfer_limit * trin_config.portal_subnetworks.len()` this would be half.

as utp_transfer_limit is 2x for income and outbound limit

PR #1350, opened 2024-07-31
fix: a few edge cases for find_enr() + distinction of same logs

KolbyML:
### What was wrong?

```nim
2024-07-31T04:32:12.149516Z ERROR trin_history::events: Error processing portal history request, responding with empty TALKRESP error=Error while building accept message: unable to find ENR for NodeId request.discv5.id=75d0ab97e1527ba1
```

Which is still popping up just a lot less, I think this is due to a few different edge cases

Also it is annoying because we use the same log in 2 different functions

### How was it fixed?

- making the same logs more distinct
- remove redundant `find_enr()` call
- add a few more possible cases to `find_enr()`

### DIFF
--- a/portalnet/src/overlay/service.rs
+++ b/portalnet/src/overlay/service.rs
@@ -948,7 +948,7 @@ where
                     // Generate a connection ID for the uTP connection.
                     let enr = self.find_enr(source).ok_or_else(|| {
                         OverlayRequestError::AcceptError(
-                            "unable to find ENR for NodeId".to_string(),
+                            "handle_find_content: unable to find ENR for NodeId".to_string(),
                         )
                     })?;
                     let enr = UtpEnr(enr);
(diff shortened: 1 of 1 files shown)


portalnet/src/overlay/service.rs
```diff
@@ -948,7 +948,7 @@ where
                     // Generate a connection ID for the uTP connection.
                     let enr = self.find_enr(source).ok_or_else(|| {
                         OverlayRequestError::AcceptError(
-                            "unable to find ENR for NodeId".to_string(),
+                            "handle_offer_content: unable to find ENR for NodeId".to_string(),
```
njgheorghita, line 951: should this be `handle_find_content`? that's the function we're inside at this point

portalnet/src/overlay/service.rs
```diff
@@ -2497,6 +2496,14 @@ where
             return Some(entry.value().clone().enr());
         }
 
+        if let kbucket::Entry::Pending(mut entry, _) = self.kbuckets.write().entry(&key) {
```
njgheorghita, line 2499: I'd add a comment explaining the purpose/motivation for checking kbuckets before `find_enr`

njgheorghita (approved):

PR #1351, opened 2024-08-01
feat(portal-bridge): add single hunter mode

njgheorghita:
### What was wrong?
Add single hunter mode so it can be more effectively deployed in our deployment scripts.

### To-Do


KolbyML (approved):
:shipit:

PR #1353, opened 2024-08-02
refactor: update state types to use address_hash instead of address to match spec + remove recursive gossip

KolbyML:
### What was wrong?
Trin doesn't match the latest state spec https://github.com/ethereum/portal-network-specs/pull/329
### How was it fixed?
updating it to match

Well updating it, recursive gossip wasn't remove yet so I did that too


trin-execution/src/storage/evm_db.rs
```diff
@@ -21,14 +21,12 @@ use super::{
...
 
-const REVERSE_HASH_LOOKUP_PREFIX: &[u8] = b"reverse hash lookup";
-
 #[derive(Debug, Clone)]
 pub struct EvmDB {
     /// State config
     pub config: StateConfig,
     /// Storage cache for the accounts used optionally for gossiping.
-    pub storage_cache: HashMap<Address, HashSet<B256>>,
+    pub storage_cache: HashMap<B256, HashSet<B256>>,
```
morph-dev, line 29: nit: might be worth updating the comment to indicate that key is keccak of address

trin-state/src/validation/trie.rs
```diff
@@ -394,8 +398,9 @@ mod tests {
     #[test]
     fn validate_account_state_just_leaf() {
         let address = Address::random();
-        let path: Vec<u8> = keccak(address)
-            .as_bytes()
+        let address_hash = keccak256(address);
```
morph-dev, line 401: nit: Considering that `address` was created randomly and used only for `address_hash`, we could just create random `B256` (I believe `B256::random()` should work). Same below.

trin-state/src/validation/trie.rs
```diff
@@ -49,11 +48,11 @@ pub fn validate_node_trie_proof<'proof, 'path>(
...
 pub fn validate_account_state(
     root_hash: Option<B256>,
-    account: &Address,
+    address_hash: &B256,
     proof: &TrieProof,
 ) -> Result<AccountState, StateValidationError> {
-    let path: Vec<u8> = keccak(account)
-        .as_bytes()
+    let path: Vec<u8> = address_hash
+        .as_slice()
```
morph-dev, line 55: nit: I believe you don't `.as_slice()` and can just call `address_hash.iter()...`.

Also few places below and in the other file.

trin-state/src/validation/trie.rs
morph-dev, line 1: I believe that the `TrieProofValidationInfo` is no longer needed or at least can be simplified (it was useful in order to more efficiently do recursive gossip).

Either remove it and clean up the code, or create an issue and assign it to me and I will do it when I get a bit more time.

morph-dev (approved):

trin-state/src/validation/trie.rs
KolbyML: I made a issue for it https://github.com/ethereum/trin/issues/1360

PR #1355, opened 2024-08-04
feat: add more metrics to trin execution

KolbyML:
Milos said I should try to take what I can from my big PR so I will do that, I will need a week to split .era2 PR more, but I will pull out anything that can be reviewed which can't break my database code for now

### What was wrong?
There wasn't enough metrics to understand things like transaction progress times, storage etc
### How was it fixed?
By updating
- the dashboard
- adding new metrics to be used

### DIFF
--- a/trin-execution/metrics/grafana/datasource.yml
+++ b/trin-execution/metrics/grafana/datasource.yml
@@ -3,7 +3,7 @@ apiVersion: 1
 datasources:
 - name: Prometheus
   type: prometheus
-  url: http://prometheus:9090 
+  url: http://localhost:9090
   isDefault: true
   access: proxy
   editable: true
--- a/trin-execution/metrics/prometheus/prometheus.yml
+++ b/trin-execution/metrics/prometheus/prometheus.yml
@@ -12,6 +12,8 @@ scrape_configs:
   # Node exporter metrics
   - job_name: "node"
     static_configs:
-      - targets: ["host.docker.internal:9091"]
+      - targets: 
+        - host.docker.internal:9091
+        - localhost:9091
         labels:
           instance: local_node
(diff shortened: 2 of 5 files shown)


trin-execution/metrics/grafana/datasource.yml
```diff
@@ -3,7 +3,7 @@ apiVersion: 1
 datasources:
 - name: Prometheus
   type: prometheus
-  url: http://prometheus:9090 
+  url: http://localhost:9090
```
morph-dev, line 6: this doesn't look right

trin-execution/metrics/prometheus/prometheus.yml
```diff
@@ -12,6 +12,8 @@ scrape_configs:
   # Node exporter metrics
   - job_name: "node"
     static_configs:
-      - targets: ["host.docker.internal:9091"]
+      - targets: 
+        - host.docker.internal:9091
+        - localhost:9091
```
morph-dev, line 17: not sure about this one either (but it might be ok, since it has both)

morph-dev (approved):
Thanks for splitting it up.

trin-execution/metrics/prometheus/prometheus.yml
KolbyML: This is so both MacOS works and Linux

trin-execution/metrics/grafana/datasource.yml
KolbyML: Maybe I need to add another separate datasource to support both macos and linux, but what I do know, is http://prometheus:9090 doesn't work on linux

KolbyML: I think because of my updates to docker-compose this should work. I don't run this on macos anymore, so if someone finds a bug we can fix it then

PR #1357, opened 2024-08-04
feat: add era pre-(fetching/processing)

KolbyML:
https://github.com/ethereum/trin/pull/1356 PR should be reviewed first

### What was wrong?
Era1 files would block executing as we had to wait for them to download then recover the senders
### How was it fixed?
download and process the next era1 file in the background while executing

### DIFF
--- a/trin-execution/src/era_manager.rs
+++ b/trin-execution/src/era_manager.rs
@@ -1,15 +1,39 @@
-use std::time::Duration;
+use std::{
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use e2store::{
     era1::{BlockTuple, Era1, BLOCK_TUPLE_COUNT},
     utils::get_shuffled_era1_files,
 };
+use ethportal_api::{
+    types::execution::{header::Header, transaction::Transaction},
+    BlockBody,
+};
+use parking_lot::Mutex;
+use rayon::iter::{IntoParallelIterator, ParallelIterator};
+use revm_primitives::Address;
 use surf::{Client, Config};
 use tokio::time::sleep;
-use tracing::warn;
+use tracing::{info, warn};
+
+#[derive(Debug, Clone)]
+pub struct TransactionsWithSenders {
+    pub transaction: Transaction,
+    pub senders_address: Address,
+}
+
+#[derive(Debug, Clone)]
+pub struct BlockWithRecoveredSenders {
+    pub header: Header,
+    pub uncles: Option<Vec<Header>>,
+    pub transactions: Vec<TransactionsWithSenders>,
+}
 
 pub struct EraManager {
-    current_era: Option<Era1>,
+    current_era: Option<Vec<BlockWithRecoveredSenders>>,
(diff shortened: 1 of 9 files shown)


trin-execution/src/era_manager.rs
```diff
@@ -23,37 +47,86 @@ impl EraManager {
...
+            }
+        };
+
         let era1_path = self
             .era1_files
             .iter()
-            .find(|file| file.contains(&format!("mainnet-{epoch_index:05}-")))
-            .expect("to be able to find era1 file");
+            .find(|file| file.contains(&format!("mainnet-{:05}-", epoch_index + 1)))
+            .expect("to be able to find era1 file")
```
morph-dev, line 105: What if `epoch_index` corresponds to the last epoch? This would fail, and so would `get_block_by_number`, which isn't expected behavior?.

trin-execution/src/era_manager.rs
```diff
@@ -23,37 +47,86 @@ impl EraManager {
...
     }
 
-    async fn download_era1_file(&self, epoch_index: u64) -> anyhow::Result<Era1> {
+    async fn download_processed_era(
+        &self,
+        epoch_index: u64,
+    ) -> anyhow::Result<Vec<BlockWithRecoveredSenders>> {
+        info!("Fetching the next era file");
+        let start = Instant::now();
+        while self.current_era.is_some() && self.next_era.lock().is_none() {
```
morph-dev, line 78: I would say that this is dangerous approach for dealing with async tasks.

For example, what if task spawn at line 109 fails (for example, network issue)? It will just finish without setting `next_era`, and this one would fail after waiting for 20 minutes.

morph-dev (changes requested):
I like the optimization idea, but I think there are few cases that are not handled correctly in regard to async tasks and usage of the EraManager (see line comments).

Here is the approach that I would take.
Instead of having `next_era: Arc<Mutex<Option<...>>>`, I would have it like this:
`next_era: Option<JoinHandle<ProcessedEra>>` (where `JoinHandle` is the object returns by `tokio::spawn`).

Also, considering that maybe not all users of `EraManager` plan to process data sequentially, I would have a parameter that would enable/disable this optimization.

Happy to have a call if you want to discuss this into more details.

trin-execution/src/era_manager.rs
```diff
@@ -1,15 +1,39 @@
...
+#[derive(Debug, Clone)]
+pub struct BlockWithRecoveredSenders {
+    pub header: Header,
+    pub uncles: Option<Vec<Header>>,
+    pub transactions: Vec<TransactionsWithSenders>,
+}
 
 pub struct EraManager {
-    current_era: Option<Era1>,
+    current_era: Option<Vec<BlockWithRecoveredSenders>>,
```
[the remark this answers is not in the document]
KolbyML: image

This is my issue with premature review's like this

KolbyML:
Me and Milos had a call. I am going to close this PR and open a new one once I've implemented Era support. I already implemented some of the feedback, and others I think is very useful. I made this PR a little pre-mature as I supporting Era requires almost fully rewriting it to an extent

PR #1361, opened 2024-08-06
feat: add support for Altair consensus types

KolbyML:
### What was wrong?
~~For `trin-execution` I need to use era files. Currently e2store/era.rs hardcoded to only work with Bellatrix, but it shouldn't be much work to support Altair, Bellatrix, Capella, Deneb (so we still won't be able to decode era files with phase0 beacon blocks, but that isn't a priority for now)~~

~~Anyways that will be my next PR~~

~~era files don't always have 8192 block's, but the era files I was trying to decode to test if they include Bellatrix blocks, but we couldn't decode Altair beacon blocks and the code would panic.~~

### How was it fixed?

Add support for Altair, and use consensus-spec-tests like ogenev did for adding support for the other forks

So now in my next PR I will be able to decode era files with altair blocks which is nice


KolbyML:
I learnt I don't need to add Altair support for what I need. It is already implemented so it is up to you guys I can close this if you want, but if we want to support all beacon chain forks, I mean the work is done so  up to you guys

ogenev:
I'm against merging this if we don't need it.

We can reopen this PR if we want to add this support in the future.

PR #1362, opened 2024-08-07
feat: allow e2store/era to decode all forks not just bellatrix

KolbyML:
### What was wrong?
`e2store/era` was hardcoded to only decode bellatrix, when I need to be able to decode all forks since execution payload was added
### How was it fixed?

by doing a if else decode pattern like we do for decoding transactions

### DIFF
--- a/e2store/src/era.rs
+++ b/e2store/src/era.rs
@@ -54,11 +54,7 @@ impl Era {
             let entry: Entry = file.entries[idx].clone();
             let fork = get_beacon_fork(next_slot);
             let beacon_block = CompressedSignedBeaconBlock::try_from(&entry, fork)?;
-            next_slot = match &beacon_block.block {
-                SignedBeaconBlock::Bellatrix(block) => block.message.slot,
-                SignedBeaconBlock::Capella(block) => block.message.slot,
-                SignedBeaconBlock::Deneb(block) => block.message.slot,
-            } + 1;
+            next_slot = beacon_block.block.slot() + 1;
             blocks.push(beacon_block);
         }
         let fork = get_beacon_fork(slot_index_state.slot_index.starting_slot);
--- a/e2store/src/era.rs
+++ b/e2store/src/era.rs
@@ -44,15 +44,25 @@ impl Era {
         let version = VersionEntry::try_from(&file.entries[0])?;
         let entries_length = file.entries.len();
         let mut blocks = vec![];
-        // Iterate over the block entries. Skip the first and last 3 entries.
-        for idx in 1..entries_length - 4 {
-            let entry: Entry = file.entries[idx].clone();
-            let beacon_block = CompressedSignedBeaconBlock::try_from(&entry)?;
-            blocks.push(beacon_block);
-        }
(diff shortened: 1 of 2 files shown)


ethportal-api/src/types/consensus/beacon_block.rs
```diff
@@ -125,8 +125,32 @@ pub struct SignedBeaconBlock {
 }
 
 impl SignedBeaconBlock {
+    /// SSZ decode without explicit fork variant.
+    pub fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, ssz::DecodeError> {
```
ogenev, line 129: My only concern about using this pattern is that it is slow because the beacon state is huge. This is why I was trying to avoid implementing it. (this is also mentioned in the lighthouse docs, I don't remember where exactly).

Can we get away by using the explicit version and get information about the fork before deserializing the beacon state? For example, if we know the slot of the block we can determine which fork it is part of.

ethportal-api/src/types/consensus/beacon_state.rs
```diff
@@ -147,7 +147,21 @@ pub struct BeaconState {
 }
 
 impl BeaconState {
-    pub fn from_ssz_bytes(bytes: &[u8], fork_name: ForkName) -> Result<Self, DecodeError> {
+    pub fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
```
ogenev, line 150: My only concern about using this pattern is that it is slow because the beacon state is huge. This is why I was trying to avoid implementing it. (this is also mentioned in the lighthouse docs, I don't remember where exactly).

Can we get away by using the explicit version and get information about the era fork before deserializing the beacon state? For example, if we know the slot of the block we can determine which fork it is part of.

KolbyML:
@ogenev ready for another view

e2store/src/era.rs
```diff
@@ -44,15 +44,25 @@ impl Era {
...
         for idx in 1..entries_length - 4 {
             let entry: Entry = file.entries[idx].clone();
-            let beacon_block = CompressedSignedBeaconBlock::try_from(&entry)?;
+            let fork = get_beacon_fork(next_slot);
+            let beacon_block = CompressedSignedBeaconBlock::try_from(&entry, fork)?;
+            next_slot = match &beacon_block.block {
+                SignedBeaconBlock::Bellatrix(block) => block.message.slot,
+                SignedBeaconBlock::Capella(block) => block.message.slot,
+                SignedBeaconBlock::Deneb(block) => block.message.slot,
+            } + 1;
```
ogenev, line 61: nit:

```suggestion
            next_slot = match &beacon_block.block {
                SignedBeaconBlock::Bellatrix(block) 
                | SignedBeaconBlock::Capella(block) 
                | SignedBeaconBlock::Deneb(block) => block.message.slot,
            } + 1;
```

ogenev (approved):
LGTM 👍

e2store/src/era.rs
KolbyML: I tried using this, but it gave me an error because the enum returns different types.

I will make a getter in `SignedBeaconBlock` should should simply this common case

PR #1366, opened 2024-08-08
feat: define state related eth_* rpc endpoints

morph-dev:
### What was wrong?

We were missing `eth_getBalance`, `eth_getCode` and `eth_getStorageAt` JSON rpc endpoints.

### How was it fixed?

Defined missing endpoints (using `todo!()` as implementation, for future PR).

We will need other endpoints in the near future (`eth_call` and `eth_estimateGass`), but they require more complicated type (`Transaction`), so I left them for future work.

### To-Do


KolbyML (approved):
:shipit:

PR #1367, opened 2024-08-08
refactor: simplify json-rpc proxy logic

morph-dev:
### What was wrong?

We proxy the JSON rpc requests to the corresponding network. This required some duplicated logic for each subnetwork (and it wasn't consistent).

### How was it fixed?

Introduced new type using generics and managed to create one functions that can proxy requests to each subnetwork.

While this is an improvement, I'm not completely satisfied with the current state because I believe that we still have some unnecessary json serialization / deserializaion. But that will be left for some potential future improvements.

### To-Do

- [x] Run hive tests locally and make sure there is no regression.

### DIFF
--- a/rpc/src/fetch.rs
+++ b/rpc/src/fetch.rs
@@ -1,58 +1,72 @@
+use std::fmt::Debug;
+
 use alloy_primitives::B256;
-/// Fetch data from related Portal networks
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use tokio::sync::mpsc;
 
 use ethportal_api::{
     types::{
         execution::{block_body::BlockBody, header::Header},
-        jsonrpc::{endpoints::HistoryEndpoint, request::HistoryJsonRpcRequest},
+        history::ContentInfo,
+        jsonrpc::{
+            endpoints::{HistoryEndpoint, SubnetworkEndpoint},
+            request::{HistoryJsonRpcRequest, JsonRpcRequest},
+        },
+        query_trace::QueryTrace,
     },
-    utils::bytes::hex_decode,
-    ContentValue, HistoryContentKey, HistoryContentValue,
+    HistoryContentKey, HistoryContentValue,
 };
 
-use crate::errors::RpcServeError;
+use crate::{
+    errors::{ContentNotFoundJsonError, RpcServeError},
+    serde::from_value,
+};
 
-pub async fn proxy_query_to_history_subnet(
-    network: &mpsc::UnboundedSender<HistoryJsonRpcRequest>,
-    endpoint: HistoryEndpoint,
-) -> Result<Value, RpcServeError> {
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ContentNotFoundError {
+    message: String,
+    trace: Option<QueryTrace>,
+}
+
+/// Fetch and deserialize data from Portal subnetwork.
+pub async fn proxy_to_subnet<TEndpoint, TOutput>(
+    network: &mpsc::UnboundedSender<JsonRpcRequest<TEndpoint>>,
+    endpoint: TEndpoint,
+) -> Result<TOutput, RpcServeError>
+where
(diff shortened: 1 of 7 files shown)


rpc/src/fetch.rs
```diff
@@ -1,58 +1,64 @@
...
 
-use crate::errors::RpcServeError;
+use crate::{errors::RpcServeError, serde::from_value};
 
-pub async fn proxy_query_to_history_subnet(
-    network: &mpsc::UnboundedSender<HistoryJsonRpcRequest>,
-    endpoint: HistoryEndpoint,
-) -> Result<Value, RpcServeError> {
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct ContentNotFoundError {
```
morph-dev, line 22: note: Not sure if there is better place to move this structure to. It would be useful when responding with this error (all `trin-{subnetwork}/src/jsonrpc.rs` files).

KolbyML (dismissed):
:shipit: I think this PR looks good and can me merged. I read the PR this morning, but I decided to wait because of the note: https://github.com/ethereum/trin/pull/1367/files#r1710320237

I think if (the note) should be relocated it will become apparent when it is important. Like if it would be useful for xyz we can and will know this move it then.

So unless anyone has a strong opinion, I think "here" is good enough for now, at least until we have an intention for expanding this to be generic with network specific error handling or something idk.

rpc/src/fetch.rs
ogenev: Maybe move this to `rpc/src/errors.rs`?

rpc/src/fetch.rs
```diff
@@ -1,58 +1,64 @@
...
-                            .expect("Failed to parse query trace"),
-                        None => None,
-                    };
-                    Err(RpcServeError::ContentNotFound { message, trace })
-                } else {
-                    Err(RpcServeError::Message(msg))
-                }
+    let Some(response) = resp_rx.recv().await else {
+        return Err(RpcServeError::Message(
+            "Internal error: No response from subnetwork".to_string(),
```
ogenev, line 41: Can we also display the name of the subnetwork here and in the errors below? I think adding a trait to all subnetwork endpoints with some method that returns the name or implementing `DIsplay` should do the trick but maybe there is a simpler approach.

ogenev (approved):
Looks like a nice improvement to me.

morph-dev:
I run hive tests locally and I didn't notice any regression. Some tests are failing, but they are failing with latest docker image as well (I think hive is very flaky on my local machine, and it also takes a long time to run).

I'm not even sure whether hive tests (and which ones) even use JSON-RPC.

KolbyML:
> I'm not even sure whether hive tests (and which ones) even use JSON-RPC.
All hive tests use the JSON-RPC

KolbyML:
> I run hive tests locally and I didn't notice any regression. Some tests are failing, but they are failing with latest docker image as well (I think hive is very flaky on my local machine, and it also takes a long time to run).

I don't think this is enough information for a 3rd party to come to a conclusion

morph-dev:
I run only `trin` client (running all clients takes a lot of time, but I might run it tonight).

There are some failing tests beacon suite, but they are the same one as the one that are failing here: https://portal-hive-experimental.ethdevops.io/, so I will ignore them.

The only other failing tests are in `history-interop`:
- `OFFER Epoch Accumulator:  trin --> trin` (x2)
- `GOSSIP blocks from A:trin --> B:trin`

The only error visible in the logs is that "Content not found in local storage"

Because they fail with latest trin docker image, and because other OFFER tests pass, I don't think it has anything to do with this PR, and I'm guessing it has something to do with the size of the payload and my machine.

What further steps do you propose me to take in order to figure out what is happening here?

KolbyML (approved):
If there isn't any extra tests failing on portal-hive

And the 3 you mentioned happen both on the master latest docker build and the docker build of this PR I think we are good and can call it a issue with your specific machine

njgheorghita (approved):
Nice refactor! :ship:

Issue #1360, opened 2024-08-06
remove TrieProofValidationInfo


KolbyML:
A description of this issue can be found here, but it is basically just removing TrieProofValidationInfo as it is no longer needed
https://github.com/ethereum/trin/pull/1353#discussion_r1705054927

PR #1368, opened 2024-08-10
refactor: state validation and trie node traversal improvements

morph-dev:
### What was wrong?

The state validation could be improved after recursive gossip removal.

fixes: #1360 

### How was it fixed?

- Improved trie node traversal logic and extract it to `ethportal-api`
  - it was previously designed only with validation in mind
  - now we can use it for other use cases as well (something we will need in order to support `eth_*` JSON rpc)
  - moved and added new tests
- Removed no longer needed types
- Extracted common logic into separate functions

Run hive tests locally (trin + fluffy clients) and all tests are passing.

### To-Do

### DIFF
--- a/ethportal-api/src/types/state_trie/mod.rs
+++ b/ethportal-api/src/types/state_trie/mod.rs
@@ -10,6 +10,7 @@ use super::bytes::{ByteList1024, ByteList32K};
 
 pub mod account_state;
 pub mod nibbles;
+pub mod trie_traversal;
 mod utils;
 
 /// The RLP encoding of a trie node.
--- a/ethportal-api/src/types/state_trie/nibbles.rs
+++ b/ethportal-api/src/types/state_trie/nibbles.rs
@@ -32,6 +32,13 @@ impl Nibbles {
     pub fn unpack_nibble_pair(packed_nibbles: &u8) -> [u8; 2] {
         [packed_nibbles >> 4, packed_nibbles & 0xF]
     }
+
+    pub fn unpack_nibbles(packed_nibbles: &[u8]) -> Vec<u8> {
+        packed_nibbles
+            .iter()
+            .flat_map(Self::unpack_nibble_pair)
+            .collect()
+    }
 }
 
 impl Encode for Nibbles {
--- a/ethportal-api/src/types/state_trie/trie_traversal.rs
+++ b/ethportal-api/src/types/state_trie/trie_traversal.rs
@@ -1,0 +1,311 @@
+use alloy_primitives::B256;
+use eth_trie::node::Node;
+use thiserror::Error;
+
+/// The explanation of how we reached empty node.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EmptyNodeInfo {
+    /// Reached [Node::Empty].
+    EmptyNode,
+    /// Consumed entire path to reach branch node without value.
+    EmptyBranchValue,
+    /// Reached extension node whose prefix is not prefix of the remaining path.
+    DifferentExtensionPrefix,
+    /// Reached leaf node whose prefix doesn't match remaining path.
+    DifferentLeafPrefix,
+}
+
+/// The error that can happen while traversing the node.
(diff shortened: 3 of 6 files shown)


KolbyML:
Are `check_traversal_result_is_node()` and `check_traversal_result_is_value()` planning to be used in the next PR?

KolbyML:
My main concern is whether this refactor will be used in follow up PR's or not. It probably will just feels a little off because nothing is using it yet to justify it's complexity over the previous implementation. Rust traits are nice because they remove a degree of OOP, but yeah. So I mean it will be used, but what if which makes me hesitate.

morph-dev:
> Are `check_traversal_result_is_node()` and `check_traversal_result_is_value()` planning to be used in the next PR?

No. They are specific for validation logic (they expect only one variant). While they are used once and can be inlined, I made them separate function in order to simplify logic of functions that are using them (not because they themself were complex). I'm completely fine with inlining their logic.

---

The next PR (implementing `eth_getBalance`, `eth_getCode`) would definitely use `NodeTraversal` in a very nice way (in my opinion). Something like this:
```rust
async fn fetch_account_state(state_root: B256, address: Address) -> Result<Option<AccountState>> {
    let mut node_hash = state_root;
    let mut path = Nibbles::unpack_nibbles(&keccak(address));
    loop {
        let node = fetch_node_from_network(node_hash).await?; // using RecursiveFindContent
        let traverse_result = node.as_trie_node()?.traverse(path);
        match traverse_result {
            TraversalResult::Empty(empty_node_info) => return Ok(None),
            TraversalResult::Value(value) => return Ok(Some(AccountState::decode(value))),
            TraversalResult::Node(next_node) => {
                node_hash = next_node.hash;
                path = next_node.remaining_path;
            },
            TraversalResult::Error(error) => return Err(error),
        }
    }
}
```

---

Regarding usage of traits. I don't have a strong preference on this. It could be simple function.
I'm considering implementing it for the `EncodedTrieNode` (as it seems very common to call `node.as_trie_node()?.traverse(path)`), which could justify usage of traits.

Another potential usage is adding `random_walk` function, something that glados (and bridge looking for missing content) would use. That way all "traversal" logic is grouped together. Of course, this can be separate function as well, but I would say it makes sense to be grouped.

KolbyML:
> The next PR (implementing `eth_getBalance`, `eth_getCode`) would definitely use `NodeTraversal` in a very nice way (in my opinion). Something like this:
> 
> ```rust
> async fn fetch_account_state(state_root: B256, address: Address) -> Result<Option<AccountState>> {
>     let mut node_hash = state_root;
>     let mut path = Nibbles::unpack_nibbles(&keccak(address));
>     loop {
>         let node = fetch_node_from_network(node_hash).await?; // using RecursiveFindContent
>         let traverse_result = node.as_trie_node()?.traverse(path);
>         match traverse_result {
>             TraversalResult::Empty(empty_node_info) => return Ok(None),
>             TraversalResult::Value(value) => return Ok(Some(AccountState::decode(value))),
>             TraversalResult::Node(next_node) => {
>                 node_hash = next_node.hash;
>                 path = next_node.remaining_path;
>             },
>             TraversalResult::Error(error) => return Err(error),
>         }
>     }
> }
> ```

This looks very nice

> Regarding usage of traits. I don't have a strong preference on this. It could be simple function.
I'm considering implementing it for the EncodedTrieNode (as it seems very common to call node.as_trie_node()?.traverse(path)), which could justify usage of traits.

I like traits as they prevent some types of abstraction I don't like, I do think their usage should have a reason of course which I think what you layed out is a valid case for them

KolbyML (approved):
Overall the PR looks good. I liked the idea of the traits usage Milos gave as an example it was pretty cool. Well I don't like one time logic being abstracted into its own function unless it is for
- encapsulating complex code
- removing repetitive duplication
- being use as a getter

I don't think it is the end of the world or should block this PR, although my tongue might really not like the taste of something. It probably doesn't have a meaningful impact on anything so my taste buds are probably being silly and I should ignore them.

njgheorghita (approved):
💯

PR #1370, opened 2024-08-12
feat(beacon-network): validate `HistoricalSummariesWithProof` against finalized state root

### LINKED ISSUE #1323: beacon-network: Validate `HistoricalSummariesWithProof` against the synced state root
Add validation for `HistoricalSummariesWithProof` against the last known state root: https://github.com/ethereum/trin/blob/284f98a8ad271d9f6b0eafc88309057bb81f48bb/trin-beacon/src/validation.rs#L160.

ogenev:
### What was wrong?
Fixes #1323 

### How was it fixed?
- add a method to get the latest finalized state root in `HeaderOracle`.
- implement validation methods in `BeaconValidator`.

Note: I had to use some creativity to write the unit tests for this validation because it is impossible to mock the jsno-rpc calls to the beacon overlay at this point.
This is why I split the validation into two methods and tested them individually, instead of one integration test of the `validate_content` method.

### To-Do

### DIFF
--- a/trin-beacon/src/validation.rs
+++ b/trin-beacon/src/validation.rs
@@ -1,21 +1,25 @@
+use alloy_primitives::B256;
 use anyhow::anyhow;
 use chrono::Duration;
-use ssz::Decode;
-use std::sync::Arc;
-
 use ethportal_api::{
     consensus::fork::ForkName,
-    types::content_value::beacon::{
-        ForkVersionedHistoricalSummariesWithProof, ForkVersionedLightClientBootstrap,
-        ForkVersionedLightClientFinalityUpdate, ForkVersionedLightClientOptimisticUpdate,
-        LightClientUpdatesByRange,
+    types::{
+        content_key::beacon::HistoricalSummariesWithProofKey,
+        content_value::beacon::{
+            ForkVersionedHistoricalSummariesWithProof, ForkVersionedLightClientBootstrap,
+            ForkVersionedLightClientFinalityUpdate, ForkVersionedLightClientOptimisticUpdate,
+            LightClientUpdatesByRange,
+        },
     },
     BeaconContentKey,
 };
 use light_client::consensus::rpc::portal_rpc::expected_current_slot;
+use ssz::Decode;
+use std::sync::Arc;
 use tokio::sync::RwLock;
-
+use tree_hash::TreeHash;
 use trin_validation::{
+    merkle::proof::verify_merkle_proof,
     oracle::HeaderOracle,
     validator::{ValidationResult, Validator},
 };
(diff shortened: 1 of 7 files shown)


KolbyML (approved):
:shipit:

trin-beacon/src/network.rs
```diff
@@ -1,6 +1,5 @@
-use std::sync::Arc;
-
 use parking_lot::RwLock as PLRwLock;
+use std::sync::Arc;
```
njgheorghita, line 2: *nitpick* but according to our guidelines, the `std` module imports are supposed to be in an isolated 1st grouping, before the external dependency imports grouping, followed by the internal dependency imports grouping. Any specific reason you changed that here (and in other places in this pr)?

trin-beacon/src/validation.rs
```diff
@@ -135,33 +139,83 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
+impl BeaconValidator {
+    /// General validation for the historical summaries with proof content
+    fn general_summaries_validation(
+        content: &[u8],
+        key: &HistoricalSummariesWithProofKey,
+    ) -> anyhow::Result<ForkVersionedHistoricalSummariesWithProof> {
+        let fork_versioned_historical_summaries =
+            ForkVersionedHistoricalSummariesWithProof::from_ssz_bytes(content).map_err(|err| {
+                anyhow!(
+                    "Historical summaries with proof has invalid SSZ bytes: {:?}",
```
njgheorghita, line 172: *nitpick* you can inline the `err` here, though this should be caught in the clippy check [here](https://github.com/ethereum/trin/blob/master/trin-beacon/src/lib.rs#L2) ... not sure why it's not being picked up?

trin-beacon/src/validation.rs
```diff
@@ -135,33 +139,83 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
+    }
 
-        Ok(ValidationResult::new(true))
+    /// Validate historical summaries against the latest finalized state root
+    async fn state_summaries_validation(
+        fork_versioned_historical_summaries: ForkVersionedHistoricalSummariesWithProof,
+        latest_finalized_root: B256,
+    ) -> anyhow::Result<()> {
+        // Verify the historical summaries against the latest finalized state root
+
```
njgheorghita, line 198: *nitpick* delete this newline, though I think you can just remove this docstring since it's basically the same as the docstring for the function

trin-beacon/src/validation.rs
```diff
@@ -135,33 +139,83 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
+        let historical_summaries_root = fork_versioned_historical_summaries
+            .historical_summaries_with_proof
+            .historical_summaries
+            .tree_hash_root();
+
+        if !verify_merkle_proof(
+            historical_summaries_root,
+            &historical_summaries_state_proof,
+            5,
+            59, // 31 + 28 = 59
```
njgheorghita, line 211: *nitpick* include what the `31` and `28` values refer to in this comment

njgheorghita (approved):
just a couple nitpicks 🚢

PR #1371, opened 2024-08-12
refactor: display decoded client type in utp logs

njgheorghita:
### What was wrong?
Manually decoding the utp client type for every log you're interested in is a bit of a pain, much easier to display the already decoded client type in the log.

### How was it fixed?
Added decoded client type to `UtpEnr` debug logs.

### To-Do

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -418,7 +418,7 @@ impl Discv5UdpSocket {
 }
 
 /// A wrapper around `Enr` that implements `ConnectionPeer`.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct UtpEnr(pub Enr);
 
 impl UtpEnr {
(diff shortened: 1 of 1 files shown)


portalnet/src/discovery.rs
```diff
@@ -433,6 +433,17 @@ impl UtpEnr {
     }
 }
 
+impl std::fmt::Debug for UtpEnr {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let peer_client_type = self.client().unwrap_or_else(|| "Unknown".to_string());
+        write!(
```
morph-dev, line 439: maybe something like this instead:
```rust
f.debug_struct("UtpEnr")
    .field("enr", &self.0)
    .field("Peer Client Type", &peer_client_type)
    .finish()
```

morph-dev (approved):

KolbyML (approved):
:shipit:

ogenev (approved):

PR #1372, opened 2024-08-12
fix(e2store): era missing a beacon block off by 1 error

KolbyML:
### What was wrong?
When I was testing reading beacon blocks from era files when I compared the result of era 600 and era 601 their execution numbers didn't line up by 1 block, when I fixed this they lined up and I could linearly scan blocks from era files

I tested this for a few other sequential era files and the issues remained the same
### How was it fixed?

change the 4 to a 3

### DIFF
--- a/e2store/src/era.rs
+++ b/e2store/src/era.rs
@@ -50,10 +50,10 @@ impl Era {
 
         // Iterate over the block entries. Skip the first and last 3 entries.
         let mut next_slot = slot_index_block.slot_index.starting_slot;
-        for idx in 1..entries_length - 4 {
-            let entry: Entry = file.entries[idx].clone();
+        for idx in 1..entries_length - 3 {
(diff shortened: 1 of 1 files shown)


e2store/src/era.rs
```diff
@@ -50,7 +50,7 @@ impl Era {
 
         // Iterate over the block entries. Skip the first and last 3 entries.
         let mut next_slot = slot_index_block.slot_index.starting_slot;
-        for idx in 1..entries_length - 4 {
+        for idx in 1..entries_length - 3 {
             let entry: Entry = file.entries[idx].clone();
```
morph-dev, line 54: nit: While you are at it, this cloning seems completely unnecessary (we can just pass `&file.entries[idx]` below).

morph-dev (approved):

PR #1374, opened 2024-08-12
Add .era2 types

KolbyML:
### What was wrong?
e2store doesn't support the `.era2` state snapshot format which will be used for bootstrapping state bridges or anything I guess

```nim
// <network-name>-<block-number>-<short-state-root>.era2
//
// era2 := Version | CompressedHeader | account*
// account :=  CompressedAccount | CompressedStorage*
// -----
// Version                     = { type: 0x3265, data: nil }
// CompressedHeader            = { type: 0x03,   data: snappyFramed(rlp(header)) }
// CompressedAccount           = { type: 0x08,   data: snappyFramed(rlp(address_hash, rlp(nonce, balance, storage_root, code_hash), raw_bytecode, storage_entry_count) } 
// CompressedStorage           = { type: 0x09,   data: snappyFramed(rlp(Vec<StorageItem { storage_index_hash, value }>)) }
// -----
// CompressedStorage can have a max of 10k storage items, records must be filled before creating a new one
```

### How was it fixed?

By implementing it, one of the following PR's will be implementing the export/import of these files in `trin-execution` the work for this is already done

### DIFF
--- a/e2store/src/era2.rs
+++ b/e2store/src/era2.rs
@@ -1,12 +1,12 @@
 use std::{
     io::{Read, Write},
-    path::PathBuf,
+    ops::Deref,
+    path::{Path, PathBuf},
 };
 
 use alloy_primitives::{hex, B256, U256};
 use alloy_rlp::{Decodable, RlpDecodable, RlpEncodable};
-use anyhow::ensure;
-use core::panic;
+use anyhow::{bail, ensure};
 use ethportal_api::types::{execution::header::Header, state_trie::account_state::AccountState};
 
 use crate::{
@@ -14,7 +14,7 @@ use crate::{
         stream::E2StoreStream,
         types::{Entry, VersionEntry},
     },
-    era1::HeaderEntry,
+    types::HeaderEntry,
 };
 
 // <network-name>-<block-number>-<short-state-root>.era2
@@ -25,12 +25,12 @@ use crate::{
 // Version                     = { type: 0x3265, data: nil }
 // CompressedHeader            = { type: 0x03,   data: snappyFramed(rlp(header)) }
 // CompressedAccount           = { type: 0x08,   data: snappyFramed(rlp(address_hash, rlp(nonce,
-// balance, storage_root, code_hash), raw_bytecode, storage_entry_count) }
+// balance, storage_root, code_hash), raw_bytecode, storage_entry_count)) }
 // CompressedStorage           = { type: 0x09,   data: snappyFramed(rlp(Vec<StorageItem {
 // storage_index_hash, value }>)) }
 // -----
-// CompressedStorage can have a max of 10k storage items, records must be filled before
-// creating a new one
+// CompressedStorage can have a max of 10 million storage items, records must be filled before
+// creating a new one, and must be sorted by storage_index_hash across all entries.
 
 /// Represents an era2 `Era2` state snapshot.
 /// Unlike era1, not all fields will be stored in the struct, account's will be streamed from an
@@ -41,12 +41,13 @@ pub struct Era2 {
 
     /// e2store_stream, manages the interactions between the era2 state snapshot
     e2store_stream: E2StoreStream,
-    storage_entries_left_to_decode: u16,
+    pending_storage_entries: u32,
+    path: PathBuf,
 }
 
 impl Era2 {
-    pub fn open(era2_path: &PathBuf) -> anyhow::Result<Self> {
-        let mut e2store_stream = E2StoreStream::open(era2_path)?;
+    pub fn open(path: PathBuf) -> anyhow::Result<Self> {
+        let mut e2store_stream = E2StoreStream::open(&path)?;
 
         let version = VersionEntry::try_from(&e2store_stream.next_entry()?)?;
         let header = HeaderEntry::try_from(&e2store_stream.next_entry()?)?;
@@ -55,26 +56,20 @@ impl Era2 {
             version,
             header,
             e2store_stream,
-            storage_entries_left_to_decode: 0,
+            pending_storage_entries: 0,
+            path,
         })
     }
 
-    pub fn create(era2_path: PathBuf, header: Header) -> anyhow::Result<Self> {
-        if era2_path.is_file() {
-            panic!(
-                "era2_path is not a directory, it is a file: {:?}",
-                era2_path
-            );
-        }
-        let era2_path = era2_path.join(format!(
+    pub fn create(path: PathBuf, header: Header) -> anyhow::Result<Self> {
+        ensure!(path.is_dir(), "era2 path is not a directory: {:?}", path);
+        let path = path.join(format!(
             "mainnet-{:010}-{}.era2",
             header.number,
             hex::encode(&header.state_root.as_slice()[..4])
         ));
-        if era2_path.exists() {
-            panic!("era2 file already exists: {:?}", era2_path);
-        }
-        let mut e2store_stream = E2StoreStream::create(&era2_path)?;
+        ensure!(!path.exists(), "era2 file already exists: {:?}", path);
+        let mut e2store_stream = E2StoreStream::create(&path)?;
 
         let version = VersionEntry::default();
         e2store_stream.append_entry(&version.clone().into())?;
@@ -86,41 +81,55 @@ impl Era2 {
             version,
             header,
             e2store_stream,
-            storage_entries_left_to_decode: 0,
+            pending_storage_entries: 0,
+            path,
         })
     }
 
+    pub fn path(&self) -> &Path {
+        self.path.as_path()
+    }
+
     #[allow(clippy::should_implement_trait)]
-    pub fn next(&mut self) -> anyhow::Result<Era2StreamEntry> {
-        if self.storage_entries_left_to_decode > 0 {
-            self.storage_entries_left_to_decode -= 1;
-            return Ok(Era2StreamEntry::Storage(StorageEntry::try_from(
+    pub fn next(&mut self) -> anyhow::Result<StreamEntry> {
+        if self.pending_storage_entries > 0 {
+            self.pending_storage_entries -= 1;
+            return Ok(StreamEntry::Storage(StorageEntry::try_from(
                 &self.e2store_stream.next_entry()?,
             )?));
         }
         let account = AccountEntry::try_from(&self.e2store_stream.next_entry()?)?;
-        self.storage_entries_left_to_decode = account.storage_count;
-        Ok(Era2StreamEntry::Account(account))
+        self.pending_storage_entries = account.storage_count;
+        Ok(StreamEntry::Account(account))
     }
 
-    pub fn append_entry(&mut self, entry: &Era2StreamEntry) -> anyhow::Result<usize> {
+    pub fn append_entry(&mut self, entry: &StreamEntry) -> anyhow::Result<usize> {
         let size = match entry {
-            Era2StreamEntry::Account(account) => {
-                if self.storage_entries_left_to_decode != 0 {
-                    panic!("Invalid append entry state: expected a storage entry, got an account entry. Still have {} storage entries left to append", self.storage_entries_left_to_decode);
-                }
+            StreamEntry::Account(account) => {
+                ensure!(
+                    self.pending_storage_entries == 0,
+                    "Invalid append entry state: expected a storage entry, got an account entry. Still have {} storage entries left to append", self.pending_storage_entries                
+                );
 
-                self.storage_entries_left_to_decode = account.storage_count;
+                self.pending_storage_entries = account.storage_count;
                 let entry: Entry = account.clone().try_into()?;
                 self.e2store_stream.append_entry(&entry)?;
                 entry.value.len()
             }
-            Era2StreamEntry::Storage(storage) => {
-                if self.storage_entries_left_to_decode == 0 {
-                    panic!("Invalid append entry state: expected an account entry, got a storage entry. No storage entries left to append for the account");
+            StreamEntry::Storage(storage) => {
+                match self.pending_storage_entries {
+                    0 => bail!("Invalid append entry state: expected an account entry, got a storage entry. No storage entries left to append for the account"),
+                    1 => ensure!(
+                        storage.len() <= 10_000_000,
+                        "Storage entry can't have more than 10 million items",
+                    ),
+                    _ => ensure!(
+                        storage.len() == 10_000_000,
+                        "Only last storage entry can have less than 10 million items",
+                    ),
                 }
 
-                self.storage_entries_left_to_decode -= 1;
+                self.pending_storage_entries -= 1;
                 let entry: Entry = storage.clone().try_into()?;
                 self.e2store_stream.append_entry(&entry)?;
                 entry.value.len()
@@ -131,7 +140,7 @@ impl Era2 {
 }
 
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub enum Era2StreamEntry {
+pub enum StreamEntry {
     Account(AccountEntry),
     Storage(StorageEntry),
 }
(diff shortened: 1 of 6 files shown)


morph-dev:
How much would you hate me if I would propose changes to the design of the era2 files? :sweat_smile: 

Why have 4 separate entries for the account, and not only one? My arguments:
 - They are always read and written together
 - The first 2 are tiny and the `bytecode` is very small in comparison to `storage`, so I don't think it adds much to the risk of making Entry too big when combined (compared to just storage)

KolbyML:
> How much would you hate me if I would propose changes to the design of the era2 files? 😅
> 
> Why have 4 separate entries for the account, and not only one? My arguments:
> 
> * They are always read and written together
> * The first 2 are tiny and the `bytecode` is very small in comparison to `storage`, so I don't think it adds much to the risk of making Entry too big when combined (compared to just storage)

It would be fine to combine the first 3 entries into one, but for storage if anything we should have 2 entries for it.

Well we haven't ran into a issue yet with storage, needing to make this change in the future and have multiple versions of .era adds work/complexity

KolbyML:
To be clear I don't know if a contract at block 20 million takes over 4GB of storage and if it does, we would need 2 entries for storage

morph-dev:
Yes, I agree that we should consider and support the case that storage entry is bigger than 4GB.
And as you said, we should have a way of splitting it into multiple entries.

One simple way to do it:
- have `AccountHeaderEntry` that will contain address_hash, account_state, bytecode, and number of storage entries
- have `StorageEntry` as you have now, but limit it to how many items it can hold

Regarding how many items to store in a storage entry, we can go with either:
1. maximum number that we know we can store in 4GB (should be somewhere around 63 millions)
2. Pick some nice number, like 1 million

I'm more in favor of the second one, as that gives reasonable max size per entry (max ~65 MB), which should cover big majority of accounts, and the overhead of the others is small (is it 8 bytes per entry?)

For the same reason, I think this should also be streamable (instead of current `Era2::next_account` which is not).

KolbyML:
Oh I think I get how the format should work so we could do an enum for storage

if 
- storage is less than 4gb include `CompressedAddressHash | CompressedAccountState | CompressedBytecode | StorageEnumThing(CompressedStorage)` in one entry
else
- have first entry be `CompressedAddressHash | CompressedAccountState | CompressedBytecode | StorageEnumThing(X amount of entries of storage)`, then the main account entry would be followed by X entries of storage.

> I'm more in favor of the second one, as that gives reasonable max size per entry (max ~65 MB), which should cover big majority of accounts, and the overhead of the others is small (is it 8 bytes per entry?)

In the current format Headers take 5.6GB for .era2 15 million, which is 7.24%.

I think the only reason for having smaller entry sizes would be the required amount of memory to import a .era2 file.

I think 1 million is too small, I don't see any major advantage saying if you were importing storage you would want them all anyways. So I could see maybe 1-2-3GB for max import GB of ram requirement reasons, but yeah.

But since if you are running a "full node" basically I think 4GB is fine honestly

KolbyML:
> Oh I think I get how the format should work so we could do an enum for storage
> 
> if
> 
> * storage is less than 4gb include `CompressedAddressHash | CompressedAccountState | CompressedBytecode | StorageEnumThing(CompressedStorage)` in one entry
>   else
> * have first entry be `CompressedAddressHash | CompressedAccountState | CompressedBytecode | StorageEnumThing(X amount of entries of storage)`, then the main account entry would be followed by X entries of storage.
> 
> > I'm more in favor of the second one, as that gives reasonable max size per entry (max ~65 MB), which should cover big majority of accounts, and the overhead of the others is small (is it 8 bytes per entry?)
> 
> In the current format Headers take 5.6GB for .era2 15 million, which is 7.24%.
> 
> I think the only reason for having smaller entry sizes would be the required amount of memory to import a .era2 file.
> 
> I think 1 million is too small, I don't see any major advantage saying if you were importing storage you would want them all anyways. So I could see maybe 1-2-3GB for max import GB of ram requirement reasons, but yeah.
> 
> But since if you are running a "full node" basically I think 4GB is fine honestly

I think this format works the best because 99.9% of accounts don't have storage anyways, and 99.9999% could fit in one entry, but this format scales to support any storage requirement and hence doesn't break if we ever have contracts which have more then 4GB of storage.

I really like it as it minimizes entries well not having a death switch to the format, like the current solution has

KolbyML:
@morph-dev ready for review with the updated format we discussed, I will prepare the next PR for you as well quick

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+/// Represents an era2 `Era2` state snapshot.
+/// Unlike era1, not all fields will be stored in the struct, account's will be streamed from an
+/// iterator as needed.
+pub struct Era2 {
+    pub version: VersionEntry,
+    pub header: HeaderEntry,
+
+    /// e2store_stream, manages the interactions between the era2 state snapshot
+    e2store_stream: E2StoreStream,
+    storage_entries_left_to_decode: u16,
```
morph-dev, line 44: nit: The `storage_entries_left_to_decode` doesn't make sense when we are writing to era2 file (because of the **decode**).
Maybe just rename it to `pending_storage_entries` or something like that.

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+            version,
+            header,
+            e2store_stream,
+            storage_entries_left_to_decode: 0,
+        })
+    }
+
+    pub fn create(era2_path: PathBuf, header: Header) -> anyhow::Result<Self> {
+        if era2_path.is_file() {
+            panic!(
```
morph-dev, line 64: Why panic and not return Err (e.g. using `ensure!`)?
Here and everywhere below.

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+                    panic!("Invalid append entry state: expected a storage entry, got an account entry. Still have {} storage entries left to append", self.storage_entries_left_to_decode);
+                }
+
+                self.storage_entries_left_to_decode = account.storage_count;
+                let entry: Entry = account.clone().try_into()?;
+                self.e2store_stream.append_entry(&entry)?;
+                entry.value.len()
+            }
+            Era2StreamEntry::Storage(storage) => {
+                if self.storage_entries_left_to_decode == 0 {
```
morph-dev, line 119: nit: Should we do something like this:

```rust
match self.storage_entries_left_to_decode {
    0 => bail!(...),
    1 => ensure!(
        storage.storage.len() <= 10_000_000,
        "Storage entry can't have more than 10 million items",
    ),
    _ => ensure!(
        storage.storage.len() == 10_000_000,
        "Only last storage entry can have less than 10 million items",
    ),
}
```

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+        let buf: Vec<u8> = vec![];
+        let mut encoder = snap::write::FrameEncoder::new(buf);
+        let _ = encoder.write(&rlp_encoded)?;
+        let encoded = encoder.into_inner()?;
+        Ok(Entry::new(0x08, encoded))
+    }
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub struct StorageEntry {
```
morph-dev, line 181: nit: maybe make this `pub struct StorageEntry(Vec<StorageItem>);`? (in which case, I would implement `Deref` trait)

If not, I would at least rename field to `items`?

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+
+        Ok(Self {
+            version,
+            header,
+            e2store_stream,
+            storage_entries_left_to_decode: 0,
+        })
+    }
+
+    pub fn create(era2_path: PathBuf, header: Header) -> anyhow::Result<Self> {
```
morph-dev, line 62: nit: Currently, there is no way to know what file we are writting to (unless we duplicate the logic for creating `era2_path`). I would just save the path in a field, and make it accessible from outside.

morph-dev (commented):
Good work. No major concerns.

Most comments are nitpicks and just suggestions (feel free to leave them as is if you don't agree with them).

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+                self.e2store_stream.append_entry(&entry)?;
+                entry.value.len()
+            }
+        };
+        Ok(size)
+    }
+}
+
+#[derive(Clone, Eq, PartialEq, Debug)]
+pub enum Era2StreamEntry {
```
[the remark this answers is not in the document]
KolbyML: > I'm also fine with leaving it as is, if you have strong preferences for this one.

I wouldn't say I have strong preferences on this. I also don't like the current name.

> How about we rename this to AccountEntry and rename current AccountEntry to AccountHeaderEntry

I don't like this because the current `AccountEntry` isn't really a header so it feels confusing

I will change `Era2StreamEntry` to `StreamEntry` for now. I don't really like `StreamEntry` too, if a good name is found I would be happy to rename it.

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+        Ok(Self {
+            version,
+            header,
+            e2store_stream,
+            storage_entries_left_to_decode: 0,
+        })
+    }
+
+    #[allow(clippy::should_implement_trait)]
+    pub fn next(&mut self) -> anyhow::Result<Era2StreamEntry> {
```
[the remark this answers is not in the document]
KolbyML: > How would we know that we finished reading the file

We get an error when we try to read an empty file

> nit: why not implement Iterator as clippy suggested?

Doesn't that require us to know how many entries there are? I don't think this is really possible, if you have a cleaver idea lets here it.

e2store/src/era2.rs
KolbyML: If this fails then it is corrupted or something extremely wrong happened and it isn't recoverable. I use panic because it is native and I am not sure it makes to pass the error up because the result will be the same.

I will change the panic!'s to ensure as I don't think it is a big difference

e2store/src/era2.rs
KolbyML: I forgot to add these checks thank you for reminding me! I really appreciate it

e2store/src/era2.rs
KolbyML: This is cool

KolbyML:
@morph-dev took a shot at resolving the concerns, ready for another look

e2store/src/era2.rs
```diff
@@ -0,0 +1,327 @@
...
+        );
+        let mut decoder = snap::read::FrameDecoder::new(&entry.value[..]);
+        let mut buf: Vec<u8> = vec![];
+        decoder.read_to_end(&mut buf)?;
+        let storage = Decodable::decode(&mut buf.as_slice())?;
+        Ok(Self(storage))
+    }
+}
+
+impl TryInto<Entry> for StorageEntry {
```
morph-dev, line 225: Should be TryFrom as well

e2store/src/era2.rs
```diff
@@ -164,22 +173,27 @@ impl TryFrom<&Entry> for AccountEntry {
...
         let _ = encoder.write(&rlp_encoded)?;
         let encoded = encoder.into_inner()?;
         Ok(Entry::new(0x08, encoded))
     }
 }
 
 #[derive(Clone, Eq, PartialEq, Debug)]
-pub struct StorageEntry {
-    pub storage: Vec<StorageItem>,
+pub struct StorageEntry(Vec<StorageItem>);
```
morph-dev, line 189: Now that I think about it, you will have to do one of those two:
- Make it `pub struct StorageEntry(**pub** Vec<StorageItem>);`
- Create constructor: `pub fn new(items: Vec<StorageItem>) -> Self {`

Otherwise, other files won't be able to create instances of it (which they will need).
But you can also add it once you have a need for it.

e2store/src/era2.rs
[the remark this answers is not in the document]
morph-dev: > We get an error when we try to read an empty file

My question is not about reading empty file, it was about reading regular file, but we finished reading the content (and there are no entries left).
This is valid case, and instead of returning error that we can't easily distinguish from other errors (e.g. disk corruption, wrong encoding, etc), I think we should return `Option` to indicate that there is no more entries to read.

> Doesn't that require us to know how many entries there are? I don't think this is really possible, if you have a cleaver idea lets here it.

Iterators in general don't know how many entries they have (you might be confusing them with slices). Iterators can even be infinite. You can read more about them [here](https://doc.rust-lang.org/std/iter/index.html).

e2store/src/era2.rs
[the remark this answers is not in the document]
morph-dev: Since neither of us likes `StreamEntry`, I will give few more suggestions but feel free to leave it as is if you don't like them:

1. rename `StreamEntry` to `AccountOrStorageEntry` (probably my favorite)
2. rename `StreamEntry` to `AccountEntry` and `AccountEntry` to `AccountInfoEntry` (since you don't like header)

morph-dev (approved):

morph-dev:
I would like to see some of the changes that I suggested (returning `Option` while reading), but it's fine if you don't do them now and we will see in future PR when you use it.
So feel free to address comments that you think make sense at this point.

:shipit:

e2store/src/era2.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+impl TryInto<Entry> for AccountEntry {
+    type Error = anyhow::Error;
+
+    fn try_into(self) -> anyhow::Result<Entry> {
+        let rlp_encoded = alloy_rlp::encode(self);
+        let buf: Vec<u8> = vec![];
+        let mut encoder = snap::write::FrameEncoder::new(buf);
+        let _ = encoder.write(&rlp_encoded)?;
```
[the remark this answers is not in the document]
KolbyML: I was more trying to say from reading the code it appears like it would write it all, but sure I can add the ensure

e2store/src/era2.rs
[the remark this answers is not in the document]
KolbyML: I think suggestion 1 is good so I will update it to that

e2store/src/era2.rs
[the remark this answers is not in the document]
KolbyML: Ok I implemented it, we will now panic if something fails, if we reach EOF we will return None

e2store/src/era2.rs
KolbyML: I will do this in the PR that uses this code then in that case

e2store/src/era2.rs
```diff
@@ -139,8 +127,51 @@ impl Era2 {
     }
 }
 
+impl Iterator for Era2 {
+    type Item = AccountOrStorageEntry;
```
morph-dev, line 131: I think it's better for `Item` to be `anyhow::Result<AccountOrStorageEntry>`.

I personally really don't like using `panic!`, because caller can no longer handle the failure.
And the `Era2` structure seems more like a library, so my feeling is that it should allow caller to handle errors.

e2store/src/era2.rs
```diff
@@ -139,8 +127,51 @@ impl Era2 {
...
+                Err(err) => panic!("Failed to decode next storage entry: {:?}", err),
+            };
+            return Some(AccountOrStorageEntry::Storage(storage_entry));
+        }
+
+        let raw_account_entry = match self.e2store_stream.next_entry() {
+            Ok(raw_account_entry) => raw_account_entry,
+            Err(err) => match err {
+                // If we read to the end of the error file we should get this
+                err if underlying_io_error_kind(&err).is_some()
```
morph-dev, line 153: it seems that this error handling should be done at the `E2StoreStream.next_entry()` level, in which case it should return `Result<Option<Entry>>` (or `Option<Result<Entry>>`, either would be fine there).

Issue #1305, opened 2024-05-22
State: `StateValidator` should verify that content belongs to the canonical chain


morph-dev:
At the moment, our [`StateValidator`](https://github.com/ethereum/trin/blob/94a7aea2241e6b92c523db61de3d46b82846b3e3/trin-state/src/validation/validator.rs#L25) doesn't verify that content value is part of the canonical chain.

The `StateContentValue` contains the `block_hash` that we should use to fetch block header. We should also verify that received header is part of the canonical chain.

From header we can get the `state_root` which is enough to use with the trie proof (which is already implemented).

morph-dev:
Fixed with #1377

PR #1377, opened 2024-08-13
feat(state): validate state content by anchoring it to the state root

morph-dev:
### What was wrong?

We currently don't anchor state content to the chain. We did this in the past as history network (and fetching block header) wasn't reliable, but we should start doing it.

Related issue: #1305 

### How was it fixed?

Fetch the block header from the network and validate the content based on the `state_root`.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2805,6 +2805,7 @@ name = "ethportal-peertest"
 version = "0.1.0"
 dependencies = [
  "alloy-primitives",
+ "alloy-rlp",
  "anyhow",
  "discv5",
  "ethereum_ssz",
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -258,6 +258,7 @@ impl TrinConfig {
     pub fn from_cli() -> Self {
         Self::new_from(env::args_os()).unwrap_or_else(|e| e.exit())
     }
+
     pub fn new_from<I, T>(args: I) -> Result<Self, clap::Error>
     where
         I: Iterator<Item = T>,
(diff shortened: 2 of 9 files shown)


morph-dev:
Ideally, this should be merged after https://github.com/ethereum/hive/pull/1152 (otherwise, hive tests will start failing).

KolbyML:
I don't think we should enable to until we have history live covering 100%

KolbyML:
One concern is what if the user isn't running the history network, there doesn't appear to be any  measures in place to prevent against this error case

KolbyML:
If we did want to enable it, we should enable it by block ranges over time

morph-dev:
Considering that in short/mid term we only plan to gossip only first 1'000'000 blocks and the most recent ones (head state), we are already covered with history network (4444 + latest). And for anything else, having history at 100% should be higher priority than having state at 100% (and also much easier).

In addition, I would say it makes sense for state bridge to first gossip block header to history network, then process that block (execute + gossip), and we avoid this problem altogether.

While I think I gave good arguments, I'm also fine with having a startup flag that can disable this check (or at least attempt it, and if we can't fine the header consider that proof is anchored). But I would leave the check enabled by default (at least for now).

---

> One concern is what if the user isn't running the history network, there doesn't appear to be any measures in place to prevent against this error case

This is good argument, but we will always have this issue (now or later). There are few simple ways around it:
- my preference: print error message at startup (if state is enabled and history is not)
- run history network but without  storage.

KolbyML:
> > One concern is what if the user isn't running the history network, there doesn't appear to be any measures in place to prevent against this error case
> 
> This is good argument, but we will always have this issue (now or later). There are few simple ways around it:
> 
> * my preference: print error message at startup and shut down
> * run history network but without  storage.

History requires beacon for validating all post capella content

So we also need to run beacon

KolbyML:
> Considering that in short/mid term we only plan to gossip only first 1'000'000 blocks and the most recent ones (head state), we are already covered with history network (4444 + latest). And for anything else, having history at 100% should be higher priority than having state at 100% (and also much easier).
> 
> In addition, I would say it makes sense for state bridge to first gossip block header to history network, then process that block (execute + gossip), and we avoid this problem altogether.
> 
> While I think I gave good arguments, I'm also fine with having a startup flag that can disable this check (or at least attempt it, and if we can't fine the header consider that proof is anchored). But I would leave the check enabled by default (at least for now).

In like 4 months when history requires beacon for post capella blocks and latest for validation of headers, if we ever want to test state we would then need to have the beacon and history network running.

Maybe this won't be a issue I guess we will see, def a lot more requirements but I that is by design

> In addition, I would say it makes sense for state bridge to first gossip block header to history network, then process that block (execute + gossip), and we avoid this problem altogether.

We can do this

KolbyML:
> > One concern is what if the user isn't running the history network, there doesn't appear to be any measures in place to prevent against this error case
> 
> This is good argument, but we will always have this issue (now or later). There are few simple ways around it:
> 
> * my preference: print error message at startup and shut down
> * run history network but without  storage.

I think the first option should be included in this PR in that case

KolbyML (approved):
:shipit:

ogenev (approved):
LGTM 👍 .

morph-dev:
Final comment. This pr should be merged only after:
- https://github.com/ethereum/trin/pull/1388 is merged
- cluster repo is updated such that state nodes have both history and state subnetworks enabled
    - I have [branch ready](https://github.com/ethereum/cluster/compare/master...morph-dev:cluster:portal_state_nodes), but it should be sent for review and only after PR from above

morph-dev:
Rebased to head and cleaned up commit history.

Will most likely merge it tomorrow morning, so I can react if something goes wrong (e.g. hive starts failing).

ogenev (approved):

PR #1378, opened 2024-08-13
feat: create json-rpc mock builder

morph-dev:
### What was wrong?

There is no easy way to mock our json-rpc request/responses which makes unit test of certain parts of the codebase very challenging.

### How was it fixed?

Created json-rpc mock builder, that is very customization.

See example of the real PR use case [here](https://github.com/ethereum/trin/blob/625839ee127798cdc6b1ffaa14a86366c1732503/trin-state/src/validation/validator.rs#L224C1-L229C24)

That case can be simplified even further (since it only cares about one response) in the following way:
```rust
let history_jsonrpc_tx = MockJsonRpcBuilder::new_builder().or_else(content_info);
```

For more complex logic, one can do something like this:
```rust
let history_jsonrpc_tx = MockJsonRpcBuilder::new_builder()
    // respond to RecursiveFindContent with one respond
    .with_response(
        HistoryEndpoint::RecursiveFindContent(content_key_1),
        content_info_1
    )
    // to another content key with another
    .with_response(
        HistoryEndpoint::RecursiveFindContent(content_key_2),
        content_info_2
    )
    // to gossip with completely different type
    .with_response(
        HistoryEndpoint::Gossip(content_key, content_value),
        42 // number of nodes that received gossip
    )
    // even custom trigger that handles all Ping requests
    .with_custom_trigger(
        |request| matches!(request, HistoryEndpoint::Ping(_)), // respond the same for every ping request
        pong_info,
    )
    // fail otherwise
    .of_fail();
```

### To-Do

### DIFF
--- a/ethportal-api/src/types/jsonrpc/json_rpc_mock.rs
+++ b/ethportal-api/src/types/jsonrpc/json_rpc_mock.rs
@@ -1,0 +1,91 @@
+use serde::Serialize;
+use serde_json::Value;
+use tokio::sync::mpsc;
+
+use super::request::JsonRpcRequest;
+
+type FilterFn<T> = Box<dyn Fn(&T) -> bool + Send>;
+
+/// Defines whether and how to respond to the request.
+struct Interaction<T> {
+    request_selector_fn: FilterFn<T>,
+    response: Result<Value, String>,
+}
+
+impl<T> Interaction<T> {
+    /// Returns `Some(response)` if selector returns `true`, `None` otherwise.
+    fn maybe_respond(&self, request: &T) -> Option<Result<Value, String>> {
+        if (self.request_selector_fn)(request) {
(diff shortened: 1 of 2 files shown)


morph-dev:
I'm not fully set on this approach as final version. I'm looking for the early feedback and thoughts.
If there are no major objections, I'm also fine with merging it like this and we can iterate as we see use cases.

KolbyML (approved):
I don't think we should or try to get it right the first time as that is idealism. If this will help us write good tests and works I think that is good for now. If it doesn't fit our needs in the future we can fix that.

Since this isn't an architecture change which determines the fate of the whole project.

I would be a fan of merging on it and iterating as required.

:shipit:

ogenev (approved):
Looks great to me 💯 !

PR #1379, opened 2024-08-13
fix: temporarily skip caching non-portal nodes in discovery

njgheorghita:
### What was wrong?
Temporary solution to validate the theory that eth2 nodes are overflowing the `node_addr_cache` on mainnet. Once the theory is validated/invalidated, this solution will be replaced with a more sustainable solution, since this solution will result in the exact same issue once the total number of portal nodes on the network exceeds current cache size.

### How was it fixed?
- only cache enrs that have a `"c"` enr key - indicating that they are participating in the portal network

### To-Do


KolbyML (approved):
:shipit:

PR #1381, opened 2024-08-15
fix(state-bridge): if there is no storage diff to gossip continue

KolbyML:
### What was wrong?
If storage was empty we would still attempt to parse the partial storage trie, but of course that isn't possible
### How was it fixed?
by adding if statements if the storage is empty skip trying to parse the trie

### DIFF
--- a/trin-execution/src/trie_walker.rs
+++ b/trin-execution/src/trie_walker.rs
@@ -1,7 +1,9 @@
 use alloy_primitives::{Bytes, B256};
+use alloy_rlp::EMPTY_STRING_CODE;
 use anyhow::ensure;
 use eth_trie::{decode_node, node::Node};
 use hashbrown::HashMap as BrownHashMap;
+use revm_primitives::keccak256;
 use serde::{Deserialize, Serialize};
 
 use super::types::trie_proof::TrieProof;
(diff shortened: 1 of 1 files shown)


KolbyML:
The code would work with only the second if case, but no harm in having both

morph-dev:
Can you be specific on what exactly was crashing/failing?

I feel like checks are done in the wrong place. Instead, I feel like `TrieWalker` should just return create empty set, as there shouldn't be anything to walk over. What do you think?

KolbyML:
if the storage diff is empty a error will be thrown that the provided trie root is missing in the provided dataset.

I mean sure I can do the suggestion it does feel weird calling a function on nothing though, but it isn't a big deal

KolbyML:

trin-execution/src/trie_walker.rs
```diff
@@ -34,6 +36,17 @@ pub struct TrieWalker {
 
 impl TrieWalker {
     pub fn new(root_hash: B256, nodes: BrownHashMap<B256, Vec<u8>>) -> Self {
+        // if the storage root is empty then there is no storage to gossip
+        if root_hash == keccak256([EMPTY_STRING_CODE]) && !nodes.is_empty() {
+            panic!("Root hash is empty but there are nodes to gossip. This should never happen.");
```
morph-dev, line 41: If you are sure this shouldn't happen, leave it like this.
Otherwise change it to logging error and return empty map (same as below).

morph-dev (approved):

trin-execution/src/trie_walker.rs
KolbyML: This should be impossible if it did happen there is def a big mistake in the codebase

PR #1382, opened 2024-08-15
feat: add non-recursive gossip logic to state bridge

KolbyML:
https://github.com/ethereum/trin/pull/1381 review this first
### What was wrong?
Our state bridge didn't gossip all iterations of a proof to the network, since recursive gossip was removed this is now required
### How was it fixed?

Making a iterator which iterates over the full proof then we just gossip all the iterations

### DIFF
--- a/portal-bridge/src/bridge/state.rs
+++ b/portal-bridge/src/bridge/state.rs
@@ -153,8 +153,10 @@ impl StateBridge {
                 let account_proof = walk_diff.get_proof(*node);
 
                 // gossip the account
-                self.gossip_account(&account_proof, block_tuple.header.header.hash())
-                    .await?;
+                for proof in walk_diff.iter_proof(*node) {
+                    self.gossip_account(&proof, block_tuple.header.header.hash())
+                        .await?;
+                }
 
                 let Some(encoded_last_node) = account_proof.proof.last() else {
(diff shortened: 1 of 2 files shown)


trin-execution/src/types/trie_proof.rs
```diff
@@ -8,6 +8,15 @@ pub struct TrieProof {
     pub proof: Vec<Bytes>,
 }
 
+impl TrieProof {
+    pub fn iter(trie_proof: TrieProof) -> impl Iterator<Item = TrieProof> {
+        (0..trie_proof.proof.len()).map(move |i| Self {
+            path: trie_proof.path[0..i].to_vec(),
```
morph-dev, line 14: I don't think it is so simple as you will have extension nodes on the way, which contribute more than one nibble to the path... I think we should leverage TrieWalker, as I believe it already takes this into consideration.

morph-dev (changes requested):
I don't think this will work. See my comment.

trin-execution/src/types/trie_proof.rs
KolbyML: I forgot thank you for reminding me I will fix it

KolbyML:
I am actually ready now!

KolbyML:
I just realized that the code already gossips all the intermidate tries, but I think I fixed a bug in this code regardless so I will close this PR and open another

PR #1384, opened 2024-08-16
refactor: simplify trie_walker get_proof code

KolbyML:
### What was wrong?
the reverse logic for path was a little overly complex
### How was it fixed?
by making the logic less complex

### DIFF
--- a/trin-execution/src/trie_walker.rs
+++ b/trin-execution/src/trie_walker.rs
@@ -1,4 +1,6 @@
-use alloy_primitives::{Bytes, B256};
+use std::collections::VecDeque;
+
+use alloy_primitives::B256;
 use alloy_rlp::EMPTY_STRING_CODE;
 use anyhow::ensure;
 use eth_trie::{decode_node, node::Node};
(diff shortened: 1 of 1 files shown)


trin-execution/src/trie_walker.rs
```diff
@@ -140,24 +140,22 @@ impl TrieWalker {
 
     pub fn get_proof(&self, node_hash: B256) -> TrieProof {
         // Build path and proof from the node to the root and reverse it at the end.
-        let mut reverse_path = vec![];
+        let mut reverse_path_parts = vec![];
```
morph-dev, line 143: How about using [VecDeque](https://doc.rust-lang.org/std/collections/struct.VecDeque.html). Something like this:

```rust
pub fn get_proof(&self, node_hash: B256) -> TrieProof {
    let mut path_parts = VecDeque::new();
    let mut proof = VecDeque::new();
    let mut next_node: Option<B256> = Some(node_hash);
    while let Some(current_node) = next_node {
        let Some(node) = self.nodes.get(&current_node) else {
            panic!("Node not found in trie walker nodes. This should never happen.");
        };
        path_parts.push_front(node.path_nibbles.clone());
        proof.push_front(node.encoded_node.clone().into());
        next_node = node.parent_hash;
    }

    TrieProof {
        path: Vec::from(path_parts).concat(),
        proof: Vec::from(proof),
    }
}
```

In my opinion, it's easier to follow logic without tracking what is reversed and what not. It's up to you (i'm fine with merging as is as well).

morph-dev (approved):

Issue #1387, opened 2024-08-19
Configuring multiple subnetworks


morph-dev:
Configuring trin to run multiple subnetworks is currently not ideal for a couple of reasons:

## Subnetwork selection

We have a flag for enabling subnetworks (`portal-subnetworks`), but not all combination should be valid.

Once #1377 is merged, state network won't be able to run without history network. In a few months, history network won't be able to run without beacon network.

Potential solutions:

1. Crash execution at the start if flags are not configured correctly (approach I took in #1377 )
2. Introduce different flag (e.g. `mode`) that enables all required subnetworks
    - For example, in the short/medium term, valid values could be `history` (beacon+history) and `state` (beacon+history+state)
        - Not sure if it makes sense to run beacon only, but we could add it if it does
    - This can get more complicated when we add more subnetworks (e.g. transactions), but we an easily support multiple modes (union of all subnetworks)

**Side note:** Validating state network requires having history network, but we don't need to store any history data ourself. However, I don't think we want to go this path (see section below).

## Storage

We have only one flag (`mb`) for configuring the storage capacity. Currently, it applies on all subnetworks separately. Meaning, if user enables all 3 networks, and specifies 1GB, we will actually set capacity of 1GB per subnetwork (meaning 3GB in total).

In the context of beacon network, my understanding is that storage capacity doesn't make sense. Firstly, it uses little storage to begin with (compare to history and state), and secondly it will never prune or do anything to stay under capacity, no matter how small it is. @ogenev is this correct?

Potential solutions:

1. Have a way to configure capacity per network
    - While this gives more granular control, it can also be complicated for the end user
2. Keep only one flag, but it should mean total capacity. It will be up to us to decide how much capacity each network has. Some ideas:
    1.  Use the same radius for all subnetworks
        - Nice and clean, but might require bigger refactoring of the current code (which we might need to do to support "big radius node" anyway)
    2. Have a fixed weight of how much capacity should be allocated for each subnetwork
        - Only consider weights of the enabled subnetworks
    3. Have a combinations of fixed weight that are dependent on the `mode` from section above (if we agree on that approach)
        - For example, let's consider two scenarios: **A: both `history` and `state` mode** and **B: only `state` mode**. In both scenarios, we would run all 3 subnetworks, but more storage would be allocated for state subnetwork in scenario B.

[...]

PR #1388, opened 2024-08-19
feat: use weight for setting storage capacity per network

morph-dev:
### What was wrong?

Currently, we only have one flag to setting storage capacity that applies to each subnetwork separately. See #1387 for more info.

### How was it fixed?

As discussed in the meeting, we will use predefined weight per subnetwork (for the short/medium term).

I put following weights as a placeholder. Let me know if you have other suggestions:
- history: 1
- state: 99
- beacon: 0 (beacon doesn't use capacity in any meaningful way (in only reports it to the metrics). 

### To-Do

### DIFF
--- a/ethportal-peertest/src/scenarios/basic.rs
+++ b/ethportal-peertest/src/scenarios/basic.rs
@@ -135,8 +135,8 @@ pub async fn test_ping(protocol: ProtocolId, target: &Client, peertest: &Peertes
 
 pub async fn test_ping_cross_network(mainnet_target: &Client, angelfood_node: &PeertestNode) {
     info!("Testing ping for history cross mainnet and angelfood discv5 protocol id");
-    let bootnode_enr = angelfood_node.enr.clone();
-    if let Ok(pong) = HistoryNetworkApiClient::ping(mainnet_target, bootnode_enr).await {
+    let angelfood_enr = angelfood_node.enr.clone();
+    if let Ok(pong) = HistoryNetworkApiClient::ping(mainnet_target, angelfood_enr).await {
         panic!("Expected ping to fail as mainnet/angelfood history nodes shouldn't be able to communicate {pong:?}");
     };
 }
--- a/ethportal-peertest/src/scenarios/basic.rs
+++ b/ethportal-peertest/src/scenarios/basic.rs
@@ -1,4 +1,4 @@
-use crate::{utils::fixture_header_with_proof, Peertest};
+use crate::{utils::fixture_header_with_proof, Peertest, PeertestNode};
 use alloy_primitives::{B256, U256};
 use ethportal_api::{
     types::{distance::Distance, portal_wire::ProtocolId},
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
:shipit:

ogenev:
@morph-dev Some of the tests are failing, could you fix them before a review, please?

morph-dev:
> @morph-dev Some of the tests are failing, could you fix them before a review, please?

Sorry about that. Some peertests were failing and the fix wasn't the simplest so it took me some time (some were relying on specific storage capacity). I ended up refactoring how peertests are initialized by allowing for custom set of subnetworks (instead of all 3 all the time).

I also noticed that capacity weigh should be applied on bytes instead of MB, so I fixed that as well.

ethportal-peertest/src/scenarios/basic.rs
```diff
@@ -133,10 +133,10 @@ pub async fn test_ping(protocol: ProtocolId, target: &Client, peertest: &Peertes
     assert_eq!(result.enr_seq, 1);
 }
 
-pub async fn test_ping_cross_network(target: &Client, peertest: &Peertest) {
+pub async fn test_ping_cross_network(mainnet_target: &Client, angelfood_node: &PeertestNode) {
     info!("Testing ping for history cross mainnet and angelfood discv5 protocol id");
-    let bootnode_enr = peertest.bootnode.enr.clone();
-    if let Ok(pong) = HistoryNetworkApiClient::ping(target, bootnode_enr).await {
+    let bootnode_enr = angelfood_node.enr.clone();
```
njgheorghita, line 138: *nitpick* maybe `angelfood_target` instead of `bootnode_enr`

trin-storage/src/config.rs
```diff
@@ -0,0 +1,81 @@
...
+            sql_connection_pool: self.sql_connection_pool.clone(),
+        }
+    }
+
+    fn get_capacity_weight(subnetwork: &str) -> u64 {
+        match subnetwork {
+            HISTORY_NETWORK => Self::HISTORY_CAPACITY_WEIGHT,
+            STATE_NETWORK => Self::STATE_CAPACITY_WEIGHT,
+            BEACON_NETWORK => Self::BEACON_CAPACITY_WEIGHT,
+            _ => panic!("Invalid subnetwork: {subnetwork}"),
```
njgheorghita, line 69: Is it worth creating a `Subnetwork` enum type? Something like [this](https://github.com/ethereum/trin/blob/master/ethportal-api/src/types/cli.rs#L32) that will cover [these](https://github.com/ethereum/trin/blob/master/ethportal-api/src/types/cli.rs#L19-L21)? So we can skip all the string ser/de?

njgheorghita: And if so, maybe also for `Network` (mainnet/angelfood)... working with enums is a bit nicer than using raw strings in various places

trin-storage/src/config.rs
```diff
@@ -0,0 +1,81 @@
...
+pub struct PortalStorageConfigFactory {
+    node_id: NodeId,
+    node_data_dir: PathBuf,
+    total_capacity_mb: u64,
+    total_capacity_weight: u64,
+    sql_connection_pool: Pool<SqliteConnectionManager>,
+}
+
+impl PortalStorageConfigFactory {
+    const HISTORY_CAPACITY_WEIGHT: u64 = 1;
```
njgheorghita, line 22: Re: 99/1 (state/history capacity)....
I guess I don't have any good, sound reasoning for why it shouldn't be that, just that this seems a little high... But, it is true that the network has much higher storage demands for state than history data, and as long as we don't update our current mainnet 4444s nodes to support state network, then 99/1 is fine by me....

Though, I wonder if we want to introduce some concept of a minimum (total) capacity? Idk, like `100mb`? (though, the only repercussion I can think of is that this minimum would break [this test](https://github.com/ethereum/trin/blob/master/ethportal-peertest/src/scenarios/gossip.rs#L298))

njgheorghita (approved):
🚀

trin-storage/src/config.rs
morph-dev: I was thinking the same, but I want to get this pr asap (so I can get another pr before the release this week). So I will leave this for the future (created #1392 ).

trin-storage/src/config.rs
morph-dev: We should definitely not enable state network on the 4444 nodes. That would be bad regardless of the ratio that we pick here.
With that in mind, I will leave the ratio as 99/1 for now (we can easily change it if there is need for it).

I don't have strong opinion about enforcing minimum capacity. It's something we can discuss in the next meeting. But I would leave it out of this pr.

Issue #1390, opened 2024-08-20
Deadlock with last log: "DEBUG portalnet::gossip: propagating validated content ids=["0x.."]"


morph-dev:
While on flamingo rotation, I noticed on glados that several machines are stuck.

I checked their logs and the very last 2 log messages on them were:

```
DEBUG discv5::service: Received RPC response: Response: Response 050108000000<redacted> to request: TALK: protocol: 500b, request: 040400000000<redacted> from: Node: 0x<redacted>, addr: <redacted>:9009
DEBUG portalnet::gossip: propagating validated content ids=["0x<redacted>"]
```
(the `<redacted>` wasn't the same but it seems irrelevant)

After restarting the docker images, they keep working fine.

It seems to me that there is some deadlock happening, most likely in the same place (and most likely during gossiping), but further investigation is needed.

morph-dev:
These lines in [portalnet/src/gossip.rs](https://github.com/ethereum/trin/blob/b4a81e34c01e24a8bdb103601cdc47e649fb3575/portalnet/src/gossip.rs#L98C62-L98C84) look like a potential candidates for the deadlock:

```rust
let permit = match utp_controller {
    Some(ref utp_controller) => match utp_controller.get_outbound_semaphore() {
        Some(permit) => Some(permit),
        None => continue,
    },
    None => None,
};
```

KolbyML:
> These lines in [portalnet/src/gossip.rs](https://github.com/ethereum/trin/blob/b4a81e34c01e24a8bdb103601cdc47e649fb3575/portalnet/src/gossip.rs#L98C62-L98C84) look like a potential candidates for the deadlock:
> 
> 
> 
> ```rust
> 
> let permit = match utp_controller {
> 
>     Some(ref utp_controller) => match utp_controller.get_outbound_semaphore() {
> 
>         Some(permit) => Some(permit),
> 
>         None => continue,
> 
>     },
> 
>     None => None,
> 
> };
> 
> ```

try_acquire_owned() doesn't block so that rules out that candidate

morph-dev:
Two more nodes got stuck, containing my last PR with log messages, so I was able to conclude that deadlock is happening at this line: [portalnet/src/gossip.rs#L61](https://github.com/ethereum/trin/blob/85239131515ad6820b2a7a96aedd25fa110eca48/portalnet/src/gossip.rs#L61)
```rust
let kbuckets = kbuckets.read();
```

[...]

PR #1391, opened 2024-08-20
feat: add debug logging to gossip logic

morph-dev:
### What was wrong?

There is a deadlock happening, mostly likely during gossip logic. See: #1390 

### How was it fixed?

Adding more debug statements to narrow down where the deadlock is happening.

I also noticed that `calculate_interested_enrs` function can be improved and simplified.

### To-Do


KolbyML (approved):
:shipit:

PR #1393, opened 2024-08-21
fix(state-bridge): storage trie changes weren't being gossiped

KolbyML:
### What was wrong?
Storage trie changes weren't being gossiped due to fetching the storage trie diff code being incorrect.

There was also another problem apparently early on in Ethereum Storage could be modified well "creating" a new contract without changing the contracts code
### How was it fixed?

- if the contract is empty don't skip gossiping storage trie changes
- correctly fetch storage trie nodes from the database

### DIFF
--- a/trin-execution/src/storage/evm_db.rs
+++ b/trin-execution/src/storage/evm_db.rs
@@ -4,7 +4,7 @@ use crate::{config::StateConfig, storage::error::EVMError};
 use alloy_consensus::constants::KECCAK_EMPTY;
 use alloy_primitives::{Address, B256, U256};
 use alloy_rlp::{Decodable, EMPTY_STRING_CODE};
-use eth_trie::{EthTrie, RootWithTrieDiff, Trie};
+use eth_trie::{EthTrie, RootWithTrieDiff, Trie, DB};
 use ethportal_api::{
     types::state_trie::account_state::AccountState as AccountStateInfo, utils::bytes::hex_encode,
 };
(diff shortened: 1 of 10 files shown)


trin-execution/src/storage/evm_db.rs
```diff
@@ -86,17 +86,29 @@ impl EvmDB {
...
+    pub fn get_storage_trie_diff(
+        &self,
+        storage_root: B256,
+        address_hash: B256,
+    ) -> BrownHashMap<B256, Vec<u8>> {
         let mut trie_diff = BrownHashMap::new();
 
+        let account_db = AccountDB::new(address_hash, self.db.clone());
+        let trie = if storage_root == keccak256([EMPTY_STRING_CODE]) {
+            EthTrie::new(Arc::new(account_db))
```
morph-dev, line 98: Can we just return empty result here?

If that's the case, I would say you can revert most of the changes in this file? Or am I missing something?

morph-dev (commented):
I understand that code can be empty and we still need to gossip storage trie.

But I can't say I understand what was wrong with fetching storage trie nodes from the database logic and how this PR fixes it. Can you elaborate it a bit more on this, or maybe we can have a call?

KolbyML:
> I understand that code can be empty and we still need to gossip storage trie.
> 
> But I can't say I understand what was wrong with fetching storage trie nodes from the database logic and how this PR fixes it. Can you elaborate it a bit more on this, or maybe we can have a call?

I just made a commit which should make it a little more clear

Storage trie nodes stored in the database is prefixed by address_hash

trin-execution/src/storage/evm_db.rs
KolbyML: This no longer applies I will click resolve, but feel free to bring it up again if you still think there is a issue

morph-dev (approved):
:ship:

PR #1394, opened 2024-08-21
feat: add jsonrpc endpoint for offer with trace

njgheorghita:
### To-Do


ethportal-api/src/beacon.rs
```diff
@@ -114,6 +114,18 @@ pub trait BeaconNetworkApi {
...
+    /// Does not store the content locally.
+    /// Returns true if the content was accepted and successfully transferred,
+    /// returns false if the content was not accepted or the transfer failed.
+    #[method(name = "traceOffer")]
+    async fn trace_offer(
+        &self,
+        enr: Enr,
+        content_key: BeaconContentKey,
+        content_value: BeaconContentValue,
+    ) -> RpcResult<bool>;
```
njgheorghita, line 127: It's possible that in the future it might be useful to know whether or not the `false` case is because the `Accept` was declined or the utp transfer failed, but for the time being a simple `bool` is already useful in the new bridge, and we can update the response type later if a more granular response is warranted.

KolbyML (approved):
:shipit:

PR #1395, opened 2024-08-24
Continue the RecursiveFindContent query until the uTP transfer is complete and the content is validated

### LINKED ISSUE #1383: Robust RecursiveFindContent | Continue the search if data transfer fails
image

Clients can have bugs, trin has had bugs, shisui has, ultralight etc. A bug in a client shouldn't stop the whole show. A bug we were seeing was trin found the content from a shisui or trin node, then the uTP transfer would fail for some unknown reason, then we would just return an error.

Instead of this we could continue the search until we find a working peer, we already did a big search to find the contents neighbourhood so it isn't out of question if the first node we try to get the data from fails, we just try another and continue the search to some failure level till we either fail X times and give up or successfully fetch the data.

carver:
### What was wrong?

Work toward #1383 

### How was it fixed?

This PR *will* delay completion of RFC until data is validated, but will *not* continue the search with other peers. I decided it was getting too big. I will do the follow-up work in another PR.

There are quite a few moving parts here, so the goal was to minimize the
number of functional changes in this refactor.

Overall: add a new query state for while the content is being validated.
Don't exit the query until validation is complete.

Some other changes along the way:

#### Moving the bigendian conversion of connection ID

Note that when handling the FindContentQueryResponse::ConnectionId in
findcontent.rs, we don't do the bigendian conversion anymore. That has
been moved to earlier in the process, when handling the Content message,
to reduce the number of places that the conversion needs to be handled.

#### The query trace has slightly different end-stage behavior

The received_from peer and the cancelled peers are not set in the query
until the content gets validated.

#### Naturally, tests need to reflect the new query structure

For example:
- the query trace test needs to have the content be marked validated to
  progress
- update overlay service tests to use new Validating query step
- immediately mark content as verified, after marking it as received,
  during the test

#### bugfix: the content search continues even if many peers don't have it

Formerly, the content query would exit out if too many peers replied
successfully (even if the responses were just more nodes to look up).

Best we can tell, the limit on the number of maximum results is
copy-pasta from the recursive-find-nodes logic. (Confirmation from
Mike)

Now, the query continues until the content is found. It doesn't track
the number of successful responses at all.

(Perhaps worth noting that tests fail without this last change. I didn't look that hard into why they weren't failing before)

### Refactor

At first, I was trying to change as little as possible. That meant that I kind of shoe-horned the pending result into the Query::Result enum. It works, but is ugly, because of having to handle the pending results in some logic, and the final results in other logic. 

So e0e54378 goes a little further and adds a new Pending result enum. It still leaves us with some assertions that certain code paths should be unreachable, but I don't see any better alternatives. I explored another Query trait type, but it required duplicating too much of `query_event_poll` logic. So I landed on this approach, largely implemented by @morph-dev 

### To-Do

- [x] trin compiles locally
- [x] Manual gut-check that RFC still works on the happy path (when this PR was first opened, it did not yet)
  - [x] always delay Finished until validation succeeds (previously it was exploring all nodes, then exiting immediately)
  - [x] Fix uTP transfer. I seem to have broken it, because it never connects successfully now. Last I checked, it was still working fine on master
- [x] Constrain active number of downloads/validations to 1
  - [x] hacky version that only allows one concurrent
  - [x] stop querying peers for content when there are active validations
  - [x] move the follow-ups to the tracking issue
- [x] Tests
  - [x] ethportal-api
  - [x] overlay service
  - [x] findcontent
  - [x] findnodes
  - [x] peertest
- [x] The FindContent queries shouldn't have a maximum number of results, it should just run till completion. In findcontent.rs, see `if *count >= self.config.num_results {`
- [x] Rebase on #1447 
  - [x] after 1447 tests pass
  - [x] after merge
- [x] After addressing PR feedback, clean up commit history and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1561,6 +1561,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-channel"
+version = "0.5.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33480d6946193aa8033910124896ca395333cae7e2d1113d1fef6c3272217df2"
+dependencies = [
+ "crossbeam-utils",
+]
+
+[[package]]
 name = "crossbeam-deque"
 version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4519,6 +4528,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "bytes",
+ "crossbeam-channel",
  "delay_map 0.4.0",
  "directories",
  "discv5",
--- a/ethportal-api/src/types/query_trace.rs
+++ b/ethportal-api/src/types/query_trace.rs
@@ -15,7 +15,7 @@ type ContentId = [u8; 32];
 #[derive(Debug, Serialize, Deserialize, Clone, Eq, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct QueryTrace {
-    /// Node ID from which the content was received. None if the content was not found.
+    /// Node ID from which the content was received. None if the content was not found & verified.
     pub received_from: Option<NodeId>,
     /// The local node.
     pub origin: NodeId,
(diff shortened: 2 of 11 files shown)


mrferris (approved):
Code LGTM, I see the vision. Good find on the unnecessary `num_results` abort. 

Tested on a local glados instance and saw virtually same results as with trin deployed to main glados instance.

Issue #1333, opened 2024-06-29
trin-execution: execute block in memory and commit it at the end


KolbyML:
https://github.com/ethereum/trin/pull/1331#discussion_r1659750741

We talked on a call

something like this will be used to solve this problem https://bluealloy.github.io/revm/docs/revm/db/states/state/struct.State.html

It is a great improvement, but we shouldn't prematurely add complexity, unless required to continue as complexity can increase the complexity of fixing bugs or finding them, which is bad if we don't already have a naive working solution to compare off when things go wrong

PR #1397, opened 2024-08-24
fix: trie_walker shouldn't assume that all nodes are useful

morph-dev:
### What was wrong?

The `TrieWalker` can't assume that all provided nodes are useful and it shouldn't check for it.

Ideally, TrieWalker should be able to make such assumption, but not with the current implementation. Currently we write to db after every transaction, and cache all contract storage writes. When the same contact storage is updated in different transactions within the same block, this results in cache containing nodes that are no longer valid.

This can be solved by keeping all state changes in memory and writing them at the end of the block (instead of after every transaction). This was already planned improvement: #1333 .

### How was it fixed?

As we are working on the proper fix (#1337), this PR disables the check.
Also adding test that executes and creates content key/value pairs up to a given block (disabled by default as it takes too long to run for anything meaningful)

### To-Do

### DIFF
--- a/trin-execution/src/trie_walker.rs
+++ b/trin-execution/src/trie_walker.rs
@@ -2,7 +2,6 @@ use std::collections::VecDeque;
 
 use alloy_primitives::B256;
 use alloy_rlp::EMPTY_STRING_CODE;
-use anyhow::ensure;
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
:shipit:

PR #1399, opened 2024-08-25
feat: implement eth_getBalance, eth_getCode and eth_getStorageAt

morph-dev:
### What was wrong?

The `eth_getBalance`, `eth_getCode` and `eth_getStorageAt` weren't implemented.

### How was it fixed?

Implemented them by fetching data from history (block header) and state network (traversing tries).

I also tested it locally by running trin myself and getting data from block 51208 (I knew that block was gossiped already).

### Future improvements

All these requests support passing block as one of:
- block number
- block hash
- tag (i.e. latest / finalized / safe / earliest / pending)

Currently we only support block hash, but in the future we want to support block number as well (e.g. when we have block index network).
I think in the future we can support tags as well, but I would say that's the lowest of priorities at the moment.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5900,7 +5900,9 @@ name = "rpc"
 version = "0.1.0"
 dependencies = [
  "alloy-primitives",
+ "alloy-rlp",
  "discv5",
+ "eth_trie",
  "ethportal-api",
  "hyper",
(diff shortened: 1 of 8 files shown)


KolbyML (approved):
:shipit:

PR #1400, opened 2024-08-25
refactor: decode ContentValue only with known ContentKey

morph-dev:
### What was wrong?

When decoding ContentValue, we try decoding all possible types until one succeed. This is dangerous as wrong type could succeed first (for example, if two values have identical encoding). This isn't the case with content keys, as they have unique prefix.

It's worth noting that Content Value should be decodable given Content Key.

### How was it fixed?

Make ContentValue decodable (either from byte slice or hex string) only with the given content key.
This required removing `Deserialize` trait from all content values, which broke many JSON things, which needed to be refactored.
I defined `RawContentValue` which can be decoded from hex string nicely, and repurposed `RawContentKey` which was seldomly used.

There are probably still many places that could be refactored so we don't use `Vec<u8>` and use `RawContentKey`/`RawContentValue`, but this PR is already big enough.

### To-Do


morph-dev:
Closing in favor of a bit simpler: #1403

PR #1401, opened 2024-08-26
feat(beacon): store `HistoricalSummariesWithProof` into db instead of cache

### LINKED ISSUE #1322: beacon-network: Store `HistoricalSummariesWithProof` in database instead of cache
Currently, we store `HistoricalSummariesWithProof` in the beacon storage cache.  

This is not very convenient because this content changes once per epoch and if the node goes offline for less than an epoch, it will be required to always get the latest `HistoricalSummariesWithProof` from the network.

ogenev:
### What was wrong?
Fixes #1322 

### How was it fixed?
- create a new `historical_summaries` table in the database on startup. This table will hold only one record.
- add insert and lookup queries
- cleanup unused cache code

### To-Do

### DIFF
--- a/trin-beacon/src/storage.rs
+++ b/trin-beacon/src/storage.rs
@@ -6,9 +6,8 @@ use ethportal_api::{
             LIGHT_CLIENT_UPDATES_BY_RANGE_KEY_PREFIX,
         },
         content_value::beacon::{
-            ForkVersionedHistoricalSummariesWithProof, ForkVersionedLightClientFinalityUpdate,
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
:shipit: looks good nothing stands out

I am just curious do you know the current rough size of the historical_summaries?, If not I can look it up in a bit myself, just thought I want ask because it is interesting and since you are working on this you might know it on the top of your head.

carver (approved):
No notes! 🚢

ogenev:
> I am just curious do you know the current rough size of the historical_summaries?

28,400 bytes

PR #1402, opened 2024-08-26
refactor(bridge): update bridge to use gossip engine

njgheorghita:
### What was wrong?
Our bridge used exclusively the gossip endpoint to broadcast data into the network. For improved performance, we need the bridge to store a view of the entire network locally and directly offer content to interested nodes.

### How was it fixed?
- Added a gossip engine to facilitate the broadcasting of content into the network
  - something to consider... there no longer is any way for the "bridges" to know whether or not a piece of content was successfully broadcast into the network. The "bridges" simply send the "gossip request" over to the gossip engine, who handles the broadcasting of the data / collecting stats. It doesn't seem like this is a problem, but I could be missing some side affects of this decision (aka we need to know that a piece of content was successfully gossiped before gossiping a second piece of content). And, it wouldn't be too hard to implement a "gossip result", informing the bridge of the result, if necessary.
- The gossip engine supports two modes..
  - `offer`: maintains a view of all network peers & directly offers content to interested peers
  - `gossip`: uses "gossip" endpoint to broadcast content into network (aka the way it worked previously)
- Moved all stats reporting into the gossip engine
  - the one downside of this is that there's no way to report stats for a given "block" / "slot"... but only for an individual piece of content. IMO, this is a fair trade-off to remove the complexity of passing the "block stats" across threads.
- Moved all semaphore logic (to track concurrent # of transfers) into gossip engine

### To-Do (in follow up prs, since this one is getting large)
- [ ] Update Metrics
  - [ ] there's still a bit more work to be done here w/ regards to metrics, since the old style of metrics (eg tracking `construct_and_gossip_block_body`) is no longer accurate, as the construct & gossip functions are now decoupled. I plan to update the metrics to reflect this decoupling (eg instead of tracking the time for `construct_and_gossip_block_body`, we'll track `construct_block_body` & `gossip_block_body` independently). But I think this is better addressed in a follow-up pr to avoid bloat here, though it should be noted that bridge metrics will not be 100% accurate until that pr lands, so I'm happy to include it here if anyone thinks that's critical to avoid.
  - [ ] `metrics.report_gossip_success` / `metrics.report_gossip_failure` is no longer accurate
  - [ ] add beacon metric for tracking last finalized slot
  - [ ] add state metric for tracking last block height
  - [ ] make sure all metric labels are accurately describing what's being measured
- [ ] reorganize all gossip logic in new gossip module, to avoid destroying the diff in this pr


njgheorghita:
Update... I'm currently in favor of #1422 as a simpler solution to this pr which uses introduces the `GossipEngine` concept. The `GossipEngine` concept worked pretty well in `latest` mode, but not so well when doing backfill/era files. Initially, the `GossipEngine` was nice bc it helped clean up some of the code in the bridges. The idea being that we could extract all of the gossiping / rate limiting / metrics / stats from the bridges and move it to the `GossipEngine`. However, the responsibility for rate limiting has to reside inside the "bridge", otherwise the `GossipEngine` will be quickly overwhelmed as gossip requests pile up. I still the `GossipEngine` concept might be valuable in the future, however imo it will require a pull model for gossip requests, rather than push (which is what we currently have). eg. The `GossipEngine` will be responsible for rate limiting, and will request the next available piece of content to gossip from the bridge when the engine is ready. That would require a fairly significant refactor which is out of scope, and I'm not necessarily convinced is ultimately worthwhile.

njgheorghita:
Closing b/c ^

PR #1407, opened 2024-08-29
Switch Cargo.toml in packages to use workspace dependencies

### LINKED ISSUE #845: refactor: use workspace to eliminate duplicate dependency definitions
See the rust docs on the workspace format: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table

Also see how reth does this specifically:
- example crate that reuses the workspace-defined dependencies: https://github.com/paradigmxyz/reth/blob/main/crates/rpc/ipc/Cargo.toml
- workspace dependencies: https://github.com/paradigmxyz/reth/blob/main/Cargo.toml

~The thing I'm least clear on: Can we still have a root Cargo file that builds and runs trin, or do we have to move it (like reth did) into a `/bin` folder? I'd rather avoid the churn.~ _This appears to be answered in the linked Rust docs now: the main Cargo.toml in the example is in the project root._

carver:
### What was wrong?

Fix #845 

### How was it fixed?

My process:
- Copied the version definitions into the workspace Cargo.toml and referred to them from each package
- If multiple versions are used in different packages, specify the highest version to use in the workspace
- If different features are used in different packages, specify the minimal subset in the workspace, and add features as needed in each package

There are several dependencies that are only used in a single package. I can see arguments for moving them to the workspace for tracking too. For example, if we don't move the dependencies, then it will be easy to accidentally create duplicates again in the future. But I landed on leaving them in place, because:
1. It highlights that removing a singular dependency from a package will completely remove it from the whole workspace
2. It feels a tiny bit burdensome to add every future dependency to the workspace when it's only in use in one place. So I wasn't ready to pick that as the new norm

But I'm open to persuasive arguments :)

### To-Do

- [x] root
- [x] e2store
- [x] ethportal-api
- [x] ethportal-peertest
- [x] light-client
- [x] portal-bridge
- [x] portalnet
- [x] rpc
- [x] trin-beacon
- [x] trin-execution
- [x] trin-history
- [x] trin-metrics
- [x] trin-state
- [x] trin-storage
- [x] trin-utils
- [x] trin-validation
- [x] utp-testing


KolbyML (approved):
:shipit: looks good

morph-dev (approved):
Looks good! :ship: 

Few sidenotes.

Since `Cargo.lock` wasn't updated, does that mean that there was no dependency that was used with different version?

Somewhat related, should we create issue(s) for:
- figuring out if we have some unused dependencies that can be removed
    - e.g. using some tools like [cargo-udeps](https://github.com/est31/cargo-udeps)
    - if we find convenient tool, maybe create a github action that runs it weekly (similar to update dependencies)
- finding out if we have multiple crates that do the same thing and we can simplify to using one (e.g. I noticed `ssz-rs` in `light-client` and I'm wondering if we can replace it with `ethereum_ssz`)
    - probably more of a manual task, but it might not be too hard not that most dependencies are in one place

carver:
> Since `Cargo.lock` wasn't updated, does that mean that there was no dependency that was used with different version?

Yeah, I think because of the regular updates, like [this one](https://github.com/ethereum/trin/pull/1398), most of the locked dependencies were further ahead of the Cargo.toml spec anyway.

> Somewhat related, should we create issue(s) for

Sure, I'll take a stab at writing some up 👍🏻

PR #1410, opened 2024-09-02
refactor: remove duplicate Withdrawal type

KolbyML:
### What was wrong?
we defined the same type for Withdrawals twice
### How was it fixed?

removed the redundant code


KolbyML:
I am closing these because some of our tests uses serde_yaml which depend on `validator_index`, well execution layer structures apparently depend on camal case? Anyways I don't think this issue is important enough for me to look to deep into it, so I am closing this PR keeping this sad reality in mind

PR #1411, opened 2024-09-02
fix: add `ssz` feature back for alloy-primitives to fix test compiling

KolbyML:
### What was wrong?
the tests I am trying to run locally are all failing, because we removed the `ssz` feature from alloy-primitives in the PR updating Cargo to properly use workplace 

so I got thousands of error's stating can't derive ssz for B256, ... 5 billion other errors etc. I was running tests in `block_body.rs`

depends
### How was it fixed?
by adding the feature back which was removed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["cryptography::cryptocurrencies"]
 description = "A Rust implementation of the Ethereum Portal Network"
 
 [dependencies]
-alloy-primitives.workspace = true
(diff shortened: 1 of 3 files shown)


morph-dev (approved):
I would even consider enabling this feature in the `[workspace.dependencies]` in root Cargo.toml.

morph-dev (approved):
Now you can remove it from `trin-validation/Cargo.toml`

PR #1412, opened 2024-09-02
fix: era `get_block_slot_indexes()` not working

KolbyML:
### What was wrong?
- `get_block_slot_indexes()` apparently it is the 0ed offset, from the start of the slot_index not 0
### How was it fixed?

implemented `get_block_slot_indexes()` to work based 0ed offset from the start of slot_index

### DIFF
--- a/e2store/src/era.rs
+++ b/e2store/src/era.rs
@@ -40,6 +40,7 @@ impl Era {
     }
 
     pub fn deserialize(buf: &[u8]) -> anyhow::Result<Self> {
+        let file_length = buf.len();
         let file = E2StoreMemory::deserialize(buf)?;
         let version = VersionEntry::try_from(&file.entries[0])?;
         let entries_length = file.entries.len();
@@ -47,12 +48,14 @@ impl Era {
 
         let slot_index_block = SlotIndexBlockEntry::try_from(&file.entries[entries_length - 2])?;
         let slot_index_state = SlotIndexStateEntry::try_from(&file.entries[entries_length - 1])?;
-        let slot_indexes = Era::get_block_slot_indexes(&slot_index_block);
+        let slot_indexes = Era::get_block_slot_indexes(file_length, &slot_index_block);
 
         // an era file has 4 entries which are not blocks
         ensure!(
             slot_indexes.len() == entries_length - 4,
-            "invalid slot index block: incorrect count"
+            "invalid slot index block: incorrect count {} {}",
+            slot_indexes.len(),
+            entries_length - 4
         );
         for (index, slot) in slot_indexes.into_iter().enumerate() {
             let entry = &file.entries[index + 1];
@@ -87,10 +90,11 @@ impl Era {
     pub fn iter_blocks(
         raw_era: Vec<u8>,
     ) -> anyhow::Result<impl Iterator<Item = anyhow::Result<CompressedSignedBeaconBlock>>> {
+        let file_length = raw_era.len();
         let file = E2StoreMemory::deserialize(&raw_era)?;
         let entries_length = file.entries.len();
         let block_index = SlotIndexBlockEntry::try_from(&file.entries[entries_length - 2])?;
-        let slot_indexes = Era::get_block_slot_indexes(&block_index);
+        let slot_indexes = Era::get_block_slot_indexes(file_length, &block_index);
 
         ensure!(
             slot_indexes.len() == entries_length - 4,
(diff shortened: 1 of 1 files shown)


e2store/src/era.rs
```diff
@@ -109,13 +109,19 @@ impl Era {
     }
 
     fn get_block_slot_indexes(slot_index_block_entry: &SlotIndexBlockEntry) -> Vec<u64> {
+        if slot_index_block_entry.slot_index.indices.is_empty() {
+            return vec![];
+        }
+        // minus 8 because the first 8 bytes are the starting slot
```
morph-dev, line 115: nit: it's minus 8 bytes for the first "entry", not "slot", right?
That should be `VersionEntry`, if I'm not wrong. Should we add something like `SERIALIZED_SIZE` to it (like we have for `Header`)?

e2store/src/era.rs
```diff
@@ -109,13 +109,19 @@ impl Era {
...
+        let beginning_of_index_record = slot_index_block_entry.slot_index.indices[0] - 8;
```
morph-dev, line 116: this logic doesn't seem right for me. Maybe I'm missing something, so if you could clarify:
1. with current logic, first slot will always be occupied (because check at line 124 will always be false). I don't think that's correct.
2. are we sure that `indices[0]` can't be lower than 8 (in which case we would panic)?

e2store/src/era.rs
KolbyML: image
`starting_slot` is 8 bytes, 

This has nothing to do with an entry

e2store/src/era.rs
KolbyML: > with current logic, first slot will always be occupied (because check at line 124 will always be false). I don't think that's correct.

The first index isn't an offset of 0, it is an offset of 8 because `starting_slot`

> are we sure that indices[0] can't be lower than 8 (in which case we would panic)?

Yes because it will always be 8 bytes higher than `starting_slot`

e2store/src/era.rs
KolbyML: image
Reading it again I am so confused, but I know it is 8 bytes

e2store/src/era.rs
morph-dev: I looked more into this, and after debugging bytes and looking for multiple example, I think I figured it out.

---

First, the indices are negative integers and we should update `indices` field in both `SlotIndexBlock` and `SlotIndexState` to `[i64; X]`. It even says so in the [spec](https://github.com/status-im/nimbus-eth2/blob/613f4a9a50c9c4bd8568844eaffb3ac15d067e56/docs/e2store.md#era-files):

> A SlotIndex record may appear in a stand-alone file which by convention ends with .e2i - in this case, the offset is counted as if the index was appened to its corresponding data file - offsets are thus negative and counted from the end of the data file. In particular, if the index is simply appended to the data file, it does not change in contents.

---

Second, if index (offset) point to the beginning of file (meaning `index == position_of_slot_index_block`), then that block is skipped.
In most cases, first block is present, and will be 8 bytes from the start of file (VersionEntry takes first 8 bytes), resulting in your math working but with the wrong reasoning (it's not because of the `starting_slot` but because of the `VersionEntry`).
However, if first block is skipped (if I'm not mistaken, epoch 819 has starting slot [6701056](https://beaconscan.com/slot/6545408) which is empty), this will not work, because first block already points to the beginning of file and we don't need to subtract 8 bytes.

---

Now, the solution. For this function to work, we need to know at what byte the `SlotIndexBlock` start in the original file, and we have 2 ways of doing so:
1. sum encoded sizes of all entries up to this point (version + blocks + state)
2. know length of the file and subtract encoded sizes of SlotIndexBlock and SlotIndexState entries (which are fixed: `8192*8 + 3*8 = 65560` and `4*8=32`)

---

I think it's a bit hard to explain this in words, but I tried my best. If something is not clear, we can have a call and discuss it.

KolbyML: I choose the 2nd option

KolbyML:

e2store/src/era.rs
```diff
@@ -108,12 +112,17 @@ impl Era {
...
+        file_length: usize,
+        slot_index_block_entry: &SlotIndexBlockEntry,
+    ) -> Vec<u64> {
+        // To calculate the beginning of the index record, we need to subtract:
+        // - 8192*8 + 3*8 = 65560 bytes for the slot index block entry
+        // - 4*8=32 bytes for the slot index state entry
+        // from the total file length.
+        // Then subtract the result from u64::MAX to get the starting slot of the index record then
+        // add 1.
+        let beginning_of_index_record = u64::MAX - (file_length as u64 - (65560 + 32)) + 1;
```
morph-dev, line 125: The indices/offsets should really be `i64`. Combined with comment above, this would be:
```rust
let beginning_of_index_record = file_length - SlotIndexBlockEntry::SERIALIZED_SIZE - SlotIndexStateEntry::SERIALIZED_SIZE;
let beginning_of_file_offset = -(beginning_of_index_record as i64);
```

And later you would just check `if *offset != beginning_of_file_offset {`

KolbyML:

morph-dev (approved):

e2store/src/era.rs
```diff
@@ -271,7 +288,9 @@ impl TryInto<Entry> for SlotIndexBlockEntry {
     fn try_into(self) -> Result<Entry, Self::Error> {
         let mut buf: Vec<u64> = vec![];
```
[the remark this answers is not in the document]
KolbyML: I think this is better

PR #1413, opened 2024-09-02
feat(beacon): store `LightClientBootstrap` in separate db table

### LINKED ISSUE #1405: beacon-network: Check for a recent bootstrap in db if `trusted-block-root` argument is not provided
Launching beacon sync in Trin requires always passing a valid trusted block root via CLI on every startup. 

This is not very feasible because if the client has been synced recently, it should already have a valid bootstrap stored in the database. We can use this bootstrap to start syncing on the next run, without expecting the user to provide a trusted block root again, as long as the bootstrap is in the weak subjectivity period.

ogenev:
### What was wrong?
This PR stores `LightClientBootstrap` beacon content in a separate table in the database. This is required because:

1. To fix #1405, we need a way to get the latest stored block root from the latest bootstrap in the DB. This is solved by storing the block root as a primary key for the new bootstrap table. (content-id is not needed for the beacon network because the node radius does not apply to this overlay network).
2. We want to purge all bootstraps older than 4 months. We can do this by adding a slot column in the bootstrap table and periodically purging all content where slot number < 4 months old.

### How was it fixed?
- create a new `lc_bootstrap` db table
- update insert and lookup queries
- update total beacon content size query

### To-Do

### DIFF
--- a/trin-beacon/src/storage.rs
+++ b/trin-beacon/src/storage.rs
@@ -1,4 +1,5 @@
 use ethportal_api::{
+    consensus::fork::ForkName,
     types::{
         content_key::beacon::{
             HISTORICAL_SUMMARIES_WITH_PROOF_KEY_PREFIX, LIGHT_CLIENT_BOOTSTRAP_KEY_PREFIX,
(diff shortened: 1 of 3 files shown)


trin-beacon/src/storage.rs
```diff
@@ -222,13 +223,12 @@ impl ContentStore for BeaconStorage {
...
-            BeaconContentKey::LightClientBootstrap(_) => {
-                let key = key.content_id();
+            BeaconContentKey::LightClientBootstrap(content_key) => {
                 let is_key_available = self
-                    .lookup_content_key(key)
+                    .lookup_lc_bootstrap_block_root(&content_key.block_hash)
                     .map_err(|err| {
                         ContentStoreError::Database(format!(
-                            "Error looking up content key: {err:?}"
+                            "Error looking up lc bootstrap block root: {err:?}"
```
KolbyML, line 231:
```suggestion
                            "Error looking up light client bootstrap block root: {err:?}"
```

I don't think we should use short hands in error's

KolbyML (approved):
I don't like LC being used as a shorthand in the code as just using light_client would be more  readable. A big distaste I have with Monero's codebase is every variable uses short hand, well this isn't as bad, it makes it hard for people who have no clue what is going on to get into our codebase.

I don't think comments justify shorthand to be clear as well. I believe in readable code!

I don't think this is a blocker (other then the comment of shorthand being used in the error) and obviously it is my opinion. Nothing looks wrong with the PR so feel free to merge it when want.

PR #1414, opened 2024-09-03
fix: alloy-primitives workspace dependency should include ssz feature

morph-dev:
### What was wrong?

The `ssz` feature of `alloy-primitives` crate should be enabled for entire workspace (not just for the `trin` crate)

### How was it fixed?

Updated workspace dependency.

### To-Do

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["cryptography::cryptocurrencies"]
 description = "A Rust implementation of the Ethereum Portal Network"
 
 [dependencies]
-alloy-primitives = { workspace = true, features = ["ssz"] }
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
:ship:

KolbyML (changes requested):
I am not sure this PR does anything. Ogy said `portal-bridge` won't compile for him. If that is true, this PR isn't even related as it never used the ssz derive feature for `alloy-primitives` in the first place

KolbyML:
If ogy says it fixes his issue we can merge it

KolbyML (approved):
nvm I see the issue now

PR #1415, opened 2024-09-03
feat: implement eth_call and refactor trin-execution to support async db

morph-dev:
### What was wrong?

The `eth_call` wasn't supported.

### How was it fixed?

Refactored logic in trin-execution crate to make functionality more reusable (and added support for async database).
Also refactored `rpc/src/eth_rpc.rs` by extracting state related functionality into separate file.

### To-Do

### DIFF
--- a/trin-execution/src/evm/blocking.rs
+++ b/trin-execution/src/evm/blocking.rs
@@ -1,0 +1,38 @@
+use revm::{
+    db::WrapDatabaseRef, inspector_handle_register, inspectors::NoOpInspector, DatabaseRef, Evm,
+    GetInspector,
+};
+use revm_primitives::{EVMResult, Env};
+
+use crate::{spec_id::get_spec_id, transaction::TxEnvModifier};
+
+/// Executes the transaction in the blocking manner.
+pub fn execute_transaction<DB: DatabaseRef>(
+    tx: &impl TxEnvModifier,
+    evm_environment: Env,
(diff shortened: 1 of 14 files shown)


trin-execution/src/evm/blocking.rs
```diff
@@ -0,0 +1,38 @@
...
+    Evm::builder()
+        .with_ref_db(database)
+        .with_env(Box::new(evm_environment))
+        .with_spec_id(get_spec_id(block_number))
+        .modify_tx_env(|tx_env| tx.modify(block_number, tx_env))
+        .with_external_context(external_context)
+        .append_handler_register(inspector_handle_register)
+        .build()
+        .transact()
+}
```
KolbyML, line 38: Realistically we wouldn't call Evm::builder() on every transaction execution. This is changed in https://github.com/ethereum/trin/pull/1337

I think this PR #1415 should be based off https://github.com/ethereum/trin/pull/1337 and #1337 should be merged first*

KolbyML (commented):
You asked for some early feedback, some it it looks fine. My main concern is this PR cements and creates abstractions for bad practices, this PR is based off latest which doesn't contain the improvements from #1337 yet.

This PR builds off of master which still calls EvmBuilder for every transaction which isn't idle. So my main feedback is to build off #1337

morph-dev:
Closing in favor of #1439

PR #1420, opened 2024-09-04
chore: add a parallel ci test to ensure all workspace crates compile

KolbyML:
### What was wrong?
When we updated cargo dependencies
https://github.com/ethereum/trin/pull/1411
https://github.com/ethereum/trin/pull/1414

failures to compile individual crates and this wasn't caught by CI
### How was it fixed?

creating a parallel ci test which checks all workspace crates

https://doc.rust-lang.org/cargo/commands/cargo-check.html here is a link to check, it basically just ensures `cargo build` works, without doing all the work so it is good for CI applications

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,32 @@ jobs:
           name: Test Trin workspace
           command: cargo test --workspace -- --nocapture
       - save-sccache-cache
+  # 'cargo check' performs all the compilation without actually building the crate, so it is quicker for the same guarantee
+  check-workspace-crates:
+    executor:
+      name: rust/default
+      tag: 1.79.0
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -224,6 +224,33 @@ jobs:
           name: Test Trin workspace
           command: cargo test --workspace -- --nocapture
       - save-sccache-cache
+  # This job is used to check if all crates build
+  # Check does all the build stuff without actually building the crate, so it is quicker for the same guarantee
```
njgheorghita, line 228: *nitpick* - `# 'cargo check' performs all the compilation without ....`

njgheorghita (approved):
:ship:

PR #1422, opened 2024-09-04
Add direct offer w/ running census process to state bridge

njgheorghita:
### What was wrong?

Convert state bridge from "gossip" mode to "offer" mode, which maintains a holistic view of network (via `Census`) and directly offers content to interested peers, rather than using `gossip` endpoint.

### How was it fixed?
I started with an alternative approach to this feature in #1402. I left a [comment](https://github.com/ethereum/trin/pull/1402#issuecomment-2331986573) explaining what I tried, and ultimately why I think this approach is better. tl;dr: This approach is simpler, but doubles down on the existing bridge architecture. That approach is cleaner, but won't work well unless we undertake a significant refactor in the entire bridge architecture (which isn't worth it imo). 

After some experimenting, it seems like the best way to move forward is to just implement offer mode for the state bridge (the only bridge where it's actually needed) and then later we can convert the other bridges to offer mode if deemed worthwhile.

### To-Do

- [ ] update bridge stats to make sure they're accurate for all gossip modes (future pr)
- [ ] update cluster deployment scripts to be compatible with these changes


portal-bridge/src/gossip.rs
```diff
@@ -95,15 +100,52 @@ pub async fn gossip_history_content(
...
+        let (resp_tx, resp_rx) = futures::channel::oneshot::channel();
+        let enrs_request = EnrsRequest {
+            content_key: ContentKey::History(content_key.clone()),
+            resp_tx,
+        };
+        census_tx.send(enrs_request)?;
+        let enrs = resp_rx.await?;
+        let total_enrs = enrs.len();
+        let mut success = 0;
+        for enr in enrs {
```
njgheorghita, line 118: One thing to consider... When running the bridge in "latest" mode, performance (measured by how many successful transfers there were ... indicating that this mode is actually better / faster at distributing content than the 'gossip' mode) really only was acceptable when I filtered out all Shisui nodes from `enrs`. It seems like we still have some utp tx issues with Shisui. So, this loop would be blocked whenever it encountered a problematic transfer. And by the time the following requests were actually made, the network had been saturated with content from the devops (aka "gossip") bridges.

One solution is to spawn tasks for each `trace_offer`, but that makes tracking the number of concurrent rpc requests (for the gossip limit) problematic. 

Imo, we **should** probably always have bridges running in both modes (eg. 'gossip' & 'offer'). And given that this is really only an issue in "latest" mode, I don't think it's worth hardcoding the Shisui exception.

portal-bridge/src/census.rs
```diff
@@ -0,0 +1,292 @@
...
+                let data_radius = Distance::from(U256::from(*data_radius));
+                if data_radius >= distance {
+                    Some(enr.clone())
+                } else {
+                    None
+                }
+            })
+            .collect()
+    }
+}
```
KolbyML, line 227: One of the reasons we want to build the big view model of the bridge is because then we can more efficiently bridge data to the network.

I think we should limit how many Enr's are returned. Because well this model gives us improvements, if there is 1_000_000 enr's we are gossiping data to it will be super slow.

portal-bridge/src/gossip.rs
KolbyML: > One solution is to spawn tasks for each trace_offer, but that makes tracking the number of concurrent rpc requests (via the gossip_semaphore) problematic.

I am not sure I see where this sentiment is coming from.

> tracking the number of concurrent rpc requests (via the gossip_semaphore) problematic.

I don't think this would be problem-matic, I don't think a gossip semaphore limit should be used on serving latest data.

I think we should limit how many Enr's we gossip to though as currently their is no limit which is a bit excess work

KolbyML: > Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer'). And given that this is really only an issue in "latest" mode, I don't think it's worth hardcoding the Shisui exception

If this is the case I think `offer` mode isn't implemented correctly.

> So, this loop would be blocked whenever it encountered a problematic transfer

There shouldn't be blocking code at that level at least.

> Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer')

A good `offer` implementation I think it only makes sense to bridge in `offer` mode.

All offer mode is, is an optimized version of `gossip` where the bridge has better information for knowing who to gossip to. With better data the bridge can send offer requests in a less wasteful way.

I pointed it out earlier, but I don't think it makes sense to offer every node we are aware of which will accept it. The bridge's job shouldn't be seeding every node on the network, won't the nodes just repeat what they got to their interested neighborhood

KolbyML (commented):
I think the census code is mostly good

> Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer'). And given that this is really only an issue in "latest" mode, I don't think it's worth hardcoding the Shisui exception. Even if all of our bridges were in offer mode, this wouldn't be fatal, it would just slightly delay the time it takes for latest data to be available in the network.

I think there is currently an issue with how the `offer` code is designed. We shouldn't be synchronously sending offer requests. This should be async a bad node shouldn't block others from getting the information.

^ with that problem fixed, especially as our network scales I don't think `gossip mode` can compete

portal-bridge/src/bridge/history.rs
```diff
@@ -147,13 +157,15 @@ impl HistoryBridge {
...
                     height,
                     None,
                     self.portal_client.clone(),
                     self.execution_api.clone(),
-                    None,
+                    permit,
                     self.metrics.clone(),
+                    self.census_tx.clone(),
                 );
             }
```
KolbyML, line 170: Why was a semaphore added to latest mode? I get why we would do it for backfill, but if we are only offering to 8 Enr's every 12 seconds we shouldn't have a issue.

If anything I would make it if we reach the gossip limit the code panics, as if we can't keep up with latest something is wrong in a big way and we need to improve the code to be able to keep up

portal-bridge/src/gossip.rs
KolbyML: `Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer')`

I disagree with this. If we have a good `offer mode` implementation a naive `gossip` implementation can't compete on performance or reliability especially as the network grows where bigger and bigger issues will be displayed by `gossip mode`

`gossip mode` will use less ram so it could be helpful if someone wants to contribute with limited hardware, but other then that I think the points against `offer mode` can easily be fixed and then it will just outperform `gossip` mode

kdeme: > I disagree with this. If we have a good `offer mode` implementation a naive `gossip` implementation can't compete on performance or reliability especially as the network grows where bigger and bigger issues will be displayed by `gossip mode`

The issue I have with this is that this then tends to go in the direction of relying only on a "few" mega bridges that can/will do all the work. If these then get attacked, the network stalls in its data injections.

I'm also not so sure if it is the good way forward if we eventually want to end up with Ethereum nodes that do this work and where the general network takes the load of gossiping things around.

KolbyML: > > I disagree with this. If we have a good `offer mode` implementation a naive `gossip` implementation can't compete on performance or reliability especially as the network grows where bigger and bigger issues will be displayed by `gossip mode`
> 
> 
> 
> The issue I have with this is that this then tends to go in the direction of relying only on a "few" mega bridges that can/will do all the work. If these then get attacked, the network stalls in its data injections.
> 
> 

You would easily be able to limit the routing table size if that is issue, for lower end machines which wanted to run a bridge. But still allowing the bridge to scale for hardware which can do more is very important.

So I don't think this design decision limits who can bridge, it just allows people with better hardware to utilize it to bridge more efficiently which below I will go over why that is important 

> 
> I'm also not so sure if it is the good way forward if we eventually want to end up with Ethereum nodes that do this work and where the general network takes the load of gossiping things around.
> 
> 

These outlines are meant to meet the needs of the State Network, which has drastically higher requirements than gossiping for the History or Beacon Network. By drastically I mean order of magnitudes.

In the state network full proofs are required to be gossiped for every changed trie node every block. As outlined during the Prague Summit.

So the general network isn't really able to take on massive burden, other than making sure its friends have the data it was gossiped. The bridge takes on a majority of the work.

On top of all this we need to account for executing the block and generating the MPT which is costly then gossip both the all the account trie and storage trie intermediate nodes in under 12 seconds ideally less. Which is no easy feat. It is unrealistic 1 bridge will be able to do this so we will likely have to split how much data is gossiped over many bridges which is expensive.


Then the other issue is for backfilling non-latest data which with the current bridge design I don't want to imagine how long that would take.

Then the third issue is how do we reliably fill holes in the network this is very hard to do, because finding the hole in the first place is a probabilistic anomaly.


So running the state network and maintaining it is very expensive and time consuming, having an efficient bridge is hence important even then it will be super expensive and time consuming. I don't think this bridge design limits lower end nodes from contributing to gossip data, it doesn't limit us in anyway like that.

It realistically enables us to hopefully sustainably run the State Network, which I don't think its requirements can be understated.

For other networks the requirements I outlined are most definitely overkill, but I am not that concerned with other networks, because if we can sustainably make the state network work, those networks are in relation a drop in the bucket.

kdeme: To be clear, my remark was about only relying on the "optimized" bridge architecture. Not about actually having those around for now.

KolbyML: > To be clear, my remark was about *only* relying on the "optimized" bridge architecture. Not about actually having those around for now.

Maybe I should have said for State it only makes sense to rely on X to be more clear, sorry for the confusion

njgheorghita:
#### `Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer')`

One reason to keep "gossip" bridges around..
- "offer" bridges each have a "complete" view of the network. "offer" bridges calculate & sort interested peers in the same way. if we limit the # of peers each "offer" bridge gossips to, then each "offer" bridge will be sending requests to the exact same set of peers. the location of an "offer" bridge in the network doesn't matter. there's no difference between running a single "offer" bridge vs many "offer" bridges, they will just offer the exact same content to the same peers
  - possible fix: we could randomize the interested peers for an "offer" bridge rather than sorting by distance to content id?
  
The location of a gossip bridge impacts how its data is broadcast into the network. Which makes running multiple gossip bridges meaningful, providing a somewhat more "natural" dissemination of data across the network.

Personally, I don't see this as a "competition" between gossip/offer bridges. It's not difficult to maintain both modes, even if it's simply to preserve a small advantage for low-resource users. 

In the short-term (aka until "offer" bridges are stabilized) it's obvious that we want to continue to support gossip bridges (to maintain current network performance), and slowly roll out "offer" bridges. So, I don't see this decision as a blocker for this pr, and probably something that's better discussed during sync when we've reached a point that deprecating "gossip" bridges is feasible.

njgheorghita:
#### `I think we should limit how many Enr's we offer to, to something like 8.`

I also considered this, and am not opposed to some solution here. 
- As I mentioned in my previous comment, since every "offer" bridge has the same view of the network, if we just take the 8 closest peers to the content id, then each bridge will perform the exact same gossip. 
- Why 8? I'm not a fan of picking some arbitrary number (as a long term solution). Maybe we could cap the limit to a % of total nodes in the `Census`?
- We could randomly sort interested enrs (rather than sorting them by distance to content id), which is the only solution I can think of that would justify running multiple "offer" bridges

portal-bridge/src/bridge/history.rs
njgheorghita: I added a semaphore primarily for consistency across the bridges. I didn't see any strong reason to **not** include a semaphore here.
- It's fairly lightweight code, and if it's never actually used to rate limit, then that's fine
- Sooner or later it would be nice to refactor this bridge entirely, so improving the consistency across bridge implementations is a step towards this direction
- Pegging the estimate here against 8 enrs / second is not ideal (as mentioned in the other comment, I'm not a huge fan of just picking 8 as the permanent limit)
  - Even if we choose 8, it would be 24 offers / 12 seconds, taking into account the various content types, but that doesn't really affect the point you're making
- It's possible that we gossip more than a block / 5 second loop. eg, maybe there's a bad http response from the provider causing us to miss a few and then process multiple blocks in a later loop 
  - If we update to a webhook provider api to get latest blocks, this would be avoided, but pandaops nodes don't support webhooks right now

njgheorghita:
#### I am not sure how tracking the number of concurrent rpc requests (via the gossip_semaphore) problematic would be problematic, I think we should do that in offer mode.

Well, I guess it's only a problem if we actually want to use the semaphore to precisely rate limit each individually spawned "offer" thread..

I tried this, it's not good... you're trying to use the same semaphore from within a process that's already holding a permit and it leads to blocking. 

The reason I went with sync offers, is then that semaphore permit is accurately limiting the # of rpc requests, since each offer is effectively a single rpc response (since they're all happening sequentially). And when they're all done, the permit is released. 

There is a problem here with a single bad rpc request blocking the others... as you pointed out and what I observed with shisui

Ok.... thinking about this more. In my head, I understood `gossip_limit` to measure the limit of active rpc requests going on at any one time. This was wrong. It's actually defined as the `The maximum number of active blocks being gossiped.`

So, as long as a single block being gossiped spawns a fixed/capped number of sub-threads, then the limit is still useful.
If a single block spawns a variable number of sub-threads, then the limit is no longer useful.

Anyways, cool, given that I was probably being a little too specific with the semaphore and my mis-reading of the `gossip-limit`, I agree that we should change this to async. But, I'd like to figure out the question of which/how many enrs to take for each offer, since that might impact whether we're spawning a fixed/variable number of threads here.

portal-bridge/src/bridge/history.rs
KolbyML: > * Sooner or later it would be nice to refactor this bridge entirely, so improving the consistency across bridge implementations is a step towards this direction

In that case I think the state bridge should be it's own implementation. Our limiting factor is the State network and I think all design decisions should be made with State in mind first as it is the hardest problem, and if we can get state to work, then the other networks will work too.

> * Pegging the estimate here against 8 enrs / second is not ideal (as mentioned in the other comment, I'm not a huge fan of just picking 8 as the permanent limit)

If we can only do X offer's a second but we have Y State content to gossip in Z seconds. If the data is successfully gossiped to a node it can handle neighborhood gossip from their, our priority is being able to gossip a latest diff in ideally less then 8 seconds. (we should aim for way less then 8 seconds)

If we gossip to even 1 enr that is enough as long as the bridge is able to gossip the whole diff or the whole diff split across multiple bridges in under the time requirement

>   * Even if we choose 8, it would be 24 offers / 12 seconds, taking into account the various content types, but that doesn't really affect the point you're making

State is at the end of the day the hardest problem, it is also the higher priority out of all the network's. If we can handle state we can more then handle the rest

> * It's possible that we gossip more than a block / 5 second loop. eg, maybe there's a bad http response from the provider causing us to miss a few and then process multiple blocks in a later loop

We should stop relying on the JSON-RPC as it is unreliable. ^ The reasons why we should do this is outlined in the quote above
  
>   * If we update to a webhook provider api to get latest blocks, this would be avoided, but pandaops nodes don't support webhooks right now

Even then it is unreliable, to a certain degree it makes sense for the State bridge to also handle gossiping History, and Beacon as it will get the data faster, as it will be directly executing and receiving the block from the CL client itself.

KolbyML:
> #### `Imo, we should probably always have bridges running in both modes (eg. 'gossip' & 'offer')`
> One reason to keep "gossip" bridges around..
> 
> * "offer" bridges each have a "complete" view of the network. "offer" bridges calculate & sort interested peers in the same way. if we limit the # of peers each "offer" bridge gossips to, then each "offer" bridge will be sending requests to the exact same set of peers. the location of an "offer" bridge in the network doesn't matter. there's no difference between running a single "offer" bridge vs many "offer" bridges, they will just offer the exact same content to the same peers
>   
>   * possible fix: we could randomize the interested peers for an "offer" bridge rather than sorting by distance to content id?

Not exactly because all the `offer bridges` we will be running will be gossiping different state data. Because we are very constrained by time and offer per second capacity the idea will be.

- 1. We execute a block and get the diff
- 2. We sort all the state content that has to be gossiped by xor(bridge.node_id, content_id)
- 3. We gossip Total_State_data_to_be_gossiped / total_offer_nodes_we_are_running  amount of content per a `offer bridge`

So the hypothesis above, doesn't hold true for the State bridge, (it does for the history and beacon network but those aren't a concern or priority in comparison as their requirement is so small, our current bridge infrastructure would be good enough)


> The location of a gossip bridge impacts how its data is broadcast into the network. Which makes running multiple gossip bridges meaningful, providing a somewhat more "natural" dissemination of data across the network.

As stated in the example above we will also utilize the bridges node_id, so there will still be a `"natural" dissemination of data across the network`

> Personally, I don't see this as a "competition" between gossip/offer bridges. It's not difficult to maintain both modes, even if it's simply to preserve a small advantage for low-resource users.

We can maintain `gossip` bridge for history and beacon, but it won't be usable for the State Network

> In the short-term (aka until "offer" bridges are stabilized) it's obvious that we want to continue to support gossip bridges (to maintain current network performance), and slowly roll out "offer" bridges. So, I don't see this decision as a blocker for this pr, and probably something that's better discussed during sync when we've reached a point that deprecating "gossip" bridges is feasible.

We can keep that gossip bridges for the Beacon and History network, well they will be more efficient when the state bridge is done, we don't necessarily need a better bridge for those network's the gossip bridge is good enough

Honestly if Beacon and History never use `offer mode` that is fine

KolbyML:
> #### `I think we should limit how many Enr's we offer to, to something like 8.`
> I also considered this, and am not opposed to some solution here.
> 
> * As I mentioned in my previous comment, since every "offer" bridge has the same view of the network, if we just take the 8 closest peers to the content id, then each bridge will perform the exact same gossip.

I outlined in a earlier reply why this wouldn't hold true for the state network

> * Why 8? I'm not a fan of picking some arbitrary number (as a long term solution). Maybe we could cap the limit to a % of total nodes in the `Census`?

``Maybe we could cap the limit to a % of total nodes in the `Census`?``

I don't think this is feasible for the State network, because of what I said in early comments. I think for spreading the gossiped data amongst nodes we should rely on `neighborhood gossip` to do some of the work. State Bridges already have enough work to do, We might be gossiping to less then 8 Enr's at that to meet the time requirements (testing will have to be done to get a more accurate number what is possible)

> * We could randomly sort interested enrs (rather than sorting them by distance to content id), which is the only solution I can think of that would justify running multiple "offer" bridges

I outlined in a earlier message why the State Network is different then the (beacon and history network) and that result wouldn't be a concern

KolbyML:
> #### `I am not sure how tracking the number of concurrent rpc requests (via the gossip_semaphore) problematic would be problematic, I think we should do that in offer mode.`
> Well, I guess it's only a problem if we actually want to use the semaphore to precisely rate limit each individually spawned "offer" thread..

I think that is potentially still open for discussion, what we need know is we will likely have a capacity of offers we can do within the time requirement and a certain amount of content we are required to gossip within the time frame. So it is fine if we don't get this right in this PR, this can be done later


> I tried this, it's not good... you're trying to use the same semaphore from within a process that's already holding a permit and it leads to blocking.

Is this referring to the latest bridge or backfill bridges? This sounds like a design problem potentially (of course there is limited context I can play with, reading this, with a more detailed write up on what exact scenario is happening)

> The reason I went with sync offers, is then that semaphore permit is accurately limiting the # of rpc requests, since each offer is effectively a single rpc response (since they're all happening sequentially). And when they're all done, the permit is released.

Maybe we can do this `The reason I went with sync offers, is then that semaphore permit is accurately limiting the # of rpc requests`, but we have do the offer's async

> There is a problem here with a single bad rpc request blocking the others... as you pointed out and what I observed with shisui

Maybe we should do a 

```nim
let handles = iter().map(tokio::spawn(offer)).collect().
handles.join().await;
```

This way we would still wait for them all to complete but none of them would block each other




> Ok.... thinking about this more. In my head, I understood `gossip_limit` to measure the limit of active rpc requests going on at any one time. This was wrong. It's actually defined as the `The maximum number of active blocks being gossiped.`

I think the semaphore limit should limit the amount of active offer requests at a time. 

> So, as long as a single block being gossiped spawns a fixed/capped number of sub-threads, then the limit is still useful. If a single block spawns a variable number of sub-threads, then the limit is no longer useful.

So a cap of the amount of offer's at a time if I understand correctly

> Anyways, cool, given that I was probably being a little too specific with the semaphore and my mis-reading of the `gossip-limit`, I agree that we should change this to async. But, I'd like to figure out the question of which/how many enrs to take for each offer, since that might impact whether we're spawning a fixed/variable number of threads here.

`But, I'd like to figure out the question of which/how many enrs to take for each offer, since that might impact whether we're spawning a fixed/variable number of threads here.`

To find this we will need to wait for Latest state and then run a | cost vs reliability | analysis where gossiping the required content amount per time frame is met

portal-bridge/src/bridge/state.rs
```diff
@@ -253,73 +219,174 @@ impl StateBridge {
...
+            content_key.to_hex(),
+        );
+        debug!(
+            "Successfully offered to {}/{} peers. Content key: {}. Success: {:?}, Fail: {:?}",
+            self.success.len(),
+            self.total,
+            content_key.to_hex(),
+            self.success,
+            self.fail,
+        );
```
KolbyML, line 390:
```suggestion
    fn report(&self, content_key: &StateContentKey) {
        if enabled!(Level::DEBUG) {
                 debug!(
                    "Successfully offered to {}/{} peers. Content key: {}. Success: {:?}, Fail: {:?}",
                    self.success.len(),
                    self.total,
                    content_key.to_hex(),
                    self.success,
                    self.fail,
                );     
        } else {
                  info!(
                      "Successfully offered to {}/{} peers. Content key: {}",
                      self.success.len(),
                      self.total,
                      content_key.to_hex(),
                  );
        }
```

Doing it this way you get the same result, but without the double logging of the same information

KolbyML (approved):
For history latest I think it is weird we are using a semaphore for it now, if we are doing that I would like an error to be printed if the limit is ever reached (or the semaphore is ever blocking), as if that happens glados will fail audits for latest history and it indicates something is wrong.

Overall for the rest of the PR it looks really good

:shipit:

portal-bridge/src/census.rs
```diff
@@ -0,0 +1,287 @@
...
+            tokio::select! {
+                // handle enrs request
+                //biased;
+                Some(request) = self.census_rx.recv() => {
+                    let enrs = self.get_interested_enrs(request.content_key).await;
+                    if let Err(err) = request.resp_tx.send(enrs) {
+                        error!("Error sending enrs response: {err:?}");
+                    }
+                }
+                // network is not initialized, this will never yield
```
morph-dev, line 72: I would remove/change this comment. If we enable history census in the future, it's likely that this comment will not be noticed (and it would be incorrect).

Same for the beacon.

portal-bridge/src/census.rs
```diff
@@ -0,0 +1,287 @@
...
+    async fn iterate_peer_routing_table(&mut self, enr: Enr) {
+        for distance in 245..257 {
+            let Ok(result) = self
+                .subnetwork
+                .find_nodes(&self.client, enr.clone(), vec![distance])
+                .await
+            else {
+                continue;
+            };
+            for found_enr in result {
```
morph-dev, line 176: nit: is it worth doing this in parallel?

portal-bridge/src/bridge/state.rs
```diff
@@ -31,52 +31,42 @@ use trin_execution::{
...
+    stats: Arc<Mutex<Stats>>,
 }
 
 impl StateBridge {
     pub async fn new(
         mode: BridgeMode,
         portal_client: HttpClient,
-        header_oracle: HeaderOracle,
-        epoch_acc_path: PathBuf,
         gossip_limit: usize,
```
morph-dev, line 57: nit: should we rename variable to `offer_limit`?

portal-bridge/src/census.rs
```diff
@@ -0,0 +1,287 @@
...
+
+/// The maximum number of enrs to return in a response,
+/// limiting the number of OFFER requests spawned by the bridge
+/// for each piece of content
+const ENRS_RESPONSE_LIMIT: usize = 8;
+
+/// The census is responsible for maintaining a list of known peers in the network,
+/// checking their liveness, updating their data radius, iterating through their
+/// rfn to find new peers, and providing interested enrs for a given content key.
+pub struct Census {
```
morph-dev, line 36: I have an idea, but feel free to ignore it if you don't like it.

Instead of having `history, state and beacon` network separately, we have one field: `networks` that is ether `HashMap<Subnetwork, Network>` of `Vec<Network>`.

The constructor would accept list of `Vec<Subnetwork>` and call `init()` for each of them (no need for separate `init()` function).

portal-bridge/src/bridge/state.rs
morph-dev: +1 for Kolby's idea.

But I would only log failed ones at the debug level.
If you think it's worth to log the enrs of successful ones as well, maybe do it at TRACE level.

portal-bridge/src/bridge/state.rs
```diff
@@ -253,73 +219,174 @@ impl StateBridge {
...
+            .clone()
+            .acquire_owned()
+            .await
+            .expect("to be able to acquire semaphore")
+    }
+}
+
+// Stats for tracking the performance of offered state content
+struct Stats {
+    keys: HashMap<StateContentKey, OfferReport>,
```
morph-dev, line 315: nit: maybe rename to `reports`?

portal-bridge/src/bridge/state.rs
```diff
@@ -253,73 +219,174 @@ impl StateBridge {
...
-                    content_key.clone(),
-                    content_value,
-                    metrics.clone()
-                )).await
+    ) {
+        let Ok(enrs) = self.request_enrs(&content_key).await else {
+            error!("Failed to request enrs for content key, skipping offer: {content_key:?}");
+            return;
+        };
+        self.stats
```
morph-dev, line 249: Do we need `stats` field?

How about we add content key to the `OfferReport`, create new one here (`Arc<Mutex<OfferReport>>`), and pass it to tasks below (the last one that updates it will print the report)?

Up to you.

morph-dev (approved):
Look good :ship: 

I left bunch of suggestions, mostly small ones. If you accept some of the bigger ones, let me know and I will take another quick look. Otherwise feel free to merge when you want.

portal-bridge/src/main.rs
```diff
@@ -40,25 +44,34 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            header_oracle,
-            epoch_acc_path,
+            bridge_config.mode.clone(),
+            portal_client.clone(),
             bridge_config.gossip_limit,
+            census_tx,
         )
         .await?;
 
+        // do we also need to push the state_bridge to the bridge_tasks?
```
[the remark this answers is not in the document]
KolbyML: > I would say yes. @KolbyML ?

The State bridge doesn't create a join handle, because we don't spawn it in a tokio::thread like the other bridges

portal-bridge/src/census.rs
njgheorghita: Probably, cause it's definitely a bit slow as it is. But it's also a bit tricky to pull it off in parallel as it's currently written since you'd need to pass a mutable reference across threads... Though, there's definitely room for improvement here. But since the state network is currently small and quick to initialize, I'm just going to leave a comment saying that it should be improved once initialization becomes too slow

portal-bridge/src/census.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+impl Census {
+    pub async fn init(&mut self) {
+        // currently, the census is only initialized for the state network
+        info!("Initializing state network census");
+        self.state.init().await;
+    }
+
+    pub async fn run(&mut self) {
```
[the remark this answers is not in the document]
njgheorghita: Not quite, iiuc. `select!` executes the next "ready" task. If there are multiple "ready" tasks, it will randomly select between which ones are available. So, if there's no expired enr, it will just execute the enr requests. If there's both an enr request & an expired enr, it'll randomly select between which one to execute. And etc if there are multiple networks with expired enrs. This was we can be sure that the census will continue to update while it's receiving a stream of enr requests.

portal-bridge/src/census.rs
njgheorghita: Mmmm, personally I prefer to have the explicit subnetwork names... makes for easier access to the `Network`, rather than 'getting' from the hashmap, or implementing a getter fn. Though, I admit it's not a really big difference...

portal-bridge/src/census.rs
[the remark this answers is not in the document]
morph-dev: I somehow missed `tokio::select!` . All is good.

PR #1423, opened 2024-09-04
chore: update depends so we can use sigp crates directly

KolbyML:
### What was wrong?
This is a part of https://github.com/ethereum/trin/commit/68f4bae39abfeec9b900563e87d6fdae9dd747ad

We were relying on forks of sigp library's during our transition to using Reth's primitive types, I made forks to update those sigp libraries to use Reth's primitives as well, they are now updated so here is a PR using them
### How was it fixed?

Updating library's as required and code as well.

Many libraries depend on each other in certain ways so I just updated everything that be needed so we can compile again, well also achieving the goal of relying on sigp's library's directly again


ethportal-api/src/types/execution/header.rs
```diff
@@ -262,20 +262,21 @@ impl From<Header> for RpcHeader {
             receipts_root,
             logs_bloom,
             difficulty,
-            number: Some(U256::from(number)),
-            gas_limit,
-            gas_used,
-            timestamp: U256::from(timestamp),
+            number,
+            gas_limit: gas_limit.to::<u128>(),
```
morph-dev, line 266: nit: I didn't check it, but I think you don't have to specify `::<u128>` as compiler should be able to figure it out. But it's fine if you want to be explicit. Here and in most other places

Cargo.toml
```diff
@@ -80,8 +80,8 @@ members = [
 
 [workspace.dependencies]
 alloy-consensus = { git = "https://github.com/alloy-rs/alloy.git", rev = "d68a6b787b2904061f0ae7fcc02ece8513e3c500"}
```
morph-dev, line 82: Can we update this as well?

morph-dev (commented):
Can we also update `revm-inspectors` in `trin-execution/Cargo.toml`?

KolbyML:
> Can we also update `revm-inspectors` in `trin-execution/Cargo.toml`?

I tried, but they have it configured weird so it wasn't easy to do

Cargo.toml
KolbyML: I will just remove it as we just use a constant from it so it is a little overkill

KolbyML:
> revm-inspectors

I think I figured it out

KolbyML:
@morph-dev ready for another review I believe I resolved all the concerns

Cargo.toml
morph-dev: nit: You could have used `alloy_consensus: v0.3.1`, as we already depend on it indirectly. Up to you.

morph-dev (approved):

trin-execution/Cargo.toml
```diff
@@ -12,6 +12,7 @@ authors = ["https://github.com/ethereum/trin/graphs/contributors"]
 
 
 [dependencies]
+alloy-consensus.workspace = true
```
morph-dev, line 15: nit: just don't forget to remove constants that you introduced

morph-dev (approved):

Issue #1419, opened 2024-09-04
Formalize all dependencies


carver:
There are several dependencies that specify repositories currently on @KolbyML 's fork. We should get these cleaned up by Devcon.

I haven't dug into all of them, but depending on the situation there are probably two main solutions:
1) merge changes upstream
2) initiate transfer of repository into the Ethereum org

KolbyML:
I think this is resolved now from https://github.com/ethereum/trin/pull/1423 and https://github.com/ethereum/trin/pull/1424 is anything still missing? If not we can close this issue

ogenev:
@KolbyML what is the current state of `tree_hash`  and `tree_hash_derive`, do we still need to use the forks?

KolbyML:
> @KolbyML what is the current state of `tree_hash`  and `tree_hash_derive`, do we still need to use the forks?

Not any more the PR I merged yesterday, uses the main crates.io releases again

ogenev:
Already resolved.

PR #1424, opened 2024-09-05
chore: update cargo.toml to depend on eth-trie.rs in ethereum org

KolbyML:
### What was wrong?
I think this is the last fix for https://github.com/ethereum/trin/issues/1419 ?
### How was it fixed?

I de-forked the latest version of eth-trie.rs we were using, then moved it to the Ethereum org

I then made a release, following the common pattern we used then made this PR

https://github.com/ethereum/eth-trie.rs here is the repo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2499,7 +2499,7 @@ dependencies = [
 [[package]]
 name = "eth_trie"
 version = "0.4.0"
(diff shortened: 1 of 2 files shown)


njgheorghita (approved):
:ship:

PR #1426, opened 2024-09-06
chore: experimental messaging from readme

KolbyML:
### What was wrong?

- the readme says unix only, but this is no longer true (we want it to work with windows and would gladly take issues now)
- I think the experimental messaging is out dated now and counter productive. We expect people Trin is ready to use, but then the readme tells me to not use it, it is counter productive
### How was it fixed?

Removing the 2 issues from the readme.

Programs change overtime for example sigp/discv5 and ethereum/go-ethereum change over time but don't have this kind of messaging in the readme.

If we verbally say we are ready for other's to take us seriously, I think messaging and this PR is the first step.

Same goes for window's support

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,22 +2,11 @@
 
 Trin is a Rust implementation of a [Portal Network](https://github.com/ethereum/portal-network-specs) client.
 
-The Portal Network is still in the research phase, and this client is experimental.
-
-**Do not rely on Trin in a production setting.**
-
-
 ## How to use Trin
 
 Check out the [Trin book](https://ethereum.github.io/trin) to quickly get up and running with Trin.
 
 > **NOTE**: This project uses [Git Submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules). If you just cloned the project, be sure to run: `git submodule update --init`. See our dedicated [page](https://ethereum.github.io/trin/developers/contributing/git/submodules.html) for more info.
-
-## Experimental Status
-
-Trin is a prototype Portal Network client. This implementation and the Portal Network specifications will continue to co-evolve.
-
-In this stage of development, Trin lacks comprehensive data validation.
 
 ## Want to help?
 


ogenev:
I don't feel that removing this note from the readme is the first step for fully supporting Windows. I think we need to first identify what are the current bottlenecks (if any) for fully supporting the platform.

For example, we can start with adding a build test for Windows in CircleCI (because cargo check doesn't catch all compilation errors).

KolbyML:
> For example, we can start with adding a build test for Windows in CircleCI (because cargo check doesn't catch all compilation errors).

Can you link a source maybe I have a misunderstanding how cargo check works?

KolbyML:
> 
> > For example, we can start with adding a build test for Windows in CircleCI (because cargo check doesn't catch all compilation errors).
> 
> Can you link a source maybe I have a misunderstanding how cargo check works?

To my understanding https://doc.rust-lang.org/cargo/commands/cargo-check.html cargo check does everything but code generation, which to my knowledge should check all compilations error's cargo build would. So it is like cargo build but faster and hence good for CI

ogenev:
> > For example, we can start with adding a build test for Windows in CircleCI (because cargo check doesn't catch all compilation errors).
> 
> Can you link a source maybe I have a misunderstanding how cargo check works?

https://rust-lang.github.io/rfcs/3477-cargo-check-lang-policy.html#decision

https://stackoverflow.com/a/69140084

KolbyML:
> > Can you link a source maybe I have a misunderstanding how cargo check works?
> 
> 
> 
> https://rust-lang.github.io/rfcs/3477-cargo-check-lang-policy.html#decision
> 
> 
> 
> https://stackoverflow.com/a/69140084

Ah ok in that case maybe we should just stop using cargo check all together. Thank you for informing me I wasn't aware of that

KolbyML:
I removed the UNIX only part in this PR already, so the experimental part is the only part left https://github.com/ethereum/trin/pull/1452

ogenev (approved):

PR #1427, opened 2024-09-06
feat(beacon): load trusted block root from database if not provided via CLI

### LINKED ISSUE #1405: beacon-network: Check for a recent bootstrap in db if `trusted-block-root` argument is not provided

ogenev:
### What was wrong?

Fixes #1405.

### How was it fixed?
- add logic to check for a recent trusted block root if not provided via CLI
- update `trusted_block_root` argument to `B256`.
- fix `lookup_lc_bootstrap_block_root` method

### To-Do

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -146,7 +146,7 @@ pub struct TrinConfig {
         value_parser = check_trusted_block_root,
         help = "Hex encoded block root from a trusted checkpoint"
     )]
-    pub trusted_block_root: Option<String>,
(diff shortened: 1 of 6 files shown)


KolbyML (approved):
:shipit:

Issue #1404, opened 2024-08-26
Move away from using Vec<u8> to represent raw content key/value


morph-dev:
Instead of using `Vec<u8>` to represent raw content keys/values, we should use something like [`alloy_primitives::Bytes`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Bytes.html) because:

- copy is faster
- (de)serializes into hex string correctly

We should either wrap or name the type, to avoid potential confusion (e.g. see `RawContentValue`).

Some of the places:
- types in `ethportal-api/src/types/portal_wire.rs`
- many functions in [OverlayProtocol](https://github.com/ethereum/trin/blob/6c7b81c5d693b686d1eabe2abda741d85218776c/portalnet/src/overlay/protocol.rs#L65)
- many functions in [OverlayService](https://github.com/ethereum/trin/blob/6c7b81c5d693b686d1eabe2abda741d85218776c/portalnet/src/overlay/service.rs#L100)
- other types in `portalnet/src/` (example `portalnet/src/find/query_info.rs`)
- many functions [ContentStore](https://github.com/ethereum/trin/blob/6c7b81c5d693b686d1eabe2abda741d85218776c/trin-storage/src/lib.rs#L47)
- [ValidationResult](https://github.com/ethereum/trin/blob/6c7b81c5d693b686d1eabe2abda741d85218776c/trin-validation/src/validator.rs#L5)

etherhood:
Can I take this one?

morph-dev:
> Can I take this one?

@etherhood, I would wait for #1403 to be merged first, as it does background for this work.
You can probably start working on top of it, but you might have to merge and resolve conflicts if there are some comments/concerns there.

etherhood:
I see its merged now, If it is fine with you, I will start on it

etherhood:
I have added some commits in WIP PR #1431 
I changed inner type of RawContentKey and fixed its uses across codebase.

Should I also change Vec<u8> type to Bytes in all places wherever Vec<u8> is being used, as I couldn't find RawContentKey being used in `portalnet/src/find/query_info.rs`, which you mentioned in original issue, but there are usage of Vec<u8>

morph-dev:
> Should I also change Vec type to Bytes in all places wherever Vec is being used, as I couldn't find RawContentKey being used in `portalnet/src/find/query_info.rs`, which you mentioned in original issue, but there are usage of Vec

[...]

PR #1431, opened 2024-09-09
refactor: change RawContentKey to alloy_primites::Bytes

etherhood:
### What was wrong?
`Vec<u8>` was being used to represent Raw Content Key.

See #1404 for details

### How was it  #fixed?
Replaced it with `alloy_primites::Bytes`

### To-Do

### DIFF
--- a/ethportal-api/src/types/content_key/state.rs
+++ b/ethportal-api/src/types/content_key/state.rs
@@ -92,18 +92,6 @@ impl OverlayContentKey for StateContentKey {
     }
 }
 
-impl From<&StateContentKey> for RawContentKey {
-    fn from(val: &StateContentKey) -> Self {
-        val.to_bytes()
-    }
-}
-
-impl From<StateContentKey> for RawContentKey {
-    fn from(val: StateContentKey) -> Self {
-        val.to_bytes()
-    }
-}
-
 impl TryFrom<RawContentKey> for StateContentKey {
     type Error = ContentKeyError;
     fn try_from(value: RawContentKey) -> Result<Self, Self::Error> {
--- a/ethportal-api/src/types/content_key/state.rs
+++ b/ethportal-api/src/types/content_key/state.rs
@@ -8,7 +8,7 @@ use std::{fmt, hash::Hash};
 use crate::{
     types::{content_key::overlay::OverlayContentKey, state_trie::nibbles::Nibbles},
     utils::bytes::hex_encode_compact,
-    ContentKeyError,
+    ContentKeyError, RawContentKey,
 };
 
 // Prefixes for the different types of state content keys:
@@ -70,7 +70,7 @@ impl OverlayContentKey for StateContentKey {
         sha256.finalize().into()
     }
 
-    fn to_bytes(&self) -> Vec<u8> {
+    fn to_bytes(&self) -> RawContentKey {
         let mut bytes: Vec<u8> = vec![];
 
         match self {
@@ -88,25 +88,25 @@ impl OverlayContentKey for StateContentKey {
             }
         }
 
-        bytes
+        bytes.into()
     }
 }
 
 impl From<&StateContentKey> for Vec<u8> {
     fn from(val: &StateContentKey) -> Self {
-        val.to_bytes()
+        val.to_bytes().to_vec()
     }
 }
 
 impl From<StateContentKey> for Vec<u8> {
     fn from(val: StateContentKey) -> Self {
-        val.to_bytes()
+        val.to_bytes().to_vec()
     }
 }
 
-impl TryFrom<Vec<u8>> for StateContentKey {
+impl TryFrom<RawContentKey> for StateContentKey {
     type Error = ContentKeyError;
-    fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
+    fn try_from(value: RawContentKey) -> Result<Self, Self::Error> {
         let Some((&selector, key)) = value.split_first() else {
             return Err(ContentKeyError::from_decode_error(
                 DecodeError::InvalidLengthPrefix {
@@ -187,7 +187,7 @@ impl fmt::Display for StateContentKey {
 mod test {
     use std::{path::PathBuf, str::FromStr};
 
-    use alloy_primitives::{keccak256, Address};
+    use alloy_primitives::{bytes, keccak256, Address, Bytes};
     use anyhow::Result;
     use rstest::rstest;
     use serde_yaml::Value;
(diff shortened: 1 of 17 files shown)


morph-dev (commented):
This is a good start, but I think that this PR should also make some changes in [OverlayContentKey](https://github.com/ethereum/trin/blob/d83cbe784b4ca7dc134f3159b4d9b8ffbcba766f/ethportal-api/src/types/content_key/overlay.rs#L12).

Here are the changes that we should do:
- remove `Into<Vec<u8>>` from supertraits set (we should use `to_bytes` directly)
- change `TryFrom<Vec<u8>, Error = ContentKeyError>` to `TryFrom<RawContentKey, Error = ContentKeyError>` in supertraits set
- `to_bytes` function should return `RawContentKey`

I think some of these are trivial, but some might require refactoring (I don't know until I try). If refactoring is not huge, I would put it into this PR, otherwise, it would be fine to split it into multiple PRs.

morph-dev:
Also, see https://www.conventionalcommits.org/en/v1.0.0/ for failing check.

etherhood:
Implemented `RawContentKey` in `OverlayContentKey` and in downstream traits as well

portalnet/src/overlay/service.rs
```diff
@@ -1029,7 +1030,7 @@ where
         let content_keys: Vec<TContentKey> = request
             .content_keys
             .into_iter()
-            .map(|k| (TContentKey::try_from)(k))
+            .map(|k| (TContentKey::try_from)(k.as_ssz_bytes().into()))
```
morph-dev, line 1033: nit: Can't this be: `TContentKey::try_from(k)`?

In which case, entire line might be:

```rust
.map(TContentKey::try_from)
```

portalnet/src/overlay/service.rs
```diff
@@ -1653,7 +1654,7 @@ where
             }
         };
         let request = Request::FindContent(FindContent {
-            content_key: content_key.clone().into(),
+            content_key: content_key.clone().to_bytes().to_vec(),
```
morph-dev, line 1657: nit: also here, can't this be: `content_key.to_bytes().to_vec()`?

portalnet/src/overlay/service.rs
```diff
@@ -2021,7 +2022,7 @@ where
         let content_keys_offered: Result<Vec<TContentKey>, TContentKey::Error> =
             content_keys_offered
                 .into_iter()
-                .map(|key| TContentKey::try_from(key))
+                .map(|key| TContentKey::try_from(key.as_ssz_bytes().into()))
```
morph-dev, line 2025: nit: also here, can't this be `TContentKey::try_from(key)`?

trin-beacon/src/storage.rs
```diff
@@ -125,7 +125,7 @@ impl ContentStore for BeaconStorage {
     type Key = BeaconContentKey;
 
     fn get(&self, key: &Self::Key) -> Result<Option<Vec<u8>>, ContentStoreError> {
-        let content_key: Vec<u8> = key.clone().into();
+        let content_key: RawContentKey = key.clone().to_bytes();
```
morph-dev, line 128: nit: and here, no need for '.clone()'

trin-beacon/src/storage.rs
```diff
@@ -217,7 +217,7 @@ impl ContentStore for BeaconStorage {
         &self,
         key: &Self::Key,
     ) -> Result<ShouldWeStoreContent, ContentStoreError> {
-        let content_key: Vec<u8> = key.clone().into();
+        let content_key: RawContentKey = key.clone().to_bytes();
```
morph-dev, line 220: nit: and here

trin-beacon/src/storage.rs
```diff
@@ -379,7 +379,7 @@ impl BeaconStorage {
         value: &Vec<u8>,
     ) -> Result<(), ContentStoreError> {
         let content_id = key.content_id();
-        let content_key: Vec<u8> = key.clone().into();
+        let content_key: RawContentKey = key.clone().to_bytes();
```
morph-dev, line 382: and here

trin-validation/src/header_validator.rs
```diff
@@ -274,7 +274,7 @@ mod test {
         // Validate content_key decodes
         let raw_ck = obj.get("content_key").unwrap().as_str().unwrap();
         let raw_ck = hex_decode(raw_ck).unwrap();
```
morph-dev, line 276: Should be `RawContentKey::from_str`

ethportal-api/src/types/content_key/state.rs
```diff
@@ -242,16 +242,20 @@ mod test {
     #[test]
     fn decode_empty_key_should_fail() {
         assert_eq!(
-            StateContentKey::try_from(vec![]).unwrap_err().to_string(),
+            StateContentKey::try_from(Bytes::new())
```
morph-dev, line 245: nit: Use `RawContentKey::new()` instead of `Bytes::new()`

portalnet/src/find/query_info.rs
```diff
@@ -73,7 +73,7 @@ impl<TContentKey: OverlayContentKey> QueryInfo<TContentKey> {
                 Request::FindNodes(FindNodes { distances })
             }
             QueryType::FindContent { ref target, .. } => Request::FindContent(FindContent {
-                content_key: target.clone().into(),
+                content_key: target.clone().to_bytes().to_vec(),
```
morph-dev, line 76: nit: we don't need `.clone()`

morph-dev (commented):
Few comments, repeating in many places:

- we should use `RawContentKey::_` instead of `Bytes::_` (example: Bytes::new())
    - this allows it easier to later replace `RawContentKey` with something else (instead of being alias) if we want to
    - it also makes code more explicit
- we should avoid unnecessary creating new objects, examples:
    - use `RawContentKey::from_str` directly (instead of using `hex_decode` to get `Vec<u8>` and then convert that into RawContentKey)
    - remove `.clone()` when not needed).

ethportal-api/src/types/content_key/state.rs
```diff
@@ -88,19 +88,19 @@ impl OverlayContentKey for StateContentKey {
             }
         }
 
-        bytes.into()
+        RawContentKey::from(bytes)
     }
 }
 
-impl From<&StateContentKey> for Vec<u8> {
+impl From<&StateContentKey> for RawContentKey {
```
morph-dev, line 95: This should be removed (similar to other ones)

ethportal-api/src/types/content_key/state.rs
```diff
@@ -88,19 +88,19 @@ impl OverlayContentKey for StateContentKey {
...
     fn from(val: &StateContentKey) -> Self {
-        val.to_bytes().to_vec()
+        val.to_bytes()
     }
 }
 
-impl From<StateContentKey> for Vec<u8> {
+impl From<StateContentKey> for RawContentKey {
```
morph-dev, line 101: This one as well.

morph-dev (approved):
Looks good! If you can address last 2 comments, and I will merge the PR.

morph-dev:
You can ignore failing commitlint check, as I will squash and merge changes anyway.

But for the future, problem is the commit message: "fix: CI fixes and resolved conflicts with master", because it starts with capital letter (maybe something that we should change in our configuration).

etherhood:
Fixed the commit message and remove From impl for StateContentKey

Issue #1406, opened 2024-08-29
beacon-network: Add `LightClientBootstrap` and `LightClientUpdate`  consensus specs validations  on protocol level


ogenev:
Trin use very basic validation for `LightClientBootstrap` and `LightClientUpdate` objects when the content is accepted.

We should add the light client validation methods from consensus [specs](https://github.com/ethereum/consensus-specs/blob/dev/specs/altair/light-client/sync-protocol.md#validate_light_client_update) (same validations that we do when syncing the light client service from a trusted block root) for validating `LightClientUpdate`.  For `LightClientBootstrap` we can compare the header with the finalized header from the recent finality update.

ToDo:

- [x] Validate `LightClientBootstrap`
- [x] Validate `LightClientUpdate`

PR #1433, opened 2024-09-10
feat(beacon): improve `LightClientBootstrap` validation

ogenev:
### What was wrong?

Fixes part 1 of #1406.

### How was it fixed?
- Add JSON-RPC endpoint to get the recently finalized beacon block header
- Compare the `LightClientBootstrap` header with the recently finalized beacon block header before gossiping and storing the content.

### To-Do


KolbyML (approved):
:shipit:

PR #1434, opened 2024-09-10
feat(beacon-bridge): delay `LightClientBootstrap` gossip with 2 slots

ogenev:
### What was wrong?

When a new beacon block root is finalized, we generate and gossip to the beacon network `LightClientBootstraps` and `LightClientFinalityUpdates`. Usually, bootstraps are generated first and propagated to the network. 

The problem is that the bootstrap validation is anchored to the header in the `LightClientFinalityUpdate` so to verify `LightClientBootstrap` content, we first must ensure the clients are synched and the latest `LightClientUpdate` is propagated **before** the bootstrap.

### How was it fixed?

- add a 2 slots delay to `LightClientBootstrap` gossip. This should be enough for all clients to receive the finality update first and validate the bootstrap later.

### To-Do

### DIFF
--- a/portal-bridge/src/bridge/beacon.rs
+++ b/portal-bridge/src/bridge/beacon.rs
@@ -70,6 +70,20 @@ impl FinalizedBeaconState {
     }
 }
 
+/// A helper struct to hold the finalized block root and whether the `LightClientBootstrap`
+/// generation is in progress.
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

PR #1435, opened 2024-09-10
refactor: reduce noisy error logs

njgheorghita:
### What was wrong?
I'm open to pushback on this.. but in my opinion the error logs when starting up a node are too noisy... `error!` messages should be reserved for critical errors. It should just be a `warn!` log if we can't connect to a given bootnode in a subnetwork. And only display the `error!` if you're node was unable to connect to any bootnodes, therefore meaning your node was completely unable to connect to the network.

### How was it fixed?
- changed log

### To-Do

### DIFF
--- a/portalnet/src/overlay/protocol.rs
+++ b/portalnet/src/overlay/protocol.rs
@@ -750,7 +750,7 @@ where
                     successfully_bonded_bootnode = true;
                 }
                 Err(err) => {
-                    error!(
+                    warn!(
                         alias = %bootnode.alias,
                         protocol = %self.protocol,
                         error = %err,
@@ -762,7 +762,7 @@ where
         if !successfully_bonded_bootnode {
             error!(
                 protocol = %self.protocol,
-                "Failed to bond with any bootnodes",
+                "Failed to bond with any bootnodes, unable to connect to subnetwork.",
             );
         }
     }


portalnet/src/overlay/protocol.rs
```diff
@@ -750,7 +750,7 @@ where
                     successfully_bonded_bootnode = true;
                 }
                 Err(err) => {
-                    error!(
+                    warn!(
```
KolbyML, line 753:
```suggestion
                    debug!(
```

I think we should only print a error if we fail connecting to all bootnodes,

A warn would still show this message every time on boot, which 1 bootnode failing isn't that note-able

^ this view is a discussion

njgheorghita: Yeah... my only question is that I still do find it pretty useful to know which bootnodes are failing... for example, right now trin's are failing. Which, otherwise I wouldn't know if we're still able to connect to a fluffy node. So I think this case is still worthy of a `warn!` level...

KolbyML: In that case `warn!()` is fine for now and we can think about moving it to debug!() in the future when things are more stable

KolbyML (approved):
:shipit:

portalnet/src/overlay/protocol.rs
KolbyML: > Maybe we could just compile a list of all "failed" bootnodes... something like... `warn!("Failed to connect to bootnodes: trin-ams3-1, fluffy1")` so that it's just a single warn log rather than many

I really like this idea

KolbyML: > Maybe we could just compile a list of all "failed" bootnodes... something like... `warn!("Failed to connect to bootnodes: trin-ams3-1, fluffy1")` so that it's just a single warn log rather than many

Maybe report the number of successful bootnodes connected to as well then, to make it not appear like things are only failing

PR #1436, opened 2024-09-10
feat: add block by number and remove epoch accumulators

KolbyML:
### What was wrong?
implements  https://github.com/ethereum/portal-network-specs/pull/334
### How was it fixed?

by removing our uses of epoch accumulators as a content type

I reworked our peertests which used  epochaccumulators  using some test data Nick gave me to use other history types instead


KolbyML:
@njgheorghita I would like your feedback on what we will do with the 5 tests which were added which uses `EpochAccumulators` in https://github.com/ethereum/trin/pull/1345 and https://github.com/ethereum/trin/pull/1341

With `EpochAccumulators` being removed which are the biggest content type, and the minimum storage requirement being 1mb. What do you think the best way forward would be

ethportal-peertest/src/scenarios/gossip.rs
```diff
@@ -90,37 +91,99 @@ pub async fn test_gossip_dropped_with_offer(peertest: &Peertest, target: &Client
...
+    )
+    .await
+    .is_err());
+    assert!(HistoryNetworkApiClient::local_content(
+        &peertest.nodes[0].ipc_client,
+        receipts_key_1.clone()
     )
     .await
     .is_err());
     // check that peertest bootnode does not have accumulator_1
```
njgheorghita, line 171: comment out of date

ethportal-peertest/src/scenarios/gossip.rs
```diff
@@ -202,18 +320,42 @@ pub async fn test_gossip_dropped_with_find_content(peertest: &Peertest, target:
...
+        receipts_value_1.encode(),
+    )
+    .await
+    .unwrap();
+    assert!(store_result);
 
-    // Store accumulator_2 locally in target
-    let (acc_key_2, acc_value_2) = fixture_epoch_acc_2();
+    // Store block_2 minus receipts locally in target
+    let (header_key_2, header_value_2) = fixture_header_by_hash_with_proof_15040709();
```
njgheorghita, line 350: I don't think you actually need to store the header locally in every target. As long as it's available in any single node, it should be available to the others for validation.

ethportal-peertest/src/scenarios/gossip.rs
```diff
@@ -90,37 +91,99 @@ pub async fn test_gossip_dropped_with_offer(peertest: &Peertest, target: &Client
         .unwrap();
     let fresh_enr = fresh_target.node_info().await.unwrap().enr;
 
-    // Store accumulator_1 locally in client that is not connected to the network
-    let (acc_key_1, acc_value_1) = fixture_epoch_acc_1();
+    // Store block_1 locally in client that is not connected to the network
```
njgheorghita, line 94: Sorry, but I realize now that the large content values I gave were not ideal, and it shows in the complexity of this test, making it difficult to follow what's going on...
I updated #1438 with new assets that should simplify this test... 
You should be able to write a test that's accurately testing the gossip-dropped feature using only the..
- receipts from block `15040641` (320kb)
- body from block  `15040708` (763kb)
- (and the 2 headers for validation)
As it stands, there's just too much going on and it's tough to follow the exact code paths that are being tested

ethportal-peertest/src/scenarios/gossip.rs
KolbyML: I am confused, am I not just storing the header in target?

KolbyML:
@njgheorghita ready for another look

ethportal-peertest/src/utils.rs
```diff
@@ -132,29 +156,100 @@ pub fn fixture_receipts() -> (HistoryContentKey, HistoryContentValue) {
...
+                "test_assets/mainnet/large_content/{block_number}/{dependent_type}.bin"
+            ))
+            .unwrap();
+            match dependent_type {
+                DependentType::BlockBody => {
+                    let content_key = HistoryContentKey::BlockBody(BlockBodyKey {
+                        block_hash: header_content_value.header.hash().0,
+                    });
+                    let content_value =
+                        HistoryContentValue::decode(&content_key, &dependent_value).unwrap();
```
njgheorghita, line 232: shouldn't clippy be complaining about this unwrap?

portal-bridge/src/api/execution.rs
```diff
@@ -75,7 +75,12 @@ impl ExecutionApi {
         &self,
         height: u64,
         epoch_acc: Option<Arc<EpochAccumulator>>,
-    ) -> anyhow::Result<(FullHeader, HistoryContentKey, HistoryContentValue)> {
+    ) -> anyhow::Result<(
+        FullHeader,
+        HistoryContentKey,
```
njgheorghita, line 80:
```rust
    HistoryContentKey, // BlockHeaderByHash
    HistoryContentKey, // BlockHeaderByNumber
```

portal-bridge/src/bridge/history.rs
```diff
@@ -341,25 +355,12 @@ impl HistoryBridge {
         &self,
```
njgheorghita, line 355: I'd also get rid of this function and just inline the relevant code since there's no more gossiping going on

portal-bridge/src/bridge/era1.rs
```diff
@@ -246,13 +250,20 @@ impl Era1Bridge {
                 panic!("Failed to get epoch from era1 file: {e}");
             }
         };
-        let epoch_acc = match self.get_epoch_acc(epoch_index).await {
+        let (_, epoch_acc) = match lookup_epoch_acc(
```
njgheorghita, line 253: It doesn't seem like we're using the `epoch_hash` that's being returned as the first value anymore, so you could just update this util fn to return the `epoch_acc`

trin-history/src/validation.rs
```diff
@@ -30,18 +29,36 @@ impl Validator<HistoryContentKey> for ChainHistoryValidator {
...
+                    .await
+                    .header_validator
+                    .validate_header_with_proof(&header_with_proof)?;
+
+                Ok(ValidationResult::new(true))
+            }
+            HistoryContentKey::BlockHeaderByNumber(key) => {
+                let header_with_proof =
+                    HeaderWithProof::from_ssz_bytes(content).map_err(|err| {
+                        anyhow!("Header with proof content has invalid encoding: {err:?}")
```
njgheorghita, line 54: Maybe it'd be useful to differentiate this error message from the one on line 35

trin-validation/src/header_validator.rs
```diff
@@ -226,82 +226,21 @@ mod test {
...
-    #[case(1_000_003)]
-    #[case(1_000_004)]
-    #[case(1_000_005)]
-    #[case(1_000_006)]
-    #[case(1_000_007)]
-    #[case(1_000_008)]
-    #[case(1_000_009)]
-    #[case(1_000_010)]
-    #[tokio::test]
-    async fn generate_and_verify_header_with_proofs(#[case] block_number: u64) {
```
njgheorghita, line 261: This test still seems somewhat useful to leave in. We still want to test that trin constructs & validates (and invalidates) these header with proofs...
I would just update this test (& `invalidate_invalid_proofs`) to lookup the epoch accs directly from disk and use `PreMergeAccumulator::construct_proof()` instead of the removed `PreMergeAccumulator::generate_proof()`

trin-history/src/validation.rs
```diff
@@ -162,7 +151,7 @@ mod tests {
         let hwp = HeaderWithProof::from_ssz_bytes(&hwp_ssz).expect("error decoding header");
```
njgheorghita, line 151: Seems worthwhile at least adding a simple validate/invalidate test for `BlockHeaderWithNumber` here

ethportal-peertest/src/utils.rs
KolbyML: This is in `ethportal-peertest`, I thought testing stuff didn't have that check

KolbyML:
@njgheorghita ready for another look

njgheorghita (commented):
Looks good! One last thing to consider before merging... we still have some number of epoch accumulators out there in the dbs of our network nodes. I think we want to drop all of those out before deploying this change. So I think we should hold off on merging this until we've done the filtering. I can dust off #1286 tomorrow and then once that's deployed / the filtering is done, we can merge this pr

njgheorghita (approved):
Moving ahead with the plan from https://github.com/ethereum/trin/pull/1286#issuecomment-2356427362, so this is good to go on my end after @ogenev comments are addressed

PR #1439, opened 2024-09-11
feat: add support for async transaction execution

morph-dev:
### What was wrong?

The `trin-execution` doesn't support async block/transaction execution. This is needed if we want to use state network to execute transactions (e.g. eth_call, eth_estimateGas).

### How was it fixed?

Created `AsyncDatabase` trait and added support for using it in order to execute transactions.

Some refactoring of the `trin-execution` crate:
- moved some files into `evm` module
- extracted shared logic (creating BlockEnv, creating Evm with/without tracer) into separate functions (simplifying BlockExecutor a bit)

### To-Do


morph-dev:
@KolbyML, I didn't do any end-to-end testing to make sure that trin-execution still works. Any recommended way to do it quickly (quickly in the sense that it doesn't take days to run it)?

KolbyML:
> @KolbyML, I didn't do any end-to-end testing to make sure that trin-execution still works. Any recommended way to do it quickly (quickly in the sense that it doesn't take days to run it)?

if you run `cargo run -p trin-execution --release`, it should be able to execute the first million or 2 in 2 hours or something, I haven't executed the dos attack in a while so we might not execute it in 1 go anymore which would make things much slower

trin-execution/src/evm/mod.rs
```diff
@@ -1,3 +1,61 @@
...
+        .map(|excess_blob_gas| BlobExcessGasAndPrice::new(excess_blob_gas.to()));
+
+    BlockEnv {
+        number: U256::from(header.number),
+        coinbase: header.author,
+        timestamp: U256::from(header.timestamp),
+        gas_limit: header.gas_limit,
+        basefee: header.base_fee_per_gas.unwrap_or_default(),
+        difficulty: header.difficulty,
+        prevrandao: header.mix_hash,
```
KolbyML, line 28: `prevrandao: header.mix_hash,` I believe we should still be checking if we are past the Merge for `prevrandao`.

mix_hash won't be none in pre-merge blocks, it was used in PoW. Maybe it will be fine? but I would rather not take chances

For difficulty I think what you did is fine as the processed header should always have the right difficulty

KolbyML (approved):
:shipit: looks mostly good, one concern with how `mix_hash` is used because the way the code is written, it wouldn't match the  code before this PR in that, it would set `BlockEnv::prevrandao` to Some() for pre-merge blocks, where before this PR we had a specifically set it to `None` for pre-merge blocks.

Maybe this might be fine, and Revm is smart enough to ignore this, but I think we should re-add the if merge {} check for `BlockEnv::prevrandao`, for sanity sake.

trin-execution/src/evm/mod.rs
morph-dev: I didn't know that `header.mix_hash` won't be none pre-merge. I fixed it.

morph-dev:
> if you run `cargo run -p trin-execution --release`, it should be able to execute the first million or 2 blocks in 2 hours or something, I haven't executed the dos attack in a while so we might not execute it in 1 go anymore which would make things much slower

I will let it run over night and merge in the morning if everything looks ok.

morph-dev:
> I will let it run over night and merge in the morning if everything looks ok.

It managed to process 3'981'311 blocks, but then it failed because it panicked while deserializing era1 file.
I downloaded problematic file and deserialization worked fine, so I assume that there was some network issue and somehow wrong file (bytes) was downloaded. Should we add hash check and retry to prevent this in the future?

Then I tried to resume execution (from the latest saved checkpoint - block 3'700'003), but that one failed. This was also not caused by this PR. I investigated it and figure out the problem (see #1440). I managed to fix the issue locally (in a bit hacky way), and I was able to continue execution.

Execution is still running, it passed 4'400'000 as I'm writing this. I will consider this a successful test, at least in regard to this PR, and I will merge it now.

Issue #1444, opened 2024-09-13
Support risc-v


come-maiz:
RISC-V is a good start to get Ethereum nodes running in free hardware, which is a necessary step for a local-first decentralized economy. Lightclients are another necessary step, so these two worlds together are very powerful.

Please support the risc-v architecture in this project.

come-maiz:
Next we have to update rustls, which is problematic because surf keeps it pinned to an old version and it is unmaintained:
https://github.com/http-rs/surf/issues/352#issuecomment-1709404352

KolbyML:
> Next we have to update rustls, which is problematic because surf keeps it pinned to an old version and it is unmaintained: [http-rs/surf#352 (comment)](https://github.com/http-rs/surf/issues/352#issuecomment-1709404352)

Yeah I guess we should remove it as a dependency and switch to something else in that case

KolbyML:
> Next we have to update rustls, which is problematic because surf keeps it pinned to an old version and it is unmaintained: [http-rs/surf#352 (comment)](https://github.com/http-rs/surf/issues/352#issuecomment-1709404352)

https://github.com/ethereum/trin/issues/1446

I moved a PR to move us to reqwest instead of surf, hopefull this helps with the mission

come-maiz:
Nice! Thank you @KolbyML. It works.

Next is ethers-providers:
https://github.com/gakonst/ethers-rs/issues/2786

KolbyML:
> Nice! Thank you @KolbyML. It works.
> 
> Next is ethers-providers: [gakonst/ethers-rs#2786](https://github.com/gakonst/ethers-rs/issues/2786)

We want to remove `etheres-rs` in Trin and replace it with `alloy`

come-maiz:
That sounds perfect.

KolbyML:
@come-maiz This PR https://github.com/ethereum/trin/pull/1456 removes our dependency on `ethers-rs`, are you able to compile on risc-v yet, or are there still some dependency issues?

come-maiz:
thanks @KolbyML!

Compilation almost finishes now. It's getting stuck in librocksdb-sys, I'll see what I can do on my machine to get it to compile.

KolbyML:
> thanks @KolbyML!
> 
> Compilation almost finishes now. It's getting stuck in librocksdb-sys, I'll see what I can do on my machine to get it to compile.

@come-maiz try installing clang and tell me if it works `sudo apt install clang`

[...]

PR #1445, opened 2024-09-13
chore: replace eth2_hashing with ethereum_hashing

come-maiz:
This is a necessary step to update ring, which is necessary to build in risc-v.

Part of #1444.

### What was wrong?

trin fails to build in risc-v

### How was it fixed?

It is still not fixed. This is just one step forward.

### To-Do


come-maiz:
Thanks to @haurog and @michaelsproul for fixing the ring dependency in ethereum_hashing:
https://github.com/sigp/ethereum_hashing/pull/8

KolbyML:
@come-maiz CI is going to fail because how how you named your commit, could you amend the commits name following these guidelines https://www.conventionalcommits.org/en/v1.0.0/

KolbyML:
'chore: replace eth2_hashing with ethereum_hashing'

^ something like this would be good

come-maiz:
Yes, done.

come-maiz:
Yes, done ^_^

KolbyML:
> Yes, done ^_^

Thank you for your first contribution

PR #1447, opened 2024-09-13
Credit first peer to return content

carver:
### What was wrong?

While working on #1395, a number of cleanups accumulated. But there's enough going on in that PR that it's nice to separate out the stuff that can be. This is a mashup of those changes, with more info in each commit.

Most prominent, and the only change in functionality, is that when content arrives twice, we ignore the 2nd to arrive (instead of ignoring the first to arrive). We would like to give credit to the fastest peer to return the content.

### How was it fixed?

Just ignore additional content to come in. This of course continues to mean that if the first content to come in is invalid, then the whole query fails. #1395 is the next step toward improving that behavior.

Everything else is just:
- more logging lines, or info in the lines
- making Node ID loggable
- skip an unnecessary clone of content (which can be quite large)
- a new clippy warning on a recent nightly build of rust

### To-Do

### DIFF
--- a/portalnet/src/find/iterators/findcontent.rs
+++ b/portalnet/src/find/iterators/findcontent.rs
@@ -73,7 +73,7 @@ struct UtpAndPeerDetails<TNodeId> {
 }
 
 #[derive(Debug, Clone)]
-pub struct FindContentQuery<TNodeId> {
+pub struct FindContentQuery<TNodeId: std::fmt::Display> {
     /// The target key we are looking for
     target_key: Key<TNodeId>,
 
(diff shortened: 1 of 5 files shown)


portalnet/src/find/iterators/findcontent.rs
```diff
@@ -525,42 +530,48 @@ mod tests {
...
 
             'finished: loop {
-                if expected.is_empty() {
+                if remaining.is_empty() {
+                    debug!("ending test: no more peers to pull from");
                     break;
                 }
                 // Split off the next up to `parallelism` expected peers.
-                else if expected.len() < max_parallelism {
+                else if remaining.len() < max_parallelism {
```
morph-dev, line 560: nit: can't we simplify this to:

```rust
expected = remaining.drain(..min(max_parallelism, remaining.len()));
```

morph-dev (approved):
Looks good. :ship:

njgheorghita (approved):

Issue #1446, opened 2024-09-13
Remove surf and replace it with reqwest


KolbyML:
https://github.com/ethereum/trin/issues/1444 Risc-V support is blocked by surf https://github.com/http-rs/surf/issues/352#issuecomment-1709404352 surf has been unmaintained for the last 2 years. 

It seems like the 2 best options are 
- hyper
- reqwest

Apparently reqwest is built on hyper, with some extra utils which is nice

morph-dev:
I just want to highlight that we already depend on both hyper (in [rpc](https://github.com/ethereum/trin/blob/ba63fb289c889a8ef8cfd63838633377adcde3f7/rpc/Cargo.toml#L19) and [ethportal-peertest](https://github.com/ethereum/trin/blob/ba63fb289c889a8ef8cfd63838633377adcde3f7/ethportal-peertest/Cargo.toml#L22) ) and reqwest (in [light-client](https://github.com/ethereum/trin/blob/ba63fb289c889a8ef8cfd63838633377adcde3f7/light-client/Cargo.toml#L25)) in our repo.

I think #1448 (replace surf with request) can go as is, but I think we should create separate PR to migrate hyper to reqwest as well.

PR #1448, opened 2024-09-13
refactor: remove unmaintained surf for reqwest

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/issues/1446 surf has been unmaintained for over 2 years now, it is also using a old version of packages which prevents us from compiling Trin to Risc-v https://github.com/ethereum/trin/issues/1444
### How was it fixed?

Just convert `surf` -> `reqwest` equivalents, which was pretty straight forward

### DIFF
--- a/portal-bridge/src/api/execution.rs
+++ b/portal-bridge/src/api/execution.rs
@@ -21,7 +21,6 @@ use ethportal_api::{
     Receipts,
 };
 use futures::future::join_all;
-use reqwest_middleware::ClientWithMiddleware;
 use serde_json::{json, Value};
 use tokio::time::sleep;
 use tracing::{debug, error, warn};
@@ -32,7 +31,7 @@ use trin_validation::{
 use url::Url;
 
 use crate::{
-    cli::url_to_client,
+    cli::{url_to_client, ClientWithBaseUrl},
     constants::{FALLBACK_RETRY_AFTER, GET_RECEIPTS_RETRY_AFTER},
     types::full_header::FullHeader,
 };
@@ -273,9 +272,6 @@ impl ExecutionApi {
         Ok(header.number)
     }
 
-    /// Used the "surf" library here instead of "ureq" since "surf" is much more capable of handling
-    /// multiple async requests. Using "ureq" consistently resulted in errors as soon as the number
-    /// of concurrent tasks increased significantly.
     async fn batch_requests(&self, obj: Vec<JsonRequest>) -> anyhow::Result<String> {
         let batched_request_futures = obj
             .chunks(BATCH_LIMIT)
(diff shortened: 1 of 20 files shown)


src/bin/test_providers.rs
```diff
@@ -425,24 +423,39 @@ impl Providers {
...
+        "CF-Access-Client-Id",
+        HeaderValue::from_str(&PANDAOPS_CLIENT_ID)
+            .map_err(|_| anyhow!("Invalid CF-Access-Client-Id header value"))?,
+    );
+    headers.insert(
+        "CF-Access-Client-Secret",
+        HeaderValue::from_str(&PANDAOPS_CLIENT_SECRET)
+            .map_err(|_| anyhow!("Invalid CF-Access-Client-Secret header value"))?,
+    );
+    let client = Client::builder()
```
morph-dev, line 438: nit: I think this can be simplified to something like:

```rust
let response = Client::new()
    .post("...")
    .headers(headers)
    .json(&request)
    .send()
    .await
    .map_err(|e| ...)?
```

src/bin/test_providers.rs
```diff
@@ -425,24 +423,39 @@ impl Providers {
...
+        "id": 1
+    });
+    let response = client
+        .post("https://geth-lighthouse.mainnet.eu1.ethpandaops.io/")
+        .json(&request_body)
+        .send()
+        .await
+        .map_err(|e| anyhow!("Request failed: {:?}", e))?;
+    let response_text = response
+        .text()
```
morph-dev, line 455: nit: we could use `.json::<serde_json::Value>()`

trin-execution/src/era/manager.rs
```diff
@@ -30,10 +33,10 @@ pub struct EraManager {
 
 impl EraManager {
     pub async fn new(next_block_number: u64) -> anyhow::Result<Self> {
-        let http_client: Client = Config::new()
-            .add_header("Content-Type", "application/xml")
-            .expect("to be able to add header")
-            .try_into()?;
+        let mut headers = HeaderMap::new();
```
morph-dev, line 36: Now that I think about it, it doesn't make sense for `CONTENT_TYPE` header to be set here and passed to `get_era_files` and `get_era1_files` (those functions are in a different crate, they control the URL and they are responsible for parsing the response).

So in this case, I would probably just create simple `Client::new()`, and each function would initialize headers that they need.

Not really part of this pr, but I think it's small enough to be included.

morph-dev (commented):
Mostly looks good, but there is one big issue in `portal-bridge` (see comment).

Also, `light-client` crate also depends on reqwest, so we should update that dependency (should use workspace version)

portal-bridge/src/api/execution.rs
```diff
@@ -51,7 +51,7 @@ pub struct ExecutionApi {
 }
 
 impl ExecutionApi {
-    pub async fn new(primary: Url, fallback: Url) -> Result<Self, surf::Error> {
+    pub async fn new(primary: Url, fallback: Url) -> Result<Self, reqwest_middleware::Error> {
         // Only check that provider is connected & available if not using a test provider.
         debug!(
             "Starting ExecutionApi with primary provider: {primary} and fallback provider: {fallback}",
```
njgheorghita, line 57: Comment on line 276 should be removed, and probably some testing to make sure that `reqwest` can handle the async load, though it's a very popular library so I expect it will

portal-bridge/src/cli.rs
```diff
@@ -155,38 +154,49 @@ pub struct BridgeConfig {
...
-            "null".to_string()
-        });
-        config = config
-            .clone()
-            .add_header("CF-Access-Client-Id", client_id)
-            .map_err(|_| "to be able to add client id header")?
-            .add_header("CF-Access-Client-Secret", client_secret)
-            .map_err(|_| "to be able to add client secret header")?
-            .set_base_url(url);
+pub fn url_to_client(url: Url) -> Result<ClientWithMiddleware, String> {
```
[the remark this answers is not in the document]
KolbyML: I will just implement base_url for Reqwest

KolbyML:
@njgheorghita @morph-dev ready for another review. I believe I resolved all the PR concerns
image
image
Here are a few pictures of history latest nothing seems wrong

Running backfill using infura I see what you mean nick when you said shisui uTP failures. It looks good as well

njgheorghita:
Cool. I'm guessing this comment predates era1 files, so back then when in `backfill` mode, we'd be making quite a bit of parallel http requests. But, now that we have era1 files (and plan to use some version of era files for everything from merge -> latest?) I actually don't think this is much of an issue anymore. Unless there's somewhere else in the bridge that I'm missing where we'll have large amounts of parallel http requests....

KolbyML:
> large amounts of parallel http requests

I just ran backfill using infura as I said before, so it was making a large amount of parallel http requests. Reqwest's get 100 times more downloads a month than surf and is also built on hyper so I don't see why it wouldn't be able to handle `large amounts of parallel http requests`

morph-dev (approved):
LGTM. But wait for Nick's approval as well.

ogenev (approved):
LGTM

njgheorghita (approved):

PR #1449, opened 2024-09-13
fix(bridge): update utp tx limit with state offer limit

njgheorghita:
### What was wrong?
First steps from looking into trin's concurrency capabilities. Running the bridge w/ a gossip limit was using the node's default `utp_transfer_limit`. These should be coupled.

I'm sorta ignoring history / era1 bridges right now. But, feel free to pushback if you think it should be addressed now & not later. The main rationale being that there is a fundamental difference in how tasks are spawned in offer / gossip mode. 
- offer mode (state bridge): we measure exactly the # of spawned jsonrpc requests 
- gossip mode (history/era1): we measure exactly the # of concurrent "gossip block" tasks, which each can individually spawn multiple concurrent jsonrpc requests. 

Given that the priority right now is the state bridge, I'm operating on the assumption that we should just leave the history/era1 bridges as-is for now, as they're working just fine for our requirements. In the state bridge, accuracy is very important due to the large volume of data we will be offering, so I just de-coupled the single "limit" into a `gossip-limit` & `offer-limit`. 

### How was it fixed?
- added cli config for 'offer-limit'
- use `offer-limit` when spawning a client in 'state' mode

### To-Do


portal-bridge/src/constants.rs
```diff
@@ -18,9 +18,16 @@ pub const BEACON_GENESIS_TIME: u64 = 1606824023;
...
-// This is the maximum number of active blocks being gossiped
+// The maximum number of active blocks being gossiped. Note that this doesn't
+// exactly mean the number of concurrent gossip jsonrpc requests, as the gossip
+// requests are batched / block.
+// Currently, this is only used in the history & era1 bridges
 pub const DEFAULT_GOSSIP_LIMIT: usize = 32;
 
+// The maximum number of concurrent offer jsonrpc requests that can be made simultaneously.
+// Currently, this is only used in the state bridge.
+pub const DEFAULT_OFFER_LIMIT: usize = 64;
```
KolbyML, line 29: Shouldn't this be `32 * 4` or `32 * 8`, as for 1 gossip limit, trin can spawn 8 offers

njgheorghita: I don't think so, since we're using the the semaphore to limit the spawning of individual offer tasks, not entire blocks. Which, imo is the key distinction between state & (history / era1) bridges.

https://github.com/ethereum/trin/blob/master/portal-bridge/src/bridge/state.rs#L248

njgheorghita: Or maybe I'm misunderstanding...

njgheorghita: I think it's cleaner to just decouple the concept of an `offer_limit` from a `gossip_limit` entirely

KolbyML: > I think it's cleaner to just decouple the concept of an `offer_limit` from a `gossip_limit` entirely

I think this is the right direction, I just think 64 is way to low

njgheorghita: Ahh, gotcha... well so far in my testing it seems to hold up pretty well up to a gossip-limit of ~100, but not so much over 150. Haven't exactly nailed down the exact number yet, but I'd say we can just leave it for now and I'll update this value later once I've got a better idea of where the actual limit is

njgheorghita: Or, I could do 100 if you feel like it's worthwhile, but if a user wants they can always manually increase this limit using a cli flag

njgheorghita: Ok, I understand your reasoning now, but I suspect that earlier when running a gossip limit of 32 (aka 32 * 4 concurrent rpc requests) that they weren't actually all successful txs

KolbyML: > Or, I could do 100 if you feel like it's worthwhile, but if a user wants they can always manually increase this limit using a cli flag

oh ok 64 or 100 is good then up to you

KolbyML (approved):
:shipit:

portal-bridge/src/constants.rs
```diff
@@ -18,9 +18,16 @@ pub const BEACON_GENESIS_TIME: u64 = 1606824023;
 // overloaded and not performing well
 pub const HTTP_REQUEST_TIMEOUT: Duration = Duration::from_secs(20);
 
-// This is the maximum number of active blocks being gossiped
+// The maximum number of active blocks being gossiped. Note that this doesn't
```
morph-dev, line 21: nit: should we say "blocks" or "content key/value" or something similar (as I believe blocks is only correct in history network context).

morph-dev (approved):

portal-bridge/src/constants.rs
njgheorghita: I think it's fine. At the bottom of this comment, I'm explicit about `Currently, this is only used in the history & era1 bridges`, which are both in the history network context, in which "blocks" is actually the correct unit that the semaphore is limiting. Aka, each block being limited contains multiple key/value pair requests

PR #1450, opened 2024-09-14
refactor: simplify TrinExecution

morph-dev:
### What was wrong?

TrinExecution has quite of bit responsibility, which I feel it shouldn't have. It also has some functionality that is no longer used.

### How was it fixed?

- "should commit early" is suited much better in BlockExecutor 
- proof generation is no longer used (removed)
- transaction tracing: I feel that this belongs in TrinExecution more than it does in BlockExecution
    - BlockToTrace now creates folders and writer, if tracing is enabled
    - BlockExecution accepts function that creates optional tracer

### To-Do

### DIFF
--- a/trin-execution/src/evm/block_executor.rs
+++ b/trin-execution/src/evm/block_executor.rs
@@ -1,13 +1,17 @@
 use std::{
     collections::HashMap,
-    fs::{self, File},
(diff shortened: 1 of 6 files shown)


KolbyML (approved):
:shipit:

Issue #1440, opened 2024-09-12
Execution can't recover after crash


morph-dev:
While running trin execution, it happened that era1 deserialization failed (irrelevant to this issue).

When I tried to resume running it, it would fail very soon afterwards with error:
`Error: database error: not found database error block_hash`

After looking a bit more into it, I found the problem.

The `BlockExecutor::manage_block_hash_serve_window` modifies the db directly after every processed block. If the execution crashes (like it happened to me) and we try to resume it, the stored block hashes will not be the correct ones (we will have 256 blocks from the moment of crash, not the saved checkpoint).

Possible solutions:
1. (preferred) Keep track of block hashes in memory and flush them to this when the rest of state is flushed.
2. Before execution starts, make sure we have all required block hashes in db (and seed them if that's not the case)

morph-dev:
Alternatively, we can just never delete block_number->block_hash from the db. Clearly, not most optimized solution, but definitely the easiest one.

It's only ~64 bytes per block, so it's not the end of the world (total of ~1.2 GB for entire chain at the moment).

KolbyML:
https://github.com/ethereum/trin/pull/1451#issuecomment-2351083111
https://github.com/ethereum/trin/pull/1451#issuecomment-2351083985

Additional comments I made on this problem, and why switching to an ACID database solves them

morph-dev:
Why can't we use RocksDB? Instead of using [`rocksdb::DB`](https://docs.rs/rocksdb/latest/rocksdb/type.DB.html), we can use [`rocksdb::TransactionDB`](https://docs.rs/rocksdb/latest/rocksdb/struct.TransactionDB.html) or [`rocksdb::OptimisticTransactionDB`](https://docs.rs/rocksdb/latest/rocksdb/type.OptimisticTransactionDB.html). 
Difference between transaction and optimistic transaction can be found here: https://github.com/facebook/rocksdb/wiki/Transactions .

I think in our case, we can even use `rocksdb::DB::write`. Might be the simplest solution.

KolbyML:
Erigon has a write up here

https://github.com/erigontech/erigon/wiki/Choice-of-storage-engine

They tried like 5 different database solution then ended up with `MDBX`.

They say it isn't ACID,

[...]

PR #1451, opened 2024-09-14
feat(trin-execution): allow ctrl+c to immediately stop execution

morph-dev:
### What was wrong?

Using Ctrl+C to stop trin_execution can take a lot of time, partially due to committing to db (which can't be avoided), and partially because execution doesn't stop until next checkpoint (which can take another 30 min).

### How was it fixed?

Refactor how TrinExecution executes blocks (built on top of #1450).

Instead of executing until the first commit (which can come early), it will now execute all the way to the provided block but it accepts `stop_signal` receiver. When stop signal is received, it will commit and return.

Note: Tested it locally (around block 6'000'000).

### Future ideas

We can speed up the shutdown by abandoning changes that are in memory (by not committing anything), but we would have the same problem that is described in #1440.

Even if implemented, it's questionable if this is desired behavior, so maybe we would need a flag that controls whether to commit on ctrl+c or not.

### To-Do

### DIFF
--- a/trin-execution/src/execution.rs
+++ b/trin-execution/src/execution.rs
@@ -4,7 +4,7 @@ use eth_trie::{RootWithTrieDiff, Trie};
 use ethportal_api::{types::execution::transaction::Transaction, Header};
 use revm::inspectors::TracerEip3155;
 use std::{path::PathBuf, sync::Arc};
-use tokio::sync::Mutex;
+use tokio::sync::{oneshot::Receiver, Mutex};
 use tracing::{info, warn};
 
 use crate::{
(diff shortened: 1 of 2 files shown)


KolbyML:
> ### Future ideas
> We can speed up the shutdown by abandoning changes that are in memory (by not committing anything), but we would have the same problem that is described in #1440.
> 
> Even if implemented, it's questionable if this is desired behavior, so maybe we would need a flag that controls whether to commit on ctrl+c or not.

https://github.com/ethereum/trin/issues/1440#issuecomment-2346615788 as said in my  comment here we would want to switch to an `ACID compliment` database.

We want ctrl+c to instantly shutdown the node, as we want to prevent the Operating System from force killing trin-execution on shutdown, which could corrupt the database. Switching to an `ACID compliment` database, resolves this class of problems to my knowledge. We should be able to shutdown even if we are in the middle of writing to the database, as we could set it to only consider the writes valid when we are done them.

`ACID compliment` database should also protect the user from power outages, force kills, etc.

Executing blocks is time consuming, RocksDB sucks in that regard because you are way more likely to get your database corrupted which sucks. `LMDB` or `MXDB` solves a lot of these problems

KolbyML (approved):
:shipit:

KolbyML:
```
ACID
In computer science, ACID is a set of properties of database transactions intended to guarantee data validity despite errors, power failures, and other mishaps. In the context of databases, a sequence of database operations that satisfies the ACID properties is called a transaction.
```

This solves that whole* class of problems

PR #1454, opened 2024-09-16
refactor: add logging for investigating #1390

morph-dev:
### What was wrong?

I was checking `state` nodes and figured out that they all deadlocked (at least all that I checked ~10 of them).
They all had the same last message, related to #1390 .

I investigated the issue by looking at the code, but I couldn't figure out that issue.

### How was it fixed?

This is an attempt to narrow down the issue, by adding more logs.

I also added "(#1390)" to logs that were already added in #1391 (so we can easily track them down and remove afterwards).

### To-Do

### DIFF
--- a/portalnet/src/gossip.rs
+++ b/portalnet/src/gossip.rs
@@ -70,7 +70,7 @@ pub fn propagate_gossip_cross_thread<TContentKey: OverlayContentKey>(
         .collect();
 
     debug!(
-        "propagating validated content: found {} nodes",
+        "propagating validated content: found {} nodes (#1390)",
         all_nodes.len()
     );
     if all_nodes.is_empty() {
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
:shipit:

njgheorghita (approved):
lmk if you want me to run an early deploy with these changes, if that'll help identify the problem that a few history nodes are having

ogenev (approved):

morph-dev:
I was able to figured our the issue while debugging locally, so I will probably abandon this PR.

> lmk if you want me to run an early deploy with these changes, if that'll help identify the problem that a few history nodes are having

I think I will have a fix today for it, so early deploy once that is in would be appreciated.

morph-dev:
I think that #1458 should fix the issue, so no need for this PR.

PR #1455, opened 2024-09-16
refactor: add from bytes trait to content key

njgheorghita:
### What was wrong?
Minor refactor to content keys

### How was it fixed?
Added `From` trait to simplify construction

### To-Do

### DIFF
--- a/ethportal-api/src/types/content_key/history.rs
+++ b/ethportal-api/src/types/content_key/history.rs
@@ -87,6 +87,12 @@ pub struct BlockHeaderByHashKey {
     pub block_hash: [u8; 32],
 }
 
+impl From<[u8; 32]> for BlockHeaderByHashKey {
+    fn from(block_hash: [u8; 32]) -> Self {
+        Self { block_hash }
+    }
+}
+
 impl From<B256> for BlockHeaderByHashKey {
     fn from(block_hash: B256) -> Self {
         Self {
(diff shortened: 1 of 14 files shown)


KolbyML:
This refactor or at least the idea of it from my perception came from reviewing my PR #1436 . Well I don't think this PR is that related directly, I think https://github.com/ethereum/trin/pull/1436 should just be reviewed, I think offshoot PR's shouldn't be made until the original PR being reviewed is reviewed. This kind of seems like leap frogging in a sense, when I think the priority should be getting #1436 reviewed and merged.

njgheorghita:
Yup, i'm in the process of reviewing #1436, it's just a pretty large pr, don't mind waiting to merge this until that's through

morph-dev (approved):
While I'm not against this, I personally like more explicit approach, something like this:

```rust
impl HistoryContentKey {
    pub fn new_block_header_with_proof(block_hash: impl Into<[u8; 32]>) -> Self {
        Self::BlockHeaderWithProof(BlockHeaderKey {
            block_hash: block_hash.into(),
        })
    }

    pub fn new_block_body(block_hash: impl Into<[u8; 32]>) -> Self {
        Self::BlockBody(BlockBodyKey {
            block_hash: block_hash.into(),
        })
    }

    pub fn new_block_receipts(block_hash: impl Into<[u8; 32]>) -> Self {
        Self::BlockReceipts(BlockReceiptsKey {
            block_hash: block_hash.into(),
        })
    }

    ... // already existing functions
}
```

I'm fine with your approach as well (or both), so pick whatever you prefer.

:ship:

KolbyML (approved):
:shipit:

PR #1458, opened 2024-09-17
fix(overlay_service): don't call ping_node while holding kbuckets lock

morph-dev:
### What was wrong?

Trin would sometimes get into deadlock. See #1390.

Everything is mostly happening in the OverplayService.
There are couple of cases when we hold `kbuckets` lock, and call `ping_node` function. The `ping_node` functions needs `store` lock in order to get our node radius (to include in the request), and will block if something else already holds it.

While this could happen in many situations, we were only deadlocking in `gossip.rs` in `propagate_gossip_cross_thread` function. I don't understand why that's the case.

### How was it fixed?

Made sure that we are holding `kbuckets` lock as little as possible and that no other blocking call happens during that time.

I would consistently get into deadlock situation on my machine within 20-30 minutes. I will let it run over night in order to be confident that problem is fixed.

I'm also thinking about refactoring how we use and handle our locks (blocking vs non-blocking), but that is for future improvements.

### To-Do

### DIFF
--- a/portalnet/src/overlay/service.rs
+++ b/portalnet/src/overlay/service.rs
@@ -20,7 +20,7 @@ use discv5::{
 };
 use futures::{channel::oneshot, future::join_all, prelude::*};
 use parking_lot::RwLock;
-use rand::seq::SliceRandom;
+use rand::seq::IteratorRandom;
 use smallvec::SmallVec;
 use ssz::Encode;
 use ssz_types::BitList;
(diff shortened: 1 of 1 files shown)


njgheorghita (approved):
🥳

KolbyML (approved):
:shipit:

morph-dev:
Did more testing, and everything looks ok on my side.

morph-dev:
Also one thing that I noticed is that it seems that I was processing (successfully) **way more** requests now.

I didn't measure this in any way, just my feeling based on how quickly my db gets full and how fast logs are scrolling on my screen :laughing:

PR #1459, opened 2024-09-17
refactor: add subnetwork & network type

### LINKED ISSUE #1392: Create enum types for `portal-subnetworks` and `network` flags
The `portal-subnetworks` and `network` flags are currently defined as plain strings.

Instead they could be enums, similar to [Web3TransportType](https://github.com/ethereum/trin/blob/cc7d99e396e5e1433e1afffa835f7eda950d371e/ethportal-api/src/types/cli.rs#L32).

njgheorghita:
### What was wrong?
Fixes #1392 

### To-Do


KolbyML:
instead of `pub enum Subnetwork`, couldn't we use `pub enum ProtocolId` as it already defines all the sub-networks, then we could have 1 enum for defining the different networks instead of 2

njgheorghita:
@KolbyML We definitely could... But, it might make things more complex than simple? I guess `Subnetwork` is meant to represent active subnetworks. So this rules out 3/6 `ProtocolId` types. There's also `ProtocolId::Utp` which is a weird gray area. A lot of the simplicity gained from `Subnetwork` is lost if we have to match on `ProtocolId` and then still have checks to make sure it's an activated network...

Idk, seems to me like `ProtocolId` is a more "granular" type than what we want for simple string serde of active subnetworks

KolbyML:
> @KolbyML We definitely could... But, it might make things more complex than simple? I guess `Subnetwork` is meant to represent active subnetworks. So this rules out 3/6 `ProtocolId` types. There's also `ProtocolId::Utp` which is a weird gray area. A lot of the simplicity gained from `Subnetwork` is lost if we have to match on `ProtocolId` and then still have checks to make sure it's an activated network...
> 
> Idk, seems to me like `ProtocolId` is a more "granular" type than what we want for simple string serde of active subnetworks


```
match protocolId {
    History => xyz
    _ => not active subnetwork
}

KolbyML:
or we could define a const in the impl of ProtocolId of active networks, I am fine if Subnetworks is used for now, just wanted to point it out

njgheorghita:
Yeah, I guess that extra `_ => ..` branch in your example is what I was trying to avoid, since `Subnetwork` is always completely matched. But it does feel pretty insignificant either way...

If we take a step back... Maybe it's the name `ProtocolId` that I'm not too keen on (I mean it's not a great name if it's often being used to refer to the string representation eg. "history" as a cli flag and not the actual ID value). But that doesn't mean there's no opportunity to consolidate these.

1. What do we prefer to refer to the "state" / "history" / etc. networks. `Protocol` or `Subnetwork`? `Subprotocol`?
    - imo, I've noticed that we've tended to go with the name `Subnetwork` to refer to these in the code/specs. I don't mind switching to `Protocol` but imo that's usually more closely tied to portal's wire protocol. Maybe you feel differently?
2. Based on 1. I don't mind trying out consolidating these types into main `Protocol` / `Subnetwork` enum that implements all the string stuff and a `id()` fn.

KolbyML:
> and a `id()` fn.

I am not sure what you mean by `id()` fn.

The Enum shouldn't implement returning a protocol_id, that should be handled in NetworkSpec.

So renaming the ProtocolId enable to `SubNetwork` would be fine.

Also Sub is a word, so I think we should use `SubNetwork`, as Sub and Network are 2 different words, and camel case is every word starts with a capital etc etc

morph-dev (approved):
I wasn't sure if I was supposed to review this PR (as review was requested, but it's still in Draft state), but I reviewed it anyway.

Regarding ProtocolId vs Subnetwork, I think I'm more inclined into merging them into one type (named either "Protocol" or "Subnetwork"). But I think it's fine to do it in a separate PR (especially if we decide to keep the name "Subnetwork").

KolbyML (approved):
:shipit:

PR #1461, opened 2024-09-18
fix(trin-execution): `ProcessBeaconBlock` calculates wrong transaction/withdrawal root (fixes stuck at block 15537397)

KolbyML:
### What was wrong?

trin-execution calculated the wrong state root at block 15537397 or 3 blocks after the merge

### How was it fixed?

So earlier I checked if our input into the EVM was wrong, I noticed that the `transaction_root` I calculated was wrong I didn't think it was a big deal as `transaction_root` isn't an input to the evm.

So to debug this further I got evm traces from a production execution client.

I compared them and seen the issue was with us storing the wrong block hash for block 15537396. The reason for us storing the wrong block hash into the database was because of the invalid transaction_root which caused the block hash to be off.

Silly me :smile: 

So I fixed the way we calculated roots in `ProcessBeaconBlock` and everything started working again :partying_face:


ethportal-api/src/types/consensus/withdrawal.rs
```diff
@@ -2,6 +2,8 @@ use alloy_primitives::Address;
 use alloy_rlp::{RlpDecodable, RlpEncodable};
 use serde::{Deserialize, Deserializer, Serialize};
 
+use crate::consensus::execution_payload::Withdrawal as ConsensusWithdrawal;
```
morph-dev, line 5: nit: considering that we are in `ethportal-api/src/types/consensus/withdrawal.rs`, wouldn't this alias be a bit confusing? Maybe change it to `ExecutionPayloadWithdrawal`. Or maybe move this file into `ethportal-api/src/types/execution`. What do you think?

I'm not sure why we even have 2 in the first place. I see some difference in deserialization, but I didn't look to check if deserialization is actually the same.

trin-execution/src/era/beacon.rs
```diff
@@ -97,21 +123,37 @@ impl ProcessBeaconBlock for SignedBeaconBlockCapella {
...
+        let transactions = process_transactions(&payload.transactions)?;
+        let transactions_root = calculate_root(
+            transactions
+                .iter()
+                .map(|transaction| &transaction.transaction),
+        )?;
+
+        let withdrawals: Vec<Withdrawal> = payload
+            .withdrawals
+            .to_vec()
```
morph-dev, line 145: nit: do we have to call `.to_ver()`, can't we just call `.iter()`?

trin-execution/src/execution.rs
```diff
@@ -104,7 +104,17 @@ impl TrinExecution {
                 if stop_signal_received {
                     info!("Stop signal received. Committing now, please wait!");
                 }
-                return self.commit(&block.header, block_executor).await;
+
+                let result = self.commit(&block.header, block_executor).await?;
+
+                if stop_signal_received {
```
morph-dev, line 110: I think you can remove this check and log your message all the time (meaning, even if we reached the last_block).

ethportal-api/src/types/consensus/execution_payload.rs
morph-dev, line 1: If I'm not mistaken, we could use similar function for calculating the transaction_root of ExecutionPayload. The only difference would be that we don't encode transactions, as they are already encoded. 

And for the withdrawals, we could use you function directly, but we would have to make them RlpDecodable as well (which should be easy).

morph-dev (commented):
Good work! No major comments.

But I do think that we should move `calculate_root` function somewhere else, and document it better.

morph-dev:
> So earlier I checked if our input into the EVM was wrong, I noticed that the `transaction_root` I calculated was wrong I didn't think it was a big deal as `transaction_root` isn't an input to the evm.

Can you maybe also add a test case for the block where we didn't calculate `transaction_root` and/or `withdrawals_root` correctly before but we do now? (e.g. using example block from the mainnet)

ethportal-api/src/types/consensus/withdrawal.rs
KolbyML: > I'm not sure why we even have 2 in the first place. I see some difference in deserialization, but I didn't look to check if deserialization is actually the same.

https://github.com/ethereum/trin/pull/1410#issuecomment-2323925871

ethportal-api/src/types/consensus/execution_payload.rs
KolbyML: I don't think these functions belong here. These are just opaque containers to the CL.

ethportal-api/src/types/execution/block_body.rs
```diff
@@ -72,6 +72,25 @@ impl ssz::Decode for BlockBody {
     }
 }
 
+/// Calculate the Merkle Patricia Trie root hash from a list of items
+pub fn calculate_root<'a, T: Encodable + 'a>(
+    items: impl Iterator<Item = &'a T>,
```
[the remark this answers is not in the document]
KolbyML: I want to avoid the clone `IntoIterator`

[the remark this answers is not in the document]
KolbyML: We can reassess this in the future if `Iterator` doesn't work for all cases, but currently `Iterator` does work for all of our usecases

KolbyML:

ethportal-api/src/types/execution/block_body.rs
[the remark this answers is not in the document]
morph-dev: You don't need to clone, and it is a simple change. Here is how to do it:

This function would change to:
```rust
pub fn calculate_merkle_patricia_root<'a, T: Encodable + 'a>(
    items: impl IntoIterator<Item = &'a T>,
) -> anyhow::Result<B256> {
    let memdb = Arc::new(MemoryDB::new(true));
    let mut trie = EthTrie::new(memdb);

    // Insert items into merkle patricia trie
    for (index, tx) in items.into_iter().enumerate() {
        let path = alloy_rlp::encode(index);
        let encoded_tx = alloy_rlp::encode(tx);
        trie.insert(&path, &encoded_tx)
            .map_err(|err| anyhow!("Error inserting into merkle patricia trie: {err:?}"))?;
    }

    trie.root_hash()
        .map_err(|err| anyhow!("Error calculating merkle patricia trie root: {err:?}"))
}
```
And in other places (e.g. ethportal-api/src/types/execution/block_body.rs), you would have:
```rust
calculate_merkle_patricia_root(&self.transactions()?)
```
instead of 
```rust
calculate_merkle_patricia_root(self.transactions()?.iter())
```
In fact, both syntax would work, but I find second one cleaner.

---

Also in trin-execution/src/era/beacon.rs, you would do `calculate_merkle_patricia_root(&withdrawals)?;`

morph-dev (approved):

morph-dev:
:ship:

ethportal-api/src/types/execution/block_body.rs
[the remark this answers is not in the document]
KolbyML: Oh cool I didn't make the link that IntoIterator could be into any type of iterator very cool

PR #1462, opened 2024-09-18
chore(deps): update jsonrpsee dependency

morph-dev:
### What was wrong?

The jsonrpsee version we are using is more than a year old. Some other dependencies already updated theirs (e.g. reth-rpc-types).

I also removed direct dependency on hyper.

### How was it fixed?

Updated the dependencies and refactor the code where needed.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -59,7 +59,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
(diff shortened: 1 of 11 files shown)


KolbyML:
We should stop using `reth-rpc-types` and use `alloy-rpc-types` instead, my PR will update this

KolbyML (approved):
:shipit:

 I kind of think we should return errors in `portal-bridge/src/gossip.rs` the code is ignoring the returned error's right now so I see why `Result<>` was removed. In the future I hope we either print a log or return an error here, but that isn't related to this PR

njgheorghita (approved):
:shipit:

PR #1463, opened 2024-09-18
fix: use default angelfood bootnodes if network is selected

### LINKED ISSUE #1460: Properly handle Bootnodes for Mainnet and Angelfood networks
Currently our test network `AngelFood` doesn't have default bootnodes. It would be ideal to have each network have it's own list of bootnodes. 

So when you boot trin with `--network angelfood`, and we use the default bootnodes option it uses bootnodes for angelfood instead of mainnet.

https://github.com/ethereum/portal-network-specs/blob/master/bootnodes.md bootnodes are list here. (currently the bootnode for angelfood isn't online)

njgheorghita:
### What was wrong?
fixes #1460

### How was it fixed?
Added angelfood bootnodes and logic to handle the correct bootnodes depending on what network's selected.

### To-Do

### DIFF
--- a/ethportal-api/src/types/bootnodes.rs
+++ b/ethportal-api/src/types/bootnodes.rs
@@ -2,12 +2,30 @@ use std::str::FromStr;
 
 use anyhow::anyhow;
 
-use crate::types::enr::Enr;
+use crate::types::{enr::Enr, network::Network};
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Bootnode {
     pub enr: Enr,
     pub alias: String,
+}
+
+impl From<Enr> for Bootnode {
+    fn from(enr: Enr) -> Self {
+        let all_bootnodes = DEFAULT_BOOTNODES
+            .clone()
+            .into_iter()
+            .chain(ANGELFOOD_BOOTNODES.clone());
+        for bootnode in all_bootnodes {
+            if bootnode.enr == enr {
+                return bootnode;
+            }
+        }
+        Bootnode {
+            enr,
+            alias: "custom".to_string(),
+        }
+    }
 }
 
 lazy_static! {
@@ -83,37 +101,26 @@ lazy_static! {
 pub enum Bootnodes {
     #[default]
     Default,
-    Angelfood,
     // use explicit None here instead of Option<Bootnodes>, since default value is
     // DEFAULT_BOOTNODES
     None,
     Custom(Vec<Bootnode>),
 }
 
-impl From<Enr> for Bootnode {
-    fn from(enr: Enr) -> Self {
-        for bootnode in DEFAULT_BOOTNODES.clone().into_iter() {
-            if bootnode.enr == enr {
-                return bootnode;
+impl Bootnodes {
+    pub fn to_enrs(&self, network: Network) -> Vec<Enr> {
+        match (self, network) {
+            (Bootnodes::Default, Network::Mainnet) => {
+                DEFAULT_BOOTNODES.iter().map(|bn| bn.enr.clone()).collect()
             }
-        }
-        Bootnode {
-            enr,
-            alias: "custom".to_string(),
-        }
-    }
-}
-
-impl From<Bootnodes> for Vec<Enr> {
-    fn from(bootnodes: Bootnodes) -> Self {
-        match bootnodes {
-            Bootnodes::Default => DEFAULT_BOOTNODES.iter().map(|bn| bn.enr.clone()).collect(),
-            Bootnodes::Angelfood => ANGELFOOD_BOOTNODES
+            (Bootnodes::Default, Network::Angelfood) => ANGELFOOD_BOOTNODES
                 .iter()
                 .map(|bn| bn.enr.clone())
                 .collect(),
-            Bootnodes::None => vec![],
-            Bootnodes::Custom(bootnodes) => bootnodes.iter().map(|bn| bn.enr.clone()).collect(),
+            (Bootnodes::None, _) => vec![],
+            (Bootnodes::Custom(bootnodes), _) => {
+                bootnodes.iter().map(|bn| bn.enr.clone()).collect()
+            }
         }
     }
 }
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
:shipit: I think it is weird to be able to use `mainnet` bootnodes on `angelfood` or the inverse. But I definitely think this PR is an big improvement on what we currently have.

morph-dev:
I also don't like that there is "angelfood" as an option. Somebody can just call `--bootnodes=angelfood` and since mainnet is default network, this would not really work in practice.

In my opinion, "default" bootnodes would depend on the Network.
The way it would work is that instead of having `impl From<Bootnodes> for Vec<Enr>`, we would have:

```rust
impl Bootnodes {
    pub fn to_enrs(&self, network: Network) -> Vec<Enr> {
        let bootnodes: &Vec<Bootnode> = match (self, network) {
            (Bootnodes::Default, Network::Mainnet) => &DEFAULT_BOOTNODES,
            (Bootnodes::Default, Network::Angelfood) => &ANGELFOOD_BOOTNODES,
            (Bootnodes::None, _) => &vec![],
            (Bootnodes::Custom(bootnodes), _) => &bootnodes,
        };
        bootnodes.iter().map(|bn| bn.enr.clone()).collect()
    }
}
```

This breaks the PortalnetConfig, which doesn't have access the the network. To solve that problem, we can either:
- change `PortalnetConfig.bootnodes` to be `Vec<Enr>`, as all uses that I found are like:
    `let bootnode_enrs: Vec<Enr> = portal_config.bootnodes.into();`
- add network field to `PortalnetConfig`

I personally might have slight preference towards the first one.

If I missed something and it's not so simple, feel free to merge it as it is now.

KolbyML:
> I also don't like that there is "angelfood" as an option. Somebody can just call `--bootnodes=angelfood` and since mainnet is default network, this would not really work in practice.
> 
> In my opinion, "default" bootnodes would depend on the Network. The way it would work is that instead of having `impl From<Bootnodes> for Vec<Enr>`, we would have:
> 
> ```rust
> impl Bootnodes {
>     pub fn to_enrs(&self, network: Network) -> Vec<Enr> {
>         let bootnodes: &Vec<Bootnode> = match (self, network) {
>             (Bootnodes::Default, Network::Mainnet) => &DEFAULT_BOOTNODES,
>             (Bootnodes::Default, Network::Angelfood) => &ANGELFOOD_BOOTNODES,
>             (Bootnodes::None, _) => &vec![],
>             (Bootnodes::Custom(bootnodes), _) => &bootnodes,
>         };
>         bootnodes.iter().map(|bn| bn.enr.clone()).collect()
>     }
> }
> ```
> 
> This breaks the PortalnetConfig, which doesn't have access the the network. To solve that problem, we can either:
> 
> * change `PortalnetConfig.bootnodes` to be `Vec<Enr>`, as all uses that I found are like:
>   `let bootnode_enrs: Vec<Enr> = portal_config.bootnodes.into();`
> * add network field to `PortalnetConfig`
> 
> I personally might have slight preference towards the first one.
> 
> If I missed something and it's not so simple, feel free to merge it as it is now.

I really like this direction

njgheorghita:
@morph-dev ready for another :eyes: here

ethportal-api/src/types/bootnodes.rs
```diff
@@ -2,14 +2,32 @@ use std::str::FromStr;
...
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Bootnode {
     pub enr: Enr,
     pub alias: String,
 }
 
+impl From<Enr> for Bootnode {
+    fn from(enr: Enr) -> Self {
+        let all_bootnodes = DEFAULT_BOOTNODES
```
morph-dev, line 15: nit: I don't think you have to clone all DEFAULT_BOOTNODES and ANGELFOOD_BOOTNODES. You can iterate over references and clone only if you find match. Something like this:

```rust
Iterator::chain(DEFAULT_BOOTNODES.iter(), ANGELFOOD_BOOTNODES.iter())
    .find(|bootnode| bootnode.enr == enr)
    .cloned()
    .unwrap_or_else(|| Self {
        enr,
        alias: "custom".to_string(),
    })
```

portalnet/src/config.rs
```diff
@@ -33,7 +35,7 @@ impl Default for PortalnetConfig {
             external_addr: None,
             private_key: B256::random(),
             listen_port: 4242,
-            bootnodes: Bootnodes::default(),
+            bootnodes: Bootnodes::default().to_enrs(Network::Mainnet),
```
morph-dev, line 38: nit: I don't like that `PortalnetConfig` implements `Default` that shouldn't be used in production code (there is a chance it can be used by mistake).
Currently it's used only in tests (as far as I can tell), so I would remove this and add something like this:

```rust
impl PortalnetConfig {
    pub fn new(trin_config: &TrinConfig, private_key: B256) -> Self {
        // current function
    }

    #[cfg(test)]
    pub fn new_for_tests() -> Self {
        Self {
            // current default
        }
    }
}
```

But, it isn't really a big deal and not really part of this PR, so feel free to ignore it (more of an idea of how we can do this better in general).
Up to you.

morph-dev (approved):
LGTM :ship: 

Left two, small and optional comments. Totally up to you.

portalnet/src/config.rs
njgheorghita: Well, we do use the `Default` trait quite a bit...
- inside `PortalnetConfig::new()`
- in "test" utils that which don't seem to be scoped to the test compilation set (https://github.com/ethereum/trin/blob/master/utp-testing/src/lib.rs#L164)
- inside our unit tests 

Ultimately, the `Default` syntax is quite handy for initializing a new instance with only a few adjusted fields (https://github.com/ethereum/trin/blob/master/portalnet/src/overlay/service.rs#L2779)...

I get your concern about using it inside "production" code, so I removed it from the `PortalnetConfig::new()` constructor. 
- the only fields that were being set by the `Default` trait were...
  - `internal_ip` - which we actually never use, so I removed it
  - `node_addr_cache_capacity` - which I just updated to use the default value. Maybe this should be added as a cli arg?

And I added a comment to the `Default` trait that it should only be used inside testing (obvs. this won't be caught by the compiler)

PR #1465, opened 2024-09-19
fix: broken master ci

njgheorghita:
### What was wrong?
Broken master ci

### How was it fixed?
Updated script to use correct `Subnetwork` type

### To-Do


KolbyML (approved):
:shipit:

PR #1466, opened 2024-09-19
fix: update dependencies and add openssl vendored

njgheorghita:
### What was wrong?
Building the releases is broken (https://github.com/ethereum/trin/actions/runs/10947101080/job/30395908924). IIUC, it probably had something to do with the changes in dependencies in #1448 

I've tested these changes locally, and was able to cross compile the release binary, so I'm fairly confident that this will resolve the issue in our broken release workflow

### How was it fixed?
- updated `rustc` version in toolchain & updated out of date libraries that wouldn't compile 
- added `vendored` version of `openssl` library

### To-Do


portal-bridge/Cargo.toml
```diff
@@ -35,7 +35,7 @@ portalnet.workspace = true
 prometheus_exporter.workspace = true
 rand.workspace = true
 reqwest.workspace = true
-reqwest-middleware = { version = "0.3.3", features = ["json"] }
+reqwest-middleware = { version = "0.3.3", default-features = false, features = ["json", "rustls-tls"] }
```
njgheorghita, line 38: My suspicion was that adding a dependency on this library in #1448 was the cause, since it was added without these features that we use on the base `reqwest` dependency. But, I added them and it didn't fix the error that was caused when building the release. I'm not sure whether or not to leave these in...

Cargo.toml
```diff
@@ -100,6 +101,7 @@ jsonrpsee = "0.20.0"
 keccak-hash = "0.10.0"
 lazy_static = "1.4.0"
 light-client = { path = "light-client" }
+openssl = { version = "0.10.66", features = ["vendored"]}
```
njgheorghita, line 104: The main downside of this solution is that `openssl` is a fairly large library, but I was unable to figure out an alternative solution (see other comment) ... but I'm open to any ideas

KolbyML: the feature `native-tls-vendored` should be enabled for reqwest and `openssl = { version = "0.10.66", features = ["vendored"]}` should be removed I believe

njgheorghita: Nice, that seems to work fine locally!

KolbyML: Could you try

```
reqwest = { version = "0.12.7", default-features = false, features = ["native-tls-vendored", "rustls-tls", "json"] }
```
changing this to

```
reqwest = { version = "0.12.7", default-features = false, features = ["rustls-tls-native-roots", "json"] }
```

KolbyML (approved):
:shipit:

PR #1467, opened 2024-09-19
Continue searching for content when an earlier transfer of content fails, or is invalidated

### LINKED ISSUE #1383: Robust RecursiveFindContent | Continue the search if data transfer fails

carver:
### What was wrong?

Fix #1383 

### How was it fixed?

- send a new message back to the query when the transfer fails, or the validation fails
- resume the search for content when the new message arrives
- expand the existing test to verify that content is eventually found, even if it sometimes is invalid

### To-Do

- [x] rebase on #1395 
- [x] make sure follow-up items from #1383 go into new issue(s)

### DIFF
--- a/portalnet/src/find/iterators/findcontent.rs
+++ b/portalnet/src/find/iterators/findcontent.rs
@@ -56,7 +56,7 @@ pub enum FindContentQueryPending<TNodeId> {
         // peer that sent the content
         peer: TNodeId,
         // channel used to send back the validated content info
-        valid_content_tx: Sender<ValidatedContent<TNodeId>>,
+        valid_content_tx: Sender<Option<ValidatedContent<TNodeId>>>,
     },
     Utp {
         connection_id: u16,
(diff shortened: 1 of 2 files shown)


portalnet/src/find/iterators/findcontent.rs
```diff
@@ -56,15 +56,15 @@ pub enum FindContentQueryPending<TNodeId> {
         // peer that sent the content
         peer: TNodeId,
         // channel used to send back the validated content info
-        valid_content_tx: Sender<ValidatedContent<TNodeId>>,
+        valid_content_tx: Sender<Option<ValidatedContent<TNodeId>>>,
```
mrferris, line 59: We spoke about how to differentiate between a failed utp and failed validation within the trace, and you mentioned it would be tough to send the trace between threads. I think we could instead pass back a wrapper datatype enum here with `ValidatedContent`, `FailedUtp`, and `FailedValidation` variants instead of `None` on failure. 

Not a blocker on this PR.

mrferris (approved):

PR #1470, opened 2024-09-20
chore: extract shared logic from trin-execution::evm into trin-evm crate

morph-dev:
### What was wrong?

Because of the `eth_call` rpc, the `trin` binary depends on the entire `trin-execution` crate.

### How was it fixed?

Extracted shared logic (which was in `trin-execution/src/evm`) into separate crate: `trin-evm`.

This reduced the size of release binary by 12% (48.9 MB -> 43.3 MB) and debug binary by 32% (685 MB -> 445 MB).

### To-Do


KolbyML:
@morph-dev since you are adding a new crate can you increase `check-workspace-crates` parallelism limit by 1 as you added one new crate

trin-evm/src/dao_fork.rs
KolbyML, line 1: Why would a list of addresses for the dao fork be moved into `trin-evm`? Wouldn't only Trin-Execution need this?

KolbyML: I don't think "block beneficiaries" is related to how the EVM works, it is a change to the state outside of the EVM's context

KolbyML (approved):

KolbyML (approved):
:shipit:

trin-evm/src/dao_fork.rs
KolbyML: I would say "block beneficiaries" is related to the state, not the evm, as "block beneficiaries" is an out of evm state change

PR #1471, opened 2024-09-20
feat(trin-execution): add JSON-RPC trait's required for the Engine API

KolbyML:
### What was wrong?
We don't implement the engine api. The first step to do this is by adding the jsonrpcee trait outline.

The engine api, exposes the `engine` namespace and also a subset of the `eth` namespace, both under JWT authentication.

I made a PR first just adding the trait to make the changes easier to review, my follow up PR's will be implementing the rest of the Engine API I am not sure how many PR's it will end up being, but I will try to make them all easily review-able sizes
### How was it fixed?

Implementing the jsonrpcee trait which outlines both the `engine` and `eth` (subset) namespaces which is required to support the engine API

### DIFF
--- a/trin-execution/src/engine/rpc.rs
+++ b/trin-execution/src/engine/rpc.rs
@@ -1,0 +1,146 @@
+use alloy_rlp::Bytes;
+use alloy_rpc_types::{Block, BlockId, Filter, Log, SyncStatus, TransactionRequest};
+use alloy_rpc_types_engine::{
+    ExecutionPayloadBodiesV1, ExecutionPayloadBodiesV2, ExecutionPayloadInputV2,
+    ExecutionPayloadV1, ExecutionPayloadV2, ExecutionPayloadV3, ExecutionPayloadV4,
+    ForkchoiceState, ForkchoiceUpdated, PayloadAttributes, PayloadId, PayloadStatus,
+    TransitionConfiguration,
+};
+use jsonrpsee::{core::RpcResult, proc_macros::rpc};
+use revm_primitives::{Address, B256, U256};
+
+/// Engine Api JSON-RPC endpoints
+#[rpc(client, server, namespace = "engine")]
+pub trait EngineApi {
+    #[method(name = "exchangeCapabilities")]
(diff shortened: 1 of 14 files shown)


trin-execution/src/engine/rpc.rs
```diff
@@ -0,0 +1,146 @@
...
+    async fn new_payload_v4(
+        &self,
+        payload: ExecutionPayloadV4,
+        versioned_hashes: Vec<B256>,
+        parent_beacon_block_root: B256,
+    ) -> RpcResult<PayloadStatus>;
+}
+
+/// A subset of Eth JSON-RPC endpoints under the Engine Api's JWT authentication
+#[rpc(client, server, namespace = "eth")]
```
morph-dev, line 110: nit: While it's only a subset, does it make sense to move this into: https://github.com/ethereum/trin/blob/master/ethportal-api/src/eth.rs ?

morph-dev (approved):
LGTM :ship: 

One thing that I'm not aware of is who is going to make requests to this API?
Should these traits be in the separate crate (e.g. ethportal-api)?

I'm asking this because if I want to build a client to call it, I don't want to depend on the `trin-execution` for it. But this is something we can easily extract in the future, when we see who is going to use it and how, so no action is expected now.

KolbyML:
> I'm asking this because if I want to build a client to call it, I don't want to depend on the `trin-execution` for it. But this is something we can easily extract in the future, when we see who is going to use it and how, so no action is expected now.

You can use `alloy` to call it. Since these calls aren't related to Portal in any context I didn't think it made sense to put it in ethportal-api

trin-execution/src/engine/rpc.rs
KolbyML: > nit: While it's only a subset, does it make sense to move this into: https://github.com/ethereum/trin/blob/master/ethportal-api/src/eth.rs ?

I think both of the traits I added should be next to each other, because they will only be used together.

I think we should have both `EngineEthApi` and `EthApi` trait, because they are used in 2 different contexts

PR #1474, opened 2024-09-22
fix(docker): add trin-evm directory to docker

morph-dev:
### What was wrong?

I forgot to add `trin-evm` to docker files (should have been part of #1470)

### How was it fixed?

I added `trin-evm` and sorted folders by name.

### To-Do


KolbyML (approved):
:shipit:

PR #1475, opened 2024-09-23
feat: enable gossip dropped by default for history network

njgheorghita:
### What was wrong?
Gossiping dropped content has been running pretty successfully now on our network nodes for the past week, so this enables it by default.

### How was it fixed?
Remove cli flag and enable feature by default

### To-Do

- [ ] update deployment scripts


KolbyML (approved):
I like keeping configuration flags, but I don't think it is a big deal and we want good defaults so :shipit:

PR #1476, opened 2024-09-23
fix: unable to execute to Cancun due to bug in post_block_beneficiaries

KolbyML:
### What was wrong?
When executing 17034871 we are generating the wrong state root
### How was it fixed?
I found the issue was we had to convert the amount in the withdrawal to the amount in the denomination of WEI. We do the same thing for block rewards.

With this fix we can now execute to Cancun


njgheorghita (approved):
:shipit:

PR #1478, opened 2024-09-23
refactor: swap ProtocolId with Subnetwork type

njgheorghita:
### What was wrong?
Consolidate `ProtocolId` & `Subnetwork` types...

I chose `Subnetwork` > `ProtocolId` ( or `Protocol`) because...
1. Imo, it's a more descriptive name & generally applicable
2. Given that we use `--portal-subnetworks` as the cli flag, that seems to establish (for users) that "subnetwork" is the preferred variable name when referencing this concept

### To-Do

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -353,6 +353,12 @@ pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>
         return Err("At least one subnetwork must be enabled".to_owned());
     }
 
+    for subnetwork in &subnetworks {
+        if !subnetwork.is_active() {
+            return Err("{subnetwork} subnetwork has not yet been activated".to_owned());
+        }
+    }
+
     Ok(Arc::new(subnetworks))
 }
 
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -346,7 +346,7 @@ pub fn network_parser(network_string: &str) -> Result<Arc<NetworkSpec>, String> 
 pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>, String> {
     let subnetworks = subnetwork_string
         .split(',')
-        .map(Subnetwork::from_str)
+        .map(Subnetwork::from_cli_arg)
         .collect::<Result<Vec<Subnetwork>, String>>()?;
 
     if subnetworks.is_empty() {
--- a/ethportal-api/src/types/network.rs
+++ b/ethportal-api/src/types/network.rs
@@ -81,4 +81,17 @@ impl Subnetwork {
             Subnetwork::Utp => "utp".to_string(),
         }
     }
+
+    /// Returns true if the subnetwork has been "fully" activated.
+    pub fn is_active(&self) -> bool {
+        match self {
+            Subnetwork::Beacon => true,
+            Subnetwork::History => true,
+            Subnetwork::State => true,
+            Subnetwork::CanonicalIndices => false,
+            Subnetwork::VerkleState => false,
+            Subnetwork::TransactionGossip => false,
+            Subnetwork::Utp => false,
+        }
+    }
 }
--- a/ethportal-api/src/types/network.rs
+++ b/ethportal-api/src/types/network.rs
@@ -29,32 +29,56 @@ impl std::str::FromStr for Network {
 }
 
 /// Enum for various different portal subnetworks in a "core" network.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub enum Subnetwork {
     Beacon,
     History,
     State,
+    CanonicalIndices,
+    VerkleState,
+    TransactionGossip,
+    Utp,
 }
 
+// Pretty printed version of the subnetwork enum, used in metrics labels.
 impl fmt::Display for Subnetwork {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            Subnetwork::Beacon => write!(f, "beacon"),
-            Subnetwork::History => write!(f, "history"),
-            Subnetwork::State => write!(f, "state"),
+            Subnetwork::Beacon => write!(f, "Beacon"),
+            Subnetwork::History => write!(f, "History"),
+            Subnetwork::State => write!(f, "State"),
+            Subnetwork::CanonicalIndices => write!(f, "Canonical Indices"),
+            Subnetwork::VerkleState => write!(f, "Verkle State"),
+            Subnetwork::TransactionGossip => write!(f, "Transaction Gossip"),
+            Subnetwork::Utp => write!(f, "uTP"),
         }
     }
 }
 
-impl std::str::FromStr for Subnetwork {
-    type Err = String;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
+// Convert camel_case cli args to/from the Subnetwork enum.
+impl Subnetwork {
+    pub fn from_cli_arg(s: &str) -> Result<Self, String> {
         match s {
             "beacon" => Ok(Subnetwork::Beacon),
             "history" => Ok(Subnetwork::History),
             "state" => Ok(Subnetwork::State),
+            "canonical_indices" => Ok(Subnetwork::CanonicalIndices),
+            "verkle_state" => Ok(Subnetwork::VerkleState),
+            "transaction_gossip" => Ok(Subnetwork::TransactionGossip),
+            "utp" => Ok(Subnetwork::Utp),
             _ => Err(format!("Unknown subnetwork: {s}")),
+        }
+    }
+
+    pub fn to_cli_arg(&self) -> String {
+        match self {
+            Subnetwork::Beacon => "beacon".to_string(),
+            Subnetwork::History => "history".to_string(),
+            Subnetwork::State => "state".to_string(),
+            Subnetwork::CanonicalIndices => "canonical_indices".to_string(),
+            Subnetwork::VerkleState => "verkle_state".to_string(),
+            Subnetwork::TransactionGossip => "transaction_gossip".to_string(),
+            Subnetwork::Utp => "utp".to_string(),
         }
     }
 }
--- a/ethportal-api/src/types/portal_wire.rs
+++ b/ethportal-api/src/types/portal_wire.rs
@@ -6,6 +6,7 @@ use std::{
 };
 
 use alloy_primitives::U256;
+use anyhow::anyhow;
 use bimap::BiHashMap;
 use once_cell::sync::Lazy;
 use rlp::Encodable;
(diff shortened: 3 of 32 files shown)


ethportal-api/src/types/portal_wire.rs
```diff
@@ -157,117 +157,68 @@ pub enum DiscoveryRequestError {
...
+    pub fn get_subnetwork_from_hex(&self, hex: &str) -> Result<Subnetwork, SubnetworkError> {
+        self.portal_subnetworks
             .get_by_right(hex)
             .copied()
-            .ok_or(ProtocolIdError::Invalid)
+            .ok_or(SubnetworkError::Invalid)
     }
 
-    pub fn get_protocol_hex_from_id(
+    pub fn get_hex_from_subnetwork(
```
KolbyML, line 179:
```suggestion
    pub fn get_protocol_identifier_from_subnetwork(
```

ethportal-api/src/types/portal_wire.rs
```diff
@@ -157,117 +157,68 @@ pub enum DiscoveryRequestError {
...
 }
 
 impl NetworkSpec {
     pub fn network(&self) -> Network {
         self.network
     }
 
-    pub fn get_protocol_id_from_hex(&self, hex: &str) -> Result<ProtocolId, ProtocolIdError> {
-        self.portal_networks
+    pub fn get_subnetwork_from_hex(&self, hex: &str) -> Result<Subnetwork, SubnetworkError> {
```
KolbyML, line 172:
```suggestion
    pub fn get_subnetwork_from_protocol_identifier(&self, hex: &str) -> Result<Subnetwork, SubnetworkError> {
```

ethportal-api/src/types/network.rs
```diff
@@ -29,32 +33,109 @@ impl std::str::FromStr for Network {
...
     Beacon,
     History,
     State,
+    CanonicalIndices,
+    VerkleState,
+    TransactionGossip,
+    Utp,
 }
 
 impl fmt::Display for Subnetwork {
```
njgheorghita, line 47: It's a little awkward here for these str conversion traits...

For `Protocolid`, we used `Display` & `FromStr` to convert cli args (aka camel case), so I implemented some helper functions to be used in these scenarios.

Ok, we could use the `Debug` trait as the "pretty" version, (the only difference being the extra space in multi-letter variants). And then use `Display`/`FromStr` to handle cli args. But imo `Display` is meant to be the "pretty" version of the string.

But, we also use `Display` to create the "metrics" label. And, if we change that, then there's a whole lot of following changes that need to be made to our dashboard

ethportal-api/src/types/network.rs
```diff
@@ -29,32 +33,109 @@ impl std::str::FromStr for Network {
...
+        let expected_u8 = 3;
+        assert_eq!(expected_u8, u8::from(protocol_id));
+
+        let protocol_id = Subnetwork::CanonicalIndices;
+        let expected_u8 = 4;
+        assert_eq!(expected_u8, u8::from(protocol_id));
+
+        let protocol_id = Subnetwork::Utp;
+        let expected_u8 = 99;
+        assert_eq!(expected_u8, u8::from(protocol_id));
```
KolbyML, line 139: These aren't protocol_identifier's. We are just mimicking an enum as an identifier for the database with this code no?

trin-storage/src/config.rs
```diff
@@ -63,6 +63,8 @@ impl PortalStorageConfigFactory {
             Subnetwork::History => Self::HISTORY_CAPACITY_WEIGHT,
             Subnetwork::State => Self::STATE_CAPACITY_WEIGHT,
             Subnetwork::Beacon => Self::BEACON_CAPACITY_WEIGHT,
+            _ => panic!("xxx"),
+            //_ => unreachable!("Unknown subnetwork: {:?}", subnetwork);
```
KolbyML, line 67: I think unreachable should still be used here I assume this is just for testing

KolbyML (commented):
Just some early feedback feel free to ignore it if you were already planning to do it, or it is premature or anything etc.

portalnet/src/overlay/protocol.rs
```diff
@@ -70,7 +71,7 @@ pub struct OverlayProtocol<TContentKey, TMetric, TValidator, TStore> {
     /// The overlay routing table of the local node.
     kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
     /// The subnetwork protocol of the overlay.
-    protocol: ProtocolId,
+    protocol: Subnetwork,
```
njgheorghita, line 74: I left the original `protocol` var alone here since it's used extensively in our logs & comments. At some point, these two phrases are interchangeable, so i'm not sure it's worth updating all the `protocol` usage in our comments to `subnetwork`

portalnet/src/overlay/service.rs
```diff
@@ -112,7 +113,7 @@ where
     /// The routing table of the local node.
     kbuckets: Arc<RwLock<KBucketsTable<NodeId, Node>>>,
     /// The protocol identifier.
-    protocol: ProtocolId,
+    protocol: Subnetwork,
```
njgheorghita, line 116: I left the original `protocol` var alone here since it's used extensively in our logs & comments. At some point, these two phrases are interchangeable, so i'm not sure it's worth updating all the `protocol` usage in our comments to `subnetwork`

KolbyML: I think it is fine to leave this as you did

ethportal-api/src/types/portal_wire.rs
KolbyML: I am just confused what the hex is, which is why I mention it. I think by saying `protocol_indentifier` instead of hex we more clearly link this to what it is in the spec

ethportal-api/src/types/network.rs
```diff
@@ -29,32 +33,109 @@ impl std::str::FromStr for Network {
...
+            Subnetwork::Beacon => 1,
+            Subnetwork::History => 0,
+            Subnetwork::State => 2,
+            Subnetwork::CanonicalIndices => 4,
+            Subnetwork::VerkleState => 5,
+            Subnetwork::TransactionGossip => 3,
+            Subnetwork::Utp => 99,
+        }
+    }
+}
```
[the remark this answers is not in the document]
KolbyML: To my knowledge this conversion `impl From<Subnetwork> for u8 {` is only used for the database

ethportal-api/src/types/portal_wire.rs
KolbyML: or maybe `get_subnetwork_hex_identifier` would be better

ethportal-api/src/types/network.rs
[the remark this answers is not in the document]
KolbyML: Maybe Milos db refactor removed the need for `impl From<Subnetwork> for u8 {`, probably something we should double check nontheless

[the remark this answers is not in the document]
njgheorghita: I can't find anywhere we actually use it, so I've just removed it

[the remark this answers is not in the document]
KolbyML: Sounds good, I assume it's usage has been removed then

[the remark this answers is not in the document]
KolbyML: One thing to note is Enum's are just `variable` to `number` links. Where the u8 value would be the order of the variables in the Enum

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum Subnetwork {
    Beacon = 0,
    History = 1,
    State = 2,
    CanonicalIndices = 3,
    VerkleState = 4,
    TransactionGossip = 5,
    Utp = 6,
}

```

So to my knowledge this would be implicit ^

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
pub enum Subnetwork {
    Beacon,
    History,
    State,
    CanonicalIndices,
    VerkleState
    TransactionGossip,
    Utp,
}

```

Both of versions are equivalent, one just has implicit indexing and the other it is declared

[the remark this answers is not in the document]
KolbyML: Yeah I think we are good

ethportal-api/src/types/cli.rs
```diff
@@ -346,7 +346,7 @@ pub fn network_parser(network_string: &str) -> Result<Arc<NetworkSpec>, String>
 pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>, String> {
     let subnetworks = subnetwork_string
         .split(',')
-        .map(Subnetwork::from_str)
+        .map(Subnetwork::from_cli_arg)
```
KolbyML, line 349: Maybe add in `pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>, String> {` a check if all the submitted `subnetwork`'s are active/supported or not. And through an error if the network isn't support yet

ethportal-api/src/types/network.rs
morph-dev: I like current approach.

I just wanted to mention that we also have [alternate formatting](https://doc.rust-lang.org/std/fmt/#sign0) at our disposal. I don't think it really fits for this use case, but you see a nice way to use it.

trin-metrics/src/labels.rs
```diff
@@ -118,16 +121,16 @@ pub enum MessageLabel {
...
-    fn from(protocol: &ProtocolId) -> Self {
-        match protocol {
-            ProtocolId::State => Self::State,
-            ProtocolId::VerkleState => Self::VerkleState,
-            ProtocolId::History => Self::History,
-            ProtocolId::TransactionGossip => Self::TransactionGossip,
-            ProtocolId::CanonicalIndices => Self::CanonicalIndices,
-            ProtocolId::Beacon => Self::Beacon,
-            ProtocolId::Utp => Self::Utp,
+impl From<&Subnetwork> for ProtocolLabel {
```
morph-dev, line 124: nit: Looking at `ProtocolLabel`, I think it's not used and we can remove it. Even if it is used, I don't see why we can't use `Subnetwork` instead.

morph-dev (approved):

ethportal-api/src/types/cli.rs
morph-dev: +1 to this

ogenev (approved):
⛵

ethportal-api/src/types/network.rs
```diff
@@ -29,32 +29,69 @@ impl std::str::FromStr for Network {
...
+    /// Returns true if the subnetwork has been "fully" activated.
+    pub fn is_active(&self) -> bool {
+        match self {
+            Subnetwork::Beacon => true,
+            Subnetwork::History => true,
+            Subnetwork::State => true,
+            Subnetwork::CanonicalIndices => false,
+            Subnetwork::VerkleState => false,
+            Subnetwork::TransactionGossip => false,
+            Subnetwork::Utp => true,
```
KolbyML, line 94:
```suggestion
            Subnetwork::Utp => false,
```

I think uTP should be false as it isn't a network you can "activate"

KolbyML (approved):
:shipit: looks good

PR #1479, opened 2024-09-23
Add peer failures to query traces

carver:
### What was wrong?

There is no way to acknowledge and track whether a peer experiences a variety of failures during a query, and how long that took.

### How was it fixed?

Add a new tracking HashMap in the QueryTrace object, to track utp and validation errors.

Still need to actually use the new QueryTrace API in the overlay service.

### To-Do

- [x] In overlay service, track failures of utp connection, utp download, and content validation

### DIFF
--- a/ethportal-api/src/types/query_trace.rs
+++ b/ethportal-api/src/types/query_trace.rs
@@ -21,6 +21,8 @@ pub struct QueryTrace {
     pub origin: NodeId,
     /// Map of a node's ID to its corresponding `QueryResponse`
     pub responses: HashMap<NodeId, QueryResponse>,
+    /// Track if and how each node fails during a query
+    pub failures: HashMap<NodeId, QueryFailure>,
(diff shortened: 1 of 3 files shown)


ethportal-api/src/types/query_trace.rs
```diff
@@ -21,6 +21,8 @@ pub struct QueryTrace {
     pub origin: NodeId,
     /// Map of a node's ID to its corresponding `QueryResponse`
     pub responses: HashMap<NodeId, QueryResponse>,
+    /// Track if and how each node fails during a query
+    pub failures: HashMap<NodeId, QueryFailure>,
```
carver, line 25: @mrferris what do you think about this structure, for tracking failures types and times, for each peer?

(PR is unfinished, I just remember you saying you had opinions on this data structure, and figured I'd get early feedback)

mrferris (approved):

PR #1481, opened 2024-09-24
fix(portal-bridge): re-initialize census if no peers are available

njgheorghita:
### To-Do


portal-bridge/src/census.rs
```diff
@@ -52,11 +59,15 @@ impl Census {
 }
 
 impl Census {
-    pub async fn init(&mut self) {
+    pub async fn init(&mut self) -> anyhow::Result<()> {
```
morph-dev, line 62: nit: Should this also be `Result<(), CensusError>`?

portal-bridge/src/main.rs
```diff
@@ -56,7 +56,9 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         let (census_tx, census_rx) = mpsc::unbounded_channel();
         let mut census = Census::new(portal_client.clone(), census_rx);
         // initialize the census to acquire critical threshold view of network before gossiping
-        census.init().await;
+        if census.init().await.is_err() {
```
morph-dev, line 59: nit: In my opinion, main can just return `anyhow::Result<()>` and you can use `anyhow!(..)` here. Or just use `census.init().await?;` and propagate error.

morph-dev (approved):
:shipit:

KolbyML (approved):
:shipit:

PR #1487, opened 2024-09-25
chore: delete unused trin-history/src/utils.rs

morph-dev:
### What was wrong?

Functionality in trin-history/src/utils.rs is not used at all.

### How was it fixed?

I removed it.


njgheorghita (approved):

carver:
Huh, interesting. It looks like we stopped expanding the shortened client names in https://github.com/ethereum/trin/pull/875, when we stopped using `bucket_entries_to_json`, but that PR description makes it look to me like it was an accident.

Expanding the client name in json responses is probably something we want to do again.

PR #1488, opened 2024-09-25
refactor: use NodeId instead of String for discv5 json types

morph-dev:
### What was wrong?

We are using String to represent NodeId in json types.
I believe this isn't necessary as NodeId serialization is the same as what we are using (hex string prefixed with "0x").

This is not a big deal, but something I encounter while refactoring kbuckets.

### How was it fixed?

Changed String to NodeId as type.

### To-Do

### DIFF
--- a/ethportal-api/src/types/discv5.rs
+++ b/ethportal-api/src/types/discv5.rs
@@ -1,5 +1,4 @@
 use super::enr::Enr;
-use crate::utils::bytes::hex_encode;
 use serde::{Deserialize, Serialize};
 
 use discv5::enr::NodeId;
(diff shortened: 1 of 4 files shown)


ethportal-api/src/types/discv5.rs
```diff
@@ -24,15 +23,15 @@ pub struct KBucketsTable {
 #[serde(rename_all = "camelCase")]
 pub struct NodeInfo {
     pub enr: Enr,
-    pub node_id: String,
+    pub node_id: NodeId,
```
njgheorghita, line 26: Might be worth adding a comment here that `NodeId` gets properly serialized as a 0x-prefixed hexstring

njgheorghita (approved):

ethportal-api/src/types/discv5.rs
morph-dev: I don't think that's necessary. Since `Serialize/Deserialize` is derived, one should assume that serialization is done correctly.

But if you feel strong about it, I can add it.

PR #1489, opened 2024-09-25
Add detailed info to offer trace

njgheorghita:
### What was wrong?
Updated `trace_offer` to return more detail about what happened during the `Offer`, specifically to identify between cases where the offered key was rejected, or the transfer failed, which is helpful to know inside the state bridge.

### How was it fixed?
- added `OfferTrace` type to capture more information about failed `trace_offer` requests

### To-Do

### DIFF
--- a/ethportal-api/src/types/portal_wire.rs
+++ b/ethportal-api/src/types/portal_wire.rs
@@ -550,7 +550,17 @@ pub struct PopulatedOfferWithResult {
     /// The offered content key & value
     pub content_item: (RawContentKey, Vec<u8>),
     /// The channel to send the result of the offer to
-    pub result_tx: tokio::sync::mpsc::UnboundedSender<bool>,
+    pub result_tx: tokio::sync::mpsc::UnboundedSender<OfferTrace>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum OfferTrace {
+    /// Offer was successful, all accepted content keys in bitlist were transferred
+    Success(BitList<typenum::U64>),
+    /// Peer is not interested in any of the offered content keys
+    Declined,
+    /// This offer failed, perhaps locally or from a timeout or transfer failure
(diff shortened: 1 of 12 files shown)


njgheorghita:
Blocked by #1485, since that's updating the `content_keys` type used here, but this is ready for review, and I'll make the adjustments once that gets merged

KolbyML (approved):
:shipit:

carver:
> Blocked by #1485, since that's updating the `content_keys` type used here, but this is ready for review, and I'll make the adjustments once that gets merged

merged

ethportal-api/src/types/portal_wire.rs
```diff
@@ -550,7 +550,14 @@ pub struct PopulatedOfferWithResult {
...
     /// The channel to send the result of the offer to
-    pub result_tx: tokio::sync::mpsc::UnboundedSender<bool>,
+    pub result_tx: tokio::sync::mpsc::UnboundedSender<OfferTrace>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum OfferTrace {
+    Success(BitList<typenum::U8>),
+    Rejected,
+    Failed,
```
carver, line 560: "Rejected" sounds to me a bit like the peer thinks we offered something in an invalid way. To me "Declined" sounds more like they just don't want the data we offered.

Also, a quick docstring would be nice. A draft for you:
```suggestion
    /// Peer is not interested in any of the offered data
    Declined,
    /// This offer failed, perhaps locally or from a timeout or transfer failure
    Failed,
```

carver (approved):
LGTM, 🚢 after the master rebase

ethportal-api/src/types/portal_wire.rs
njgheorghita: Nice, gonna start applying this mindset to the women who don't want to go out on a date with me...

PR #1491, opened 2024-09-26
feat(beacon): validate `HistoricalSummariesWithProof` if finalized state root is available

ogenev:
### What was wrong?
When the beacon light client is not synced, we can't process a `HistoricalSummariesWithProof` content to a `FindContent` request. 

### How was it fixed?
Validate `HistoricalSummariesWithProof` against the state root if the light client finalized root is available.

### To-Do

### DIFF
--- a/trin-beacon/src/validation.rs
+++ b/trin-beacon/src/validation.rs
@@ -26,6 +26,7 @@ use light_client::{
 use ssz::Decode;
 use std::sync::Arc;
 use tokio::sync::RwLock;
+use tracing::debug;
 use tree_hash::TreeHash;
 use trin_validation::{
     merkle::proof::verify_merkle_proof,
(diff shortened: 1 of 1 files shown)


trin-beacon/src/validation.rs
```diff
@@ -262,13 +262,15 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
                     .get_finalized_state_root()
-                    .await?;
+                    .await;
 
-                Self::state_summaries_validation(
-                    fork_versioned_historical_summaries,
-                    latest_finalized_root,
-                )
-                .await?;
+                if let Ok(latest_finalized_root) = latest_finalized_root {
```
njgheorghita, line 267: Is it worth logging a `warn!` here in the case that the `latest_finalized_root` is not available?

njgheorghita (approved):

KolbyML (approved):
:shipit:

trin-beacon/src/validation.rs
ogenev: Added a `debug!` log because this is expected to happen if the light client is not synced.

PR #1492, opened 2024-09-26
docs: add note about state bridges to deployment docs

morph-dev:
### What was wrong?

Deployment docs don't include information that state bridges should most likely not be updated, and it doesn't provide instructions how to deploy other state network nodes.

### How was it fixed?

Updated docs.

### To-Do


KolbyML (approved):
:shipit:

njgheorghita (approved):

carver (approved):

PR #1494, opened 2024-09-28
refactor: extract kbuckets into separate structure

morph-dev:
### What was wrong?

The `kbuckets` are widely used and they are wrapped in `Arc<RwLock<>>` for asynchronous access.

This can be dangerous combination as it's easy to hold the `kbuckets` lock and try to acquire some other lock or do `.await`. This was the root problem of the #1390 .

### How was it fixed?

I wrapped `Arc<RwLock<KBucketsTable<NodeId, Node>>>` in a structure (that I named `SharedKBucketsTable`).
Every time lock is acquired, it's held only for the minimum amount of time, during one function call.

This approached is recommended in https://draft.ryhl.io/blog/shared-mutable-state/ , which is referenced from: https://tokio.rs/tokio/tutorial/shared-state#restructure-your-code-to-not-hold-the-lock-across-an-await

I tried to minimize functionality changes. Here are highlights:
- `interested_enrs` and `batch_interested_enrs` are just moved from `gossip.rs`
- `insert_or_update_discovered_nodes` - re-implementation of previous `OverlayService::process_discovered_enrs`
    - original implementation was handling a case that is not possible so I removed it (for reference, inserted nodes can't be inserted as pending because they are inserted as disconnected)
- `nodes_by_distance` - original implementation wasn't working correctly
    - it would call `KBucketsTable::nodes_by_distances` but it would potentially later filter entries, resulting in returning less results than desired

Note: While this PR is big, I would say that at least 2/3 are test related changes. If desired, I can try to split it up into smaller PRs for easier review.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4341,6 +4341,7 @@ dependencies = [
  "smallvec",
  "ssz_types",
  "stunclient",
+ "tempfile",
  "test-log",
  "thiserror",
  "tokio",
--- a/portalnet/Cargo.toml
+++ b/portalnet/Cargo.toml
@@ -53,6 +53,7 @@ env_logger.workspace = true
 quickcheck.workspace = true
 rstest.workspace = true
 serial_test.workspace = true
+tempfile.workspace = true
 test-log.workspace = true
 tokio-test.workspace = true
 tracing-subscriber.workspace = true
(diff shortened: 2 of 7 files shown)


KolbyML (approved):
:shipit: looks good I didn't see anything wrong when reviewing. I also ran the PR through hive and all tests pass there as well.

PR #1495, opened 2024-09-28
feat: add "last-block" command line flag to trin execution

morph-dev:
### What was wrong?

It's not possible to configure trin-execution to stop early (it will execute all blocks up to Cancun fork or until stopped).

### How was it fixed?

Added optional command line flag to support early stop.

### To-Do


KolbyML (approved):
:shipit:

PR #1497, opened 2024-09-29
feat(state-bridge): add support for gossiping blocks from a range

morph-dev:
### What was wrong?

State Bridge only supports gossiping blocks starting from 0.
We know that content from first ~500'000 blocks are saturated, and we might want to skip them to speed up the process.

### How was it fixed?

Add support for range gossip.
We will still execute blocks starting from zero, but we will gossip only blocks from the range.

Specifying single block would imply gossiping all blocks starting from zero and ending with provided block.
This was left unmodified for backward compatibility, until we upgrade deployment scripts.
In my opinion, single block should gossip entire state trie at the specified block (but this requires more work).

### To-Do

### DIFF
--- a/portal-bridge/README.md
+++ b/portal-bridge/README.md
@@ -1,4 +1,5 @@
 # Portal-Bridge
+
 Process to feed the portal network by gossiping data retrieved from a trusted provider. This is only compatible with `Trin` clients.
 
 ex.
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
You should add an example to the readme for the bridge other then that the PR looks good

PR #1498, opened 2024-09-30
chore: automatic cargo update + manual alloy-* and revm-* update

morph-dev:
### What was wrong?

Automatic `cargo update` failed (see #1496 and #1508).

### How was it fixed?

- manually updated "alloy-*" and "revm-*" dependencies
- updated dependencies in `trin-execution/src/storage/evm_db.rs`
    - we now use `alloy_primitives::map::FbHashMap` which is optimized when keys are fixed byte arrays
- removed dependency on `alloy` and added dependency only on crates that we use (`alloy` just exports other crates)
    -  alternatively, we can use `alloy` everywhere and only enable features that we need, where we need them

Note: first commit is automated `cargo update` (#1508), second one is my fix.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,31 +108,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5c6cb57a04249c6480766f7f7cef5467412af1490f8d1e243141daddada3264f"
 
 [[package]]
-name = "alloy"
-version = "0.3.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8367891bf380210abb0d6aa30c5f85a9080cb4a066c4d5c5acadad630823751b"
-dependencies = [
- "alloy-consensus",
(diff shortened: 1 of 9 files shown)


morph-dev:
It seems that new `alloy-primitives` updated the way `Address` is serialized.
The new string representation uses checksum string, which makes some of our tests fail (they expect lowercase).

pipermerriam:
In our old python utilities we had a `is_same_address` function that was able to take the addresses in all of their possible formats (binary, hex, 0x-hex, checksum) and tell us if they were the same to avoid this kind of headache....

morph-dev:
When I updated dependencies again (today), issue that I reported last week with checksum was no longer there (but there were more deps to update now).
I think there was a process of updating many different crates and last week was "worl in the progress" any some things weren't working as expected. But now everything seems fine.

njgheorghita (approved):

PR #1499, opened 2024-09-30
feat(bridge): increase total request timeout value via CLI flag

ogenev:
### What was wrong?
The total request timeout for the bridge was set to 20 seconds. However, getting beacon state data from the Pandaops nodes usually takes more time, which causes the transfer to fail with a `Timeout`  error.

### How was it fixed?
- Add a CLI flag to set a custom total request timeout value. The default is the current value (20 sec).

### To-Do


njgheorghita:
I'm not sure it's worth increasing the timeout for all non-beacon related requests simply because it's necessary for beacon requests. Any thoughts about updating this constant to be a cli flag?

ogenev:
> I'm not sure it's worth increasing the timeout for all non-beacon related requests simply because it's necessary for beacon requests. Any thoughts about updating this constant to be a cli flag?

@njgheorghita ready for review!

KolbyML (approved):
:shipit:

PR #1500, opened 2024-10-01
feat(trin-execution): add support for cancun (we can now execute to latest*)

KolbyML:
### What was wrong?

We didn't support cancun/deneb this blocked us from executing to latest.

### How was it fixed?

by enabling c-kzg which is required to execute cancun blocks.

It was also fixed by adding support for cancun.

### To-Do in future PR's

This PR doesn't allow us to keep up with the chain that would require
- engine API (communication between EL/CL the CL drives EL execution)
- blockchain tree, basically a structure which handles re-orgs and such for latest blocks.

I am working on this now my focus will being able to bridge latest state first aka portal priorities before I even attempt to implement other engine API functionality required for block building and such.

### DIFF
--- a/trin-execution/src/evm/pre_block_contracts.rs
+++ b/trin-execution/src/evm/pre_block_contracts.rs
@@ -1,4 +1,4 @@
-use alloy_eips::eip4788::{self, BEACON_ROOTS_ADDRESS};
+use alloy_eips::eip4788;
 use anyhow::anyhow;
 use ethportal_api::Header;
 use revm::{db::State, DatabaseCommit, Evm};
@@ -26,7 +26,7 @@ fn apply_beacon_root_contract(
     // Update transaction environment to call the beacon roots contract
     evm.context.evm.env.tx = TxEnv {
         caller: eip4788::SYSTEM_ADDRESS,
-        transact_to: TxKind::Call(BEACON_ROOTS_ADDRESS),
+        transact_to: TxKind::Call(eip4788::BEACON_ROOTS_ADDRESS),
         data: parent_beacon_block_root.0.into(),
         ..Default::default()
     };
@@ -38,8 +38,9 @@ fn apply_beacon_root_contract(
         Err(err) => return Err(anyhow!("Failed to call beacon roots contract: {err:?}")),
     };
 
-    state.remove(&eip4788::SYSTEM_ADDRESS);
-    state.remove(&evm.block().coinbase);
+    // Remove all the keys except the beacon roots address which is the only state change we want to
+    // keep
+    state.retain(|address, _| address == &eip4788::BEACON_ROOTS_ADDRESS);
 
     evm.context.evm.db.commit(state);
 


trin-execution/src/evm/pre_block_contracts.rs
```diff
@@ -0,0 +1,59 @@
...
+    };
+    evm.context.evm.env.block.gas_limit = U256::from(evm.context.evm.env.tx.gas_limit);
+    evm.context.evm.env.block.basefee = U256::ZERO;
+
+    let mut state = match evm.transact() {
+        Ok(result) => result.state,
+        Err(err) => return Err(anyhow!("Failed to call beacon roots contract: {err:?}")),
+    };
+
+    state.remove(&eip4788::SYSTEM_ADDRESS);
```
morph-dev, line 41: I assume we remove these two because of the gas usage, right? Maybe add a comment about it.

Alternatively, I assume that `BEACON_ROOTS_ADDRESS` is the only remaining updated account. Maybe remove all except that one.

trin-execution/src/evm/pre_block_contracts.rs
```diff
@@ -0,0 +1,59 @@
...
+        return Err(anyhow!("Parent beacon block root is missing"));
+    };
+
+    // Save the previous environment
+    let previous_env = Box::new(evm.context.evm.env().clone());
+
+    // Update transaction environment to call the beacon roots contract
+    evm.context.evm.env.tx = TxEnv {
+        caller: eip4788::SYSTEM_ADDRESS,
+        transact_to: TxKind::Call(BEACON_ROOTS_ADDRESS),
```
morph-dev, line 29:
```suggestion
        transact_to: TxKind::Call(eip4788::BEACON_ROOTS_ADDRESS),
```

nit: for consistency with previous line

morph-dev (approved):
:shipit:

trin-execution/src/main.rs
```diff
@@ -11,6 +9,7 @@ use trin_execution::{
 use trin_utils::{dir::setup_data_dir, log::init_tracing_logger};
 
 const APP_NAME: &str = "trin-execution";
+const LATEST_BLOCK: u64 = 20_868_946;
```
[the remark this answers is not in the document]
KolbyML: The latest block is a rolling window, soon we shouldn't need a static variable like this as the CL should tell us the latest block. If we wanted to find the last block contained in era files we wouldn't use a constant, but we won't be using constants soon at all for block height

trin-execution/src/evm/pre_block_contracts.rs
KolbyML: > I assume we remove these two because of the gas usage, right? Maybe add a comment about it.

Mostly because those accounts are touched, and we don't want to  commit that to the trie

> Alternatively, I assume that `BEACON_ROOTS_ADDRESS` is the only remaining updated account. Maybe remove all except that one.

sure

PR #1502, opened 2024-10-01
chore(book): add a note about resolving `docker_compose_v2` error in ansible

ogenev:
### What was wrong?
I dealt with `docker_compose_v2` error on my Linux and Mac machines when running ansible for Glados.

### How was it fixed?
Add a step in the book with a command to fix the issue.

### To-Do


carver (approved):
It's totally unfamiliar to me, so kind of a rubber stamp. I like docs 👍🏻

PR #1503, opened 2024-10-01
refactor(census): split census into multiple files and remove "Subnetwork"

morph-dev:
### What was wrong?

Currently all census logic is in one file. This makes any future changes harder to review (as more files will be created and history might not be preserved).

### How was it fixed?

This PR has 3 small commits:

- move `Network` and `Subnetwork` into separate file
- remove `Subnetwork` type and use similar one from `ethportal-api` crate
- select peers randomly
    - I believe that current implementation always selects the same peers (at least until `peers` collection is updated, which isn't frequent)

### To-Do


KolbyML (approved):
:shipit:

PR #1505, opened 2024-10-02
refactor(trin-execution): state import/export don't need TrinExecution

morph-dev:
### What was wrong?

The import/export state in `trin-execution` don't need TrinExecution. They only need `EvmDb` and `ExecutionPosition`.

No tests that test import/export end to end.

### How was it fixed?

Refactor how Importer/Exporter are initialized.

Also added test that:
- executes fist 1000 blocks
- exports state to era2 file
- imports state from era2 file (into new directory)
- executes another 1000 blocks

### To-Do

### DIFF
--- a/trin-execution/src/cli.rs
+++ b/trin-execution/src/cli.rs
@@ -4,6 +4,8 @@ use clap::{Args, Parser, Subcommand};
 
 use crate::types::block_to_trace::BlockToTrace;
 
+pub const APP_NAME: &str = "trin-execution";
+
 #[derive(Parser, Debug, Clone)]
 #[command(name = "Trin Execution", about = "Executing blocks with no devp2p")]
 pub struct TrinExecutionConfig {
@@ -54,12 +56,16 @@ pub enum TrinExecutionSubCommands {
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ImportStateConfig {
+    #[arg(long, help = "The directory where to import the snapshot data.")]
+    pub data_dir: Option<PathBuf>,
     #[arg(long, help = "path to where the era2 state snapshot is located")]
     pub path_to_era2: PathBuf,
 }
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ExportStateConfig {
+    #[arg(long, help = "The directory from where to export the snapshot data.")]
+    pub data_dir: Option<PathBuf>,
     #[arg(long, help = "path to where the era2 state snapshot is located")]
     pub path_to_era2: PathBuf,
 }
--- a/trin-execution/src/cli.rs
+++ b/trin-execution/src/cli.rs
@@ -56,16 +56,12 @@ pub enum TrinExecutionSubCommands {
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ImportStateConfig {
-    #[arg(long, help = "The directory where to import the snapshot data.")]
-    pub data_dir: Option<PathBuf>,
     #[arg(long, help = "path to where the era2 state snapshot is located")]
     pub path_to_era2: PathBuf,
 }
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ExportStateConfig {
-    #[arg(long, help = "The directory from where to export the snapshot data.")]
-    pub data_dir: Option<PathBuf>,
     #[arg(long, help = "path to where the era2 state snapshot is located")]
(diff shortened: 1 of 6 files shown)


trin-execution/src/cli.rs
```diff
@@ -54,12 +56,16 @@ pub enum TrinExecutionSubCommands {
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ImportStateConfig {
+    #[arg(long, help = "The directory where to import the snapshot data.")]
+    pub data_dir: Option<PathBuf>,
```
KolbyML, line 60: Shouldn't we just use the `data_dir` from `TrinExecutionConfig`. As with this the user can input 2 different data_dirs potentially and the CLI won't complain, but then the user will be confused. So I think we should use the one in `TrinExecutionConfig`

trin-execution/src/cli.rs
```diff
@@ -54,12 +56,16 @@ pub enum TrinExecutionSubCommands {
...
     #[arg(long, help = "path to where the era2 state snapshot is located")]
     pub path_to_era2: PathBuf,
 }
 
 #[derive(Args, Debug, Default, Clone, PartialEq)]
 pub struct ExportStateConfig {
+    #[arg(long, help = "The directory from where to export the snapshot data.")]
+    pub data_dir: Option<PathBuf>,
```
KolbyML, line 68: same comment about this, I think the ``TrinExecutionConfig` should be used I don't think we should add a param here as we already have one

KolbyML (commented):
Most of the PR looks good, but I don't think the subcommand's should have a param `data_dir`, the param from TrinExecutionConfig should be used.

`./trin --data-dir="/" import-state --data-dir="/hi"`
^ this is a valid commandline in this PR

I think this is very misleading to the user. 

`./trin --data-dir="/" import-state`
^ I think we should have this

KolbyML:
Also I think the first commit is a refactor not a feat as no functionality is changed

KolbyML:
Also to note `Data dir import export` is already  supported before this PR using 

`./trin --data-dir="/" import-state` This matches how other execution clients like `go-ethereum` do it as well for there era1 import/export

morph-dev:
You are right. I was trying:

```
./target/release/trin-execution export-state --data-dir="/my_dir" --path-to-era2="/era_dir"
```

and that didn't work. And I interpreted the error that I was getting as that none of the other cli arguments are available if subcommand is selected.
But you are right. It works if you specify the argument before subcommand.

If I knew this, I probably wouldn't even do this PR. But I think it's nice improvement (main is a bit cleaner) and we have a test.

KolbyML:
> You are right. I was trying:
> 
> ```
> ./target/release/trin-execution export-state --data-dir="/my_dir" --path-to-era2="/era_dir"
> ```
> 
> and that didn't work. And I interpreted the error that I was getting as that none of the other cli arguments are available if subcommand is selected. But you are right. It works if you specify the argument before subcommand.
> 
> If I knew this, I probably wouldn't even do this PR. But I think it's nice improvement (main is a bit cleaner) and we have a test.

Yeah, I guess we should add a few examples in the readme to document this. You can do this in this PR, or I can make a PR for this after this is merged, or something doesn't matter. But it would be nice to prevent confusion surrounding this in the future

trin-execution/src/main.rs
```diff
@@ -17,45 +15,44 @@ async fn main() -> anyhow::Result<()> {
...
+                );
                 return Ok(());
             }
         }
     }
 
+    // Initialize prometheus metrics
+    if let Some(addr) = trin_execution_config.enable_metrics_with_url {
+        prometheus_exporter::start(addr)?;
+    }
```
KolbyML, line 51: This should be before the subcommand code so we get logs

KolbyML (approved):
:shipit: everything looks good. Don't forget to 

```rust
    // Initialize prometheus metrics
    if let Some(addr) = trin_execution_config.enable_metrics_with_url {
        prometheus_exporter::start(addr)?;
    }
```

To before the sub command code so logs work (ideally it is as high as it can go).

Other then the the PR looks good

morph-dev:
> Yeah, I guess we should add a few examples in the readme to document this. You can do this in this PR, or I can make a PR for this after this is merged, or something doesn't matter. But it would be nice to prevent confusion surrounding this in the future

Unless I missed it, I don't think we have documentation for import/export at all. I would leave that for another PR.

KolbyML (approved):

PR #1506, opened 2024-10-02
feat(bridge): add bridge id cli flag for state bridges

njgheorghita:
### What was wrong?
Pretty simple approach to "horizontally scaling" our state bridge nodes. Introduced a `bridge_id` flag, which calculates whether or not an individual bridge is responsible for gossiping a piece of content.

- using a simple id is easier than managing a bridge's node id and using that to calculate distance to the content
- downsides...
  - this approach requires every bridge to iterate over every node in the trie/diff to figure out whether or not it's responsible for gossiping that content. A smarter solution will be needed for gossiping an imported era2 file, or if there's a better solution for chunking the diff, lmk and I'll implement it here.
  - there's no redundancy built in, unless you adjust the `bridge_id` accordingly. aka if we have four bridges, use a `bridge_total` of 2
 
### How was it fixed?
- added `bridge_id` flag and support for filtering offers based on this value

### To-Do

### DIFF
--- a/portal-bridge/src/bridge/state.rs
+++ b/portal-bridge/src/bridge/state.rs
@@ -33,6 +33,7 @@ use trin_utils::dir::create_temp_dir;
 use crate::{
     bridge::history::SERVE_BLOCK_TIMEOUT,
     census::{ContentKey, EnrsRequest},
+    cli::BridgeId,
     types::mode::{BridgeMode, ModeType},
 };
 
@@ -47,6 +48,8 @@ pub struct StateBridge {
     census_tx: mpsc::UnboundedSender<EnrsRequest>,
     // Global offer report for tallying total performance of state bridge
     global_offer_report: Arc<Mutex<GlobalOfferReport>>,
+    // Bridge id used to determine which content keys to gossip
+    bridge_id: BridgeId,
 }
 
 impl StateBridge {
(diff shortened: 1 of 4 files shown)


KolbyML (approved):
:shipit:

portal-bridge/src/bridge/state.rs
```diff
@@ -47,6 +47,10 @@ pub struct StateBridge {
     census_tx: mpsc::UnboundedSender<EnrsRequest>,
     // Global offer report for tallying total performance of state bridge
     global_offer_report: Arc<Mutex<GlobalOfferReport>>,
+    // Bridge id used to determine which content keys to gossip
+    bridge_id: u64,
```
morph-dev, line 51: Why not make this: `bridge_id: BridgeId`?

You can also move `should_gossip_content(&self, content_idx: usize) -> bool` inside BridgeId. But maybe rename it to `is_selected(&self, index:usize) -> bool`. Up to you?

portal-bridge/src/bridge/state.rs
```diff
@@ -66,6 +70,8 @@ impl StateBridge {
             offer_semaphore,
             census_tx,
             global_offer_report: Arc::new(Mutex::new(global_offer_report)),
+            bridge_id: 1,
```
morph-dev, line 73: Shouldn't this get value from `BridgeConfig`?

morph-dev (approved):
few small comments

:ship:

PR #1507, opened 2024-10-02
feat: update `portal_*Offer` to handle multiple pieces of content

KolbyML:
### What was wrong?
updates Trin to match proposed spec changes https://github.com/ethereum/portal-network-specs/pull/342
### How was it fixed?
updating our code to match the spec

### DIFF
--- a/ethportal-api/src/types/portal.rs
+++ b/ethportal-api/src/types/portal.rs
@@ -32,6 +32,8 @@ pub struct PongInfo {
 
 pub type FindNodesInfo = Vec<Enr>;
 
+pub const MAX_AMOUNT_OF_OFFERED_CONTENT_KEYS: usize = 64;
+
 /// Response for Offer endpoint
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
--- a/ethportal-api/src/types/portal.rs
+++ b/ethportal-api/src/types/portal.rs
@@ -32,7 +32,7 @@ pub struct PongInfo {
 
 pub type FindNodesInfo = Vec<Enr>;
 
-pub const MAX_AMOUNT_OF_OFFERED_CONTENT_KEYS: usize = 64;
+pub const MAX_CONTENT_KEYS_PER_OFFER: usize = 64;
 
 /// Response for Offer endpoint
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
--- a/ethportal-peertest/src/scenarios/offer_accept.rs
+++ b/ethportal-peertest/src/scenarios/offer_accept.rs
@@ -81,8 +81,7 @@ pub async fn test_populated_offer(peertest: &Peertest, target: &Client) {
     let result = target
         .offer(
             Enr::from_str(&peertest.bootnode.enr.to_base64()).unwrap(),
-            content_key.clone(),
-            content_value.encode(),
+            vec![(content_key.clone(), content_value.encode())],
         )
         .await
         .unwrap();
@@ -144,8 +143,7 @@ pub async fn test_offer_propagates_gossip(peertest: &Peertest, target: &Client) 
     target
         .offer(
             peertest.bootnode.enr.clone(),
-            content_key.clone(),
-            content_value.encode(),
+            vec![(content_key.clone(), content_value.encode())],
         )
         .await
         .unwrap();
@@ -221,8 +219,7 @@ pub async fn test_offer_propagates_gossip_multiple_content_values(
     target
         .offer(
             peertest.bootnode.enr.clone(),
-            header_key.clone(),
-            header_value.encode(),
+            vec![(header_key.clone(), header_value.encode())],
         )
         .await
         .unwrap();
--- a/ethportal-peertest/src/scenarios/offer_accept.rs
+++ b/ethportal-peertest/src/scenarios/offer_accept.rs
@@ -18,32 +18,8 @@ use ethportal_api::{
     ContentValue, Discv5ApiClient, HistoryNetworkApiClient,
 };
 
-pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
-    info!("Testing Unpopulated OFFER/ACCEPT flow");
-
-    let (content_key, content_value) = fixture_header_by_hash();
-
-    // Send wire offer request from testnode to bootnode
-    let result = target
-        .offer(
-            Enr::from_str(&peertest.bootnode.enr.to_base64()).unwrap(),
-            vec![(content_key.clone(), content_value.encode())],
-        )
-        .await
-        .unwrap();
-
-    // Check that ACCEPT response sent by bootnode accepted the offered content
-    assert_eq!(hex_encode(result.content_keys.into_bytes()), "0x03");
-
-    // Check if the stored content value in bootnode's DB matches the offered
-    assert_eq!(
-        content_value,
-        wait_for_history_content(&peertest.bootnode.ipc_client, content_key).await,
-    );
-}
-
-pub async fn test_populated_offer(peertest: &Peertest, target: &Client) {
-    info!("Testing Populated Offer/ACCEPT flow");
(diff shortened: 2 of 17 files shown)


ethportal-peertest/src/scenarios/offer_accept.rs
morph-dev, line 1: Should we add test here as well?

rpc/src/beacon_rpc.rs
```diff
@@ -179,23 +179,35 @@ impl BeaconNetworkApiServer for BeaconNetworkApi {
...
         &self,
         enr: Enr,
-        content_key: BeaconContentKey,
-        content_value: RawContentValue,
+        content_items: Vec<(BeaconContentKey, RawContentValue)>,
     ) -> RpcResult<AcceptInfo> {
-        let content_value = BeaconContentValue::decode(&content_key, &content_value)
-            .map_err(RpcServeError::from)?;
-        let endpoint = BeaconEndpoint::Offer(enr, content_key, content_value);
+        if !(1..=64).contains(&content_items.len()) {
```
morph-dev, line 191: We should probably define `64` as a constant somewhere in the `ethportal-api`.

portalnet/src/overlay/protocol.rs
```diff
@@ -509,13 +509,10 @@ where
...
         enr: Enr,
-        content_key: RawContentKey,
-        content_value: Vec<u8>,
+        content_items: Vec<(RawContentKey, Vec<u8>)>,
     ) -> Result<Accept, OverlayRequestError> {
         // Construct the request.
-        let request = Request::PopulatedOffer(PopulatedOffer {
-            content_items: vec![(content_key, content_value)],
-        });
+        let request = Request::PopulatedOffer(PopulatedOffer { content_items });
```
morph-dev, line 515: In my opinion, check that number of content items is 1..=64 should happen here (in which case it should return `Err(OverlayRequestError::InvalidRequest(..))`).

In that case, it's probably not needed in `*_rpc.rs`.

morph-dev (commented):
It looks good in general, but there are few things I want to discuss.

The *TraceOffer is not defined in spec. Should we make it accept multiple items as well?

The *WireOffer is also not defined in spec, and if I'm not wrong, it was introduced because there was no way to test sending multiple content items using portal network OFFER request. Do we still need it? Can we remove it?

If you feel this can be left for future PR, I'm fine with that as well (but in that case, we should have an issue for it).

portalnet/src/overlay/protocol.rs
KolbyML: I don't think we should move the request check to `portalnet/src/overlay/protocol.rs`. Because then we would be sending possibly over 64 key-value pairs through the `*_rx` channel which I think should be avoided. Yes this suggestion would reduce code duplication, but I don't think we should push validating length checks after a channel. We should do the range check as soon as possible to avoid that send. Why waste extra resources when it can be avoided.

ethportal-peertest/src/scenarios/offer_accept.rs
KolbyML: I will remove `WireOffer` then `Offer` will take it's place for tests which should be sufficient for test coverage

KolbyML:
> It looks good in general, but there are few things I want to discuss.
> 
> The *TraceOffer is not defined in spec. Should we make it accept multiple items as well?


I will remove `WireOffer` and replace it with `Offer`. I think `TraceOffer` is a custom endpoint that will likely be removed, assuming trin-bridge stops using the JSON-RPC and integrates Trin instead of running Trin as a binary. Because `TraceOffer` is only used for bridging/internal use, I don't think we need to change it unless something requires it.

> The *WireOffer is also not defined in spec, and if I'm not wrong, it was introduced because there was no way to test sending multiple content items using portal network OFFER request. Do we still need it? Can we remove it?
> 
> If you feel this can be left for future PR, I'm fine with that as well (but in that case, we should have an issue for it).

I will remove WireOffer and use Offer directly

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -22,19 +22,12 @@ pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
     info!("Testing Unpopulated OFFER/ACCEPT flow");
```
morph-dev, line 22: This is no longer "unpopulated" flow. If I'm not mistaken, we no longer have "unpopulated" flow.

Either rename/update current test, or remove it if you think it's redundant.

portalnet/src/overlay/protocol.rs
morph-dev: The reason why I think it should be there is not because I want to avoid code duplication, but because this type of check should be at lower level. Reason being is that what if there is alternative call to this function (or is added in the future).

Now that I think about it, we might want it at even lower level (e.g. right before sending offer request over network). Because poke, and gossip on deletion (and maybe some other mechanics) wouldn't have this check.

But, regarding your point, it's fine to have it on upper level as well (and not make calls that we know would fail).

morph-dev (commented):
Look good from my point of view. But I will let @njgheorghita to give approval as I think he should take a look as well.

I still think that `TraceOffer` should behave the same as `Offer` but with trace. But I'm not going to insist on it.

ethportal-api/src/types/portal.rs
```diff
@@ -32,6 +32,8 @@ pub struct PongInfo {
 
 pub type FindNodesInfo = Vec<Enr>;
 
+pub const MAX_AMOUNT_OF_OFFERED_CONTENT_KEYS: usize = 64;
```
njgheorghita, line 35: *nit* `MAX_CONTENT_KEYS_PER_OFFER`

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -49,40 +42,14 @@ pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
...
-        Ok(_) => panic!("Unpopulated offer should have failed"),
-        Err(e) => {
-            assert!(e
-                .to_string()
-                .contains("Content key not found in local store"));
-        }
-    }
-}
-
 pub async fn test_populated_offer(peertest: &Peertest, target: &Client) {
```
njgheorghita, line 45: Along with @morph-dev comment above, you can update the name of this test & log to just `test_offer` ... and remove any reference to `unpopulated`/`populated` in other test names / logs / comments in this file

ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -49,40 +42,14 @@ pub async fn test_unpopulated_offer(peertest: &Peertest, target: &Client) {
...
     info!("Testing Populated Offer/ACCEPT flow");
```
njgheorghita, line 46: *nit* `Testing OFFER/ACCEPT flow` / `Testing Offer/Accept flow` (& in other test's `info!` logs)

portalnet/src/overlay/protocol.rs
njgheorghita: We do have a check at the lower level https://github.com/ethereum/trin/blob/master/portalnet/src/gossip.rs#L103-L115 ... but i'd include an update in this pr to use the `MAX_AMOUNT_OF_OFFERED_CONTENT_KEYS` const in that code

njgheorghita (approved):
:shipit: 

> I still think that TraceOffer should behave the same as Offer but with trace. But I'm not going to insist

Since the bridges are only offering a single content item at a time, there isn't any immediate benefit to this change. So I'm not gonna insist either. But it might end up being useful in the state bridge fairly soon assuming that we'll want to gossip more than 1 content item -> peer per content type in each diff (or when gossiping an era2 file) which seems likely imo. Though this can all be handled in a future pr

portalnet/src/overlay/protocol.rs
KolbyML: I will update the code to use that const. I think it is fine if we have a lower and higher level check

PR #1509, opened 2024-10-06
feat: split E2StoreStream into reader and writer

morph-dev:
### What was wrong?

The `E2StoreStream` uses `File` structure for reading/writing into file. This can be more optimized by using `BufReader`/`BufWriter`.

### How was it fixed?

I split `E2StoreStream` into `E2StoreStreamReader` and `E2StoreStreamWriter` that use `BufReader`/`BufWriter`. I also had to split `Era2` into `Era2Reader` and `Era2Writer`.

In my opinion, this split makes more sense as it would most likely be wrong to read and write into the same file at the same time.

To measure if this actually improves performance as I expect, I did the following:

0. generated era2 files for blocks 3 million, 4 million and 5 million
1. built release version of `trin-execution` binary at both `master` and this `pr`
2. for both builds (master/pr), imported each era2 state into clean directory
3. exported era2 from directory from step2

Here are the results

| block number | import master | import pr | export master | export pr |
|--------------|---------------|-----------|---------------|-----------|
| 3'000'000    | 2m11s         | 2m10s     | 37s           | 18s       |
| 4'000'000    | 8m55s         | 9m01s     | 2m44s         | 1m35s     |
| 5'000'000    | 54m22s        | 52m55s    | 43m38s        | 37m02s    |

We can see that there is no big difference when in comes to importing era2 file (reading), but there is noticeable improvement for exporting (writing). 

### To-Do

### DIFF
--- a/e2store/README.md
+++ b/e2store/README.md
@@ -23,4 +23,4 @@ TODO: Add chart of snapshot size at every million block interval.
 
 ## What is the difference between `e2store/memory.rs` and `e2store/stream.rs`
 
-The `e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This is required for `.era2` format for storing full flat state snapshots.
+`e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This is required for `.era2` format for storing full flat state snapshots.
--- a/e2store/README.md
+++ b/e2store/README.md
@@ -23,4 +23,4 @@ TODO: Add chart of snapshot size at every million block interval.
 
 ## What is the difference between `e2store/memory.rs` and `e2store/stream.rs`
 
-`e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This will be required in `.era2` a format for storing full flat state snapshots.
+The `e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This is required for `.era2` format for storing full flat state snapshots.
--- a/e2store/src/e2store/stream.rs
+++ b/e2store/src/e2store/stream.rs
@@ -7,13 +7,12 @@ use std::{
 use super::types::{Entry, Header};
 
 /// Streaming reader for e2store files.
-pub struct E2StoreStreamReader {
-    reader: BufReader<File>,
+pub struct E2StoreStreamReader<R> {
+    reader: BufReader<R>,
 }
 
-impl E2StoreStreamReader {
-    pub fn new(e2store_path: &Path) -> anyhow::Result<Self> {
-        let reader = BufReader::new(File::open(e2store_path)?);
+impl<R: Read> E2StoreStreamReader<R> {
+    pub fn new(reader: BufReader<R>) -> anyhow::Result<Self> {
         Ok(Self { reader })
     }
 
@@ -27,14 +26,19 @@ impl E2StoreStreamReader {
     }
 }
 
-/// Streaming writer for e2store files.
-pub struct E2StoreStreamWriter {
-    writer: BufWriter<File>,
+impl E2StoreStreamReader<File> {
+    pub fn open(path: &Path) -> anyhow::Result<Self> {
+        Self::new(BufReader::new(File::open(path)?))
+    }
 }
 
-impl E2StoreStreamWriter {
-    pub fn new(e2store_path: &Path) -> anyhow::Result<Self> {
-        let writer = BufWriter::new(File::create(e2store_path)?);
+/// Streaming writer for e2store files.
+pub struct E2StoreStreamWriter<W: Write> {
+    writer: BufWriter<W>,
+}
+
+impl<W: Write> E2StoreStreamWriter<W> {
+    pub fn new(writer: BufWriter<W>) -> anyhow::Result<Self> {
         Ok(Self { writer })
     }
 
@@ -47,6 +51,12 @@ impl E2StoreStreamWriter {
 
     pub fn flush(&mut self) -> anyhow::Result<()> {
         Ok(self.writer.flush()?)
+    }
+}
+
+impl E2StoreStreamWriter<File> {
+    pub fn create(path: &Path) -> anyhow::Result<Self> {
+        Self::new(BufWriter::new(File::create(path)?))
     }
 }
 
@@ -70,7 +80,7 @@ mod tests {
             .join(format!("{}.e2store_stream_test", random_number));
 
         // create a new e2store file and write some data to it
-        let mut e2store_stream_writer = E2StoreStreamWriter::new(&tmp_path)?;
+        let mut e2store_stream_writer = E2StoreStreamWriter::create(&tmp_path)?;
 
         let version = VersionEntry::default();
         e2store_stream_writer.append_entry(&version.clone().into())?;
@@ -82,7 +92,7 @@ mod tests {
         drop(e2store_stream_writer);
 
         // read results and see if they match
-        let mut e2store_stream_reader = E2StoreStreamReader::new(&tmp_path)?;
+        let mut e2store_stream_reader = E2StoreStreamReader::open(&tmp_path)?;
         let read_version_entry = VersionEntry::try_from(&e2store_stream_reader.next_entry()?)?;
         assert_eq!(version, read_version_entry);
         let read_entry = e2store_stream_reader.next_entry()?;
--- a/e2store/src/e2store/stream.rs
+++ b/e2store/src/e2store/stream.rs
@@ -1,51 +1,59 @@
 use std::{
     fs::File,
-    io::{Read, Write},
-    path::PathBuf,
+    io::{BufReader, BufWriter, Read, Write},
+    path::Path,
 };
 
 use super::types::{Entry, Header};
 
-/// e2store/memory.rs was built to load full .era/.era2 files into memory and provide a simple API
-/// to access the data. The issue for this is for larger files this wouldn't be feasible, as the
-/// entire file would need to be loaded into memory. This is where e2store_file.rs comes in, it
-/// provides a way to read and write e2store files in a streaming fashion.
-pub struct E2StoreStream {
-    pub e2store_file: File,
+/// Streaming reader for e2store files.
+pub struct E2StoreStreamReader {
+    reader: BufReader<File>,
 }
 
-impl E2StoreStream {
-    pub fn open(e2store_path: &PathBuf) -> anyhow::Result<Self> {
-        let e2store_file = File::open(e2store_path)?;
-        Ok(Self { e2store_file })
-    }
-
-    pub fn create(e2store_path: &PathBuf) -> anyhow::Result<Self> {
-        let e2store_file = File::create(e2store_path)?;
-        Ok(Self { e2store_file })
+impl E2StoreStreamReader {
+    pub fn new(e2store_path: &Path) -> anyhow::Result<Self> {
+        let reader = BufReader::new(File::open(e2store_path)?);
+        Ok(Self { reader })
     }
 
     pub fn next_entry(&mut self) -> anyhow::Result<Entry> {
         let mut buf = vec![0; 8];
-        self.e2store_file.read_exact(&mut buf)?;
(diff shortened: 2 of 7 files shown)


KolbyML:
If there is no difference in performance using `BufReader` that should be removed from the PR

KolbyML:
How big are the .era2 files for 3,4,5 million can you upload them to era2.ethportal.net.

Will there be a performance difference at block 10,15,20 million well exporting?

morph-dev:
> If there is no difference in performance using `BufReader` that should be removed from the PR

I would say that it probably has improvements, but they might not be significant. And in my setup, what took the most time was recreating the storage trie and writing it to rocks db.

Sure, since it might not be as relevant, I can remove `BufReader`. But I would still keep `E2StoreStreamReader` and `E2StoreStreamWriter` separate (as they need to be because of the `BufWriter`). In which case, we might just keep `BufReader` because it should be better than using `File`.


> How big are the .era2 files for 3,4,5 million can you upload them to era2.ethportal.net.

3 million - 358 MB / 4 million - 1.5 GB / 5 million - 7.5 GB

> Will there be a performance difference at block 10,15,20 million well exporting?

I didn't try as it would take too long, but I assume there will be.
To have good results of my benchmark, I would have to not use my computer during that period (as there is a lot of CPU/RAM usage while processing trie), and me using the computer can affect it.

I guess I can do a test in which I'm just reading/writing to RAM memory (not actually processing the storage) and remove the noise. I can look into it if you think it's necessary.

morph-dev:
> How big are the .era2 files for 3,4,5 million can you upload them to era2.ethportal.net.

And yes, I plan to upload them, once I export few more.

KolbyML:
> > If there is no difference in performance using `BufReader` that should be removed from the PR
> 
> I would say that it probably has improvements, but they might not be significant. And in my setup, what took the most time was recreating the storage trie and writing it to rocks db.
> 
> Sure, since it might not be as relevant, I can remove `BufReader`. But I would still keep `E2StoreStreamReader` and `E2StoreStreamWriter` separate (as they need to be because of the `BufWriter`). 

I am fine with this

> In which case, we might just keep `BufReader` because it should be better than using `File`.

Unless there is an improvement that can be measured I don't see why we would add an extra wrapper. "`BufReader` because it should be better than using `File`"  with the measurements above these is no clear indication in these circumstances there is an improvement if there is an improvement in the future we can add the wrapper back but until then I don't see why would have the extra needless lines.

> > How big are the .era2 files for 3,4,5 million can you upload them to era2.ethportal.net.
> 
> 3 million - 358 MB / 4 million - 1.5 GB / 5 million - 7.5 GB

ty you for the stats

> > Will there be a performance difference at block 10,15,20 million well exporting?
> 
> I didn't try as it would take too long, but I assume there will be. To have good results of my benchmark, I would have to not use my computer during that period (as there is a lot of CPU/RAM usage while processing trie), and me using the computer can affect it.
> 
> I guess I can do a test in which I'm just reading/writing to RAM memory (not actually processing the storage) and remove the noise. I can look into it if you think it's necessary.

No worries it would be cool to know but not a blocker to adding  `BufWriter` which has measurable improvements

morph-dev:
I created a binary that just reads the era2 file and collects stats on how many account entries, storage entries, storage items and non-empty bytecodes there are.
Exported commit [here](https://github.com/morph-dev/trin/commit/e3c0c057bf47d0efbb75ab01ae84ef7c3c954a0b), and I think it can be useful and I plan to make a pr, unless you have some objection.

This is the cleanest benchmark for `File` vs `BufReader` in context of era2 file that I can think of. Here are the result:
| block_number | master | pr     |
|--------------|--------|--------|
| 3'000'000    | 18s    | 17s    |
| 4'000'000    | 1m15s  | 1m12s  |
| 5'000'000    | 5m15s  | 5m0s   |
| 10'000'000   | 21m2s  | 19m51s |

The benefits of `BufReader` are noticeable, but they are not huge.

According to [documentation](https://doc.rust-lang.org/std/io/struct.BufReader.html): `BufReader<R> can improve the speed of programs that make small and repeated read calls to the same file or network socket`. I think this fits nicely in our use case, expect that we have a mix of small (account_entry) and big reads (storage_entry). 
If we decide to support Stream reader on some type that is not era2, we might have bigger improvements. Or if we decide to stream it directly from the internet.

No matter how I look at it, it is or can be beneficial to use `BufReader`, without downside (except that it requires extra line of code for initialization).

KolbyML:
> I created a binary that just reads the era2 file and collects stats on how many account entries, storage entries, storage items and non-empty bytecodes there are. Exported commit [here](https://github.com/morph-dev/trin/commit/e3c0c057bf47d0efbb75ab01ae84ef7c3c954a0b), and I think it can be useful and I plan to make a pr, unless you have some objection.

You can make a PR for it if you want, I don't see a use case for it right now but that is me.

> This is the cleanest benchmark for `File` vs `BufReader` in context of era2 file that I can think of. Here are the result:
> 
> block_number	master	pr
> 3'000'000	18s	17s
> 4'000'000	1m15s	1m12s
> 5'000'000	5m15s	5m0s
> 10'000'000	21m2s	19m51s
> The benefits of `BufReader` are noticeable, but they are not huge.

I think these numbers are within the margin of error i.e. it doesn't make a difference

> According to [documentation](https://doc.rust-lang.org/std/io/struct.BufReader.html): `BufReader<R> can improve the speed of programs that make small and repeated read calls to the same file or network socket`. I think this fits nicely in our use case, expect that we have a mix of small (account_entry) and big reads (storage_entry). 

In testing there is no difference within the margin of error

> If we decide to support Stream reader on some type that is not era2, we might have bigger improvements. 

I don't see how that is related to this PR

> Or if we decide to stream it directly from the internet.

Is there a benchmark for this?

> No matter how I look at it, it is or can be beneficial to use `BufReader`, without downside (except that it requires extra line of code for initialization).

If there was no downside to `BufReader` it would be the default functionality of `File`. This PR is an optimization PR so if a performance increase can't be measured (10 seconds on 10 million isn't a performance increase) then it doesn't make sense to add the complexity.


If you want to keep it go for it, but fundamentally it doesn't make sense to keep it unless there is something measurable. Maybe testing it on a harddrive would help for a worse case possibility it could help?

KolbyML:
Why is importing at 5 million takes over 50 minutes, but 10 million takes over 20 minutes?

morph-dev:
> I don't see how that is related to this PR

It's related because change to `BufReader`/`BufWriter` is done in `e2store/src/e2store/stream.rs`. The `Era2` is currently the only user and motivation for this change, but any future usage of it will be beneficial as well (at that point, nobody is going to go back and look how `E2StoreStream*` works and if it can be optimized).

> Why is importing at 5 million takes over 50 minutes, but 10 million takes over 20 minutes?

It's measuring different things.

Measurement in the PR description (50min for block 5 million) is importing entire state from era2 file, meaning reading era2 file, constructing account and all contract tries, saving everything into RocksDB (using import-state subcommand).

The second one, from my last [reply](https://github.com/ethereum/trin/pull/1509#issuecomment-2396676486) (20 minutes for block 10 million) is just reading era2 and doing some counting (number of accounts, storage entries, storage items).

> ... (10 seconds on 10 million isn't a performance increase) ...

It's not 10 seconds, it's **1 min 10 sec** for block 10 million (which takes ~21 min). That is benefit of 5%. And if we look closely at other numbers, they are consistently 5% lower.

Yes, in the context of importing entire state from era2 file (first measurement), then saving 15 seconds on reading era2 file is negligible in comparison of the entire process (processing that data and saving to RocksDB). That's the reason we didn't see significant improvements for import-state, and the reason I did second measurement that is focused only on reading era2 file.

> I think these numbers are within the margin of error i.e. it doesn't make a difference

To declare it is or not within margin of error, I would have to repeat the measurement many times, under many different situations, and I don't want to spend time doing that. The fact that it's 5% across several measurements (I did few more than presented here, and numbers were stable) is enough for me that there is benefit (together with official documentation that supports that claim).

> If there was no downside to BufReader it would be the default functionality of File.

Even the documentation of [File](https://doc.rust-lang.org/std/fs/struct.File.html) says that wrapping it in `BufReader`/`BufWriter` should be considered for efficiency when performing many small read/write calls.
The downside is that you can't (easily) switch between read/write mode (we don't support this use case anyway) or if for some other reason you want to read/write exact amount of data (and not buffer it).

> Maybe testing it on a harddrive would help for a worse case possibility it could help?

I only have SSD on my new machine, but I can give it a try on old machine with HDD (if I have time, I can do it tomorrow).

e2store/README.md
```diff
@@ -23,4 +23,4 @@ TODO: Add chart of snapshot size at every million block interval.
 
 ## What is the difference between `e2store/memory.rs` and `e2store/stream.rs`
 
-`e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This will be required in `.era2` a format for storing full flat state snapshots.
+The `e2store/memory.rs` provides an api to load a full e2store file such as `.era`/`.era1` and manipulate it in memory. For smaller e2store files this approach works well. The issue comes when dealing with e2store files of much greater size loading the whole file into memory at once often isn't possible. This is where `e2store/stream.rs` comes in where you can stream the data you need from a e2store file as you need it. This is required for `.era2` format for storing full flat state snapshots.
```
KolbyML, line 26: I don't think the makes gramatical sense here

e2store/src/e2store/stream.rs
```diff
@@ -1,51 +1,59 @@
...
-    pub fn open(e2store_path: &PathBuf) -> anyhow::Result<Self> {
-        let e2store_file = File::open(e2store_path)?;
-        Ok(Self { e2store_file })
-    }
-
-    pub fn create(e2store_path: &PathBuf) -> anyhow::Result<Self> {
-        let e2store_file = File::create(e2store_path)?;
-        Ok(Self { e2store_file })
+impl E2StoreStreamReader {
+    pub fn new(e2store_path: &Path) -> anyhow::Result<Self> {
```
KolbyML, line 15: could we maintain the `open` naming scheme here as we are opening a file still

e2store/src/e2store/stream.rs
```diff
@@ -1,51 +1,59 @@
...
     }
+}
+
+/// Streaming writer for e2store files.
+pub struct E2StoreStreamWriter {
+    writer: BufWriter<File>,
+}
+
+impl E2StoreStreamWriter {
+    pub fn new(e2store_path: &Path) -> anyhow::Result<Self> {
```
KolbyML, line 36: same here but for create even though we are using a `BufWriter` we are still creating a new file so I think that would be more clear in this context

e2store/src/era2.rs
```diff
@@ -21,71 +48,41 @@ use crate::{
...
-            e2store_stream,
-            pending_storage_entries: 0,
-            path,
-        })
-    }
-
-    pub fn create(path: PathBuf, header: Header) -> anyhow::Result<Self> {
-        fs::create_dir_all(&path)?;
+impl Era2Writer {
+    pub fn new(path: &Path, header: Header) -> anyhow::Result<Self> {
```
KolbyML, line 65: I think we should still use create here instead of new

e2store/src/era2.rs
```diff
@@ -123,22 +120,60 @@ impl Era2 {
...
+    pub version: VersionEntry,
+    pub header: HeaderEntry,
+
+    reader: E2StoreStreamReader,
+    pending_storage_entries: u32,
+    path: PathBuf,
+}
+
+impl Era2Reader {
+    pub fn new(path: &Path) -> anyhow::Result<Self> {
```
KolbyML, line 149: I think we should still do open instead of new, because we are opening an era2 file

KolbyML (approved):
I misread the chat and thought it was a 10 second difference sorry about that.

:shipit: most of the PR looks good. I think we should maintain `create` and `open` initialization conventions as we are either opening or creating the files. I don't think `new` is descriptive enough in this case. 99% of causes I would agree we should use `new` to instantiate a struct, but I don't think it makes sense here. I think following the conventions of `File` like we were doing before makes the most sense

morph-dev:
> I think we should maintain `create` and `open` initialization conventions as we are either opening or creating the files. I don't think `new` is descriptive enough in this case. 99% of causes I would agree we should use `new` to instantiate a struct, but I don't think it makes sense here. I think following the conventions of `File` like we were doing before makes the most sense

I made `E2StoreStreamReader` and `E2StoreStreamWriter` generic, so they can read/write from any type of source (not only file).
I created `new` accept any object that implements `Read`/`Write` (wrapped in BufReader/BufWriter), and I created `open`/`create` functions for working with files.  I think this is the best outcome. What do you think?

KolbyML:
> > I think we should maintain `create` and `open` initialization conventions as we are either opening or creating the files. I don't think `new` is descriptive enough in this case. 99% of causes I would agree we should use `new` to instantiate a struct, but I don't think it makes sense here. I think following the conventions of `File` like we were doing before makes the most sense
> 
> I made `E2StoreStreamReader` and `E2StoreStreamWriter` generic, so they can read/write from any type of source (not only file). I created `new` accept any object that implements `Read`/`Write` (wrapped in BufReader/BufWriter), and I created `open`/`create` functions for working with files. The era2 structures have only `open`/`create` and only support working with files (at least for now).
> 
> I think this is the best outcome. What do you think?

I like this. I think I see how we could do the downloading while processing example you gave earlier assuming the HTTP client we use implements the Read and Write trait.

morph-dev:
> I like this. I think I see how we could do the downloading while processing example you gave earlier assuming the HTTP client we use implements the Read and Write trait.

I think that `reqwest`'s blocking version supports it (at least it seems that `Response` object implements `Read` ([source](https://docs.rs/reqwest/latest/reqwest/blocking/struct.Response.html))).

The async (default) version also has support for streaming (using `Response::bytes_stream()` [link](https://docs.rs/reqwest/latest/reqwest/struct.Response.html#method.bytes_stream)), but I think we can't use current approach to integrate with it. If we want, we can probably support this flow as well (with a bit of refactoring).

PR #1511, opened 2024-10-07
fix(state bridge): gossip contract storage tries

morph-dev:
### What was wrong?

The state bridge is not gossiping contract storage content.

After some debugging, I figured out that the issue is that `EvmDB` is cloned, but `storage_cache` is not behind `Arc`.
Because of that, the `TrinExecution.database.storage_cache` is never updated (instead, only `BlockExecutor.evm.db_mut().database` is).

### How was it fixed?

Wrapped `storage_cache` in `Arc<Mutex<_>>`.

Added additional metric to `test_we_can_generate_content_key_values_up_to_x` to confirm that issue was fixed and measure the difference.

Also tested locally and can confirm that state bridge now gossips contract storage content.

### To-Do

### DIFF
--- a/portal-bridge/src/bridge/state.rs
+++ b/portal-bridge/src/bridge/state.rs
@@ -112,7 +112,7 @@ impl StateBridge {
                 .process_range_of_blocks(start_block - 1, None)
                 .await?;
             // flush the database cache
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
:shipit: looks good

PR #1512, opened 2024-10-07
fix(beacon): revert column name change to `historical_summaries` db table

ogenev:
### What was wrong?
Deploying beacon on testnet bootnodes and state regular nodes, causes Trin to fail because of an implemented db column change without a proper migration. This change was introduced [here](https://github.com/ethereum/trin/pull/1413/files#diff-d79c9290a61b26e8d078e761fe5397feb9ace5b8a73ef8694b7fb74dbe4b905cR57).

I planned to modify the testnet databases with a bash script and ansible, but I had issues connecting to the running docker containers sqlite files and went into a rabbit hole.

### How was it fixed?
Revert the column name change causing the issue back to the original.

This is a quick and dirty?! fix to deploy beacon on testnet without db issues until we implement a proper migration for the beacon db.

### To-Do


KolbyML (approved):
:shipit:

PR #1513, opened 2024-10-08
feat: add binary for simple era2 analysis

morph-dev:
Adding binary that reads era2 file and prints most basic stats.

I used it to measure performance improvements of #1509, but I think it can't hurt to merge it as well (as it can easily be modified to do more complex analysis).

### To-Do

### DIFF
--- a/e2store/src/bin/era2-stats.rs
+++ b/e2store/src/bin/era2-stats.rs
@@ -26,7 +26,8 @@ struct Stats {
 /// Reads the era2 file and prints stats about it.
 ///
 /// It can be run with following command:
-/// ```
+///
+/// ```bash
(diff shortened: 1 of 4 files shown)


e2store/src/bin/era2-stats.rs
```diff
@@ -0,0 +1,72 @@
...
+struct Stats {
+    account_count: usize,
+    bytecode_count: usize,
+    storage_entry_count: usize,
+    storage_item_count: usize,
+}
+
+/// Reads the era2 file and prints stats about it.
+///
+/// It can be run with following command:
```
njgheorghita, line 28: *nit* i'd also copy this into the `README`

njgheorghita (approved):

PR #1514, opened 2024-10-08
test: add test for default utp concurrency limit

njgheorghita:
### What was wrong?
Added a test to make sure that trin will always manage at least the default # of concurrent utp txs. The test/constant can also easily be increased to test trin's real tx capacity (currently somewhere around 100 from what I've seen, but still working to see if this can be improved)

### How was it fixed?
Added the test

### To-Do

### DIFF
--- a/ethportal-peertest/src/utils.rs
+++ b/ethportal-peertest/src/utils.rs
@@ -29,7 +29,7 @@ where
     Fut: Future<Output = Result<O>>,
 {
     // If content is absent an error will be returned.
-    for counter in 0..60 {
+    for counter in 0..10 {
         let result = f().await;
         match result {
             Ok(val) => return val,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2235,6 +2235,7 @@ dependencies = [
  "alloy-rlp",
  "anyhow",
  "discv5",
+ "e2store",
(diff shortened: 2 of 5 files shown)


KolbyML (approved):
:shipit:

ethportal-peertest/src/utils.rs
```diff
@@ -29,7 +29,7 @@ where
     Fut: Future<Output = Result<O>>,
 {
     // If content is absent an error will be returned.
-    for counter in 0..60 {
```
morph-dev, line 32: I think this made some tests flaky, at least when running on circleci. The test that is frequently failing (with "Retried too many times" panic) is `peertest_history_offer_propagates_gossip_multiple_large_content_values`.

Test passes fine when I run it locally.

njgheorghita: Hmmm, thanks for the heads up. I'll look into it

njgheorghita: Well, I made this change since it's a real pain to wait for a minute if the test fails locally. But, I have no problem with you reverting the `10` -> `60` in #1515 . It's fairly simple to make this change when testing locally if it becomes a pain in the future

PR #1515, opened 2024-10-09
refactor(build): use alloy as dependency instead of invidivual crates

morph-dev:
## Issue

Currently we use many different `alloy-*` crates. Adding new `alloy-*` crate or updating existing ones can be tricky, as we don't want many different versions (and sometimes different versions don't work well together).

## Proposed solution

Idea of this PR is that trin should depend only on `alloy` [crate](https://docs.rs/crate/alloy/latest), which is Meta-crate of all others. It includes only `alloy-primitives` by default, and other "sub-crates" can be enabled using features.

I'm not sure if this is something we want, but I wanted to crate this PR and we can discuss and make a decision.
I think this makes maintaining and updating dependencies much easier, and I don't see too many drawbacks.

### Sidenote

It seems that `RlpDecodable` and `RlpEncodable` derive macros don't work unless `alloy-rlp` is declared as dependency directly. I will try to create minimal reproducible case of this and open an issue at alloy github.

In addition, it seems that "serde" feature doesn't include "serde" feature of "alloy-rpc-types-eth".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,11 +13,7 @@ categories = ["cryptography::cryptocurrencies"]
 description = "A Rust implementation of the Ethereum Portal Network"
 
 [dependencies]
-alloy-eips.workspace = true
-alloy-primitives.workspace = true
-alloy-provider = { version = "0.4.2", features = ["ipc", "ws"]}
-alloy-pubsub = { version = "0.4.2" }
-alloy-rpc-types.workspace = true
+alloy = { workspace = true, features = ["eips", "provider-ipc", "provider-ws", "pubsub", "reqwest", "rpc-types"] }
 anyhow.workspace = true
 clap.workspace = true
 dirs = "5.0.1"
@@ -80,11 +76,8 @@ members = [
 ]
 
 [workspace.dependencies]
-alloy-consensus = "0.4.2"
-alloy-eips = "0.4.2"
-alloy-primitives = { version ="0.8.7" , features = ["map-hashbrown"] }
-alloy-rlp = "0.3.8"
-alloy-rpc-types = "0.4.2"
+alloy = { version = "0.4.2", default-features = false, features = ["std"] }
+alloy-rlp = { version = "0.3.8", default-features = false, features = ["derive"] }
 anyhow = "1.0.68"
 async-trait = "0.1.68"
 bytes = "1.3.0"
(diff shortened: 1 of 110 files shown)


KolbyML (approved):
:shipit:

trin-execution/Cargo.toml
```diff
@@ -12,11 +12,9 @@ authors = ["https://github.com/ethereum/trin/graphs/contributors"]
 
 
 [dependencies]
-alloy-consensus.workspace = true
-alloy-eips.workspace = true
-alloy-primitives.workspace = true
+alloy = { workspace = true, features = ["eips", "rpc-types-engine", "serde"] }
 alloy-rlp.workspace = true
-alloy-rpc-types = { workspace = true, features = ["engine"]}
+alloy-rpc-types = "0.4.2"
```
carver, line 17: Can this line be deleted, by adding the `rpc-types` feature in the alloy dependency above?

Cargo.toml
```diff
@@ -80,11 +76,8 @@ members = [
 ]
 
 [workspace.dependencies]
-alloy-consensus = "0.4.2"
-alloy-eips = "0.4.2"
-alloy-primitives = { version ="0.8.7" , features = ["map-hashbrown"] }
+alloy = { version = "0.4.2", default-features = false, features = ["std"] }
 alloy-rlp = "0.3.8"
```
carver, line 80: Should `alloy-rlp` get bumped to "0.4.2" also?

ethportal-api/Cargo.toml
```diff
@@ -11,10 +11,9 @@ categories = ["cryptography::cryptocurrencies"]
 authors = ["https://github.com/ethereum/trin/graphs/contributors"]
 
 [dependencies]
-alloy-consensus.workspace = true
-alloy-primitives.workspace = true
+alloy = { workspace = true, features = ["consensus", "rlp", "rpc-types", "serde"] }
 alloy-rlp.workspace = true
-alloy-rpc-types.workspace = true
+alloy-rpc-types-eth = "0.4.2"
```
carver, line 16: Huh, so are the `eth_*` types included by default in the `alloy-rpc-types` crate, but not included in the `rpc-types` feature?

carver (approved):
Yeah, i very much support the direction. Let the alloy team figure out which versions are supposed to work in tandem :)

Weird that alloy-rlp needs to be explicitly included. This is one of those times where I miss the explicit imports of python. It still feels pretty haphazard for me to try to track down why symbols aren't available in rust. If you find out anything new about alloy-rlp, I'd be curious to hear an update!

trin-execution/Cargo.toml
morph-dev: Sadly, no.
The "rpc-types-engine" feature already enables "rpc-types", so adding it wouldn't solve the problem. The problem is that these features don't enable "serde" feature on "alloy-rpc-types" / "alloy-rpc-types-eth" crates.

ethportal-api/Cargo.toml
morph-dev: It's about "serde" feature of "alloy-rpc-types-eth" crate, which isn't enabled by following features: "rpc-types", "serde" (and I believe it should). This line solves it because that feature is enabled by default (but when imported via "alloy", they don't enable default features).

Cargo.toml
morph-dev: The latest version of that one is 0.3.8 (and other 0.4.2 crates do depend on this one, so no duplicates).

PR #1517, opened 2024-10-09
fix(portal-bridge): Temporarily disable `HistoricalSummariesWithProof` gossip

ogenev:
### What was wrong?
Requesting a beacon state over JSON causes the Pandaops nodes to run OOM.

### How was it fixed?
Disable temporary  `HistoricalSummariesWithProof` until we implement an option for ssz requests or a new debug endpoint is implemented in Nimbus, which will not require a download the full beacon state.

### To-Do


KolbyML (approved):
:shipit:

ogenev:
Rage merging with red CircleCI :)

PR #1518, opened 2024-10-10
fix(trin-beacon): fix light client finality update validation

ogenev:
### What was wrong?
We were incorrectly validating the `LightClientFinalityUpdate` content key against the value.

### How was it fixed?
Update the comparison because the portal beacon protocol is designed always to return the latest finality update with finalized slot that is bigger or equal to the requested content key.

### To-Do

### DIFF
--- a/trin-beacon/src/validation.rs
+++ b/trin-beacon/src/validation.rs
@@ -164,9 +164,9 @@ impl Validator<BeaconContentKey> for BeaconValidator {
                 // slot
                 let finalized_slot = lc_finality_update.get_finalized_slot();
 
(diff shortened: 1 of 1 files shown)


ogenev:
`peertest_history_offer_propagates_gossip_multiple_large_content_values` looks like a flaky test

morph-dev:
> `peertest_history_offer_propagates_gossip_multiple_large_content_values` looks like a flaky test

Indeed. I commented here, as I think this caused it to be flaky: https://github.com/ethereum/trin/pull/1514#discussion_r1794839654

PR #1519, opened 2024-10-10
fix: revert changed const in peertest utils

njgheorghita:
### What was wrong?
Reverts a change in #1514 that introduced a flaky test

### How was it fixed?
changed const

### To-Do


KolbyML (approved):
:shipit:

carver:
```
thread 'validation::tests::test_validate_light_client_finality_update' panicked at trin-beacon/src/validation.rs:452:14:
called `Result::unwrap_err()` on an `Ok` value: ValidationResult { valid_for_storing: true, additional_content_to_propagate: None }
```

PR #1520, opened 2024-10-10
fix: revert changed const in peertest utils and fix light client update validation test

ogenev:
### To-Do

### DIFF
--- a/trin-beacon/src/validation.rs
+++ b/trin-beacon/src/validation.rs
@@ -441,10 +441,11 @@ mod tests {
 
         assert!(result.valid_for_storing);
 
-        // Expect error because the finalized slot does not match the content key finalized slot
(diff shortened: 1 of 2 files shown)


trin-beacon/src/validation.rs
```diff
@@ -441,10 +441,11 @@ mod tests {
 
         assert!(result.valid_for_storing);
 
-        // Expect error because the finalized slot does not match the content key finalized slot
+        // Expect error because the content key finalized slot is greaten than the light client
+        // update  finalized slot
```
carver, line 445:
```suggestion
        // Expect error because the content key finalized slot is greater than the light client
        // update finalized slot
```

But I don't think it's worth kicking off the tests again, since I'm waiting to run the deploy until this PR gets into master.

PR #1521, opened 2024-10-11
feat: add support for eth_getBlockByNumber

morph-dev:
### What was wrong?

Trin currently doesn't support eth_getBlockByNumber or eth_call with block specified as number.

### How was it fixed?

Added support for `eth_*` JSON rpc request for fetching block by number.

### To-Do

### DIFF
--- a/rpc/src/eth_rpc.rs
+++ b/rpc/src/eth_rpc.rs
@@ -1,11 +1,11 @@
 use alloy::{
     primitives::{Address, Bytes, B256, U256},
-    rpc::types::{Block, BlockId, BlockTransactions, TransactionRequest},
+    rpc::types::{Block, BlockId, BlockNumberOrTag, BlockTransactions, TransactionRequest},
 };
 use ethportal_api::{
     jsonrpsee::types::{error::CALL_EXECUTION_FAILED_CODE, ErrorObjectOwned},
     types::{
-        execution::block_body::BlockBody,
+        execution::{block_body::BlockBody, transaction::Transaction},
         jsonrpc::{
             endpoints::HistoryEndpoint,
             request::{HistoryJsonRpcRequest, StateJsonRpcRequest},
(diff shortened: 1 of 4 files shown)


rpc/src/eth_rpc.rs
```diff
@@ -144,8 +127,20 @@ impl EthApiServer for EthApi {
         )
         .await
         .map_err(|err| RpcServeError::Message(err.to_string()))?;
-        let output = result_and_state.result.into_output().unwrap_or_default();
-        Ok(output)
```
carver, line 148: *nit* I think this change is clearer if it goes into its own commit. I was trying to figure out how it was related for a bit, until I went back to the PR summary to realize that it wasn't 😆 .

carver (approved):
It would be great to add at least one test to `tests/rpc_server.rs`.

If you feel like it, I think it would be cool to add a test that confirms that the "latest" tag fails and/or that hydrated transactions fail. It can act as a local reminder for whichever dev adds that feature, that they should flesh out that test.

I think the test is important, but I'm not asking to review it. Happy if you want me to, though!

rpc/src/eth_rpc.rs
```diff
@@ -180,6 +193,37 @@ impl EthApi {
...
+                "replying with all transaction bodies is not supported yet".to_string(),
+            ));
+        }
+
+        let body = self.fetch_block_body(header.hash()).await?;
+        let transactions = match &body {
+            BlockBody::Legacy(body) => &body.txs,
+            BlockBody::Merge(body) => &body.txs,
+            BlockBody::Shanghai(body) => &body.txs,
+        };
```
KolbyML, line 212: I just wanted to mention body has a getter for transactions, I think we could remove the Result and clones from the getter, but not a big deal either way

KolbyML (approved):
:shipit:

morph-dev:
> It would be great to add at least one test to `tests/rpc_server.rs`.
> 
> If you feel like it, I think it would be cool to add a test that confirms that the "latest" tag fails and/or that hydrated transactions fail. It can act as a local reminder for whichever dev adds that feature, that they should flesh out that test.
> 
> I think the test is important, but I'm not asking to review it. Happy if you want me to, though!

I added few tests. Would be good if you can take a look at it as well (as I refactored existing test as well).

morph-dev:
Rebased to master and I will merge it now.

@carver if you look at the tests and find comments, let me know and I will do a followup PR

PR #1522, opened 2024-10-12
feat: eth_call should return if execution fails

morph-dev:
### What was wrong?

The `eth_call` doesn't properly handle the case when execution returns the error. Currently, it only returns empty byte array.

### How was it fixed?

Transforms execution result into proper json response.

### To-Do


KolbyML (approved):
:shipit:

PR #1523, opened 2024-10-12
refactor: block body fn: transactions(), uncles() and withdrawals()

morph-dev:
### What was wrong?

Some `BlockBody`'s functions return `anyhow::Result` type when then don't need to, and they clone objects, instead of returning references.

### How was it fixed?

Changed function signature to return `&[T]` instead.

**Note:** I think that `calculate_merkle_patricia_root` (used `transactions_root` and `withdrawals_root`) should return `B256` (instead of `anyhow::Result<B256>`).

The api that we use (`eth-trie`) can fail, but it shouldn't in this case (we create new trie and use in-memory db), so I'm in favor of panicing instead of returning `Result`.

I didn't add it in this PR in order not to complicate things, but I'm willing to add it if that's direction we want to go with.

### To-Do

### DIFF
--- a/ethportal-api/src/types/execution/block_body.rs
+++ b/ethportal-api/src/types/execution/block_body.rs
@@ -1,10 +1,9 @@
 use std::vec;
 
-use alloy_primitives::B256;
+use alloy_primitives::{keccak256, B256};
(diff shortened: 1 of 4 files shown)


KolbyML (approved):
:shipit:

PR #1525, opened 2024-10-14
fix: set Enr sequence to current epoch timestamp

KolbyML:
### What was wrong?
last year I attempted to fix a bug where every time trin booted our sequence number was always default (0 or something) which was bad because that meant that peers would think our Enr we were sending  was outdated as they would have an enr of our node already with a higher sequence https://github.com/ethereum/trin/pull/937

The issue with this solution was sigp/discv5 would sometimes update the Enr and incremeant the sequence because this is done on the sigp/discv5 level we wouldn't save the sequence number to a file to restore on reboot so the same bug could occur which I tried to fix before.

sigp/discv5 sometimes updates the IP of the node which is good because most households have dynamic IP address or what if Trin is running on a laptop we need to be able to handle IP changes.
### How was it fixed?

By setting the sequence number to the current epoch timestamp on Trin's initalization, even if sigp/discv5 increments the sequence number of our Enr when we reboot Trin, we will set the sequence to the latest timestamp again, which will always be higher than what the sequence was before boot.

Hence the problem is fixed! I meant to make this PR months ago but I was distracted by higher-priority issues

### DIFF
--- a/ethportal-peertest/src/scenarios/basic.rs
+++ b/ethportal-peertest/src/scenarios/basic.rs
@@ -120,6 +120,7 @@ pub async fn test_lookup_enr(subnetwork: Subnetwork, peertest: &Peertest) {
 pub async fn test_ping(subnetwork: Subnetwork, target: &Client, peertest: &Peertest) {
     info!("Testing ping for {subnetwork}");
     let bootnode_enr = peertest.bootnode.enr.clone();
+    let bootnode_sequence = bootnode_enr.seq();
     let result = match subnetwork {
         Subnetwork::Beacon => BeaconNetworkApiClient::ping(target, bootnode_enr),
         Subnetwork::History => HistoryNetworkApiClient::ping(target, bootnode_enr),
@@ -132,7 +133,7 @@ pub async fn test_ping(subnetwork: Subnetwork, target: &Client, peertest: &Peert
         result.data_radius,
         U256::from_be_slice(Distance::MAX.as_ssz_bytes().as_slice())
     );
-    assert_eq!(result.enr_seq, 1);
+    assert_eq!(result.enr_seq, bootnode_sequence);
 }
 
 pub async fn test_ping_cross_network(mainnet_target: &Client, angelfood_node: &PeertestNode) {
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -1,9 +1,8 @@
 use std::{
-    fmt, fs,
+    fmt,
     hash::{Hash, Hasher},
     io,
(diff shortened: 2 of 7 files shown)


morph-dev:
I will let somebody else with better knowledge of discv5 protocol review this PR, but I just wanted to provide alternative approach.

If I'm not mistaken, `Discv5` will emit [`Event::SocketUpdated`](https://docs.rs/discv5/latest/discv5/enum.Event.html#variant.SocketUpdated) when enr is changed. We handle different types of events [here](https://github.com/ethereum/trin/blob/f4eb3e19cf827a83a61acc45722e83b2b7e83e9a/portalnet/src/discovery.rs#L228-L260), but we are ignoring this one. Can we just update enr that we store locally when this event happens?
Maybe also update it when `Discovery` structure is dropped?! (this alone isn't enough because of crashes and non-graceful shutdowns, but I don't see downsides to have it).

Again, I don't have extensive knowledge, so my idea might not be good / might not even work.

What I like about your approach is that we don't need extra file to store this information. But I'm not big fan of sequence number being set to timestamp.

KolbyML:
> I will let somebody else with better knowledge of discv5 protocol review this PR, but I just wanted to provide alternative approach.
> 
> 
> 
> If I'm not mistaken, `Discv5` will emit [`Event::SocketUpdated`](https://docs.rs/discv5/latest/discv5/enum.Event.html#variant.SocketUpdated) when enr is changed. We handle different types of events [here](https://github.com/ethereum/trin/blob/f4eb3e19cf827a83a61acc45722e83b2b7e83e9a/portalnet/src/discovery.rs#L228-L260), but we are ignoring this one. Can we just update enr that we store locally when this event happens?
> 
> Maybe also update it when `Discovery` structure is dropped?! (this alone isn't enough because of crashes and non-graceful shutdowns, but I don't see downsides to have it).
> 
> 
> 
> Again, I don't have extensive knowledge, so my idea might not be good / might not even work.
> 
> 
> 
> What I like about your approach is that we don't need extra file to store this information. But I'm not big fan of sequence number being set to timestamp.

Using the epoch timestamp is what Geth's implementation of Discv5 does. Second this solution has less code to maintain for the same result, so I don't see the need for storing a trin.enr when this is a cleaner/simpler solution

KolbyML:
> Only concern is that this will leave dangling enr files in all data dirs that precede this pr... I'm having trouble deciding if this is a problem worth implementing a check for... It could be nice to have a "temporary" util that checks for the enr file and deletes it if it's found, and then we can remove that util a couple releases down the road. But i'll leave that decision up to you

Me and Nick talked and we will remove the `trin.enr` files from our deployment script

PR #1526, opened 2024-10-14
refactor: update jsonrpc endpoint from *RecursiveFindContent to *GetContent

KolbyML:
### What was wrong?
updates our codebase to match spec https://github.com/ethereum/portal-network-specs/pull/344
### How was it fixed?

by renaming `portal_*RecursiveFindContent` to `portal_*GetContent`

### DIFF
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -78,16 +78,13 @@ pub trait BeaconNetworkApi {
         -> RpcResult<ContentInfo>;
 
     /// Lookup a target content key in the network
-    #[method(name = "beaconRecursiveFindContent")]
-    async fn recursive_find_content(&self, content_key: BeaconContentKey)
-        -> RpcResult<ContentInfo>;
+    #[method(name = "beaconGetContent")]
+    async fn get_content(&self, content_key: BeaconContentKey) -> RpcResult<ContentInfo>;
 
     /// Lookup a target content key in the network. Return tracing info.
-    #[method(name = "beaconTraceRecursiveFindContent")]
-    async fn trace_recursive_find_content(
-        &self,
-        content_key: BeaconContentKey,
-    ) -> RpcResult<TraceContentInfo>;
+    #[method(name = "beaconTraceGetContent")]
+    async fn trace_get_content(&self, content_key: BeaconContentKey)
+        -> RpcResult<TraceContentInfo>;
 
     /// Pagination of local content keys
     #[method(name = "beaconPaginateLocalContentKeys")]
--- a/ethportal-api/src/beacon.rs
+++ b/ethportal-api/src/beacon.rs
@@ -77,7 +77,8 @@ pub trait BeaconNetworkApi {
     async fn find_content(&self, enr: Enr, content_key: BeaconContentKey)
         -> RpcResult<ContentInfo>;
 
-    /// Lookup a target content key in the network
+    /// First checks local storage if content is not found lookup a target content key in the
+    /// network
     #[method(name = "beaconGetContent")]
     async fn get_content(&self, content_key: BeaconContentKey) -> RpcResult<ContentInfo>;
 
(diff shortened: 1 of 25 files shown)


ethportal-api/src/beacon.rs
```diff
@@ -78,16 +78,13 @@ pub trait BeaconNetworkApi {
         -> RpcResult<ContentInfo>;
 
     /// Lookup a target content key in the network
```
morph-dev, line 80: I think it would be good to clarify that it will first check local storage.
Also in other places.

ethportal-api/src/types/portal.rs
```diff
@@ -14,7 +14,7 @@ pub type RawContentValue = Bytes;
 pub type DataRadius = U256;
 pub type Distance = U256;
 
-/// Part of a TraceRecursiveFindContent response
+/// Part of a TraceGetContent response
 #[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct NodeInfo {
```
morph-dev, line 20: This object doesn't seem to be used at all.

ethportal-api/src/types/portal.rs
```diff
@@ -53,7 +53,7 @@ pub struct TraceGossipInfo {
     pub transferred: Vec<String>,
 }
 
-/// Response for FindContent & RecursiveFindContent endpoints
+/// Response for FindContent & GetContent endpoints
```
morph-dev, line 56: While it's true that this is the response type, RecursiveFindContent would only return `Content` variant (this is also according to spec). Should we create separate type for it?

I also believe that `ConnectionId` variant is never used (not even for `FindContent`). Should we remove it?

While it's not strictly related to this pr (renaming endpoint), I think that this is something that we should improve. If you feel like it, maybe do it in this PR, otherwise create an issue for it.

ethportal-peertest/src/scenarios/gossip.rs
```diff
@@ -299,7 +299,7 @@ pub async fn test_gossip_dropped_with_find_content(peertest: &Peertest, target:
     // send find_content request from fresh target to target
```
morph-dev, line 299: nit: Unrelated to this pr, but this comment is wrong. Or at least, it looks like it was correct, but than line 301 was  commented out and line 302 is added. Would be nice to fix it if you know what is happening here.

morph-dev (approved):
:ship: 

I general it looks good to me. Couple of comments that maybe don't even belong to this PR, but I wanted to start discussion (if they are small enough, I wouldn't be against including them into this PR).

Since this is causing some API changes, please wait for at least one more approval.

ethportal-api/src/types/portal.rs
KolbyML: I will make a follow up PR to resolve this

ethportal-peertest/src/scenarios/gossip.rs
KolbyML: I removed it, that seems like an artifact left when writing the test

morph-dev: The comment should still be updated to

ethportal-api/src/beacon.rs
```diff
@@ -77,17 +77,15 @@ pub trait BeaconNetworkApi {
...
-    /// Lookup a target content key in the network
-    #[method(name = "beaconRecursiveFindContent")]
-    async fn recursive_find_content(&self, content_key: BeaconContentKey)
-        -> RpcResult<ContentInfo>;
+    /// First checks local storage if content is not found lookup a target content key in the
+    /// network
+    #[method(name = "beaconGetContent")]
+    async fn get_content(&self, content_key: BeaconContentKey) -> RpcResult<ContentInfo>;
 
     /// Lookup a target content key in the network. Return tracing info.
```
morph-dev, line 85: nit: I think you forgot to update this doc (you did it in other files)

morph-dev (approved):

njgheorghita (approved):
:shipit:

PR #1527, opened 2024-10-14
fix: add uncles and withdrawals to eth_getBlockByNumber and eth_getBlockByHash

morph-dev:
### What was wrong?

The eth_getBlockByNumber and eth_getBlockByHash don't populate uncles and withdrawals fields of the response (but we have data available).

### How was it fixed?

Populate required fields.

**Note:** I don't know how `size` field should be calculated. Let me know if you do.

### To-Do

### DIFF
--- a/ethportal-api/src/types/execution/header.rs
+++ b/ethportal-api/src/types/execution/header.rs
@@ -277,7 +277,7 @@ impl From<Header> for RpcHeader {
             excess_blob_gas: excess_blob_gas.map(|v| v.to()),
             hash,
             parent_beacon_block_root: parent_beacon_block_root.map(|h264| h264.0.into()),
-            total_difficulty: Some(difficulty),
+            total_difficulty: None,
             requests_root: None,
         }
     }
--- a/ethportal-api/src/types/execution/header.rs
+++ b/ethportal-api/src/types/execution/header.rs
@@ -277,6 +277,7 @@ impl From<Header> for RpcHeader {
             excess_blob_gas: excess_blob_gas.map(|v| v.to()),
             hash,
             parent_beacon_block_root: parent_beacon_block_root.map(|h264| h264.0.into()),
+            // We don't have access to total_difficulty
             total_difficulty: None,
             requests_root: None,
         }
--- a/tests/rpc_server.rs
+++ b/tests/rpc_server.rs
@@ -4,6 +4,7 @@ use std::fs;
 use std::net::{IpAddr, Ipv4Addr};
 
 use alloy::{
+    primitives::U256,
     providers::{IpcConnect, Provider, ProviderBuilder, RootProvider},
     pubsub::PubSubFrontend,
     rpc::types::{BlockNumberOrTag, BlockTransactions, BlockTransactionsKind, Header as RpcHeader},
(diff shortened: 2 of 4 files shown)


KolbyML:
> Note: I don't know how size field should be calculated. Let me know if you do.

https://github.com/ethereum/go-ethereum/blob/add5709cb5a02183df8541505fc8e0f8c569faea/core/types/block.go#L429-L439

So I look at Geth and apparently, the size is the `len(rlp(header, uncles, transactions, withdrawals))`

I think it makes sense to include the size in this PR as it is supposed to be a required field of this endpoint

morph-dev:
Added `size` as well.
Also removed setting `total_difficulty` in Header as current logic is not correct (and I believe we don't have this info available).
Ready for another view.

ethportal-api/src/types/execution/header.rs
```diff
@@ -277,7 +277,7 @@ impl From<Header> for RpcHeader {
             excess_blob_gas: excess_blob_gas.map(|v| v.to()),
             hash,
             parent_beacon_block_root: parent_beacon_block_root.map(|h264| h264.0.into()),
-            total_difficulty: Some(difficulty),
+            total_difficulty: None,
```
njgheorghita, line 280: Could you just add a comment here explaining why this is set to `None`

tests/rpc_server.rs
```diff
@@ -111,7 +112,7 @@ async fn test_eth_get_block_by_number() {
         .expect("specified block not found");
 
     assert_header(&block.header, &hwp.header);
-    assert_eq!(block.size, None);
+    assert_eq!(block.size, Some(U256::from(37729)));
```
njgheorghita, line 115: Hmmm... this is for block `17510000`? Seems like there's a different size value on etherscan: https://etherscan.io/block/17510000

carver (approved):
LGTM, minus the size bug

ethportal-api/src/types/execution/header.rs
morph-dev: I think they are still part of the accumulators (based on quick look [here](https://github.com/ethereum/trin/blob/4692e6bde873243884f3820a8b417616efca26c9/trin-validation/src/accumulator.rs)), but they are not part of the content value object.

So yes, you are right that we don't have access to them.
If block is post-merge, we could set it as value is now always the same (basically have hardcoded value), but since we don't have it pre-merge, I think it's better to just say it's unsupported.

tests/rpc_server.rs
morph-dev: After couple hours of debugging I figured out the problem.

The Transactions should be encoded with envelope (additional rlp header for non-legacy transactions). As far as I can tell, there is no easy way to do so with current code, so I'm leaving `size` as `None` for now (but I added comment for future fix).

KolbyML (approved):
> After couple hours of debugging I figured out the problem.
> 
> The Transactions should be encoded with envelope (additional rlp header for non-legacy transactions). As far as I can tell, there is no easy way to do so with current code, so I'm leaving `size` as `None` for now (but I added comment for future fix).

I wrote code to handle this so I will make a PR to add size I guess and we can merge this now

njgheorghita (approved):

PR #1528, opened 2024-10-14
refactor! clean up ContentInfo type, remove unexpected paths

KolbyML:
### What was wrong?
fixes https://github.com/ethereum/trin/pull/1526#discussion_r1799174704

Basically the `ContentInfo` type we were using for our `FindContent` and `GetContent` RPC endpoints were leaking information implementation details not related to the endpoints
### How was it fixed?
Split `ContentInfo` into 
- `FindContentInfo`
- `GetContentInfo`

and only expose the datatypes specified in the `portal-network-specs` json-rpc specification

### DIFF
--- a/ethportal-api/src/types/portal.rs
+++ b/ethportal-api/src/types/portal.rs
@@ -49,9 +49,7 @@ pub struct TraceGossipInfo {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
 #[allow(clippy::large_enum_variant)]
-pub enum ContentInfo {
-    #[serde(rename_all = "camelCase")]
-    ConnectionId { connection_id: u16 },
+pub enum FindContentInfo {
     #[serde(rename_all = "camelCase")]
     Content {
         content: RawContentValue,
@@ -61,8 +59,15 @@ pub enum ContentInfo {
     Enrs { enrs: Vec<Enr> },
 }
 
-/// Parsed response for TraceGetContent endpoint
-///
+/// Response for GetContent endpoints
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetContentInfo {
+    pub content: RawContentValue,
+    pub utp_transfer: bool,
+}
+
+/// Parsed response for TraceRecursiveFindContent endpoint
 /// This struct represents the content info, and is only used
 /// when the content is found locally or on the network.
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
--- a/ethportal-api/src/types/portal.rs
+++ b/ethportal-api/src/types/portal.rs
@@ -45,10 +45,9 @@ pub struct TraceGossipInfo {
     pub transferred: Vec<String>,
 }
 
-/// Response for FindContent & GetContent endpoints
+/// Response for the FindContent endpoint
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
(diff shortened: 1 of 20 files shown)


morph-dev:
Since this is somewhat related to changes in #1526, I will review if once that one is merged.

Alternatively, if you rebase this one on top of #1526 (as separate commit), I can take a look.

KolbyML:
> Since this is somewhat related to changes in #1526, I will review if once that one is merged.
> 
> Alternatively, if you rebase this one on top of #1526 (as separate commit), I can take a look.

rebased

carver:
*nit* since it's a breaking change, it should use a `!` instead of `:` to match the conventional commit standard.

Also, it would be good to get that breaking change updated in glados as soon as possible (and Hive, but I assume that was already on your mind).

src/bin/poll_latest.rs
morph-dev, line 1: I think that many of changes in this file are not needed. While it does preserve original functionality, I think original implementation was wrong (because it wasn't handling errors correctly).

Meaning, if `GetContent` fails (content not available) it should keep retrying, right?
So `timeout` and `backoff` should still be used.

In my opinion, in `audit_content_key` we should have something like:
```rust
while Instant::now() - timestamp < timeout {
    match client.get_content(content_key.clone()).await {
        Ok(_) => return Ok(Instant::now()),
        _ => {
            attempts += 1;
            let sleep_time = match backoff {
                Backoff::Exponential => attempts * 2,
                Backoff::Linear(delay) => delay,
            };
            sleep(Duration::from_secs(sleep_time)).await;
        }
    }
}
```
which is closer to the original implementation.

---

Otherwise, if you think new implementation is correct, than you should cleanup more of this file (remove Backoff, update documentation on how to run it, etc).

morph-dev (commented):
Looks mostly ok, except that changes in `src/bin/poll_latest.rs`. See comment there.

src/bin/poll_latest.rs
KolbyML: The new implementation is the same as the old implementation, but I agree that your example is most likely what was originally intended

portal-bridge/src/bridge/era1.rs
```diff
@@ -368,7 +368,7 @@ impl Era1Bridge {
             let header_hash = block_tuple.header.header.hash();
             let header_content_key = HistoryContentKey::new_block_header_by_hash(header_hash);
             let header_content_info = portal_client.get_content(header_content_key.clone()).await;
-            if let Ok(ContentInfo::Content { .. }) = header_content_info {
+            if let Ok(GetContentInfo { .. }) = header_content_info {
```
morph-dev, line 371: nit: this could also be `header_content_info.is_ok()`.

Also in 3 more places in this file below.

ethportal-api/src/types/portal.rs
```diff
@@ -49,9 +49,7 @@ pub struct TraceGossipInfo {
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 #[serde(untagged)]
 #[allow(clippy::large_enum_variant)]
-pub enum ContentInfo {
-    #[serde(rename_all = "camelCase")]
-    ConnectionId { connection_id: u16 },
+pub enum FindContentInfo {
```
morph-dev, line 52: The documentation is still not updated.

morph-dev (approved):

carver:
> _nit_ since it's a breaking change, it should use a `!` instead of `:` to match the conventional commit standard.

BTW, sorry I misspoke about the conventional commits standard. For next time: it is supposed to be to *add* a `!` before the `:`. Like: `refactor!: clean up`...

PR #1529, opened 2024-10-14
feat: shorter utp timeout, for more reliable query

carver:
Shrink the idle timeout from the default of 60 seconds to 10. This comes up most often in failed connections, which are perpetually idle, and keep trying until the timeout expires.

During a find-content call, if a uTP transfer is initiated and the remote peer fails to connect, the default timeout waits for 60 seconds. The problem is that 60s is also our entire query timeout. So if the utp connection fails by timeout, then there is no time to attempt downloads from alternate peers.

By shrinking the timeout to 10 seconds, we leave plenty of room for contacting other peers for the same content. It still allows us to make a few attempts at connection (after 1s, then another 1.5s, then another 2.25s, then another 3.375s).

### Historical Context

This timeout was previously 32 seconds, then increased to 60s: https://github.com/ethereum/trin/pull/909

The reason I feel differently now from before is that reattempts make it actually *more* likely for success, when shrinking the timeout. Before reattempts were implemented, we were only decreasing our chance of success by shrinking the timeout.

### To-Do

### DIFF
--- a/portalnet/src/utp_controller.rs
+++ b/portalnet/src/utp_controller.rs
@@ -1,7 +1,7 @@
 use crate::discovery::UtpEnr;
 use anyhow::anyhow;
 use lazy_static::lazy_static;
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
 use tokio::sync::{OwnedSemaphorePermit, Semaphore};
 use tracing::debug;
 use trin_metrics::{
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

carver:
Hm, does this change make the `peertest_offer_concurrent_utp_transfer_limit` test flaky? Investigating...

carver:
Yes, I believe the stress test in CI is showing that 5 seconds of waiting for connections is not enough sometimes (it failed 1/3 of the time that CI was run).  I'm doubling it to 10, guessing that that should be enough. 🤷🏻‍♂️ 

I think this choice of 10s is more about passing the test than picking the right production value. Unfortunately, the timeout is set in a static variable, so it's not currently easy to change only during testing. But even though CI seems unrealistically bad performance-wise at times, I still like the premise of testing with the same configs that run in production.

carver:
Ok, I ran CI a couple times without any failures. I'll keep an eye on it, but think that should be sufficient for now.

PR #1532, opened 2024-10-16
test: ignore flaky utp concurrency limit test

carver:
### What was wrong?

Master turned red, `peertest_offer_concurrent_utp_transfer_limit` is failing sporadically. #1529 seemed to uncover an issue of flakiness in the `peertest_offer_concurrent_utp_transfer_limit` test.

### How was it fixed?

Ignore the test, until we can identify the problem with the test. This should get master reliably passing CI again.

Also had to remove the `#[serial]` attribute.

### To-Do

### DIFF
--- a/tests/self_peertest.rs
+++ b/tests/self_peertest.rs
@@ -81,7 +81,7 @@ async fn peertest_offer_with_trace() {
 }
 
 #[tokio::test(flavor = "multi_thread")]
-#[serial]
+#[ignore = "test is flaky: fails in some environments and in CI sporadically. Re-add #[serial] when re-enabling"]
 async fn peertest_offer_concurrent_utp_transfer_limit() {
     let (peertest, target, handle) = setup_peertest_bridge(&[Subnetwork::History]).await;
     peertest::scenarios::offer_accept::test_offer_concurrent_utp_transfer_limit(&peertest, target)


carver:
Hm, the ignore attribute isn't ignoring on my local tests...

carver:
Ok, serial was interfering. Removed it.

KolbyML (approved):
:shipit: it would be cool if githubs `squash and merge` feature would be used to merge the PR. As it leaves a cleaner master commit history which is useful for tracking if changes broke something or seeing when changes happened.

Using `Create a Merge commit` breaks master commit history. It causes the commits merged to be put into the commit history when they all happened which means if a PR has 5 commits all made on 5 different days they would be in the commit history on those commit dates. This makes the history hard to track as commits from 1 PR will be scattered everywhere. For example the commits which were merged with `Create a Merge commit` will be between other Merged PR's in the history which is very confusing.

It seems like most contributors to Trin already use `Squash and Merge` and other projects like `geth`, `reth`, and `hive` all exclusively use `Squash and Merge` I think we should standardize the practice for `Trin` too! Having a cleaner history on Master would be very nice!

carver:
When it's reasonable, I can push the commit to master.

I disagree with a lot of the premise, though, especially:

> Using `Create a Merge commit` breaks master commit history.

I see no reasonable way to call it broken. Merging is the way of git.

> This makes the history hard to track as commits from 1 PR will be scattered everywhere.

Have you tried `git log --graph`?

I think it's really valuable to have the commits broken into logical chunks, for understanding the history of why a piece of code was changed during a git blame or bisect.

For single commit changes that are on top of head, it seems reasonable to just push them to master. Though github's interface doesn't include that option, so I have to drop to CLI to do it. There was one other annoying thing about that, but I can't remember this moment. Maybe github doesn't support the 'click to delete branch' option when you do it that way?

KolbyML:
> When it's reasonable, I can push the commit to master.
> 
> I disagree with a lot of the premise, though, especially:
> 
> > Using `Create a Merge commit` breaks master commit history.
> 
> I see no reasonable way to call it broken. Merging is the way of git.

Github PR's aren't git native. Real git projects like the linux kernal submit diffs. This whole PR process was created by Github in 2008. Before 2008 all git projects were done through submitting diffs. So I don't see `I see no reasonable way to call it broken. Merging is the way of git.` as a fair statement as we are using a special mechanism which isn't apart of git for managing proposed code changes.

> > This makes the history hard to track as commits from 1 PR will be scattered everywhere.
> 
> Have you tried `git log --graph`?

It breaks history on Github. It creates an under ordered commit history on master as in. 

- If I merge PR 1 first
- Then someone merges PR 2 second, with `Create a Merge commit`, but the commit date was first

Then PR 2 commit will before PR 1 on the commit history. Now imagine PR 2 has multiple commits if you want to see which PR broken master or added a deadlock it is then a lot harder because it isn't easy to revert the PR, you must revert multiple commits from the PR which have dependencies before and after the commit multiple times for all the commits of that PR.

If all PR's are merged via `Squash and Merge` this isn't a problem.

`Create a Merge commit` creates an un-ordered merge history. I think when the PR is merged matters more then when the commits were committed for Master's commit history

> I think it's really valuable to have the commits broken into logical chunks, for understanding the history of why a piece of code was changed during a git blame or bisect.

PR's can be viewed for that if someone is looking through the commits on github they can their is always a link to the PR

image
at the bottom left

> For single commit changes that are on top of head, it seems reasonable to just push them to master. Though github's interface doesn't include that option, so I have to drop to CLI to do it. But I can live with that.

 yes it does it is called `rebase and merge` and this retains PR commit ordering unlike `Create a merge commit` which breaks it

carver:
I think we've gone past this conversation being productive over comments. Let's talk about it at the end of the next call.

PR #1535, opened 2024-10-18
docs: examples for eth_* RPC calls

carver:
### What was wrong?

When trying to use `eth_blockNumber` as recommended in the docs, it failed. It probably should work, especially when enabling beacon, but that's a problem for another day.

### How was it fixed?

- Replace eth_blockNumber with eth_getBlockByNumber example
- Add an eth_call example to show off state access
- Remind people why they want to use a web3-like tool

### To-Do


KolbyML (approved):
:shipit:

PR #1534, opened 2024-10-17
fix: add size to eth_getBlockByNumber and eth_getBlockByHash

KolbyML:
This PR should be reviewed after https://github.com/ethereum/trin/pull/1527

### What was wrong?
followup PR to https://github.com/ethereum/trin/pull/1527 which adds size
### How was it fixed?
properly handle rlp encoding for the block to get the size

PR #1536, opened 2024-10-18
refactor: transaction rlp encoding

morph-dev:
### What was wrong?

Transaction can be rlp encoded/decoded with (less common) or without (most common) additional rlp header.

Currently, we have to deal with less common case manually, which isn't great. As of now, we have two use cases for it: inside `BlockBody` and as block size inside `eth_getBlockBy*` calls (which is not implemented).

### How was it fixed?

Added const generic argument to the `Transaction` struct.

The default value represents the most common case, which means that we don't have to specify it in most of the cases.

### To-Do

### DIFF
--- a/ethportal-api/src/types/execution/block_body.rs
+++ b/ethportal-api/src/types/execution/block_body.rs
@@ -121,8 +121,6 @@ impl BlockBody {
     }
 
     /// Returns reference to uncle headers.
-    ///
-    /// Returns None post Merge fork.
     pub fn uncles(&self) -> &[Header] {
         match self {
             BlockBody::Legacy(body) => &body.uncles,
@@ -191,7 +189,7 @@ impl Decodable for BlockBodyLegacy {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         let mut payload_view = rlp::Header::decode_bytes(buf, true)?;
         let txs = rlp_decode_transaction_list_with_header(&mut payload_view)?;
-        let uncles = Vec::<Header>::decode(buf)?;
+        let uncles = Vec::<Header>::decode(&mut payload_view)?;
         Ok(Self { txs, uncles })
     }
 }
@@ -353,7 +351,7 @@ impl Decodable for BlockBodyShanghai {
     fn decode(buf: &mut &[u8]) -> alloy_rlp::Result<Self> {
         let mut payload_view = rlp::Header::decode_bytes(buf, true)?;
         let txs = rlp_decode_transaction_list_with_header(&mut payload_view)?;
-        let withdrawals = Vec::<Withdrawal>::decode(buf)?;
+        let withdrawals = Vec::<Withdrawal>::decode(&mut payload_view)?;
         Ok(Self { txs, withdrawals })
     }
 }
--- a/ethportal-api/src/types/execution/block_body.rs
+++ b/ethportal-api/src/types/execution/block_body.rs
@@ -2,7 +2,7 @@ use std::vec;
 
 use alloy::{
     primitives::{keccak256, B256},
-    rlp::{Decodable, Encodable, Error as RlpError, Header as RlpHeader},
+    rlp::{self, Decodable, Encodable},
 };
 use anyhow::{anyhow, bail};
 use serde::Deserialize;
@@ -45,7 +45,7 @@ impl Decodable for BlockBody {
         } else if let Ok(val) = BlockBodyShanghai::decode(buf) {
             Ok(BlockBody::Shanghai(val))
         } else {
-            Err(RlpError::Custom("Invalid block body rlp"))
+            Err(rlp::Error::Custom("Invalid block body rlp"))
         }
     }
 }
@@ -123,17 +123,16 @@ impl BlockBody {
     /// Returns reference to uncle headers.
     ///
     /// Returns None post Merge fork.
-    pub fn uncles(&self) -> Option<&[Header]> {
+    pub fn uncles(&self) -> &[Header] {
         match self {
-            BlockBody::Legacy(body) => Some(&body.uncles),
-            BlockBody::Merge(_) => None,
-            BlockBody::Shanghai(_) => None,
+            BlockBody::Legacy(body) => &body.uncles,
+            BlockBody::Merge(_) => &[],
+            BlockBody::Shanghai(_) => &[],
         }
     }
 
     pub fn uncles_root(&self) -> B256 {
-        let uncles = self.uncles().unwrap_or_default().to_vec();
-        let encoded_uncles = alloy_rlp::encode(uncles);
+        let encoded_uncles = alloy_rlp::encode(self.uncles().to_vec());
         keccak256(&encoded_uncles)
     }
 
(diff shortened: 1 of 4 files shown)


KolbyML:
> I'm not extremely happy with this approach, as it complicates the Transaction object, which is very commonly used (even tho complexity is mostly hidden away).
> 
> The only other reasonable approach that I see is to have another type (e.g. `TransactionWithRlpHeader`, up for a better name suggestion) that would be a wrapper around existing `Transaction` with its own `Encodable`/`Decodable` implementation.

Couldn't we implement alloy_rlp length trait with the pre-existing code as well. If neither of the proposed refactors are a "happy approaches" we can just implement the length trait and call it a day no? which would also be much similar than the proposed changes.

I think the manual rlp calls instead of deriving them is fine because as you said the "manual" path is hardly used. So is it worth the complexity to implement auto derives for a rare usecase in the grand scheme of things.

carver:
Oh yeah this is starting to come back to me, it's been about 3 years!

Since you're looking for alternatives, I'll share what we did in py-evm. We found it helpful to have a unified pre-RLP-encoded representation. So that means the primitives of byte-strings and lists (er, vectors), which can contain more byte-strings and vectors, no matter the variant of transaction. When we correctly serialize into this format, then the RLP library should handle the actual encoding for us (by prepending the length of the payload recursively).

So the [approach we took in py-evm](https://github.com/ethereum/py-evm/blob/f2219d6df83ede8f9574eab48cbe4bc33197066f/eth/vm/forks/berlin/transactions.py#L489) was to offer two separate methods on the global transaction type:
- `serialize()`: generate primitives that are suitable for RLP encoding
- `encode()`: generate a self-contained byte-string representing a single transaction, useful for saving to a database, etc

For a legacy transaction, that looks like:
- serialize: return a vector of byte-strings, with an entry for each field of the transaction
- encode: `rlp.encode(self.serialize())`

For a typed transaction, split it into two layers, a wrapping layer that identifies the transaction type and a payload layer that is fully defined by that type ID. The payload layer works effectively the same as a legacy transaction, for serialization.

In the wrapping layer:
- serialize: return the `self.encode()` value as a byte-string at the root level, with no vector
- encode: prepend the type byte to the output of `payload.encode()`

Then the standard RLP encoding should work on a list of transactions that have each been run through `transaction.serialize()`. I haven't messed around with the alloy::rlp library much yet, but I expect it to be straightforward once you are supplying RLP-primitive values.

carver:
*nit* since this is adding a new ability in the rpc, I think naming commit something like this would be clearer:
`feat: add size to block returned in eth_getBlock*`

carver:
> Since you're looking for alternatives, I'll share what we did in py-evm.

Hm, the longer I look at this, the more I think it's potentially an orthogonal problem. I guess it's possible we never dealt with this funny case just when calculating block size, where just the typed transactions are double-rlp-encoded (it doesn't seem familiar to me). So I guess if I don't have any better ideas. 🤷🏻‍♂️

KolbyML:
https://docs.rs/alloy-rlp/latest/src/alloy_rlp/encode.rs.html#15-30

image

This PR doesn't provide a performance important over https://github.com/ethereum/trin/pull/1534 as both PR's fully do the RLP encode to get the length. So that are equivalent in that regard.

@carver https://github.com/ethereum/trin/pull/1534#issuecomment-2423432588

So this PR is purely a stylistic refactor. I am fine closing my PR, but I just wanted to make it clear that there isn't really a performance difference here.

KolbyML:
> > Since you're looking for alternatives, I'll share what we did in py-evm.
> 
> Hm, the longer I look at this, the more I think it's potentially an orthogonal problem. I guess it's possible we never dealt with this funny case just when calculating block size, where just the typed transactions are double-rlp-encoded (it doesn't seem familiar to me). So I guess if I don't have any better ideas. 🤷🏻‍♂️
> 
> I guess it's somewhat related how we have to add an extra rlp-encoding to legacy transactions when [calculating the transaction root](https://github.com/ethereum/trinity/blob/6383280c5044feb06695ac2f7bc1100b7bcf4fe0/trinity/_utils/trie.py#L19-L23).

This 

> orthogonal

We also don't need a refactor to calculate block size as shown here https://github.com/ethereum/trin/pull/1534 . If we want the code to fit a certain style that is fine, but it isn't like we are solving novel problems here

KolbyML:
Nvm I think I am wrong about the performance claims I made as I think after reading the implementation for the `RlpEncodable` macro it will implement the `Encodable` traits length() function at compile time with a static length. My bad.

So with this refactor it should use this pre-computed length.

KolbyML:
I gave this PR a brief review. We should implement `length()` for the `header` as currently that will still use `encode()` to get the length as we don't use `RlpEncodable` macro for `header`.

I will give a more thorough once all the tests are passing. Sorry about the confusion in my earlier messages.

ethportal-api/src/types/execution/block_body.rs
```diff
@@ -123,17 +123,16 @@ impl BlockBody {
     /// Returns reference to uncle headers.
     ///
     /// Returns None post Merge fork.
```
KolbyML, line 125:
```suggestion
    /// Returns None post Merge fork.
```

this doc is no longer true

rpc/src/eth_rpc.rs
```diff
@@ -210,27 +211,36 @@ impl EthApi {
...
             .map(|withdrawals| withdrawals.iter().map(Withdrawal::from).collect());
 
-        // TODO: Add calculation for the block's size:
+        // Calculate block size:
         //   len(rlp(header, transactions, uncles, withdrawals))
-        // NOTE: Transactions should be encoded with envelope
-        let size = None;
+        // Note: transactions are encoded with header
+        let size = {
+            let payload_size = header.length()
```
KolbyML, line 223:
```suggestion
            let payload_size = header.length()
```

Our Header type here doesn't use the macro or implement length in its implementation of `Encodable` so we are stilling encoding to find the length of `Header`. Would you want to implement `length()` for header in this PR or make an issue for it any we can solve it in another PR.


This would be the only place in calculating size where we still are encoding to get the length I believe.

KolbyML (approved):
:shipit: PR looks good. I think this solution is fine as I would want to avoid a wrapper type as it just seems kinda feels off to me.

rpc/src/eth_rpc.rs
morph-dev: I will do it in a separate PR (no need to open issue about it, I will do it right now).

morph-dev:
@KolbyML @carver I created another PR (very similar to this one) where I used separate type that is just a wrapper around `Transaction` with different RLP encoding.

Now that I see both of them, I like that one a bit more because it doesn't complicate the `Transaction` object that is very frequently used. But I can be convinced otherwise. So let me know what you think about both of them.

KolbyML:
> @KolbyML @carver I created another PR (very similar to this one) where I used separate type that is just a wrapper around `Transaction` with different RLP encoding.
> 
> Now that I see both of them, I like that one a bit more because it doesn't complicate the `Transaction` object that is very frequently used. But I can be convinced otherwise. So let me know what you think about both of them.

I like it more too. I thought it would look weirder in my head.

carver (changes requested):
Prefer the other approach

morph-dev:
Closing in favor of #1539

PR #1538, opened 2024-10-19
feat: set default storage size to 1GB

carver:
1GB seems like a reasonable default number. Big enough to help the network, small enough to not be a nuisance on most setups.

---
For context: this PR used to be about turning on state by default. Now it's just about the storage size default.

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -18,7 +18,7 @@ pub const DEFAULT_DISCOVERY_PORT: u16 = 9009;
 pub const DEFAULT_UTP_TRANSFER_LIMIT: usize = 50;
 const DEFAULT_SUBNETWORKS: &str = "history";
 pub const DEFAULT_NETWORK: &str = "mainnet";
-pub const DEFAULT_STORAGE_CAPACITY_MB: &str = "100";
+pub const DEFAULT_STORAGE_CAPACITY_MB: &str = "1000";
 pub const DEFAULT_WEB3_TRANSPORT: &str = "ipc";
 
 use crate::dashboard::grafana::{GrafanaAPI, DASHBOARD_TEMPLATES};


KolbyML (approved):
I think a storage increase would be fine

carver:
I think @mrferris felt the strongest about keeping storage size small, so wanted to invite you to the conversation.

morph-dev:
As I mentioned in the discord, I don't think that we should enable state as default just yet.

Main reasons:
- the actual content on the state subnetwork is not that useful
- we split dedicated storage to 1% history - 99% state, which doesn't reflect current situation

I wouldn't be against it if at least one of the two things is resolved, meaning:
- we start gossiping head state
- we refactor our code such that state and history use the same radius (my understanding is that beacon doesn't use radius)

KolbyML:
> * we refactor our code such that state and history use the same radius (my understanding is that beacon doesn't use radius)

Why would we want to use the same radius for both networks? Don't the overlay services store the radius's independently. What value does the state and history network using the same radius provide? Would this replace `we split dedicated storage to 1% history - 99% state, which doesn't reflect current situation` this scheme?

KolbyML:
> we split dedicated storage to 1% history - 99% state, which doesn't reflect current situation
 
 Maybe a dynamic system would work then?

morph-dev:
> What value does the state and history network using the same radius provide?

It means that we would have unified radius for our node. Meaning, we would store the same percentage of each subnetwork.
So as we start backfilling state data, percentage wise, more of it will be stored locally.

> Would this replace `we split dedicated storage to 1% history - 99% state, which doesn't reflect current situation` this scheme?

Yes, it would replace this. This hardcoded split was never intended as the long term solution. Instead, we want dynamic version of it, which is unified radius.

Alternative is that storage capacity is configured separately per subnetwork. But my understanding is that we don't want this because it's very bad user experience.

morph-dev:
> > we split dedicated storage to 1% history - 99% state, which doesn't reflect current situation
> 
> Maybe a dynamic system would work then?

Yes. And unified radius is the dynamic solution. Or do you have an idea for another one?

KolbyML:
> > What value does the state and history network using the same radius provide?
> 
> It means that we would have unified radius for our node. Meaning, we would store the same percentage of each subnetwork. So as we start backfilling state data, percentage wise, more of it will be stored locally.

I am still a little confused and some of the details. Like why would this scheme cause more state to be stored. But I agree a dynamic solution is better then a static one so that would be cool if it worked out

KolbyML:
`we would store the same percentage of each subnetwork` how can this be true if state is supposed to store more?

morph-dev:
> I am still a little confused and some of the details. Like why would this scheme cause more state to be stored. But I agree a dynamic solution is better then a static one so that would be cool if it worked out

> `we would store the same percentage of each subnetwork` how can this be true if state is supposed to store more?

Maybe we are confusing terminology. I will give an example:

Let's say storage capacity is 5GB

1. And let's say that currently entire history subnetwork is 500GB, and entire state subnetwork is 10GB
    - unified radius would be ~1% -> we store 1% of each subnetwork -> ~1GB history and ~100MB state (at bit less)
2. Let's assume we start gossip a lot more state nework, so the entire state network is also 500GB
    - radius would now be 0.5% -> we would store ~2500MB history and ~2500MB state
3. We keep gossiping state and now entire state network is 4500GB
    - radius would be 0.1% -> we would store 500MB history and 4500MB state

So even tho radius went down, we are storing more state and less history content locally. They are always the same percentage of the entire subnetwork, but local split changes because they grow with different speed.

KolbyML:
> Maybe we are confusing terminology. I will give an example:
> 
> Let's say local storage capacity is 5GB
> 
> 1. And let's say that currently entire history subnetwork is 500GB, and entire state subnetwork is 10GB
>    
>    * unified radius would be ~1% -> we store 1% of each subnetwork -> ~4900MB history and ~100MB state
> 2. Let's assume we start gossip a lot more state nework, so the entire state network is also 500GB
>    
>    * radius would now be 0.5% -> we would store ~2500MB history and ~2500MB state
> 3. We keep gossiping state and now entire state network is 4500GB
>    
>    * radius would be 0.1% -> we would store 500MB history and 4500MB state
> 
> So even tho radius went down, we are storing more state and less history content locally. They are always the same percentage of the entire subnetwork, but local split changes because they grow with different speed.


There isn't a way for a node to know how much total storage each network takes. So would it be determined by how many offers you get per a network?

State content is much smaller so you would likely get more offers as there is a higher likelihood the content is within your radius.

morph-dev:
> There isn't a way for a node to know how much total storage each network takes. So would it be determined by how many offers you get per a network?

If you fix local storage capacity and force radius to be the same, you are going to get this behavior.
You don't need to "know" total storage per network and you don't need to monitor number of offers, how much local storage you are using per network or do anything special.

The tricky part (with current design) is how to keep radius the same and how to do "incremental" pruning.

This is assuming you stay active/online for a long time in order to get real view of the network (meaning, there is no content that falls within your radius that you don't have/was never offered to you).

> Maybe this would be easier to go over in a call or something

Sure, let's have a call

carver:
> I wouldn't be against it if at least one of the two things is resolved, meaning:
> 
>     * we start gossiping head state

Is it head in particular, or you just mean to deposit a lot more data? Let's say we start gossiping state at block 19M or 20M forward, and if we don't catch up to head quickly. Would you be happy that we are starting to move toward the natural balance point of state network being much heavier than history?

> 
>     * we refactor our code such that state and history use the same radius (my understanding is that beacon doesn't use radius)

I haven't dug in, but I'm guessing it will be a significant refactor to support this. (But I agree it's the best long-term solution we've identified)

KolbyML:
> > I wouldn't be against it if at least one of the two things is resolved, meaning:
> > 
> > * we start gossiping head state
> 
> Is it head in particular, or you just mean to deposit a lot more data? Let's say we start gossiping state at block 19M or 20M forward, and if we don't catch up to head quickly. Would you be happy that we are starting to move toward the natural balance point of state network being much heavier than history?

the state for block 19M or 20M is no different in value then block 1M. I would caution against backfilling not latest state like that until we have big radius nodes.

morph-dev:
> Is it head in particular, or you just mean to deposit a lot more data? Let's say we start gossiping state at block 19M or 20M forward, and if we don't catch up to head quickly. Would you be happy that we are starting to move toward the natural balance point of state network being much heavier than history?

I think I would be fine as soon as their is any significant state data present, that is at least somewhat useful.

But as @KolbyML  said, I don't think we should start by backfilling the data. We should start with a block that is very close to the head (coincidently, mainnet passed block 21M ~4h ago), and follow head.

But in principle, I would be fine as long as there is some meaningful useful data. For example, if we would be gossiping all state data from genesis to merge (which I think is less useful than the head state). And I would be fine even before we finish gossiping, just that we are doing something meaningful.

---

> I haven't dug in, but I'm guessing it will be a significant refactor to support this.

I was thinking about it. If we just want to support it, we can do it in a bit hacky way, which might not scale that well (e.g. when we start adding more subnetworks).

But I think it would be the great opportunity to do it when we start supporting big radius nodes, as we would need some big refactoring for that anyway. They both touch the concept of radius and how we handle it, maybe from slightly different angle, but they definitely have some overlap.

pipermerriam:
I didn't read everything you guys wrote.  I came here to write that if we have a flag for `--radius.max` which is defaulted to a low value but can be set to `100` for things like hive, that might address the issues this would introduce for running tiny networks with only a few nodes where it is appropriate for radius to be set to a high value.

PR #1541, opened 2024-10-20
feat: implement rlp::Encodable::length for Header

morph-dev:
### What was wrong?

The `Header` doesn't implement the [`length`](https://docs.rs/alloy-rlp/0.3.8/alloy_rlp/trait.Encodable.html#method.length) function of the [`alloy::rlp::Encodable`](https://docs.rs/alloy-rlp/0.3.8/alloy_rlp/trait.Encodable.html) trait.

This function has default implementation, so explicit implementation is not required.
However, the [default implementation](https://docs.rs/alloy-rlp/0.3.8/src/alloy_rlp/encode.rs.html#25-29) is not optimized (it encodes the object and returns the length of the byte vector), and even the [documentation](https://docs.rs/alloy-rlp/0.3.8/alloy_rlp/trait.Encodable.html#method.length) says that it's recommended to override it.

### How was it fixed?

Added private `rlp_payload_length` that calculates the length of the rlp payload (encoding without rlp header),
and use it in `encode` and `length` functions.

Also refactored rlp decoding to handle some error cases.

### To-Do

### DIFF
--- a/ethportal-api/src/types/execution/header.rs
+++ b/ethportal-api/src/types/execution/header.rs
@@ -1,8 +1,9 @@
 use alloy::{
     primitives::{keccak256, Address, Bloom, Bytes, B256, B64, U256, U64},
-    rlp::{Decodable, Encodable, Header as RlpHeader},
+    rlp::{self, Decodable, Encodable},
     rpc::types::Header as RpcHeader,
 };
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit:

PR #1542, opened 2024-10-20
feat: support "size" in eth_getBlockBy*

morph-dev:
### What was wrong?

The `eth_getBlockBy*` JSON rpc calls didn't return `size` field.

### How was it fixed?

The #1541 added support for calculating `size` field easily, so we are using it here.

Tested it locally by trying dozens of blocks (both recent and pre-merge) and comparing size field with the ones returned by: https://sandbox.alchemy.com/?method=eth_getBlockByNumber

### To-Do


KolbyML (approved):
:shipit:

carver (approved):
🚢

PR #1546, opened 2024-10-22
test: relax e2store test that depends on website

carver:
### What was wrong?

We saw master failing because https://mainnet.era.nimbus.team/ stopped listing these epochs: 701, 707, 709, 719, 725, 731, 759, 779, and 780. We can't just leave it red.

Also, when the epoch list is incomplete, the current message doesn't help you identify which one.

### How was it fixed?

Convert the test failure to a log warning.

Bonus: add the epoch index into the log.

### Possible follow-up work

Multi-source the data?

### DIFF
--- a/e2store/src/utils.rs
+++ b/e2store/src/utils.rs
@@ -4,6 +4,7 @@ use anyhow::{ensure, Error};
 use rand::{seq::SliceRandom, thread_rng};
 use reqwest::Client;
 use scraper::{Html, Selector};
+use tracing::warn;
 
 const ERA_DIR_URL: &str = "https://mainnet.era.nimbus.team/";
 const ERA1_DIR_URL: &str = "https://era1.ethportal.net/";
@@ -48,10 +49,9 @@ pub async fn get_era_files(http_client: &Client) -> anyhow::Result<HashMap<u64, 
     let era_files = download_era_links(http_client, ERA_DIR_URL).await?;
     ensure!(!era_files.is_empty(), "No era files found at {ERA_DIR_URL}");
     for epoch_num in 0..era_files.len() {
-        ensure!(
-            era_files.contains_key(&(epoch_num as u64)),
-            "Epoch indices are not starting from zero or not consecutive: missing epoch {epoch_num}",
-        );
+        if !era_files.contains_key(&(epoch_num as u64)) {
+            warn!("Epoch indices are not starting from zero or not consecutive: missing epoch {epoch_num}");
+        }
     }
     Ok(era_files)
 }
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
I am assuming this change is to make it so the error is clearer

carver:
Yeah exactly. Sorry I didn't have a PR description before.

You can now see in the test output that epoch 701 is missing:
```
thread 'utils::tests::test_get_era_file_download_links' panicked at e2store/src/utils.rs:98:59:
called `Result::unwrap()` on an `Err` value: Epoch indices are not starting from zero or not consecutive: missing epoch 701
test utils::tests::test_get_era_file_download_links ... FAILED
```

carver:
Getting this ready to go. If it's not fixed by tomorrow, let's turn master green again.

KolbyML (commented):
> Getting this ready to go. If it's not fixed by tomorrow, let's turn master green again.

Nobody is going to see a warning in CI, so I think we might as well remove the test in that case. I guess I will add a panic in Trin Execution as well

e2store/src/utils.rs
```diff
@@ -47,10 +48,11 @@ pub async fn download_era_links(
 pub async fn get_era_files(http_client: &Client) -> anyhow::Result<HashMap<u64, String>> {
     let era_files = download_era_links(http_client, ERA_DIR_URL).await?;
     ensure!(!era_files.is_empty(), "No era files found at {ERA_DIR_URL}");
-    ensure!(
-        (0..era_files.len()).all(|epoch| era_files.contains_key(&(epoch as u64))),
-        "Epoch indices are not starting from zero or not consecutive",
-    );
+    for epoch_num in 0..era_files.len() {
+        if !era_files.contains_key(&(epoch_num as u64)) {
+            warn!("Epoch indices are not starting from zero or not consecutive: missing epoch {epoch_num}");
```
KolbyML, line 53: This shouldn't be a warning. This code is used elsewhere. It isn't only for the CI. It ensures whoever calls this function gets a complete sequential era_file list, the warning would be unexpected behavior from a user's perspective.

KolbyML (changes requested):
I made a PR https://github.com/ethereum/trin/pull/1547 to remove the tests. If there is a problem with having tests which download from HTTP servers that is fine. We can remove the tests but removing the ensure!() here is not ok, because then Trin Execution will start with an era_files list which is incomplete which isn't ok. This check was added to ensure `get_era_files()` always return a valid sequential list with no files missing. Hence replacing this `ensure!()` with an warning isn't ok.

KolbyML:
I am going to set up our own mirror as well with the help of Paul

carver:
Yup, I'm happy to close in favor of #1547

PR #1547, opened 2024-10-23
chore: remove/ignore tests which download from a remote server

KolbyML:
### What was wrong?
We had tests which would fail due to missing data on a website, some tests still fail too due to this, but we want those tests so I set them to ignore
### How was it fixed?

removing the tests or ignore them. Trin Execution will still exit on start with the errors if any era files are missing so we are good


carver (approved):
Yeah, sounds good. I'll close #1546 

I do think we still want to have the error include which index is missing, like this commit: https://github.com/ethereum/trin/pull/1546/commits/531b0e14a140b0b67cf3f860b1985e44cdd1859c

KolbyML:
> Yeah, sounds good. I'll close #1546
> 
> I do think we still want to have the error include which index is missing, like this commit: [531b0e1](https://github.com/ethereum/trin/commit/531b0e14a140b0b67cf3f860b1985e44cdd1859c)

Sounds good I made the error look like `Epoch indices are not starting from zero or not consecutive: missing epochs [50, 500]` as it looks nice

Issue #1545, opened 2024-10-22
Monitor CPU usage and internally rate limit


carver:
## What's wrong?

Especially with the state network, and when onboarding a new node, trin can redline the CPU while accepting offers and re-offering to peers. (I tended to see my fans spinning up when trin reached about 10 offers accepted per second).

Just like with storage, we don't want to overuse CPU. We prefer the client to feel light, and as something that can be constantly running in the background.

## Possible solution

Monitor `cpu_time::ProcessTime` inside trin, and cap it at some small % by default (2%? 5%? 10%?). Add a CLI flag to manually configure it.

Every 10s that the CPU is above the limit, shrink the data radius of the state network by 10% of its current level. (This is making the assumption, based on the current experience, that state is the only offender for high CPU usage). Every 10s that the CPU is below the limit by at least half, and the data storage is under target, then grow the radius by 10%. I think we want to be quite responsive, which this accomplishes by being willing to cut the radius in ~half in 1 minute.

This approach might actually accelerate state nodes finding their natural "true" radius point faster, and mitigate the fill & dump behavior when launching a fresh client (which is more slow and painful on state than history).

## Challenges

trin could use CPU for other reasons, like due to user interaction (ie~ when someone is using the RPC API). It is easy to imagine that CPU usage will spike then, and it would be wrong to mess with state radius at that point. We probably cannot punt on this, and will need to include a solution with the first implementation.

Another awkward aspect of this approach is that it's hard to tell which network is using too much CPU. We probably don't want to adjust every network's radius at once if CPU is high. Right now, it's only ever state, so I think we can punt this challenge. But if multiple networks start using a lot of CPU, we might want a clever way to measure computation that isn't just checking `cpu_time::ProcessTime`.

## Timeline

I won't try to get this into the imminent stable release, of course, just planning ahead. This is another good reason not to enable state by default.

KolbyML:
> Do we know (and if not, can we measure) what exactly is causing high CPU usage? Is it the process of validating state content, gossiping that content or something else?
> 
> Instead of shrinking/growing radius, should we maybe reduce/increase number of parallel requests that we are accepting?
> 
> I think your proposed solution is good first step.

> Do we know (and if not, can we measure) what exactly is causing high CPU usage?

I think this is the key takeaway is ^. Normally performance problems like this are small problems that manifest in big ways. If is likely 1 hot coldpath which can be fixed. This could be the redhairing which is preventing us from scaling Trin's concurrency limit.

[...]

PR #1549, opened 2024-10-24
feat: log cpu utilization in state network report

carver:
### What was wrong?

Related to #1545 -- to start we at least want to measure CPU usage. However we decide to respond to it, it is a thing we will want to respond to. So this is an experiment to see how well the library works at observing usage, to lay the groundwork for later steps.

### How was it fixed?

Add `cpu_time` to track the amount of cpu usage that trin has. It's a bit hacky to put it in the `trin-state` report, because it's just measuring everything happening in the process. But since state is not stabilizing (and seems to be the culprit for high usage currently), it seems like a good place to add the log.

It's already been a convenient way to watch my machine react as I experiment with the CPU effect of various changes.

Example log:
```
2024-10-24T04:53:32.129663Z  INFO trin_state: reports~ data: radius=12% content=53.3/495mb #=131860 disk=204.9mb; msgs: offers=5235/5522, accepts=7070/7070, validations=706/707; cpu=8.4%
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1452,6 +1452,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
+name = "cpu-time"
+version = "1.0.0"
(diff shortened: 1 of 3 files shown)


ogenev (approved):
Looks reasonable to me.

PR #1552, opened 2024-10-24
feat: alloy bytes conversion

etherhood:
### What was wrong?
Replace Vec<u8> with alloy::Bytes or bytes::Bytes based on usage

### To-Do
Replace intermediary usage of Vec<u8> in some places

### DIFF
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -37,7 +37,7 @@ const TALKREQ_CHANNEL_BUFFER: usize = 100;
 /// ENR key for portal network client version.
 pub const ENR_PORTAL_CLIENT_KEY: &str = "c";
 
-pub type ProtocolRequest = Bytes;
+pub type ProtocolRequest = Vec<u8>;
 
 /// The contact info for a remote node.
 #[derive(Clone, Debug)]
@@ -464,7 +464,7 @@ impl AsyncUdpSocket<UtpEnr> for Discv5UdpSocket {
     async fn send_to(&mut self, buf: &[u8], target: &UtpEnr) -> io::Result<usize> {
         let discv5 = Arc::clone(&self.discv5);
         let target = target.0.clone();
-        let data = Bytes::from(buf.to_vec());
+        let data = buf.to_vec();
         tokio::spawn(async move {
             match discv5.send_talk_req(target, Subnetwork::Utp, data).await {
                 // We drop the talk response because it is ignored in the uTP protocol.
--- a/portalnet/src/discovery.rs
+++ b/portalnet/src/discovery.rs
@@ -10,6 +10,7 @@ use std::{
 
 use anyhow::anyhow;
 use async_trait::async_trait;
+use bytes::Bytes;
 use discv5::{
     enr::{CombinedKey, Enr as Discv5Enr, NodeId},
     ConfigBuilder, Discv5, Event, ListenConfig, RequestError, TalkRequest,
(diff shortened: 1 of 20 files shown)


trin-storage/src/lib.rs
```diff
@@ -125,10 +128,11 @@ impl ContentStore for MemoryContentStore {
         &mut self,
         key: Self::Key,
         value: V,
-    ) -> Result<Vec<(Self::Key, Vec<u8>)>, ContentStoreError> {
+    ) -> Result<Vec<(Self::Key, RawContentValue)>, ContentStoreError> {
         let content_id = key.content_id();
         let value: &[u8] = value.as_ref();
-        self.store.insert(content_id.to_vec(), value.to_vec());
+        self.store
+            .insert(content_id.to_vec(), Bytes::from(value.to_vec()));
```
morph-dev, line 135: This should be: `Bytes::copy_from_slice(value)`

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -265,7 +265,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
             named_params! {
                 ":content_id": content_id,
                 ":content_key": content_key,
-                ":content_value": content_value,
+                ":content_value": content_value.to_vec(),
```
morph-dev, line 268: Instead of `to_vec()` we should use `as_ref()`.

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -510,7 +510,11 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
             self.usage_stats.entry_count -= to_delete;
             self.usage_stats.total_entry_size_bytes -= deleted_content_size;
             self.usage_stats.report_metrics(&self.metrics);
-            deleted_content.extend(deleted_content_values);
+            deleted_content.extend(
```
morph-dev, line 513: instead of doing it here, conversion should be done while reading data from db. Somewhere around line 481.
For example, something like this:
```rust
let key_bytes: Vec<u8> = row.get("content_key")?;
let value_bytes: Vec<u8> = row.get("content_value")?;
let value = RawContentValue::from(value_bytes);
let size: u64 = row.get("content_size")?;
TContentKey::try_from_bytes(key_bytes)
    .map(|key| (key, value, size))
    .map_err(|e| {
        rusqlite::Error::FromSqlConversionFailure(0, Type::Blob, e.into())
    })
```

trin-storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -811,12 +815,15 @@ mod tests {
         assert!(!store.has_content(&id)?);
         let usage_stats = store.usage_stats();
 
-        store.insert(&key, value.clone())?;
+        store.insert(&key, value.clone().into())?;
```
morph-dev, line 818: Instead of calling `value.into()` here and in many places below, we should refactor `generate_key_value` and `generate_key_value_with_content_size` so they return `RawContentValue`

portalnet/src/discovery.rs
```diff
@@ -36,7 +37,7 @@ const TALKREQ_CHANNEL_BUFFER: usize = 100;
 /// ENR key for portal network client version.
 pub const ENR_PORTAL_CLIENT_KEY: &str = "c";
 
-pub type ProtocolRequest = Vec<u8>;
+pub type ProtocolRequest = Bytes;
```
morph-dev, line 40: After looking more into how this is used, I think it's better if this is not changed to `Bytes` and left as `Vec<u8>`.

morph-dev (changes requested):
A lot of good work, but also a lot of small comments.

etherhood:
> A lot of good work, but also a lot of small comments.

Addressed all the comments, ready for review

morph-dev:
Looks good. Thanks!

PR #1553, opened 2024-10-24
refactor: census peer discovery algorithm

morph-dev:
## What was wrong?

The census initialization and new peer discovery is not very optimal.

#### Initialization

Runs one RFN (recursive find content) and then "processes" discovered peers in order.

#### Peer processing

- if peer is not already "known"
  - sends ping request. if successful:
    - mark is as "known"
    - send many "find nodes" requests in order to fully discover peers routing table
    - each "not-known" discovered peer is pinged and marked as "known" if ping is successful

#### New peers discovering

Each "known" peer expires after 1h. At that point, it's marked as "not-known" and processed.

#### Issues

Initialization would end up "processing" only 1 or 2 nodes because most of the peers discovered by RFN would be marked as "known" while processing first one. This leads to discovering small part of the network.

"Peer processing" is relatively slow. At least it's slow enough if we have to do it for every peer. This leads to disconnected peers being considered "known" for too long (1h).

Peers are discovered in bursts. First 1h after initialization, we won't discover any new peers (we already cover small-ish view of the network). After 1h, all peers starts expiring and we start "processing" all of them, and marking them as "known". This is bad because:
- next peer processing will happen 1h later (no new peers are discovered or removed in the meantime)
- this intense work affects bridge's gossip

It's hard to implement peer scoring when peer discovery is tied with processing "known" peers

## How was it fixed?

Approach initialization and peer discovery from completely different angle:

#### Initialization

1. Run 16 RFN in parallel
2. Ping all discovered peers that are "not-known" and consider them "known" if ping is successful
3. If number of new "known" peers from step 2 is more than 10% of total number of "known" peers, start from step 1 again

#### Peer discovery

Once per minute, run 1 RFN, ping "not-known" discovered peers, mark them as "known" if successful

#### Removing disconnected peers

Every "known" peer is pinged once every 5 min, and kept as "known" if ping is successful.

This can create similar "burst" problem described above, but it's not as severe, because:
- it only pings peers (which is much cheaper than "processing")
- this will be removed in next PR, when peers start being aware of other non-ping requests that are sent to them (needed for scoring)
  - this couldn't be done originally as they "processing" was tied to discovering new peers and re-checking their liveness

## Stats

I run initialization with original and new implementation, for both History and State network. The summary:

**History network (average of 10)**

|  | original implementation | new implementation |
|---|---|---|
| peers | 59.7 | 255.7 |
| duration (seconds) | 16.7 s | 10.6 s |

**State network (average of 10)**

|  | original implementation | new implementation |
|---|---|---|
| peers | 43.2 | 107.7 |
| duration (seconds) | 7.2 s | 10.3 s |

New implementation discovers majority of the network (significant more than original implementation), while not running noticeable longer.

### To-Do


KolbyML (approved):
:shipit: looks good

njgheorghita (approved):
:ship: Those numbers are looking good to me! Nothing jumps out as a potential pitfall with the new approach

PR #1554, opened 2024-10-25
fix(beacon): update `LightClientHeader` fork version

ogenev:
### What was wrong?
Portal bridges are injecting light client bootstraps into beacon testnet with an older fork version (Capella instead of Deneb)

### How was it fixed?
- Update `LightClientHeader` to support the Deneb fork which will allow correct deserialization for recent bootstraps.
- Add more light client tests for Deneb.

### To-Do


ethportal-api/src/types/consensus/light_client/bootstrap.rs
```diff
@@ -152,4 +153,39 @@ mod test {
         LightClientBootstrap::from_ssz_bytes(&expected, ForkName::Capella).unwrap();
         assert_eq!(content.as_ssz_bytes(), expected);
     }
+
+    #[rstest]
+    #[case("case_0")]
+    #[case("case_1")]
+    fn serde_light_client_bootstrap_deneb(#[case] case: &str) {
```
carver, line 160: Am I reading this right that these are just two randomly generated examples? They aren't some specific variants being tested?

If it's random, then I guess I have no notes. But if there's a particular difference being tested, it would be nice to make that clearer.

trin-beacon/src/test_utils.rs
```diff
@@ -51,33 +51,33 @@ pub fn get_light_client_update(number: u8) -> ForkVersionedLightClientUpdate {
...
+        LightClientFinalityUpdate::from_ssz_bytes(&lc_finality_update, ForkName::Deneb).unwrap();
 
     ForkVersionedLightClientFinalityUpdate {
         fork_name: ForkName::Deneb,
         update: lc_finality_update,
     }
 }
 
-// Valid number range for the test cases is 0..4
+// Valid number range for the test cases is 0..1
```
carver, line 71: *Tiniest of nits*: this looks just like rust range syntax, but then it's off by one. So I guess it's slightly better to use something like:
```suggestion
// Valid number range for the test cases is 0..2
```

or 
```suggestion
// Valid number range for the test cases is 0..=1
```

carver (approved):
LGTM

ethportal-api/src/types/consensus/light_client/bootstrap.rs
ogenev: They are random.

PR #1555, opened 2024-10-25
feat: add format validation to era2-stats binary

morph-dev:
### What was wrong?

The "era2-stats" binary doesn't check validity of the file.

### How was it fixed?

Added validation. The only thing missing is checking that state data actually results in the canonical state trie, but that is covered by `import-state` subcommand of the trin-execution.

### To-Do

### DIFF
--- a/e2store/src/bin/era2-stats.rs
+++ b/e2store/src/bin/era2-stats.rs
@@ -1,5 +1,6 @@
 use std::path::PathBuf;
 
+use alloy::primitives::{keccak256, B256};
 use anyhow::bail;
 use clap::Parser;
 use e2store::era2::{AccountOrStorageEntry, Era2Reader};
@@ -8,7 +9,7 @@ use tracing::info;
 #[derive(Debug, Parser)]
 #[command(
     name = "Era2 stats",
-    about = "Reads the era2 file and prints stats about it."
+    about = "Reads the era2 file, validates it, and prints stats about it."
 )]
 struct Config {
     #[arg(help = "The path to the era2 file.")]
@@ -23,12 +24,12 @@ struct Stats {
     storage_item_count: usize,
 }
 
-/// Reads the era2 file and prints stats about it.
+/// Reads the era2 file, validates it, and prints stats about it.
 ///
 /// It can be run with following command:
 ///
 /// ```bash
-/// cargo run -p e2store --bin era2-stats --features era2-stats-binary -- <path>
+/// cargo run --release -p e2store --bin era2-stats --features era2-stats-binary -- <path>
 /// ```
 fn main() -> anyhow::Result<()> {
     trin_utils::log::init_tracing_logger();
@@ -41,6 +42,8 @@ fn main() -> anyhow::Result<()> {
         reader.header.header.number
     );
 
+    let mut last_address_hash = B256::ZERO;
+
     let mut stats = Stats::default();
     while let Some(entry) = reader.next() {
         let AccountOrStorageEntry::Account(account) = entry else {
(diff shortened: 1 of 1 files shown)


KolbyML:
Checking for sequentiality and if the content is valid is 2 different things no. I guess you could say we are checking if the files our format specification which includes sequentially. But I would say just saying "validation" is misleading.

morph-dev:
> Even if the accounts are out of order, the data can still be valid and generate the canonical state trie, so I think the wording is misleading.

How would you phrase it?

e2store/src/bin/era2-stats.rs
```diff
@@ -23,12 +24,12 @@ struct Stats {
     storage_item_count: usize,
 }
 
-/// Reads the era2 file and prints stats about it.
+/// Reads the era2 file, validates it, and prints stats about it.
```
KolbyML, line 27:
```suggestion
/// Reads the era2 file, validates the format, and prints stats about it.
```
or

```suggestion
/// Reads the era2 file, verifies keys are sequential, and prints stats about it.
```

e2store/src/bin/era2-stats.rs
```diff
@@ -8,7 +9,7 @@ use tracing::info;
 #[derive(Debug, Parser)]
 #[command(
     name = "Era2 stats",
-    about = "Reads the era2 file and prints stats about it."
+    about = "Reads the era2 file, validates it, and prints stats about it."
```
KolbyML, line 12:
```suggestion
    about = "Reads the era2 file, validates the format, and prints stats about it."
```

or 

```suggestion
    about = "Reads the era2 file, verifies keys are sequential, and prints stats about it."
```

KolbyML (commented):
These are just 2 idea's maybe there is a better one? I just would like it to be more specific so there isn't the implication this verifies `generate the canonical state trie` we can generate the expected trie.

So phrase any phase which is more specific I think would be fine.

morph-dev:
Updated as suggested.

KolbyML (approved):
:shipit:

e2store/src/bin/era2-stats.rs
```diff
@@ -23,12 +24,12 @@ struct Stats {
...
 }
 
-/// Reads the era2 file and prints stats about it.
+/// Reads the era2 file, validates the format, and prints stats about it.
 ///
 /// It can be run with following command:
 ///
 /// ```bash
-/// cargo run -p e2store --bin era2-stats --features era2-stats-binary -- <path>
+/// cargo run --release -p e2store --bin era2-stats --features era2-stats-binary -- <path>
```
carver, line 32: I assume this is just because the operation is slow, so `--release` helps speed it up?

e2store/src/bin/era2-stats.rs
```diff
@@ -61,6 +78,24 @@ fn main() -> anyhow::Result<()> {
...
             stats.storage_entry_count += 1;
             stats.storage_item_count += storage.len();
+
+            if index + 1 < account.storage_count {
+                assert_eq!(
+                    storage.len(),
+                    10_000_000,
+                    "Non-final storage entry doesn't have 10M items",
+                );
+            }
```
carver, line 88: It took me a minute to grok this (granted, I was unfamiliar with the era format). Maybe it's the double-negative in the error message? A quick comment might help speed up comprehension for era noobs. Something like:
```suggestion
            // Validate that every chunk of storage is full, with 10M items (skipping the final chunk)
            if index + 1 < account.storage_count {
                assert_eq!(
                    storage.len(),
                    10_000_000,
                    "Non-final storage entry doesn't have 10M items",
                );
            }
```

e2store/src/bin/era2-stats.rs
```diff
@@ -61,6 +78,24 @@ fn main() -> anyhow::Result<()> {
             };
             stats.storage_entry_count += 1;
             stats.storage_item_count += storage.len();
+
+            if index + 1 < account.storage_count {
+                assert_eq!(
+                    storage.len(),
+                    10_000_000,
+                    "Non-final storage entry doesn't have 10M items",
```
carver, line 86: A bit more info in the error message would be sweet. Probably something like: the storage index, the actual storage length and which account it was in.

carver (approved):
LGTM

e2store/src/bin/era2-stats.rs
morph-dev: Yes, that's correct.

e2store/src/bin/era2-stats.rs
morph-dev: Added comment, removed double-negative and added more details to the message.

Issue #1548, opened 2024-10-23
Subnetwork and storage command line flags


morph-dev:
Given our recent discussion in a team meeting, I think we should adjust command line flags to reflect our long term direction before we make "v0.1" next week.

I'm going to propose couple of ideas and give my opinion on them, but I encourage everybody to give their opinion, propose alternatives, etc.

One thing that we should keep in mind is that certain dependency between networks should be enforced (in order for content to be validated).
For example, if state is enabled, history has to be as well (and probably soon if history is enabled, beacon has to be enabled as well).
This is not major issue as we can just shut down at startup and print the message.

# Type A - Subnetwork and total capacity (different radius per network)

What we have at the moment, but with smarter distribution of storage per network (e.g. if only 100 MB is given, split it 50/50, if 10GB split it 90/10, etc)

Example: `--portal_subnetworks=history,state --mb=1000`

Pros: Simple user for beginners experience (for beginners)
Cons: Not custom enough for expert users, weird behavior if different subnetworks are enabled between restarts

# Type B - Explicit capacity for each network

Each subnetwork has its own storage capacity that is configured separately

Examples (1GB for history, 2GB for storage):
1. `--portal_subnetworks=history:1000,state:2000`
2. `--portal_subnetworks=history:1000 --portal_subnetworks=state:2000`
3. `--history=1000 --state=2000`

**Note:** We can support examples 1 and/or 2, or alternatively only 3, or if somebody has some other ideas.

Some ideas for making it easier to use (not required for first implementation): instead of explicitly specifying the capacity, one can set "auto" (or not set anything, in which case "auto" is default).
Meaning of "auto" can be discussed separately. Some ideas:
- 1GB if all capacities are "auto", otherwise average of all non-auto values
- "auto" is allowed only if all capacities are "auto", in which case it means 1GB per subnetwork

Pros: explicit, fully customizable (good for expert users)
Cons: harder to define default values and make it clear to the user 

# Type C - Subnetwork and total capacity (unified radius per network)

It seems that we don't want to go with this approach, but I added it for completeness.

Similar to "Type A", but capacity per subnetwork will be dynamic.

All subnetwork will have the same radius, and the amount they store per network will be byproduct of that and having fixed total capacity.

Pros: simple user experience, user's storage distribution should be the same as total storage capacity of each subnetwork
Cons: weird behavior if users enables/disables subnetwork between restarts, can cause weird behavior on global scale (content from one network can disappear because another network gossiped a lot of new content)

---

[...]

PR #1558, opened 2024-10-27
feat: add cli flags for storage capacity per subnetwork

morph-dev:
### What was wrong?

Based on #1548 and discussion in the last meeting, we agreed that we following cli flags regarding storage capacity:

1. `storage.total` flag that specifies total storage capacity (in MB)
    - this capacity should be shared between all enabled networks
    - current flag `mb` should be set as alias (until we migrate in other places)
2. `storage.beacon`, `storage.history` and `storage.state` that specifies storage capacity (in MB) per subnetwork explicitly

Some restrictions:
- `storage.total` can't be combined with other flags
    - other flags can/should be used together, see next point
- if `storage.{subnetwork}` is specified for one enabled subnetwork, it should be specified for all enabled networks

For future work:
- if user didn't enable subnetwork, but we detect that it's using some storage on disk (e.g. from previous run), print warning message (or do something else, TBD).

### How was it fixed?

- created cli flags
    - added detailed documentation on how they should be used
- added checks when `TrinConfig` is created
    -  added tests
- adapted logic for creating `PortalStorageConfig`
    - added tests

### To-Do


KolbyML (approved):
@morph-dev over all the PR looks good

trin-storage/src/config.rs
```diff
@@ -76,3 +111,121 @@ pub struct PortalStorageConfig {
...
+    #[case::history_beacon(vec![Subnetwork::History], 100, Subnetwork::Beacon, None)]
+    #[case::history_history(vec![Subnetwork::History], 100, Subnetwork::History, Some(100_000_000))]
+    #[case::history_state(vec![Subnetwork::History], 100, Subnetwork::State, None)]
+    #[case::historystate_beacon(vec![Subnetwork::History, Subnetwork::State], 100, Subnetwork::Beacon, None)]
+    #[case::historystate_history(vec![Subnetwork::History, Subnetwork::State], 100, Subnetwork::History, Some(1_000_000))]
+    #[case::historystate_state(vec![Subnetwork::History, Subnetwork::State], 100, Subnetwork::State, Some(99_000_000))]
+    #[case::beaconhistorystate_beacon(vec![Subnetwork::Beacon, Subnetwork::History, Subnetwork::State], 100, Subnetwork::Beacon, Some(0))]
+    #[case::beaconhistorystate_history(vec![Subnetwork::Beacon, Subnetwork::History, Subnetwork::State], 100, Subnetwork::History, Some(1_000_000))]
+    #[case::beaconhistorystate_state(vec![Subnetwork::Beacon, Subnetwork::History, Subnetwork::State], 100, Subnetwork::State, Some(99_000_000))]
+    fn combined_capacity_config(
```
carver, line 135: nice coverage 👍🏻

ethportal-api/src/types/cli.rs
```diff
@@ -312,8 +357,77 @@ impl TrinConfig {
...
+                    }
+                    if storage_flag.is_some() && !config.portal_subnetworks.contains(&subnetwork) {
+                        return Err(Error::raw(
+                            ErrorKind::ArgumentConflict,
+                            format!(
+                                "--storage.{} is set but {subnetwork} subnetwork is not enabled",
+                                subnetwork.to_cli_arg(),
+                            ),
+                        ));
+                    }
```
carver, line 406: I'm not sure we *have* to error out in this case. I do like the impulse to be strict to avoid accidents.

But I can also imagine a scenario where I'm flipping between enabling and disabling a network, and having to do that in two places in the command being a little annoying.

I'm slightly in favor of dropping this if you're neutral.

ethportal-api/src/types/cli.rs
```diff
@@ -256,7 +301,7 @@ impl TrinConfig {
 
     pub fn new_from<I, T>(args: I) -> Result<Self, clap::Error>
     where
-        I: Iterator<Item = T>,
+        I: IntoIterator<Item = T>,
```
carver, line 304: I do really like this change. It also added noise to a big PR. I think it belongs in its own commit/PR.

ethportal-api/src/types/cli.rs
```diff
@@ -709,6 +804,247 @@ mod test {
...
+                "--storage.beacon",
+                "0",
+                "--portal-subnetworks",
+                "history,state,beacon",
+            ])
+            .unwrap();
+            assert_eq!(
+                config.storage_capacity_config(),
+                StorageCapacityConfig::Specific {
+                    beacon_mb: Some(0),
```
carver, line 969: I don't think there are any actual bugs here, but as a practice the test for setting a specific value should probably pick one that isn't the value you would arrive at by default.

carver (approved):
Love all the tests!

*nit* I just did a quick grep, and didn't see any tests that the `--mb` alias still works. I'd be in favor of a single simple one to make sure we don't accidentally disable it in the future.

morph-dev:
Rebased, fixed default capacity tests, added `--mb` and zero capacity tests, and reverted partially reverted refactoring changes. 
I will merge it like this.

Let me know if you have other comments and I will address them in the followup PR.

Issue #1396, opened 2024-08-24
Smart Contract's bytecode is gossiped every time balance or storage is updated.


morph-dev:
Currently, we will try to gossip Smart Contract's code even if it is not modified.
If contract's balance or storage was updated, entire AccountState (and leaf in the account trie) will be updated as well.

The following code would then gossip bytecode as well: https://github.com/ethereum/trin/blob/3fc0ed81fc165961f8c587a95fb5de66a9df7b7e/portal-bridge/src/bridge/state.rs#L180-L191

I'm not sure if there is easy way for this to be fixed.

KolbyML:
I don't think this is a milestone for devcon 2024

KolbyML:
Just in case someone wanted to work on this, https://github.com/ethereum/trin/pull/1337 this PR should be merged first as it would invalidate the work from this PR.

I think in the grand scheme of things there is a lot that should be done before this issue though

morph-dev:
> I don't think this is a milestone for devcon 2024

While I think it's not critical for the devcon and there are other important issue, I think we should have this if we want to gossip head blocks (because there will be a lot of contracts that are touched at every block).

morph-dev:
@KolbyML , the #1559 fixed this, right?

KolbyML:
Yes

PR #1559, opened 2024-10-28
feat: only bridge newly created contracts

KolbyML:
### What was wrong?
We would always bridge contracts, which isn't good because the contract should already be gossiped on the network. This takes up a lot of bridge bandwidth, which is redundant as the contracts should already be on the network.
### How was it fixed?

By using a `cache` to store newly created contracts like we do for `storage_trie` diffs.

### DIFF
--- a/trin-execution/src/storage/evm_db.rs
+++ b/trin-execution/src/storage/evm_db.rs
@@ -39,8 +39,10 @@ fn start_processing_timer(name: &str) -> HistogramTimer {
 pub struct EvmDB {
     /// State config
     pub config: StateConfig,
-    /// Storage cache for the accounts used optionally for gossiping, keyed by address hash.
-    pub storage_cache: Arc<Mutex<FbHashMap<32, HashSet<B256>>>>,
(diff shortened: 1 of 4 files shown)


trin-execution/src/storage/evm_db.rs
```diff
@@ -102,6 +106,15 @@ impl EvmDB {
         trie_diff
     }
 
+    pub fn get_newly_created_contract_if_available(&self, code_hash: B256) -> Option<Bytecode> {
```
morph-dev, line 109: nit: I think that `get_newly_created_contract` is clear enough in this case. Up to you.

morph-dev (approved):
:shipit: 

Add reference to #1396 and close it once it's merged.

portal-spec-tests
morph-dev, line 1: I don't think this should be here. I assume you rebased to master, but didn't handle submodules correctly.
I don't know why all tests are still passing, so I could be wrong.

Just wanted to highlight it, so you can double check what is happening.

KolbyML: I didn't rebase that is an artifact from me switching from master

PR #1561, opened 2024-10-29
feat: update state:history storage to 1:1

carver:
_I don't have strong feelings about what the new ratio should be. I just picked one, mostly to kick off discussion. Now, we have more information than we did at the 99:1 ratio selection._

The new ratio is still a loose guess. But it is based on some recently acquired state data, specifically:

A recent average amount of new state generated is about 1.75MB / block or ~375 GB per month. Let's make room for 10 years of intremental state at that rate, which gives us a network size of: ~45 TB.

The typical history node can fill up 35GB of storage with a ~2% radius, implying a network size of: ~1.75 TB.

45 TB / 1.75 TB ~= 25:1 ratio

Inspired by conversation and data collection here: https://discord.com/channels/890617081744220180/1089234065816821860/1300370058366947370

### DIFF
--- a/trin-storage/src/config.rs
+++ b/trin-storage/src/config.rs
@@ -20,7 +20,7 @@ pub struct PortalStorageConfigFactory {
 
 impl PortalStorageConfigFactory {
     const HISTORY_CAPACITY_WEIGHT: u64 = 1;
-    const STATE_CAPACITY_WEIGHT: u64 = 99;
+    const STATE_CAPACITY_WEIGHT: u64 = 1;
     const BEACON_CAPACITY_WEIGHT: u64 = 0; // Beacon doesn't care about given capacity
 
     pub fn new(


morph-dev:
I wanted to get #1558 merged first and I would have started this topic myself :). I agree that 99:1 is broken and we should do something about it.

Based on the discussion during meeting, my impression is that we want different ratio based on total storage capacity (different ratio for 100 MB vs 1 GB vs 10 GB). I also think that we should start measuring and dedicating some storage to Beacon as well.

With this being said, I think that at the moment we don't have to do any special math, as we will have to redo it again, most likely soon. So I'm in favor of just doing 1:1 or even dedicating more storage towards history (as this is closer to actual current usage).

---

Somewhat related to this, now that we have cli flags to specify storage per network, I would like to restart the discussion about enabling state network by default. My original concerns were:
- state network not yet storing very useful data - this is still my concern
- implementing unified radius - this is no longer my concern because:
    - we no longer plan to use unified radius
    - have a way to specify storage per subnetwork (#1558 )
    - plan to change default storage ratio (this PR)

So my only remaining concern is that state network doesn't have very useful data, yet. So if you or anybody else still feel like we should enable it, I wouldn't be against it.

carver:
> So I'm in favor of just doing 1:1 or even dedicating more storage towards history (as this is closer to actual current usage).

Yeah, let's do 1:1 to just not overthink it. We can put the extra thought into whatever scheme we come up with next :)

> I wanted to get #1558 merged first and I would have started this topic myself :)

Hah, yeah I spawned this PR when reviewing that one.

---

> So my only remaining concern is that state network doesn't have very useful data, yet. So if you or anybody else still feel like we should enable it, I wouldn't be against it.

Yeah, looking back I was having a mini manic moment on trying to enable it then. I think it's pretty clear that the stability of state isn't really approaching history. So I at least would want to wait past Devcon. Some other reference points I might think about for calling it stable enough to turn on by default:
- 21M state & up added
- A couple weeks of glados showing >=99.9% on state audits for the first 1M and the 21M+ state
- A glados audit for recent state (Maybe ~3 block lag is okay for now)

morph-dev (approved):
:shipit:

PR #1564, opened 2024-10-29
feat(cli): set query timeout value via CLI argument

ogenev:
### What was wrong?
When a beacon trin node request a light client content (while keeping in sync with the latest header), the query can take more time than the beacon slot duration (12 sec), because the default query timeout duration is set to 60 sec. 

For beacon, we want to set this timeout to no more than 12 sec, so we can always request the light client content for the next slot and not miss slots.

### How was it fixed?
Set the query timeout via CLI argument.

### To-Do

### DIFF
--- a/ethportal-api/src/types/cli.rs
+++ b/ethportal-api/src/types/cli.rs
@@ -20,6 +20,7 @@ const DEFAULT_SUBNETWORKS: &str = "history";
 pub const DEFAULT_NETWORK: &str = "mainnet";
 pub const DEFAULT_STORAGE_CAPACITY_MB: &str = "1000";
 pub const DEFAULT_WEB3_TRANSPORT: &str = "ipc";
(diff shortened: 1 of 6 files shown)


carver:
I didn't add a 12s timeout for specific beacon queries, but #1566 should do most of the heavy lifting. On top of that, it should just be a line or two to drop into `light-client/src/consensus/rpc/portal_rpc.rs`. (Is it just the optimistic updates that should have a custom limit?)

Anyway, I figured just having the refactor available would be helpful for you to do whatever kind of testing you want to do with a 12s limit while running beacon, @ogenev . So I think we can merge the refactor without it, if the approach looks good enough to everyone.

ogenev:
> Wouldn't setting this flag to 12 seconds (I assume desired when beacon is enabled), change behavior for other networks as well?

This is a good point, we may want different timeouts per network and content. I'm closing this in favor of https://github.com/ethereum/trin/pull/1566.

PR #1566, opened 2024-10-29
refactor: ability to set timeout per-content-query

carver:
The feature is not actually used anywhere yet, just laying the groundwork. Inspired by discussion in #1564


portalnet/src/overlay/command.rs
```diff
@@ -40,3 +42,17 @@ pub enum OverlayCommand<TContentKey> {
     /// Handle an event sent from another overlay.
     Event(EventEnvelope),
 }
+
+/// A configuration for a query to find content.
+#[derive(Debug, Default)]
+pub struct FindContentConfig {
```
ogenev, line 48: nit: I think we can move this struct to `/overlay/config.rs` but I don't have very strong opinion.

ogenev (approved):
I like the refactoring and bundling of `is_trace` with the query timeout in a config 👍 .

PR #1567, opened 2024-10-29
feat: implement new trie_walker which can iterate full+partial tries

KolbyML:
### What was wrong?
The old `TrieWalker` implementation I did only worked for trie-diffs as it did everything in memory. This is bad because we want to be able to gossip the whole state trie onto the network.

My new `trie_walker` implementation will also allow us to reuse code for horizontal scaling (i.e. walking the trie with ranged scopes). So either way having 1 trie implementation which is more flexible should be easier to maintain, keeping both is redundent.
### How was it fixed?
Implement a trie-walker which builds iterative proofs sequentially, instead of building a map of back traces from every node like the old trie-walker did.

Using this new trie-walker I was able to iterate whole the whole state at block 21 million while only using 1.4GB of ram which is really good as the whole state is almost 400GB

I did some testing and both implementations generate the same amount of content I tested it from 0-100k
```
Using new trie walker
Oct 29 16:33:25.467  INFO test_we_can_generate_content_key_values_up_to_x: content_generation: Finished all 100000 blocks: Stats { content_count: 742202, gossip_size: 1135779983, storage_size: 304058469, account_trie_count: 728452, contract_storage_trie_count: 13476, contract_bytecode_count: 274 }
Using old trie walker
Oct 29 16:34:14.857  INFO test_we_can_generate_content_key_values_up_to_x: content_generation: Finished all 100000 blocks: Stats { content_count: 742202, gossip_size: 1135779983, storage_size: 304058469, account_trie_count: 728452, contract_storage_trie_count: 13476, contract_bytecode_count: 274 }
```


trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+                None => {
+                    // If we are walking a partial trie, some nodes won't be available in the
+                    // database
+                    if self.is_partial_trie {
+                        return Ok(());
+                    }
+                    return Err(anyhow::anyhow!("Node not found in the database"));
+                }
+            };
+            let decoded_node = decode_node(&mut value.as_ref())?;
```
morph-dev, line 101: nit: I guess we want to do this to make sure that it's valid node. How about something like this:

```rust
// check that node decodes correctly and to correct variant
if matches!(
        decode_node(&mut value.as_ref())?,
        Node::Empty | Node:: Hash(_)) {
    // log warning and return Ok
    // or maybe return Err
}
partial_proof.push(value.into());
self.stack.push(TrieProof { path, proof: partial_proof });
```

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+                        extension.prefix.get_data().to_vec().as_slice(),
+                    ]
+                    .concat(),
+                )
+                .expect("Failed to process node");
+            }
+            Node::Branch(branch) => {
+                let branch = branch.read().expect("Branch node must be readable");
+
+                // We don't need to check the branches value as it is already encoded in the proof
```
morph-dev, line 149: nit: I think this comment is not needed. We are iterating Trie nodes, not Trie values.

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+    async fn test_state_walker() {
+        let temp_directory = create_temp_test_dir().unwrap();
+        let db = Arc::new(setup_rocksdb(temp_directory.path()).unwrap());
+        let trie = Arc::new(Mutex::new(EthTrie::new(Arc::new(TrieRocksDB::new(
+            false,
+            db.clone(),
+        )))));
+        {
+            let mut trie = trie.lock();
+            trie.insert(
```
morph-dev, line 198: Can we do this in a loop?

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+    use std::{str::FromStr, sync::Arc};
+    use tracing_test::traced_test;
+    use trin_utils::dir::create_temp_test_dir;
+
+    #[tokio::test]
+    #[traced_test]
+    async fn test_state_walker() {
+        let temp_directory = create_temp_test_dir().unwrap();
+        let db = Arc::new(setup_rocksdb(temp_directory.path()).unwrap());
+        let trie = Arc::new(Mutex::new(EthTrie::new(Arc::new(TrieRocksDB::new(
```
morph-dev, line 192: Do we need this to be `Arc<Mutex<...>>`?

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+            trie.insert(
+                B256::from(U256::from(18)).as_slice(),
+                B256::from(U256::from(18)).as_slice(),
+            )
+            .unwrap();
+            trie.root_hash().unwrap();
+        }
+
+        let root_hash = trie.lock().root_hash().unwrap();
+        let walker = TrieWalker::new(root_hash, trie.lock().db.clone()).unwrap();
```
morph-dev, line 292: Isn't `trie.lock().db.clone()` just `db`?

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+            };
+
+            let Node::Leaf(leaf) =
+                decode_node(&mut encoded_last_node.as_ref()).expect("Failed to decode node")
+            else {
+                continue;
+            };
+            leaf_count += 1;
+
+            // reconstruct the address hash from the path so that we can fetch the
```
morph-dev, line 309: I'm confused by this comment. What are we fetching from the database?

morph-dev (commented):
No major concerns, but several non-trivial things.

morph-dev:
Also, what do you think about moving this to `ethportal-api`?
For example here: https://github.com/ethereum/trin/tree/master/ethportal-api/src/types/state_trie

trin-execution/src/walkers/trie_walker.rs
KolbyML: It is an Arc of the traited object

KolbyML:
> Also, what do you think about moving this to `ethportal-api`? For example here: https://github.com/ethereum/trin/tree/master/ethportal-api/src/types/state_trie

I don't see value into moving this into `ethportal-api` as the only usecases are
- bridging content
- generating stats for bridging content

So I don't think it makes sense to store it in `ethportal-api`

trin-execution/src/walkers/memory_db.rs
```diff
@@ -0,0 +1,47 @@
...
+
+use anyhow::anyhow;
+use eth_trie::DB;
+use hashbrown::HashMap;
+use parking_lot::Mutex;
+use revm_primitives::B256;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum ReadOnlyMemoryDBError {
```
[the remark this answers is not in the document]
KolbyML: Now that we are using our own trait we can, the DB trait constrained us to using thiserror

trin-execution/src/walkers/memory_db.rs
```diff
@@ -0,0 +1,47 @@
...
+    #[error("read only memory db error {0}")]
+    ANYHOW(#[from] anyhow::Error),
+}
+
+#[derive(Debug)]
+pub struct ReadOnlyMemoryDB {
+    storage: Arc<Mutex<HashMap<B256, Vec<u8>>>>,
```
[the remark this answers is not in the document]
KolbyML: We don't need to now

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
+use std::sync::Arc;
+
+use alloy::consensus::EMPTY_ROOT_HASH;
+use eth_trie::{decode_node, node::Node, DB};
+use revm_primitives::{Bytes, B256};
+
+use crate::types::trie_proof::TrieProof;
+
+/// This is used for walking the whole state trie and partial tries (forward state diffs)
```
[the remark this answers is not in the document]
KolbyML: Ok I think I did what you asked

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+/// use cases are
+/// 1. gossiping the whole state trie
+/// 2. gossiping the forward state diffs
+/// 3. getting stats about the state trie
+pub struct TrieWalker<TrieDB: DB> {
+    is_partial_trie: bool,
+    trie: Arc<TrieDB>,
```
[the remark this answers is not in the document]
KolbyML: Yes becuase `trie.lock().db.clone()` gives us an Arc

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+        Ok(Self {
+            is_partial_trie: false,
+            trie,
+            stack: vec![root_proof],
+        })
+    }
+
+    pub fn new_partial_trie(root_hash: B256, trie: TrieDB) -> anyhow::Result<Self> {
+        // if the storage root is empty then there is no storage to gossip
+        if root_hash == EMPTY_ROOT_HASH {
```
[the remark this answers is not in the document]
KolbyML: It should work yeah

trin-execution/src/walkers/trie_walker.rs
KolbyML: So it is Arc<DB>

trin-execution/src/walkers/trie_walker.rs
```diff
@@ -0,0 +1,378 @@
...
+                // can't be a hash node because we just decoded it
+                Node::Empty | Node::Hash(_) => (),
+            }
+        }
+        Ok(())
+    }
+}
+
+impl<TrieDB: DB> Iterator for TrieWalker<TrieDB> {
+    type Item = TrieProof;
```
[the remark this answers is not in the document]
KolbyML: I added a commit for why it would panic, the only reason it would panic is if the trie database was corrupted or a database read failed, which at that point you wouldn't be able to continue anyways as both of those are unrecoverable states.

KolbyML:

trin-execution/src/trie_walker/mod.rs
```diff
@@ -98,23 +88,26 @@ impl<TrieDB: DB> TrieWalker<TrieDB> {
...
+            if matches!(
+                decode_node(&mut value.as_ref())?,
+                Node::Empty | Node::Hash(_)
+            ) {
+                return Ok(());
             }
+
+            self.stack.push(TrieProof {
+                path,
+                proof: [partial_proof, vec![value]].concat(),
```
morph-dev, line 102: I think there is no reason to use `concat()`, as I believe that it would iterate and clone partial_proof (it should be cheap, but still not needed). We can:

```rust
let mut proof = partial_proof
proof.push(value);
self.stack.push(TrieProof { path, proof });
```

or if you prefer:
```rust
partial_proof.push(value);
self.stack.push(TrieProof {
    path,
    proof: partial_proof,
});
```

trin-execution/src/trie_walker/mod.rs
```diff
@@ -98,23 +88,26 @@ impl<TrieDB: DB> TrieWalker<TrieDB> {
...
+            });
         }
         Ok(())
     }
 }
 
-impl<TrieDB: DB> Iterator for TrieWalker<TrieDB> {
+/// Panics if the trie is corrupted
```
morph-dev, line 109: This comment should probably be on `TrieWalker`, as I think this will never be shown in IDE or elsewhere.

trin-execution/src/trie_walker/mod.rs
```diff
@@ -170,126 +161,37 @@ impl<TrieDB: DB> Iterator for TrieWalker<TrieDB> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    use eth_trie::{EthTrie, RootWithTrieDiff, Trie};
+    use revm_primitives::{keccak256, Address, B256, U256};
```
morph-dev, line 166: nit: `alloy::primitives`

trin-execution/src/trie_walker/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+            };
+
+            let Node::Leaf(leaf) =
+                decode_node(&mut encoded_last_node.as_ref()).expect("Failed to decode node")
+            else {
+                continue;
+            };
+            leaf_count += 1;
+
+            // reconstruct the address hash from the path so that we can fetch the
```
morph-dev, line 211: nit: this comment is still not clear to me. what are we fetching from the database?
aren't we fetching proof directly from the trie?

trin-execution/src/trie_walker/mod.rs
```diff
@@ -0,0 +1,276 @@
...
+                panic!("Account proof is empty");
+            };
+
+            let Node::Leaf(leaf) =
+                decode_node(&mut encoded_last_node.as_ref()).expect("Failed to decode node")
+            else {
+                continue;
+            };
+
+            // reconstruct the address hash from the path so that we can fetch the
```
morph-dev, line 260: also here

morph-dev (approved):

morph-dev:
> I don't see value into moving this into `ethportal-api` as the only usecases are
> 
> * bridging content
> * generating stats for bridging content
> 
> So I don't think it makes sense to store it in `ethportal-api`.

Wouldn't it make sense to be in `portal-bridge`? Especially if you plan to add horizontal scaling.

I thought that `ethportal-api` might be okay and maybe glados can use it at some point.
But it's probably not the exact use case, and API might not work very well. So we can definitely wait until we get there and reevaluate.

KolbyML:
> > I don't see value into moving this into `ethportal-api` as the only usecases are
> > 
> > * bridging content
> > * generating stats for bridging content
> > 
> > So I don't think it makes sense to store it in `ethportal-api`.
> 
> Wouldn't it make sense to be in `portal-bridge`? Especially if you plan to add horizontal scaling.
I plan to use it in `Trin Execution` and `portal-bridge` will depends on `Trin Execution` so that would create a recursive dependency. Soo I would say no as the horizontal scaling part of the walker isn't related to the bridge it is more of just scoping the walked path.


> I thought that `ethportal-api` might be okay and maybe glados can use it at some point. But it's probably not the exact use case, and API might not work very well. So we can definitely wait until we get there and reevaluate.

Sure, I think a different design would work better for Glados's use case

Issue #1501, opened 2024-10-01
state-bridge: Census peer scoring


morph-dev:
## Problem

The Census only keeps track of all available peers, and their liveness is checked once per hour.

When peers are requested for a given content id, census filters out all peers whose radius doesn't include provided content, and returns first X (4 at the moment of writing).

Ideally, we should have some reputation or scoring system for the peers and use it when selecting them.

## High level idea

Firstly, we should still filter out peers that are not interested in the content (as we do now). Then, each peer will be assigned weight, and X peers will be selected randomly based on their weight.

The weight can be calculated as product of following: distance, reputation and recency weights

Note: All constants can be configurable parameters. Given values are picked by feeling and could be a good starting point (unless somebody has some other suggestions).

### Distance weight

This is the main weigh: peers should be selected based on their distance to the content.

```
distance_weight = u16::MAX - distance(peer_id, content_id).as_u16()
```

The `as_u16` function takes 16 most significant bits (2 bytes) of the `U256` distance metric

### Reputation weight

This is the main penalty weight: peers with recent failed attempts should be penalized.

```
reputation_weight = 1 / (2^recent_failed_attempts) as f64
```

The `recent_failed_attempts` is the number of recently failed OFFER requests to the peer. Only last 16 OFFER requests should be considered. Rejected offers shouldn't be counted as failures . 

Each failed attempt lowers the weight by factor of 2.

### Recency weight

This is the main recovery function: it "rewards" peers that weren't offered content in a long time.
Peers that were failing requests in the past but have recovered in the meantime would have hard time being selected, and this weight should help with that.

```
recency_weight = 2 ^ ((now - last_offer_timestamp).as_seconds() / 300)
```

For every 5 minutes that we didn't offer content to the peer, their weight increases by factor of 2

Assuming all peers are healthy, this shouldn't play big factor because of the high volume and random distribution of content ids (meaning all peers should be selected before this weight grows too big).

This only starts playing important role for peers that are penalized by `reputation_weight`, in which case it takes 5 minutes to recover one failed attempt.

## Future improvements

Some ideas omitted from the initial approach as I want to start with something simple and add more complexity afterwards if needed.

- Latency weight - reward/penalize peers based on their latency/bandwidth
- How to handle peers that seem offline (e.g. failed PING request)? Should we keep pinging them (e.g. with exponential backoff period) until we eventually drop their info.

Maliksb11:
ok

pipermerriam:
[...]

PR #1575, opened 2024-10-31
feat(census): keep track of liveness checks

morph-dev:
### What was wrong?

Census doesn't keep track of previous liveness checks or other request sent to peers.

Needed for: #1501 

### How was it fixed?

This PR does ground work focused on liveness checks, and followup PR will introduce other requests (that will later be used for peer scoring).

### DIFF
--- a/portal-bridge/src/census/network.rs
+++ b/portal-bridge/src/census/network.rs
@@ -23,15 +23,15 @@ use crate::{
 
 use super::peers::Peers;
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 /// The result of the liveness check.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum LivenessResult {
     /// We pinged the peer successfully
     Pass,
     /// We failed to ping peer
     Fail,
-    /// Peer is already registered and not expired (we didn't try to ping the peer)
-    AlreadyRegistered,
+    /// Peer is already known and doesn't need liveness check
+    Fresh,
 }
 
 #[derive(Debug, Clone)]
@@ -98,7 +98,7 @@ impl Network {
         NetworkManager::new(self.clone())
     }
 
-    // Look up all known interested enrs for a given content id
+    /// Look up interested enrs for a given content id
     pub fn get_interested_enrs(&self, content_id: &[u8; 32]) -> Result<Vec<Enr>, CensusError> {
         if self.peers.is_empty() {
             error!(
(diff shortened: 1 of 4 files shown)


morph-dev:
Followup PR: #1576

portal-bridge/src/census/peer.rs
```diff
@@ -0,0 +1,109 @@
...
+
+    pub fn record_successful_liveness_check(&mut self, enr: &Enr, radius: Distance) {
+        if self.enr.seq() > enr.seq() {
+            error!(
+                "successful_liveness_check: received outdated enr: {enr} (existing enr: {})",
+                self.enr.seq()
+            );
+        } else {
+            self.enr = enr.clone();
+        }
```
KolbyML, line 86:
```suggestion
        } else if self.enr.seq() < enr.seq()  {
            self.enr = enr.clone();
        }
```

it probably doesn't matter, but wouldn't we only need to do the clone if a updated enr was given, which should be fairly infrequent so we might as well avoid the clone

portal-bridge/src/census/network.rs
```diff
@@ -438,17 +430,15 @@ impl NetworkManager {
...
         for enr in enrs {
-            if self.network.liveness_check(enr).await == LivenessResult::Pass {
-                new_peers += 1;
-            }
+            self.network.liveness_check(enr).await;
         }
-
-        let total_peers = self.network.peers.len();
+        let ending_peers = self.network.peers.len();
+        let new_peers = ending_peers - starting_peers;
```
KolbyML, line 438: I am assuming the possibility that `new_peers` can be negative as a feature to display we lost peers? Because isn't there a case where `ending_peers` can be less then `starting_peers` i.e. if more peers were removed then added?

This doesn't really look like an issue, just wanted to point it out.

KolbyML (approved):
:shipit: looks good

portal-bridge/src/census/peer.rs
morph-dev: I changed function signature so it accepts `Enr`, in which case we never clone it.

portal-bridge/src/census/network.rs
morph-dev: I was thinking about it and concluded that it shouldn't happen. Unless there is some weird race-condition and/or Census is performing very badly.

But, you are right, it's technically possible. It's also possible that we change logic in the future and forget to update and this would panic (or difference no longer means that these are added).

So I updated the code to not panic and changed the message.

PR #1576, opened 2024-10-31
feat(census): keep track of offer requests and their status

morph-dev:
### What was wrong?

Census doesn't keep track of previous offer requests, their success, duration, etc.

Needed for #1501 

### How was it fixed?

Added recording information of those events. They will be used in followup PR for peer scoring.


KolbyML (approved):
:shipit:

PR #1577, opened 2024-10-31
fix: properly decode Attestations from json while decoding BeaconBlock

KolbyML:
### What was wrong?
When I was trying to decode a block from a CL client through the block endpoint it was failing to decode it as the `slot` and `index` fields in the JSON are u64, but encoded as a string.

We handle this higher up in our block decoding code it just seems to be forgotten here, when I added these 2 lines of serde decoding the blocks worked for me
### How was it fixed?
By adding the 2 serde lines which are used in other places to decode the u64 strings.


KolbyML:
> > We handle this higher up in our block decoding code it just seems to be forgotten here, when I added these 2 lines of serde decoding the blocks worked for me
> 
> Is this other handling redundant now? Can it be removed?

What I meant is `#[serde(deserialize_with = "as_u64")]` used in all the other structs for decoding consensus types, except here

KolbyML:
> Can you add a test that deserializes an attestation?

in a bit

KolbyML:
@carver ready for another review I added a test

ogenev (commented):
We already have two random `SignedBeaconBlock` yaml test files for Deneb in `test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/` and tests that encode/decode it. 

What we can do to cover the use case here is to update the attestation data fields to string in the yaml file (for example adding quotes [here](https://github.com/ethereum/trin/blob/bcb921d0402c97be7caab8aa1a7e8c9313190718/test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/case_0/value.yaml#L86)). Then the existing test should catch the string ser/de.

KolbyML:
> We already have two random `SignedBeaconBlock` yaml test files for Deneb in `test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/` and tests that encode/decode it.
> 
> What we can do to cover the use case here is to update the attestation data fields to strings in the yaml file (for example adding quotes [here](https://github.com/ethereum/trin/blob/bcb921d0402c97be7caab8aa1a7e8c9313190718/test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/case_0/value.yaml#L86)). Then the existing test should catch the string ser/de for those fields.

Done

test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/case_0/value.yaml
```diff
@@ -91,8 +91,8 @@ message:
         signature: '0xfb09452cee022c73ae85b69c45c7cb45a20c369eeb985c31183124acb770eaf77728708286c6e68b9da0a63b197843b31c9d40ea3637a39915dee862b75922a4c72d2647cea8441e68cb7b0dca52eff1c84b38cfc8a8cabf8250e34334ed34b4'
       - aggregation_bits: '0x6806'
         data:
-          slot: 4535242405270891733
-          index: 9093395152080509436
+          slot: "4535242405270891733"
+          index: "9093395152080509436"
```
ogenev, line 95: Those should be updated in other places below. Also, we need to do the same with the `case_1/value.yaml` file.

ethportal-api/src/types/consensus/beacon_block.rs
```diff
@@ -276,8 +276,10 @@ mod test {
         .expect("cannot find test asset");
         let value: Value = serde_yaml::from_str(&value).unwrap();
         let content: SignedBeaconBlockDeneb = serde_json::from_value(value.clone()).unwrap();
-        let serialized = serde_json::to_value(content).unwrap();
-        assert_eq!(serialized, value);
+        let serialized = serde_json::to_value(content.clone()).unwrap();
+        let deserialized_content: SignedBeaconBlockDeneb =
```
ogenev, line 280: This seems redundant with line 278.

KolbyML: Are we removing the assert then?

test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/case_0/value.yaml
KolbyML: I don't see why it should be updated everywhere

ethportal-api/src/types/consensus/beacon_block.rs
KolbyML: `fix: can't compare Value's now as one has a number the other has type String`

test_assets/beacon/deneb/SignedBeaconBlock/ssz_random/case_0/value.yaml
carver: If the attestation format is well-specified as always being a string, then we should have all our tests match the spec, and we don't need to cover the int case. If it's not well-specified and it comes in different types from different sources, then it makes sense to have a string and int version of every test case.

What's our go-to source for these specs? I see:
- https://www.quicknode.com/docs/ethereum/eth-v1-beacon-blocks-block_id-attestations which says string, but it's a 3rd-party source instead of a primary source
- https://github.com/ethereum/consensus-specs/blob/1b408e9354358cd7f883c170813e8bf93c922a94/specs/phase0/beacon-chain.md#attestationdata that seems to suggest an int value, but this isn't necessarily the same as the RPC spec

KolbyML: > If the attestation format is well-specified as always being a string, then we should have all our tests match the spec, and we don't need to cover the int case. If it's not well-specified and it comes in different types from different sources, then it makes sense to have a string and int version of every test case.
> 
> What's our go-to source for these specs? I see:
> 
> * https://www.quicknode.com/docs/ethereum/eth-v1-beacon-blocks-block_id-attestations which says string, but it's a 3rd-party source instead of a primary source
> * https://github.com/ethereum/consensus-specs/blob/1b408e9354358cd7f883c170813e8bf93c922a94/specs/phase0/beacon-chain.md#attestationdata that seems to suggest an int value, but this isn't necessarily the same as the RPC spec

The yaml format in `consensus-spec-tests` uses int's. Those yaml files are taken from `consensus-spec-tests`.

The beacon-api uses strings as JSON doesn't support u64.

So I don't see why we would deviate from `consensus-specs-repo` in how they define their format.

> then it makes sense to have a string and int version of every test case.

Then I will just re-add my json tests back, instead of modifying the yaml files

KolbyML: All the beacon yaml files use numbers instead of strings and we have over 10 yaml files at least from `consensus-spec-tests`

KolbyML:
I added a json test again.

ogenev (approved):
:shipit:

PR #1579, opened 2024-11-01
feat(census): add peer scoring

morph-dev:
### What was wrong?

Census doesn't use information about previous successful/failed requests when selecting peers (instead, it selects them randomly). 
Previous PRs made this information available.

Part of #1501

### How was it fixed?

Added `Weight` trait that can be used to calculate peers weights, and `PeerSelector` that picks peers based on weights.

Current implementation of `Weight` is based on [Piper's suggestion](https://github.com/ethereum/trin/issues/1501#issuecomment-2386493645):
- starting weight is 200
- we keep track of up to 255 latest requests per peer
- each successful request in the last 15 min increases weight by 5
- each failed request in the last 15 min decreases weight by 10
- final weight is restricted to [0, 400] range

In the future, we can implement multiple versions of `Weight`, make them configurable via cli flags, and compare them in practice.

### DIFF
--- a/portal-bridge/src/census/peer.rs
+++ b/portal-bridge/src/census/peer.rs
@@ -9,18 +9,18 @@ use tracing::error;
 
 #[derive(Debug, Clone)]
 pub struct LivenessCheck {
-    success: bool,
-    #[allow(dead_code)]
-    timestamp: Instant,
+    pub success: bool,
+    pub timestamp: Instant,
 }
 
-#[allow(dead_code)]
 #[derive(Debug, Clone)]
 pub struct OfferEvent {
-    success: bool,
-    timestamp: Instant,
-    content_value_size: usize,
-    duration: Duration,
+    pub success: bool,
+    pub timestamp: Instant,
+    #[allow(dead_code)]
+    pub content_value_size: usize,
+    #[allow(dead_code)]
+    pub duration: Duration,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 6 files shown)


portal-bridge/src/census/scoring.rs
```diff
@@ -0,0 +1,115 @@
...
+    pub fn select_peers<'a>(
+        &self,
+        content_id: &[u8; 32],
+        peers: impl IntoIterator<Item = &'a Peer>,
+    ) -> Vec<Enr> {
+        let weighted_peers = self.weight.weight_all(content_id, peers).collect_vec();
+
+        weighted_peers
+            .choose_multiple_weighted(&mut thread_rng(), self.limit, |(_peer, weight)| *weight)
+            .expect("choosing random sample shouldn't fail")
```
pipermerriam, line 111: Does this fail if there are zero peers?

morph-dev: No, it doesn't.
That can fail if weight is negative (api allows any type that can be converted to f64), which can't happen in our case because we use.

Weight 0 would also be picked, but only if there isn't enough weights that are non-zero, which is what we want.

One thing that we don't differentiate at the moment is that weight can be 0 if:
- content doesn't fall within radius
- peer has very negative weight

It might be better to never select peer whose radius doesn't include content. But I left that for future optimization

portal-bridge/src/census/peer.rs
```diff
@@ -9,18 +9,18 @@ use tracing::error;
...
-    success: bool,
-    timestamp: Instant,
-    content_value_size: usize,
-    duration: Duration,
+    pub success: bool,
+    pub timestamp: Instant,
+    #[allow(dead_code)]
+    pub content_value_size: usize,
+    #[allow(dead_code)]
+    pub duration: Duration,
```
KolbyML, line 23: Are there plans to use these in the future?

KolbyML (approved):
:shipit: looks good

portal-bridge/src/census/peer.rs
morph-dev: Yes.
Idea is to measure bandwidth and prefer peers that are faster than average.

But I didn't want to complicate initial work, and I don't think it works well with current implementation of "Weight" because peers that are working would quickly increase their weight to maximum and this wouldn't make a difference.

PR #1581, opened 2024-11-04
refactor: remove trin-utils dependency in ethportal-api

njgheorghita:
### What was wrong?
We can't release another `ethportal-api` version as long as it has a dependency on `trin-utils`

### How was it fixed?
Extracted relevant "version" utils into `ethportal-api`

### To-Do


KolbyML (approved):
:shipit:

PR #1591, opened 2024-11-27
fix: properly encode/decode `started_at_ms` in QueryTrace object

KolbyML:
### What was wrong?

fixes the second part of https://github.com/ethereum/portal-network-specs/issues/352

We were exposing the implementation details of `SystemTime` for the `started_at_ms` when encoding the `QueryTrace` object. This doesn't comply with our Json-RPC spec.

As `started_at_ms` is supposed to be encoded as a milisecond timestamp.
Json only allows for u64's which will last us for millions of years.

for example this is what we are currently returning before this fix

```rust
"startedAtMs": {
    "secs_since_epoch": 1730857625,
    "nanos_since_epoch": 447534202
}
```

but the result is supposed to be like
```rust
"startedAtMs": 1730857625000,
```

### How was it fixed?

- updating the returned type for `started_at_ms` to a u64
- adding a test to make sure things don't break in the future

### DIFF
--- a/ethportal-api/src/types/query_trace.rs
+++ b/ethportal-api/src/types/query_trace.rs
@@ -1,4 +1,7 @@
-use std::{collections::HashMap, time::SystemTime};
+use std::{
+    collections::HashMap,
+    time::{Duration, SystemTime, UNIX_EPOCH},
+};
 
 use alloy::primitives::B256;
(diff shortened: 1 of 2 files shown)


ogenev (approved):
LGTM 👍

PR #1592, opened 2024-11-27
feat: add state snapshot bridge

KolbyML:
### What was wrong?

We didn't have code to gossip a state snapshot from an Era2 file

### How was it fixed?

implementing a era2 snapshot bridging mode

### To-Do in future PR's

- add a feature to only gossip a portion of the snapshot so that bridging the snapshot can be horizontally scaled

### The Plans for the latest state bridging

Currently Trin Execution is used as a dependency for `portal-bridge`.

Since `Trin-Execution` will bridge latest diffs by default, and will also eventually run Trin as a library by default. I think I am going to move the State Bridge code into Trin Execution.

So `Trin Execution` will then depend on the `portal-bridge` crate then. Bridging state depends on Trin Execution heavily, whereas the History and Beacon bridge use a JSON-RPC provider. Currently only `portal-bridge` beacon and history can be ran together whereas state must be ran alone regardless.

It is important to note as well the state bridge doesn't rely on using a 3rd party JSON-RPC provider like the History and Beacon network do.

So since `Trin-Execution` will be bridging by default (if a users uses Trin Execution it will gossip a small random section of the state diff), It is my opinion that all state bridging belongs in Trin Execution. I don't want to support latest bridging in both `trin-execution` and `portal-bridge`. Supporting latest bridging in `portal-bridge` would just be running `trin-execution` either way so it doesn't really make any sense, when it would be better UX to just use `trin-execution` to bridge directly.

After I finish `Trin execution keeping up with latest,` there will be quite a few PRs to review.

By default, `trin-execution` will gossip a small amount of the latest state, but you can configure it to do more and horizontally bridge and such. That is my current plan.

It is also always for
- better UX, as the commandline options be more focused
- Trin Execution is becoming more service-based; having the state bridge a part of trin-execution simplifies logic and reduces code duplication.
- to put it plainly, it would simplify things.

### DIFF
--- a/e2store/src/utils.rs
+++ b/e2store/src/utils.rs
@@ -4,6 +4,7 @@ use anyhow::{ensure, Error};
 use rand::{seq::SliceRandom, thread_rng};
 use reqwest::Client;
 use scraper::{Html, Selector};
+use url::Url;
 
 const ERA_DIR_URL: &str = "https://mainnet.era.nimbus.team/";
 const ERA1_DIR_URL: &str = "https://era1.ethportal.net/";
@@ -19,10 +20,9 @@ pub fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
     None
 }
 
-pub async fn download_era_links(
+async fn download_era_links(
     http_client: &Client,
     url: &str,
-    append_base_url: bool,
 ) -> anyhow::Result<HashMap<u64, String>> {
     let index_html = http_client.get(url).send().await?.text().await?;
     let index_html = Html::parse_document(&index_html);
@@ -40,21 +40,19 @@ pub async fn download_era_links(
                 .expect("to be able to get epoch")
                 .parse::<u64>()
                 .expect("to be able to parse epoch");
-            (
-                epoch_index,
-                if append_base_url {
-                    format!("{}{}", url, href)
-                } else {
-                    href.to_string()
-                },
-            )
+            let url = Url::parse(url)
+                .and_then(|url| url.join(href))
+                .unwrap_or_else(|_| {
+                    panic!("to construct valid url from base ({url}) and href ({href}).")
+                });
+            (epoch_index, url.to_string())
         })
         .collect();
     Ok(era_files)
 }
 
 pub async fn get_era_files(http_client: &Client) -> anyhow::Result<HashMap<u64, String>> {
-    let era_files = download_era_links(http_client, ERA_DIR_URL, true).await?;
+    let era_files = download_era_links(http_client, ERA_DIR_URL).await?;
     ensure!(!era_files.is_empty(), "No era files found at {ERA_DIR_URL}");
     let missing_epochs: Vec<String> = (0..era_files.len())
         .filter(|&epoch_num| !era_files.contains_key(&(epoch_num as u64)))
@@ -70,7 +68,7 @@ pub async fn get_era_files(http_client: &Client) -> anyhow::Result<HashMap<u64, 
 }
 
 pub async fn get_era1_files(http_client: &Client) -> anyhow::Result<HashMap<u64, String>> {
-    let era1_files = download_era_links(http_client, ERA1_DIR_URL, true).await?;
+    let era1_files = download_era_links(http_client, ERA1_DIR_URL).await?;
     ensure!(
         era1_files.len() == ERA1_FILE_COUNT,
         format!(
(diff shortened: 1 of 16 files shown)


trin-execution/src/subcommands/era2/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+        let chunk = chunk?;
+        file.write_all(&chunk)?;
+        downloaded += chunk.len() as u64;
+
+        // Check if enough time has passed since the last report
+        if last_report_time.elapsed() >= Duration::from_secs(2) {
+            // Report every 2 seconds
+            // Calculate elapsed time
+            let elapsed = start_time.elapsed().as_secs_f64();
+            let speed = downloaded as f64 / elapsed; // Bytes per second
```
morph-dev, line 65: Wouldn't speed be average speed of the whole download? I think it would be more useful to have current download speed (e.g. average speed since last printed message, i.e. over 2 sec).

trin-execution/src/subcommands/era2/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+            let remaining_time = if total_size > 0 {
```
morph-dev, line 66: Don't we already have a check that `total_size` is greater than zero? Can we simplify this logic?

trin-execution/src/subcommands/era2/utils.rs
```diff
@@ -0,0 +1,116 @@
...
+                let remaining_bytes = total_size.saturating_sub(downloaded) as f64;
+                Some(remaining_bytes / speed)
+            } else {
+                None
+            };
+
+            // Log progress with estimated time remaining
+            if total_size > 0 {
```
morph-dev, line 74: also here

morph-dev (commented):
In general, it looks good.

I have one idea that I would like to discuss.

Technically, we don't have to fully download the era2 file, before we start processing it. Yes, we would have to refactor e2store stream reader to support this, but I don't think it's too much work.

Downside is that doing it that way might is more prone to error that are harder to recover (e.g. losing connection), but upside is that it can improve performance.

What do you think?

Of course, this doesn't (shouldn't) be done as part of this PR. I just wanted to start  the discussion.

KolbyML:
> Technically, we don't have to fully download the era2 file, before we start processing it. Yes, we would have to refactor e2store stream reader to support this, but I don't think it's too much work.
> 
> Downside is that doing it that way might is more prone to error that are harder to recover (e.g. losing connection), but upside is that it can improve performance.
> 
> What do you think?
> 
> Of course, this doesn't (shouldn't) be done as part of this PR. I just wanted to start the discussion.

Currently from my testing of running the snapshot bridge at block 21 million

- downloading the era2 file takes 1 hour
- importing the era2 file takes ~24 hours
- seeding the snapshot takes 1,200 days

So this optimization would only realistically save 1 hour at most, so I don't think it would be a priority for a very long time.

If we wanted to make optimizations I think focusing on `seeding the snapshot takes 1,200 days` as it takes 28800 times hours in comparison.


So with the
- higher complexity to implement that optimization
- not being the bottleneck

I don't see the benefit in making this change for the time being, as there are other tasks which ship the state network or in other words help us reach our MVP's for state faster.

e2store/src/utils.rs
```diff
@@ -21,6 +22,7 @@ pub fn underlying_io_error_kind(error: &Error) -> Option<io::ErrorKind> {
 pub async fn download_era_links(
     http_client: &Client,
     url: &str,
+    append_base_url: bool,
```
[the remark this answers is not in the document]
KolbyML: Very cool!

trin-execution/src/subcommands/era2/utils.rs
KolbyML: I removed the 1 unneeded check, the first one is a warning. The last one is needed

trin-execution/src/subcommands/era2/utils.rs
KolbyML: Addressed in comment above

e2store/src/utils.rs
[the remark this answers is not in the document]
KolbyML: Clippy gave me an error and it told me to do

```rust
            let url = Url::parse(url)
                .and_then(|url| url.join(href))
                .unwrap_or_else(|_| {
                    panic!("to construct valid url from base ({url}) and href ({href}).")
                });
                
```

instead

KolbyML:

portal-bridge/README.md
```diff
@@ -44,6 +44,7 @@ cargo run -p portal-bridge -- --executable-path ./target/debug/trin --epoch-accu
     - before gossiping a individual piece of content, the bridge will perform a lookup to see if the content is already in the portal network. If it is, the content will not be gossiped.
 - `"--mode fourfours:single_hunter:10:50`: sample size = 10, threshold = 50
     - same as the above hunter mode, but it will only gossip a single era1 file before exiting
+- `"--mode snapshot:1000000"`: gossips a state snapshot at the respective block, in this example the state snapshot at block 1,000,000 will be gossiped. This mode is only used for the State Network.
```
morph-dev, line 47: Shouldn't this be under State Subnetwork ?

morph-dev (approved):

PR #1598, opened 2024-12-05
feat: add Filter to TrieWalker

KolbyML:
### What was wrong?
Walking a whole state trie is very time consuming
### How was it fixed?

By making it so you can filter the trie to only walk a slice of it, this PR will be used in my followup PR, which adds a random walking state snapshot mode.

This mode will be useful for the random slice state snapshot bridge mode I will be adding

### DIFF
--- a/trin-execution/src/trie_walker/filter.rs
+++ b/trin-execution/src/trie_walker/filter.rs
@@ -1,0 +1,72 @@
+use alloy::primitives::{B256, U256};
+use rand::{thread_rng, Rng};
+
+use crate::utils::partial_nibble_path_to_right_padded_b256;
+
+#[derive(Debug, Clone)]
+pub struct Filter {
+    start_prefix: B256,
+    end_prefix: B256,
+}
+
+impl Filter {
+    pub fn new_random_filter(slice_count: u16) -> Self {
+        // if slice_count is 0 or 1, we want to include the whole trie
+        if slice_count == 0 || slice_count == 1 {
+            return Self {
+                start_prefix: B256::ZERO,
+                end_prefix: B256::from(U256::MAX),
+            };
+        }
+
+        let slice_size = U256::MAX / U256::from(slice_count);
+        let random_slice_index = thread_rng().gen_range(0..slice_count);
+
+        let start_prefix = U256::from(random_slice_index) * slice_size;
+        let end_prefix = if random_slice_index == slice_count - 1 {
+            U256::MAX
+        } else {
+            start_prefix + slice_size - U256::from(1)
+        };
+
+        Self {
+            start_prefix: B256::from(start_prefix),
+            end_prefix: B256::from(end_prefix),
+        }
+    }
+
+    pub fn is_included(&self, path: &[u8]) -> bool {
+        let path = partial_nibble_path_to_right_padded_b256(path);
+        (self.start_prefix..=self.end_prefix).contains(&path)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_new_random_filter() {
+        let filter = Filter::new_random_filter(0);
+        assert_eq!(filter.start_prefix, B256::ZERO);
+        assert_eq!(filter.end_prefix, B256::from(U256::MAX));
+
+        let filter = Filter::new_random_filter(1);
+        assert_eq!(filter.start_prefix, B256::ZERO);
+        assert_eq!(filter.end_prefix, B256::from(U256::MAX));
+    }
+
+    #[test]
+    fn test_is_included() {
+        let filter = Filter {
+            start_prefix: partial_nibble_path_to_right_padded_b256(&[0x1]),
+            end_prefix: partial_nibble_path_to_right_padded_b256(&[0x3]),
+        };
+
+        assert!(!filter.is_included(&[0x00]));
+        assert!(filter.is_included(&[0x01]));
+        assert!(filter.is_included(&[0x02]));
+        assert!(filter.is_included(&[0x03]));
+        assert!(!filter.is_included(&[0x04]));
+    }
+}
--- a/trin-execution/src/trie_walker/filter.rs
+++ b/trin-execution/src/trie_walker/filter.rs
@@ -5,8 +5,8 @@ use crate::utils::partial_nibble_path_to_right_padded_b256;
 
 #[derive(Debug, Clone)]
 pub struct Filter {
-    start_prefix: B256,
-    end_prefix: B256,
+    start_prefix: U256,
+    end_prefix: U256,
 }
 
 impl Filter {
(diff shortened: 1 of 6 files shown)


trin-execution/src/trie_walker/filter.rs
```diff
@@ -0,0 +1,120 @@
...
+        };
+
+        Self {
+            start_prefix,
+            end_prefix,
+        }
+    }
+
+    fn partial_prefix(prefix: U256, shift_amount: usize) -> B256 {
+        prefix
```
morph-dev, line 39: Any reason not to do one of the following:
1. `prefix | (U256::MAX << shift_amount)` - I think this is the cleanest
    - I would go one step further and say that `shift_amount` is not clear what it means. Instead, I would use opposite value:
    - ```rust
        let prefix_length = 4 * path.len();
        let partial_prefix_mask = !(U256::MAX >> prefix_length);
        let partial_start_prefix = self.start_prefix & partial_prefix_mask;
        let partial_end_prefix = self.end_prefix & partial_prefix_mask;
        ```
2. `(prefix >> shift_amount) << shift_amount` - same idea as yours, but simpler

And in either of those cases, this is simple enough and I would inline it below.

trin-execution/src/trie_walker/filter.rs
```diff
@@ -0,0 +1,120 @@
+use alloy::primitives::{B256, U256};
+use rand::{thread_rng, Rng};
+
+use crate::utils::partial_nibble_path_to_right_padded_b256;
+
+#[derive(Debug, Clone)]
+pub struct Filter {
+    start_prefix: U256,
```
morph-dev, line 8: nit: why do we call this `_prefix`? Wouldn't `start` and `end` be sufficient and more correct?

trin-execution/src/utils.rs
```diff
@@ -8,15 +20,11 @@ pub fn full_nibble_path_to_address_hash(key_path: &[u8]) -> B256 {
...
-            raw_address_hash.push(key_path[i] << 4);
-        } else {
-            raw_address_hash[i / 2] |= key_path[i];
-        }
-    }
-    B256::from_slice(&raw_address_hash)
+    B256::from_slice(&compress_nibbles(key_path))
+}
+
+pub fn partial_nibble_path_to_right_padded_b256(partial_nibble_path: &[u8]) -> B256 {
```
morph-dev, line 26: Not a big fan of converting them to `Vec` and then to `B256`.

How about:
```rust
pub fn nibbles_to_right_padded_b256(nibbles: &[u8]) -> B256 {
    let mut result = B256::ZERO;
    for (i, nibble) in nibbles.iter().enumerate() {
        if i % 2 == 0 {
            result[i / 2] |= nibble << 4;
        } else {
            result[i / 2] |= nibble;
        };
    }
    result
}
```

We also don't need `compress_nibbles` and we can modify `full_nibble_path_to_address_hash` to:
```rust
pub fn full_nibble_path_to_address_hash(key_path: &[u8]) -> B256 {
    if key_path.len() != 64 {
        panic!(
            "Key path should always be 64 bytes long: {}",
            key_path.len()
        );
    }

    nibbles_to_right_padded_b256(key_path)
}
```

morph-dev (commented):
Few minor comments. Nothing deal breaking but I would say they are nice improvements.

morph-dev:
Actually, there is one thing, and we can decide if it is a problem or not.

Let's say we have two Filters ( `slice_count` is 512)
- `FilterA { start: 0xab10..00, end: 0xab7f..ff }`
- `FilterB { start: 0xab80..00, end: 0xabff..ff }`

And let's say that trie node at path `[0xa, 0xb]` is a Leaf node, with prefix: `[0xc, 0xd, ...]` (making entire path for the data stored inside the leaf node: `0xabcd...`).

Intuitively, one would expect that this trie node (since it's leaf) would be visited only by TrieWalker that has FilterB, and not be visited by TrieWalker with FilterA. But it would actually be visited by both of them.

I personally don't think this is an issue. But I wanted to raise awareness and get your opinion as well.

trin-execution/src/trie_walker/filter.rs
KolbyML: I choose 2 as it is a simpler to understand

trin-execution/src/trie_walker/filter.rs
KolbyML: sure

KolbyML:
> Actually, there is one thing, and we can decide if it is a problem or not.
> 
> Let's say we have two Filters ( `slice_count` is 512)
> 
> * `FilterA { start: 0xab10..00, end: 0xab7f..ff }`
> * `FilterB { start: 0xab80..00, end: 0xabff..ff }`
> 
> And let's say that trie node at path `[0xa, 0xb]` is a Leaf node, with prefix: `[0xc, 0xd, ...]` (making entire path for the data stored inside the leaf node: `0xabcd...`).
> 
> Intuitively, one would expect that this trie node (since it's leaf) would be visited only by TrieWalker that has FilterB, and not be visited by TrieWalker with FilterA. But it would actually be visited by both of them.
> 
> I personally don't think this is an issue. But I wanted to raise awareness and get your opinion as well.

I made Filter more inclusive, because I thought it was simpler to reason about that way. I would rather the implementation be more inclusive instead of by chance having an edge case and miss something if I made a mistake in my implementation.

KolbyML:

trin-execution/src/trie_walker/filter.rs
```diff
@@ -1,120 +1,131 @@
...
+    start: U256,
+    end: U256,
 }
 
 impl Filter {
-    pub fn new_random_filter(slice_count: u16) -> Self {
+    /// Create a new filter that includes the whole trie
+    /// Slice index must be less than slice count or it will panic
+    pub fn new(slice_index: u16, slice_count: u16) -> Self {
+        assert!(
```
morph-dev, line 16: This condition would fail if `slice_count` is zero (which seems to be supported case otherwise).

trin-execution/src/trie_walker/filter.rs
```diff
@@ -1,120 +1,131 @@
...
-        let filter = Filter::new_random_filter(1);
-        assert_eq!(filter.start_prefix, U256::ZERO);
-        assert_eq!(filter.end_prefix, U256::MAX);
+        let filter = Filter::random(1);
+        assert_eq!(filter.start, U256::ZERO);
+        assert_eq!(filter.end, U256::MAX);
     }
 
     #[test]
     fn test_is_included() {
```
morph-dev, line 82: nit: should be renamed to `contains`

trin-execution/src/trie_walker/filter.rs
```diff
@@ -1,120 +1,131 @@
...
 }
 
 #[cfg(test)]
 mod tests {
-    use alloy::hex::FromHex;
-
     use super::*;
 
     #[test]
     fn test_new_random_filter() {
```
morph-dev, line 71: nit: names of tests usually don't start with `test_` (it's redundant since they are already inside `tests` module)

morph-dev (approved):
:shipit:

Issue #1596, opened 2024-12-03
Content that is available in the network "not found" depending on network key / node id


kdeme:
This is an issue that I first encountered when running a Fluffy node, but the same is reproducible with a Trin node.

### The issue:
For some content key(s) there are deterministic failures when doing a `RecursiveFindContent` from a node with a specific `NodeId` / network key.  This traces back to the node(s) that store the content failing in setting up a uTP connection to retrieve the data.
The node that is looking for the content does end up requesting it to the node that has the content.

When then directly targeting the node that has the content with a `FindContent` request, this will fail.
Changing network key and running the same call will most likely work (unless perhaps you hit by chance another NodeId or NodeId range that does not work? I'm not sure about this exactly as I don't understand what causes this in the first place).

### How to reproduce the issue
Here is a way to reproduce this for one specific content key / content id combo. 

1. Build Trin: `cargo build`
2. Run trin with specific network key: `cargo run -- --web3-transport http --unsafe-private-key "0x24b336eb9522dbe3afa8f31a2ec277333538ff211480d32633c45c29477bebe8"`
3. Request specific block with `FindContent` method: `curl -s -X POST -H 'Content-Type: application/json' -d '{"jsonrpc":"2.0","id":"1","method":"portal_historyFindContent","params":["enr:-LS4QJi4_7PnH3BsKPej9J8--A1AWczIYojJIU_NWdWCIgFbflGJ4f4x35InkIMyQuBdz_XCaNx44V2X1O-dw_zprFKEZyImyGOqdCA2ZmQ0NzRmYWQ2MjNhNWY2MzVmOTFiOWVmZGE2MjI3NmJiYjE2MmZkgmlkgnY0gmlwhI_0qIWJc2VjcDI1NmsxoQIIu0viHReNIKnjhuSCXDMcC6TLKEDF5oMmqogiPISN4oN1ZHCCIzE", "0x018a24f51c42f5c1e216351c6c2ab29d2ae25fc4f366ea690a4e13c640844412e7"]}' http://localhost:8545 | jq`

For Trin this will result in this, not so great, error message:
```
    "message": "Failed to parse error message from history subnetwork: Error(\"expected value\", line: 1, column: 1)"
```

But if you look at the Trin node logs, there is also this error:
```
[...]

PR #1600, opened 2024-12-07
fix: add debug logs to OverlayService::find_enr

morph-dev:
### What was wrong?

It seems that `OverlayService::find_enr` sometimes finds wrong Enr.

See https://github.com/ethereum/trin/issues/1596#issuecomment-2525302159 for more details

### How was it fixed?

This isn't a fix. I'm just adding extra logging to confirm my suspicious and try to find out the root cause of the problem.

### To-Do


KolbyML (approved):
:shipit:

PR #1601, opened 2024-12-07
fix: fallback to custom error if we can't parse ContentNotFound error

morph-dev:
### What was wrong?

The error message of failed JSON response can contain `"Unable to locate content on the network"` text, but error itself is not `ContentNotFound` error.

This can happen if `ContentNotFound` error is nested inside some other message.

Currently, returned response doesn't contain details of the original failure (instead it tells us why parsing failed, in not so useful way).

### How was it fixed?

If parsing `ContentNotFoundJsonError` fails, fallback to generic error message instead.

### To-Do


KolbyML (approved):
:shipit:

PR #1603, opened 2024-12-08
style: enable fmt linting for grouped imports by std external and crates

KolbyML:
### What was wrong?

We don't have the fmt linting option for the import format we all use.

The format we normally use as a team is

-1 std imports grouped up
-2 external imports grouped up
-3 crates imports grouped up

### How was it fixed?

It turns out fmt has a flag for enabling this formatting so I enabled the lint option in rustfmt.toml

### DIFF
--- a/book/src/developers/contributing/rust/imports.md
+++ b/book/src/developers/contributing/rust/imports.md
@@ -6,7 +6,7 @@
   3. self, super and crate imports.
 - Alphabetize imports in `Cargo.toml`
 
-```rust
+```rust,ignore
 use alloc::alloc::Layout;
 use core::f32;
 use std::sync::Arc;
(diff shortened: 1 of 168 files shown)


morph-dev (approved):
Rubber stamp approval (checked dozens of files and they look fine).

:shipit: 

---

Sidenote: Our [trin book page](https://ethereum.github.io/trin/developers/contributing/rust/imports.html) says that groups should be:
- _Imports from 'std'_
- _Imports from external crates_
- _Imports from crates within trin_

I think this can be interpreted that other _trin_ crates should be imported inside 3rd group, something like this:

```rust
// e.g. inside trin-storage/src
use alloy::{...};

use crate::{..};
use ethportal_api::{..};
```

But in reality, `ethportal_api` would be considered external crate, and be part of the 2nd group.

I think that we can't setup linter to do  the above, so maybe while you are at it, can you update docs page to clarify.

PR #1606, opened 2024-12-12
docs: publishing ethportal-api, plus a Cargo.lock bump

carver:


book/src/developers/contributing/releases/release_checklist.md
```diff
@@ -15,17 +15,36 @@ anything.
...
+```bash
+cd ethportal-api
+cargo publish --no-verify
+```
+
+We would like to get rid of the no-verify ASAP, but for now Cargo.lock is
+causing issues that we have no other workaround for. `cargo publish` generates
+a new lock file, and then complains that the lockfile is new. See this
+[StackOverflow
+post](https://stackoverflow.com/questions/79276315/how-to-build-a-cargo-lock-file-for-a-package-within-a-workspace).
```
KolbyML, line 37:
```suggestion
[StackOverflow post](https://stackoverflow.com/questions/79276315/how-to-build-a-cargo-lock-file-for-a-package-within-a-workspace).
```

It is a little weird that these are on 2 different lines

KolbyML (approved):
:shipit: looks good

book/src/developers/contributing/releases/release_checklist.md
carver: yeah, vim just did that with gqq, but I don't mind manually changing it

PR #1609, opened 2024-12-17
fix: trin cli flag displays wrong version info

njgheorghita:
### What was wrong?
https://github.com/ethereum/trin/pull/1581#issuecomment-2455933987

### How was it fixed?
After a little research into our options, I agree with @carver 's suggestion to just remove the version dependency inside `ethportal-api`. There might be an option with configuring the `PROJECT_ROOT` based on a flag, but that route introduces some management overhead that IMO is not worthwhile. This solution plays nicely with docker and all our current config (afaik)... and I'm not sure it's a huge loss to remove the `VERSION` info from the cli. The only side affect that I can tell is that running
```
cargo run -p trin -- --version
```
will just return the string I replaced the version with, and if somebody needs to know the version, running the client quickly seems like a reasonable compromise.

### To-Do


KolbyML (dismissed):
:shipit: we can do this for now if it is blocking anything.

I want to remove the `shadow-rs` stuff and replace it with a different solution as I don't think `shadow-rs` is needed to properly do versioning.

I also figured out how to make it so we report the right version for Trin which is nice. It is a little to much to discuss on this PR, but I need it for ping extensions. So I will probably make a PR after this one

KolbyML:
I made a PR to supersede this one https://github.com/ethereum/trin/pull/1615

the `shadow-rs` code doesn't properly report the clients version regardless. I would argue it is broken. I need working version code for my `ping extensions` PR. 

My PR above fixes the publish code, and gives the desired commit hash result, and will easily allow me to do what I need for my Ping Extensions PR.

I am not trying to step on any toes or one-up or anything. I feel a little awkward right now.

KolbyML (changes requested):
I made a PR to supersede this one https://github.com/ethereum/trin/pull/1615

the `shadow-rs` code doesn't properly report the clients version regardless. I would argue it is broken. I need working version code for my `ping extensions` PR. 

My PR above fixes `publishing` not working, and gives the desired commit hash result and Trin version, and will easily allow me to do what I need for my Ping Extensions PR.

We don't need to break `cargo run -p trin -- --version` and I don't want the version code to be moved to `trin_utils` since I need it for my `ping extension PR`. My PR resolves https://github.com/ethereum/trin/pull/1581#issuecomment-2455933987 this comment, well still keeping Trin's version information accessible in `ethportal-api`

PR #1614, opened 2024-12-26
refactor: create separate binary crates and move TrinConfig to trin crate

KolbyML:
### What was wrong?

https://github.com/ethereum/trin/pull/1609 trin reporting versioning is broken. I need versioning working for ping extensions.

This is one PR, but I am going to make another PR fixing versioning.

One issue was that, TrinConfig shouldn't be in Ethportal-api and Trin's binary shouldn't be defined in the global cargo file

This causes a recursive import in ethportal-peertests.

Anyways I am making this PR for multiple reasons, I just haven't felt enough pain to clean things up until now

### How was it fixed?
- move `src/bin/*`'s into separate crates
- move TrinConfig into Trin's binaries crate. It really shouldn't be in ethportal-api


ethportal-api/Cargo.toml
```diff
@@ -2,7 +2,6 @@
 name = "ethportal-api"
 version = "0.4.0"
 description = "Definitions for various Ethereum Portal Network JSONRPC APIs"
-build = "build.rs"
```
ogenev, line 5: Why the removal of this? In vergen docs it is required: https://docs.rs/vergen/latest/vergen/#usage

ethportal-api/src/types/cli.rs
ogenev, line 1: Why are we leaving this file in `ethportal-api` and we are not moving all helper methods to `trin/src/cli.rs` ?

trin-storage/src/config.rs
```diff
@@ -1,14 +1,28 @@
...
+use ethportal_api::types::network::Subnetwork;
 use r2d2::Pool;
 use r2d2_sqlite::SqliteConnectionManager;
 
 use crate::{error::ContentStoreError, utils::setup_sql, DistanceFunction};
 
 const BYTES_IN_MB_U64: u64 = 1000 * 1000;
 
+/// The storage capacity configurtion.
+#[derive(Debug, Clone, PartialEq, Eq)]
```
ogenev, line 13: What is the reason for this change?

bin/trin/src/cli.rs
```diff
@@ -0,0 +1,1141 @@
...
+    for template_path in DASHBOARD_TEMPLATES {
+        let dashboard_url = grafana.create_dashboard(template_path, &prometheus_uid)?;
+        println!("Dashboard successfully created: {dashboard_url}");
+    }
+
+    Ok(())
+}
+
+impl TrinConfig {
+    pub fn portalnet_config(&self, private_key: B256) -> PortalnetConfig {
```
ogenev, line 464: Not sure if I like this pattern, I prefer implementing the From trait for `PortalnetConfig` and `RpcConfig`.

```rust
impl From<TrinConfig> for PortalnetConfig {
...
}

impl From<TrinConfig> for RpcConfig {
...
}

```

rpc/src/config.rs
```diff
@@ -0,0 +1,13 @@
+use std::{path::PathBuf, sync::Arc};
+
+use alloy::transports::http::reqwest::Url;
+use ethportal_api::types::{cli::Web3TransportType, network::Subnetwork};
+
+pub struct RpcConfig {
```
ogenev, line 6: nit: a docstring is always nice

ogenev (commented):
Having the trin cli config in `ethportal-api` is something that bugged me for a long time, great!

ethportal-api/Cargo.toml
KolbyML: I think we should discuss this in this PR https://github.com/ethereum/trin/pull/1615

I did some testing and if you are building it doesn't seem required. I also checked other projects which used vergen such as Reth and they removed `build = "build.rs"` as well.

`build = "build.rs"` isn't friendly with publishing crates to crates.io and it doesn't seem needed for our usecase.

ethportal-api/src/types/cli.rs
KolbyML: portal-bridge is using these helper functions too, maybe we can move them to trin-utils instead? What do you think?

bin/trin/src/cli.rs
KolbyML: I can do From for `RpcConfig`, but I can't for `impl From<TrinConfig> for PortalnetConfig {` as it requires the `private_key: B256` argument

KolbyML:
@ogenev ready for another review, I want to review this PR first https://github.com/ethereum/trin/pull/1615

bin/trin/src/cli.rs
ogenev: OK, we can then change the function name to `to_portalnet_config` because we are not consuming `self` and will follow the naming conventions.

ethportal-api/Cargo.toml
ogenev: Fair enough 👍 .

ethportal-api/src/types/cli.rs
ogenev: Hmm, yeah, if more than one crate uses them, `trin-utils` seems a reasonable candidate.

ogenev (approved):
Left a comment for moving the helper functions to `trin-utils`.

Anyways, the PR looks good 👍 .

Cargo.toml
```diff
@@ -1,62 +1,11 @@
...
-ethportal-peertest = { path = "ethportal-peertest" }
-serde_yaml.workspace = true
-serial_test.workspace = true
-ureq.workspace = true
-
 [workspace]
 members = [
+    "bin/historical-batch", 
+    "bin/poll-latest", 
+    "bin/purge-invalid-history-content",
```
njgheorghita, line 5: hmmm. now it seems like the overhead for including these single-task binaries is higher than before. given that we don't really use these in production, I'd be in favor of removing all of them, except for one, to leave a framework for how to do it in future scenarios

KolbyML: ok which one do you want to keep

KolbyML: For tooling we could make x,y,z subcommands in the trin cli, which would make there versions be tied. I think the purge script makes sense as a Trin sub command specifically

bin/trin/src/cli.rs
```diff
@@ -7,75 +7,39 @@ use clap::{
     error::{Error, ErrorKind},
     Args, Parser, Subcommand,
 };
-use url::Url;
-
-use crate::{
-    types::{bootnodes::Bootnodes, network::Subnetwork},
-    version::FULL_VERSION,
+use ethportal_api::{
```
njgheorghita, line 10: I feel like this refactor should have removed some dependencies from `ethportal-api`? At least `clap` ... maybe `url`?

Cargo.toml
njgheorghita: Idk... honestly does it make sense to keep any of them around? None of them are used / have been in use recently afaik. And maybe we should migrate these kinds of tasks to the "subcommand" pattern, which might enforce a bit more of a thorough design/plan rather than these single-use scripts, which are a bit hacky....

njgheorghita (approved):

Cargo.toml
KolbyML: Ok I will remove them all then, as I don't have any specific attachment to them either

bin/trin/src/cli.rs
KolbyML: Good call!

njgheorghita (commented):
You'll also need to remove the refs to the deleted scripts in the dockerfiles

KolbyML:
> You'll also need to remove the refs to the deleted scripts in the dockerfiles

good catch

PR #1615, opened 2024-12-26
fix: trin cli flag displays wrong version info

KolbyML:
### What was wrong?
This PR supercedes https://github.com/ethereum/trin/pull/1609

***I need to properly get the client version and such for my ping extensions PR so I am inclined to get fetching the version and commit properly***

Basically when you build Trin it will report the wrong commit hash. And the shadow setup breaks publishing the `ethportal-api` crate.

I fixed the problem by removing shadow and using `vergen` instead.

I ran `cargo publish --dry-run` and it ran successfully so publish works with this PR

I also ran `cargo run -p trin -- --version`

and it works
I get `trin b123384e`

which is the commit of the build which is good that is what we want

image

### How was it fixed?

using vergen


ethportal-api/build.rs
```diff
@@ -1,36 +1,15 @@
...
-        "unknown"
-    }
-}"#;
-
-    writeln!(file, "{}", hook_fn)?;
+use std::error::Error;
+
+use vergen::EmitBuilder;
+
+fn main() -> Result<(), Box<dyn Error>> {
```
njgheorghita, line 5: I think you're going to either have to set a `GIT_HASH` env var here, or update our 2 dockerfiles so that it'll pick up the env var [here](https://github.com/ethereum/trin/blob/master/docker/Dockerfile#L11)

KolbyML: https://github.com/sigp/lighthouse/blob/stable/Dockerfile
https://github.com/paradigmxyz/reth/blob/main/Dockerfile

both Lighthouse and Reth copy the whole folder, if we do that it would resolve this issue and we could remove

https://github.com/ethereum/trin/blob/ddbd9ff54df221c2ccf6967fab8340553896b723/docker/Dockerfile#L8-L11

and passing the git hash in with the CI, 

As only copying in the crates would be problematic for vergen

Reth uses this https://github.com/LukeMathWalker/cargo-chef?tab=readme-ov-file#benefits-of-cargo-chef package, which I think will resolve our caching problems we were trying to solve by copying crates individually, this should also make it so we have to update our crates less so the code is more maintainable.

KolbyML: I will make a PR updating our dockerfiles with what I said above in mind

KolbyML: https://github.com/ethereum/trin/pull/1619

Here is a PR which resolves your concern, once both of these PR's are merged it will work as expected

ethportal-api/src/version.rs
```diff
@@ -1,4 +1,31 @@
...
+    const_format::str_split!(env!("VERGEN_CARGO_TARGET_TRIPLE"), "-")[2];
+
+/// The architecture of the build, x86_64, aarch64, etc.
+pub const BUILD_ARCHITECTURE: &str =
+    const_format::str_split!(env!("VERGEN_CARGO_TARGET_TRIPLE"), "-")[0];
+
+// /// The version of the programming language used to build the binary.
+pub const PROGRAMMING_LANGUAGE_VERSION: &str = env!("VERGEN_RUSTC_SEMVER");
+
+pub const VERSION: &str = const_format::formatcp!(
```
ogenev, line 19: nit:

```suggestion
pub const FULL_VERSION: &str = const_format::formatcp!(
```

ogenev (approved):
LGTM 👍

PR #1619, opened 2025-01-06
refactor: update docker images to use cargo-chef

KolbyML:
### What was wrong?

@njgheorghita brought up a concern that https://github.com/ethereum/trin/pull/1615#discussion_r1904266409 with my PR to update how we get the client info, to use vergen instead of shadow-rs https://github.com/ethereum/trin/pull/1615 that our docker images would no longer get the right git commitment.

He brought up a very good point! The way our dockerfiles were setup vergen wouldn't work properly. 

### How was it fixed?

Remove the requirement to pass `GITHASH` to the docker build, by copying the whole working directory, since then the working directory has access to the .git folder, we no longer need to pass it in.

Another thing is now we don't need to update our dockerfile everytime we add a new crate, which will improve maintainability.

Because we are now copying the whole workspace, to not lose caching performance I added cargo-chef (which reth uses) which will changes more info can be found at https://github.com/LukeMathWalker/cargo-chef

I tested these changes locally with https://github.com/ethereum/trin/pull/1615 and it works
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ jobs:
       - run:
           name: Build Docker image
           no_output_timeout: 30m
-          command: docker build -f ./docker/Dockerfile -t $IMAGE_NAME:latest --build-arg GIT_HASH=$CIRCLE_SHA1 .
+          command: docker build -f ./docker/Dockerfile -t $IMAGE_NAME:latest
       - run:
           name: Check that Docker container has no defects
           command: docker run $IMAGE_NAME:latest -h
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - run:
           name: Build Docker image
           no_output_timeout: 30m
-          command: docker build -f ./docker/Dockerfile -t $IMAGE_NAME:latest --build-arg GIT_HASH=$CIRCLE_SHA1 .
+          command: docker build -f ./docker/Dockerfile -t $IMAGE_NAME:latest
```
njgheorghita, line 69: Maybe I'm missing where this happens, but are you sure we have git initialized at this point? Otherwise there's no `.git` folder for docker to pick up

PR #1620, opened 2025-01-07
fix: can't serialize transactions from nethermind when running history bridge

KolbyML:
### What was wrong?
geth includes "parityY and parityY aliased to the value of `v`" for transactions types 1,2,3

Nethermind also has "parityY and parityY aliased to the value of `v`", but only for transactions types 1 and 2

so when I try and run a history bridge using a nethermind node as the EL provider it fails when I call `eth_getBlockByNumber` as our code expects `v` to be available in transaction type 3, but it isn't included

### How was it fixed?

Since Geth and Nethermind use parityY (as they should `v` is only supported in legacy transactions), I removed `#[serde(rename(deserialize = "v"))]` for all 3 transaction types which fixes the bug. types 1,2,3 don't support v regardless they support parityY so I think we should decode for parityY


morph-dev (approved):
What does the spec says about this?

Should we maybe use [`serde(alias = "v")`](https://serde.rs/field-attrs.html#alias) ?

In general, I don't have strong opinion on whether we remove it or use an "alias".

ogenev:
Some tests are failing.

KolbyML:
> What does the spec says about this?
> 
> Should we maybe use [`serde(alias = "v")`](https://serde.rs/field-attrs.html#alias) ?
> 
> In general, I don't have strong opinion on whether we remove it or use an "alias".

https://github.com/ethereum/execution-apis/blob/10f58fbface95676780ee7328091a494e9584a6e/src/schemas/transaction.yaml#L424

The spec says it has been deprecated for 2 years now

KolbyML:
> Some tests are failing.

It seemed like the test data wasn't conforming to the spec. I tested getting the data from Geth,Besu,Nethermind,Reth and they all conformed to the spec, so I just ended up updating the testdata

KolbyML:
> What does the spec says about this?
> 
> Should we maybe use [`serde(alias = "v")`](https://serde.rs/field-attrs.html#alias) ?
> 
> In general, I don't have strong opinion on whether we remove it or use an "alias".

Since the spec says parityY is required and "v" is deprecated, I think we are file just removing it

PR #1621, opened 2025-01-07
refactor: use generics for consensus test specs testing to reduce code duplication

KolbyML:
### What was wrong?

Currently every fork we must add thousands of lines of test data from ethereum/consensus-spec-tests and copy pastes tons of tests

This bloats our repo with large files, but also creates a bunch of almost duplicate tests. 

This method also makes our test coverage limited as adding tests taking a lot of work to add and maintain so we often avoided adding additional test cases

This makes it very annoying and tedious to write tests for future forks

### How was it fixed?

Creating a `ef-testing` crate which runs tests on data mostly generically, but manually for some cases.
- we add a Makefile to download the data from `ethereum/consensus-spec-tests`
- use generics to reduce code duplication and makes it easier to test future forks

These tests will also run in the CI which is good

Lighthouse and Reth use similar strategies for dealing with EF test data

**In short**
- test coverage and maintainability is now highly improved
- adding future tests and supporting new forks is easier

# Future work

This is a big improvement, but of course we can always still improve. Anyways I want to move `ethportal-peertest` to the testing folder, and probably the test folder into it as well. What do you guys think?

I also want to move `utp-testing` into the testing folder


testing/ef-tests/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+Run Tests this will automatically download test data
+```bash
+make test
+```
+
+Clean test files
+```bash
+make clean
+```
```
njgheorghita, line 12: In general, I think this README could use a fair amount more useful information....
- outline of the crate structure
- what kinds of datatypes are being tested?
- how to add tests in future forks & make sure that they're tested
- any other info you think would be useful to someone who just looks at this crate, and doesn't want to go digging through the code to understand what's going on

KolbyML: Thank you for the feedback I will update the readme :saluting_face:

KolbyML:
@njgheorghita ready for another look, I tried to address the feedback, let me know if there is anything else you think would need improvement

testing/ef-tests/Makefile
ogenev, line 1: I'm more in favor of having a Makefile in the root dir. I would like, for example, to run a `make test` from the root to do all data downloads and all testing for me locally.  We can have different commands for more granular testing, something similar to what reth are doing [here](https://github.com/paradigmxyz/reth/blob/986c75434a92ca34fdd37d0fa1d9cf53d7e74477/Makefile#L200) and [here](https://github.com/paradigmxyz/reth/blob/986c75434a92ca34fdd37d0fa1d9cf53d7e74477/Makefile#L491).

ogenev (commented):
I think it will be better if we have a root Makefile, instead of always going inside `testing/ef-test` if we want to download the test data.

KolbyML:
@ogenev ready for another review, I believe I addressed all your concerns, let me know if I missed something or anything

morph-dev:
I didn't take a closer look, but I wanted to ask why use Makefile and downloading test files instead of using submodule (like we do for `portal-spec-tests`)? Am I missing something?

In general, I'm in favor of this PR.

KolbyML:
> I didn't take a closer look, but I wanted to ask why use Makefile and downloading test files instead of using submodule (like we do for `portal-spec-tests`)? Am I missing something?
> 
> In general, I'm in favor of this PR.

Both lighthouse and reth download the test files using makefiles. 

The test data is over 500MB's and uses git lfs, so I think using a submodule would make it less friendly to clone our project or run tests.

Doing it this way git cloning is still fast, and the annoying thing of git submodules changing branches is avoided.

Makefile
```diff
@@ -97,10 +97,50 @@ build-release-tarballs: ## Create a series of `.tar.gz` files in the BIN_DIR dir
...
+	fi
+
+.PHONY: ef-tests
+ef-tests: extract_test_data ## Runs Ethereum Foundation tests.
+	@echo "Running tests..."
+	@cargo test -p ef-tests --features ef-tests
+	@echo "Tests complete."
+
+.PHONY: test
+test: ## Runs workspace tests.
```
ogenev, line 137: What do you think about having a make command like `test-workspace` (or `test-trin`?), where we will run `cargo test --workspace -- --nocapture` and then having `make test`  to run all tests (`ef-tests` + `test-workspace`)?

KolbyML: I wouldn't want `make test` to run `ef-tests` as it downloads files

neither lighthouse or reth uses `make test` to run `ef-tests`

Lighthouse puts `ef-tests` behind `make test-full`, I will do that as then it is more intentional you want to run the expanded test suite, which would involve downloading files.

Both lighthouse and reth use `make test` for testing the workspace

ogenev: ok fair enough 👍 .

ogenev (approved):
⛵

PR #1624, opened 2025-01-07
refactor: move crates into crates folder and remove redundant prefixes

KolbyML:
### What was wrong?

- Having all the crates in / made it so we always had to update our dockerfiles/dockerignore when new crates are added
- Our root project directory was getting pretty big
- we had a lot of redundant prefixes on our paths, trin-* x,y,z

Both https://github.com/paradigmxyz/reth and https://github.com/lambdaclass/ethrex

use this general structure, I think it is very clean and makes it easier to expand our project going forward

### How was it fixed?

moving folders, renaming a few as well

### To-Do in a future PR
I am leaving
- portal bridge and trin execution for now, eventually we can put there binaries in the bin folder and make crates out of them and put them in the crates folder. I kinda like trin-execution being in its own folder though, but I think the refactoring would make sense

I want to move `utp-testing` and `ethportal-peertests` to the testing folder added in https://github.com/ethereum/trin/pull/1621

But https://github.com/ethereum/trin/pull/1621 would need to be merged first so that will probably happen in a follow up PR


Cargo.toml
```diff
@@ -1,21 +1,21 @@
...
     "bin/trin", 
-    "ethportal-api",
+    "crates/ethportal-api",
+    "crates/e2store",
+    "crates/light-client",
+    "crates/metrics",
+    "crates/rpc",
+    "crates/beacon",
+    "crates/evm",
+    "crates/history",
```
njgheorghita, line 11: *nitpick* ... maybe we could leave the prefix on the `history` / `state` / `beacon` crates? Sometimes when you have to make a change to one of these, you typically have to replicate it across all three, so it's nice when they're grouped together alphabetically

njgheorghita (commented):
I'm 👍 for the direction of this pr, but I also think this needs a :+1: from everyone before merging

Cargo.toml
KolbyML: I was thinking of putting them in a folder
so
``crates/subnetworks/history``
``crates/subnetworks/state``
``crates/subnetworks/history``

What do you think? I am open to ideas, I am fine doing the suggestion above, I think a folder would be a little cleaner though

njgheorghita: yeah, works for me!

KolbyML:
the book ci is failing here is a PR to fix it https://github.com/ethereum/trin/pull/1628

ethportal-api/src/test_utils/constants.rs
ogenev, line 1: I still see the `ethportal-api/src/test_utils` folder in the root folder on my GitHub review dashboard, could you confirm that the whole ethportal-api folder is removed from the root?

ogenev (approved):
I'm OK with these changes, but as Nick suggested, everyone from the team needs to review this.

ethportal-api/src/test_utils/constants.rs
KolbyML: image

I don't see it, maybe it is a visual bug on your dashboard?

morph-dev (approved):
LGTM

I like this change :ship:

ethportal-api/src/test_utils/constants.rs
ogenev: yeah, maybe, It looks good now.

carver (approved):
Discussed live. GTG with follow-ups to match crate names to paths, and have consistent crate names.

Issue #1551, opened 2024-10-24
Trin do not create dashboard


r4f4ss:
I am following trin book to run dashborad using docker: https://ethereum.github.io/trin/users/monitoring.html#metrics-setup-with-docker

In the last step, creation of dashboard (step 5 `cargo run -p trin -- create-dashboard`) I can not create a dashboard:

(logs with `RUST_BACKTRACE=1`)

```
      Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.43s
     Running `target/debug/trin create-dashboard`
thread 'main' panicked at ethportal-api/src/types/cli.rs:266:17:
Creating dashboard failed template is missing value for placeholder "protocol"
stack backtrace:
   0: rust_begin_unwind
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/std/src/panicking.rs:665:5
   1: core::panicking::panic_fmt
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/panicking.rs:74:14
   2: ethportal_api::types::cli::TrinConfig::new_from
             at ./ethportal-api/src/types/cli.rs:266:17
   3: ethportal_api::types::cli::TrinConfig::from_cli
             at ./ethportal-api/src/types/cli.rs:254:9
   4: trin::main::{{closure}}
             at ./src/main.rs:12:23
   5: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/eeb90cda1969383f56a2637cbd3037bdf598841c/library/core/src/future/future.rs:123:9
   6: tokio::runtime::park::CachedParkThread::block_on::{{closure}}
             at <PATH>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/park.rs:281:63
   7: tokio::runtime::coop::with_budget
             at <PATH>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/coop.rs:107:5
   8: tokio::runtime::coop::budget
             at <PATH>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/coop.rs:73:5
   9: tokio::runtime::park::CachedParkThread::block_on
             at <PATH>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.40.0/src/runtime/park.rs:281:31
[...]

PR #1625, opened 2025-01-08
fix: use Grafana provisioning for metrics

Ktl-XV:
### What was wrong?
As reported in #1551 the `create-dashboard` trin sub command stopped working.

The issue arose because Grafana and `nanotemplate` both use brackets for template placeholders. A change in the dashboard introduced new metrics which included Grafana placeholders, but `nanotemplate` was parsing them as it's own.

### How was it fixed?
Since Grafana includes a provisioning system, switched to using that and removed the `create-dashboard` trin sub command. This also removed the need for `nanotemplate` as a dependency.

Additionally, updated the dashboard's panels to display State network data (and any networks that might be added later) 

### To-Do

- [ ] [etc/grafana/dashboards/trin-bench-trin-metrics-dashboard.json.template](/ethereum/trin/blob/master/etc/grafana/dashboards/trin-bench-trin-metrics-dashboard.json.template) is sort of dangling now, and it is also [not strictly needed in this repo](https://github.com/ethereum/trin/pull/1008#discussion_r1378880317) should it be kept, moved or deleted?


KolbyML:
I am not the biggest fan of this approach because it makes the code/instructions platform specific, whereas before you could use the subcommand on Windows as well

Ktl-XV:
Thanks for the feedback.

That's right, didn't think about Windows. Since currently the monitoring instructions are Linux and Docker only, that is where I focused.

Could a solution be if/when adding Windows instructions, add a step to edit the provisioning file and point to the appropriate location?

A different option could be to include an importable dashboard where the user pastes the JSON in Grafana and the web UI guides them to select a Prometheus data source (which can still be provisioned automatically). This approach would work in any OS while also making it easier to add the Trin Dashboard to Grafana instances without having to input credentials into the terminal. This template could also be uploaded to the [Grafana Dashboards website](https://grafana.com/grafana/dashboards/) to make it easier to import.

Or we could also replace `nanotemplate` with another templating engine which can handle this use case properly and keep the subcommand. I don't like this approach much because, imho, configuring a Grafana instance should be out of scope of the main software, but I'm willing to do it if you prefer this option.

KolbyML:
> Thanks for the feedback.
> 
> That's right, didn't think about Windows. Since currently the monitoring instructions are Linux and Docker only, that is where I focused.
> 
> Could a solution be if/when adding Windows instructions, add a step to edit the provisioning file and point to the appropriate location?
> 
> A different option could be to include an importable dashboard where the user pastes the JSON in Grafana and the web UI guides them to select a Prometheus data source (which can still be provisioned automatically). This approach would work in any OS while also making it easier to add the Trin Dashboard to Grafana instances without having to input credentials into the terminal. This template could also be uploaded to the [Grafana Dashboards website](https://grafana.com/grafana/dashboards/) to make it easier to import.
> 
> Or we could also replace `nanotemplate` with another templating engine which can handle this use case properly and keep the subcommand. I don't like this approach much because, imho, configuring a Grafana instance should be out of scope of the main software, but I'm willing to do it if you prefer this option.

Assuming the subcommand worked (it used to), didn't it work on Windows?

KolbyML:
If you want a think I good solution would be doing what I did for Trin Execution https://github.com/ethereum/trin/tree/master/trin-execution/metrics

You can rename the folder from etc to something like metrics or dashboards to be more readable.

This way anybody could run metrics easier then our current system for trin, just `docker compose up` and they would be live

Ktl-XV:
The subcommand should have worked on Windows before it broke. We can fix the subcommand by improving how we handle the templates.

But I believe setting up a Grafana Dashboard + Prometheus Datasource should be out of scope for the trin binary (most programs I'm familiar with don't have such a command). That is why this PR uses provisioning instead of fixing the subcommand, and propose adding special instructions for Windows users in the book, if/when a "Monitoring in Windows" guide is added.

That being said, if you think the subcommand should stay, I can fix it.

Ktl-XV:
I agree, using `docker compose up` is way simpler.
I'll update the PR.
Thanks

book/src/users/monitoring.md
```diff
@@ -39,9 +39,29 @@ htop
...
+```
+- The addresses must be bound to 0.0.0.0, because 127.0.0.1 only allows internal requests to
+  complete, and requests from docker instances are considered external.
+- The `--enable-metrics-with-url` parameter is the address that Trin exports metrics to, and should be equal to the port to which your Prometheus server is targeting at the bottom of `prometheus/prometheus.yml`
+- The `--web-transport http` will allow Grafana to request routing table information from Trin via JSON-RPC over HTTP
+
+4. Navigate to http://localhost:3000/d/trin-metrics/trin-metrics. Use `admin`/`admin` to login.
+
 
 ## Metrics setup no docker
```
KolbyML, line 66: If you want you can also delete the `## Metrics setup no docker`, I don't think we should keep it as
- it is more complex to maintain especially as people are unlikely to use it so nobody will know how it works so they can fix it
- the docker solution is just better

No worries if you don't want to include that in this PR, we can always remove it on a different date

Ktl-XV: I'll go ahead and remove it, that way the structure of the new `metrics` dir can be the same as the one in `trin-execution`

Ktl-XV:
Fixed the commit messages.

Thanks @KolbyML for fixing CI

KolbyML:
once the PR is rebased I will give it a review

metrics/prometheus/prometheus.yml
```diff
@@ -12,14 +12,15 @@ alerting:
...
   scrape_timeout: 10s
   metrics_path: /metrics
   scheme: http
   follow_redirects: true
   enable_http2: true
+  fallback_scrape_protocol: PrometheusText0.0.4
   static_configs:
   - targets:
-    - host.docker.internal:9100
+    - localhost:9100
```
KolbyML, line 26:
```suggestion
    - localhost:9100
    - host.docker.internal:9100
```

please add both so the dashboard works on both linux and macos

KolbyML (approved):
looks good I just have one comment

KolbyML:
could you fix the https://github.com/ethereum/trin/actions/runs/12709833285/job/35429737140?pr=1625 ci as well

KolbyML:
Thank you for the PR :saluting_face:

PR #1634, opened 2025-01-14
refactor: move crate `portal-bridge` to `bin/portal-bridge`

KolbyML:
### What was wrong?
binary crates should be in the bin folder. I said I was going to make this PR here  #1624 and #1621

### How was it fixed?
moving the crate

in the future we can crate out some parts of portal-bridge as well, but for now this move is good enough


njgheorghita (approved):
:shipit:

PR #1636, opened 2025-01-14
fix: improve light client testing coverage

njgheorghita:
### What was wrong?
There was a problem with our hive beacon test fixtures, but those have been updated in `portal-spec-tests`. It wasn't a problem with our decoding / encoding logic, just the actual fixture data. But these are some improvements to our testing that at the very least increase our coverage (to both capella & deneb types), and remove some redundant test-fixtures (the test fixtures that were inside `/test-assets/portalnet/content/beacon/*` are available in `portal-spec-tests`)

### How was it fixed?
- removed redundant test fixtures
- added tests for both capella and deneb types
- updated `portal-spec-tests` submodule

### To-Do


KolbyML (approved):
:shipit: looks good and much cleaner

PR #1637, opened 2025-01-15
refactor: perform local db content lookup at OverlayService level

### LINKED ISSUE #1329: portalnet/overlay: `FindContent` query doesn't check for the content in database before making network requests
The core `FindContent` query in `OvelrayService` [does not check](https://github.com/ethereum/trin/blob/8e36266c8a43d70fe81eb57d933058ea5d2f9c99/portalnet/src/overlay/service.rs#L2433) local database for the content before making network requests.

This cause an issue when we send commands to the `OverlayService`  from [external services](https://github.com/ethereum/trin/blob/8e36266c8a43d70fe81eb57d933058ea5d2f9c99/light-client/src/consensus/rpc/portal_rpc.rs#L55).

For example, when we request a content in the `light-client` module, the `OverlayService` ignores all content in the database and relies solely on network requests.

On the other side, when we do a `RecursiveFindContent` JSON-RPC call, we check for the content in the database but this is [implemented](https://github.com/ethereum/trin/blob/8e36266c8a43d70fe81eb57d933058ea5d2f9c99/trin-history/src/jsonrpc.rs#L96) only on the JSON-RPC level.

To allow external modules to communicate with `OverlayService` in a way to get the available content  via either database or network request, it seems that we have two options:

**Option 1:** Check for content availability in database before sending network request on `OverlayService` level and remove the locality check from the JSON-RPC module.

**Option 2:** Implement a new `OverlayService` [command](https://github.com/ethereum/trin/blob/8e36266c8a43d70fe81eb57d933058ea5d2f9c99/portalnet/src/overlay/command.rs#L15) that will explicitly check only for content availability in the database. External services then will have to send two commands to `OverlayService` - one to check for the content in the database and one to get the content from the network.

njgheorghita:
### What was wrong?
Fixes #1329 
This feature is needed to more efficiently perform beacon sync inside hive tests.

### How was it fixed?
- Perform local store content lookup at `OverlayService` level rather than at `JsonRpc` level.

### To-Do

### DIFF
--- a/crates/portalnet/src/overlay/service.rs
+++ b/crates/portalnet/src/overlay/service.rs
@@ -344,23 +344,10 @@ impl<
                         OverlayCommand::Request(request) => self.process_request(request),
                         OverlayCommand::Event(event) => self.process_event(event),
                         OverlayCommand::FindContentQuery { target, callback, config } => {
-                            if let Some(query_id) = self.init_find_content_query(target.clone(), Some(callback), config) {
-                                trace!(
-                                    query.id = %query_id,
-                                    content.id = %hex_encode_compact(target.content_id()),
-                                    content.key = %target,
-                                    "FindContent query initialized"
-                                );
-                            }
+                            self.init_find_content_query(target.clone(), Some(callback), config);
                         }
                         OverlayCommand::FindNodeQuery { target, callback } => {
-                            if let Some(query_id) = self.init_find_nodes_query(&target, Some(callback)) {
-                                trace!(
-                                    query.id = %query_id,
-                                    node.id = %hex_encode_compact(target),
-                                    "FindNode query initialized"
-                                );
-                            }
+                            self.init_find_nodes_query(&target, Some(callback));
                         }
                         OverlayCommand::RequestEventStream(callback) => {
                             if callback.send(self.event_stream.subscribe()).is_err() {
@@ -1017,7 +1004,6 @@ impl<
             request.discv5.id = %request_id,
             "Handling FindContent message",
         );
-
         let content_key = match TContentKey::try_from_bytes(&request.content_key) {
             Ok(key) => key,
             Err(_) => {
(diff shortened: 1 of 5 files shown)


crates/portalnet/src/overlay/service.rs
```diff
@@ -2369,9 +2358,27 @@ impl<
...
+                let _ = callback.send(Ok((
+                    RawContentValue::from(content),
+                    false,
+                    Some(query_trace),
+                )));
+                return;
+            } else {
+                warn!("Local content found but no callback to return it to overlay protocol, continuing with network query & ignoring local content.");
+            }
+        }
```
KolbyML, line 2380:
```suggestion
        // Lookup content locally before querying the network.
        if let Ok(Some(content)) = self.store.read().get(&target) {
            let local_enr = self.local_enr();
            let mut query_trace = QueryTrace::new(&local_enr, target.content_id().into());
            query_trace.node_responded_with_content(&local_enr);
            query_trace.content_validated(local_enr.into());
            if let Some(callback) = callback {
                let _ = callback.send(Ok((
                    RawContentValue::from(content),
                    false,
                    Some(query_trace),
                )));
                return;
            } else {
                warn!("Local content found but no callback to return it to overlay protocol, continuing with network query & ignoring local content.");
            }
        }
```

I think we should remove the Option on the callback, the option is only used for testing, which tastes weird, also we should remove this `warn!` and check if there is a channel.

But as I said before we should just remove the option, I don't think we should have special arguments just for writing unit tests.

It doesn't make sense to do a storage lookup, if we aren't going to use it.

crates/portalnet/src/overlay/service.rs
```diff
@@ -2369,9 +2358,27 @@ impl<
         target: TContentKey,
         callback: Option<oneshot::Sender<RecursiveFindContentResult>>,
```
KolbyML, line 2359:
```suggestion
        callback: oneshot::Sender<RecursiveFindContentResult>,
```

I don't think this should be an option

KolbyML (approved):
:shipit: this PR looks very good

Issue #1571, opened 2024-10-30
Add an rpc test confirming that we support batched requests


carver:
For the automated test, here is a batch example in alloy here:
https://crates.io/crates/alloy-rpc-client

It would be nice to run a manual gut-check test that the same works in web3.py, a la:
https://web3py.readthedocs.io/en/stable/web3.main.html#batch-requests

oac1771:
Hello! Can I be assigned to this issue?

ogenev:
> Hello! Can I be assigned to this issue?

No need to be assigned, open a PR when you are ready 😸.

oac1771:
Sweet, I will work on it now. I do have a follow up question though. 

How exactly to you run the tests found in the root [test](https://github.com/ethereum/trin/tree/master/tests) directory? I looked through the trin book, but didnt see anything and I also ran the `make test` commands that I found in the [circle ci](https://github.com/ethereum/trin/blob/master/.circleci/config.yml#L225)  config file which doesnt seem to run those tests. Thanks!

KolbyML:
> Sweet, I will work on it now. I do have a follow up question though.
> 
> How exactly to you run the tests found in the root [test](https://github.com/ethereum/trin/tree/master/tests) directory? I looked through the trin book, but didnt see anything and I also ran the `make test` commands that I found in the [circle ci](https://github.com/ethereum/trin/blob/master/.circleci/config.yml#L225) config file which doesnt seem to run those tests. Thanks!

That should be fix in this PR https://github.com/ethereum/trin/pull/1632

The `tests` folder must be winthin a crate, currently it isn't.

For the reasoning why #1632 fixes it is here https://github.com/ethereum/trin/pull/1633#issuecomment-2588329265

It should be merged tomorrow, so you can begin your work then

oac1771:
Sweet thanks, ill keep an eye out

oac1771:
Hello, I have opened a [PR](https://github.com/ethereum/trin/pull/1638) for this

ogenev:
Resolved in #1638.

PR #1638, opened 2025-01-16
test: add a rpc test confirming that we support batched requests

oac1771:
### What was wrong?
Adds test case described in #1571 which adds a test case ensuring batch calls are supported.

### How was it fixed?
used allow http rpcclient to batch call trin node

### To-Do


testing/ethportal-peertest/tests/rpc_server.rs
```diff
@@ -62,6 +66,50 @@ async fn setup_web3_server() -> (RpcServerHandle, RootProvider<PubSubFrontend>,
...
+    let client_version_fut = batch
+        .add_call::<(), serde_json::Value>("web3_clientVersion", &())
+        .unwrap();
+    let node_info_fut = batch
+        .add_call::<(), serde_json::Value>("discv5_nodeInfo", &())
+        .unwrap();
+
+    batch.send().await.unwrap();
+    client_version_fut.await.unwrap();
+    node_info_fut.await.unwrap();
```
KolbyML, line 108: what is fut? I think it is fine if we just write `future` there, makes it more readable for people

oac1771: Sweet, just made the change

KolbyML (approved):
:shipit: PR looks good. Thank you for your contribution :partying_face:

PR #1640, opened 2025-01-17
refactor: add trin prefix to most crates

KolbyML:
### What was wrong?

in the meeting on monday we agreed to add trin- as a prefix to most of the crates

### How was it fixed?

adding the prefix

I wasn't sure if I was supposed to add the prefix to `ethportal-api`. If I missed something please let me know


KolbyML:
closing as even though I was told to make this PR in our meeting, there doesn't appear to be actual support for this change, I agree with not merging this so I will close the PR

PR #1643, opened 2025-01-17
fix: cross build ci required for releases + remove const_format

KolbyML:
### What was wrong?

There was a bug in https://github.com/ethereum/trin/pull/1615 which caused https://github.com/ethereum/trin/actions/runs/12834319209/job/35791388762 release ci to fail, on trying to build.

### How was it fixed?

I removed const_format cause it wasn't needed and also fixed some indexing

I tested on my forked github repo and it works

release: https://github.com/KolbyML/trin/releases
release branch: https://github.com/KolbyML/trin/tree/fix-cross-build
ci working: https://github.com/KolbyML/trin/actions/runs/12835484962/job/35796339052

image

I tested the binary built locally

image

and the short commit matches

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1440,27 +1440,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c2459377285ad874054d797f3ccebf984978aa39129f6eafde5cdc8315b612f8"
 
 [[package]]
-name = "const_format"
-version = "0.2.34"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "126f97965c8ad46d6d9163268ff28432e8f6a1196a55578867832e3049df63dd"
-dependencies = [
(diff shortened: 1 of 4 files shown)


njgheorghita (approved):
Nice, the only thing I can think of is to make the environment variable names their own variables, so we can just use the variable across files and not worry about possible spelling errors breaking something. Though, it seems like these are the only places these env vars will be used, so it might end up being too much overhead for not much utility. Anyways, leaving that call up to you

KolbyML:
> Nice, the only thing I can think of is to make the environment variable names their own variables, so we can just use the variable across files and not worry about possible spelling errors breaking something. Though, it seems like these are the only places these env vars will be used, so it might end up being too much overhead for not much utility. Anyways, leaving that call up to you

I am open to it being done in a future PR, I want to deploy the release now and CI takes 10 minutes xD

PR #1646, opened 2025-01-20
feat: add uTP duration metrics

KolbyML:
### What was wrong?
For the most part we have no idea into how uTP is performing, are things working right? Idk lets add some metrics
### How was it fixed?

add a timer attached to the semaphore. This metric won't include uTP transfers initiated over json-rpc, but I am not really concerned with the uTP transfers over json-rpc

### DIFF
--- a/crates/portalnet/src/utp/timed_semaphore.rs
+++ b/crates/portalnet/src/utp/timed_semaphore.rs
@@ -1,0 +1,17 @@
+use tokio::sync::OwnedSemaphorePermit;
+use trin_metrics::timer::DiscardOnDropHistogramTimer;
+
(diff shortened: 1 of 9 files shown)


crates/portalnet/src/utp/timed_semaphore.rs
```diff
@@ -0,0 +1,15 @@
+use tokio::sync::OwnedSemaphorePermit;
+use trin_metrics::timer::DiscardOnDropHistogramTimer;
+
+#[derive(Debug)]
+pub struct OwnedTimedSemaphorePermit {
```
ogenev, line 5: nit: Some doc strings are always nice

ogenev (approved):
LGTM 👍

Issue #1574, opened 2024-10-31
Estimate radius at <100% when under-utilizing space


carver:
## Problem

When running trin before storage hits the allocation limit, we currently run at 100% radius. If the steady state radius will land at 3% for a particular storage allocation, then 97% of the stored content we receive early on (in the period before radius starts to shrink) will end up being deleted. During that period, 97% of the network transfer, cpu for verification, and storage I/O is wasted.

### Pre-shrink the radius

We can't know the size of the network with arbitrary precision ahead of time, but we can bake in a rough guess to the binary. This rough guess can massively shrink the wasted data.

### Asymmetry in the guess

The effect of incorrectly estimating the radius is highly asymmetrical:
- don't preshrink the radius enough, and ... we're still much better off than before
- preshrink the radius too much, and you never store as much as you are willing to store (without us adding a more complicated feature, to bump up the radius)

So we probably want to take our radius estimate and double it. We still get a plenty large benefit, with only a small downside when our baked-in numbers get stale over time (assuming it's quite rare for a network total storage to shrink).

### Real example

Our current fleet of history nodes at steady state show a ~2% radius with 35 GB allocated each, which implies a 1.75 TB total network size.

If you launch a fresh trin instance for history with 17.5 GB of space, we can estimate that the radius will end up at 1%. With a 2x buffer, we could preshrink the radius to 2%. That means that we waste ~1% of total resources compared to the status quo of ~99%. (During the period before the radius shrinks below 100%) 99x gains are good. The less storage you allocate, the more there is to gain with this change. Those are the folks who probably care the most about performance.

History approaches the correct radius relatively quickly in practice, but state takes a while, and is more resource-intensive. So state is probably where we will feel the benefit of this the most.

pipermerriam:
This is another one that will probably cause hive test failures.  A way around this would be to have this value be able to be configured at the CLI so that in hive tests it can effectively be disabled.  Seems potentially as simple as something like `--history.total_network_data=600_000`.  This would allow us to take the storage limit they provided (or the default) and extrapolate the appropriate starting radius.  For hive, we might also want to introduce something like `--history.max_allowed_radius=auto` as the default, which means it will extrapolate the value from the total network size.  In hive, we can then do `--history.max_allowed_radius=100` to have it do the *default* thing it does now and start the radius at 100%.

[...]

PR #1647, opened 2025-01-22
feat: add max-radius param with 5% default

KolbyML:
### What was wrong?

For a larger description of the problem can be viewed here https://github.com/ethereum/trin/issues/1574

The TL:DR is currently we start at 100 percent radius when the node will end up at 1-3%. This means we download and purge content our node won't end up storing which wastes a lot of CPU cycles, but also takes our uTP transfer limit which could be used for content within our final 1-3% final radius.

This is a big issue as on the state network, currently we are trying to bridge latest state, but our nodes uTP limit is always full so often `trin-execution` offer's of state diffs are declined as all State nodes are at transfer capacity.

image
^ here is a picture I took awhile ago, at the bottom you can see we successfully offered to 0 peers and the content was declined 8 times. This is due to the State nodes being at uTP transfer capacity (or in other words hitting our rate limit)

Recently this issue has gotten worse, but even if it didn't this PR is required if we want to bridge latest state. Bridging state content is very expensive, because there is just so much state to gossip. If we want Portal and the State network to be successful we need a 10x performance improvement at least in Trin as currently Trin can only handle 5 Mbps up and down which is way to slow, we won't be able to take a load on the network with this.

So this PR is the first step in a road of changes which will hopefully greatly increase the performance in Trin. I don't think we have a choice we need to be performent if we want users.

### How was it fixed?

Add a cli called `max-radius`, by default the max radius is 5 percent, assuming we have a network of over 20 nodes we should have max coverage. Since the nodes radius's should end up below 3% especially as the network storage requirements will keep growing over time. So I don't think we should have a default bigger then 5%

For testing on hive we will use `--max-radius 100`

I think this solution is the best one for the foreseeable future, we could do a more complex scheme but that wouldn't be worth it right now, instead we should focus on finding other bottlenecks in Trin as this PR is good enough and gets us most of want we want for the foreseeable future. A more complex scheme would give diminishing returns, so I wouldn't expect us to look into it until Trin is very optimized so it is a very low priority to further optimize this with a complex solution.

### DIFF
--- a/bin/trin/src/lib.rs
+++ b/bin/trin/src/lib.rs
@@ -7,7 +7,9 @@ use std::sync::Arc;
 
 use cli::TrinConfig;
 use ethportal_api::{
-    types::network::Subnetwork, utils::bytes::hex_encode, version::get_trin_version,
+    types::{distance::Distance, network::Subnetwork},
+    utils::bytes::hex_encode,
+    version::get_trin_version,
 };
 use portalnet::{
     discovery::{Discovery, Discv5UdpSocket},
@@ -103,7 +105,7 @@ pub async fn run_trin(
                 &discovery,
                 utp_socket.clone(),
                 portalnet_config.clone(),
-                storage_config_factory.create(&Subnetwork::State)?,
+                storage_config_factory.create(&Subnetwork::State, trin_config.max_radius)?,
                 header_oracle.clone(),
             )
             .await?
@@ -123,7 +125,7 @@ pub async fn run_trin(
             &discovery,
             utp_socket.clone(),
             portalnet_config.clone(),
-            storage_config_factory.create(&Subnetwork::Beacon)?,
+            storage_config_factory.create(&Subnetwork::Beacon, Distance::MAX)?,
             header_oracle.clone(),
         )
         .await?
@@ -146,7 +148,7 @@ pub async fn run_trin(
             &discovery,
             utp_socket.clone(),
             portalnet_config.clone(),
-            storage_config_factory.create(&Subnetwork::History)?,
+            storage_config_factory.create(&Subnetwork::History, trin_config.max_radius)?,
             header_oracle.clone(),
         )
         .await?
--- a/crates/storage/src/versioned/id_indexed_v1/store.rs
+++ b/crates/storage/src/versioned/id_indexed_v1/store.rs
@@ -84,12 +84,12 @@ impl<TContentKey: OverlayContentKey> VersionedContentStore for IdIndexedV1Store<
         let pruning_strategy = PruningStrategy::new(config.clone());
 
         let mut store = Self {
-            config,
-            radius: Distance::MAX,
+            radius: config.max_radius,
             pruning_strategy,
             usage_stats: UsageStats::default(),
             metrics: StorageMetricsReporter::new(subnetwork),
             _phantom_content_key: PhantomData,
+            config,
         };
         store.init()?;
         Ok(store)
@@ -133,11 +133,12 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
         } else {
             debug!(
                 Db = %self.config.content_type,
-                "Used capacity ({}) is below target capacity ({}) -> Using MAX radius",
+                "Used capacity ({}) is below target capacity ({}) -> Using MAX radius ({})",
                 self.usage_stats.total_entry_size_bytes,
-                self.pruning_strategy.target_capacity_bytes()
+                self.pruning_strategy.target_capacity_bytes(),
+                self.config.max_radius,
             );
-            self.radius = Distance::MAX;
+            self.radius = self.config.max_radius;
             self.metrics.report_radius(self.radius);
         }
 
@@ -420,7 +421,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
 
     /// Sets `self.radius` to the distance to the farthest stored content.
     ///
-    /// If no content is found, it sets radius to `Distance::MAX`.
+    /// If no content is found, it sets radius to `config.max_radius`.
     fn set_radius_to_farthest(&mut self) -> Result<(), ContentStoreError> {
         match self.lookup_farthest()? {
             None => {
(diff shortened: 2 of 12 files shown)


bin/trin/src/lib.rs
```diff
@@ -146,7 +148,7 @@ pub async fn run_trin(
             &discovery,
             utp_socket.clone(),
             portalnet_config.clone(),
-            storage_config_factory.create(&Subnetwork::History)?,
+            storage_config_factory.create(&Subnetwork::History, trin_config.max_radius)?,
```
njgheorghita, line 151: Is there any reason to leave this as `Distance::MAX`? Since it seems like the bottlenecking volume is only on the state network iiuc.

KolbyML: The bottlenecking is only more extreme on State. This issue occurs on History too, it is just the radius's are settled now, if we are targeting low hardware requirements I think this setting should be used on the history network as well.

KolbyML: Also since we run History and State at the same time, they affect each other in terms of performance, so we should do both as bad performance on the History network would degrade performance of the State network.

crates/storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -432,7 +433,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
                     self.radius = Distance::ZERO;
                 } else {
                     error!(Db = %self.config.content_type, "Farthest not found!");
-                    self.radius = Distance::MAX;
+                    self.radius = self.config.max_radius;
                 }
             }
             Some(farthest) => {
```
morph-dev, line 439: Should we `self.radius` to not be higher than `max_radius` in this case as well.

morph-dev (commented):
In general, I'm in favor of this direction.

Few concerns:

1. we currently have only ~25 state nodes. Making their radius to 5% would leave us with gaps.
    - we can/should get more state nodes
    - once live, big storage nodes should make this less of an issue (it would still be nice to have entire domain space covered without them)
2. Once this is merged, with the code as is, nothing would change regarding already running nodes, which is what we want to accomplish in the first place. If you look at `IdIndexedV1Store::init`, I will fall under [this condition](https://github.com/ethereum/trin/blob/83ceaed350baf165143043130a75fddea4362585/crates/storage/src/versioned/id_indexed_v1/store.rs#L115-L125), which means that it will initialize radius to distance of the farthest content in the db. See my comment as well.
3. Even when 2 is fixed, we have a bit of not properly tested scenario where we prune a lot of data and trin startup time. This might not be a problem at all (it's one time thing), but I wanted to highlight it.

KolbyML:
> In general, I'm in favor of this direction.
> 
> Few concerns:
> 
> 1. we currently have only ~25 state nodes. Making their radius to 5% would leave us with gaps.
>    
>    * we can/should get more state nodes
>    * once live, big storage nodes should make this less of an issue (it would still be nice to have entire domain space covered without them)

All 25 state nodes are expected to settle well below 1% radius, we can add more state nodes there wouldn't be any push back against this.

crates/storage/src/versioned/id_indexed_v1/store.rs
KolbyML: Sure I will do min()

KolbyML: Ok I added the min so it will not return higher then the `max_radius` now

morph-dev (approved):
> All 25 state nodes are expected to settle well below 1% radius, we can add more state nodes there wouldn't be any push back against this.

When do you think we should add more nodes?

---

In general, I'm ok with this being merged. But would like to see 1-2 more approvals from others.

I'm still concerned that we might have holes in the domain space. But luckily we have big_storage nodes and there are other clients as well, so none of the content should be lost.

Once this is merged, and before it's deployed, we have to make sure that ansible configuration for big_storage nodes is updated with this flag.

KolbyML:
> > All 25 state nodes are expected to settle well below 1% radius, we can add more state nodes there wouldn't be any push back against this.
> 
> When do you think we should add more nodes?

We can double our fleet now if you want, I am not necessarily in a rush, but it couldn't hurt

crates/utils/src/cli.rs
```diff
@@ -72,3 +73,30 @@ pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>
...
+        }
+    };
+
+    if max_radius_percentage > U256::from(100) {
+        return Err(format!(
+            "Invalid max radius percentage, expected 0 to 100, received: {max_radius_percentage}"
+        ));
+    }
+
+    // If the max radius is 100%, return the maximum distance as without it we will be 35 less than
```
njgheorghita, line 93: Not sure I understand this comment?

crates/utils/src/cli.rs
```diff
@@ -72,3 +73,30 @@ pub fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>
 
     Ok(Arc::new(subnetworks))
 }
+
+pub fn max_radius_parser(max_radius_str: &str) -> Result<Distance, String> {
+    let max_radius_percentage = match max_radius_str.parse::<U256>() {
```
njgheorghita, line 78:
```rust
let Ok(max_radius_percentage) = max_radius_str.parse::<U256> else {
    return Err(...)
}
```

njgheorghita (approved):
Looks good! Just a few nitpicks. And I'll echo what was already said, but feels pretty important
> Once this is merged, and before it's deployed, we have to make sure that ansible configuration for big_storage nodes is updated with this flag.

crates/utils/src/cli.rs
KolbyML: I want to print the error from parse

crates/utils/src/cli.rs
KolbyML: I tried to clarify it a bit

KolbyML:
https://github.com/ethereum/cluster/pull/1144 the ansible script is updated

KolbyML:
I am going to merge this PR on monday after the cluster PR is merged

PR #1648, opened 2025-01-25
fix: docker portal-bridge can't find genesis.json

KolbyML:
### What was wrong?
Milos noticed that the 1M state bridges weren't running after the refactor
image

### How was it fixed?

Stop reading the genesis file and use `include_str!` I am using this upstream for trin-execution and it is also common practice in reth for the genesis file.

Just to point out we can't use a const I believe, `include_str` errored saying we must use a string literal


morph-dev (approved):
:ship:  (assuming all tests are passing)

PR #1649, opened 2025-01-25
fix: set gas_price when handling eth_call

morph-dev:
### What was wrong?

We can't execute transaction post EIP-1559 without setting gas price. And gas price has to be at least base fee.

### How was it fixed?

If unset, set the gas price to base fee.

Also added `debug` logging for easier debugging eth_call requests.


KolbyML (approved):
:shipit: looks good

PR #1652, opened 2025-01-28
feat: update utp-rs dependency

morph-dev:
### What was wrong?

It's possible for wrong Enr to be passed to utp library if it was changed recently.
See #1596 for more details.

### How was it fixed?

The utp-rs library is updated (see https://github.com/ethereum/utp/pull/136) so only peer id (NodeId) is required when receiving the utp packet, while entire peer (Enr) is passed when calling `accept_with_cid`.

The biggest change is in `crates/portalnet/src/discovery.rs`, where we now call `find_enr` only if Enr is unknown when we have to send them packet (which should rarely happen), instead of every time we receive packet.

Worth highlighting that due to the nature of #1596, we can't easily verify that this actually fixes it until it's deployed to the servers. I did some manual testing and I'm confident that this should fix it (but I plan to verify once this is merged and deployed).

### To-Do
- [x] Update dependency to utp repo once https://github.com/ethereum/utp/pull/136 is merged

### DIFF
--- a/crates/portalnet/src/discovery.rs
+++ b/crates/portalnet/src/discovery.rs
@@ -383,7 +383,7 @@ impl UtpPeer {
 impl std::fmt::Debug for UtpPeer {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let peer_client_type = self.client().unwrap_or_else(|| "Unknown".to_string());
-        f.debug_struct("EnrPeer")
+        f.debug_struct("UtpPeer")
             .field("enr", &self.0)
             .field("Peer Client Type", &peer_client_type)
             .finish()
--- a/crates/portalnet/src/discovery.rs
+++ b/crates/portalnet/src/discovery.rs
@@ -1,8 +1,7 @@
 use std::{
-    fmt,
-    hash::{Hash, Hasher},
-    io,
+    fmt, io,
     net::{Ipv4Addr, SocketAddr},
+    ops::Deref,
     str::FromStr,
     sync::Arc,
     time::Duration,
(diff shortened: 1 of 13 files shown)


crates/portalnet/src/discovery.rs
```diff
@@ -357,59 +358,13 @@ impl Discv5UdpSocket {
...
-    }
 }
 
 /// A wrapper around `Enr` that implements `ConnectionPeer`.
 #[derive(Clone)]
-pub struct UtpEnr(pub Enr);
+pub struct EnrPeer(pub Enr);
 
-impl UtpEnr {
+impl EnrPeer {
```
KolbyML, line 367: Is this struct only going to be used with uTP or will it be used elsewhere, as we should be specific if the usecase is specific, we UtpEnr is going to start being used more broadly I think it is fine to rename it. EnrPeer is just extremely broad, so if it is still exclusive for utp maybe keeping the old name makes sense as it is more to the point.

crates/portalnet/src/discovery.rs
```diff
@@ -421,49 +376,51 @@ impl UtpEnr {
...
+        let peer = peer.clone();
         let discv5 = Arc::clone(&self.discv5);
-        let target = target.0.clone();
+        let enr_cache = Arc::clone(&self.enr_cache);
+        let header_oracle = Arc::clone(&self.header_oracle);
         let data = buf.to_vec();
         tokio::spawn(async move {
-            match discv5.send_talk_req(target, Subnetwork::Utp, data).await {
+            let enr = match peer.peer() {
+                Some(enr) => enr.0.clone(),
```
KolbyML, line 414: we are already cloning the Enr above, shouldn't we be able to avoid cloning it twice?

KolbyML (commented):
Looks pretty good I am only concerned with cloning the enr twice in a hot path, and the specifics of renaming the UtpEnr struct, as EnrPeer sounds pretty broad and ambiguous if its only use is uTP

KolbyML (approved):
:shipit: looks good

crates/portalnet/src/discovery.rs
KolbyML: I think `UtpPeer` looks good

carver (approved):
LGTM

PR #1656, opened 2025-01-31
refactor: rename `era2` to `e2ss`

KolbyML:
### What was wrong?
So I talked with @arnetheduck and we will be renaming `era2` to `e2ss` for 2 reasons
- so `era` will be reserved for Ethereum chain history whether that is EL/CL/Both
- era2 store didn't follow the era format, era refers to a period of 8192 blocks or slots, era2 was a 1 block snapshot of the state.
### How was it fixed?

renaming era2 to e2ss

Paul will be updating  era2.ethportal.net to e2ss.ethportal.net and updating the file type to e2ss, I won't merge this PR until that is done


crates/e2store/README.md
```diff
@@ -15,20 +15,20 @@ era is a format for storing beacon chain data more information can be found here
 
 era1 is a format for storing all of Ethereum's pre merge blocks. It contains block headers, block bodies, and receipts for pre-merge block history which ranges block 0-15537394
 
-## What is era2?
+## What is e2ss?
```
njgheorghita, line 18: Like here we use `e2ss`...

crates/e2store/README.md
```diff
@@ -15,20 +15,20 @@ era is a format for storing beacon chain data more information can be found here
...
 
-era2 is a format made to store full flat state snapshots, one of our first uses of this will be using to bootstrap Portal State Network bridges. Unlike `.era`/`.era1` era2 files will only store 1 block's worth of state per file. The reason for this choice is a snapshot of the state is quite large.
+e2ss is a format made to store full flat state snapshots, one of our first uses of this will be using to bootstrap Portal State Network bridges. Unlike `.era`/`.era1` e2ss files will only store 1 block's worth of state per file. The reason for this choice is a snapshot of the state is quite large.
 
 TODO: Add chart of snapshot size at every million block interval.
 
-### Era2 analysis tool
+### E2ss analysis tool
```
njgheorghita, line 24: And here we use `E2ss`... Though, I get it's the first word which is usually capitalized... but, I'd still rather see `e2ss` or `E2SS` here

njgheorghita (approved):
*nitpick* I'm not a huge fan of `E2ss`... I'd prefer to user either `e2ss` or `E2SS` depending on the context / usecase

morph-dev (approved):
:ship: 

Not blocking, but just wondering. Should we maybe add in the README file what `e2ss` stands for?

Issue #1578, opened 2024-10-31
Improvements to *NetworkApiClient ergonomics


carver:
Inspired by a bug I introduced in glados, and thoughts about how to make it less likely in the future:
https://github.com/ethereum/glados/pull/335#pullrequestreview-2408660979

The current usage pattern goes something like:
```rs
let key = StateContentKey::AccountTrieNode(...);
let response = match StateNetworkApiClient::get_content(transport, key) {
  Ok(response) => response,
  Err(e) => return handle_err(e),
};
let decoded = match StateContentValue::decode(key, response.content) {
  Ok(decoded) => decoded,
  Err(e) => return handle_err(e),
};
if let StateContentValue::TrieNode(trie_node) = decoded {
  // do_something(trie_node)
}
```

An ideal API probably looks something closer to this (at least as an option):
```rs
let key = StateContentKey::AccountTrieNode(...);
let trie_node = match StateNetworkApiClient::get_decoded_content(transport, key) {
  Ok(trie_node) => trie_node, // only a single type can be returned by any key type, so we can skip the enum
  Err(e) => return handle_err(e),
};
// do_something(trie_node)
```

I don't know the rust magic to make that happen ^ but it should be straightforward to do a middle-ground solution, like:

```rs
let key = StateContentKey::AccountTrieNode(...);
let trie_node = match StateNetworkApiClient::get_decoded_content(transport, key) {
  Ok(StateContentValue::TrieNode(trie_node)) => trie_node, // only a single type can be returned by any key type, so we can skip the enum
  Ok(wrong_type) => panic!("Should never get a different value when looking up a trie node. Got: {wrong_type}"),
  Err(e) => return handle_err(e),
};
// do_something(trie_node)
```

I do see that there could be reasons to keep both `get_content` and `get_decoded_content` around in the API. Maybe this is pretty low priority, since it amounts to having to make one less function call. (unless we can get the magic solution working)
---

At the very least, we can add some docs to `EncodedTrieNode` that say: if you're trying to decode this by hand after making an RPC request, you're probably doing it wrong, and should be using `StateContentValue::decode()` instead

morph-dev:
Somewhat related to this, we have logic in Census where we do something like this (also for other api calls, like ping, recursive_find_nodes):
```rust
match self.subnetwork {
    Subnetwork::History => {
        HistoryNetworkApiClient::find_nodes(&self.client, enr.clone(), distances)
    }
    Subnetwork::State => {
[...]

PR #1658, opened 2025-02-01
feat: Add get_decoded_content to NetworkClientApi

7suyash7:
Fix: #1578 

Add `get_decoded_content` to NetworkClientApi to simplify content retrieval.

Instead of having to manually handle content fetching and decoding in multiple steps like this:
```rust
let key = StateContentKey::AccountTrieNode(...);
let response = match StateNetworkApiClient::get_content(transport, key) {
    Ok(response) => response,
    Err(e) => return handle_err(e),
};
let decoded = match StateContentValue::decode(key, response.content) {
    Ok(decoded) => decoded,
    Err(e) => return handle_err(e),
};
if let StateContentValue::TrieNode(trie_node) = decoded {
    // do_something(trie_node)
}
```

You can now just do:
```rust
let key = StateContentKey::AccountTrieNode(...);
let decoded = match StateNetworkApiClient::get_decoded_content(transport, key) {
    Ok(value) => value,
    Err(e) => return handle_err(e),
};
// decoded value is ready to use
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7088,7 +7088,7 @@ checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 [[package]]
 name = "utp-rs"
 version = "0.1.0-alpha.8"
-source = "git+https://github.com/ethereum/utp?tag=v0.1.0-alpha.15#ef62dd992cf67662319ee69d6fe8caa2af15aeae"
+source = "git+https://github.com/ethereum/utp?tag=v0.1.0-alpha.16#a4da8630ba0f56eb04a4ead6c374a850724445ac"
 dependencies = [
  "async-trait",
  "delay_map 0.3.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,7 +95,7 @@ tree_hash_derive = "0.8.0"
 uds_windows = "1.0.1"
 ureq = { version = "2.5.0", features = ["json"] }
 url = "2.3.1"
-utp-rs = { tag = "v0.1.0-alpha.15", git = "https://github.com/ethereum/utp" }
+utp-rs = { git = "https://github.com/ethereum/utp", tag = "v0.1.0-alpha.16" }
 
 # Trin workspace crates
 e2store = { path = "crates/e2store" }
--- a/crates/portalnet/src/discovery.rs
+++ b/crates/portalnet/src/discovery.rs
@@ -1,8 +1,7 @@
 use std::{
-    fmt,
-    hash::{Hash, Hasher},
-    io,
+    fmt, io,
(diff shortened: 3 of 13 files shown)


morph-dev:
There are few issues with this PR, but there are also good stuff.

1. We probably want the same functionality for all subnetworks (not just state). So whatever we do, we will have to do for other types as well. Therefore, newly introduced  types should probably be moved to shared location and should be more general
2.  `StateContentValue` shouldn't be (de)serialized as key-value object. Instead, it should be pure hex string. As such, it can't be deserialized directly, because the same hex string can be the valid representation of multiple content types. For this reason, we have somewhat complicated logic for encoding/decoding and serialization/deserialization where `content_key` is needed for correct decoding/deserialization.
3. The `ContentKeyType` is something that I believe we can reuse. I don't think we need both `try_decode_primary` and `try_decode_proof` as they are always used one after another (except in tests).

With comments  above, I propose the following changes:
1. Move `ContentKeyType` into `crates/ethportal-api/src/types/content_key/overlay.rs` (or maybe separate file) and make it more generic, something like:
    ```rust
    pub trait ContentKey {
        type K: OverlayContentKey;
    
        fn decode(&self, buf: &[u8]) -> Result<Self::K, ContentKeyError>;
    }
    ```
    - This can be enriched to include more logic that is shared between types
2. Remove `get_decoded_content` from `StateNetworkApi` as `StateContentValue` can't be (de)serializable
3. Introduce new trait that will have function similar to `get_decoded_content`, but be generic. This would remove restrictions on returning object being (de)serializable
    - This trait should be implemented by all `<Subnetwork>NetworkApi` structures

---

I think this work should be split into separate PRs for easier review and discussion. For example, I would start with just step 1 from above. I understand that it can be difficult to work on step 1 without knowing the exact usecase from step 2. Usually my workflow would be one of:
- Do step 1 as well as possible, clearly explaining future use for it. And if needed, adapt/fix if when working on step 2
- Work on both steps at the same time, but commit changes in separate commits and send two PRs at the same time (one PR would contain only first commit, and other would contain both)

7suyash7:
Hi, @morph-dev, thanks for the review!

1. I'll remove the Serialize/Deserialize from the StateContentValue and make `try_decoded_primary` + `try_decode_proof` into a single `decode()` method. Also, move the new trait to a shared file (like content_key/overlay.rs) or maybe a new file, so it's not state-specific. This will be PR-1
2. Remove the `get_decoded_content` method from the JSON-RPC and keep it as a local trait that would implement single-step decoding so the network-level API remains raw bytes only. This would be PR-2
3. Then if we want to do the same approach for History or Beacon, I can follow up with additional PRs reusing the new trait design.

If this looks good, I'll close this PR and work on it separately.

PR #1659, opened 2025-02-01
feat: implement Debug for PingError so messages display

KolbyML:
### What was wrong?

The default derived Debug implementation printed the message as bytes

### How was it fixed?

Implement Debug manually to decode the Ping Message, there is no point in decoding it unless we print it so I think this strikes a pretty good balance.

### DIFF
--- a/crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
+++ b/crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
@@ -1,12 +1,14 @@
+use std::fmt::Debug;
+
 use anyhow::anyhow;
 use ssz::Encode;
 use ssz_derive::{Decode, Encode};
 use ssz_types::{typenum::U300, VariableList};
 
-use crate::types::portal_wire::CustomPayload;
+use crate::{types::portal_wire::CustomPayload, utils::bytes::hex_encode};
 
 /// Used to respond to pings which the node can't handle
-#[derive(PartialEq, Debug, Clone, Encode, Decode)]
+#[derive(Eq, PartialEq, Clone, Encode, Decode)]
 pub struct PingError {
     pub error_code: u16,
     pub message: VariableList<u8, U300>,
@@ -27,6 +29,18 @@ impl PingError {
                 anyhow!("PingError can only handle messages up to 300 bytes, received {err:?}")
             })?,
         })
+    }
+}
+
+impl Debug for PingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let message = String::from_utf8(self.message.to_vec())
+            .unwrap_or_else(|_| format!("Invalid utf8 string: {}", hex_encode(&*self.message)));
+
+        f.debug_struct("PingError")
+            .field("error_code", &self.error_code)
+            .field("message", &message)
+            .finish()
     }
 }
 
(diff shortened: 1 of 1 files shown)


crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
```diff
@@ -30,6 +32,20 @@ impl PingError {
     }
 }
 
+impl Debug for PingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let string = String::from_utf8(self.message.to_vec())
+            .map_err(|_| {
+                ssz::DecodeError::BytesInvalid(format!("Invalid utf8 string: {:?}", self.message))
+            })
+            .map_err(|_| std::fmt::Error)?;
```
morph-dev, line 41: Wouldn't this `map_err` completely invalidate previous one and basically always return `std::fmt::Error`?

How about we do something like this:

```rust
let message = String::from_utf8(&self.message)
    .unwrap_or_else(|_| format!("Invalid utf8 string: {}", hex_encode(&self.message)));

f.debug_struct("PingError")
    .field("error_code", &self.error_code)
    .field("message", &message)
    .finish()
```

crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
```diff
@@ -6,7 +8,7 @@ use ssz_types::{typenum::U300, VariableList};
 use crate::types::portal_wire::CustomPayload;
 
 /// Used to respond to pings which the node can't handle
-#[derive(PartialEq, Debug, Clone, Encode, Decode)]
+#[derive(PartialEq, Clone, Encode, Decode)]
```
morph-dev, line 11: nit: In my opinion, if structure implements `PartialEq`, it should also implement [`Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html) if it applies for it.

morph-dev (approved):
Few small comments. 

:shipit:

crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
KolbyML: Good catch

crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
```diff
@@ -30,6 +32,18 @@ impl PingError {
     }
 }
 
+impl Debug for PingError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let message = String::from_utf8(self.message.to_vec())
```
morph-dev, line 37: nit: you should be able to use `&*self.message` here as well (instead of calling `to_vec()`)

morph-dev (approved):

crates/ethportal-api/src/types/ping_extensions/extensions/type_65535.rs
KolbyML: That doesn't compile `String::from_utf8` only takes a Vec

KolbyML: &* gives you a slice

KolbyML: Looks like milos was looking at str::from_utf8 well I was looking at String::from_utf8

PR #1660, opened 2025-02-01
feat: add benchmark for trin

KolbyML:
### What was wrong?

We didn't have a reliable way to test if we made a performance improvement or regression in Trin.

### How was it fixed?

By building `trin-bench` which contains three parts
- Trin Sender, a trin instance which sends offers
- Trin Receiver a trin instance which recieves offers
- Trin Bench the coordinator which gets the data for the test and makes the requests.

You can set the era1 range you want to test with, then run the test which will generate flamegraphs so you can review what functions are using all our CPU cycles

### TODO: in future PR's

add support for other Portal clients, so we can see how performent each client is relative to each other


KolbyML:
setting this PR to draft as I am doing some profiling on it

PR #1664, opened 2025-02-03
fix: generate_spaced_private_keys doesn't generate even keyspace

KolbyML:
### What was wrong?
If you set the count to 128, only half the key range is covered. Milos found the issue

image

### How was it fixed?

Use this first 16 bits instead of first 8 bits, this makes the script slower, but it is more then fast enough. I was able to generate 128 private keys in 10 minutes, we don't generate spaced keys often so I think this accuracy increase is a good trade off

### DIFF
--- a/bin/portal-bridge/src/utils.rs
+++ b/bin/portal-bridge/src/utils.rs
@@ -16,12 +16,12 @@ use serde::{Deserialize, Serialize};
 
 /// Generates a set of N private keys, with node ids that are equally spaced
 /// around the 256-bit keys space.
-// count as u8 is used as a safety precaution against foot-gunning, since it could
+// count as u16 is used as a safety precaution against foot-gunning, since it could
 // take a long time with a bigger input. Separately, a value larger than 256
 // stops getting evenly spread, because of how the code spreads the data by
 // inspecting the first byte from the node ID. Also, values between 16 to
 // 256 would be spread, but less and less evenly.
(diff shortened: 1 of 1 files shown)


morph-dev:
I'm not sure I like this approach. It probably works the best when number of nodes is not power of two, but I think it has wrong properties when it is.

For example, let's say that we want to generate 4 NodeIds. To me, that means that NodeIds should have unique highest 2 bits, and all remaining bits are fully random.

In this case, we will get unique 2 bits (and even this is not 100% guaranteed with current implementation), and next 6 bits will be close to each other. On the first thought, one might think that having other 6 bits the same (or close to each other) is desired property (they seem equally distributed), but I believe that's not desired because we are dealing with XOR distance (I can go into details on this if needed).

KolbyML:
> I'm not sure I like this approach. It probably works the best when number of nodes is not power of two, but I think it has wrong properties when it is.
> 
> For example, let's say that we want to generate 4 NodeIds. To me, that means that NodeIds should have unique highest 2 bits, and all remaining bits are fully random.
> 
> In this case, we will get unique 2 bits (and even this is not 100% guaranteed with current implementation), and next 6 bits will be close to each other. On the first thought, one might think that having other 6 bits the same (or close to each other) is desired property (they seem equally distributed), but I believe that's not desired because we are dealing with XOR distance (I can go into details on this if needed).

I am not too committed to this so I am going to close the PR. Next time I need to generate node_id's I just want the code to work well enough, if you want to make a PR to fix this it would be much appreciated

Issue #1666, opened 2025-02-04
Update `HeaderWithProof` Tracking Issue


njgheorghita:
# Updating the `HeaderWithProof` type

There's been an update to the specs for the `HeaderWithProof` type, which is breaking. All existing `HeaderWithProof`s in our prod dbs will need to be updated / removed. Also, trin needs support for ephemeral proofs. 

@morph-dev & @njgheorghita are taking lead on this

# TODOs
- [ ] Turn off history bridges (to prevent invalid data flowing into the network)
- [ ] Temporarily disable trin node's ability to store header types (to prevent further propagation of invalid data)
- [ ] Write a script to update headers in prod db databases
  - Convert valid headers with proofs. delete any headers that don't have proofs
- [ ] Update trin & bridge to support new header / proof generation (eg. non-ephemeral)
- [ ] Add support for generating / storing / updating ephemeral proofs
  - [ ] Investigate storage strategy (does this require a separate db?)
- [ ] Can we fix incorrect storage reporting metrics in Trin during this update?
- [ ] Deploy changes and restart bridges

# Related Issues
- [Cross-Client Implementation / Deployment Tracking Issue](https://github.com/ethereum/portal-network-specs/issues/363)
- [Spec PR for updated Header type](https://github.com/ethereum/portal-network-specs/pull/362)
- [Storage reporting issue in Trin](https://github.com/ethereum/trin/issues/1653)

PR #1667, opened 2025-02-04
feat: disable state network history header validation

KolbyML:
### What was wrong?
We are in a transition period of enabling full validation for headers on the history network https://github.com/ethereum/trin/issues/1666

image

Since the bridges for the History network will be shutdown during this transition period, we will be disabling the State Networks requirement for History Network headers until the transition period is done

### How was it fixed?

add a constant


njgheorghita (approved):
🚢

PR #1668, opened 2025-02-04
fix: add temporary flag to disable storing headers

njgheorghita:
### What was wrong?
Part of #1666 ...

### How was it fixed?
- Added flag for disabling the storage of all `HeaderWithProof` types
- I spoke with @morph-dev offline about not using a flag and just hardcoding this exception, but I figured that would probably mess with our tests somewhere so the flag seems like the easier solution.

### To-Do

### DIFF
--- a/bin/trin/src/cli.rs
+++ b/bin/trin/src/cli.rs
@@ -216,6 +216,13 @@ pub struct TrinConfig {
         value_parser = max_radius_parser,
     )]
     pub max_radius: Distance,
+
+    #[arg(
+        long = "disable-storing-headers",
+        help = "Disable storing headers in the database. This is a temporary flag used for upgrading the network, and should be removed once the upgrade is complete.",
+        default_value = "false"
+    )]
+    pub disable_storing_headers: bool,
 }
 
 impl Default for TrinConfig {
@@ -249,6 +256,7 @@ impl Default for TrinConfig {
             network: MAINNET.clone(),
             max_radius: max_radius_parser(DEFAULT_MAX_RADIUS)
                 .expect("Parsing static DEFAULT_MAX_RADIUS to work"),
+            disable_storing_headers: false,
         }
     }
 }
--- a/bin/trin/src/cli.rs
+++ b/bin/trin/src/cli.rs
@@ -218,11 +218,11 @@ pub struct TrinConfig {
     pub max_radius: Distance,
 
     #[arg(
-        long = "disable-storing-headers",
-        help = "Disable storing headers in the database. This is a temporary flag used for upgrading the network, and should be removed once the upgrade is complete.",
+        long = "disable-history-storage",
+        help = "Disable storing all history data locally. This is a temporary flag used for upgrading the network, and should be removed once the upgrade is complete.",
         default_value = "false"
     )]
-    pub disable_storing_headers: bool,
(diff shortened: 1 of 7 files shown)


crates/subnetworks/history/src/storage.rs
```diff
@@ -35,6 +36,12 @@ impl ContentStore for HistoryStorage {
         key: &HistoryContentKey,
     ) -> Result<ShouldWeStoreContent, ContentStoreError> {
         let content_id = ContentId::from(key.content_id());
+        // temporarily disable storing all block headers
+        if self.disable_storing_headers {
+            if let HistoryContentKey::BlockHeaderByHash(_) = key {
+                return Ok(ShouldWeStoreContent::NotWithinRadius);
+            }
+        }
```
KolbyML, line 44:
```suggestion
        // temporarily disable storing all block headers
        if self.disable_storing_headers {
            if let HistoryContentKey::BlockHeaderByHash(_) | HistoryContentKey::BlockHeaderByNumber(_) = key {
                return Ok(ShouldWeStoreContent::NotWithinRadius);
            }
        }
```

Don't you need to add a check for BlockHeaderByNumber as well

KolbyML (approved):
:shipit: looks good, there was a check missing for HeaderByNumber, I assume the cluster repo will be updated with --disable-storing-header=true

bin/trin/src/cli.rs
```diff
@@ -216,6 +216,13 @@ pub struct TrinConfig {
         value_parser = max_radius_parser,
     )]
     pub max_radius: Distance,
+
+    #[arg(
+        long = "disable-storing-headers",
+        help = "Disable storing headers in the database. This is a temporary flag used for upgrading the network, and should be removed once the upgrade is complete.",
```
morph-dev, line 222: nit: maybe add link to tracking issue

crates/storage/src/config.rs
```diff
@@ -127,6 +129,7 @@ pub struct PortalStorageConfig {
     pub distance_fn: DistanceFunction,
     pub sql_connection_pool: Pool<SqliteConnectionManager>,
     pub max_radius: Distance,
+    pub disable_storing_headers: bool,
```
morph-dev, line 132: I know that this is only temporarily, but I don't like that this is history network specific thing and it's present in the  PortalStorageConfig.

Instead, I believe we can:
1. add parameter to `initialize_state_network`
2. pass it to `HistoryNetwork::new`
3. pass it to `HistoryStorage::new`

This would make change smaller and also easier to revert.

Again, since this is temporarily, I won't block/insist on this.

morph-dev (commented):
I think we should disable storing all history content (for the time being).

Reason for not storing new bodies and receipts until migration is completed, is that I'm not sure we can verify them if we can't fetch correct headers from the network. Meaning, we might end up doing a lot of unnecessary work..

Also, I'm not sure this content will be gossiped until migration is complete either.

bin/trin/src/cli.rs
njgheorghita: I added this pr to the list of todos in the tracking issue so I'm confident this won't slip through the cracks

morph-dev (approved):
LGTM :ship:

PR #1672, opened 2025-02-07
Update `HeaderWithProof` with latest spec changes

njgheorghita:
### What was wrong?
Specs have been updated to the `HeaderWithProof` type
- Removing the `None` type for the proof
- Removes the ssz union

### How was it fixed?
- Updated the `HeaderWithProof` type
- Renamed `PreMergeAccumulatorProof` -> `HistoricalHashesAccumulatorProof` to better match the specs
- I tested these locally against the test vectors in https://github.com/ethereum/portal-spec-tests/pull/34 but this pr will have failing tests until that is merged
- `portal-bridge` / some test utils have a *temporary* invalid proof. The next step is to update the bridge to generate valid proofs, but I didn't want to combine these two prs otherwise it would get pretty large. Next, I'll work on updating the bridge and then update all the invalid proofs to be valid

### To-Do


njgheorghita:
Ready for an intial review on this pr. The failing tests pass locally when I use https://github.com/ethereum/portal-spec-tests/pull/34 as a source for test vectors. But I'm happy to wait until that's merged to merge this one through

KolbyML:
> * `portal-bridge` / some test utils have a _temporary_ invalid proof. The next step is to update the bridge to generate valid proofs, but I didn't want to combine these two prs otherwise it would get pretty large. Next, I'll work on updating the bridge and then update all the invalid proofs to be valid


I am not aware of a reason why you would need to `but I didn't want to combine these two prs otherwise` if you updated the bridge to generate valid proofs first

KolbyML (approved):
:shipit: PR looks good. I am a little confused by this statement `But I'm happy to wait until that's merged to merge this one through`, because I thought it was assumed we wait for CI passing to merge, so the message is making me second guess myself, but I will ignore that feeling.

crates/ethportal-api/src/types/execution/header_with_proof.rs
```diff
@@ -40,21 +46,26 @@ impl ssz::Encode for HeaderWithProof {
 }
 
 #[derive(Debug, Clone, PartialEq, Decode, Encode, Deserialize)]
-#[ssz(enum_behaviour = "union")]
+#[ssz(enum_behaviour = "transparent")]
```
morph-dev, line 49: We will need both old type (with union and SszNone) until we are done with migration. Otherwise, I won't be able to do the migration (meaning, I won't be able to successfully decode content from the db).

I would also remove `Decode` from derive, because now that it's no longer union, we can't decode it (we need header timestamp). Technically, this is not accurate as format is different enough (I believe) so that only one would decode correctly. But we shouldn't try to decode this object directly, so better not to implement the trait.

crates/ethportal-api/src/types/execution/header_with_proof.rs
morph-dev, line 1: I checked the [spec](https://github.com/ethereum/portal-network-specs/blob/1a7e929283c9578a9a7c4b5c513df726ff41b868/history/history-network.md#block-header) and it seems that `ExecutionBlockProof` is different for `BlockProofHistoricalRoots` and `BlockProofHistoricalSummaries`, while we still use the some type.

Are all updated specs tests passing? Do we cover all these types in them?

crates/ethportal-api/src/types/execution/header_with_proof.rs
```diff
@@ -64,21 +75,33 @@ impl ssz::Decode for HeaderWithProof {
...
         let mut decoder = builder.build()?;
 
         let header_rlp: Vec<u8> = decoder.decode_next()?;
-        let proof = decoder.decode_next()?;
+        let proof: Vec<u8> = decoder.decode_next()?;
         let header: Header = Decodable::decode(&mut header_rlp.as_slice()).map_err(|_| {
             ssz::DecodeError::BytesInvalid("Unable to decode bytes into header.".to_string())
         })?;
-
+        let proof = if header.timestamp <= MERGE_TIMESTAMP {
```
morph-dev, line 87: Should this be `<=` or `<`? At least intuitively for me, MERGE_TIMESTAMP should be the timestamp of the first post-merge block (but I don't know if that's try in practice).

Either way, I think we should have tests to cover these corner cases, meaning have a test case for the last pre-merge and first post-merge blocks (also shanghai)

crates/ethportal-api/src/types/execution/header_with_proof.rs
```diff
@@ -40,21 +46,26 @@ impl ssz::Encode for HeaderWithProof {
...
-    None(SszNone),
-    PreMergeAccumulatorProof(PreMergeAccumulatorProof),
+    HistoricalHashesAccumulatorProof(HistoricalHashesAccumulatorProof),
     HistoricalRootsBlockProof(HistoricalRootsBlockProof),
     HistoricalSummariesBlockProof(HistoricalSummariesBlockProof),
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-pub struct PreMergeAccumulatorProof {
+pub struct HistoricalHashesAccumulatorProof {
```
morph-dev, line 59: Not strictly part of this PR, but can't we replace this structure with:
```rust
pub type HistoricalHashesAccumulatorProof = FixedVector<B256, typenum::U15>;
```

morph-dev (commented):
Left some high level comments.

I plan to work on top of this PR, and I might suggest some changes if they are blocking my work.

crates/ethportal-api/src/types/execution/header_with_proof.rs
morph-dev: Thinking more about this, I think that old type should remain the same, and the new type should have new name

crates/ethportal-api/src/types/execution/header_with_proof.rs
njgheorghita: All the updated specs tests are passing now. I actually don't think the test vectors cover all cases (aka, 1 pre-merge, 1 capella, 1 post-capella). This is worth addressing. 

Also, looking at the spec, it looks like our naming convention is inverse of what's in the spec, which we should update to be 1-1. Though, if I'm reading everything correctly, I actually think our logic is sound. "our" `HistoricalSummariesProof` type and `HistoricalRootsProof` type (which I think are the types you're referring to?) seem to be up to spec. But, honestly it's confusing with all the variable naming schemes. I don't quite consider this a blocker for this pr? I'd rather address it in a follow-up pr where I'll update all the names to be spec-compliant and add missing test vectors? But can do it here if you think that's worthwhile.

njgheorghita:
@KolbyML Sorry if I was confusing, I just wanted to avoid bloating this pr with too many changes. And you're right we shouldn't be merging with failing ci, I was just saying so funnily

@morph-dev I moved the new types into an `updated` module. I incorporated most of your suggestions from your branch here. Lmk how you feel about updating the naming/test vectors in this pr or a follow up. I've copied the updated test vectors from the pending pr here so merging is no longer blocked by a failing ci.

crates/ethportal-api/src/types/updated/content_value.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        Err(ContentValueError::UnknownContent {
+            bytes: hex_encode(buf),
+            subnetwork: Subnetwork::History,
+        })
+    }
+}
+
+#[cfg(test)]
+mod test {
```
morph-dev, line 59: Even tho it might be tested in some other places indirectly, I think we should have tests here that test that encoding and decoding works as expected.

This doesn't have to be done in this PR (we should have test vectors for this).

morph-dev (approved):
Looks good. Few suggestions, but I'm not going to insist on them:
- I would move new `content_value.rs` file into `ethportal-api/src/types/content_value/history_pr1666.rs`
    - I'm not the biggest fan of `_pr1666` suffix, and either `_new` or `_updated` is fine as well. All of this is only temporarily.
- We can also move new `header_with_proof.rs` one level up and name it: `header_with_proof_pr1666.rs`
- If I'm not mistaken, the test files (in `validation/src/asssets/fluffy`) are currently only used with this new `header_with_proof.rs`, so I would move them somewhere in `ethportal-api` crate.

crates/ethportal-api/src/types/updated/content_value.rs
njgheorghita: I removed the existing tests since they were duplicates. But I guess the tests you're referencing are for encoding/decoding the `HistoryContentValue` and not necessarily the inner type. I'll circle back to this pr and add these later

njgheorghita:
> If I'm not mistaken, the test files (in validation/src/asssets/fluffy) are currently only used with this new header_with_proof.rs, so I would move them somewhere in ethportal-api crate.

I left them where they were because I'd like to remove that entire test asset folder after the "upgrade" since they're all duplicates from the `portal-spec-tests` repo (and once the updated vectors get merged, we won't need these files anymore). (one exception might be `epoch_acc.bin` but I'll find a new home for that file if necessary).

I updated the modules to the new naming convention (`*_new.rs`). 

> We can also move new header_with_proof.rs one level up and name it: 

I ended up moving it into `src/types/execution/` so that it sits at the same level as it's "older" version

PR #1673, opened 2025-02-08
refactor: use util functions for reading from portal-spec-tests submodule

morph-dev:
### What was wrong?

We have several places where we reference files in portal-spec-tests submodule directly.
Instead, we should use utility functions.

### How was it fixed?

Added few more utility functions and remove all direct references to files.
Also remove some files that were duplicates.


KolbyML (approved):
:shipit: looks good

njgheorghita (approved):
🚢

PR #1674, opened 2025-02-08
feat: remove LegacyHistory store version

morph-dev:
### What was wrong?

The `LegacyHistory` store version is not used for a long time.  The only remaining use was the migration from it to `IdIndexedV1` store, which is no longer needed.

### How was it fixed?

Removed it, and clean up MockContentStore type.


KolbyML (approved):
:shipit: looks good

Issue #1653, opened 2025-01-29
Storage capacity estimates


morph-dev:
We are wrongly estimating our storage capacity usage. Goal of this issue is to enumerate all known issues on this topic, so we can start discussion on how to address it.

## We underestimate how much space is used

Currently, we estimate how much disk space is used by adding lengths of `content_id`, `content_key` and `content_value`. However, there is extra overhead, I believe caused by:
- other columns (rowid, distance_short, content_size)
- data needed for indexes
- other "empty" space in DB pages

I strongly believe that overhead is dependent on the number of stored items, not the size of the items. Because average size of the content on the state network is fraction of average size of the content on the history network (see data below), our estimate is significantly worse for the state network data.

I extracted the data from 3 different types of prod nodes:

#### History node (4444 nodes) - actual disk usage 36 183 mb

| network | radius | content_count | content_used | content_capacity |
|---|---|---|---|---|
| history | 1.9% | 1 344 329 | 34 999 mb | 35 000 mb |

#### State nodes - actual disk usage 54 198 mb

| network | radius | content_count | content_used | content_capacity |
|---|---|---|---|---|
| history | 1.2% | 761 237 | 17 500 mb | 17 500 mb |
| state | 37% | 60 700 610 | 17 498 mb | 17 500 mb |
| beacon | 0% | 326 | 434 mb | 0mb |

#### Big storage nodes (21m state) - actual disk usage 639 759 mb

| network | radius | content_count | content_used | content_capacity |
|---|---|---|---|---|
| history | 13% | 218 968 | 5 000 mb | 5 000 mb |
| state | 100% | 2 019 082 740 | 392 787 mb | 500 000 mb |
| beacon | 0% | 129 | 15 mb | 0 mb |

We can observe that actual usage in nodes that have both history and state content is significantly worse (up to 60%) then when we only have history network (only ~3% extra)

## Beacon network declares that it doesn't need any space

While beacon clearly uses some disk space, it's unclear how much that is. Also, it never reserves space from the one provided via flag (meaning, we just assume it uses zero and distribute specified capacity between history and state).

I would also highlight that I didn't investigate if beacon correctly reports it's usage (I know there are multiple tables, and I don't know if we are actually reporting stats from all of them).

## Disk usage never goes down

Something to be tested, but I'm pretty sure that disk usage never goes down, even if we remove a lot of content at once.
My understanding is that sqlite keeps free database pages around, so it can reuse them later.

[...]

PR #1676, opened 2025-02-09
feat: improve estimated disk usage

morph-dev:
### What was wrong?

We underestimate the disk usage. See #1653 for details.

### How was it fixed?

Based on the analysis in #1653, I added fixed number of bytes per content item that should be used when estimating disk usage (together with already counted number of bytes for content_id, content_key and content_value).

This way, the data that is actually store in the db (`content_size = content_id.len() + content_key.len() + content_value.len()`) doesn't have to change and we can easily update our estimate in the future if we need to.

### How was it tested?

I had local trin client that was running for some time (~20h) last week. Storage was configured to use 1GB of data, and all 3 subnetworks are enabled (state and history had 500MB capacity each). The actual size of the DB was 1447 MB (45% than desired).

Starting trin client with this PR would trigger pruning at startup. Pruning would run until our estimated disk usage is below 1GB. This however doesn't affect the actual disk usage. To do so, we would have to run "VACUUM".

Before running this code, I made a copy of db. That way I was able to re-run the code, and confirm that results are stable.

#### Run 1: Prune + Vacuum

I started new trin client (with this PR) and observed that:

- estimated disk usage was:
    - 1063 MB for state + 526 MB for history (compared to 500 MB each before) = 1590 MB (compared to actual 1447 MB)
- pruning executed at startup
- estimated disk usage after pruning: 498 MB for state + 495 MB for history (both under specified capacity)
    - the actual DB size didn't decrease because we don't use Vacuum
- I stop trin and run Vacuum manually
    - the db size decreased to 785 MB

#### Run 2: Vacuum + Prune + Vacuum

I run Vacuum on the original DB which decreased db size to 1154 MB.
Then I repeated steps from the first run and observed that:

- The estimated disk usage was the same (as expected)
- Pruning ended with the same estimated usage
- DB size after pruning was still 1.154 MB (expected, again because of the vacuum)
- DB size after vacuum was also 785 MB.

### Conclusion

It's worth observing that running Vacuum on the original db decreased it's size from 1447 MB to 1154 MB. That means that defragmentation added ~25% disk usage.  If we assume the same defragmentation, resulting db would occupy 984 MB (instead of 785 MB), which would leave us under close to the target capacity (with a bit of room).

This defragmentation ratio is not fixed. It depends on the content that is stored (it's different between state and history content) and how long client is running (longer running client tends to have higher defragmentation, but probably there is a limit). The original DB had radius between 2% and 4% for both history and state network, which implies that there was significant defragmentataion (if I remember correctly, I run the code before the max 5% radius PR was merged), which would imply that defragmentation might not be significantly higher even in the worse case.

### Things to consider

This would result in all our nodes (or at least the ones that reached their capacity) to trim content at startup. Different nodes are affected differently:

- history nodes - they wouldn't be significantly affected, as disk usage estimate wasn't as bad (in comparison to state)
    - in my local example, I deleted only ~6% of history content
- regular state nodes - Since they were restarted recently with 5% max radius, they didn't reach they capacity yet. This change would just change how much capacity they estimated they use, but they wouldn't prune anything (unless they get full by the time this PR is deployed)
- big storage nodes - These would be impacted the most because their defragmentation is very low (so far they only insert and never delete content). Running this PR would cause them to delete 1/3 of their content, which would:
    - run pruning very long (making them unresponsive for quite some time)
    - leave it with significant overestimated disk usage (I believe good in the long run)

### DIFF
--- a/crates/storage/src/versioned/id_indexed_v1/store.rs
+++ b/crates/storage/src/versioned/id_indexed_v1/store.rs
@@ -80,7 +80,11 @@ impl<TContentKey: OverlayContentKey> VersionedContentStore for IdIndexedV1Store<
         let mut store = Self {
             radius: config.max_radius,
             pruning_strategy,
-            usage_stats: UsageStats::default(),
+            usage_stats: UsageStats::new(
+                /* entry_count= */ 0,
+                /* total_entry_size_bytes= */ 0,
+                extra_disk_usage_per_content_bytes(&config.content_type),
+            ),
             metrics: StorageMetricsReporter::new(subnetwork),
             _phantom_content_key: PhantomData,
             config,
@@ -102,7 +106,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
             debug!(
                 Db = %self.config.content_type,
                 "High storage usage ({}) -> Pruning",
-                self.usage_stats.total_entry_size_bytes,
+                self.usage_stats.estimated_disk_usage_bytes(),
             );
             // ignore dropped content...
             self.prune()?;
@@ -113,8 +117,8 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
             debug!(
                 Db = %self.config.content_type,
                 "Used capacity ({}) is above target capacity ({}) -> Using distance to farthest for radius",
-                self.usage_stats.total_entry_size_bytes,
-                self.pruning_strategy.target_capacity_bytes()
+                self.usage_stats.estimated_disk_usage_bytes(),
+                self.pruning_strategy.target_capacity_bytes(),
             );
             self.set_radius_to_farthest()?;
         } else if self.config.storage_capacity_bytes == 0 {
@@ -128,7 +132,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
             debug!(
                 Db = %self.config.content_type,
                 "Used capacity ({}) is below target capacity ({}) -> Using MAX radius ({})",
-                self.usage_stats.total_entry_size_bytes,
+                self.usage_stats.estimated_disk_usage_bytes(),
                 self.pruning_strategy.target_capacity_bytes(),
                 self.config.max_radius,
             );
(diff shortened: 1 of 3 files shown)


crates/storage/src/versioned/id_indexed_v1/store.rs
```diff
@@ -252,7 +256,7 @@ impl<TContentKey: OverlayContentKey> IdIndexedV1Store<TContentKey> {
 
         let content_id = content_id.to_vec();
         let content_key = content_key.to_bytes().to_vec();
-        let content_size = content_id.len() + content_key.len() + content_value.len();
+        let content_size = Self::calculate_content_size(&content_id, &content_key, &content_value);
```
KolbyML, line 259: this helper function seems a little overkill

KolbyML (approved):
:shipit: overall the PR looks good

carver (approved):
LGTM! 🚢 

>big storage nodes - These would be impacted the most because their defragmentation is very low (so far they only insert and never delete content). Running this PR would cause them to delete 1/3 of their content, which would:
>    run pruning very long (making them unresponsive for quite some time)

Do you think this is something we need to address/prevent? One option would be to simultaneously increase the config of how much data the big nodes are allowed to store, so we don't trigger a huge pruning. But maybe it's just fine to trigger a huge prune.

---

I always get a little nervous when there is a refactor mixed in with a functional change, that there will be bug that looks like an innocuous part of the refactor. It would help me if it were split into commits next time, if you don't want to split into different PRs.

carver:
Interesting how big an effect the vacuum can have, from just the fragmentation! I guess I can make some sense of that, when each record can have such different size. That would make it harder to reuse a deleted record, than for a more uniform database.

I don't think it's a top priority, but I can imagine wanting to schedule an occasional vacuum. I still lean against turning autovacuum on, though.

morph-dev (commented):
> Do you think this is something we need to address/prevent? One option would be to simultaneously increase the config of how much data the big nodes are allowed to store, so we don't trigger a huge pruning. But maybe it's just fine to trigger a huge prune.

We will have to update flags. Currently we specify capacity to be 500 GB, because estimated capacity is ~400 GB while actual usage is ~640 GB. This was done so that when estimated capacity grows up to 500 GB, with decent amount of fragmentation, we stay under disk limit.

With this change, estimated disk usage would grow to 1.4 GB (while we would still use only 640GB).
If I set storage capacity to 900GB (that is the plan), we would have to prune around 1/3 of the content. 

Rollout plan would be the following:

- pick one big-storage node
    - backup db on my machine (just in case, hopefully not needed)
    - deploy and observe logs while pruning is happening
    - once it's done, check radius, remaining content, estimated disk usage and verify that they are at expected level
    - if not, rollback original db and debug
- if everything went ok and pruning takes too long (most likely), deploy in 2-3 phases to other  big-storage nodes

> I always get a little nervous when there is a refactor mixed in with a functional change, that there will be bug that looks like an innocuous part of the refactor. It would help me if it were split into commits next time, if you don't want to split into different PRs.

I will try to do it in the future.

PR #1677, opened 2025-02-09
refactor: remove unused Distance type

morph-dev:
### What was wrong?

The `ethportal_api::types::portal::Distance` type is not used.

My auto complete confuses it frequently with `ethportal_api::types::distance::Distance` (the actual type that we use).

### How was it fixed?

Removed the type.


KolbyML (approved):
:shipit: looks good

PR #1678, opened 2025-02-11
refactor: update names to be spec compliant for new hwp

njgheorghita:
### What was wrong?
The naming scheme we used for the `HeaderWithProof` types doesn't match the [spec](https://github.com/ethereum/portal-network-specs/blob/master/history/history-network.md#block-header) and it can get very confusing. 

### How was it fixed?
- Updated the names (only in the "new" `HeaderWithProof` type, don't think this is necessary for the old type we're about to sunset) and docs to match the spec.

### To-Do

### DIFF
--- a/crates/ethportal-api/src/types/execution/header_with_proof_new.rs
+++ b/crates/ethportal-api/src/types/execution/header_with_proof_new.rs
@@ -13,17 +13,21 @@ use crate::{
     Header,
 };
 
-/// The accumulator proof for the pre-merge blocks.
-pub type HistoricalHashesAccumulatorProof = FixedVector<B256, typenum::U15>;
+/// The accumulator proof for EL BlockHeader for the pre-merge blocks.
+pub type BlockProofHistoricalHashesAccumulator = FixedVector<B256, typenum::U15>;
 
-/// Proof that execution header root is part of BeaconBlock, post-Merge and pre-Capella
-pub type BeaconBlockProofPreCapella = FixedVector<B256, typenum::U11>;
-/// Proof that execution header root is part of BeaconBlock, post-Capella
-pub type BeaconBlockProof = VariableList<B256, typenum::U12>;
-/// Proof that BeaconBlockHeader root is part of HistoricalRoots
-pub type HistoricalRootsProof = FixedVector<B256, typenum::U14>;
-/// Proof that BeaconBlockHeader root is part of HistoricalSummaries
-pub type HistoricalSummariesProof = FixedVector<B256, typenum::U13>;
+/// Proof that EL block_hash is in BeaconBlock -> BeaconBlockBody -> ExecutionPayload
+/// for TheMerge until Capella
+pub type ExecutionBlockProof = FixedVector<B256, typenum::U11>;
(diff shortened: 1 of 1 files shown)


crates/ethportal-api/src/types/execution/header_with_proof_new.rs
```diff
@@ -82,47 +85,61 @@ impl ssz::Encode for BlockHeaderProof {
...
+            BlockHeaderProof::HistoricalSummaries(proof) => proof.ssz_bytes_len(),
         }
     }
 }
 
 /// The struct holds a chain of proofs. This chain of proofs allows for verifying that an EL
 /// `BlockHeader` is part of the canonical chain. The only requirement is having access to the
 /// beacon chain `historical_roots`.
-// Total size (8 + 1 + 3 + 1 + 14) * 32 bytes + 4 bytes = 868 bytes
+/// Total size (8 + 1 + 3 + 1 + 14) * 32 bytes + 4 bytes = 868 bytes
```
morph-dev, line 112: I don't think that this formula is correct. Shouldn't it be: `(14 + 1 + 11) * 32 + 8` ?
The original formula is also weird because it's unclear where values come from

crates/ethportal-api/src/types/execution/header_with_proof_new.rs
```diff
@@ -13,17 +13,19 @@ use crate::{
...
-pub type BeaconBlockProof = VariableList<B256, typenum::U12>;
-/// Proof that BeaconBlockHeader root is part of HistoricalRoots
-pub type HistoricalRootsProof = FixedVector<B256, typenum::U14>;
-/// Proof that BeaconBlockHeader root is part of HistoricalSummaries
-pub type HistoricalSummariesProof = FixedVector<B256, typenum::U13>;
+/// The accumulator proof for EL BlockHeader for the pre-merge blocks.
+pub type BlockProofHistoricalHashesAccumulator = FixedVector<B256, typenum::U15>;
+
+/// Proof that EL block_hash is in BeaconBlock -> BeaconBlockBody -> ExecutionPayload
+pub type ExecutionBlockProof = FixedVector<B256, typenum::U11>;
```
morph-dev, line 20: I would add a comment to clarify that this is Merge -> Capella

crates/ethportal-api/src/types/execution/header_with_proof_new.rs
```diff
@@ -13,17 +13,19 @@ use crate::{
...
+/// Proof that EL block_hash is in BeaconBlock -> BeaconBlockBody -> ExecutionPayload
+pub type ExecutionBlockProofCapella = VariableList<B256, typenum::U12>;
```
morph-dev, line 22: Also here: Post-Capella

crates/ethportal-api/src/types/execution/header_with_proof_new.rs
njgheorghita: Yeah, I agree with your math, but not sure I see the point of this comment overall so I just deleted it... If someone is actually concerned with the size of the proof then it's fairly simple to recalculate it

PR #1670, opened 2025-02-06
chore: pin crunchy for Windows compatibility

carver:
### What was wrong?

Windows build is failing: https://github.com/ethereum/trin/actions/runs/13189513029/job/36819458339

### How was it fixed?

Pin crunchy for all dependencies

PR #1680, opened 2025-02-12
chore: add cross-compile checks to every PR

carver:
Recently, cross was not working because of a dependency conflict. See #1670 

We only found out at the moment of release, which caused a bunch of downstream issues. For example:
- Either way that you try to trigger the follow-up release will cause
  the Release Draft to be missing all the real changes (Tried repushing
  the tag, and skipping the version into the next one)
- There's no other good way to test if whatever was broken the first
  time is fixed. You don't want to test your change by pushing to
  master.
- It makes a messy list of commits on master.

This change tests cross compiles at every PR, to reduce the chances of a similar issue happening again.

~I don't know a good way to test this locally. Any suggestions besides merging and finding out?~ Ah, great, it runs immediately in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,3 +25,61 @@ jobs:
 
       - name: Validate PR commits with commitlint
         run: sudo npx commitlint --from ${{ github.event.pull_request.head.sha }}~${{ github.event.pull_request.commits }} --to ${{ github.event.pull_request.head.sha }} --verbose
+
+  check-cross-build:
+    # This is a stripped-down version of the full release build in .github/workflows/release.yml
+    # Some key changes:
+    # - No need to extract the version of the build
+    # - No need to upload artifacts
+    # - Fewer architectures, just to verify that some custom cases work
+    name: Verify cross-compilation
+    strategy:
+      matrix:
+        arch:
+          [
+            x86_64-unknown-linux-gnu,
+            aarch64-apple-darwin,
+            x86_64-pc-windows-gnu,
+          ]
+        include:
+          - arch: x86_64-unknown-linux-gnu
+            platform: ubuntu-20.04
+            profile: release
+          - arch: aarch64-apple-darwin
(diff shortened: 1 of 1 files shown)


carver:
Hm, I'm wondering if I should build it as "debug", which I expect to run faster. The downside is that it introduces another difference between what we are testing in the PR and what is actually running for a trin release.

KolbyML:
> Hm, I'm wondering if I should build it as "debug", which I expect to run faster. The downside is that it introduces another difference between what we are testing in the PR and what is actually running for a trin release.

I wonder how much faster it would be if at all, if it is only like a 3 minute different might as well do release, but we will see

carver:
> > Hm, I'm wondering if I should build it as "debug", which I expect to run faster. The downside is that it introduces another difference between what we are testing in the PR and what is actually running for a trin release.
> 
> I wonder how much faster it would be if at all, if it is only like a 3 minute different might as well do release, but we will see

Interesting. Yeah, the release build took ~7.5min. I guess that's fast enough. It failed when I tried to use a debug release and I'm sure I could figure it out, but I think it's better to just match the final release build anyway, so I'll revert it and mark ready for review.

KolbyML (approved):
:shipit: PR looks good

carver:
I guess it is reasonable to drop the windows build in Circle, now. Thoughts?

carver:
> One is a native build being built on Windows itself and one is a cross compile

This is a good callout. I imagine we probably want to run some integration tests on the windows build on a windows machine. I guess I slightly lean toward leaving it in place, to support new Windows tests.

KolbyML:
> > One is a native build being built on Windows itself and one is a cross compile
> 
> This is a good callout. I imagine we probably want to run some integration tests on the windows build on a windows machine. I guess I slightly lean toward leaving it in place, to support new Windows tests.

Sounds good

njgheorghita (approved):
🚢

PR #1683, opened 2025-02-13
fix: update deprecated ubuntu base dependencies

njgheorghita:
### What was wrong?
Fixes the broken build step in our release workflow.

### How was it fixed?
- Update ubuntu base version, which is in the process of being [browned out](https://github.com/actions/runner-images/issues/11101)
- We also need to update our node version in the `commitlint` workflow to avoid a different deprecation error

### To-Do


carver:
> * avoid a different deprecation error

Any more info? Can you paste the error that was happening

> Fixes the broken build step in our release workflow.

From our side-chat: I am still not convinced that anything was broken in the release flow.

carver (approved):
LGTM, ready to merge when CI turns green

testing/utp/docker/circleci/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04
+FROM ubuntu:24.04
```
KolbyML, line 1: why can't we use latest here?

.github/workflows/release.yml
```diff
@@ -15,7 +15,7 @@ env:
 jobs:
   extract-version:
     name: extract version
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
```
KolbyML, line 18: same with here?

testing/utp/docker/circleci/Dockerfile
njgheorghita: What are your thoughts? @carver seems to think that it's less stable to pin to `latest` than to a specific version

KolbyML: Isn't ubuntu latest always a stable release of ubuntu? Isn't latest just the latest stable release

KolbyML: image

carver: > why can't we use latest here?

That was my suggestion. The primary benefit I see of having a stable target (stable in the sense of best-tested, but also not changing), is that it eliminates uncertainty at new Ubuntu releases. By using latest, our release flow could break without warning or any change on our end, if some feature or tool works differently on a new Ubuntu release.

It's sort of like the equivalent of using a Cargo.lock file, so that we can choose when we're ready to try a new dependency. The base OS is our dependency in the release flow.

KolbyML: So latest tracks the latest LTS release so currently that is 24.04 the next one will be 26.04

KolbyML (approved):
I think we should use latest, for the docker builds and CI, for the release builds it is fine to tie it to 24.04

I don't think it matters that much at the end of the day as the release cycle is 2 years anyway

So I am trying to say I am not blocking the PR, go for it

njgheorghita:
Cool, yeah going ahead with the merge. I think the difference between the two options is negligible

PR #1684, opened 2025-02-15
feat: create fn for history content migration

morph-dev:
### What was wrong?

We need logic for history content migration. See #1666 for details.

### How was it fixed?

Created new `ContentType`s that will be used for new content types.
This way, we can use the same implementation of content store, because new type will correspond to the new sql table.

This also makes the migration simple: we have to delete content from original table, and either discard or insert into new table.

### To-Do

The migration function is not used at the moment, so this PR can be merged as is.
The followup work, that can be done either as part of this PR or as a followup PR (depending on the review status of this PR and when I finish it):
- Add tests for migration
- Add migration metrics/stats that can be printed periodically
- Decide what we do regarding post-merge bodies and receipts

### DIFF
--- a/crates/subnetworks/history/src/storage_migration.rs
+++ b/crates/subnetworks/history/src/storage_migration.rs
@@ -1,0 +1,173 @@
+use ethportal_api::{
+    types::{
+        content_value::{
+            history::HistoryContentValue as OldHistoryContentValue,
+            history_new::HistoryContentValue as NewHistoryContentValue,
+        },
+        execution::{header_with_proof as old_hwp, header_with_proof_new as new_hwp},
+        network::Subnetwork,
+    },
+    ContentValue, HistoryContentKey, OverlayContentKey, RawContentValue,
+};
+use rusqlite::{named_params, types::Type};
+use tracing::{debug, error, info};
+use trin_storage::{
+    error::ContentStoreError,
+    versioned::{create_store, ContentType, IdIndexedV1Store, IdIndexedV1StoreConfig},
+    PortalStorageConfig,
+};
+
+mod sql {
+    pub const TABLE_EXISTS: &str =
+        "SELECT name FROM sqlite_master WHERE type='table' AND name='ii1_history';";
+
+    pub const BATCH_DELETE: &str = "
+        DELETE FROM ii1_history
+        WHERE rowid IN (
+            SELECT rowid
+            FROM ii1_history
+            ORDER BY content_id
+            LIMIT :limit
+        )
+        RETURNING content_key, content_value;
+    ";
+
+    pub const DROP_TABLE: &str = "DROP TABLE ii1_history;";
+}
+
+const BATCH_DELETE_LIMIT: usize = 100;
+
+#[allow(unused)]
+pub fn maybe_migrate(config: &PortalStorageConfig) -> Result<(), ContentStoreError> {
+    let conn = config.sql_connection_pool.get()?;
+
+    if !conn.prepare(sql::TABLE_EXISTS)?.exists(())? {
+        info!("Legacy history table doesn't exist!");
+        return Ok(());
+    }
+
+    info!("Legacy history table exists. Starting migration.");
+
+    let config = IdIndexedV1StoreConfig::new(
+        ContentType::HistoryEternal,
+        Subnetwork::History,
+        config.clone(),
+    );
+
+    let mut store: IdIndexedV1Store<HistoryContentKey> = create_store(
+        ContentType::HistoryEternal,
+        config.clone(),
+        config.sql_connection_pool,
+    )?;
+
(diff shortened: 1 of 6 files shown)


crates/subnetworks/history/src/storage_migration.rs
```diff
@@ -0,0 +1,173 @@
+use ethportal_api::{
+    types::{
+        content_value::{
+            history::HistoryContentValue as OldHistoryContentValue,
+            history_new::HistoryContentValue as NewHistoryContentValue,
+        },
+        execution::{header_with_proof as old_hwp, header_with_proof_new as new_hwp},
```
KolbyML, line 7:
```suggestion
        execution::{header_with_proof::BlockHeaderProof as OldBlockHeaderProof, header_with_proof_new::{BlockProofHistoricalHashesAccumulator, BlockHeaderProof, BlockProofHistoricalRoots, BlockProofHistoricalSummaries, HeaderWithProof}},
```

old_hwp is only used for `header_with_proof::BlockHeaderProof as OldBlockHeaderProof`, so I think we can pretty easily avoid using `hwp`

crates/subnetworks/history/src/storage_migration.rs
```diff
@@ -0,0 +1,173 @@
...
+                        beacon_block_root: historical_summaries_proof.beacon_block_root,
+                        execution_block_proof,
+                        slot: historical_summaries_proof.slot,
+                    };
+                    new_hwp::BlockHeaderProof::HistoricalSummaries(proof)
+                }
+            };
+            let new_content_value =
+                NewHistoryContentValue::BlockHeaderWithProof(new_hwp::HeaderWithProof {
+                    header: old_hwp.header,
```
KolbyML, line 163:
```suggestion
        OldHistoryContentValue::BlockHeaderWithProof(old_header_with_proof) => {
            let proof = match old_header_with_proof.proof {
                OldBlockHeaderProof::None(_) => return Ok(None),
                OldBlockHeaderProof::PreMergeAccumulatorProof(
                    pre_merge_accumulator_proof,
                ) => {
                    let proof = BlockProofHistoricalHashesAccumulator::new(
                        pre_merge_accumulator_proof.proof.to_vec(),
                    )
                    .map_err(|err| ContentStoreError::InvalidData {
                        message: format!("Invalid HistoricalHashes proof: {err:?}"),
                    })?;
                    BlockHeaderProof::HistoricalHashes(proof)
                }
                OldBlockHeaderProof::HistoricalRootsBlockProof(historical_roots_proof) => {
                    // Note: there is inconsistency between field names because old types are not
                    // in sync with most recent spec.
                    let proof = BlockProofHistoricalRoots {
                        beacon_block_proof: historical_roots_proof.historical_roots_proof,
                        beacon_block_root: historical_roots_proof.beacon_block_root,
                        execution_block_proof: historical_roots_proof.beacon_block_proof,
                        slot: historical_roots_proof.slot,
                    };
                    BlockHeaderProof::HistoricalRoots(proof)
                }
                OldBlockHeaderProof::HistoricalSummariesBlockProof(
                    historical_summaries_proof,
                ) => {
                    // Note: there is inconsistency between field names because old types are not
                    // in sync with most recent spec.
                    let execution_block_proof = new_hwp::ExecutionBlockProofCapella::new(
                        historical_summaries_proof.beacon_block_proof.to_vec(),
                    )
                    .map_err(|err| ContentStoreError::InvalidData {
                        message: format!("Invalid ExecutionBlockProofCapella proof: {err:?}"),
                    })?;
                    let proof = BlockProofHistoricalSummaries {
                        beacon_block_proof: historical_summaries_proof.historical_summaries_proof,
                        beacon_block_root: historical_summaries_proof.beacon_block_root,
                        execution_block_proof,
                        slot: historical_summaries_proof.slot,
                    };
                    BlockHeaderProof::HistoricalSummaries(proof)
                }
            };
            let new_content_value =
                NewHistoryContentValue::BlockHeaderWithProof(HeaderWithProof {
                    header: old_header_with_proof.header,
```

I think we can avoid using `hwp` here

KolbyML (approved):
:shipit: overall looks good, 

I think we can fairly easily avoid using `hwp`, the acronym is very specific to our project, so avoiding it would help keep the code readable for people not as familiar with the project.

Also the code is a little more consistent with `OldHistoryContentValue`

morph-dev:
> I think we can fairly easily avoid using `hwp`, the acronym is very specific to our project, so avoiding it would help keep the code readable for people not as familiar with the project.

At the start, I expected to have way more overlapping types, so I wanted to import modules with different names (so it's always clear what type they refer to). And I wanted module name to be short, so I went with `hwp`. I didn't consider readability for people not familiar with the project as this is only temporary code and should go away soon after migration...

But you are right. I ended up not needing it. So I applied your suggestions.

PR #1686, opened 2025-02-16
fix(ci): make dependencies update bot commit under github-actions account instead of users

KolbyML:
### What was wrong?
image
Currently every time our CI dependencies bot runs it commits under a user account, instead of github actions bot.

This PR and commit isn't being made by a user, it is being made by github actions bot account, so we probably shouldn't be comitting it under a user's account
### How was it fixed?

I tested this locally on my fork https://github.com/KolbyML/trin/pull/5

image

Now it commits the update dependencies commit under the bots account

### DIFF
--- a/.github/workflows/dependencies.yml
+++ b/.github/workflows/dependencies.yml
@@ -31,6 +31,9 @@ jobs:
   update:
     name: Update
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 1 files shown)


.github/workflows/dependencies.yml
```diff
@@ -31,6 +31,9 @@ jobs:
   update:
     name: Update
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
ogenev, line 35: Why do we need the both permissions set to write?

KolbyML: I just removed the unneeded permissions as it appears to work without it https://github.com/KolbyML/trin/pull/6

I added it as apparently in certain conditions it can be required it looks like

https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token
image

PR #1687, opened 2025-02-17
feat(beacon): enable pruning for outdated light client bootstrap data

ogenev:
### What was wrong?
The beacon node keeps all the light client bootstrap data in the DB, but we want to serve and retain only bootstraps that are within the weak subjectivity period ~= 4 months old.

### How was it fixed?
Add a pruning task for light client bootstrap data that runs on every start, and once a day when the client is running.

### To-Do

### DIFF
--- a/crates/subnetworks/beacon/src/storage.rs
+++ b/crates/subnetworks/beacon/src/storage.rs
@@ -1,4 +1,4 @@
-use std::path::PathBuf;
+use std::{path::PathBuf, time::Duration};
 
 use alloy::primitives::B256;
 use ethportal_api::{
@@ -15,12 +15,13 @@ use ethportal_api::{
     },
     BeaconContentKey, OverlayContentKey, RawContentValue,
 };
+use light_client::consensus::rpc::portal_rpc::expected_current_slot;
 use r2d2::Pool;
 use r2d2_sqlite::{rusqlite, SqliteConnectionManager};
 use rusqlite::params;
 use ssz::{Decode, Encode};
 use ssz_types::{typenum::U128, VariableList};
-use tracing::debug;
+use tracing::{debug, error, info};
 use tree_hash::TreeHash;
 use trin_metrics::storage::StorageMetricsReporter;
 use trin_storage::{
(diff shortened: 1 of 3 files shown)


crates/subnetworks/beacon/src/storage.rs
```diff
@@ -661,6 +677,28 @@ impl BeaconStorage {
...
 }
 
+/// Prune old bootstrap data from the database
+fn prune_old_bootstrap_data(
+    sql_connection_pool: &Pool<SqliteConnectionManager>,
+) -> Result<(), ContentStoreError> {
+    let conn = sql_connection_pool.get()?;
+    let expected_current_slot = expected_current_slot();
+    // One slot is 12 sec, and we want the threshold to be 4 months before the current slot
+    let slot_threshold = expected_current_slot - (4 * 30 * 24 * 60 * 60 / 12);
```
KolbyML, line 687: This time period should be more like 4.8 months. It is defined here https://github.com/ethereum/consensus-specs/blob/9d1e7087b3d0e5d3861b67dcf932158e26332fe3/specs/phase0/p2p-interface.md?plain=1#L206

```suggestion
    // One slot is 12 sec, and we want the threshold to be 4 months before the current slot
    let slot_threshold = expected_current_slot - (33042 * 32);
```

crates/subnetworks/beacon/src/storage.rs
```diff
@@ -861,4 +899,34 @@ mod test {
...
     }
+
+    #[test]
+    fn test_beacon_storage_prune_old_bootstrap_data() {
+        let (_temp_dir, config) = create_test_portal_storage_config_with_capacity(10).unwrap();
+        let mut storage = BeaconStorage::new(config).unwrap();
+        let mut value = test_utils::get_light_client_bootstrap(0);
+        // Set the bootstrap slot to outdated value
+        let current_slot = expected_current_slot();
+        let outdated_slot = current_slot - (4 * 30 * 24 * 60 * 60 / 12) - 1;
```
KolbyML, line 910:
```suggestion
        let outdated_slot = current_slot - (33042 * 32) - 1;
```

due to the reasoning above

KolbyML (approved):
:shipit: PR mostly looks good. I think we should use the constant from the spec to determine when we drop content https://github.com/ethereum/consensus-specs/blob/9d1e7087b3d0e5d3861b67dcf932158e26332fe3/specs/phase0/p2p-interface.md?plain=1#L206

Because the spec defines we are supposed to hold the content for more so 4.86 months or ~5 months

crates/subnetworks/beacon/src/storage.rs
ogenev: There is no fixed time for the weak subjectivity period, it depends on network conditions (participation rate, slashing events, etc.) and market factors (e.g., how quickly they can buy, borrow, or bribe stakers). 

While the protocol doesn’t fix an official “weak subjectivity period,” the commonly referenced window for Ethereum mainnet today is around 2–4 months, so I think 4 months is a reasonable threshold.

ogenev:
> :shipit: PR mostly looks good. I think we should use the constant from the spec to determine when we drop content https://github.com/ethereum/consensus-specs/blob/9d1e7087b3d0e5d3861b67dcf932158e26332fe3/specs/phase0/p2p-interface.md?plain=1#L206
> 
> Because the spec defines we are supposed to hold the content for more so 4.82 months or ~5 months

There is no fixed time for the weak subjectivity period, it depends on network conditions (participation rate, slashing events, etc.) and market factors (e.g., how quickly they can buy, borrow, or bribe stakers).

While the protocol doesn’t fix an official “weak subjectivity period,” the commonly referenced window for Ethereum mainnet today is around 2–4 months, so I think 4 months is a reasonable threshold.

We can always change this number in the future if we decide.

PR #1688, opened 2025-02-18
feat(history_migration): don't migrate post-merge headers and adding tests

morph-dev:
### What was wrong?

We decided that we don't want to migrate post-merge headers with proofs (this content was never gossiped, and our original encoding was wrong).

We were also missing tests for migration

### How was it fixed?

Removed migration code for post-merge headers.
Added tests.

### To-Do

For future PRs:
- add migration metrics
- add logic for deleting post-merge bodies and receipts


KolbyML (approved):
:shipit: looks good, I looked over the PR yesterday I must have forgotten to approve it

PR #1689, opened 2025-02-18
feat(beacon): load trusted block root from embedded file

ogenev:
### What was wrong?
To facilitate deployments and UX, on the last call we agreed to rely on an embedded trusted block root file as a fallback mechanism (if the user does not provide it via a `--trusted-block-root`  flag and if the beacon network is enabled).

### How was it fixed?
1. Embed a recent trusted block root into a file.
2. Load the embedded root as a fallback.

The trusted block root is:  [0x107ebc42ce260968c32fba62e2fa6c779f188a0a8c22db0d8afc75801c692fc7](https://beaconcha.in/slot/11094208).

### To-Do


Cargo.toml
```diff
@@ -71,6 +71,7 @@ reth-ipc = { tag = "v1.0.8", git = "https://github.com/paradigmxyz/reth.git"}
 revm = { version = "14.0.3", default-features = false, features = ["std", "secp256k1", "serde-json", "c-kzg"] }
 revm-primitives = { version = "10.0.0", default-features = false, features = ["std", "serde"] }
 rstest = "0.18.2"
+rust-embed = "8.5.0"
```
KolbyML, line 74: We don't need to add another dependency we can use 
```suggestion
```

KolbyML (changes requested):
I think we should use `include_str!("../../resources/genesis/mainnet.json")` instead of adding another dependency

ogenev:
> I think we should use `include_str!("../../resources/genesis/mainnet.json")`, RustEmbed seems pretty overkill, for something `include_str!()` can accomplish easily and is also apart of the Rust Core Library

I changed it to `include_str!()` as it is simpler, and we are not planning to add more files to the folder for now. Ready for re-review.

crates/subnetworks/beacon/src/network.rs
```diff
@@ -113,3 +104,93 @@ impl BeaconNetwork {
...
+            block_root = %db_block_root,
+            "No trusted block root provided. Using latest block root from storage."
+        );
+        return Ok(Some(db_block_root));
+    }
+
+    // 3) If there's still nothing, log and attempt to load the default checkpoint file
+    info!("No trusted block root provided and no block root found in storage. Loading default checkpoint.");
+
+    let hash_bytes = hex_decode(DEFAULT_TRUSTED_BLOCK_ROOT.trim())
```
morph-dev, line 131: This can be simplified to:
```rust
let trusted_block_root = B256::from_str(DEFAULT_TRUSTED_BLOCK_ROOT.trim()).expect("...");
```

You might have to add `use std::str::FromStr`

crates/subnetworks/beacon/src/network.rs
```diff
@@ -113,3 +104,93 @@ impl BeaconNetwork {
...
+    use tempfile::TempDir;
+    use trin_storage::{config::StorageCapacityConfig, PortalStorageConfigFactory};
+
+    use super::*;
+
+    /// Test that embedded "trusted_block_root.txt" can be loaded & parsed into bytes.
+    #[test]
+    fn test_embedded_file_loading_and_parsing() {
+        // Decode from hex to bytes
+        let hash_bytes = hex_decode(DEFAULT_TRUSTED_BLOCK_ROOT.trim())
```
morph-dev, line 154: Similarly here (you can use B256::from_str).

And it would be nice to extract parsing `DEFAULT_TRUSTED_BLOCK_ROOT` into separate function (that way it's easy to test it).

morph-dev (approved):

crates/subnetworks/beacon/src/network.rs
```diff
@@ -77,39 +78,123 @@ impl BeaconNetwork {
...
+
+    // 3) If there's still nothing, log and attempt to load the default checkpoint file
+    info!("No trusted block root provided and no block root found in storage. Loading default checkpoint.");
+    let block_root = parse_trusted_block_root(DEFAULT_TRUSTED_BLOCK_ROOT.trim())?;
+
+    Ok(block_root)
+}
+
+/// Parse the trusted block root from a string.
+fn parse_trusted_block_root(trusted_block_root: &str) -> anyhow::Result<B256> {
```
morph-dev, line 134: nit: since this is used only for `DEFAULT_TRUSTED_BLOCK_ROOT`, you can just rename it to `get_default_trusted_root` and you don't need argument.

crates/subnetworks/beacon/src/network.rs
```diff
@@ -77,39 +78,123 @@ impl BeaconNetwork {
...
+            "No trusted block root provided. Using latest block root from storage."
+        );
+        return Ok(db_block_root);
+    }
+
+    // 3) If there's still nothing, log and attempt to load the default checkpoint file
+    info!("No trusted block root provided and no block root found in storage. Loading default checkpoint.");
+    let block_root = parse_trusted_block_root(DEFAULT_TRUSTED_BLOCK_ROOT.trim())?;
+
+    Ok(block_root)
```
morph-dev, line 130:
```suggestion
    info!("No trusted block root provided and no block root found in storage. Loading default checkpoint.");
    parse_trusted_block_root(DEFAULT_TRUSTED_BLOCK_ROOT.trim())
```

Or if you apply other suggestion, than simply:

```suggestion
    info!("No trusted block root provided and no block root found in storage. Loading default checkpoint.");
    get_default_trusted_root()
```

morph-dev (approved):
:rocket:

crates/subnetworks/beacon/src/network.rs
```diff
@@ -52,7 +53,7 @@ impl BeaconNetwork {
         header_oracle: Arc<RwLock<HeaderOracle>>,
     ) -> anyhow::Result<Self> {
         let config = OverlayConfig {
-            bootnode_enrs: portal_config.bootnodes,
+            bootnode_enrs: portal_config.bootnodes.clone(),
```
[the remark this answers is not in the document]
ogenev: Yeah, because portal_config is used again below.

PR #1690, opened 2025-02-19
feat(history_migration): add history migration metrics

morph-dev:
**Note**: This is built on top of #1688. That PR should be reviewed first.

### What was wrong?

We don't have good way of monitoring history migration and knowing if everything went as expected with migration. 

### How was it fixed?

Added metrics and logging statements.


KolbyML (approved):
:shipit: looks good

PR #1691, opened 2025-02-19
feat: use new history proof types and content value

morph-dev:
### What was wrong?

We have to migrate to new encoding for `HeaderWithProof` and `HistoryContentValue`.
See #1666 for more details.

### How was it fixed?

Type and db migration is already done in previous PRs.

Important things related to this PR:

- updates the HistoryContentValue type that is re-exported in `ethportal-api`
    - `HeaderWithProof` is not re-exported, so we have to change imports
    - old types should no longer be used, except during db migration
- Updates portal-spec-tests submodule to the latest commit (that contains updated test vectors)
- Header validator is updated to use new types 
- Some state network tests will not be passing when header validation is re-enabled
    - This is caused by test vectors containing only header, without proof, and should be fixed by adding header proof
    - Affected tests (as fast I could find): state validator and state peer-to-peer


KolbyML (approved):
:shipit: looks good

PR #1694, opened 2025-02-21
chore: add github action to monthly update trusted block root

ogenev:
### What was wrong?
We must regularly update the trusted block root embedded into Trin to keep it within the weak subjectivity period.

### How was it fixed?
Add GitHub action that gets the latest finalized block root and auto opens a PR updating the trusted block root file.
This workflow runs on the 1st of every month.
This action has been tested, and this is the example [PR](https://github.com/ethereum/trin/pull/1696) that will be open every month.

### To-Do

### DIFF
--- a/.github/workflows/monthly_block_root_update.yml
+++ b/.github/workflows/monthly_block_root_update.yml
@@ -1,0 +1,75 @@
+name: "Monthly Block Root Update"
+
+on:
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

njgheorghita (approved):
🚢

PR #1704, opened 2025-02-24
chore: avoid duplicate crates

rex4539:


KolbyML (changes requested):
crates should only be in the main toml file if they are used in multiple crates, could you revert the changes for the crates only used in one file?

rex4539:
My approach sets the correct foundation in order to prevent duplicate crates even in the future.

KolbyML:
> My approach sets the correct foundation in order to prevent duplicate crates even in the future.

We don't want to use rocksdb in any crate but trin-execution. If a duplicate create is added this refactor can happen then

rex4539:
Sure, it's your repo. Feel free to take over from here :)

KolbyML:
Ok I am closing this PR then. If you ever want to deduplicate duplicate imports in the future, that would be much appreciated

PR #1706, opened 2025-02-26
feat: add block header by number content value

ogenev:
### What was wrong?
Block Header by Number content value implementation is missing. Here are the [specs](https://github.com/ethereum/portal-network-specs/blob/master/history/history-network.md#block-header-by-number).

To reserve the **0x03** selector, we have to implement this for `HistoryContentValue` before ephemeral headers.

### How was it fixed?
Add `BlockHeaderByNumber` history content value variant and update the db migration code.

It seems the `BlockHeaderWithProof`  variant should be renamed to `BlockHeaderByHash` but we can discuss this and it is for another PR.

### To-Do

### DIFF
--- a/crates/ethportal-api/src/types/content_value/history.rs
+++ b/crates/ethportal-api/src/types/content_value/history.rs
@@ -16,6 +16,7 @@ pub enum HistoryContentValue {
     BlockHeaderWithProof(HeaderWithProof),
     BlockBody(BlockBody),
     Receipts(Receipts),
+    BlockHeaderByNumber(HeaderWithProof),
 }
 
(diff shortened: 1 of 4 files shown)


KolbyML (changes requested):
BlockHeaderByNumber isn't a ContentValue variant, both BlockHeaderByHash and BlockHeaderByNumber both use the ContentValue BlockHeaderWithProof.

BlockHeaderByHash and BlockHeaderByNumber are ContentKey's not ContentValues. 0x3 is already reserved for BlockHeaderByNumber.

KolbyML:
image
Notice in the spec both have different content_key's but the same content_value

PR #1707, opened 2025-02-26
docs: update README.md to use correct content_value

morph-dev:
### What was wrong?

The ethportal-api's README file was using old content value.

### How was it fixed?

Updated the content value.

Also renamed `content_item` to `content_value` for consistency.


KolbyML (approved):
:shipit: looks good

PR #1708, opened 2025-02-26
feat!: remove old history content value types and migration

morph-dev:
### What was wrong?

See #1666 for related work.

We still have definition of the old HistoryContentType (and HeaderWithProof). These were needed for db migration.

### How was it fixed?

Deleted old types, db migration, db migration metrics.

Also deleted test vectors from `crates/validation/src/assets/fluffy/...` as we can use the ones from `portal-spec-tests` directly.

### To-Do


crates/ethportal-api/src/types/content_value/history.rs
KolbyML, line 1: if this is being deleted shouldn't history_new.rs be renamed to history.rs now

crates/ethportal-api/src/types/execution/header_with_proof.rs
KolbyML, line 1: same with this if this file is being removed shouldn't `header_with_proof_new.rs` be renamed to `header_with_proof.rs`

crates/ethportal-api/src/types/execution/header_with_proof_new.rs
KolbyML, line 1: this file should be renamed to `header_with_proof.rs`

morph-dev:
@KolbyML  I plan to rename files and do remaining cleanup in the followup pr. 
I wanted to split removal of the old types and renaming of the new types into separate PR for easier review and cleaner commit history.

KolbyML:
> @KolbyML I plan to rename files and do remaining cleanup in the followup pr. I wanted to split removal of the old types and renaming of the new types into separate PR for easier review and cleaner commit history.

I don't think it makes it easier to review or makes the history cleaner so idk, I would expect that to be done in this PR. In the future we should call the old type `_old.rs` instead of using `_new.rs` to avoid this I guess idk.

ogenev (approved):
🛥️

KolbyML (approved):
:shipit:

PR #1709, opened 2025-02-26
feat: remove old store ContentType::History

morph-dev:
### What was wrong?

The old history content type (`ContentType::History`) used for storage is no longer needed.

Instead, we should use either `ContentType::HistoryEternal` or `ContentType::HistoryEphemeral`.

### How was it fixed?

Removed no longer used type.

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1710, opened 2025-02-27
feat: implement `portal_*Ping` rpc with ping extension support

KolbyML:
### What was wrong?
implements https://github.com/ethereum/portal-network-specs/pull/369

tl;dr we weren't exposing ping extensions to the json-rpc. So users couldn't use ping extensions and we couldn't write tests in hive

### How was it fixed?

implementing the spec update


morph-dev:
Ping this PR again once newest changes are in.

morph-dev (commented):
I really don't like that we use and rely on `serde_json::Value`. I also don't think that `crates/portalnet/src/overlay/protocol.rs` should be responsible for decoding it.
And while we do use it mostly in json context, do we know that it will work as expected with other formats if needed.

Why can't we use `PingExtension` instead? Is it because fields of enum variants have the same name?

I believe we can make deserialization work if we use [#[serde(deny_unknown_fields)]](https://serde.rs/container-attrs.html#deny_unknown_fields). This approach might not work if we have two ping extensions with the exactly the same fields in the future, but I think I know how we can do that in slightly more complicated way (in short, we can combine `payload_type` and `payload`, use custom (de)serialization and use [#[serde(flatten)]](https://serde.rs/field-attrs.html#flatten)).

To be honest, I'm leaning towards second approach even if it is more complicated, as it will handle (de)serializing one/two optional fields and a better way. Happy to chat about it.

KolbyML:
We talked on a call a bit, Milos said he will do some more thinking and get back to me.

morph-dev:
We discussed again. Conclusion is:
- Use `serde_json::Value` only as parameter (`portal_*Ping`) and response (`PongInfo`) for JSON rpc server
- Use ping extension type instead of u16 (manually implement (de)serialization)

KolbyML:
> * Use ping extension type instead of u16 (manually implement (de)serialization)

The PR is ready for review. I don't think  this suggestion ^ makes sense on the json-rpc level or protocol level due to how ping extensions work. If we get a ping with an invalid `ping extension type` we are supposed to send a ping error, but if do `Use ping extension type instead of u16 (manually implement (de)serialization)` we wouldn't fail and return a log before we get to the process ping code

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -370,17 +376,68 @@ impl<
...
+                        "Payload type mismatch".into(),
+                    ));
+                }
+                (payload_type, payload.into())
+            }
+            (Some(payload_type), None) => (payload_type, self.default_payload(payload_type)?),
+            (None, Some(payload)) => (payload.payload_type(), payload.into()),
+            (None, None) => (
+                PingExtensionType::Capabilities,
+                ClientInfoRadiusCapabilities::new(
```
morph-dev, line 430: should this be `self.default_payload(PingExtensionType::Capabilities)`?

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -370,17 +376,68 @@ impl<
...
+        payload: Option<PingExtension>,
+    ) -> Result<Pong, OverlayRequestError> {
         // Construct the request.
         let enr_seq = self.discovery.local_enr().seq();
         let data_radius = self.data_radius();
-        let payload =
-            ClientInfoRadiusCapabilities::new(data_radius, self.ping_extensions.raw_extensions())
-                .into();
+
+        let (payload_type, payload) = match (payload_type, payload) {
```
morph-dev, line 417: In my opinion, the return type of this `match` block should just be `ping_extension` (of PingExtension type).

And later you can have
```rust
let request = Ping {
    enr,
    payload_type: ping_extension.payload_type().into(),
    ping_extension.into(),
}
```

crates/portalnet/src/types/node.rs
```diff
@@ -42,13 +42,13 @@ impl Node {
...
     /// Returns the capabilities of the node.
-    pub fn capabilities(&self) -> Option<&[Extensions]> {
+    pub fn capabilities(&self) -> Option<&[PingExtensionType]> {
         self.capabilities.as_deref()
     }
 
     /// Compares the capabilities of the node with the given capabilities.
     /// Returns true if the capabilities are the same.
-    pub fn compare_capabilities(&self, capabilities: &[Extensions]) -> bool {
+    pub fn compare_capabilities(&self, capabilities: &[PingExtensionType]) -> bool {
```
morph-dev, line 51: nit: I think that this function will return true if provided capabilities are subset of `self.capabilities`, not if they are the same (which is stated in the documentation).
Don't know how this function is used, so don't know which is desired behavior and if this is a problem.

crates/rpc/src/errors.rs
```diff
@@ -47,21 +48,37 @@ pub enum RpcServeError {
         message: String,
         trace: Option<Box<QueryTrace>>,
     },
+    /// PayloadTypeNotSupported
+    #[error("Payload type not supported: {message}")]
+    PayloadTypeNotSupported {
```
morph-dev, line 53: All 3 of these messages refer to "payload". I think they should refer to "Ping payload" to avoid confusion.

crates/rpc/src/errors.rs
```diff
@@ -47,21 +48,37 @@ pub enum RpcServeError {
         message: String,
         trace: Option<Box<QueryTrace>>,
     },
+    /// PayloadTypeNotSupported
```
morph-dev, line 51: Also, these doc strings should explain a bit more when this error happens (not just repeat the name of the error).

crates/rpc/src/ping_extension.rs
```diff
@@ -0,0 +1,48 @@
...
+        (Some(payload_type), Some(payload)) => Some(
+            PingExtension::decode_json(payload_type, payload).map_err(|err| {
+                RpcServeError::FailedToDecodePayload {
+                    message: format!("Failed to decode payload {err:?}"),
+                }
+            })?,
+        ),
+        (None, Some(_)) => {
+            return Err(RpcServeError::PayloadTypeRequired {
+                message: "If the payload is specified the payload type must be as well".to_string(),
```
morph-dev, line 42:
```suggestion
                message: "If the 'payload' is specified the 'payloadType' must be as well".to_string(),
```

nit. Up to you.

testing/ethportal-peertest/src/scenarios/ping.rs
morph-dev, line 1: :heart: 

nit: Would be nice to have at least some tests when only payload_type is specified (we have all 3 other combination)

testing/ethportal-peertest/src/scenarios/ping.rs
```diff
@@ -0,0 +1,196 @@
...
+    let bootnode_enr = peertest.bootnode.enr.clone();
+    let bootnode_sequence = bootnode_enr.seq();
+    let result = match subnetwork {
+        Subnetwork::Beacon => BeaconNetworkApiClient::ping(target, bootnode_enr, None, None),
+        Subnetwork::History => HistoryNetworkApiClient::ping(target, bootnode_enr, None, None),
+        Subnetwork::State => StateNetworkApiClient::ping(target, bootnode_enr, None, None),
+        _ => panic!("Unexpected subnetwork: {subnetwork}"),
+    }
+    .await
+    .unwrap_or_else(|_| panic!("Ping failed {subnetwork}"));
```
morph-dev, line 36: Instead of using
```rust
.unwrap_or_else(|_| panic!(...));
```

use

```rust
.expect(...); // you don't have to worry about "err", it will be included automatically
```

Also everywhere below.

crates/ethportal-api/src/types/distance.rs
```diff
@@ -1,12 +1,27 @@
 use std::{fmt, ops::Deref};
 
 use alloy::primitives::U256;
+use serde::{Deserialize, Serialize};
 use ssz_derive::{Decode, Encode};
 
 pub type DataRadius = U256;
```
morph-dev, line 7: Just wondering if we should actually use this type instead of `Distance` (in ping/pong context). What do you think?

morph-dev (commented):
Lots of comments, but no major concerns.

crates/ethportal-api/src/types/ping_extensions/extensions/type_0.rs
```diff
@@ -49,10 +51,10 @@ impl ClientInfoRadiusCapabilities {
         }
     }
 
-    pub fn capabilities(&self) -> Result<Vec<Extensions>, ExtensionError> {
+    pub fn capabilities(&self) -> Result<Vec<PingExtensionType>, ExtensionError> {
```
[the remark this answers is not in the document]
KolbyML: Oh ok I think I need to add new value to the enum being `UnspecifiedType(u16)`, or else if someone sends us a type we don't support `capabilities()` would fail every time

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -370,17 +376,68 @@ impl<
         enr
     }
 
+    /// Generate a default payload for a respective payload type.
+    pub fn default_payload(
+        &self,
+        payload_type: PingExtensionType,
+    ) -> Result<CustomPayload, OverlayRequestError> {
+        Ok(match payload_type {
+            PingExtensionType::Capabilities => ClientInfoRadiusCapabilities::new(
```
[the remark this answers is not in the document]
KolbyML: The new() function generates the client info for trin so we wouldn't need to pass it in

crates/portalnet/src/overlay/protocol.rs
KolbyML: I took what you said in mind and this is what I came up with

```
        if let (Some(payload_type), Some(payload)) = (payload_type, &payload) {
            if payload_type != payload.ping_extension_type() {
                return Err(OverlayRequestError::InvalidRequest(
                    "Payload type mismatch".into(),
                ));
            }
        }

        let payload_type = payload_type.unwrap_or(PingExtensionType::Capabilities);
        let ping_extension = match payload {
            Some(payload) => payload,
            None => self.default_ping_extension(payload_type)?,
        };

        let request = Ping {
            enr_seq: self.discovery.local_enr().seq(),
            payload_type: payload_type.into(),
            payload: ping_extension.into(),
        };
```

It is a little different for the Ping struct you envisioned, but I think it accomplishes the same goal you envisioned of simplifying the code

crates/rpc/src/ping_extension.rs
KolbyML: Sure

KolbyML:
> > * Use ping extension type instead of u16 (manually implement (de)serialization)
> 
> The PR is ready for review. I don't think this suggestion ^ makes sense on the json-rpc level or protocol level due to how ping extensions work. If we get a ping with an invalid `ping extension type` we are supposed to send a ping error, but if do `Use ping extension type instead of u16 (manually implement (de)serialization)` we would fail and return a log before we get to the process ping code to properly handle the invalidly sent data.

Never mind I figured out how we can do it!

KolbyML:
@morph-dev The PR is ready for another look

morph-dev (commented):
Looks much better! Almost there

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -410,35 +414,24 @@ impl<
...
-
-        let (payload_type, payload) = match (payload_type, payload) {
-            (Some(payload_type), Some(payload)) => {
-                if payload_type != payload.payload_type() {
-                    return Err(OverlayRequestError::InvalidRequest(
-                        "Payload type mismatch".into(),
-                    ));
-                }
-                (payload_type, payload.into())
+        if let (Some(payload_type), Some(payload)) = (payload_type, &payload) {
```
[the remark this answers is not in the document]
KolbyML: I don't think this is good practice and there isn't value in just writing what the is doing especially when it is this simple.

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -410,35 +414,24 @@ impl<
...
+            Some(payload) => payload,
+            None => self.default_ping_extension(payload_type)?,
         };
 
         let request = Ping {
-            enr_seq,
-            payload_type: u16::from(payload_type),
-            payload,
+            enr_seq: self.discovery.local_enr().seq(),
+            payload_type,
```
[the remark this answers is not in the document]
KolbyML: The code in the suggestion above doesn't fix a bug.

```
        if ping_extension_type.is_none() && ping_extension.is_some() {
            return Err(OverlayRequestError::InvalidRequest(
                "`ping_extension_type`, must be specified if `ping_extension` is specified".into(),
            ));
        }
```

I added this validation step above

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -410,35 +414,24 @@ impl<
...
-                    self.ping_extensions.raw_extensions(),
-                )
-                .into(),
-            ),
+        }
+
+        let payload_type = payload_type.unwrap_or(PingExtensionType::Capabilities);
+        let ping_extension = match payload {
+            Some(payload) => payload,
+            None => self.default_ping_extension(payload_type)?,
```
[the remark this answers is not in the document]
KolbyML: I changed it to an `unwrap_or` so the ? would work, as we don't need a closure here.

crates/portalnet/src/overlay/service/ping/mod.rs
```diff
@@ -67,48 +67,47 @@ impl<
...
-                    "Received non-supported extension type in ping message: {err:?}",
-                );
-                return self.create_pong(
-                    PingExtensionType::Error.into(),
-                    PingError::new(ErrorCodes::ExtensionNotSupported).into(),
-                );
-            }
+        let extension_type = request.payload_type;
+
+        if let PingExtensionType::NonSupportedExtension(non_supported_extension) = extension_type {
```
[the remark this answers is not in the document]
KolbyML: > Can't we merge this with !self.ping_extensions.is_supported check below?

I would prefer to keep them separate cause they are distinct log cases, but sure

> Or just move it instead match block below (instead of unreachable!(...) at line 130)

This would be unreachable regardless

crates/rpc/src/errors.rs
```diff
@@ -125,3 +127,13 @@ pub enum WsHttpSamePortError {
         ws_modules: Vec<PortalRpcModule>,
     },
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, thiserror::Error)]
+pub enum PayloadTypeNotSupportedReason {
```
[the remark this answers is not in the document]
KolbyML: > Now that I see how this is used, it shouldn't implement Error (and there is no need to `thiserror:Error`).
> My mistake. Please remove it.

I am leaving it the way it is now, it needs to implement `Display` and `thiserror:Error` implements `Display`

KolbyML:
> Instead of using
> 
> ```rust
> .unwrap_or_else(|_| panic!(...));
> ```
> 
> use
> 
> ```rust
> .expect(...); // you don't have to worry about "err", it will be included automatically
> ```
> 
> Also everywhere below.

Now fmt is asking to do it the way I initially did it before `.unwrap_or_else(|err| panic!("Ping should succeed. Subnetwork: {subnetwork}, {err:?}"));`

Anyways I am not overly concerned with how these errors look, but it is tedious to need to switch it back to how I had it originally

KolbyML:

crates/portalnet/src/overlay/protocol.rs
```diff
@@ -408,29 +408,38 @@ impl<
...
         &self,
         enr: Enr,
-        payload_type: Option<PingExtensionType>,
-        payload: Option<PingExtension>,
+        ping_extension_type: Option<PingExtensionType>,
+        ping_extension: Option<PingExtension>,
     ) -> Result<Pong, OverlayRequestError> {
-        if let (Some(payload_type), Some(payload)) = (payload_type, &payload) {
-            if payload_type != payload.ping_extension_type() {
+        if ping_extension_type.is_none() && ping_extension.is_some() {
```
morph-dev, line 421: Now that we have 2 failure conditions, and then another long line with multiple `.unwrap_or`, I find it not so easy to read. I think something what you had initially works much better, but just returning ping extension:

```rust
let ping_extension = match (ping_extension_type, ping_extension) {
    (None, None) => self.default_ping_extension(PingExtensionType::default()),
    (Some(ping_extension_type), None) => self.default_ping_extension(ping_extension_type),
    (Some(ping_extension_type), Some(ping_extension)) => {
        if ping_extension_type != ping_extension.ping_extension_type() {
            return Err(..);
        }
        ping_extension
    }
    (None, Some(ping_extension)) => return Err(..),
};

let request = Ping {
    enr_seq: self.discovery.local_enr().seq(),
    payload_type: ping_extension.ping_extension_type(),
    payload: ping_extension.into(),
};
```

crates/portalnet/src/overlay/protocol.rs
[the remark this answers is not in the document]
morph-dev: I suggested writing it, specifically because I don't think it was simple and obvious for someone reading the code the first time. But with new suggestion, and function documentation, it's clearer.

crates/rpc/src/errors.rs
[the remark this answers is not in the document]
morph-dev: No, this is not ok. The `thiserror::Error` has a use case, and this isn't it.
If you want to implement `Display`, just do that (it's simple enough). If you don't want to do it manually, use `strum::Display`. But don't use `thiserror::Error` to implement `Display`.

morph-dev (commented):
One thing crossed my mind. I know it's according to spec, but does it make sense for all clients to include Capabilities and Error in their supported extension types? Both of these are standard types and all clients should support them on all networks.

Reasoning:
- responding to someone's Capabilities Ping and saying the I support Capabilities extension type is meaningless - they can already figure that out :).
- what does it mean to say that you support Error type? You don't support Pings that are Error types, and I will always reply with Error Pong (when applicable), even if you don't say you support it

I'm fine having this discussion on spec repo and modifying the spec to say that clients don't have to include these in their responses.

crates/rpc/src/errors.rs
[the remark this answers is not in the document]
KolbyML: This is annoying I have to revert to what I had originally

crates/portalnet/src/overlay/protocol.rs
KolbyML: I will do this, but it is a little annoying to go back to this, after switching it 2 times already

morph-dev: My initial [suggestion](https://github.com/ethereum/trin/pull/1710#discussion_r2006451280) wasn't to not have a `match`, but that match should only return `PingExtension` object (as `PingExtensionType` can be derived from it). Pretty much what I suggested again.

In the middle, you diverged from it, and it wasn't working. But it was simple enough and I didn't mind. But then you made it more complicated by adding additional error handling. So I reverted pretty much to my initial suggestion.

KolbyML: I combined these 2 lines, as it is a little fatiguing reading both

```
    (None, None) => self.default_ping_extension(PingExtensionType::default()),
    (Some(ping_extension_type), None) => self.default_ping_extension(ping_extension_type),
```

KolbyML: > My initial [suggestion](https://github.com/ethereum/trin/pull/1710#discussion_r2006451280) wasn't to not have a `match`, but that match should only return `PingExtension` object (as `PingExtensionType` can be derived from it). Pretty much what I suggested again.
> 
> In the middle, you diverged from it, and it wasn't working. But it was simple enough and I didn't mind. But then you made it more complicated by adding additional error handling. So I reverted pretty much to my initial suggestion.

oh my bad

KolbyML:
> One thing crossed my mind. I know it's according to spec, but does it make sense for all clients to include Capabilities and Error in their supported extension types? Both of these are standard types and all clients should support them on all networks.
> 
> Reasoning:
> 
> * responding to someone's Capabilities Ping and saying the I support Capabilities extension type is meaningless - they can already figure that out :).
> * what does it mean to say that you support Error type? You don't support Pings that are Error types, and I will always reply with Error Pong (when applicable), even if you don't say you support it
> 
> I'm fine having this discussion on spec repo and modifying the spec to say that clients don't have to include these in their responses.

You can have 400 different extensions active at any given time, so I don't think it makes a difference if we include the 2 standard extensions in the capabilities list. I agree logistically it makes a little more sense, code wise this is a little simpler as we can just do a `.contains(x)`

KolbyML:
> One thing crossed my mind. I know it's according to spec, but does it make sense for all clients to include Capabilities and Error in their supported extension types? Both of these are standard types and all clients should support them on all networks.
> 
> Reasoning:
> 
> * responding to someone's Capabilities Ping and saying the I support Capabilities extension type is meaningless - they can already figure that out :).
> * what does it mean to say that you support Error type? You don't support Pings that are Error types, and I will always reply with Error Pong (when applicable), even if you don't say you support it
> 
> I'm fine having this discussion on spec repo and modifying the spec to say that clients don't have to include these in their responses.

This is a bit out of scope for this PR as well, so I don't think this should block this PR

crates/rpc/src/errors.rs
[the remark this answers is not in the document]
morph-dev: I already acknowledged that I made a mistake for asking/suggesting the change. But it was during my first pass of the PR, without having context on how it's going to be used.

It didn't help that PR is 1.5K lines long, it was located in `errors.rs` together with other error types, that it was defined in portalnet crate but used in rpc.

Sorry for being annoying and sometimes too nitpicky. I believe that it's important to be very detailed with reviews as it matters for the health of the codebase in the long run. I think you would agree that we did improve this pr significantly.

morph-dev (approved):
Thank you for bearing with me! :rocket:

crates/rpc/src/errors.rs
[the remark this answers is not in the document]
KolbyML: >  think you would agree that we did improve this pr significantly.

I agree and I appropriate you reviewing my  PR's. Thank you very much! 1.5k :triumph: a bit of that is testing also a lot of stuff was updated during review which was a little out of scope for the PR, but I do agree a lot of this discussion has lead to improvements.

I was just a little sleepy when I woke up at like 6am!

KolbyML:
> Thank you for bearing with me! 🚀

Thank you for bearing with me and thank you for reviewing my PR's :fire:. I appropriate your feedback and time.

morph-dev:
> This is a bit out of scope for this PR as well, so I don't think this should block this PR

Yes, totally agree. Just wanted to raise the topic.

> You can have 400 different extensions active at any given time, so I don't think it makes a difference if we include the 2 standard extensions in the capabilities list. I agree logistically it makes a little more sense, code wise this is a little simpler as we can just do a `.contains(x)`

I'm not sure when you would check for it. I think we always assume that Capabilities are supported , and we never check Error type. We only check for the _Base_ extension types, which is valid case.

And 400 extensions are possible, but I would be surprised if in practice we have more than 5 (non-standard) on the same network. And it's not about saving bandwith as 4 bytes wouldn't make much difference. It's more about making the spec clean.

PR #1711, opened 2025-02-27
Unify reth & alloy dependencies

carver:
### What was wrong?

There are a couple distinct versions of reth/revm libraries that we depend on. Would be better to sync them all up.

### How was it fixed?

Just have to hunt down everywhere in the tree that we depend on different versions

Also:
- update rust to support newer dependencies (only needed 1.82, but figured we would just go to the latest)
- update the engine API to drop some old (now unused) types that are now unavailable

TODO:
- [x] address new clippy issues, after rust ugrade


KolbyML (approved):
:shipit: looks good

PR #1714, opened 2025-02-27
docs: release ethportal using verification

carver:
Frankly, I don't know what changed since --no-verify was required. It works without the flag now, so we should definitely not use it. Verification seems good.


morph-dev (approved):
I noticed that you bumped cargo version both in workspace Cargo.toml and in ethportal-api. Should we document this? Should we say that it's ok to merge that pr without approval from others (assuming all tests are passing)?

ogenev (approved):

carver:
Yeah, I'll add a line about updating the workspace Cargo.

I think it would probably generally be good to get a double-check that the new version number is right, so I might not codify that it's expected to skip review. But I would think of skipping a review like jaywalking on a busy street. We don't officially bless it, and expect you to recognize that you messed up if you get run over, but it will probably happen sometimes without incident and it's not a felony offense.

It is deeply annoying to me when projects screw up semantic versioning and cause downstream issues because they didn't correctly bump a major version. So I'd like our project to not inflict that pain on others.

PR #1719, opened 2025-03-10
chore: update docker compose environment variable syntax

0xpsyduck:
the environment variable section is now be declared according to array syntax described in [docker docs](https://docs.docker.com/reference/compose-file/services/#environment).

the orginal declaration
```yaml
environment:
    - RUST_LOG: info
```
is giving this syntax error
```bash
➜  docker compose up
services.trin.environment.[0]: unexpected type map[string]interface {}
```


KolbyML (approved):
:shipit: looks good

PR #1720, opened 2025-03-13
feat: implement Accept codes for detailed offer declines for protocol versioning v1

KolbyML:
### What was wrong?
implements spec change https://github.com/ethereum/portal-network-specs/pull/370

> A offer can be declined for various reasons, using a generic decline message is very cryptic which makes it immensely hard to understand why we can't find nodes to accept the content, without controlling a majority of the network and logging every single implementation which doesn't make sense due to the complexity.

### How was it fixed?

Implementing the spec change


crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
...
+#[serde(transparent)]
+pub struct ByteList64(pub VariableList<u8, typenum::U64>);
+
+impl ByteList64 {
+    pub fn with_capacity(num_bytes: usize) -> Result<ByteList64, ssz_types::Error> {
+        if num_bytes <= 64 {
+            Ok(Self(VariableList::from(vec![0; num_bytes])))
+        } else {
+            Err(ssz_types::Error::OutOfBounds { i: 64, len: 64 })
+        }
```
morph-dev, line 39: I believe you get exactly this by using `VariableList::new(vec![0; num_bytes])`.

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
 pub type ByteList2048 = VariableList<u8, typenum::U2048>;
 pub type ByteList32K = VariableList<u8, typenum::U32768>;
 pub type ByteList1G = VariableList<u8, typenum::U1073741824>;
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(transparent)]
+pub struct ByteList64(pub VariableList<u8, typenum::U64>);
```
morph-dev, line 31: Why do we have to wrap VariableList?

Since we are wrapping it, and `set` and `get` functions use `AcceptCode`, should we call it, `AcceptCodeList` instead, and maybe wrap `VariableList<AcceptCode>`?

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
 pub type ByteList2048 = VariableList<u8, typenum::U2048>;
 pub type ByteList32K = VariableList<u8, typenum::U32768>;
 pub type ByteList1G = VariableList<u8, typenum::U1073741824>;
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
```
morph-dev, line 29: What would be the desired serialize json value of this object. Let's assume there are two AcceptCodes: accepted and declined (generic). Which one should it be:
```json
{
    "connection_id": 1234,
    "content_keys": [
        0,
        1
    ]
}
```
or
```json
{
    "connection_id": 1234,
    "content_keys": "0x0001"
}
```
or something else?

I feel like we serialize our types the "easy" way and this is frequently not desirable for `VariableList<u8, _>` types, so I just want to raise the question in this case.

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
...
+        let len = self.len();
+
+        if let Some(byte) = self.0.get(i) {
+            Ok(AcceptCode::from(*byte))
+        } else {
+            Err(ssz_types::Error::OutOfBounds { i, len })
+        }
+    }
+
+    pub fn is_not_one(&self) -> bool {
```
morph-dev, line 63: I find `is_not_one` name ambiguous. How about `is_at_least_one_accepted` or `all_declined` or if you have better alternative. Also, add brief doc string.

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
...
+        self.0.iter().all(|byte| *byte != 1)
+    }
+
+    pub fn into_bytes(self) -> Vec<u8> {
```
morph-dev, line 67: What is the use case for this? I'm worried it might be used instead of `as_ssz_bytes()`.

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
...
+
+impl Deref for ByteList64 {
+    type Target = VariableList<u8, typenum::U64>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Encode for ByteList64 {
```
morph-dev, line 80: This is not needed. You can add `Encode` into `#[derive(..)]`, together with:
```
#[ssz(struct_behaviour = "transparent")]
```

Same for `Decode`.

crates/ethportal-api/src/types/bytes.rs
```diff
@@ -21,3 +25,130 @@ pub type ByteList1100 = VariableList<u8, U1100>;
...
+    /// The content was rejected because the content is not within the node's data radius
+    NotWithinRadius,
+    /// Unspecified accept code, this should not be used
+    Unspecified,
+}
+
+impl From<AcceptCode> for u8 {
+    fn from(code: AcceptCode) -> u8 {
+        match code {
+            AcceptCode::Declined => 0,
```
morph-dev, line 131: This should be updated (reflecting recent changes in the PR)

morph-dev (commented):
Looks mostly ok, but I have several suggestions.

What is the rollout plan? We need coordination between teams, right?!

KolbyML:
> Looks mostly ok, but I have several suggestions.
> 
> What is the rollout plan? We need coordination between teams, right?!

I am not sure yet, the spec PR needs consensus first before that is discussed, we could potentially use the new system piper made an issue for, I am not that picky but will look more into it once the PR is more finalized

crates/ethportal-api/src/types/bytes.rs
KolbyML: It was used in a few tests, but I switched them to use `as_ssz_bytes()` now

crates/ethportal-api/src/types/bytes.rs
KolbyML: Sure I wanted to avoid the work of implementing Serialize/Encode for AcceptCode, but it wasn't the worst

crates/ethportal-api/src/types/bytes.rs
KolbyML: I see I wrote a custom serializer to fix this issue

KolbyML:

crates/ethportal-api/src/types/accept_code.rs
```diff
@@ -0,0 +1,152 @@
...
+
+use serde::{Deserialize, Serialize};
+use ssz::{Decode, Encode};
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum, VariableList};
+
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq, Encode, Decode)]
+#[serde(transparent)]
+#[ssz(struct_behaviour = "transparent")]
+pub struct AcceptCodeList(pub VariableList<AcceptCode, typenum::U64>);
```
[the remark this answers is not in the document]
KolbyML: I want to keep this struct and remove the auto derives of `Encode, Decode` as they can now be error prone. So I will implement functions `encode` and `decode` 
- `encode(&self, protocol_version: ProtocolVersion)`
- `decode(protocol_version: ProtocolVersion, raw_content_keys: Bytes)`

crates/ethportal-api/src/types/accept_code_hex.rs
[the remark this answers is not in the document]
KolbyML: :+1: I named the mod accept_code_hex to not conflict with the serde namespace

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -618,7 +619,61 @@ impl From<PopulatedOfferWithResult> for Offer {
...
+            Bytes::from(v0_content_keys.as_ssz_bytes())
+        };
+
+        Ok(Self {
+            connection_id,
+            content_keys,
+        })
+    }
+
+    pub fn decode_accept_code_list(
```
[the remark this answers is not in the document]
KolbyML: I ended up moving this function to `AcceptCodeList`, as I don't make it makes sense in the Accept struct, due to people being able to use AcceptCodeList wrong, so this concern is resolved now.

crates/ethportal-api/src/types/accept_code.rs
```diff
@@ -0,0 +1,152 @@
...
+    pub fn get(&self, index: usize) -> Result<AcceptCode, ssz_types::Error> {
+        if let Some(accept_code) = self.0.get(index) {
+            Ok(*accept_code)
+        } else {
+            Err(ssz_types::Error::OutOfBounds {
+                i: index,
+                len: self.len(),
+            })
+        }
+    }
```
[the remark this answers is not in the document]
KolbyML: > nit: you could avoid having these if you would implement: [`Index`](https://doc.rust-lang.org/std/ops/trait.Index.html) and [`IndexMut`](https://doc.rust-lang.org/std/ops/trait.IndexMut.html) traits. You would also be able to access/modify elements with `access_code_list[i]` syntax.

I removed `get()` as since we implement deref we already can do `access_code_list[i]` syntax

> The only difference would be that using get and set would panic if index is out of bounds, which I guess is ok in most cases.
> 
> Up to you.
> 

I ended up keeping our set implementation, but using the panic semantics of  `Vec::insert`

> Unrelated to previous comment, but `ssz_types::Error` sounds like the wrong error type. On the other hand, creating new error type might be too much. If you don't want to implement Index and IndexMut, why don't we just expose `get` and `get_mut` the same way that the underlying object.

We are using `VariableList`'s get and we are implementing our own `set` functions following the same semantics as `Vec::insert` now

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -618,7 +619,61 @@ impl From<PopulatedOfferWithResult> for Offer {
...
-    pub content_keys: BitList<typenum::U64>,
+    pub content_keys: Bytes,
+}
+
+impl Accept {
+    pub fn new(
+        protocol_version: ProtocolVersion,
+        connection_id: u16,
+        content_keys: AcceptCodeList,
+    ) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
KolbyML: > * `fn new(..)` - we can just call `.expect(..)` as indices should always be correct
done
> * decode function can use `ssz::DecodeError`
The decode function now uses `AcceptCodeListError`, but the decode function is now also not apart of this struct, but `AcceptCodeList`

KolbyML:

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -888,29 +854,29 @@ mod test {
...
         {
-            let content_keys = AcceptCodeList::from_ssz_bytes(&content_keys).unwrap();
+            let content_keys = AcceptCodeList::decode(protocol_version, content_keys).unwrap();
             assert_eq!(decoded_connection_id, connection_id);
             assert_eq!(content_keys.len(), 64);
-            assert_eq!(content_keys.get(0).unwrap(), AcceptCode::Accepted);
-            assert_eq!(content_keys.get(1).unwrap(), AcceptCode::Declined);
-            assert_eq!(content_keys.get(62).unwrap(), AcceptCode::Declined);
-            assert_eq!(content_keys.get(63).unwrap(), AcceptCode::Accepted);
+            assert_eq!(content_keys.first().unwrap(), &AcceptCode::Accepted);
```
morph-dev, line 860: nit: while this is technically correct, I would still use `.get(0)` for consistency with the rest of the code.

morph-dev (approved):
:shipit:

crates/ethportal-api/src/types/portal_wire.rs
KolbyML: clippy or fmt fails if you do that, so I will switch all the tests to parse with [] then, if you want consistency

PR #1724, opened 2025-03-19
refactor: rename history_new and header_with_proof_new by removing _new

morph-dev:
### What was wrong?

We have to do some cleanup after header migration (see #1666 for details) 

### How was it fixed?

Renamed `history_new->history` and `header_with_proof_new->header_with_proof`.

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1725, opened 2025-03-19
refactor: read e2store header type as big endian

KolbyML:
### What was wrong?
I am creating a repo to formalize all of the e2store formats and to avoid duplicate types with @arnetheduck and @advaita-saha https://github.com/eth-clients/e2store-format-specs

Anyways currently we are reading the type as little endian which doesn't make sense.

For example the type for the `VersionEntry` is a vector [65, 32] which in ascii is `e2`, but we are reading it backwards as `2e`.

Our code is ends up working to the same result, but it is fairly unidle as we also have 0x3265 in our documentation when it should be 0x6532 or [65, 32]

### How was it fixed?

use be and update numbers.

### DIFF
--- a/crates/e2store/src/e2ss.rs
+++ b/crates/e2store/src/e2ss.rs
@@ -44,6 +44,7 @@ use crate::{
         stream::{E2StoreStreamReader, E2StoreStreamWriter},
         types::{Entry, VersionEntry},
     },
+    entry_types,
     types::HeaderEntry,
     utils::underlying_io_error_kind,
 };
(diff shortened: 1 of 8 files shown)


morph-dev (approved):
I think that all these constants should be declared as constants, either at the beginning of each file, or all of them in the same place.

I prefer all of them in separate files, with clear documentation. For example, we could put them in `crates/e2store/src/entry_types.rs`, and have something like this for each of them:

```rust
/// Snappy compressed rlp execution header
///
/// data: snappyFramed(rlp(header))
pub const COMPRESSED_HEADER: u16 = 0x0300;
```

That way, we can later access them with:

```rust
use e2store::entry_types;

Ok(Entry::new(entry_types::COMPRESSED_HEADER, encoded))
```

PR #1726, opened 2025-03-20
chore: update e2store so it can be published

KolbyML:
### What was wrong?

I wanted to publish the crate to secure the name e2store on crates.io https://crates.io/crates/e2store

@morph-dev do you want to keep the e2ss stats util you made? If we want to keep it I will move it to the bin folder up to you I am fine with keeping it. As we can't depend on `trin-utils`

### How was it fixed?

update cargo.toml to use the proper readme
remove trin-utils usage so it can be published.


morph-dev:
Regarding `e2ss-stats.rs`, do you think that binary shouldn't be part of the crate or is dependency to `trin-utils` the only issue?
If dependency is the only issue, I think we can remove it. Currently we just use it to setup logging, and we can do that directly in the binary.
But if you think binary shouldn't be there, then please move it to bin folder (I think it can be useful).

We should consider making similar binaries for other formats. They can be used to quickly check that files are well formatted, and to collect some basic stats about them.

njgheorghita (approved):

morph-dev:
> it doesn't really matter much to me, if that is the only thing `logs` I will look into it

I think we should preserve `e2ss-stats` binary, and removing `trin-utils` as dependency should be trivial.

KolbyML:
> > it doesn't really matter much to me, if that is the only thing `logs` I will look into it
> 
> I think we should preserve `e2ss-stats` binary, and removing `trin-utils` as dependency should be trivial.

That is why I haven't merged the PR yet from Nick's review. I am going to try removing `trin-utils` and then try and publish `e2store` with `e2ss-stats` and if it works I will update the PR to retain the bin.

KolbyML:

morph-dev (approved):
:ship:

PR #1727, opened 2025-03-21
feat: add e2hs-writer binary and file format

njgheorghita:
### What was wrong?
Support for new `E2HS` file format, and tooling to help generate these files. 

### How was it fixed?
- added file format implementation
- wrote a cli tool to help gather the relevant data to create `e2hs` files. 
- proof generation for post-merge `HeaderWithProof`

### To-Do
- Add support for post-dencun eras
- Add capability for this to be used in live tooling architecture to generate "latest" e2hs files
- Improve performance for post-merge eras (binary search is time-consuming, but I'm not sure if there's a better solution here... though at the head of the chain this isn't an issue)


crates/ethportal-api/src/types/execution/header.rs
```diff
@@ -309,6 +313,63 @@ impl PartialEq for Header {
     }
 }
 
+impl From<ExecutionPayloadBellatrix> for Header {
```
njgheorghita, line 316: This logic is somewhat duplicated from https://github.com/ethereum/trin/blob/master/bin/trin-execution/src/era/beacon.rs#L39 but I wasn't able to find a great way to use integrate the two use-cases together

crates/ethportal-api/src/assets/test/proofs/17062257/state.ssz
njgheorghita, line 1: These test assets are somewhat large... but I do think what they're testing is fairly critical

crates/ethportal-api/src/types/execution/header.rs
KolbyML: I think this level of code duplication is fine, but I do think we should get this PR in first https://github.com/ethereum/trin/pull/1721

KolbyML:
> * Improve performance for post-merge eras (binary search is time-consuming, but I'm not sure if there's a better solution here... though at the head of the chain this isn't an issue)

Instead of binary searching each time you can just do it once then use era manager which can collect blocks from era or era1 files https://github.com/ethereum/trin/blob/master/bin/trin-execution/src/era/manager.rs

crates/ethportal-api/src/assets/test/proofs/17062257/state.ssz
morph-dev: I don't like that so big files are part of trin repo. Can we move them to different repo (and use them using submodules)?

If not [portal-spec-tests](https://github.com/ethereum/portal-spec-tests), then maybe something similar in the context of [e2store-format-specs](https://github.com/eth-clients/e2store-format-specs)?

KolbyML: I think we should stop using submodules if we want to test big files we should do what we did for the consensus-spec-tests https://github.com/ethereum/trin/tree/master/testing/ef-tests

morph-dev: If I'm not mistaken, `ef-tests` will always run against the head of the `ethereum/consensus-spec-tests` repo.
While that is ok in some cases, sometimes you want more control over which version of the spec you run your tests against (e.g. when spec changes more frequently, you don't want tests to start failing out of nowhere). So for `portal-spec-tests`, I would still prefer to use submodules.

In this particular case, I don't have strong opinion. Other than that they should probably live in different repo.

KolbyML: > If I'm not mistaken, ef-tests will always run against the head of the ethereum/consensus-spec-tests repo.

This is a choice like any other, and it doesn't run against the head it is currently set to use the latest release zip

KolbyML:
> @KolbyML @morph-dev Apologies for the massive pr... A lot of it is boilerplate `e2store` stuff, but I don't mind splitting this up into smaller pieces if it makes life easier, but I had to tackle it in this initial lump to generate the new `e2hs` files initially.

I am fine reviewing this in one PR

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,381 @@
...
+        })?;
+        let header_with_proof = HeaderWithProof {
+            header,
+            proof: BlockHeaderProof::HistoricalHashes(proof),
+        };
+        Ok(AllBlockData {
+            block: block_number,
+            header_with_proof,
+            body,
+            receipts,
```
KolbyML, line 118:
```suggestion
        let epoch_acc = self.epoch_accumulator.clone().unwrap_or_else(|| {
            panic!("Epoch accumulator not found for pre-merge block: {block_number}")
        });
        let proof = PreMergeAccumulator::construct_proof(&header, &epoch_acc)?;
        let proof = BlockProofHistoricalHashesAccumulator::new(proof.into()).map_err(|e| {
            anyhow!("Unable to convert proof to BlockProofHistoricalHashesAccumulator: {e:?}")
        })?;
        let header_with_proof = HeaderWithProof {
            header,
            proof: BlockHeaderProof::HistoricalHashes(proof),
        };
        Ok(AllBlockData {
            block: block_number,
            header_with_proof,
            body: tuple.body.body,
            receipts: tuple.receipts.receipts,
```

These variables aren't used, so I think we shouldn't define them to simplify the code

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,381 @@
...
+        let payload = block.body.execution_payload.clone();
+        let header = pre_capella_execution_payload_to_header(payload.clone());
+        let historical_batch = HistoricalBatch {
+            state_roots: era.era_state.state.state_roots().clone(),
+            block_roots: era.era_state.state.block_roots().clone(),
+        };
+        let slot = block.slot;
+        // beacon block proof
+        let hb_proof = historical_batch.build_block_root_proof(slot % EPOCH_SIZE);
+        let beacon_block_proof: FixedVector<B256, typenum::U14> = hb_proof.into();
```
KolbyML, line 144:
```suggestion

        // beacon block proof
        let beacon_block_proof = historical_batch.build_block_root_proof(slot % EPOCH_SIZE);
        let beacon_block_proof: FixedVector<B256, typenum::U14> = beacon_block_proof.into();
        
```

We can add some strategic spaces for readability and code sectioning and I don't know what hb means, so if we could expand on that it would be cool

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,381 @@
...
+            beacon_block_root: block.tree_hash_root(),
+            slot,
+            execution_block_proof,
+        };
+        let header_with_proof = HeaderWithProof {
+            header,
+            proof: BlockHeaderProof::HistoricalRoots(proof),
+        };
+        let body = BlockBody(AlloyBlockBody {
+            transactions: decode_transactions(&payload.transactions)?,
```
KolbyML, line 163:
```suggestion
        let transactions = decode_transactions(&payload.transactions)?;
        let header = pre_capella_execution_payload_to_header(payload.clone(), &transactions);
        let historical_batch = HistoricalBatch {
            state_roots: era.era_state.state.state_roots().clone(),
            block_roots: era.era_state.state.block_roots().clone(),
        };
        let slot = block.slot;
        // beacon block proof
        let hb_proof = historical_batch.build_block_root_proof(slot % EPOCH_SIZE);
        let beacon_block_proof: FixedVector<B256, typenum::U14> = hb_proof.into();
        // execution block proof
        let mut execution_block_hash_proof = block.body.build_execution_block_hash_proof();
        let body_root_proof = block.build_body_root_proof();
        execution_block_hash_proof.extend(body_root_proof);
        let execution_block_proof: FixedVector<B256, typenum::U11> =
            execution_block_hash_proof.into();

        let proof = BlockProofHistoricalRoots {
            beacon_block_proof,
            beacon_block_root: block.tree_hash_root(),
            slot,
            execution_block_proof,
        };
        let header_with_proof = HeaderWithProof {
            header,
            proof: BlockHeaderProof::HistoricalRoots(proof),
        };
        let body = BlockBody(AlloyBlockBody {
            transactions,
```


`pre_capella_execution_payload_to_header` calls `decode_transactions()` and it is called below, maybe we can decode the transactions once instead of twice?

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,381 @@
...
+        let header_with_proof = HeaderWithProof {
+            header,
+            proof: BlockHeaderProof::HistoricalSummaries(proof),
+        };
+        let withdrawals: Vec<Withdrawal> =
+            payload.withdrawals.iter().map(Withdrawal::from).collect();
+        let body = BlockBody(AlloyBlockBody {
+            transactions: decode_transactions(&payload.transactions)?,
+            ommers: vec![],
+            withdrawals: Some(Withdrawals::new(withdrawals)),
```
KolbyML, line 229:
```suggestion
        let transactions = decode_transactions(&payload.transactions)?;
        let withdrawals: Vec<Withdrawal> =
            payload.withdrawals.iter().map(Withdrawal::from).collect();
        let header = pre_deneb_execution_payload_to_header(payload.clone(), &transactions, &withdrawals);

        let historical_batch = HistoricalBatch {
            state_roots: era.era_state.state.state_roots().clone(),
            block_roots: era.era_state.state.block_roots().clone(),
        };
        let slot = block.slot;
        
        let beacon_block_proof = historical_batch.build_block_root_proof(slot % EPOCH_SIZE);
        let beacon_block_proof: FixedVector<B256, typenum::U13> = beacon_block_proof.into();
        
        // execution block proof
        let mut execution_block_hash_proof = block.body.build_execution_block_hash_proof();
        let body_root_proof = block.build_body_root_proof();
        execution_block_hash_proof.extend(body_root_proof);
        let execution_block_proof: VariableList<B256, typenum::U12> =
            execution_block_hash_proof.into();
        let proof = BlockProofHistoricalSummaries {
            beacon_block_proof,
            beacon_block_root: block.tree_hash_root(),
            slot,
            execution_block_proof,
        };

        let header_with_proof = HeaderWithProof {
            header,
            proof: BlockHeaderProof::HistoricalSummaries(proof),
        };
        let body = BlockBody(AlloyBlockBody {
            transactions,
            ommers: vec![],
            withdrawals: Some(Withdrawals::new(withdrawals)),
```

We can do some of the improvements here as well

KolbyML (commented):
Left some initial feedback on improvements and a few bugs I seen, as currently this PR will generate invalid headers.

I think some of the functions could use some strategic spaces to section related parts of the code to make it a little easier to read.

After this feedback is addressed I will give it a bit of a deeper look.

njgheorghita:
Thanks for the review @KolbyML - ready for another round of :eyes: when you are

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+            false => None,
+        };
+        Ok(Self {
+            starting_block,
+            ending_block,
+            epoch_accumulator,
+            execution_api,
+            era_provider,
+        })
+    }
```
KolbyML, line 94:
```suggestion
impl EpochReader {
    pub async fn new(
        epoch_index: u64,
        epoch_acc_path: PathBuf,
        el_provider_url: Url,
    ) -> anyhow::Result<Self> {
        let execution_api = ExecutionApi::new(el_provider_url.clone(), el_provider_url, 10).await?;
        let latest_block = execution_api.get_latest_block_number().await?;
        let maximum_epoch = latest_block / EPOCH_SIZE;
        ensure!(
            epoch_index <= maximum_epoch,
            "Epoch {epoch_index} is greater than the maximum epoch {maximum_epoch}"
        );
        
        let starting_block = epoch * EPOCH_SIZE;
        let epoch_accumulator = match starting_block < MERGE_BLOCK_NUMBER {
            true => {
                let header_validator = HeaderValidator::new();
                Some(Arc::new(
                    lookup_epoch_acc(
                        epoch_index,
                        &header_validator.pre_merge_acc,
                        &epoch_acc_path,
                    )
                    .await?,
                ))
            }
            false => None,
        };
        
        Ok(Self {
            starting_block,
            ending_block: starting_block + EPOCH_SIZE,
            epoch_accumulator,
            execution_api,
            era_provider: EraProvider::new(epoch).await?,
        })
    }
```

The way the code is layed out makes it hard for me to read, I think the adjustments I made above makes the code a lot easier to follow ^

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    fn get_pre_merge_block_data(&self, block_number: u64) -> anyhow::Result<AllBlockData> {
+        let raw_era1 = self.era_provider.get_era1_for_block(block_number)?;
+        let block_index = block_number % EPOCH_SIZE;
+        let tuple = Era1::get_tuple_by_index(&raw_era1, block_index);
+        let header = tuple.header.header;
+        let epoch_acc = self.epoch_accumulator.clone().ok_or_else(|| {
+            anyhow!("Epoch accumulator not found for pre-merge block: {block_number}")
+        })?;
+        let proof = PreMergeAccumulator::construct_proof(&header, &epoch_acc)?;
```
KolbyML, line 104:
```suggestion
        let Some(epoch_acc) = &self.epoch_accumulator else {
            bail!("Epoch accumulator not found for pre-merge block: {block_number}");
        };
        let proof = PreMergeAccumulator::construct_proof(&header, epoch_acc)?;
```

I think this is a little simpler/cleaner

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+            } else if current_block < CANCUN_BLOCK_NUMBER {
+                yield self.get_pre_deneb_block_data(current_block).await;
+            } else {
+                yield Err(anyhow!("Invalid block number: {current_block}"));
+            }
+            current_block += 1;
+        }
+        }
+    }
+}
```
KolbyML, line 267:
```suggestion
    pub fn iter_blocks(self) -> impl Stream<Item = anyhow::Result<AllBlockData>> {
        stream! {
            for current_block in (self.starting_block..self.ending_block) {
                if current_block < MERGE_BLOCK_NUMBER {
                    yield self.get_pre_merge_block_data(current_block);
                } else if current_block < SHANGHAI_BLOCK_NUMBER {
                    yield self.get_pre_capella_block_data(current_block).await;
                } else if current_block < CANCUN_BLOCK_NUMBER {
                    yield self.get_pre_deneb_block_data(current_block).await;
                } else {
                    yield Err(anyhow!("Invalid block number: {current_block}"));
                }
            }
        }
    }
}
```

It looks like indentation is missing here, also I think if we use an iterator the code would be a little simpler, then a while loop and maintaining a mut variable

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+                test_vector["execution_block_proof"].clone(),
+            )
+            .unwrap(),
+            slot: serde_yaml::from_value(test_vector["slot"].clone()).unwrap(),
+        };
+
+        let test_assets_dir =
+            format!("../../portal-spec-tests/tests/mainnet/history/headers_with_proof/beacon_data/{block_number}");
+        let hb_path = format!("{test_assets_dir}/historical_batch.ssz");
+        let hb_raw = std::fs::read(hb_path).unwrap();
```
KolbyML, line 318: I am not sure what hb means

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+
+    #[rstest::rstest]
+    #[case(15539558, 4702208, 575, "block_proofs_bellatrix/beacon_block_proof-15539558-cdf9ed89b0c43cda17398dc4da9cfc505e5ccd19f7c39e3b43474180f1051e01.yaml")]
+    #[case(15547621, 4710400, 576, "block_proofs_bellatrix/beacon_block_proof-15547621-96a9313cd506e32893d46c82358569ad242bb32786bd5487833e0f77767aec2a.yaml")]
+    #[case(15555729, 4718592, 577, "block_proofs_bellatrix/beacon_block_proof-15555729-c6fd396d54f61c6d0f1dd3653f81267b0378e9a0d638a229b24586d8fd0bc499.yaml")]
+    #[tokio::test]
+    async fn test_pre_capella_proof_generation(
+        #[case] block_number: u64,
+        #[case] slot: u64,
+        #[case] _epoch: u64,
```
KolbyML, line 295:
```suggestion
```

if `_epoch` isn't used do we need it?

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,384 @@
...
+        17062257,
+        6238210,
+        762,
+        "block_proofs_capella/beacon_block_proof-17062257.yaml"
+    )]
+    #[tokio::test]
+    async fn test_pre_deneb_proof_generation(
+        #[case] block_number: u64,
+        #[case] slot: u64,
+        #[case] _epoch: u64,
```
KolbyML, line 351:
```suggestion
```

if `_epoch` isn't used do we need it?

crates/e2store/src/e2hs.rs
```diff
@@ -0,0 +1,429 @@
...
+        let count = u64::from_le_bytes(
+            entry.value[(block_tuple_count * 8 + 8)..(block_tuple_count * 8 + 16)].try_into()?,
+        );
+        Ok(Self {
+            starting_number,
+            indices,
+            count,
+        })
+    }
+}
```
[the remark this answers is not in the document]
KolbyML: Ok maybe the structs should have a different name or something to prevent confusion, but I think it is fine if we use the era1 so either option is good imo

bin/e2hs-writer/src/provider.rs
```diff
@@ -0,0 +1,128 @@
...
+        Ok(Self {
+            first_source,
+            second_source,
+        })
+    }
+
+    pub fn get_era1_for_block(&self, block_number: u64) -> anyhow::Result<Arc<Vec<u8>>> {
+        if block_number >= MERGE_BLOCK_NUMBER {
+            bail!("Invalid logic, tried to lookup era1 file for post-merge block");
+        }
```
KolbyML, line 82:
```suggestion
        ensure!(block_number < MERGE_BLOCK_NUMBER, "Invalid logic, tried to lookup era1 file for post-merge block"); 
```

I think this is nicer

bin/e2hs-writer/src/provider.rs
```diff
@@ -0,0 +1,128 @@
...
+            EraSource::PostMerge(_) => {
+                bail!("Era1 file not found for block number: {block_number}",)
+            }
+        })
+    }
+
+    pub fn get_era_for_block(&self, block_number: u64) -> anyhow::Result<Arc<Era>> {
+        if block_number < MERGE_BLOCK_NUMBER {
+            bail!("Invalid logic, tried to lookup era file for pre-merge block");
+        }
```
KolbyML, line 94:
```suggestion
        ensure!(block_number >= MERGE_BLOCK_NUMBER, "Invalid logic, tried to lookup era file for pre-merge block");
```

bin/e2hs-writer/src/provider.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+async fn fetch_era_file_for_block(
+    http_client: Client,
+    block_number: u64,
+) -> anyhow::Result<Arc<Era>> {
+    if block_number < MERGE_BLOCK_NUMBER {
+        bail!("Invalid logic, tried to lookup era file for pre-merge block");
+    }
```
KolbyML, line 115:
```suggestion
    ensure!(block_number >= MERGE_BLOCK_NUMBER, "Invalid logic, tried to lookup era file for pre-merge block");
```

bin/e2hs-writer/src/reader.rs
njgheorghita: I just thought it might be useful information to leave there for future reference, so people don't need to go digging to figure out what epoch we're operating on. But, i've updated it to just a comment

bin/e2hs-writer/src/reader.rs
njgheorghita: It seems pretty obvious to me that it's `historical_batch` given the format string there...

KolbyML: then this variable should be called `historical_batch`

bin/e2hs-writer/src/utils.rs
```diff
@@ -0,0 +1,74 @@
...
+    })
+}
+
+pub fn pre_deneb_execution_payload_to_header(
+    payload: ExecutionPayloadCapella,
+    transactions: &[TxEnvelope],
+    withdrawals: Vec<Withdrawal>,
+) -> anyhow::Result<Header> {
+    let transactions_root = calculate_transaction_root(transactions);
+    let withdrawals_root = calculate_withdrawals_root(&withdrawals);
```
KolbyML, line 50:
```suggestion
    withdrawals: &[Withdrawal],
) -> anyhow::Result<Header> {
    let transactions_root = calculate_transaction_root(transactions);
    let withdrawals_root = calculate_withdrawals_root(withdrawals);
```

I tested locally and we shouldn't need a Vec<> or to clone the withdrawals

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,388 @@
...
+            .message_capella()
+            .map_err(|e| anyhow!("Unable to decode capella block: {e:?}"))?;
+        let payload = block.body.execution_payload.clone();
+        let transactions = decode_transactions(&payload.transactions)?;
+        let withdrawals: Vec<Withdrawal> =
+            payload.withdrawals.iter().map(Withdrawal::from).collect();
+        let header = pre_deneb_execution_payload_to_header(
+            payload.clone(),
+            &transactions,
+            withdrawals.clone(),
```
KolbyML, line 204:
```suggestion
        let withdrawals: Vec<Withdrawal> =
            payload.withdrawals.iter().map(Withdrawal::from).collect();
        let header = pre_deneb_execution_payload_to_header(
            payload.clone(),
            &transactions,
            &withdrawals,
```

I tested locally and we shouldn't need to clone withdrawals here

bin/e2hs-writer/src/reader.rs
njgheorghita: yeah, i made that change. but sometimes i do think there is a fair trade off between lengthy and simplified var names, given that the reader can understand the meaning without too much effort. especially in tests.

crates/e2store/src/e2hs.rs
```diff
@@ -0,0 +1,339 @@
...
+            E2HSBlockIndexEntry::try_from(file.entries.last().expect("missing block index entry"))
+                .expect("invalid block index entry")
+                .block_index;
+        let index = block_index.indices[block_number as usize];
+        let header_bytes = raw_e2hs
+            .get((index as usize)..(index as usize + 8))
+            .unwrap();
+        let header = Header::deserialize(header_bytes).expect("invalid header");
+        let hwp_length = header.length as u64;
+        let hwp_bytes = raw_e2hs
```
KolbyML, line 330: I am confused I don't think these are headers with proof no?

njgheorghita: .... i'm pretty sure they are? i deserialize it just a few lines below

KolbyML (approved):
I wonder what @morph-dev would think, most of the apparent problems have been addressed. I do think how some things were structured makes it a little hard for me to read, but overall I assume the code will work and serve its purpose, we can always iterate on it as time goes by.

crates/e2store/src/e2hs.rs
KolbyML: Oh my bad I confusing the imports `use crate::e2store::types::Header;` and `use crate::types::Header;`

bin/e2hs-writer/src/writer.rs
```diff
@@ -0,0 +1,87 @@
...
+        let mut offset = 8;
+        let mut indices = vec![];
+        for block_tuple in &block_tuples {
+            indices.push(offset);
+            let entry: [Entry; 3] = block_tuple.clone().try_into()?;
+            let length = entry.iter().map(|entry| entry.length() as u64).sum::<u64>();
+            offset += length;
+        }
+        let starting_header = &block_tuples[0].header_with_proof.header_with_proof.header;
+        let short_hash = starting_header.hash_slow().as_slice()[0..4].to_vec();
```
morph-dev, line 63: nit: you can merge this and next line into:

```rust
let short_hash = hex_encode(&starting_header.hash_slow()[..4]);
```

I would personally go one step further and not use `hex_encode` and do:

```rust
let short_hash = (&starting_header.hash_slow()[..4]).encode_hex()
```

And then later you don't need to strip prefix.

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,385 @@
...
+    header_validator::HeaderValidator,
+};
+use url::Url;
+
+use crate::{
+    provider::EraProvider,
+    utils::{pre_capella_execution_payload_to_header, pre_deneb_execution_payload_to_header},
+};
+
+pub struct AllBlockData {
```
morph-dev, line 40: nit: Can this be BlockTuple from e2store crate?

bin/e2hs-writer/src/reader.rs
```diff
@@ -0,0 +1,385 @@
...
+        };
+        Ok(AllBlockData {
+            block_number,
+            header_with_proof,
+            body: tuple.body.body,
+            receipts: tuple.receipts.receipts,
+        })
+    }
+
+    async fn get_pre_capella_block_data(&self, block_number: u64) -> anyhow::Result<AllBlockData> {
```
morph-dev, line 119: Most of this and `get_pre_deneb_block_data` is very similar, if not exactly the same. I assume that would be the case for post deneb as well.

I would merge them into one function (e.g. `get_post_merge_block_data`), and have one or two `match block_number` blocks for logic that is different.

morph-dev (approved):
I left bunch of small comments, as I want this to be merged before you go on the leave.

I assume it would be useful if we could generate consecutive e2hs files (instead of 1 at the time). With that in mind, we could have optimized and restructred things differently (so we don't have to download each era file twice). But this is something that we can do in a separate PR, if/when we start supporting that use case.

bin/e2hs-writer/src/reader.rs
njgheorghita: We could, but I'm not sure it's worth it in this case. It would require a fair amount of wrapping/unwrapping the `Entry` types when using non-era1 source data (aka all receipts & era data).

njgheorghita:
> I assume it would be useful if we could generate consecutive e2hs files (instead of 1 at the time). 

I see your point here... but off the top of my head, I'm not sure there's a serious need for this optimization. I expect that the primary use case for this code will be "tracking" the head and generating new `e2hs` files for each new epoch. I don't think it will be used in the "genesis -> head" generation more than a handful of times to initially generate the files, so I'm not sure the performance improvement you suggest is worth implementing at this point in time

PR #1730, opened 2025-03-26
refactor: ownership/from traits in e2store

njgheorghita:
### What was wrong?
Addressed some remaining comments from #1727 related to ownership patterns we use in `e2store`

### How was it fixed?
- removed some clones & updated ownership in conversion traits

### To-Do


KolbyML (approved):
:shipit: looks good

morph-dev (approved):
I found several places where we use `impl TryFrom<& X > for Entry`, but conversion never fails. I know that this is similar to original implementation (using `TryInto`), and not something that you introduced.

I would say we should use `impl From<& X > for Entry` whenever we can.

Looking into other types and their conversion into Entry, it seems that error can happen only during snappy encoding. But since we are always writing to `Vec`, I think even that can't fail, and I wouldn't be against just using `.expect` in that case. But I'm not going to insist on this as it's cleaner to return Error if it can fail, but I would maybe change error type to `std::io::Error` in that case (instead of generic anyhow::Error).

PR #1733, opened 2025-03-27
refactor: remove redundant test vectors

njgheorghita:
### What was wrong?
We're duplicating some test assets in `trin` which are no longer necessary after they've been added to `portal-spec-tests`

### How was it fixed?
- removed redundant assets

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1734, opened 2025-03-27
refactor: print debug logs for silent errors in `handle_offer()`

KolbyML:
### What was wrong?
The Samba team has spent days trying to debug why there `put_content` implementation isn't working, Trin wasn't printing any logs so they had no way to tell why Trin was failing to accept their offer
### How was it fixed?

If these functions return error's print them on the debug log

### DIFF
--- a/crates/portalnet/src/overlay/service/manager.rs
+++ b/crates/portalnet/src/overlay/service/manager.rs
@@ -1129,11 +1129,13 @@ impl<
                                 // If the fallback FINDCONTENT task fails, that's fine for now.
                                 // In the future, we might want to cycle through all available
                                 // fallback peers on an error.
-                                let _ = Self::fallback_find_content(
+                                if let Err(err) = Self::fallback_find_content(
                                     content_key.clone(),
                                     utp_processing,
                                 )
-                                .await;
+                                .await {
+                                    debug!(%err, content_key = ?content_key, "Fallback FINDCONTENT task failed, after uTP transfer failed");
+                                }
                             })
                         })
                         .collect();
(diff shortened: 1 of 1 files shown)


crates/portalnet/src/overlay/service/manager.rs
```diff
@@ -1187,7 +1192,12 @@ impl<
                             None => {
                                 // Spawn a fallback FINDCONTENT task for each content key
                                 // that failed individual processing.
-                                let _ = Self::fallback_find_content(key, utp_processing).await;
+                                if let Err(err) = Self::fallback_find_content(
+                                    key, utp_processing,
+                                )
+                                .await {
+                                    debug!(%err, "Fallback FINDCONTENT task failed, after validating and storing content failed");
```
morph-dev, line 1199: nit: maybe add content key here as well?

crates/portalnet/src/overlay/service/manager.rs
```diff
@@ -1129,11 +1129,13 @@ impl<
...
                                 // In the future, we might want to cycle through all available
                                 // fallback peers on an error.
-                                let _ = Self::fallback_find_content(
+                                if let Err(err) = Self::fallback_find_content(
                                     content_key.clone(),
                                     utp_processing,
                                 )
-                                .await;
+                                .await {
+                                    debug!(%err, content_key = ?content_key, "Fallback FINDCONTENT task failed, after uTP transfer failed");
```
morph-dev, line 1137:
```suggestion
                                    debug!(%err, ?content_key, "Fallback FINDCONTENT task failed, after uTP transfer failed");
```
nit: I think this would work. Same at line 1164

morph-dev (approved):
:ship:

ogenev (approved):

PR #1735, opened 2025-03-28
feature: add e2hs mode to portal-bridge

njgheorghita:
### What was wrong?
With the introduction of the new [e2hs](https://github.com/eth-clients/e2store-format-specs/pull/6) file format, we need a bridge mode to support these files.

This is a very basic bridge mode, that won't really work unless you have all necessary `e2hs` files in a local `test-e2hs` dir, inside the root level of the trin dir. But the intention is that this will be updated to use a centralized server that hosts all `e2hs` files, once they are generated and made available. 

The existing `BridgeMode` type has turned into a bit of a frankenstein, as a result of us trying to push too much into a single cli arg. After this new bridge mode matures, I expect a fair amount of legacy bridge code to be deleted, that's why I decided to create new, isolated cli args for the `e2hs` mode to use, rather than try and squeeze them into the old `BridgeMode` type. 

There's another trade-off to consider. I opted to require the user to provide the entire block range, rather than using a live el provider to detect the head of the chain. I'm not married to this, but I think it's easier if a user just wants to casually run the bridge, without an infura / pandaops account. 

But both of these decisions are fairly trivial, and can be changed as new use-cases for the bridge are implemented. 

### Usage
`"--mode e2hs --e2hs-range 1000-10000 --e2hs-randomize"`: randomize the order in which epochs from block range are gossiped

### How was it fixed?
- added `e2hs` mode to portal-bridge

### To-Do

### DIFF
--- a/bin/portal-bridge/src/bridge/e2hs.rs
+++ b/bin/portal-bridge/src/bridge/e2hs.rs
@@ -1,0 +1,439 @@
+use std::{
+    str::FromStr,
+    sync::{Arc, Mutex},
+};
+
+use alloy::primitives::B256;
+use anyhow::{anyhow, ensure};
+use clap::Parser;
+use e2store::e2hs::{BlockTuple, E2HS};
+use ethportal_api::{
+    jsonrpsee::http_client::HttpClient,
+    types::execution::{
+        block_body::BlockBody, header_with_proof::HeaderWithProof, receipts::Receipts,
+    },
+    HistoryContentKey, HistoryContentValue,
+};
+use futures::future::join_all;
+use rand::seq::SliceRandom;
+use tokio::{
+    sync::{OwnedSemaphorePermit, Semaphore},
+    task::JoinHandle,
+    time::{sleep, timeout, Duration},
+};
+use tracing::{debug, error, info, warn};
+use trin_metrics::bridge::BridgeMetricsReporter;
+use trin_validation::header_validator::HeaderValidator;
+
+use crate::{
+    bridge::history::{HEADER_SATURATION_DELAY, SERVE_BLOCK_TIMEOUT},
+    put_content::gossip_history_content,
+    stats::{HistoryBlockStats, StatsReporter},
+    types::range::block_range_to_epochs,
+};
+
+/// Looks for the e2hs file for the given epoch index in a local directory.
+/// TODO: this will need to be updated to fetch files from a hosted
+/// server once we have a proper source for the files.
+fn get_e2hs_file(epoch_index: u64) -> anyhow::Result<String> {
+    let e2hs_dir = "./test-e2hs";
+    let all_files = std::fs::read_dir(e2hs_dir)?;
+    let files: Vec<String> = all_files
+        .filter_map(|entry| {
+            let entry = entry.ok()?;
+            let path = entry.path();
+            let path_str = path.to_str()?;
+            if path_str.contains(&format!("mainnet-{epoch_index:05}")) {
+                Some(path_str.to_string())
+            } else {
+                None
+            }
+        })
+        .collect();
+    let file_count = files.len();
+    ensure!(
+        file_count == 1,
+        "Expected 1 file for epoch {epoch_index}, found {file_count}"
+    );
+    let file = files.first().expect("to be able to get first file");
+    Ok(file.clone())
+}
+
+#[derive(Debug)]
+pub struct E2HSBridge {
+    portal_client: HttpClient,
+    metrics: BridgeMetricsReporter,
+    gossip_semaphore: Arc<Semaphore>,
+    header_validator: HeaderValidator,
+    block_range: BlockRange,
+    random_fill: bool,
+}
+
+impl E2HSBridge {
+    pub fn new(
+        portal_client: HttpClient,
+        gossip_limit: usize,
+        block_range: BlockRange,
+        random_fill: bool,
+    ) -> anyhow::Result<Self> {
+        let gossip_semaphore = Arc::new(Semaphore::new(gossip_limit));
+        let mode = format!("{}-{}:{}", block_range.start, block_range.end, random_fill);
+        let metrics = BridgeMetricsReporter::new("e2hs".to_string(), &mode);
+        let header_validator = HeaderValidator::new();
+        Ok(Self {
+            portal_client,
+            metrics,
+            gossip_semaphore,
+            header_validator,
+            block_range,
+            random_fill,
+        })
+    }
+
+    pub async fn launch(&self) {
+        info!("Launching E2HS bridge");
+        let epochs = block_range_to_epochs(self.block_range.start, self.block_range.end);
+        let mut epoch_indexes: Vec<u64> = epochs.keys().cloned().collect();
+        if self.random_fill {
+            epoch_indexes.shuffle(&mut rand::thread_rng());
+        }
+        for index in epoch_indexes {
+            let block_range = epochs.get(&index).expect("to be able to get block range");
+            self.gossip_epoch(index, *block_range).await;
+        }
+    }
+
+    async fn gossip_epoch(&self, epoch_index: u64, block_range: Option<(u64, u64)>) {
+        match block_range {
+            Some(_) => {
+                info!("Gossiping block range: {block_range:?} from epoch {epoch_index}",);
+            }
+            None => {
+                info!("Gossiping entire epoch {epoch_index}");
+            }
+        }
+        let Ok(e2hs_path) = get_e2hs_file(epoch_index) else {
+            panic!("Failed to find e2hs file for epoch {epoch_index}");
+        };
+        let raw_e2hs = std::fs::read(e2hs_path).expect("Failed to read e2hs file");
+        let mut serve_block_tuple_handles = vec![];
+        let block_stream = E2HS::iter_tuples(raw_e2hs).expect("to be able to iter tuples");
+        for block_tuple in block_stream {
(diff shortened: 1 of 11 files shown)


bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+                    .header_with_proof
+                    .header_with_proof
+                    .header
+                    .receipts_root,
+            "Receipts root mismatch"
+        );
+        let body = &block_tuple.body.body;
+        body.validate_against_header(&block_tuple.header_with_proof.header_with_proof.header)?;
+        Ok(())
+    }
```
KolbyML, line 417:
```suggestion
    fn validate_block_tuple(&self, block_tuple: &BlockTuple) -> anyhow::Result<()> {
        let header_with_proof = &block_tuple.header_with_proof.header_with_proof;
        self.header_validator
            .validate_header_with_proof(header_with_proof)?;
            
        let body = &block_tuple.body.body;
        body.validate_against_header(&header_with_proof.header)?;            
        
        let receipts = &block_tuple.receipts.receipts;
        ensure!(
            receipts.root()
                == header_with_proof
                    .header
                    .receipts_root,
            "Receipts root mismatch"
        );        
        Ok(())
    }
```

I think ordering it like this would be a bit more readable. Also I don't think we need the extra variable `receipts_root`

Idk up to you

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+        })
+    }
+
+    async fn serve_block_tuple(
+        portal_client: HttpClient,
+        block_tuple: BlockTuple,
+        block_stats: Arc<Mutex<HistoryBlockStats>>,
+        metrics: BridgeMetricsReporter,
+    ) -> anyhow::Result<()> {
+        let header_number = block_tuple
```
KolbyML, line 202:
```suggestion
        let block_number = block_tuple
```
I think this is more clear

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+            .header_with_proof
+            .header_with_proof
+            .header
+            .number;
+
+        info!("Serving block tuple for block #{header_number}");
+
+        let header_hash = block_tuple
```
KolbyML, line 210:
```suggestion
        let block_hash = block_tuple
```

I think this is more clear

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+        };
+        let raw_e2hs = std::fs::read(e2hs_path).expect("Failed to read e2hs file");
+        let mut serve_block_tuple_handles = vec![];
+        let block_stream = E2HS::iter_tuples(raw_e2hs).expect("to be able to iter tuples");
+        for block_tuple in block_stream {
+            let block_number = block_tuple
+                .header_with_proof
+                .header_with_proof
+                .header
+                .number;
```
KolbyML, line 126: We are doing this 3 times with 3 different variable names
- block_number
- number
- header_number

Could we only do it once, then just pass it as a variable and use `block_number` as I think that is the standard terminology.

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    async fn gossip_epoch(&self, epoch_index: u64, block_range: Option<(u64, u64)>) {
+        match block_range {
+            Some(_) => {
+                info!("Gossiping block range: {block_range:?} from epoch {epoch_index}",);
+            }
+            None => {
+                info!("Gossiping entire epoch {epoch_index}");
+            }
+        }
```
KolbyML, line 114:
```suggestion
        match block_range {
            Some(_) => info!("Gossiping block range: {block_range:?} from epoch {epoch_index}"),
            None => info!("Gossiping entire epoch {epoch_index}"),
        }
```

Perhaps we could condense it now like this?

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+        })
+        .collect();
+    let file_count = files.len();
+    ensure!(
+        file_count == 1,
+        "Expected 1 file for epoch {epoch_index}, found {file_count}"
+    );
+    let file = files.first().expect("to be able to get first file");
+    Ok(file.clone())
+}
```
KolbyML, line 60:
```suggestion
fn get_e2hs_file(epoch_index: u64) -> anyhow::Result<String> {
    let e2hs_dir = "./test-e2hs";
    let all_files: Vec<String> = std::fs::read_dir(e2hs_dir)?
        .filter_map(Result::ok)
        .filter_map(|entry| entry.path().to_str().map(str::to_string))
        .collect();
    let e2hs_path = all_files
        .into_iter()
        .find(|path| path.contains(&format!("mainnet-{epoch_index:05}-")))
        .ok_or_else(|| anyhow::anyhow!("Failed to find e2hs file for epoch {epoch_index}"))?;
    Ok(e2hs_path)
}
```

What do you think about doing it like this?

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+            hwp_by_hash_content_key.clone(),
+            hwp_content_value,
+            block_stats,
+        )
+        .await
+        {
+            error!(
+                "Failed to gossip history content key: {:?} - {:?}",
+                hwp_by_hash_content_key, err
+            );
```
KolbyML, line 322:
```suggestion
            error!("Failed to gossip history content key: {hwp_by_hash_content_key:?} - {err:?}");
```

We inline the variables in the error logs in most functions, maybe we should do it here too to be consistent?

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+            hwp_by_number_content_key.clone(),
+            hwp_content_value,
+            block_stats,
+        )
+        .await
+        {
+            error!(
+                "Failed to gossip history content key: {:?} - {:?}",
+                hwp_by_number_content_key, err
+            );
```
KolbyML, line 347:
```suggestion
            error!("Failed to gossip history content key: {hwp_by_number_content_key:?} - {err:?}");
```

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+            body_content_key.clone(),
+            body_content_value,
+            block_stats,
+        )
+        .await
+        {
+            error!(
+                "Failed to gossip history content key: {:?} - {:?}",
+                body_content_key, err
+            );
```
KolbyML, line 371:
```suggestion
            error!("Failed to gossip history content key: {body_content_key:?} - {err:?}");
```

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+            receipts_content_key.clone(),
+            receipts_content_value,
+            block_stats,
+        )
+        .await
+        {
+            error!(
+                "Failed to gossip history content key: {:?} - {:?}",
+                receipts_content_key, err
+            );
```
KolbyML, line 395:
```suggestion
            error!("Failed to gossip history content key: {receipts_content_key:?} - {err:?}");
```

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+
+impl FromStr for BlockRange {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let parts: Vec<&str> = s.split('-').collect();
+        if parts.len() != 2 {
+            return Err(anyhow!("Invalid block range format"));
+        }
```
KolbyML, line 434:
```suggestion
        ensure!(parts.len() == 2, "Invalid block range format");
```

bin/portal-bridge/src/types/range.rs
```diff
@@ -0,0 +1,95 @@
...
+/// The resulting HashMap contains:
+/// - For epochs that are completely covered by the block range: None
+/// - For epochs that are partially covered: Some((start_block, end_block)) where start_block and
+///   end_block represent the actual block numbers that overlap with the epoch
+///
+/// # Returns
+/// HashMap mapping epoch indexes to either None or Some((start_block, end_block))
+pub fn block_range_to_epochs(start_block: u64, end_block: u64) -> HashMap<u64, Option<(u64, u64)>> {
+    let mut epoch_map = HashMap::new();
+    const EPOCH_SIZE: u64 = 8192;
```
KolbyML, line 15:
```suggestion
```

Isn't this a pretty common constant couldn't we import this from somewhere else?

KolbyML (approved):
:shipit: looks good. I think it is fine if we iterate on this like you explained in the PR's description.

bin/portal-bridge/src/bridge/e2hs.rs
njgheorghita: could be nice if you find a way to add this to the linting rules

KolbyML: I think we have the linting rule enabled, but they are still working on fmt to cover all the edge cases in where it misses it

PR #1736, opened 2025-03-28
fix: update short hash used in e2hs file path

njgheorghita:
### What was wrong?
The [spec](https://github.com/eth-clients/e2store-format-specs/pull/6) for `e2hs` was just updated to to use the final block header hash in a file (as opposed to the first) to be used in the file path. 

### How was it fixed?
- updated the `e2hs-writer`

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1739, opened 2025-03-31
fix: ensure our ENR cache has enough room for max sessions

KolbyML:
### What was wrong?

Discv5 uses an timed LRU cache for all of its sessions. In the past we had bugs where a discv5 session was established, we would try and send a message to the node but it wouldn't be in our ENR cache. From talking to Milos on our call today. The `Discv5.session_capacity` should be less than or equal to our `NodeAddrCache` capacity, to prevent this bug.

### How was it fixed?
- I manually set `session_cache_capacity()` well initializing Discv5, to ensure both of our caches use the same value. I also added a comment for context why we are doing this.
- I updated the `NODE_ADDR_CACHE_CAPACITY` const from 512 Enr's to 1000, to be consistent with Discv5's session capacity default, if we want to save RAM I think it is fine to leave it at 512, but I think matching Discv5's default is a better story then, `32 full k-buckets`. They don't declare a const for `1000`, so I just wrote it.

This fix is important as we need it for the new protocol versioning update. As e.x. before this PR

- node contacts us
- node is flushed from our LRU cache as we are spammed with new connections
- node remains in sessions, due to session capacity being double our Enr capacity
- we try to fetch ENR from NodeID, well trying to figure out which PV they use, but get a cache miss
- **Failure** everything worked, except we weren't able to find the Peer's ENR from their NodeID, so we can't handle their messages

This PR fixes this problem as if our ENRCaches capacity is => then the session capacity, we will always have the ENR's for all the active sessions.

### DIFF
--- a/crates/portalnet/src/config.rs
+++ b/crates/portalnet/src/config.rs
@@ -5,9 +5,9 @@ use ethportal_api::types::{enr::Enr, network::Network};
 
 use crate::{bootnodes::Bootnodes, constants::DEFAULT_UTP_TRANSFER_LIMIT};
 
-/// Capacity of the cache for observed `NodeAddress` values.
+/// Discv5 session cache capacity.
 /// Provides capacity for 1000 nodes, to match Discv5's default session_cache_capacity value.
-pub const NODE_ADDR_CACHE_CAPACITY: usize = 1000;
+pub const DISCV5_SESSION_CACHE_CAPACITY: usize = 1000;
 
 #[derive(Clone)]
 pub struct PortalnetConfig {
@@ -17,7 +17,7 @@ pub struct PortalnetConfig {
     pub bootnodes: Vec<Enr>,
     pub no_stun: bool,
     pub no_upnp: bool,
-    pub node_addr_cache_capacity: usize,
+    pub discv5_session_cache_capacity: usize,
     pub disable_poke: bool,
     pub trusted_block_root: Option<B256>,
     /// the max number of concurrent utp transfers
@@ -34,7 +34,7 @@ impl Default for PortalnetConfig {
             bootnodes: Bootnodes::default().to_enrs(Network::Mainnet),
             no_stun: false,
             no_upnp: false,
-            node_addr_cache_capacity: NODE_ADDR_CACHE_CAPACITY,
+            discv5_session_cache_capacity: DISCV5_SESSION_CACHE_CAPACITY,
             disable_poke: false,
             trusted_block_root: None,
             utp_transfer_limit: DEFAULT_UTP_TRANSFER_LIMIT,
--- a/crates/portalnet/src/config.rs
+++ b/crates/portalnet/src/config.rs
@@ -6,9 +6,8 @@ use ethportal_api::types::{enr::Enr, network::Network};
 use crate::{bootnodes::Bootnodes, constants::DEFAULT_UTP_TRANSFER_LIMIT};
 
 /// Capacity of the cache for observed `NodeAddress` values.
-/// Provides capacity for 32 full k-buckets. This capacity will be shared among all active portal
-/// subnetworks.
-pub const NODE_ADDR_CACHE_CAPACITY: usize = discv5::kbucket::MAX_NODES_PER_BUCKET * 32;
+/// Provides capacity for 1000 nodes, to match Discv5's default session_cache_capacity value.
+pub const NODE_ADDR_CACHE_CAPACITY: usize = 1000;
 
 #[derive(Clone)]
 pub struct PortalnetConfig {
--- a/crates/portalnet/src/config.rs
+++ b/crates/portalnet/src/config.rs
@@ -20,7 +20,7 @@ pub struct PortalnetConfig {
     pub node_addr_cache_capacity: usize,
     pub disable_poke: bool,
     pub trusted_block_root: Option<B256>,
-    // the max number of concurrent utp transfers
+    /// the max number of concurrent utp transfers
     pub utp_transfer_limit: usize,
 }
 
--- a/crates/portalnet/src/discovery.rs
+++ b/crates/portalnet/src/discovery.rs
@@ -159,6 +159,11 @@ impl Discovery {
 
         let discv5_config = ConfigBuilder::new(listen_config)
             .request_timeout(Duration::from_secs(3))
+            // Set the session cache capacity to match our node address cache capacity. If our cache
+            // is smaller then the session cache capacity, it can lead to problems where we can't
+            // send replies to nodes that we have a session with, as we wouldn't have enough room in
+            // to store all the Enr's from all of our current established connections.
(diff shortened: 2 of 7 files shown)


KolbyML:
> I think this is good, and I think this will work great, but I would also:
> 
> * try to revert [fix: temporarily skip caching non-portal nodes in discovery #1379](https://github.com/ethereum/trin/pull/1379) and make sure things still work well
>   
>   * I think this might be a proper fix for original issue
>   * we can't relly on ENR containing client info anymore (it's now part of ping), and we can get broken at any point
>   * we should test it with much lower value of capacity (e.g. 50) as default value is likely to store entire network

I will look into this ^ and share the results when I get them

> * consider making our local cache twice as big as  discv5 session cache
>   
>   * might not be needed, but I don't see big downside to it

I am not sure of a benefit of our cache being twice as big as the discv5 session cache? Wouldn't we just be using ram needlessly? I can't think of a reason to go any higher then the `session_cache_capacity`

morph-dev (approved):
Approving as this PR can go as is. My questions/comments are more of a followup, but it's up to you if you want to include them in this PR, wait with this PR until issue is checked or however you want to approach it.

morph-dev:
> I am not sure of a benefit of our cache being twice as big as the discv5 session cache? Wouldn't we just be using ram needlessly? I can't think of a reason to go any higher then the `session_cache_capacity`. As shouldn't the capacities being equal offer identical functionality to our ENR cache being twice the size?

Because we both use LRU cache, there can be inconsistency. For example:
- let's say cache size is 5, and both are the same: [A, B, C, D, E]
- the discv5 update's it's cache (for some reason, let's say it received discv5 PING from node E) and is now: [E, A, B, C, D]
  - our LRU cache is still [A, B, C, D, E]
- new node (F) is added -> our cache: [F, A, B, C, D] discv5 cache: [F, E, A, B, C]
- we receive message from node E, discv5 has session and can handle it, but we can't

KolbyML:
> > I am not sure of a benefit of our cache being twice as big as the discv5 session cache? Wouldn't we just be using ram needlessly? I can't think of a reason to go any higher then the `session_cache_capacity`. As shouldn't the capacities being equal offer identical functionality to our ENR cache being twice the size?
> 
> Because we both use LRU cache, there can be inconsistency. For example:
> 
> * let's say cache size is 5, and both are the same: [A, B, C, D, E]
> * the discv5 update's it's cache (for some reason, let's say it received discv5 PING from node E) and is now: [E, A, B, C, D]
>   
>   * our LRU cache is still [A, B, C, D, E]
> * new node (F) is added -> our cache: [F, A, B, C, D] discv5 cache: [F, E, A, B, C]
> * we receive message from node E, discv5 has session and can handle it, but we can't

Ok it wouldn't be bad for our cache to hold 2000 Enr's it would only be `0.6 megabytes`. I will talk to Age and see what he thinks about the problem

crates/portalnet/src/config.rs
```diff
@@ -6,9 +6,8 @@ use ethportal_api::types::{enr::Enr, network::Network};
 use crate::{bootnodes::Bootnodes, constants::DEFAULT_UTP_TRANSFER_LIMIT};
 
 /// Capacity of the cache for observed `NodeAddress` values.
-/// Provides capacity for 32 full k-buckets. This capacity will be shared among all active portal
-/// subnetworks.
-pub const NODE_ADDR_CACHE_CAPACITY: usize = discv5::kbucket::MAX_NODES_PER_BUCKET * 32;
+/// Provides capacity for 1000 nodes, to match Discv5's default session_cache_capacity value.
+pub const NODE_ADDR_CACHE_CAPACITY: usize = 1000;
```
morph-dev, line 10: In my opinion, we should rename this and cli flag to: DISCV5_SESSION_CACHE_CAPACITY, because that's what it's used for.
The node_addr_cache_capacity is twice that size, but direct usage of the value is discv5 cache capacity.

crates/portalnet/src/discovery.rs
```diff
@@ -484,48 +470,13 @@ impl AsyncUdpSocket<UtpPeer> for Discv5UdpSocket {
...
-    if let Some(cached_enr) = enr_cache.write().await.get(node_id).cloned() {
-        return Ok(cached_enr);
-    }
-
-    if let Some(enr) = discv5.find_enr(node_id) {
-        enr_cache.write().await.put(*node_id, enr.clone());
-        return Ok(enr);
-    }
-
+async fn find_enr(node_id: &NodeId, discv5: &Arc<Discovery>) -> io::Result<Enr> {
```
morph-dev, line 473: This doesn't have to be async anymore, right?

crates/portalnet/src/overlay/service/manager.rs
```diff
@@ -2281,6 +2281,11 @@ impl<
 
     /// Returns an ENR if one is known for the given NodeId.
     pub fn find_enr(&self, node_id: &NodeId) -> Option<Enr> {
+        // Check whether this node id is in our discovery ENR cache
+        if let Some(node_addr) = self.discovery.cached_node_addr(node_id) {
```
morph-dev, line 2285: Do we need both this and `self.discovery.find_enr` at line 2295? Wouldn't find_end first check `cached_node_addr` anyway?

Does order matter?

KolbyML: That is a different `find_enr` function then the one you are thinking about. This `find_enr` only checks discv5 routing tables kbuckets, not our cache

KolbyML: There are higher odds the `cached_node_addr` have the enr vs the kbuckets

crates/portalnet/src/discovery.rs
```diff
@@ -159,6 +158,11 @@ impl Discovery {
 
         let discv5_config = ConfigBuilder::new(listen_config)
             .request_timeout(Duration::from_secs(3))
+            // Set the session cache capacity to match our node address cache capacity. If our cache
```
morph-dev, line 161: nit: I think this comment is a bit outdated.
Because we renamed the cli flag and config field, I don't think we need any special explanation here.

crates/portalnet/src/discovery.rs
```diff
@@ -173,7 +177,9 @@ impl Discovery {
                 .map_err(|e| format!("Failed to add bootnode enr: {e}"))?;
         }
 
-        let node_addr_cache = LruCache::new(portal_config.node_addr_cache_capacity);
+        // We set the cache capacity to double the node address cache capacity pad for
```
morph-dev, line 180: nit: I would rephrase this to:
_Set local NodeAddress cache to twice the size of discv5 session cache to compensate for inconsistency between them_

morph-dev (commented):
Nice work! :rocket: 
Found a bit outdated documentation. No big deal

KolbyML:
I will add these updated comments to my next PR

PR #1740, opened 2025-04-01
fix: we use invalid BlockIndex type 6232 instead of 6632

KolbyML:
### What was wrong?

@ScottyPoi noticed we were encoding the `blockIndex` type wrong

### How was it fixed?

fix it


KolbyML:
I need to make a quick PR to portal-spec-tests as there is invalid test data there

PR #1743, opened 2025-04-02
feat: add Protocol Version support

KolbyML:
### What was wrong?
We don't support Protocol versioning https://github.com/ethereum/portal-network-specs/pull/379
### How was it fixed?

By implementing the specification change. Currently Trin on Mainnet is only set to support `v0`. I set Angelfood to `v0, v1` just because, no body uses AngelFood, but this allows us to easily test changes for V1 in Hive potentially. In the future we could probably just pass a custom `NetworkSpec` through the CLI for testing this kind of stuff in hive

`v1` should be added to the `NetworkSpec.supported_protocol_versions` once we support it and want to enable it on the network.

### To-Do in followup PR's

This PR doesn't implement the `V1`

`V1` will be 2 changes
- https://github.com/ethereum/trin/pull/1720 I will rebase this PR and modify it to exist in the `Protocol Versions` context
- https://github.com/ethereum/portal-network-specs/pull/383 varint size prefix for `FindContent/Content request/response`

So I will implement these 2 features in separate PR's, then we can enable `V1` when we want to

### DIFF
--- a/crates/ethportal-api/src/types/portal_wire.rs
+++ b/crates/ethportal-api/src/types/portal_wire.rs
@@ -4,7 +4,7 @@ use std::{
     sync::Arc,
 };
 
-use alloy::primitives::{map::HashSet, U256};
+use alloy::primitives::U256;
 use alloy_rlp::Decodable;
 use anyhow::anyhow;
 use bimap::BiHashMap;
@@ -21,7 +21,9 @@ use validator::ValidationError;
 use super::{
     bytes::ByteList1100,
     ping_extensions::extension_types::PingExtensionType,
-    protocol_versions::{ProtocolVersion, ProtocolVersionList, ENR_PROTOCOL_VERSION_KEY},
+    protocol_versions::{
+        ProtocolVersion, ProtocolVersionError, ProtocolVersionList, ENR_PROTOCOL_VERSION_KEY,
+    },
 };
 use crate::{
     types::{
@@ -169,8 +171,7 @@ impl NetworkSpec {
         supported_protocol_versions: ProtocolVersionList,
     ) -> anyhow::Result<Self> {
         // Ensure supported protocol versions are ordered chronologically with no duplicates.
-        supported_protocol_versions.verify_ordered()?;
-        supported_protocol_versions.verify_unique()?;
+        supported_protocol_versions.is_strictly_sorted_and_specified();
 
         Ok(Self {
             portal_subnetworks,
@@ -209,26 +210,23 @@ impl NetworkSpec {
     pub fn latest_common_protocol_version(
         &self,
         enr: &Enr,
-    ) -> anyhow::Result<Option<ProtocolVersion>> {
-        let Some(their_supported_versions) = enr
+    ) -> Result<ProtocolVersion, ProtocolVersionError> {
+        let Some(other_supported_versions) = enr
             .get_decodable::<ProtocolVersionList>(ENR_PROTOCOL_VERSION_KEY)
-            .transpose()?
+            .transpose()
+            .map_err(|_| ProtocolVersionError::FailedToDecode)?
         else {
-            return Ok(Some(ProtocolVersion::V0));
+            return Ok(ProtocolVersion::V0);
         };
-
-        // Convert `their_supported_versions` to a HashSet for O(1) lookups.
-        let their_supported_versions = their_supported_versions.iter().collect::<HashSet<_>>();
 
         // The NetworkSpec's `supported_protocol_versions` are ordered chronologically.
         // Hence, we iterate in reverse order to find the latest common version.
-        for version in self.supported_protocol_versions.iter().rev() {
-            if their_supported_versions.contains(version) {
-                return Ok(Some(*version));
-            }
-        }
-
-        Ok(None)
+        self.supported_protocol_versions
+            .iter()
+            .rev()
+            .find(|v| other_supported_versions.contains(v))
+            .copied()
+            .ok_or(ProtocolVersionError::NoMatchingVersion)
     }
 }
 
--- a/crates/ethportal-api/src/types/portal_wire.rs
+++ b/crates/ethportal-api/src/types/portal_wire.rs
@@ -4,7 +4,7 @@ use std::{
     sync::Arc,
 };
 
-use alloy::primitives::U256;
+use alloy::primitives::{map::HashSet, U256};
 use alloy_rlp::Decodable;
 use anyhow::anyhow;
 use bimap::BiHashMap;
@@ -201,7 +201,10 @@ impl NetworkSpec {
             return Ok(Some(ProtocolVersion::V0));
         };
 
-        // We assume our supported protocol versions are ordered from most old to most recent.
+        // Convert `their_supported_versions` to a HashSet for O(1) lookups.
+        let their_supported_versions = their_supported_versions.iter().collect::<HashSet<_>>();
+
+        // We assume our supported protocol versions are ordered chronologically.
         // Hence, we iterate in reverse order to find the latest common version.
         for version in self.supported_protocol_versions.iter().rev() {
             if their_supported_versions.contains(version) {
--- a/crates/ethportal-api/src/types/portal_wire.rs
+++ b/crates/ethportal-api/src/types/portal_wire.rs
@@ -163,6 +163,22 @@ pub struct NetworkSpec {
 }
 
 impl NetworkSpec {
+    pub fn new(
+        portal_subnetworks: BiHashMap<Subnetwork, String>,
+        network: Network,
+        supported_protocol_versions: ProtocolVersionList,
+    ) -> anyhow::Result<Self> {
+        // Ensure supported protocol versions are ordered chronologically with no duplicates.
+        supported_protocol_versions.verify_ordered()?;
+        supported_protocol_versions.verify_unique()?;
+
+        Ok(Self {
+            portal_subnetworks,
+            network,
+            supported_protocol_versions,
+        })
+    }
+
     pub fn network(&self) -> Network {
         self.network
     }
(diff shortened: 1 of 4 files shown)


crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -179,6 +189,28 @@ impl NetworkSpec {
...
+        enr: &Enr,
+    ) -> anyhow::Result<Option<ProtocolVersion>> {
+        let Some(their_supported_versions) = enr
+            .get_decodable::<ProtocolVersionList>(ENR_PROTOCOL_VERSION_KEY)
+            .transpose()?
+        else {
+            return Ok(Some(ProtocolVersion::V0));
+        };
+
+        // We assume our supported protocol versions are ordered from most old to most recent.
```
ogenev, line 204: With `Ord` and `PartialOrd` implemented, we can change this to something more concise and idiomatic:

 ```rust
 pub fn latest_common_protocol_version(
        &self,
        enr: &Enr,
    ) -> anyhow::Result<Option<ProtocolVersion>> {
        let Some(their_supported_versions) = enr
            .get_decodable::<ProtocolVersionList>(ENR_PROTOCOL_VERSION_KEY)
            .transpose()?
        else {
            return Ok(Some(ProtocolVersion::V0));
        };

        // Find the maximum version that exists in both our and their supported versions
        self.supported_protocol_versions
            .iter()
            .filter(|version| their_supported_versions.contains(version))
            .max()
            .copied()
            .ok_or_else(|| anyhow::anyhow!("No common protocol version found"))
            .map(Some)
            .map_err(|e| e)
    } 

 ```

KolbyML: I don't think this recommendation is more concise or more idiomatic. I mean functional paradigms are cool and all, but I don't think going for the naive solution which is suggestion above is optimal. 

The solution recommended above always goes through the all permutations, which is a little overkill, first off since we know our list is sorted we can started with the biggest numbers first. So I do think I can implement a more optimal solution then I have now, but I don't think what is suggested above is it, it is a regression on my current implementation.

KolbyML: I just updated `latest_common_protocol_version()` to `O(n)` time complexity, whereas my original solution and the solution above are `O(n^2)` time complexity.

So I think my new implementation I just pushed is optimal and should be the one we go with

crates/ethportal-api/src/types/protocol_versions.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
+#[ssz(struct_behaviour = "transparent")]
+pub struct ProtocolVersionList(pub VariableList<ProtocolVersion, U8>);
+
+impl ProtocolVersionList {
+    pub fn new(versions: Vec<ProtocolVersion>) -> Self {
+        Self(VariableList::from(versions))
```
morph-dev, line 84: nit: The difference between `VariableList::from` and `VariableList::new` is that if you pass vector that is longer than allowed (in this case 8), the `from` will just truncate it, while `new` will return error.

I don't like use `from` and having side effect (actually losing some elements), but I don't like dealing with errors. If it was me, I would have probably deal with errors (even if it is simple `.expect(...)`.

Up to you

crates/ethportal-api/src/types/protocol_versions.rs
```diff
@@ -0,0 +1,138 @@
...
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum::U8, VariableList};
+
+pub const ENR_PROTOCOL_VERSION_KEY: &str = "pv";
+
+/// Portal Protocol Versions
+///
+/// https://github.com/ethereum/portal-network-specs/blob/master/protocol-version-changelog.md
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ProtocolVersion {
```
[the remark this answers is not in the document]
morph-dev: I agree with @ogenev  that this enum should implement `Ord` and `PartialOrd`.

But instead of doing `Ord` manually, I would convert `ProtocolVersion` to `u8` and compare those.

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -179,6 +205,31 @@ impl NetworkSpec {
                 "Cannot find protocol identifier for subnetwork: {subnetwork}"
             ))
     }
+
+    pub fn latest_common_protocol_version(
+        &self,
+        enr: &Enr,
+    ) -> anyhow::Result<Option<ProtocolVersion>> {
+        let Some(their_supported_versions) = enr
```
morph-dev, line 213: nit: I think it's more common to call this variable: `other_supported_versions`. Up to you.

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -179,6 +205,31 @@ impl NetworkSpec {
...
+            .get_decodable::<ProtocolVersionList>(ENR_PROTOCOL_VERSION_KEY)
+            .transpose()?
+        else {
+            return Ok(Some(ProtocolVersion::V0));
+        };
+
+        // Convert `their_supported_versions` to a HashSet for O(1) lookups.
+        let their_supported_versions = their_supported_versions.iter().collect::<HashSet<_>>();
```
morph-dev, line 221: Considering that both this and `self.supported_protocol_versions` will be of length 1 or 2 in most cases, I don't think there is value in making it a `HashSet`. In fact, I think that might make it slower in most of the cases.

So I would just remove this line.

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -179,6 +205,31 @@ impl NetworkSpec {
                 "Cannot find protocol identifier for subnetwork: {subnetwork}"
             ))
     }
+
+    pub fn latest_common_protocol_version(
+        &self,
+        enr: &Enr,
+    ) -> anyhow::Result<Option<ProtocolVersion>> {
```
morph-dev, line 212: nit: I'm not sure how you plan to use this function. I can see that we might want to just return `None` if there is an error decoding the field (maybe print warning here). Up to you

crates/ethportal-api/src/types/protocol_versions.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    pub fn verify_unique(&self) -> anyhow::Result<()> {
+        let mut seen = HashSet::new();
+        ensure!(
+            self.0.iter().all(|version| seen.insert(version)),
+            "Version list contains duplicates"
+        );
+        Ok(())
+    }
```
morph-dev, line 106: If you implement Ord, you can replace both of these functions with:

```rust
pub fn is_strictly_sorted_and_specified(&self) -> bool {
    for i in 0..self.len() {
        if matches!(self.0[i], ProtocolVersion::UnspecifiedVersion(_)) {
            return false;
        }
        if i > 0 && self.0[i - 1] >= self.0[i] {
            return false;
        }
    }
    true
}
```

crates/ethportal-api/src/types/portal_wire.rs
```diff
@@ -152,15 +157,36 @@ pub enum DiscoveryRequestError {
...
 }
 
 impl NetworkSpec {
+    pub fn new(
+        portal_subnetworks: BiHashMap<Subnetwork, String>,
+        network: Network,
+        supported_protocol_versions: ProtocolVersionList,
+    ) -> anyhow::Result<Self> {
+        // Ensure supported protocol versions are ordered chronologically with no duplicates.
+        supported_protocol_versions.verify_ordered()?;
```
morph-dev, line 172: We should also check that none of the versions is `Unspecified` (check my other comment).

morph-dev (approved):
Mostly looks ok, but as you will see in the comments, I think that we:
- Should implement Ord for ProtocolVersion
- Have simpler check for uniqueness and sorted properties
- Add check that each protocol version is one of the specified versions
- Use of HashSet is a waste since we will have lists of only 1-2 elements (and only 1 in most of the cases)

crates/ethportal-api/src/types/portal_wire.rs
KolbyML: I just remembered the max size is 8 elements per list so 64 combinations. Saying that we will have a length of likely 2 that is a worst case of 16. I don't think we can make assumptions on what a peer does like this, but since the size of their list is limited, I think you are right here

crates/ethportal-api/src/types/portal_wire.rs
KolbyML: I will use an enum. In the None case we would want to desecore maybe ban the peer? But for the error case we would 100% want to ban them

KolbyML: I ended up just using a Result instead of an Option, with an Error Enum I made

crates/ethportal-api/src/types/portal_wire.rs
KolbyML: Hmm because Milos has justified us not using a HashSet. 

I will go with this

```
    pub fn latest_common_protocol_version(
        &self,
        enr: &Enr,
    ) -> Result<ProtocolVersion, ProtocolVersionError> {
        let other_supported_versions = enr
            .get_decodable::<ProtocolVersionList>(ENR_PROTOCOL_VERSION_KEY)
            .transpose()
            .map_err(|_| ProtocolVersionError::FailedToDecode)?
            .ok_or(ProtocolVersionError::NoMatchingVersion)?;

        self.supported_protocol_versions
            .iter()
            .rev()
            .find(|v| other_supported_versions.contains(v))
            .copied()
            .ok_or(ProtocolVersionError::NoMatchingVersion)
    }
```

It is like the solution you suggested as in it uses functional stuff, expect it will exit early if it finds a match like my original approach, so I hope this is considered a good middle ground

KolbyML:
> * Use of HashSet is a waste since we will have lists of only 1-2 elements (and only 1 in most of the cases)

I wrote this a bit in my comment above, but we have no control over how many elements a peer sends us, but since it is capped at 8 elements, I agree it is likely not needed

PR #1751, opened 2025-04-10
feat: enable protocol V1 on mainnet

KolbyML:
### What was wrong?
we have now fully implemented V1, but it isn't enabled on mainnet. V1 is high value for us because we maintain majority of the bridging infrastructure. V1 includes `AcceptCodes` which massive improves our bridges insights which should allow for a higher understanding of Bridge performance, but also more intelligent bridges to be written
### How was it fixed?

- I moved NetworkSpecs into it's own file, as it is fairly unrelated to `portal-wire.rs`
- I enabled V1 on mainnet, and disable V1 on angelfood, so we still run tests for V0 & V1


morph-dev (approved):
Can you run hive tests locally using this PR before you merge it?
I would like to make sure that there is no weird interaction with other clients.

I guess we would be missing a case where two trin clients communicate, where one support V0 and other supports V1 (old and new release). But it should be ok if that's the only case that we are missing.

KolbyML:
I ran hive tests and the results were as expected, so I am going to merge this PR

PR #1754, opened 2025-04-11
perf: make e2hs writer 44x faster for generating E2HS files from Era1 files

KolbyML:
### What was wrong?

We where deserializing the Era1 file every time for every block retrieved from the Era1 file. So we were deserializing the Era1 file 8192 times. 

Before time: 8m27.011s

After my fix: 0m11.451s

^ these results came from generating E2HS file 1655
### How was it fixed?

I removed `get_tuple_by_index()` as the function is just not good and encourages bad patterns. Instead of using `get_tuple_by_index` I decoded the Era1 file once, then parsed the `block_tuple` out when needed


morph-dev (approved):
:ship:

PR #1755, opened 2025-04-12
fix: e2hs-writer can't generate merge transition e2hs file

KolbyML:
### What was wrong?

When I tried to generate E2HS 1896 which is the E2HS file which includes the merge boundary. So we use both ERA and ERA1 files in generating this E2HS file unlike all the other files. But when I attempted to build E2HS file 1896 it failed. This was due to E2HS 1896 requiring access to ERA1 1896 and ERA 574&575 to get all the blocks required to build E2HS 1896.

### How was it fixed?

I made it so the provider code could optionally download a 3rd Era file, only if e2hs-writer is building the E2hs file 1896

### DIFF
--- a/bin/e2hs-writer/src/provider.rs
+++ b/bin/e2hs-writer/src/provider.rs
@@ -22,6 +22,8 @@ use trin_validation::constants::{EPOCH_SIZE, MERGE_BLOCK_NUMBER};
 pub struct EraProvider {
     first_source: EraSource,
     second_source: Option<EraSource>,
+    /// used for merge-boundary as 1 era1 file and 2 era files are needed
+    third_source: Option<EraSource>,
 }
 
 pub enum EraSource {
@@ -67,14 +69,29 @@ impl EraProvider {
             }
             false => {
                 // todo: if first source is era1, we know the era for next range
+                info!("Fetching second e2store file for epoch: {epoch}");
                 let era = fetch_era_file_for_block(http_client.clone(), ending_block).await?;
-                info!("Second e2store file fetched for epoch: {epoch}");
                 Some(EraSource::PostMerge(era))
             }
         };
+
+        let third_source = match !(starting_block..ending_block).contains(&MERGE_BLOCK_NUMBER) {
+            true => {
+                info!("No third e2store file required for epoch: {epoch}");
+                None
+            }
+            false => {
+                // todo: if first source is era1, we know the era for next range
+                info!("Fetching third e2store file for epoch: {epoch}");
+                let era = fetch_era_file_for_block(http_client.clone(), MERGE_BLOCK_NUMBER).await?;
+                Some(EraSource::PostMerge(era))
+            }
+        };
+
         Ok(Self {
             first_source,
             second_source,
+            third_source,
         })
     }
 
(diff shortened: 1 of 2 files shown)


bin/e2hs-writer/src/provider.rs
```diff
@@ -21,6 +21,8 @@ use trin_validation::constants::{EPOCH_SIZE, MERGE_BLOCK_NUMBER};
 pub struct EraProvider {
     first_source: EraSource,
     second_source: Option<EraSource>,
+    /// used for merge-boundary as 1 era1 file and 2 era files are needed
```
morph-dev, line 24: Instead of having first, second and third source, can we just have: `sources: Vec<EraSource>`? 

We would have to modify logic below to something like (combined with other pr):

```rust
let mut next_block = starting_block;
let mut sources = vec![];

while next_block < ending_block {
    let source  = if next_block < MERGE_BLOCK_NUMBER {
        ...
        let raw_era1 = fetch_bytes(http_client.clone(), era1_path).await?;
        let era1 = Era1::deserialize(&raw_era1)?;
        EraSource::PreMerge(Arc::new(era1))
    } else {
        let era = fetch_era_file_for_block(http_client.clone(), starting_block).await?;
        EraSource::PostMerge(era)
    }
    next_block = source.last_block() + 1; // we would have to implement this, but it should be simple
    sources.push(source);
}
```

This simplifies and removes duplicated logic.

I think this can also simplify `get_era_for_block`.

What do you think?

morph-dev (commented):
It feels like we are just trying to make it work, without actually making it work well.

If you are strongly against it, I'm not going to insist. But we have to opportunity to improve logic a bit.

KolbyML:
> It feels like we are just trying to make it work, without actually making it work well.

I don't think there is that much value in improving this code. I think a lot of this code should be changed and traits should be used to get rid of the duplicate code. I am mostly interested in increasing the performance of generating E2HS files from ERA files, as currently it takes over 60 minutes.

So I think
- fixing the performance (which I will do it should be fairly straight forward)
- writing it so it generates a new e2hs file every 27 hours 
are higher value.

I will implement the suggestion above, but saying `without actually making it work well`, this code doesn't add much value, the greatest value is it working at all.

KolbyML:

bin/e2hs-writer/src/provider.rs
```diff
@@ -17,11 +17,10 @@ use trin_validation::constants::{EPOCH_SIZE, MERGE_BLOCK_NUMBER};
...
-// - merge-boundary: will have an era1 & era file
+// - merge-boundary: will have an era1 & 2 era files
 // - post-merge: all epochs will likely have 2 era files
 pub struct EraProvider {
-    first_source: EraSource,
-    second_source: Option<EraSource>,
+    sources: Vec<EraSource>,
 }
 
 pub enum EraSource {
```
morph-dev, line 26: nit: line 28 - part of the previous pr, but I didn't notice it. Should probably not say "raw era1"

morph-dev (approved):
:shipit:

bin/e2hs-writer/src/provider.rs
KolbyML: good catch

PR #1759, opened 2025-04-15
feat: make E2HS bridge fetch from http endpoint, instead of using local directory

KolbyML:
### What was wrong?

We didn't have an HTTP endpoint hosting E2HS files so we temporarily used a local directory system

### How was it fixed?

By using `https://e2hs.ethportal.net/` in the same way we used to use `https://era1.ethportal.net/` for our ERA1 bridge. Currently `https://e2hs.ethportal.net/` is broken because all of the links point to the index page, but that should be fixed tomorrow. The PR itself is good to go

### DIFF
--- a/crates/e2store/src/utils.rs
+++ b/crates/e2store/src/utils.rs
@@ -8,6 +8,7 @@ use url::Url;
 
 const ERA_DIR_URL: &str = "https://mainnet.era.nimbus.team/";
 const ERA1_DIR_URL: &str = "https://era1.ethportal.net/";
+const E2HS_DIR_URL: &str = "https://e2hs.ethportal.net/";
 const E2SS_DIR_URL: &str = "https://e2ss.ethportal.net/index.html";
 pub const ERA1_FILE_COUNT: usize = 1897;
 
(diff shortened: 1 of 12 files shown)


crates/e2store/src/utils.rs
```diff
@@ -67,8 +68,27 @@ pub async fn get_era_files(http_client: &Client) -> anyhow::Result<HashMap<u64,
     Ok(era_files)
 }
 
+pub async fn get_e2hs_files(http_client: &Client) -> anyhow::Result<HashMap<u64, String>> {
+    let e2hs_files = download_e2store_links(http_client, E2HS_DIR_URL).await?;
+    ensure!(
+        !e2hs_files.is_empty(),
+        "No era files found at {ERA_DIR_URL}"
+    );
+    let missing_epochs: Vec<String> = (0..e2hs_files.len())
```
morph-dev, line 77: nit: Should we move this check into `download_e2store_links`?

morph-dev (approved):
Looks good!

:ship:

crates/e2store/src/utils.rs
KolbyML: E2SS doesn't have continuous files generated so we can't

KolbyML: I moved the checks to a function called `assert_nonempty_and_consecutive_epochs`

PR #1760, opened 2025-04-15
feat: update all bridges to use census/offer instead of `TracePutContent`/`TraceGossip`

KolbyML:
### What was wrong?

We no longer want to support `trace_gossip`/`TracePutContent`, it isn't that useful compared to using `trace_offer` and harmful in regards to that you can't properly rate limit a bridge using `trace_gossip`/`TracePutContent`. I want to remove `trace_gossip`/`TracePutContent` as I don't think it is useful anymore and is tech debt.

### How was it fixed?

- Updating the History and Beacon Bridge to use the consensus/offer instead of trace_gossip/trace_put_content
- Making it so the user can only enable 1 bridge at a time. We never really used the run multiple bridges at the same time feature, and at this point I don't think it something we would want to support

### To-Do in followup PR

Remove `trace_gossip`/`TracePutContent` https://github.com/ethereum/trin/pull/1760


bin/portal-bridge/src/bridge/e2hs.rs
morph-dev, line 1: Not related to this PR, but I think we have a problem with gossip logic (at least for this content).


If I'm not mistaken, we start async tasks that would gossip all headers, bodies and receipts at the same time.

I see this as faulty, because we might start gossiping bodies and receipts before headers. Peers that accept them would fail to verify them (because they can't find headers on the network) and drop them. Meaning, we successfully offered the content, but peers don't actually keep the content.

I think approach should be:
- gossip headers (by hash and number) and wait until this is fully finished
- gossip bodies and receipts

bin/portal-bridge/src/cli.rs
```diff
@@ -374,15 +402,13 @@ impl ClientWithBaseUrl {
...
 
 // parser for subnetworks, makes sure that the state network is not ran alongside other subnetworks
-fn subnetwork_parser(subnetwork_string: &str) -> Result<Arc<Vec<Subnetwork>>, String> {
-    let active_subnetworks: Vec<Subnetwork> = subnetwork_string
-        .split(',')
-        .map(Subnetwork::from_str)
-        .collect::<Result<Vec<Subnetwork>, String>>()?;
-    if active_subnetworks.contains(&Subnetwork::State) && active_subnetworks.len() > 1 {
-        return Err("The State network doesn't support being ran with other subnetwork bridges at the same time".to_string());
+fn subnetwork_parser(subnetwork_string: &str) -> Result<Subnetwork, String> {
```
morph-dev, line 405: Why not use: `Subnetwork::from_str`?
In fact, I think we don't need this and you can even remove it from above (line 66), since we already have a check for supported networks in the other file.

morph-dev (approved):
Looks mostly ok. No big comments.

:shipit:

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -389,6 +214,114 @@ impl E2HSBridge {
         );
         Ok(())
     }
+
+    // spawn individual offer tasks of the content key for each interested enr found in Census
+    async fn spawn_offer_tasks(
```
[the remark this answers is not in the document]
KolbyML: Something like this would happen during future refactors maybe ones related to accept codes

bin/portal-bridge/src/bridge/e2hs.rs
KolbyML: > If I'm not mistaken, we start async tasks that would gossip all headers, bodies and receipts at the same time.

```

        // Sleep for 10 seconds to allow headers to saturate network,
        // since they must be available for body / receipt validation.
        sleep(Duration::from_secs(HEADER_SATURATION_DELAY)).await;
```

10 seconds is more then enough time for small headers to propagate. So I don't think what is stated above is a problem here. A problem here though each block_tuple is blocking each other due to the sleep. I will push a quick fix to remove the sleep well maintaining the status quo, of waiting a bit

KolbyML: I don't think there is a reason to wait for header by number, as bodies and receipts don't depend on them for verification

bin/portal-bridge/src/bridge/beacon.rs
```diff
@@ -266,44 +294,30 @@ impl BeaconBridge {
...
-
-        if let Ok(stats) = slot_stats.lock() {
-            stats.report();
-        } else {
-            warn!("Error displaying beacon gossip stats. Unable to acquire lock.");
-        };
+        self.global_offer_report
+            .lock()
+            .expect("Failed to get global offer report lock")
+            .report();
```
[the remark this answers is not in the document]
morph-dev: You can also create something like a heartbeat that will print it periodically.

bin/portal-bridge/src/bridge/e2hs.rs
morph-dev: Ah, yes, I missed the `sleep`. But as you said, every block_header is blocking everything else...
I guess it's fine to leave it as is for now (we can improve it later if we want to).

And I agree, others don't depend on header_by_number. It's just more logical for me to gossip header_by_hash and header_by_number together (but I can also see the argument that we should gossip only what is needed (ie by hash), and gossip everything else afterwards)

bin/portal-bridge/src/bridge/beacon.rs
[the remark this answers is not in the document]
KolbyML: I already removed it, but maybe we can add a heart beat for it in the future

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -157,48 +155,91 @@ impl E2HSBridge {
...
-        self.gossip_body(block_tuple.body.body, block_hash).await;
-        self.gossip_receipts(block_tuple.receipts.receipts, block_hash)
+
+        let census = self.census.clone();
+        let portal_client = self.portal_client.clone();
+        let metrics = self.metrics.clone();
+        let global_offer_report = self.global_offer_report.clone();
+        let offer_semaphore = self.offer_semaphore.clone();
+        tokio::spawn(async move {
+            if let Err(err) = is_finished_rx.await {
```
morph-dev, line 177: I don't like this because there isn't enough time for header to propagate over network. Yes, we passed it to some peers usign offer, but there was no time for recursive gossip to play its role.

I think 10 seconds timer was better from this point of view, and I would say it's good enough

---

Probably the best solution would be:
- gossip all headers
    - we can do this while still fetching receipts, but probably too complicated for very little gain
- wait 10 (or maybe even more) seconds after last offer finishes
- gossip the rest

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -214,114 +255,118 @@ impl E2HSBridge {
...
+#[allow(clippy::too_many_arguments)]
+async fn spawn_offer_tasks(
+    content_key: HistoryContentKey,
+    content_value: HistoryContentValue,
+    census: Census,
+    portal_client: HttpClient,
+    metrics: BridgeMetricsReporter,
+    global_offer_report: Arc<Mutex<GlobalOfferReport>>,
+    offer_semaphore: Arc<Semaphore>,
+    is_finished_tx: Option<oneshot::Sender<()>>,
```
morph-dev, line 270: I think you made this a bit too complicated.
Previous implementation was much simpler and could have been easily modified to sometghin like this:

```rust
let mut tasks = vec![];
for enr in enrs.clone() {
    // ...
    tasks.push(tokio::spawn(async move {
        // ...
    }));
}
tasks.into_iter().collect::<JoinAll<_>>().await
```

With this change, the `spawn_offer_tasks` would finish when all offers finish (not when they are started), and whoever call it can decide if they want to wait or not.

In practice, we would wait for headers, but not for others (or maybe we wait for all others using `JoinAll` as well).

morph-dev (commented):
In principle, if PR is approved but there are more complicated changes (like this one), I think it's better to wait another look from a reviewer.

KolbyML:
me and milos talked on a call and he said he wants to try some ideas locally and potentially make a pr

PR #1761, opened 2025-04-15
refactor: remove `TracePutContent` json-rpc endpoint

KolbyML:
This PR should be reviewed first https://github.com/ethereum/trin/pull/1760

### What was wrong?

`TracePutContent` or originally `TraceGossip` was an endpoint created so that we could build better bridges. In terms of knowing what happened and rate limiting the bridge. It was a time when we were still trying to reach 100% on Glados. `TracePutContent` was superseded by `TraceOffer` which ended up being the superior solution to our original effort.

### How was it fixed?

I removed it

### DIFF
--- a/testing/ethportal-peertest/src/scenarios/put_content.rs
+++ b/testing/ethportal-peertest/src/scenarios/put_content.rs
@@ -9,12 +9,73 @@ use trin::cli::TrinConfig;
 
 use crate::{
     utils::{
-        fixture_block_body_15040708, fixture_header_by_hash_with_proof_15040641,
-        fixture_header_by_hash_with_proof_15040708, fixture_receipts_15040641,
-        wait_for_history_content,
+        fixture_block_body_15040708, fixture_header_by_hash,
+        fixture_header_by_hash_with_proof_15040641, fixture_header_by_hash_with_proof_15040708,
+        fixture_receipts_15040641, wait_for_history_content,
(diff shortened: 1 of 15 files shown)


morph-dev:
Had a quick look. Looks mostly ok. Ping again when #1760 is merged.

KolbyML:

testing/ethportal-peertest/src/scenarios/put_content.rs
```diff
@@ -9,78 +9,12 @@ use trin::cli::TrinConfig;
...
-        fixture_block_body_15040708, fixture_header_by_hash,
-        fixture_header_by_hash_with_proof_15040641, fixture_header_by_hash_with_proof_15040708,
-        fixture_receipts_15040641, wait_for_history_content,
+        fixture_block_body_15040708, fixture_header_by_hash_with_proof_15040641,
+        fixture_header_by_hash_with_proof_15040708, fixture_receipts_15040641,
+        wait_for_history_content,
     },
     Peertest,
 };
-pub async fn test_gossip_with_trace(peertest: &Peertest, target: &Client, network: Network) {
```
morph-dev, line 18: I think we should keep the test as is, but use put_content instead of trace_put_content (probably also rename test to reflect that).

morph-dev (approved):
:rocket:

PR #1762, opened 2025-04-15
feat: implement Ephemeral Header Content Types

KolbyML:
### What was wrong?
We don't implement the Ephemeral Header Content Types https://github.com/ethereum/portal-network-specs/pull/387
### How was it fixed?
Implementing them and adding tests https://github.com/ethereum/portal-network-specs/pull/390

### DIFF
--- a/crates/ethportal-api/src/types/content_key/history.rs
+++ b/crates/ethportal-api/src/types/content_key/history.rs
@@ -18,8 +18,8 @@ pub const HISTORY_BLOCK_HEADER_BY_HASH_KEY_PREFIX: u8 = 0x00;
 pub const HISTORY_BLOCK_BODY_KEY_PREFIX: u8 = 0x01;
 pub const HISTORY_BLOCK_RECEIPTS_KEY_PREFIX: u8 = 0x02;
 pub const HISTORY_BLOCK_HEADER_BY_NUMBER_KEY_PREFIX: u8 = 0x03;
-pub const HISTORY_EPHEMERAL_HEADERS_BY_FIND_CONTENT_KEY_PREFIX: u8 = 0x04;
-pub const HISTORY_EPHEMERAL_HEADER_BY_OFFER_KEY_PREFIX: u8 = 0x05;
+pub const HISTORY_EPHEMERAL_HEADERS_FIND_CONTENT_KEY_PREFIX: u8 = 0x04;
+pub const HISTORY_EPHEMERAL_HEADER_OFFER_KEY_PREFIX: u8 = 0x05;
 
 /// A content key in the history overlay network.
 #[derive(Clone, Debug, Eq, PartialEq)]
@@ -32,8 +32,16 @@ pub enum HistoryContentKey {
     BlockBody(BlockBodyKey),
     /// The transaction receipts for a block.
     BlockReceipts(BlockReceiptsKey),
-    EphemeralHeadersByFindContent(EphemeralHeadersByFindContentKey),
-    EphemeralHeaderByOffer(EphemeralHeaderByOfferKey),
+    /// Ephemeral headers by find content.
+    ///
+    /// This is used to find the headers of a block and its ancestors.
+    /// This response can only be used by FindContent requests.
+    EphemeralHeadersFindContent(EphemeralHeadersFindContentKey),
+    /// Ephemeral header offer.
+    ///
+    /// This is used to offer a block header to a peer. This response can only be used by Offer
+    /// requests. This type contains a single header.
+    EphemeralHeaderOffer(EphemeralHeaderOfferKey),
 }
 
 impl HistoryContentKey {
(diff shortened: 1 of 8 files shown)


crates/ethportal-api/src/types/execution/ephermeral_header.rs
```diff
@@ -0,0 +1,65 @@
+use alloy::{consensus::Header, rlp};
+use ssz::{Decode, Encode, SszDecoderBuilder, SszEncoder};
+use ssz_derive::{Decode, Encode};
+use ssz_types::{typenum::U256, VariableList};
+
+use crate::types::{bytes::ByteList2048, execution::header_with_proof::ssz_header};
```
morph-dev, line 6: nit: maybe we should move `execution::header_with_proof::ssz_header` into its own file (now that its used in multiple places)

crates/ethportal-api/src/types/execution/ephermeral_header.rs
```diff
@@ -0,0 +1,65 @@
...
+}
+
+impl Encode for EphemeralHeadersByFindContent {
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
+
+    fn ssz_append(&self, buf: &mut Vec<u8>) {
+        let offset = <VariableList<ByteList2048, U256> as Encode>::ssz_fixed_len();
+        let mut encoder = SszEncoder::container(buf, offset);
```
morph-dev, line 20: I think this is wrong and we shouldn't use the `SszEncoder::container`. Meaning, I think we should just do:

```rust
self
    .headers
    .iter()
    .map(|header| ByteList2048::new(rlp::encode(header)).expect("Header should be less than 2KB"))
    .collect::<Vec<_>>();
let headers =
    VariableList::<ByteList2048, U256>::new(headers).expect("Input has same length");
headers.ssz_append(buf)
```

Similar for decoding.

morph-dev (commented):
I think we do encoding/decoding wrongly

crates/ethportal-api/src/types/execution/ephermeral_header.rs
morph-dev: Actually, just checked the spec, and I think your implementation is correct.

Maybe maybe spec is wrong in a sense that we have one extra useless container as a wrapper...

KolbyML: I think you are right

crates/ethportal-api/src/types/content_key/history.rs
```diff
@@ -30,6 +32,8 @@ pub enum HistoryContentKey {
     BlockBody(BlockBodyKey),
     /// The transaction receipts for a block.
     BlockReceipts(BlockReceiptsKey),
+    EphemeralHeadersByFindContent(EphemeralHeadersByFindContentKey),
```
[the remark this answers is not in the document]
KolbyML: I choose ``EphemeralHeadersFindContent, EphemeralHeaderOffer (the same as current without "By")`` as the other ones are ephemeral what?

KolbyML:

crates/ethportal-api/src/types/content_key/history.rs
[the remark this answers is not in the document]
morph-dev: I still prefer `EphemeralHeadersLookup` but this is fine as well :)

crates/ethportal-api/src/types/content_key/history.rs
```diff
@@ -32,8 +32,16 @@ pub enum HistoryContentKey {
     BlockBody(BlockBodyKey),
     /// The transaction receipts for a block.
     BlockReceipts(BlockReceiptsKey),
-    EphemeralHeadersByFindContent(EphemeralHeadersByFindContentKey),
-    EphemeralHeaderByOffer(EphemeralHeaderByOfferKey),
+    /// Ephemeral headers by find content.
+    ///
+    /// This is used to find the headers of a block and its ancestors.
+    /// This response can only be used by FindContent requests.
```
morph-dev, line 38:
```suggestion
    /// Ephemeral headers used for FindContent request.
    ///
    /// This is used to find the headers of a block and its ancestors.
    /// This content type can only be used by FindContent requests.
```

crates/ethportal-api/src/types/content_key/history.rs
```diff
@@ -32,8 +32,16 @@ pub enum HistoryContentKey {
...
+    EphemeralHeadersFindContent(EphemeralHeadersFindContentKey),
+    /// Ephemeral header offer.
+    ///
+    /// This is used to offer a block header to a peer. This response can only be used by Offer
+    /// requests. This type contains a single header.
```
morph-dev, line 43:
```suggestion
    /// Ephemeral header used for Offer request..
    ///
    /// This is used to offer an ephemeral block header to a peer. This content type can only be used by Offer
    /// requests. The type contains a single header, but more headers should be included in the single Offer request.
```

I would add more info. Feel free to rephrase it \ formulate it better.

crates/ethportal-api/src/types/content_key/history.rs
```diff
@@ -187,16 +195,16 @@ impl fmt::Display for HistoryContentKey {
...
+                    "EphemeralHeadersFindContent {{ block_hash: {}, ancestor_count: {} }}",
                     hex_encode_compact(ephemeral_headers.block_hash),
                     ephemeral_headers.ancestor_count
                 )
             }
-            Self::EphemeralHeaderByOffer(ephemeral_header) => {
+            Self::EphemeralHeaderOffer(ephemeral_header) => {
                 format!(
-                    "EphemeralHeadersByOffer {{ block_hash: {} }}",
+                    "EphemeralHeadersOffer {{ block_hash: {} }}",
```
morph-dev, line 207:
```suggestion
                    "EphemeralHeaderOffer {{ block_hash: {} }}",
```

morph-dev (approved):
:ship:

crates/ethportal-api/src/types/content_key/history.rs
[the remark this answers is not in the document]
KolbyML: Look up feels too broad to me

PR #1763, opened 2025-04-16
feat: add Electra consensus types

ogenev:
### To-Do


bin/trin-execution/src/era/beacon.rs
```diff
@@ -167,6 +168,51 @@ impl ProcessBeaconBlock for SignedBeaconBlockDeneb {
...
+            timestamp: payload.timestamp,
+            extra_data: payload.extra_data.to_vec().into(),
+            mix_hash: payload.prev_randao,
+            nonce: B64::ZERO,
+            base_fee_per_gas: Some(payload.base_fee_per_gas.to()),
+            withdrawals_root: Some(withdrawals_root),
+            blob_gas_used: Some(payload.blob_gas_used),
+            excess_blob_gas: Some(payload.excess_blob_gas),
+            parent_beacon_block_root: Some(self.message.parent_root),
+            requests_hash: None,
```
KolbyML, line 204:
```suggestion
            requests_hash: None,
```

To calculate the requests_hash, you can take the `execution_requests` from the `BeaconBlockBody` convert them to this alloy type https://github.com/alloy-rs/alloy/blob/main/crates/eips/src/eip7685.rs then call `.requests_hash()`

I am not sure if this is helpful, but I hope it is

PR #1764, opened 2025-04-17
feat: reduce e2hs-writer memory usage by 5.48x for era to e2hs files

KolbyML:
### What was wrong?

My computer is freezing because my computer is running out of memory. I have seen instances where `e2hs-writer` would use over 35GB's of ram.

### How was it fixed?

- Reducing the memory usage caused by downloading receipts. I think there used to be a memory leak as well, which I fixed.
- Reducing the memory usage caused by downloading Era files, we were doing a lot of redownloading&re-deserialization as well, which was slowing things down. I think there was a memory leak here as well as for some reason memory wasn't being cleared when large structures went out of scope.
- I changed the batch_limit to 10 as I was getting rate limited on generating quite a few e2hs files. I tried 20/30 and it didn't work, we can work on a more specification solution in the future, but this is good enough for the time being.
- Don't generate the whole E2HS file in memory before writing to disk. Stream the data to the file. From testing there was no time performance degradation, just a big memory save.
- Decompressed Era files are really big, only store what we need to, to a reasonable degree
- Don't binary search the era_list multiple times as it is slow

#### testing on master generating epoch 2062
```rust
Number from HTOP: 33.4GB RAM usage
dhat: Total:     278,560,633,350 bytes in 656,615,524 blocks
dhat: At t-gmax: 17,793,603,521 bytes in 101,603,533 blocks
dhat: At t-end:  556,688 bytes in 924 blocks
```

#### testing using this PR generating epoch 2062
```rust 
Number from HTOP: 6.095GB RAM usage
dhat: Total:     194,542,923,155 bytes in 512,040,056 blocks
dhat: At t-gmax: 4,594,085,926 bytes in 5,771,296 blocks
dhat: At t-end:  555,632 bytes in 922 blocks
```

I don't understand why the memory wasn't being cleared, but overall I fixed of downloading the same data multiple times, and deseralizing the same data multiple times in this PR.

5.48x in reduce memory usage is massive, this will unblock me from continuing to generate `post-merge to pre-deneb` E2HS files. I have a script which runs N different instances of `e2hs-writer` in parallel using N different json-rpc endpoints

### DIFF
--- a/bin/e2hs-writer/src/provider.rs
+++ b/bin/e2hs-writer/src/provider.rs
@@ -45,12 +45,24 @@ pub struct MinimalEra {
 }
 
 impl MinimalEra {
-    pub fn contains_block(&self, block_number: u64) -> bool {
+    pub fn get_block(
+        &self,
+        block_number: u64,
+    ) -> Option<(CompressedSignedBeaconBlock, &MinimalEra)> {
         let first_block_number = self.blocks[0].block.execution_block_number();
         let last_block_number = self.blocks[self.blocks.len() - 1]
             .block
             .execution_block_number();
-        (first_block_number..=last_block_number).contains(&block_number)
+        if (first_block_number..=last_block_number).contains(&block_number) {
+            if let Some(block) = self
+                .blocks
+                .iter()
+                .find(|block| block.block.execution_block_number() == block_number)
+            {
+                return Some((block.clone(), self));
+            }
+        }
+        None
     }
 }
 
@@ -149,7 +161,12 @@ impl EraProvider {
         );
         Ok(match &self.sources[0] {
             EraSource::PreMerge(era1) => {
-                era1.block_tuples[(block_number % EPOCH_SIZE) as usize].clone()
+                let block_tuple = era1.block_tuples[(block_number % EPOCH_SIZE) as usize].clone();
+                ensure!(
+                    block_tuple.header.header.number == block_number,
+                    "Block number mismatch",
+                );
+                block_tuple
             }
             EraSource::PostMerge(_) => {
                 bail!("Era1 file not found for block number: {block_number}",)
@@ -167,14 +184,8 @@ impl EraProvider {
         );
         for sources in self.sources.iter() {
             if let EraSource::PostMerge(era) = sources {
-                if era.contains_block(block_number) {
-                    if let Some(block) = era
-                        .blocks
-                        .iter()
-                        .find(|block| block.block.execution_block_number() == block_number)
-                    {
-                        return Ok((block.clone(), era));
-                    }
+                if let Some(block) = era.get_block(block_number) {
+                    return Ok(block);
                 }
             }
         }
--- a/bin/e2hs-writer/src/provider.rs
+++ b/bin/e2hs-writer/src/provider.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 use anyhow::{anyhow, bail, ensure};
 use e2store::{
     era::{CompressedSignedBeaconBlock, Era},
-    era1::Era1,
+    era1::{BlockTuple, Era1},
     utils::{get_era1_files, get_era_files},
 };
 use ethportal_api::consensus::beacon_state::HistoricalBatch;
@@ -42,6 +42,16 @@ impl EraSource {
 pub struct MinimalEra {
     pub blocks: Vec<CompressedSignedBeaconBlock>,
     pub historical_batch: HistoricalBatch,
+}
+
+impl MinimalEra {
+    pub fn contains_block(&self, block_number: u64) -> bool {
+        let first_block_number = self.blocks[0].block.execution_block_number();
+        let last_block_number = self.blocks[self.blocks.len() - 1]
+            .block
+            .execution_block_number();
+        (first_block_number..=last_block_number).contains(&block_number)
+    }
 }
 
 impl From<Era> for MinimalEra {
@@ -132,20 +142,22 @@ impl EraProvider {
         Ok(Self { sources })
     }
 
-    pub fn get_era1_for_block(&self, block_number: u64) -> anyhow::Result<Arc<Era1>> {
+    pub fn get_pre_merge(&self, block_number: u64) -> anyhow::Result<BlockTuple> {
         ensure!(
             block_number < MERGE_BLOCK_NUMBER,
             "Invalid logic, tried to lookup era1 file for post-merge block"
         );
         Ok(match &self.sources[0] {
-            EraSource::PreMerge(era1) => era1.clone(),
+            EraSource::PreMerge(era1) => {
+                era1.block_tuples[(block_number % EPOCH_SIZE) as usize].clone()
+            }
             EraSource::PostMerge(_) => {
                 bail!("Era1 file not found for block number: {block_number}",)
             }
         })
     }
 
-    pub fn get_block(
+    pub fn get_post_merge(
         &self,
         block_number: u64,
     ) -> anyhow::Result<(CompressedSignedBeaconBlock, &MinimalEra)> {
@@ -155,11 +167,7 @@ impl EraProvider {
         );
         for sources in self.sources.iter() {
             if let EraSource::PostMerge(era) = sources {
-                let first_block_number = era.blocks[0].block.execution_block_number();
-                let last_block_number = era.blocks[era.blocks.len() - 1]
-                    .block
-                    .execution_block_number();
-                if (first_block_number..=last_block_number).contains(&block_number) {
+                if era.contains_block(block_number) {
                     if let Some(block) = era
                         .blocks
                         .iter()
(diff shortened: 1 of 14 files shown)


bin/e2hs-writer/src/provider.rs
```diff
@@ -94,40 +145,35 @@ impl EraProvider {
...
+    ) -> anyhow::Result<(CompressedSignedBeaconBlock, &MinimalEra)> {
         ensure!(
             block_number >= MERGE_BLOCK_NUMBER,
             "Invalid logic, tried to lookup era file for pre-merge block"
         );
         for sources in self.sources.iter() {
             if let EraSource::PostMerge(era) = sources {
-                if era.contains(block_number) {
-                    return Ok(era.clone());
+                let first_block_number = era.blocks[0].block.execution_block_number();
```
morph-dev, line 158: nit: would be nice if this logic was in MinimalEra, but it's not big deal since this is the only place where it's used.
Up to you.

bin/e2hs-writer/src/writer.rs
```diff
@@ -46,49 +62,47 @@ impl EpochWriter {
...
+                        .header_with_proof
+                        .header_with_proof
+                        .header
+                        .hash_slow(),
+                );
+            }
+
+            e2hs_writer.append_entry(&BlockTupleOrIndexEntry::BlockTuple(block_tuple))?;
+
+            // Flush the writer to ensure memory usage is kept low.
```
morph-dev, line 84: in my opinion, this shouldn't be needed. Or if it is needed, the e2hs_writer should do it.

Checked the implementation of `E2StoreStreamWriter` and by default, we use `BufWriter` with 8KB buffer. So flushing to ensure low memory usage shouldn't be needed.

crates/e2store/src/e2hs.rs
```diff
@@ -44,13 +46,98 @@ use crate::{
...
+            writer,
+            temp_path,
+        })
+    }
+
+    pub fn temp_path(&self) -> &Path {
+        self.temp_path.as_path()
+    }
+
+    pub fn append_entry(&mut self, entry: &BlockTupleOrIndexEntry) -> anyhow::Result<usize> {
```
morph-dev, line 92: In my opinion, this function should just accept `&Entry`. Meaning, I think we don't need `BlockTupleOrIndexEntry` at all and this function would be simple:

```rust
pub fn append_entry(&mut self, entry: &Entry) -> anyhow::Result<usize> {
    self.writer.append_entry(&entry) // We can make E2StoreStreamWriter::append_entry return number of written bytes
}
```

Then in bin/e2hs-writer/src/writer.rs, we would have:

```rust
// around line 66
block_index_indices.push(block_index_offset);
for entry in <[Entry; 3]>::try_from(&block_tuple)? {
    block_index_offset += e2hs_writer.append_entry(&entry)?;
}

...

// around line 91
let block_index = BlockIndex {
    starting_number: self.epoch_index * EPOCH_SIZE,
    indices: block_index_indices,
    count: BLOCK_TUPLE_COUNT as u64,
};
e2hs_writer.append_entry(&E2HSBlockIndexEntry::new(block_index).into())?;
```

morph-dev: Alternatively, if you really want E2HSWriter to be more structured, then you could:
- add `epoch_index` , `block_index_offset` and `block_index_indices` as fields of `E2HSWriter`
- add
    ```rust
    fn append_block_tuple(&mut self, block_tuple: &BlockTuple) -> anyhow::Result<()> {
        let entries = <[Entry; 3]>::try_from(block_tuple)?;

        self.block_index_indices.push(self.block_index_offset);
        for entry in entries {
            self.block_index_offset += self.writer.append_entry(entry)?;
        }
    }
    ```

- add
    ```rust
    fn finish(mut self, last_block_hash: B256) -> anyhow::Result<()> {
        ensure!(self.block_index_indices == BLOCK_TUPLE_COUNT);
        let block_index = BlockIndex {
            starting_number: self.epoch_index * EPOCH_SIZE,
            indices: self.block_index_indices,
            count: BLOCK_TUPLE_COUNT as u64,
        };
        self.writer.append_entry(&E2HSBlockIndexEntry::new(block_index).into())?;

        self.write.flush()?;

        let short_hash = &last_block_hash.encode_hex()[..8];
        // the rest of the current `finish` function
        // ...
    }
    ``` 

To be honest, I prefer this. Up to you.

morph-dev (commented):
Mostly ok. Added some suggestions for improving E2HSWriter

bin/e2hs-writer/src/provider.rs
```diff
@@ -94,40 +145,35 @@ impl EraProvider {
         })
     }
 
-    pub fn get_era_for_block(&self, block_number: u64) -> anyhow::Result<Arc<Era>> {
+    pub fn get_block(
```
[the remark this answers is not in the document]
KolbyML: I will call it `get_beacon_block` and `get_execution_block`

As era block and era1 block doesn't make sense to me

crates/e2store/src/e2hs.rs
KolbyML: > In my opinion, this function should just accept `&Entry`. Meaning, I think we don't need `BlockTupleOrIndexEntry` at all and this function would be simple:
> 
> ```rust
> pub fn append_entry(&mut self, entry: &Entry) -> anyhow::Result<usize> {
>     self.writer.append_entry(&entry) // We can make E2StoreStreamWriter::append_entry return number of written bytes
> }
> ```
> 
> Then in bin/e2hs-writer/src/writer.rs, we would have:
> 
> ```rust
> // around line 66
> block_index_indices.push(block_index_offset);
> for entry in <[Entry; 3]>::try_from(&block_tuple)? {
>     block_index_offset += e2hs_writer.append_entry(&entry)?;
> }
> 
> ...
> 
> // around line 91
> let block_index = BlockIndex {
>     starting_number: self.epoch_index * EPOCH_SIZE,
>     indices: block_index_indices,
>     count: BLOCK_TUPLE_COUNT as u64,
> };
> e2hs_writer.append_entry(&E2HSBlockIndexEntry::new(block_index).into())?;
> ```

I don't think this is an option because then it is very easy to create invalid files

KolbyML: > Alternatively, if you really want E2HSWriter to be more structured, then you could:
> 
> * add `epoch_index` , `block_index_offset` and `block_index_indices` as fields of `E2HSWriter`
> * add
>   ```rust
>   fn append_block_tuple(&mut self, block_tuple: &BlockTuple) -> anyhow::Result<()> {
>       let entries = <[Entry; 3]>::try_from(block_tuple)?;
>   
>       self.block_index_indices.push(self.block_index_offset);
>       for entry in entries {
>           self.block_index_offset += self.writer.append_entry(entry)?;
>       }
>   }
>   ```
> * add
>   ```rust
>   fn finish(mut self, last_block_hash: B256) -> anyhow::Result<()> {
>       ensure!(self.block_index_indices == BLOCK_TUPLE_COUNT);
>       let block_index = BlockIndex {
>           starting_number: self.epoch_index * EPOCH_SIZE,
>           indices: self.block_index_indices,
>           count: BLOCK_TUPLE_COUNT as u64,
>       };
>       self.writer.append_entry(&E2HSBlockIndexEntry::new(block_index).into())?;
>   
>       self.write.flush()?;
>   
>       let short_hash = &last_block_hash.encode_hex()[..8];
>       // the rest of the current `finish` function
>       // ...
>   }
>   ```
> 
> To be honest, I prefer this. Up to you.

this is reasonable

bin/e2hs-writer/src/provider.rs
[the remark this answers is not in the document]
morph-dev: How about `get_pre_merge` and `get_post_merge`?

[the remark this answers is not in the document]
KolbyML: sure

KolbyML:

KolbyML:
Tested locally and everything works

bin/e2hs-writer/src/provider.rs
```diff
@@ -44,6 +44,16 @@ pub struct MinimalEra {
     pub historical_batch: HistoricalBatch,
 }
 
+impl MinimalEra {
+    pub fn contains_block(&self, block_number: u64) -> bool {
```
morph-dev, line 48: nit: you might as well put the rest of the logic and call it `get_block`. Up to you.

bin/e2hs-writer/src/provider.rs
```diff
@@ -132,20 +142,22 @@ impl EraProvider {
...
-    pub fn get_era1_for_block(&self, block_number: u64) -> anyhow::Result<Arc<Era1>> {
+    pub fn get_pre_merge(&self, block_number: u64) -> anyhow::Result<BlockTuple> {
         ensure!(
             block_number < MERGE_BLOCK_NUMBER,
             "Invalid logic, tried to lookup era1 file for post-merge block"
         );
         Ok(match &self.sources[0] {
-            EraSource::PreMerge(era1) => era1.clone(),
+            EraSource::PreMerge(era1) => {
+                era1.block_tuples[(block_number % EPOCH_SIZE) as usize].clone()
```
morph-dev, line 152: nit: might be good to double check that block's number matches the requested one

morph-dev (approved):
:ship:

PR #1765, opened 2025-04-17
chore: bump ethportal-api & e2store version

KolbyML:
### What was wrong?
we want to release ethportal-api and e2store which both have breaking api changes
### How was it fixed?

bump the minor versions

### DIFF
--- a/crates/ethportal-api/Cargo.toml
+++ b/crates/ethportal-api/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "ethportal-api"
-version = "0.7.0"
+version = "0.8.0"
 description = "Definitions for various Ethereum Portal Network JSONRPC APIs"
 authors.workspace = true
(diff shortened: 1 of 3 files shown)


crates/ethportal-api/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ethportal-api"
-version = "0.7.0"
+version = "0.8.0"
```
morph-dev, line 3: nit: do we have some breaking changes since the last version? If not, then I would go with `0.7.1`.

Some for e2store.

morph-dev (approved):

crates/ethportal-api/Cargo.toml
KolbyML: we talked and Milos said I am fine to proceed

PR #1767, opened 2025-04-18
chore(book): fix final PoW block number

Ktl-XV:
### What was wrong?
The "last PoW block" in the book was off by one
Block 15,537,394 is actually the [first PoS block](https://ethereum.org/en/history/#paris)

### How was it fixed?
Changing it to the right block number.
Verify that the other amounts in that section of the book are correct.


KolbyML (approved):
:shipit: looks good

PR #1770, opened 2025-04-20
refactor: e2hs bridge logic

morph-dev:
### What was wrong?

The e2hs bridge wasn't designed very well:

- We would just spawn many tokio tasks and have no way of tracking them, resulting in:
    - they can fail (e.g. if they panic) and we have no way of knowing 
    - we can't wait for them to finish, resulting in the last content that is gossiped not having enough time before tasks is aborted (entire process finishes)
- We use `GlobalOfferReport::is_finished_tx` to detect when headers were gossiped. This has two drawbacks:
    - If any of the offer tasks fails, the `is_finished_tx` will never be called
    - We don't allow any time for headers to propagate through network using recursive gossip

### How was it fixed?

Create `Gossiper` struct, responsible for gossiping the content. It's cheap to clone (which means we can clone it and use it in tasks). We now have access to each task (in the form of `JoinHandle`), and at each gossip granularity (epoch, block_tuple, content key/value,  individual offer)

Removed `is_finished_tx` and starting using HEADER_SATURATION_DELAY again, to wait for header to saturate the network.

### To-Do

### DIFF
--- a/bin/portal-bridge/src/bridge/e2hs.rs
+++ b/bin/portal-bridge/src/bridge/e2hs.rs
@@ -7,7 +7,9 @@ use std::{
 
 use alloy::primitives::B256;
 use anyhow::ensure;
+use bytes::Bytes;
 use clap::Parser;
+use discv5::Enr;
 use e2store::{
     e2hs::{BlockTuple, E2HSMemory},
     utils::get_e2hs_files,
@@ -17,17 +19,19 @@ use ethportal_api::{
     types::{
         execution::header_with_proof::HeaderWithProof, network::Subnetwork, portal_wire::OfferTrace,
     },
-    ContentValue, HistoryContentKey, HistoryContentValue, HistoryNetworkApiClient,
-    OverlayContentKey,
+    BlockBody, ContentValue, HistoryContentKey, HistoryContentValue, HistoryNetworkApiClient,
+    OverlayContentKey, RawContentValue, Receipts,
 };
+use futures::future::join_all;
 use rand::seq::SliceRandom;
 use reqwest::{
     header::{HeaderMap, HeaderValue, CONTENT_TYPE},
     Client,
 };
 use tokio::{
-    sync::{oneshot, OwnedSemaphorePermit, Semaphore},
-    time::timeout,
+    sync::{OwnedSemaphorePermit, Semaphore},
+    task::JoinHandle,
+    time::{sleep, timeout},
 };
 use tracing::{debug, error, info, warn};
 use trin_metrics::bridge::BridgeMetricsReporter;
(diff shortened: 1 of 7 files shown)


bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -143,103 +147,29 @@ impl E2HSBridge {
...
-        )
-        .await;
+            let gossiper = self.gossiper.clone();
+            let block_permit = self.block_semaphore.clone().acquire_owned().await;
+            tasks.push(tokio::spawn(async move {
+                gossiper.gossip_block(block_number, block_tuple).await;
+                drop(block_permit);
+            }));
+        }
+        join_all(tasks).await;
```
KolbyML, line 158: Ideally we aren't blocking at the end here, we can be downloading the next e2hs file and already be gossipping it before this is finished, hence we are wasting time where we could be gossipping content

KolbyML:
> * We would just spawn many tokio tasks and have no way of tracking them, resulting in:
>   
>   * they can fail (e.g. if they panic) and we have no way of knowing

Why would `spawn_offer_tasks()` panic? 

>   * we can't wait for them to finish, resulting in the last content that is gossiped not having enough time before tasks is aborted 

We have dealt with this in the past, by going doing a sleep the same length as the offer_timeout in the `launch()` function, I don't think we need to await these spawns to achieve the same result.

KolbyML:
`HEADER_SATURATION_DELAY` on this regard, we should be able to get headers,bodies,receipts onto the network in under 4 seconds. 15 seconds is a bit much. Especially as this sleep is only relevant on the first pass of data on the network. After the first pass of initially gossiping this content on the network, the bridge will run through the block range a lot faster, as most blocks should already be filled, and after the first pass the bridge is moreso looking to fill holes, declines are cheap

bin/portal-bridge/src/bridge/e2hs.rs
morph-dev: I think that never calling await is bad practice, but I get your point. So, I will return the result of `join_all` and keep track of tasks of the previous epoch one level above.

That way, we can start downloading and gossiping next epoch before previous one fully finishes, but we won't do 3 epochs in parallel. I would say this is acceptable.

morph-dev:
> Why would `spawn_offer_tasks()` panic?

It's called "panic" for a reason, and it's better to be defensive. Even if our code is always absolutely flawless (even if it is now, future changes might break it), we use external libraries that could introduce random weird bugs. It's better to be able to detect if something is failing than to just fail silently.

> We have dealt with this in the past, by going doing a sleep the same length as the offer_timeout in the `launch()` function, I don't think awaiting these tasks is required to achieved the desired result here
>
> You could also just wait until `offer_semaphore.available_permits() == offer_limit`, which is another reason why we wouldn't need to track spawned tokio tasks

True, there are other ways of dealing with this. I think that waiting for tasks to actually finish is the cleanest.

---

Regarding `HEADER_SATURATION_DELAY` I'm fine if we lower it back to 10 seconds.
I'm also fine that we use some value that depends on the success/rejection/failed values. For example:
- if all offers were rejected: we don't wait (recursive gossip wouldn't happen)
- if 50% or more offers were rejected, we wait 2 seconds (we assume headers were previously gossiped, but it's clear that no everybody has it)
- otherwise we wait 10 seconds.

What do you think?

KolbyML:
> Regarding `HEADER_SATURATION_DELAY` I'm fine if we lower it back to 10 seconds. I'm also fine that we use some value that depends on the success/rejection/failed values. For example:
> 
> * if all offers were rejected or failed: we don't wait (recursive gossip wouldn't happen)
> * if 50% or more offers were rejected, we wait 2 seconds (we assume headers were previously gossiped, but it's clear that no everybody has it)
> * otherwise we wait 10 seconds.
> 
> What do you think?

With decline codes you know if the headers are already stored or not. Most of this sounds fine, `otherwise we wait 10 seconds` I don't think we should wait 10 seconds regardless, it takes less then 4 seconds for full CL Blocks to propagate on mainnet, it shouldn't take over 2 seconds for headers to propagate recursively on Portal

KolbyML:
Here is a visualizer for CL block propagation/attesting/aggregation etc https://lab.ethpandaops.io/beacon/slot/live

full CL blocks are consistently propagated in 1-3 seconds across the world, I don't see why headers which are 1 packet, would take longer then this

PR #1771, opened 2025-04-21
feat: change EnrOfferLimit from 2 to 8

KolbyML:
### What was wrong?

Today in the call I said I would add the granular offer code I wrote for the StateBridge to the E2HS bridge, but after thinking about that I think that would be completely overkill, expessially when the History Network has 100-1000x less content for us to offer.

- Census code was originally added with a offer limit of 8 https://github.com/ethereum/trin/pull/1422
- Offer limit lowered to 4 https://github.com/ethereum/trin/pull/1490
- Offer limit lowered to 2 https://github.com/ethereum/trin/pull/1597

Because long term the state bridge won't be apart of `portal-bridge` and I am already running the 4 E2HS bridges I am running with `--enr-offer-limit=8` with good results. I think just increasing the default is good enough for the E2HS bridge

### How was it fixed?

changed the default limit from 2 to 8. It isn't worth the additional complexity, as the E2HS bridge doesn't face the same challenges which warrant it

### DIFF
--- a/bin/portal-bridge/src/cli.rs
+++ b/bin/portal-bridge/src/cli.rs
@@ -32,9 +32,9 @@ use crate::{
 pub const DEFAULT_SUBNETWORK: &str = "history";
 pub const DEFAULT_EXECUTABLE_PATH: &str = "./target/debug/trin";
 
-/// The maximum number of enrs to return in a response,
-/// limiting the number of OFFER requests spawned by the bridge
-/// for each piece of content
+/// The maximum number of peers to send each piece of content.
+///
+/// This is used as a parameter in census, which selects the peers.
 pub const ENR_OFFER_LIMIT: usize = 8;
 
 #[derive(Parser, Debug, Clone)]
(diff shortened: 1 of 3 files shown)


bin/portal-bridge/src/cli.rs
```diff
@@ -33,6 +32,11 @@ use crate::{
 pub const DEFAULT_SUBNETWORK: &str = "history";
 pub const DEFAULT_EXECUTABLE_PATH: &str = "./target/debug/trin";
 
+/// The maximum number of enrs to return in a response,
+/// limiting the number of OFFER requests spawned by the bridge
+/// for each piece of content
```
morph-dev, line 37:
```suggestion
/// The maximum number of peers to send each piece of content.
///
/// This is used as a parameter in census, which selects the peers.
```

nit: I think this comment was written mostly from census point of view, and now is a bit outdated. Feel free to update it.

PR #1772, opened 2025-04-22
feat: remove `disable_history_storage`

KolbyML:
### What was wrong?

Our nodes had a flag which was temporarily added called `--disable-history-storage`. This was a temporary step which was supposed to be removed after we upgraded our proof format
image

### How was it fixed?

remove it

### DIFF
--- a/bin/trin/src/cli.rs
+++ b/bin/trin/src/cli.rs
@@ -216,13 +216,6 @@ pub struct TrinConfig {
         value_parser = max_radius_parser,
     )]
     pub max_radius: Distance,
-
-    #[arg(
-        long = "disable-history-storage",
-        help = "Disable storing all history data locally. This is a temporary flag used for upgrading the network, and should be removed once the upgrade is complete.",
-        default_value = "false"
-    )]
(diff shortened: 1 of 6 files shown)


KolbyML:
> The main reason we introduced flag is because we weren't verifying all content types. We still don't.
> 
> For example, we don't verify any historical summaries proofs at the moment. We just assume they are valid:
> 
> https://github.com/ethereum/trin/blob/9d6db3d0e632467c7e3c934aa4b77599d36631af/crates/validation/src/header_validator.rs#L77
> 
> Can you double check and make sure that all history content types are validated. If we don't support their validation yet, we shouldn't store that content.
> 
> Also, can you reference #1666 and update it once this is merged (mark corresponding part as done).

I knew we didn't validate Capella HistoricalSummaries, but I seen an error return was set and assumed it was accounted for.

I updated the Ok(()) to an Err(()) so the PR should be ok to merge now

KolbyML:
Everything looks good now

morph-dev (approved):

PR #1774, opened 2025-04-22
feat: enable post-capella HistoricalSummary Header validation

KolbyML:
### What was wrong?
HistoricalSummary HeaderWithProof validation wasn't enabled
### How was it fixed?

Enabled it, the E2HS bridge has no way to verify post-capella HeaderWithProofs as it doesn't have access to a provider and we don't want to give it access to one. So the assumption is the the E2HS bridge is downloading the files from a trusted source.

For this to work, it requires the BeaconNetwork or validation will fail

### DIFF
--- a/bin/e2hs-writer/src/reader.rs
+++ b/bin/e2hs-writer/src/reader.rs
@@ -25,9 +25,7 @@ use ethportal_api::{
 use portal_bridge::api::execution::ExecutionApi;
 use tokio::try_join;
 use trin_execution::era::beacon::decode_transactions;
-use trin_validation::{
-    accumulator::PreMergeAccumulator, constants::EPOCH_SIZE, header_validator::HeaderValidator,
-};
+use trin_validation::{accumulator::PreMergeAccumulator, constants::SLOTS_PER_HISTORICAL_ROOT};
 use url::Url;
 
 use crate::{
@@ -63,27 +61,27 @@ impl EpochReader {
     ) -> anyhow::Result<Self> {
         let execution_api = ExecutionApi::new(el_provider_url.clone(), el_provider_url, 10).await?;
         let latest_block = execution_api.get_latest_block_number().await?;
-        let maximum_epoch = latest_block / EPOCH_SIZE;
+        let maximum_epoch = latest_block / SLOTS_PER_HISTORICAL_ROOT;
         ensure!(
             epoch_index <= maximum_epoch,
             "Epoch {epoch_index} is greater than the maximum epoch {maximum_epoch}"
         );
 
-        let starting_block = epoch_index * EPOCH_SIZE;
+        let starting_block = epoch_index * SLOTS_PER_HISTORICAL_ROOT;
         let epoch_accumulator = if network_spec().is_paris_active_at_block(starting_block) {
             None
         } else {
             Some(Arc::new(
                 lookup_epoch_acc(
                     epoch_index,
-                    &HeaderValidator::new_without_historical_summaries().pre_merge_acc,
+                    &PreMergeAccumulator::default(),
                     &epoch_acc_path,
                 )
                 .await?,
             ))
         };
 
-        let ending_block = starting_block + EPOCH_SIZE;
+        let ending_block = starting_block + SLOTS_PER_HISTORICAL_ROOT;
         let is_paris_active = network_spec().is_paris_active_at_block(ending_block);
         let receipts_handle = tokio::spawn(async move {
             if is_paris_active {
--- a/bin/e2hs-writer/src/reader.rs
+++ b/bin/e2hs-writer/src/reader.rs
@@ -76,7 +76,7 @@ impl EpochReader {
             Some(Arc::new(
                 lookup_epoch_acc(
                     epoch_index,
-                    &HeaderValidator::new().pre_merge_acc,
+                    &HeaderValidator::new_without_historical_summaries().pre_merge_acc,
                     &epoch_acc_path,
                 )
                 .await?,
--- a/bin/portal-bridge/src/bridge/e2hs.rs
+++ b/bin/portal-bridge/src/bridge/e2hs.rs
@@ -15,7 +15,9 @@ use e2store::{
 };
 use ethportal_api::{
     types::{
-        execution::header_with_proof::HeaderWithProof, network::Subnetwork, portal_wire::OfferTrace,
+        execution::header_with_proof::{BlockHeaderProof, HeaderWithProof},
+        network::Subnetwork,
+        portal_wire::OfferTrace,
     },
     BlockBody, ContentValue, HistoryContentKey, HistoryContentValue, OverlayContentKey,
     RawContentValue, Receipts,
@@ -95,7 +97,7 @@ impl E2HSBridge {
         Ok(Self {
             gossiper,
             block_semaphore,
-            header_validator: HeaderValidator::new(),
+            header_validator: HeaderValidator::new_without_historical_summaries(),
             block_range,
             random_fill,
             e2hs_files,
@@ -171,7 +173,7 @@ impl E2HSBridge {
                     continue;
                 }
             }
-            if let Err(err) = self.validate_block_tuple(&block_tuple) {
+            if let Err(err) = self.validate_block_tuple(&block_tuple).await {
                 error!("Failed to validate block tuple: {err:?}");
                 continue;
             }
@@ -206,10 +208,19 @@ impl E2HSBridge {
             .unwrap_or_else(|err| panic!("unable to read e2hs file at path: {e2hs_path:?} : {err}"))
     }
 
-    fn validate_block_tuple(&self, block_tuple: &BlockTuple) -> anyhow::Result<()> {
+    async fn validate_block_tuple(&self, block_tuple: &BlockTuple) -> anyhow::Result<()> {
         let header_with_proof = &block_tuple.header_with_proof.header_with_proof;
-        self.header_validator
-            .validate_header_with_proof(header_with_proof)?;
+        // The E2HS bridge doesn't have access to a provider so it can't validate historical summary
+        // Header with Proofs
+        if !matches!(
+            header_with_proof.proof,
+            BlockHeaderProof::HistoricalSummariesCapella(_)
+                | BlockHeaderProof::HistoricalSummariesDeneb(_)
+        ) {
+            self.header_validator
+                .validate_header_with_proof(header_with_proof)
+                .await?;
+        }
         let body = &block_tuple.body.body;
         body.validate_against_header(&header_with_proof.header)?;
         let receipts = &block_tuple.receipts.receipts;
--- a/bin/portal-bridge/src/bridge/e2hs.rs
+++ b/bin/portal-bridge/src/bridge/e2hs.rs
@@ -14,6 +14,7 @@ use e2store::{
     utils::get_e2hs_files,
 };
 use ethportal_api::{
+    consensus::historical_summaries::HistoricalSummaries,
     types::{
         execution::header_with_proof::{BlockHeaderProof, HeaderWithProof},
         network::Subnetwork,
@@ -97,7 +98,9 @@ impl E2HSBridge {
         Ok(Self {
             gossiper,
             block_semaphore,
-            header_validator: HeaderValidator::new_without_historical_summaries(),
+            header_validator: HeaderValidator::new_with_historical_summaries(
+                HistoricalSummaries::default(),
+            ),
             block_range,
             random_fill,
             e2hs_files,
--- a/crates/validation/src/header_validator.rs
+++ b/crates/validation/src/header_validator.rs
@@ -4,7 +4,7 @@ use alloy::{consensus::Header, primitives::B256};
 use alloy_hardforks::EthereumHardforks;
 use anyhow::bail;
 use ethportal_api::{
-    consensus::historical_summaries::{HistoricalSummaries, HistoricalSummary},
+    consensus::historical_summaries::HistoricalSummaries,
     types::{
         execution::header_with_proof::{
             BeaconBlockProofHistoricalRoots, BeaconBlockProofHistoricalSummaries, BlockHeaderProof,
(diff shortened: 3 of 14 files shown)


morph-dev (commented):
Few comments. Would like @ogenev  to take a look as well.

crates/validation/src/header_validator.rs
```diff
@@ -153,15 +201,21 @@ impl HeaderValidator {
         let gen_index = EPOCH_SIZE + block_root_index;
         let historical_summary_index =
```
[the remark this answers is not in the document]
KolbyML: > Are you sure that this index works both pre and post Dencun? 

The index calculation starts at the epoch of the `CAPELLA_FORK_EPOCH`, so you are basically just calculating 8192 slot increments from the fork. The intervals of HistoricalSummaries remain constant from this point.

> Do we have tests that cover all forks?

We don't have proof generation yet, I am assuming that is something you would need to add over time

crates/validation/src/header_validator.rs
```diff
@@ -14,8 +17,50 @@ use crate::{
...
     merkle::proof::verify_merkle_proof,
+    oracle::HeaderOracle,
 };
 
+#[derive(Debug, Clone)]
+pub enum HistoricalSummariesProvider {
+    /// The historical summaries are provided by the header oracle.
+    HeaderOracle(Arc<RwLock<HeaderOracle>>),
+    /// The historical summaries are provided by passed historical summaries.
+    HistoricalSummaries(HistoricalSummaries),
```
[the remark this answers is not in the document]
KolbyML:
```
        let historical_summaries = HistoricalSummaries::from_ssz_bytes(&historical_summaries_bytes)
            .expect("cannot decode HistoricalSummaries bytes");
```

I don't think we should/need to make this a collect, as HistoricalSummaries are indexed by 0, there are also only like 40k bytes, super tiny in terms of test data we will add. So I don't think we need to hyper optimize this to reduce our test data sizes, as this is a drop in the bucket compared to our real offenders.

KolbyML:
@morph-dev ready for another look. @ogenev do you think you will be able to review this PR before you go on leave?

crates/validation/src/header_validator.rs
```diff
@@ -14,43 +17,96 @@ use crate::{
...
 
+#[derive(Debug, Default, Clone)]
+pub enum HistoricalSummariesProvider {
+    /// The historical summaries are provided by the header oracle.
+    HeaderOracle(Arc<RwLock<HeaderOracle>>),
+    /// The historical summaries are provided by passed historical summaries.
+    HistoricalSummaries(HistoricalSummaries),
+    /// We don't have historical summaries, for example the E2HS bridge doesn't use a provider and
+    /// hence can't get HistoricalSummaries. HistoricalSummary proof validation will fail.
+    #[default]
```
morph-dev, line 31: I don't think this should be default. Probably this type shouldn't implement Default.

The same applies for `HeaderValidator`. Few places that currently use default `HeaderValidator` should explicitly create one with `HistoricalSummariesProvider::None`.

If you want "easy" way to create `HeaderValidator` without HistoricalSummariesProvider, then you can have constructor that would be documented: `HeaderValidator::new_without_historical_summaries()`.

bin/e2hs-writer/src/reader.rs
```diff
@@ -70,17 +70,14 @@ impl EpochReader {
...
-                        &header_validator.pre_merge_acc,
-                        &epoch_acc_path,
-                    )
-                    .await?,
-                ))
-            }
+            true => Some(Arc::new(
+                lookup_epoch_acc(
+                    epoch_index,
+                    &HeaderValidator::default().pre_merge_acc,
```
morph-dev, line 76: Why don't we just use `PreMergeAccumulator::default()`?

crates/validation/src/header_validator.rs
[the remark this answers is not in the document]
morph-dev: I think I'm not familiar with how exactly historical summaries work, so correctly me if I'm wrong.

The `get_historical_summary` function accepts epoch for a reason. So whatever historical summaries we use, we should know that they are valid only up to some slot.

After reading a bit about it, I agree that we probably don't need collection. But I would still include epoch.
I would personally prefer: `HistoricalSummaries { epoch: u64, historical_summaries: HistoricalSummaries }` but I'm not against simple `HistoricalSummaries(u64, HistoricalSummaries(`.

This way, the `get_historical_summary` can return descriptive error, which should be much better than what would happen otherwise (I guess error while checking merkle proof).

crates/validation/src/header_validator.rs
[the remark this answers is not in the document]
morph-dev: Sorry, I marked wrong lines.

I think that calling `Self::verify_beacon_block_proof` few lines above wouldn't work correctly. It even says that it's for pre-capella. This is primary reason why I want @ogenev to take a look at this, as I believe he originally implemented it.

---

And I'm not comfortable with us shipping code that doesn't have a single test for all forks. I would rather wait with this PR until we have those tests vectors and we can do it properly.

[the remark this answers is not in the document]
KolbyML: > I think that calling `Self::verify_beacon_block_proof` few lines above wouldn't work correctly. It even says that it's for pre-capella. This is primary reason why I want @ogenev to take a look at this, as I believe he originally implemented it.

https://github.com/ethereum/trin/pull/1748/files#diff-06484484ad3bff928c828d2fc2013746b289cb23a51c53c42bab5a13966f87a9R105-R123

The comment is just out of date, the `beacon_block_body_root_proof` doesn't change between forks as since in the PR lines above, the `beacon_block_body_root_proof` proof is the same format for bellatrix, capella, deneb, pectra



> And I'm not comfortable with us shipping code that doesn't have a single test for all forks. I would rather wait with this PR until we have those tests vectors and we can do it properly.

We have tests for Capella which is good enough for now no? If something is wrong post-capella will fail validation anyways and nothing wrong will happen.

I don't think there is a reason to wait, even in the worst case nothing wrong happens

crates/validation/src/header_validator.rs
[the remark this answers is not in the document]
KolbyML: I don't understand why we would need to return the epoch. `get_historical_summary())` returns the HistoricalSummaries for the given epoch or a more up to date HistoricalSummaries. HistoricalSummaries are already proven, `get_historical_summary()` will error if it can't get the HistoricalSummaries of at least the given epoch.

Also I am a little confused what you are referring to now.
- what HistoricalSummariesProvider::get_historical_summary() returns? I don't think it makes sense to return it here as if we got the epoch we know it contains the expected data for that epoch
- what HeaderOracle::get_historical_summary() returns? For our usecase it doesn't matter what epoch of HistoricalSummaries the database returns, as it either returns the requested version or a more up to date version

bin/e2hs-writer/src/reader.rs
KolbyML: HeaderValidator in the future would need to support multiple networks, PreMergeAccumulator currently only supports mainnet, but I don't see why we wouldn't add sepolia in the future

crates/validation/src/header_validator.rs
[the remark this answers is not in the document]
KolbyML: I am looking forward to @ogenev review as well.

> And I'm not comfortable with us shipping code that doesn't have a single test for all forks. 

We already have tests for Capella. ^ I think it is unreasonable to expect tests for all forks currently and this request is a little unreasonable.

> I would rather wait with this PR until we have those tests vectors and we can do it properly.

I don't think it would make a difference, as if there was a mistake Validation would fail and that would be our worst case, the best case is things just work. We have been waiting for a lot of this code for quite some time. We can always fix things in the future, so I think this fear is unwarranted in this specific circumstances

bin/e2hs-writer/src/reader.rs
KolbyML: I will make the recommended change for the time being

crates/validation/src/header_validator.rs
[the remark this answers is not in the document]
KolbyML: Me and Milos talked in dm's. He didn't see we already have tests for capella, I will double check the tests, and see if I can use the test data from https://github.com/ethereum/portal-spec-tests/pull/45 to test deneb, Milos didn't care if he had tests for pectra yet.

I will do this and ping when I have some time

ogenev (commented):
Did an initial review, other than some naming comments I don't see any major issues.

However, we should add deneb tests before merging to make sure that everything works as expected (those validation has not been tested on Deneb but in theory should work).

ogenev:
> @morph-dev ready for another look. @ogenev do you think you will be able to review this PR before you go on leave?

Yeah, even if I go to leave sooner, I will find time to review it.

KolbyML:

bin/e2hs-writer/src/reader.rs
```diff
@@ -76,7 +76,7 @@ impl EpochReader {
             Some(Arc::new(
                 lookup_epoch_acc(
                     epoch_index,
-                    &HeaderValidator::new().pre_merge_acc,
+                    &HeaderValidator::new_without_historical_summaries().pre_merge_acc,
```
morph-dev, line 79: Seems that we are creating `HeaderValidator` just to reference default `PreMergeAccumulator`. Why not use `PreMergeAccumulator::default()` directly?

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -95,7 +97,7 @@ impl E2HSBridge {
         Ok(Self {
             gossiper,
             block_semaphore,
-            header_validator: HeaderValidator::new(),
+            header_validator: HeaderValidator::new_without_historical_summaries(),
```
morph-dev, line 100: Can't we use the logic from #1777 and fetch historical summaries before we start gossiping? HistoricalSummaries fetched at the start of the bridge should be sufficient for all content that we want to gossip, right?

crates/validation/src/header_validator.rs
```diff
@@ -13,36 +15,107 @@ use ethportal_api::{
...
+#[derive(Debug, Clone)]
+pub enum HistoricalSummariesSource {
+    /// The historical summaries are provided by the header oracle.
+    HeaderOracle(Arc<RwLock<HeaderOracle>>),
+    /// The historical summaries are provided by passed historical summaries.
+    HistoricalSummaries(HistoricalSummaries),
+}
+
+#[derive(Debug, Clone)]
+pub struct HistoricalSummariesProvider {
```
morph-dev, line 38: In my opinion, this and `HistoricalSummariesSource` should be in a separate file, as they have some non-trivial logic as well.

crates/validation/src/header_validator.rs
```diff
@@ -13,36 +15,107 @@ use ethportal_api::{
...
+    pub fn new(source: HistoricalSummariesSource) -> Self {
+        Self {
+            cache: Arc::new(RwLock::new(HistoricalSummaries::default())),
+            source,
+        }
+    }
+
+    pub async fn get_historical_summary(
+        &self,
+        epoch: u64,
```
morph-dev, line 53: If I'm not mistaken, based on the calculation below, there is relationship between epoch and historical_summary_index:
`historical_summary_index := epoch - CAPELLA_FORK_EPOCH*SLOTS_PER_EPOCH/EPOCH_SIZE` ¹

I think we should pass either slot or epoch, and let HistoricalSummariesProvider figure out how to get correct HistoricalSummary.

---

¹ Side note, `SLOTS_PER_EPOCH=32` and `EPOCH_SIZE=8192`, which is very confusing...
I would say that the most common usage of EPOCH refers to 32 slots, and if we are referring to 8192, we should use one of:
- in the context of historical roots/summaries - `SLOTS_PER_HISTORICAL_ROOT` (which we already have declared as constant)
- in the context of PreMergeAccumulator, maybe PRE_MERGE_ACCUMULATOR_EPOCH? open for suggestions

Nothing particularly to do with this PR.

crates/validation/src/header_validator.rs
```diff
@@ -624,7 +729,7 @@ mod test {
     }
 
     fn get_mainnet_header_validator() -> HeaderValidator {
-        let header_validator = HeaderValidator::default();
+        let header_validator = HeaderValidator::new_without_historical_summaries();
```
morph-dev, line 732: I would just move `read_historical_summaries` into this function and change it to:
```rust
fn create_header_validator() -> HeaderValidator {
    let historical_summaries = read_ssz_portal_spec_tests_file(
        PathBuf::from(SPEC_TESTS_DIR)
            .join("headers_with_proof/beacon_data/historical_summaries_at_slot_11476992.ssz"),
    )
    .unwrap();

    HeaderValidator::new(
        HistoricalSummariesProvider::new(
            HistoricalSummariesSource::HistoricalSummaries(historical_summaries),
        ),
    )
}
``` 

And all tests can use it (as they were before).

Also, the `pre_merge_acc.tree_hash_root()` check doesn't really fit in this function anyway, but it should be separate test on its own.

crates/validation/src/header_validator.rs
```diff
@@ -13,36 +15,107 @@ use ethportal_api::{
...
 
+    pub fn new_without_historical_summaries() -> Self {
         Self {
-            pre_merge_acc,
-            historical_roots_acc,
+            pre_merge_acc: PreMergeAccumulator::default(),
+            historical_roots_acc: HistoricalRootsAccumulator::default(),
+            historical_summaries_provider: HistoricalSummariesProvider::new(
+                HistoricalSummariesSource::HistoricalSummaries(HistoricalSummaries::default()),
+            ),
```
morph-dev, line 114: I think with my other suggestions, we no longer need `new_without_historical_summaries()`. Because constructing `HistoricalSummariesProvider` is tedious, I would instead have these 2 constructors:

```rust
pub fn new_with_header_oracle(header_oracle: Arc<RwLock<HeaderOracle>>) {
...
}

pub fn new_with_historical_summaries(historical_summaries: HistoricalSummaries) {
...
}
```

Or you can leave `HeaderValidator::new` as is, and put these constructors in `HistoricalSummariesProvider`

morph-dev (approved):
No major concerns, but several suggestions that I think would be really nice.

:ship:

bin/portal-bridge/src/bridge/e2hs.rs
KolbyML: Me and Milos talked and this concern will be dealt with in a followup PR

crates/validation/src/header_validator.rs
KolbyML: I think we should just use `SLOTS_PER_HISTORICAL_ROOT` as that is the origin of the constant

PR #1775, opened 2025-04-23
feat: make beacon required network if history is enabled

morph-dev:
### What was wrong?

In order to validate post-capella and Ephemeral content, we need beacon network enabled.

### How was it fixed?

Adding check to ensure that beacon network is enabled when history is enabled.

### To-Do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2561,6 +2561,7 @@ dependencies = [
  "ethportal-api",
  "futures",
  "hex",
+ "itertools 0.13.0",
  "jsonrpsee",
  "portal-bridge",
  "portalnet",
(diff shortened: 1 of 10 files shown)


KolbyML:
> In order to validate Ephemeral content, we need beacon network enabled.

Don't forget about post-bellatrix content

morph-dev:
> > In order to validate Ephemeral content, we need beacon network enabled.
> 
> Don't forget about post-bellatrix content

You mean post-capella, right? And yes, you are right, we need it for that as well.

KolbyML (approved):
:shipit: looks good

ogenev (approved):
⛵

PR #1777, opened 2025-04-24
feat: enable HistoricalSummaries bridge + use SSZ instead of JSON Beacon API by default

KolbyML:
### What was wrong?

- We weren't gossiping HistoricalSummaries
- using the beacon api with json is slow, use ssz if avaliable

### How was it fixed?

- Currently we don't have access to a viable `HistoricalSummaries` endpoint so we need to use `debug*/state`, I implemented downloading BeaconAPI calls by SSZ which halfed our `HistoricalSummaries` download time, it should also make things a lot more reliable. I was able to download and decode the state in 25 seconds, instead of over 1 minute using the JSON Beacon API
- The ConsensusAPI will try to download SSZ by default, but it is up to the CL node to determine what it will send us. Some endpoints don't support SSZ for example
- I made our decode logic a lot cleaner, and avoided a lot of the messy clones we were doing

### TODO in future
- use Nimbus/Lodestar `HistoricalSummaries` endpoint once they are in production


bin/portal-bridge/src/api/consensus/rpc_types.rs
```diff
@@ -0,0 +1,60 @@
...
+            finalized: None,
+            data,
+        }
+    }
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(untagged)]
+pub enum VersionedDataResult<T> {
+    Result {
```
morph-dev, line 29: Can't this be: `Result(VersionedDataResponse<T>)`?
I think deserialization would work as expected.

bin/portal-bridge/src/api/utils.rs
```diff
@@ -0,0 +1,112 @@
...
+pub const SSZ_CONTENT_TYPE: &str = "application/octet-stream";
+pub const ACCEPT_PRIORITY: &str = "application/octet-stream;q=1.0,application/json;q=0.9";
+
+#[derive(Debug, Clone, Copy)]
+pub enum ContentType {
+    Json,
+    Ssz,
+}
+
+pub fn url_to_client(
```
morph-dev, line 22: If I'm not mistaken, `ClientWithBaseUrl::new` is not used outside of this function., so this function feels like a proper constructor. So maybe move this logic there?

bin/portal-bridge/src/api/utils.rs
morph-dev, line 1: In general `utils` filenames are bad practice because they don't tell us much about what exactly is there.
In this case, we can easily name this `client.rs` or `http_client.rs` or something like that.

bin/portal-bridge/src/api/consensus/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+                    .await
+                    .map_err(|err| {
+                        anyhow!("Unable to request consensus data from fallback provider: {err:?}")
+                    })
+            }
+        }
+    }
+
+    /// request() should be used instead of this function.
+    async fn request_no_fallback<T>(
```
morph-dev, line 141: nit: Maybe move this logic can be moved inside ClientWithBaseUrl (e.g. `fn get_data_response(&self, ...)`)?

bin/portal-bridge/src/api/consensus/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+        );
+        let client =
+            url_to_client(primary.clone(), request_timeout, ContentType::Json).map_err(|err| {
+                anyhow!("Unable to create primary client for consensus data provider: {err:?}")
+            })?;
+        if let Err(err) = check_provider(&client).await {
+            warn!("Primary consensus data provider may be offline: {err:?}");
+        }
+        Ok(Self {
+            primary,
```
morph-dev, line 49: nit: would it make sense to actually save `client` instead of primary url? That way we don't have to create one every time we want to make a request.

We could potentially do the same for fallback, but maybe it's not so important as we should use primary most of the time...

bin/portal-bridge/src/constants.rs
```diff
@@ -11,7 +11,7 @@ pub const BEACON_GENESIS_TIME: u64 = 1606824023;
 
 /// The timeout in seconds is applied from when the request starts connecting until the response
 /// body has finished. Also considered a total deadline.
-pub const DEFAULT_TOTAL_REQUEST_TIMEOUT: u64 = 20;
+pub const DEFAULT_TOTAL_REQUEST_TIMEOUT: u64 = 60;
```
morph-dev, line 14: If i'm not mistaken, we use the same request timeout regardless of what we are requesting and how big it can be.

Would it make sense to introduce separate timeouts

 I don't think all timeouts should be cli flag, but it makes sense for the detault one to be. Others can be just const and used when appropriate.

morph-dev (commented):
Looks mostly ok, not big concerns.
@ogenev Do you think you have time to take a look as well?

if i'm not mistaken, with this PR, we will start gossiping HistoricalSummaries and potentially be able to prove all history content, right?

bin/portal-bridge/src/constants.rs
KolbyML: The default one is already a flag, I made a seperate const for BEACON_STATE which is only used on that specific request

bin/portal-bridge/src/api/consensus/mod.rs
KolbyML: This logic is beacon api specific, so I don't think it belongs in ClientWithBaseUrl

bin/portal-bridge/src/api/utils.rs
KolbyML: sure

bin/portal-bridge/src/api/consensus/mod.rs
KolbyML: I ended up doing it for both. The old check_provider() didn't work as it would return working even if the request was blocked by not providing api keys.

KolbyML:

bin/portal-bridge/src/api/consensus/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+    ) -> anyhow::Result<VersionedDataResponse<T>>
+    where
+        T: Decode + DeserializeOwned,
+    {
+        let request = client.get(&endpoint)?;
+        let request = if let Some(timeout) = custom_timeout {
+            request.timeout(timeout)
+        } else {
+            request
+        };
```
morph-dev, line 168:
```suggestion
        let request = match custom_timeout {
            Some(timeout) => request.timeout(timeout),
            None => request,
        };
```

tiniest nitpick: I found suggested code a bit more readable. up to you

morph-dev (approved):

bin/portal-bridge/src/api/consensus/rpc_types.rs
```diff
@@ -26,29 +27,14 @@ impl<T> VersionedDataResponse<T> {
...
-                finalized,
-                data,
-            } => Ok(VersionedDataResponse {
-                version,
-                execution_optimistic,
-                finalized,
-                data,
-            }),
+            VersionedDataResult::Result(versioned_data_response) => Ok(versioned_data_response),
             VersionedDataResult::Error(err) => bail!("Failed to deserialize json {err:?}"),
```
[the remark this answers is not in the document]
KolbyML: > The type of the `err` is `serde_json::Value`, which is a bit misleading. Can you rename it? Was it even intentional to print entire json object?

If we fail to decode `VersionedDataResponse` it is because we received an unexpected result which is an error

> Thinking a bit more about usefulness of the entire `VersionedDataResult` struct. Do we have it only so that we can print json response if decoding into `VersionedDataResponse` fails? Or is there some other use for it?

By default serde_json has absolutely useless errors. The endpoint can return anything as an error, which is why we print it.

I will change it from Error to `UnexpectedResponse`.  `VersionedDataResult` is extremely useful, due to how horrible serde_json is

bin/portal-bridge/src/api/consensus/rpc_types.rs
```diff
@@ -26,29 +27,14 @@ impl<T> VersionedDataResponse<T> {
...
-        execution_optimistic: Option<bool>,
-        finalized: Option<bool>,
-        data: T,
-    },
+    Result(VersionedDataResponse<T>),
     Error(Value),
 }
 
 impl<T> VersionedDataResult<T> {
     pub fn to_result(self) -> anyhow::Result<VersionedDataResponse<T>> {
```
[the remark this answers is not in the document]
KolbyML: I will rename name it to `result` for now, if you really want I can change it to `response` I don't like `data_response` as it sounds confusing. If you have other ideas feel free to suggest them

KolbyML:
@morph-dev I just tested and double checked all the endpoints worked. I found a issue with `get_light_client_updates()` as I didn't know how to calculate the period initially, so I would end up getting empty results when testing which did pass. To fix it I add a request function for lists of objects. Currently requesting lists from the BeaconAPI only supports JSON, we can add SSZ support in the future if needed. I added a todo

KolbyML:
@ogenev Milos told me ``ping Ogi directly if he thinks he will have time to review it. If not, you can merge it when you want``. Do you think you have time to review this? If you don't have time no worries I think the main concern would be the very minimal changes in `serve_historical_summaries_with_proof()`, but since you wrote that code and I didn't really change it, I don't think there would be any real concerns.

I just double checked if all endpoints worked as expected and I didn't find any issues (after I fixed `get_light_client_updates()` as stated previously). I tested running all endpoints sequentially and with this PR it takes `17.88s`, only using json it takes `67.09s` a 4x performance gain is pretty nice

KolbyML:
```rust
#[tokio::test]
async fn hi() {
    let api = ConsensusApi::new(
        // Url::from_str("[https://lodestar.mainnet.eu1.ethpandaops.io/").unwrap()](https://lodestar.mainnet.eu1.ethpandaops.io/%22).unwrap()),
        Url::from_str("[https://nimbus.mainnet.eu1.ethpandaops.io/").unwrap()](https://nimbus.mainnet.eu1.ethpandaops.io/%22).unwrap()),
        Url::from_str("[https://nimbus.mainnet.eu1.ethpandaops.io/").unwrap()](https://nimbus.mainnet.eu1.ethpandaops.io/%22).unwrap()),
        90,
    )
    .await
    .unwrap();

    assert!(api.get_beacon_state().await.is_ok());
    assert!(api.get_node_version().await.is_ok());
    assert!(api
        .get_light_client_bootstrap(
            B256::from_hex("0x51e17921dc1204147a86c5bc229c02894052db8435c63290f93e275c95b2c507")
                .unwrap()
        )
        .await
        .is_ok());
    assert!(api
        .get_light_client_updates(11572384 / (32 * 256), 1)
        .await
        .is_ok());
    assert!(api.get_light_client_optimistic_update().await.is_ok());
    assert!(api.get_light_client_finality_update().await.is_ok());
    assert!(api.get_beacon_block_root("head").await.is_ok());
    assert!(api.get_beacon_state_finalized_root().await.is_ok());
}
```

Here is a minimal test. I double checked the results visually as well

ogenev (approved):
I don't see anything that stand out to me, nice PR :shipit: .

PR #1779, opened 2025-04-24
feat: always vector for ExecutionBlockProof types

morph-dev:
# [DEPRECATED] I will split this into multiple PRs. No need for review!

### What was wrong?

The spec is changing the type of ExecutionBlockProof to always be Vector. See https://github.com/ethereum/portal-network-specs/pull/396

### How was it fixed?

Switch to using Vector for ExecutionBlockProof types.
Add proof generation for Deneb.

Note that tests are passing with combination of following PRs:
- https://github.com/ethereum/portal-spec-tests/pull/45 (took only beacon state and block ssz files)
- https://github.com/ethereum/portal-spec-tests/pull/47
-  https://github.com/ethereum/portal-spec-tests/pull/48

### To-Do

### DIFF
--- a/crates/ethportal-api/src/types/execution/header_with_proof.rs
+++ b/crates/ethportal-api/src/types/execution/header_with_proof.rs
@@ -3,36 +3,53 @@ use jsonrpsee::core::Serialize;
 use serde::Deserialize;
 use ssz::SszDecoderBuilder;
 use ssz_derive::{Decode, Encode};
-use ssz_types::{typenum, FixedVector, VariableList};
+use ssz_types::{typenum, FixedVector};
 use tree_hash::TreeHash;
 
-use crate::types::{
-    bytes::ByteList1024,
-    consensus::{
-        beacon_block::{BeaconBlockBellatrix, BeaconBlockCapella},
-        beacon_state::{BeaconStateCapella, HistoricalBatch},
-        proof::build_merkle_proof_for_index,
-    },
-    execution::{
-        block_body::{MERGE_TIMESTAMP, SHANGHAI_TIMESTAMP},
-        ssz_header,
+use crate::{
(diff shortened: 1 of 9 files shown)


crates/ethportal-api/src/types/execution/header_with_proof.rs
```diff
@@ -3,36 +3,53 @@ use jsonrpsee::core::Serialize;
...
+/// - Block number: 17034870
+///     - Note that frst Shapella block is created at slot 6209538 (timestamp: 1681338479)
+pub const SHAPELLA_TIMESTAMP: u64 = 1681338455;
+
+/// The timestamp of the first Dencun (Cancun-Deneb) slot.
+///
+/// - Slot: 8626176
+/// - Epoch: 269568
+/// - Block number: 19426587
+pub const DENCUN_TIMESTAMP: u64 = 1710338135;
```
KolbyML, line 37: One thing to note is long term instead of using these constants we would be required to pass NetworkSpec which would hold the fork timestamps per network mainnet,sepolia etc

KolbyML: So in short going forward if we want to support testnets we won't be able to use any global consts for forks

KolbyML: I will make a followup PR for using `NetworkSpec` instead of const's. I don't think I have concerns with this PR anymore

KolbyML: Here is a PR for it https://github.com/ethereum/trin/pull/1780

morph-dev:
No need to review this PR. I will split it into smaller PRs and merge separately (I made some changes that are no longer needed, and I want to organize code a bit better)

PR #1780, opened 2025-04-25
feat: use NetworkSpec for fork activation checks instead of harded coded values

KolbyML:
### What was wrong?

- This ended up not being a concern here, but https://github.com/ethereum/trin/pull/1779 I do feel anxious when we use hardcoded timestamps to decode things, as we could create a hard situation for ourselves when we want to support testnets.
- We use hardcode block_numbers/timestamps throughout our code, this isn't good if we want to support testnets, as fork_conditions are different for testnets
- I don't want to wait any longer to make this change, as we want to eventually support test networks, and I fear we might accidentally dig a hole for ourselves

### How was it fixed?

Replace all instances of a FORK timestamp/block_number being used, with checking `NetworkSpec` if a fork is enabled or not.

### TO DO in the future
This is a required step in the direction if we want to support testnets (which we do). This PR also helps us avoid us digging traps for ourselves which makes this harder to do in the future, which was one of my main concerns of https://github.com/ethereum/trin/pull/1779 initially

There would be additional work if we would want to support testnets, but this is an initial step


KolbyML:
Milos said using OnceLock might be cleaner so I am going to give it a shot

KolbyML:

I had to use `LazyLock<RwLock<Arc<NetworkSpec>>` instead to make it so the tests work. I also had to remove the `peertest_ping_cross_discv5_protocol_id()` test as we can't run 2 networks at the same time anymore, but to be honest I don't that that test was that important so we should be fine.

crates/validation/src/header_validator.rs
```diff
@@ -516,6 +542,18 @@ mod test {
         EpochAccumulator::from_ssz_bytes(&epoch_acc_bytes).unwrap()
     }
 
+    #[tokio::test]
+    async fn header_oracle_bootstraps_with_default_pre_merge_acc() {
```
morph-dev, line 546: The test name doesn't make sense.

crates/utils/src/cli.rs
```diff
@@ -46,13 +46,17 @@ pub fn check_private_key_length(private_key: &str) -> Result<B256, String> {
...
+    let network_spec = match network_string {
+        "mainnet" => MAINNET.clone(),
+        "angelfood" => ANGELFOOD.clone(),
+        _ => {
+            return Err(format!(
+                "Not a valid network: {network_string}, must be 'angelfood' or 'mainnet'"
+            ))
+        }
+    };
+    set_network_spec(network_spec.clone());
```
morph-dev, line 58: I don't think we should do this during parsing (I don't think there should be side effects).
Instead, I think we should explicitly do it as soon as it's finished parsing.

crates/subnetworks/state/src/validation/validator.rs
```diff
@@ -209,7 +209,8 @@ mod tests {
 
         let header_oracle = HeaderOracle {
```
morph-dev, line 210: nit: I would replace this entire block with:

```rust
let mut header_oracle = HeaderOracle::new();
header_oracle.history_jsonrpc_tx = Some(history_jsonrpc_tx);

```

testing/ethportal-peertest/src/scenarios/ping.rs
```diff
@@ -182,15 +182,6 @@ pub async fn test_ping_failed_to_decode_payload(target: &Client, peertest: &Peer
     assert!(error.contains("-39005"));
 }
 
-pub async fn test_ping_cross_network(mainnet_target: &Client, angelfood_node: &PeertestNode) {
```
morph-dev, line 185: I would be nice to have at least one end-to-end test that checks that ping from another network (i.e. angelfood) fails.

We can't do this easily with current NetworkSpec setup, but I think we can still do it if we manually encode the PING message (instead of using `Discovery`) and send it directly.

Don't know if this is easy and definitely not something I would expect you to do it in this PR.

I just wanted to mention it, and maybe we can create an issue to track it.

morph-dev (commented):
No big comments. Looks mostly ok.

---

Do we have any tests that change NetworkSpec to something else (this was the whole purpose of using RwLock instead of OnceLock)?

I believe that RwLock and how it's currently setup doesn't work well for most unit tests, because they usually run in parallel and most would expect that NetworkSpec has default value (and if one test changes it and it doesn't revert it at the end, other tests would fail).

It probably works better for peertests because they run one at the time, and they all explicitly set value at startup.

This is probably not big issue as of now and this is a great start. I just want to voice my opinion for the future.

Cargo.toml
```diff
@@ -38,6 +38,7 @@ version = "0.2.6"
 
 [workspace.dependencies]
 alloy = { version = "0.12", default-features = false, features = ["std", "serde"] }
+alloy-hardforks = "0.2.0"
```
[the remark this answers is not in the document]
KolbyML: I made a PR to bump all the versions https://github.com/ethereum/trin/pull/1794

crates/validation/src/oracle.rs
```diff
@@ -17,35 +17,24 @@ use ethportal_api::{
...
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct HeaderOracle {
     // We could simply store the main portal jsonrpc tx channel here, rather than each
     // individual channel. But my sense is that this will be more useful in terms of
     // determining which subnetworks are actually available.
     pub history_jsonrpc_tx: Option<mpsc::UnboundedSender<HistoryJsonRpcRequest>>,
     pub beacon_jsonrpc_tx: Option<mpsc::UnboundedSender<BeaconJsonRpcRequest>>,
     pub state_jsonrpc_tx: Option<mpsc::UnboundedSender<StateJsonRpcRequest>>,
-    pub header_validator: HeaderValidator,
```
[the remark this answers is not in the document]
KolbyML: I will make a new PR to move it then, because it doesn't belong here. It isn't used and it shouldn't be behind a RwLock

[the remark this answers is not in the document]
KolbyML: After this comment https://github.com/ethereum/trin/pull/1755#pullrequestreview-2762112442 I have been trying to make things work more well, but maybe the above just is just a little too offtopic

crates/validation/src/header_validator.rs
KolbyML: You asked me to move it back, so I will move it back https://github.com/ethereum/trin/pull/1780#discussion_r2065834655

crates/validation/src/oracle.rs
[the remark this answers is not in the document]
KolbyML: I guess this change is in my other PR anyways so I will just remove it from this PR for now

testing/ethportal-peertest/src/scenarios/ping.rs
KolbyML: > I would be nice to have at least one end-to-end test that checks that ping from another network (i.e. angelfood) fails.
> 
> We can't do this easily with current NetworkSpec setup, but I think we can still do it if we manually encode the PING message (instead of using `Discovery`) and send it directly.
> 
> Don't know if this is easy and definitely not something I would expect you to do it in this PR.
> 
> I just wanted to mention it, and maybe we can create an issue to track it.

Ok I will do it in a follow up PR then

crates/utils/src/cli.rs
KolbyML: Ok I moved the call for it to the main function

KolbyML:
> No big comments. Looks mostly ok.
> 
> Do we have any tests that change NetworkSpec to something else (this was the whole purpose of using RwLock instead of OnceLock)?
> 
> I believe that RwLock and how it's currently setup doesn't work well for most unit tests, because they usually run in parallel and most would expect that NetworkSpec has default value (and if one test changes it and it doesn't revert it at the end, other tests would fail).

`this was the whole purpose of using RwLock instead of OnceLock` the whole purpose was for ethportal-peertests no?

> It probably works better for peertests because they run one at the time, and they all explicitly set value at startup.
> 
> This is probably not big issue as of now and this is a great start. I just want to voice my opinion for the future.

:+1:

KolbyML:
@morph-dev ready for another review, if you want we can merge this pr first https://github.com/ethereum/trin/pull/1794 then merge this one, that will avoid the different alloy-primitive versions being used

crates/ethportal-api/src/types/network_spec.rs
```diff
@@ -12,19 +14,40 @@ use super::{
...
+
+pub fn network_spec() -> Arc<NetworkSpec> {
+    NETWORK_SPEC.read().clone()
+}
+
+/// `NetworkSpec` is a struct that contains the network configuration for the portal network and the
+/// respective Ethereum Network.
+///
+/// It includes the mapping of subnetworks to protocol id hex strings, supported protocol versions,
+/// and hardforks.
```
morph-dev, line 35:
```suggestion
/// It includes the mapping of subnetworks to protocol id hex strings, supported protocol versions,
/// and hardforks.
/// 
/// It should be initialized at the start of the application using [set_network_spec] and can be
/// accessed from anywhere using [network_spec]. The tests are exception to this case, as they
/// can set value at the start (if they test something other than mainnet data) and reset the value
/// at the end, while making sure that they are no other tests running in parallel (e.g. using
/// `serial_test` crate).
```

crates/ethportal-api/src/types/network_spec.rs
```diff
@@ -12,19 +14,40 @@ use super::{
     },
 };
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+static NETWORK_SPEC: LazyLock<RwLock<Arc<NetworkSpec>>> =
+    LazyLock::new(|| RwLock::new(MAINNET.clone()));
+
+/// `set_network_spec` should be called only once at the start of the application.
+///
+/// The only exception is for testing purposes, where it can be called multiple times.
```
morph-dev, line 22:
```suggestion
/// Should be called only once at the start of the application to initialize static [NetworkSpec].
///
/// The static `NetworkSpec` can be accessed using [network_spec].
///
/// Tests can also use this, but should be more careful. See [NetworkSpec] for details.
```

bin/e2hs-writer/src/reader.rs
```diff
@@ -241,15 +244,28 @@ impl EpochReader {
...
 
     pub fn iter_blocks(mut self) -> impl Iterator<Item = anyhow::Result<AllBlockData>> {
-        (self.starting_block..self.ending_block).map(move |current_block| match current_block {
-            0..MERGE_BLOCK_NUMBER => self.get_pre_merge_block_data(current_block),
-            MERGE_BLOCK_NUMBER..SHANGHAI_BLOCK_NUMBER => {
+        (self.starting_block..self.ending_block).map(move |current_block| {
+            if current_block
+                < EthereumHardfork::Paris
+                    .activation_block(network_spec().network().into())
+                    .expect("Paris is available")
```
morph-dev, line 251:
```suggestion
                    .expect("Paris should be available")
```

nit: If I would see the error/panic message "Paris is available", I would assume that is what caused the error (not what should be).

Same below.

crates/validation/src/header_validator.rs
```diff
@@ -519,6 +545,10 @@ mod test {
         EpochAccumulator::from_ssz_bytes(&epoch_acc_bytes).unwrap()
     }
 
+    fn slot_to_timestamp(slot: u64) -> u64 {
```
morph-dev, line 548: nit: This feels like useful function in general. Since it's also dependent of the chain, It would make sense to be in NetworkSpec as well. What do you think?

morph-dev (approved):
Looks great :fire: 

Thank you for being patient, with my reviews.

crates/validation/src/header_validator.rs
KolbyML: sure

testing/ethportal-peertest/src/scenarios/ping.rs
KolbyML: Here is an issue https://github.com/ethereum/trin/issues/1795

PR #1781, opened 2025-04-25
feat(bridge): run trin ourselves instead of other process using json-rpc

KolbyML:
### What was wrong?

- we are currently depending on a separate binary to run the bridge, this is annoying because if the Trin and Portal-bridge binary are out of sync you can get weird errors like if there is a change to ethportal-api, which is a common problem if you switch branches
- we are currently using HTTP JSON-RPC which is very slow for making all requests to trin which is extremely slow
- The beacon bridge test code isn't really testing anything, we already removed the history test code, I think we should remove it too.
- I need the performance gains for the state bridge, also a separate binary is bad UX. Ideally I don't want to need to add support for the census to use HTTP and HeaderOracle, like I did upstream in my Trin-Execution branch, this change will make it easier for me to downstream my changes

### How was it fixed?

- run Trin internally as a library instead of running a separately compiled trin binary
- remove beacon bridge test which really isn't testing much


crates/validation/src/oracle.rs
morph-dev, line 1: I really don't think that most of this functionality should be in structure called "HeaderOracle".

If you want, you can get `mpsc::UnboundedSender<HistoryJsonRpcRequest>` from the oracle and have it in a separate object that will let you send these requests.

bin/portal-bridge/src/handle.rs
```diff
@@ -1,60 +1,193 @@
...
-    }
-    if let Some(client_metrics_url) = bridge_config.client_metrics_url {
-        let url: String = client_metrics_url.to_string();
-        command.args(["--enable-metrics-with-url", &url]);
-    }
-
-    Ok(command.spawn()?)
+const APP_NAME: &str = "portal-bridge";
+
+pub async fn run_trin(bridge_config: &BridgeConfig) -> anyhow::Result<Arc<RwLock<HeaderOracle>>> {
```
morph-dev, line 31: Isn't this very similar to `run_trin` in bin/trin/src/lib.rs ? Can we refactor that one and make it reusable?

morph-dev (commented):
I reviewed entire PR quickly and I don't want all this functionality in HeaderOracle.

Also, if I'm not mistaken, we are still converting requests/responses to/from JSON, right? What do you think about getting `Arc<OverlayProtocol<...>>` (e.g. `history_handler.network.overlay.clone()`) and using it directly? Or do you think that is value in using high level API?

---

I agree that current bridge peertest isn't testing much. But I believe that we should have some bridge peertests in place. I don't expect this to be done in this PR. I don't expect this to be done in the next PR because it isn't a priority. But I believe we should make a issue to track it so we don't completely forget about it.

bin/portal-bridge/Cargo.toml
```diff
@@ -54,6 +53,17 @@ trin-utils.workspace = true
 trin-validation.workspace = true
 url.workspace = true
 
+# Trin dependencies
```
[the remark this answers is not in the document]
KolbyML: Ok I moved Trin crates under `Trin dependencies`, and moved the generic ones above

KolbyML:
> I reviewed entire PR quickly and I don't want all this functionality in HeaderOracle.
> 
> Also, if I'm not mistaken, we are still converting requests/responses to/from JSON, right? What do you think about getting `Arc<OverlayProtocol<...>>` (e.g. `history_handler.network.overlay.clone()`) and using it directly? Or do you think that is value in using high level API?
 
I think this is a great idea :rocket: 

> I agree that current bridge peertest isn't testing much. But I believe that we should have some bridge peertests in place. I don't expect this to be done in this PR. I don't expect this to be done in the next PR because it isn't a priority. But I believe we should make a issue to track it so we don't completely forget about it.

Sure I can make an issue

crates/validation/src/oracle.rs
KolbyML: Removed it

bin/portal-bridge/src/handle.rs
KolbyML: Done

KolbyML:

KolbyML:
I made an issue https://github.com/ethereum/trin/issues/1793

bin/trin/src/lib.rs
```diff
@@ -3,42 +3,37 @@
...
+use trin_history::{initialize_history_network, network::HistoryNetwork};
+use trin_state::{initialize_state_network, network::StateNetwork};
+use trin_storage::{config::StorageCapacityConfig, PortalStorageConfigFactory};
 #[cfg(windows)]
 use trin_utils::cli::Web3TransportType;
 use trin_validation::oracle::HeaderOracle;
 use utp_rs::socket::UtpSocket;
 
-pub async fn run_trin(
+pub async fn run_trin_from_trin_config(
```
morph-dev, line 34: Seeing that many places use this function and then have to extract and handle `Optional` for rpc_server_handle, how about we do following:

- `TrinHandle::rpc_server_handle` would no longer be `Optional`
- rename `run_trin` to `ruin_trin_internal` (maybe even move it in a separate file)
    - make it non public
    - change return type to `(SubnetworkOverlays, Optional<RpcServerHandle>)`
- create new `run_trin` that doesn't accept `RpcConfig` and only returns `SubnetworkOverlays`
- create new `run_trin_with_rpc` that accepts `RpcConfig` and returns `TrinHandle`
    - `run_trin_from_trin_config` would use `run_trin_with_rpc`

What do you think?

bin/trin/src/lib.rs
```diff
@@ -146,23 +170,38 @@ pub async fn run_trin(
...
+            )
+            .await?,
+        )
+    } else {
+        None
+    };
+
+    let mut trin_handle = TrinHandle {
+        rpc_server_handle,
+        ..Default::default()
```
morph-dev, line 192: Why not create SubnetworkOverlays directly:
```rust
SubnetworkOverlays {
    history: history_handler.as_ref().map(|handler| handler.network.clone()),
    // ...
}
```

bin/portal-bridge/src/bridge/beacon.rs
```diff
@@ -94,7 +91,7 @@ impl FinalizedBootstrap {
 pub struct BeaconBridge {
     consensus_api: ConsensusApi,
     mode: BridgeMode,
-    portal_client: HttpClient,
+    portal_client: SubnetworkOverlays,
```
morph-dev, line 94: I don't think BeaconBridge should have `SubnetworkOverlays` (as it's annoying to deal with optional).

We would still pass `SubnetworkOverlays` in constructor, but it should only extract and keep `beacon_network: Arc<BeaconNetwork>` (if that's all it needs).

Same for other bridges.

bin/portal-bridge/src/census/network.rs
```diff
@@ -70,15 +71,19 @@ impl Default for NetworkInitializationConfig {
 ///
 /// The [Network::init] should be used to initialize our view of the network, and [NetworkManager]
 /// should be used in a background task to keep it up-to-date.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub(super) struct Network {
     peers: Peers<AdditiveWeight>,
-    client: HttpClient,
+    client: SubnetworkOverlays,
```
morph-dev, line 77: We should probably do similar thing here (just keep one network itself), but it's not as easy...
We could probably use `OverlayProtocol` type, but I think we would have to introduce 5 generics (which probably isn't that bad). What do you think?

If you decide that it's too ugly/too much work, we could at least do this:

```rust
pub fn get_overlay_protocol(&self) -> Arc<OverlayProtocol> {
    match self.subnetwork {
        Subnetwork::History => {
            self.client
                .history_overlay()
                .expect(...)
                .overlay
                .clone()
        // ...
    }
}
```
This would make ping, find_nodes and recursive_find_nodes functions almost not needed.

---

Also, "client" name makes no sense anymore.

bin/portal-bridge/src/cli.rs
```diff
@@ -182,6 +181,23 @@ pub struct BridgeConfig {
     pub data_dir: Option<PathBuf>,
 }
 
+impl BridgeConfig {
+    pub fn to_node_runtime_config(&self, node_data_dir: PathBuf) -> NodeRuntimeConfig {
```
morph-dev, line 185: nit: maybe `as_node_runtime_config` ? The "to" and "into" usually take ownership of `self` while `as_` don't.

Of just implement `impl From<&BridgeConfig> for NodeRuntimeConfig`.

bin/trin/src/cli.rs
```diff
@@ -381,6 +383,32 @@ impl TrinConfig {
             },
         }
     }
+
+    pub fn to_portalnet_config(&self, private_key: B256) -> PortalnetConfig {
```
morph-dev, line 387: nit: Also here, maybe `as_portalnet_config`?

bin/trin/src/cli.rs
```diff
@@ -381,6 +383,32 @@ impl TrinConfig {
...
+            no_stun: self.no_stun,
+            no_upnp: self.no_upnp,
+            discv5_session_cache_capacity: DISCV5_SESSION_CACHE_CAPACITY,
+            disable_poke: self.disable_poke,
+            trusted_block_root: self.trusted_block_root,
+            utp_transfer_limit: self.utp_transfer_limit,
+        }
+    }
+
+    pub fn to_node_runtime_config(&self, node_data_dir: PathBuf) -> NodeRuntimeConfig {
```
morph-dev, line 402: nit: and here

morph-dev (commented):
Looks much better.
Several suggestions, nothing too major.

bin/portal-bridge/Cargo.toml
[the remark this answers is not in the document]
KolbyML: > Why not keep them together?
> 
> If you want to propose that crates from this repo should be separated from external crates, then we should do it for all crates, not just bridge (and that should definitely be separate PR). I personally don't see much value in doing this and don't think that we should do it, but I wouldn't oppose it.

It is easier to what what crates are ours

> Also, why is `utp-rs` here and `eth_trie` is not?

accident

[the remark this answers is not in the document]
KolbyML: I will revert it for now I guess

bin/portal-bridge/src/cli.rs
KolbyML: > nit: maybe `as_node_runtime_config` ? The "to" and "into" usually take ownership of `self` while `as_` don't.

sure

> Of just implement `impl From<&BridgeConfig> for NodeRuntimeConfig`.

I am passing in `node_data_dir` so I can't use the from trait

bin/trin/src/lib.rs
KolbyML: Sure sounds good

bin/portal-bridge/src/bridge/beacon.rs
KolbyML: I passed in the respective network into the constructor, as `SubnetworkOverlays` is a bit to broad I feel for the constructor, we might as well fully commit if we go this path. Which I agree it is good

bin/portal-bridge/src/census/network.rs
KolbyML: I ended up just moving those functions into the `SubnetworkOverlays` struct
image

5 generics is way too much, I would like to avoid it if possible

KolbyML:

morph-dev (commented):
Almost there. Few small comment.

bin/trin/src/handle.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub state: Option<Arc<StateNetwork>>,
+    pub beacon: Option<Arc<BeaconNetwork>>,
+}
+
+impl SubnetworkOverlays {
+    pub fn history_overlay(&self) -> anyhow::Result<Arc<HistoryNetwork>> {
+        match &self.history {
+            Some(overlay) => Ok(overlay.clone()),
+            None => bail!("History network is not available"),
+        }
```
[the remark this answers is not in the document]
KolbyML: I will name it history, as I don't want to return `OverlayProtocol` due to the generics

bin/trin/src/builder.rs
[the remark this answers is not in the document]
KolbyML: I ended up just re-exporting the 3

bin/trin/src/lib.rs
```diff
@@ -1,194 +1,6 @@
 #![warn(clippy::unwrap_used)]
```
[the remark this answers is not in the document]
KolbyML: I talked to Milos and dm's and there was just some confusion on what this did. He said to ignore this comment

KolbyML:

bin/trin/src/lib.rs
```diff
@@ -1,6 +1,8 @@
...
 #![warn(clippy::uninlined_format_args)]
 
-pub mod builder;
 pub mod cli;
 pub mod handle;
+pub mod run;
```
morph-dev, line 6: Since we are re-exporting all these functions, no need for this module to be public.

To be fair, we can re-export all public structures/functions from this crate, as there aren't that many of them. We moved them into separate files only because this one was growing big.
It makes sense for other crates that want to use these types/functions, to do so at the root (e.g. `trin::TrinConfig` instead of `trin::cli::TrinConfig`).

morph-dev (approved):
LGTM :rocket:

bin/trin/src/lib.rs
KolbyML: Sure, done

PR #1785, opened 2025-04-28
feat: add Deneb header proof type and generation

morph-dev:
### What was wrong?

We don't have Deneb header proof, generation and validation.

### How was it fixed?

I created Deneb header proof type and generation code.

I also added support for it in e2hs binary.

### To-Do

### DIFF
--- a/bin/e2hs-writer/src/utils.rs
+++ b/bin/e2hs-writer/src/utils.rs
@@ -10,6 +10,7 @@ use alloy::{
 };
 use anyhow::{anyhow, ensure};
 use ethportal_api::{
+    consensus::execution_payload::ExecutionPayloadDeneb,
     types::{
         consensus::execution_payload::{ExecutionPayloadBellatrix, ExecutionPayloadCapella},
         execution::accumulator::EpochAccumulator,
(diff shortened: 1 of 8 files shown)


bin/e2hs-writer/src/utils.rs
```diff
@@ -94,6 +95,44 @@ pub fn pre_deneb_execution_payload_to_header(
...
+        parent_beacon_block_root: None,
+        requests_hash: None,
+    };
+
+    ensure!(
+        payload.block_hash == header.hash_slow(),
+        "Block hash mismatch"
+    );
+    Ok(header)
+}
```
KolbyML, line 134: This code is correctly generating the wrong `block_hash` due to missing the `parent_beacon_block_root` field, you can reference this code https://github.com/ethereum/trin/blob/07f7bd09d1022750573a5864468e577a68f8bc50/bin/trin-execution/src/era/beacon.rs#L125-L168 to see where you can find this root

morph-dev: Good catch!
I remember that I checked this, and I though it was introduced in Pectra. But I checked again and it's actually introduced in Dencun.

ogenev (approved):
LGTM ⛵

KolbyML (approved):
:shipit: looks good

PR #1786, opened 2025-04-28
feat: add Deneb header validation

morph-dev:
This is followup of #1785 , which ~should be reviewed and merged first~ is now merged.

### What was wrong?

We are missing validation for Deneb header proofs.

### How was it fixed?

Add proof validation.

Also several improvements  in tests:
- group tests per fork
- split valid and invalid test cases
- more descriptive test  names 
- extract logic for reading and deserializing/decoding files from portal_spec_tests submodule

### To-Do


morph-dev:
This PR is ready for review.

crates/validation/src/header_validator.rs
```diff
@@ -182,6 +182,50 @@ impl HeaderValidator {
...
+            gen_index as usize,
+            historical_summary,
+        ) {
+            return Err(anyhow!(
+                "Merkle proof validation failed for BlockProofHistoricalSummariesDeneb"
+            ));
+        }
+
+        Ok(())
+    }
```
KolbyML, line 227: It seems like the only real difference is `verify_bellatrix_to_deneb_beacon_block_proof` ` and `verify_post_deneb_beacon_block_proof` and the timestamp fork boundary checks above, I am fine with this level code duplication. I am a bit surprised though :grin:

morph-dev: You are absolutely correct :smiley: . While at it, I renamed other functions and made some changes to error messages.

crates/validation/src/header_validator.rs
```diff
@@ -96,37 +102,34 @@ impl HeaderValidator {
...
-
-        let block_root_index = proof.slot % EPOCH_SIZE;
-        let gen_index = 2 * EPOCH_SIZE + block_root_index;
-        let historical_root_index = proof.slot / EPOCH_SIZE;
-        let historical_root =
-            self.historical_roots_acc.historical_roots[historical_root_index as usize];
+        ) {
+            return Err(anyhow!(
+                "Execution block proof verification failed for Merge-Capella header"
+            ));
```
KolbyML, line 122:
```suggestion
            bail!("Execution block proof verification failed for Merge-Capella header");
```

Couldn't we just use `bail!(` instead of `return Err(anyhow!(` everywhere?

crates/validation/src/header_validator.rs
```diff
@@ -148,51 +142,77 @@ impl HeaderValidator {
...
+        ) {
+            bail!("Beacon block proof verification failed for post-Deneb header");
+        }
+
+        Ok(())
+    }
+
+    /// Verify that the execution block header is included in Bellatrix/Capella beacon block
+    #[must_use]
+    fn verify_bellatrix_to_deneb_execution_block_proo(
```
KolbyML, line 210:
```suggestion
    fn verify_bellatrix_to_deneb_execution_block_proof(
```

missing an f

crates/validation/src/header_validator.rs
```diff
@@ -207,26 +227,92 @@ impl HeaderValidator {
...
+        let historical_summary_index = (slot - CAPELLA_FORK_EPOCH * SLOTS_PER_EPOCH) / EPOCH_SIZE;
+        let historical_summary =
+            historical_summaries[historical_summary_index as usize].block_summary_root;
+
+        verify_merkle_proof(
+            beacon_block_root,
+            beacon_block_proof,
+            13,
+            gen_index as usize,
+            historical_summary,
```
KolbyML, line 313:
```suggestion
        let block_summary_root =
            historical_summaries[historical_summary_index as usize].block_summary_root;

        verify_merkle_proof(
            beacon_block_root,
            beacon_block_proof,
            13,
            gen_index as usize,
            block_summary_root,
```

maybe we should just call that variable `block_summary_root` as that is what it is

crates/validation/src/header_validator.rs
```diff
@@ -248,261 +334,359 @@ mod test {
...
+        #[case::block_number_1_000_006(1_000_006)]
+        #[case::block_number_1_000_007(1_000_007)]
+        #[case::block_number_1_000_008(1_000_008)]
+        #[case::block_number_1_000_009(1_000_009)]
+        #[case::block_number_1_000_010(1_000_010)]
+        #[tokio::test]
+        async fn generate_and_verify_header_with_proofs(#[case] block_number: u64) {
+            // Use fluffy's proofs as test data to validate that trin
+            // - generates proofs which match fluffy's
+            // - validates hwps
```
KolbyML, line 369:
```suggestion
        async fn generate_and_verify_header_with_proofs(#[case] block_number: u64) {
            // Use fluffy's proofs as test data to validate that trin
            // - generates proofs which match fluffy's
            // - validates header_with_proof's
```

It would be nice if we avoided the short hand in the comments at least

KolbyML (approved):
:shipit: looks good

crates/validation/src/header_validator.rs
morph-dev: I will do this:

```rust
let historical_summary = &historical_summaries[historical_summary_index as usize];
verify_merkle_proof(
    beacon_block_root,
    beacon_block_proof,
    13,
    gen_index as usize,
    historical_summary.block_summary_root,
)
```

Both name is correct and I think it's clearer (to me, block_summary_root is a bit vague).

crates/validation/src/header_validator.rs
morph-dev: I ended up refactoring all pre-merge tests:
- no more "hwp"
- proof generation part of the tests is moved into `accumulator.rs`
- removed no longer needed asset file and some constants

KolbyML (approved):
:shipit: looks good

PR #1794, opened 2025-04-29
chore: bump alloy/revm/ssz

KolbyML:
### What was wrong?
- We want to avoid importing multiple versions of alloy crates
- I want to use alloy-hardforks a crate which was made 3 months ago, which implements all the Ethereum Fork stuff
- We use a quite old version of alloy and revm, and ssz, all of which is required to be upgraded as they depend on each other

### How was it fixed?

Bumping the versions of
- alloy
- revm
- ssz
- ckzg

I also had to update some code to match the upstream changes

### DIFF
--- a/crates/evm/src/tx_env_modifier.rs
+++ b/crates/evm/src/tx_env_modifier.rs
@@ -129,7 +129,7 @@ impl TxEnvModifier for TransactionRequest {
         }
         tx_env.gas_priority_fee = self.max_priority_fee_per_gas;
         if let Some(max_fee_per_blob_gas) = self.max_fee_per_blob_gas {
-            tx_env.gas_price = max_fee_per_blob_gas;
+            tx_env.max_fee_per_blob_gas = max_fee_per_blob_gas;
         }
         if let Some(gas) = self.gas {
             tx_env.gas_limit = gas;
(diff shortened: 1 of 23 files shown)


crates/evm/src/tx_env_modifier.rs
```diff
@@ -115,16 +119,18 @@ impl TxEnvModifier for TransactionRequest {
             tx_env.caller = from;
         }
         if let Some(to) = self.to {
-            tx_env.transact_to = to;
+            tx_env.kind = to;
         }
         if let Some(gas_price) = self.gas_price {
-            tx_env.gas_price = U256::from(gas_price);
+            tx_env.gas_price = gas_price;
```
morph-dev, line 125: nit: I noticed that we are potentially setting `tx_env.gas_price` multiple times, more precisely if more than one of `self.gas_price, self.max_fee_per_gas and self.max_fee_per_blob_gas` are set.

Can you explain why this is correct?

KolbyML: max_fee_per_gas is a replacement for gas_price in later transaction types

but yeah max_fee_per_blob_gas should equal max_fee_per_blob_gas, not gas_price. good catch

PR #1797, opened 2025-05-01
feat: update portal-spec-tests and refactor reading from test files

morph-dev:
### What was wrong?

We were missing test vectors around fork boundaries. These were added to portal-spec-tests in ethereum/portal-spec-tests#49

### How was it fixed?

Updated portal-spec-tests submodule, and simplified/refactored some tests in ehtportal-api crate.

### To-Do

### DIFF
--- a/crates/ethportal-api/src/test_utils/mod.rs
+++ b/crates/ethportal-api/src/test_utils/mod.rs
@@ -1,18 +1,52 @@
 use std::{
-    fs, io,
+    fs::{self, File},
+    io::{self, BufReader},
     path::{Path, PathBuf},
 };
+
+use anyhow::anyhow;
+use serde::de::DeserializeOwned;
+use ssz::Decode;
 
 use self::constants::PORTAL_SPEC_TESTS_SUBMODULE_PATH;
 
 pub mod constants;
+pub mod types;
 
-/// Reads a file from a "portal-spec-tests" submodule.
-pub fn read_file_from_tests_submodule<P: AsRef<Path>>(path: P) -> io::Result<String> {
-    fs::read_to_string(PathBuf::from(PORTAL_SPEC_TESTS_SUBMODULE_PATH).join(path))
+fn portal_spec_tests_path(path: impl AsRef<Path>) -> PathBuf {
+    PathBuf::from(PORTAL_SPEC_TESTS_SUBMODULE_PATH).join(path)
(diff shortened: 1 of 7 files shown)


crates/ethportal-api/src/test_utils/types.rs
```diff
@@ -0,0 +1,37 @@
...
+    pub content_key: K,
+    #[serde(rename = "content_value")]
+    pub raw_content_value: RawContentValue,
+}
+
+impl<K: OverlayContentKey> ContentItem<K> {
+    pub fn content_value<V: ContentValue<TContentKey = K>>(&self) -> Result<V, ContentValueError> {
+        V::decode(&self.content_key, &self.raw_content_value)
+    }
+}
```
KolbyML, line 20: Isn't this type already defined in `crates/utils/src/testing.rs`, shouldn't we only define it once?

crates/ethportal-api/src/test_utils/mod.rs
```diff
@@ -1,18 +1,52 @@
...
+
+/// Reads ssz file from a "portal-spec-tests" submodule
+pub fn read_ssz_portal_spec_tests_file<T: Decode>(path: impl AsRef<Path>) -> anyhow::Result<T> {
+    let bytes = read_binary_portal_spec_tests_file(&path)?;
+    T::from_ssz_bytes(&bytes).map_err(|err| {
+        anyhow!(
+            "Error decoding ssz file: {}. Error: {err:?}",
+            path.as_ref().display()
+        )
+    })
```
KolbyML, line 51: Isn't this code already defined in `crates/utils/src/submodules.rs` shouldn't we only define it once?

KolbyML (commented):
Most of the PR looks good, I just have questions on the types we already have defined

KolbyML (approved):
:shipit: looks good. Milos messaged me on discord for context

morph-dev:
> :shipit: looks good. Milos messaged me on discord for context

For context for others, these files and logic is duplicated because we don't want to put it in `ethportal-api` because we publish this crate. And `ethportal-api` can't depend on `crates/utils` because of circular dependency.  So the setup is:
- we have module in `ethportal-api` that is protected by `#[cfg(test)]` that is used only within that crate
- we have `crates/utils` for all other crates

PR #1799, opened 2025-05-02
feat: add electra types

morph-dev:
### What was wrong?

We don't have any Pectra types defined.

### How was it fixed?

I rebased #1763 and added few more things on top (rebase is first commit, mine is second):
- Added proper (I think) fork digest
    - I added test that checks that all fork digests are as expected (had to move some stuff around)
- Added pectra fork as part of the config
- Added more types to ef-tests
- Few other smaller changes

Planned work for the followup PRs:
- Add functions for creating `HeaderWithProof` based on Pectra types
- Add support for Pectra for HistoricalSummariesWithProof (see ethereum/portal-network-specs#397)
- Investigate if we have to update anything regarding validating beacon content post Pectra

### To-Do


KolbyML:
@morph-dev please look at my review of the previous PR here, https://github.com/ethereum/trin/pull/1763#discussion_r2046599179

```
impl ProcessBeaconBlock for SignedBeaconBlockElectra {
    fn process_beacon_block(&self) -> anyhow::Result<ProcessedBlock> {
```

is generating invalid execution headers

morph-dev:
> @morph-dev please look at my review of the previous PR here, [#1763 (comment)](https://github.com/ethereum/trin/pull/1763#discussion_r2046599179)

Yes, that's what I was planning on doing.

However, requests inside BeaconBlockBody are grouped by type, and I'm not sure in which order they should be iterated for execution header. I assume it would be "deposit requests" then "withdrawals requests" then "consolidations requests", but I'm not 100% sure.

Will look into it tomorrow...

KolbyML (approved):
I haven't verified if `ForkName::Electra => [0xad, 0x53, 0x2c, 0xeb]` is correct but other then that the PR looks good

:shipit:

PR #1800, opened 2025-05-02
feat: create header proof using BeaconBlockElectra

morph-dev:
This is a followup of #1799

### What was wrong?

We are missing logic for creating header proof using `BeaconBlockElectra`.

### How was it fixed?

Added function for creating proof.
Note that we don't have any test vectors to verify if this works as intended.

### To-Do

### DIFF
--- a/bin/e2hs-writer/src/reader.rs
+++ b/bin/e2hs-writer/src/reader.rs
@@ -5,9 +5,16 @@ use alloy::{
     eips::eip4895::{Withdrawal, Withdrawals},
     primitives::B256,
 };
-use alloy_hardforks::{EthereumHardfork, EthereumHardforks};
+use alloy_hardforks::EthereumHardforks;
 use anyhow::{anyhow, bail, ensure};
 use ethportal_api::{
+    consensus::{
+        beacon_block::{
+            BeaconBlockBellatrix, BeaconBlockCapella, BeaconBlockDeneb, BeaconBlockElectra,
+            SignedBeaconBlock,
+        },
+        beacon_state::HistoricalBatch,
+    },
     types::{
         execution::{
             accumulator::EpochAccumulator,
@@ -32,8 +39,7 @@ use crate::{
     provider::EraProvider,
     utils::{
         bellatrix_execution_payload_to_header, capella_execution_payload_to_header,
-        lookup_epoch_acc, post_deneb_execution_payload_to_header,
-        post_electra_execution_payload_to_header,
+        deneb_execution_payload_to_header, electra_execution_payload_to_header, lookup_epoch_acc,
     },
 };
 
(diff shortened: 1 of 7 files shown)


bin/e2hs-writer/src/reader.rs
```diff
@@ -241,6 +242,48 @@ impl EpochReader {
         })
     }
 
+    #[allow(dead_code)]
+    fn get_electra_block_data(&mut self, block_number: u64) -> anyhow::Result<AllBlockData> {
```
morph-dev, line 246: This logic is triggered after processing era file. Do you think this will be needed?
Also, at the moment, we decide which fork to use based on block number, and we don't have one for Pectra yet.

KolbyML: We have the timestamp it will be enabled already

bin/e2hs-writer/src/utils.rs
```diff
@@ -134,6 +137,46 @@ pub fn post_deneb_execution_payload_to_header(
...
+        timestamp: payload.timestamp,
+        extra_data: payload.extra_data.to_vec().into(),
+        mix_hash: payload.prev_randao,
+        nonce: B64::ZERO,
+        base_fee_per_gas: Some(payload.base_fee_per_gas.to()),
+        withdrawals_root: Some(withdrawals_root),
+        blob_gas_used: Some(payload.blob_gas_used),
+        excess_blob_gas: Some(payload.excess_blob_gas),
+        parent_beacon_block_root: Some(parent_beacon_block_root),
+        requests_hash: None, // TODO
```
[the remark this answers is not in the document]
KolbyML: https://github.com/ethereum/trin/pull/1799#issuecomment-2848163541

or just read https://eips.ethereum.org/EIPS/eip-7685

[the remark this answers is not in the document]
KolbyML: You can just generate a dummy header from reth or geth

bin/e2hs-writer/src/reader.rs
KolbyML: There are post merge and pre-merge blocks, all post merge blocks use ERA, I don't think this current loop makes much sense tbh

KolbyML: We could do a match case on what CL fork the BeaconBlock is for, instead of blindly parsing it like we are now

morph-dev: I refactored how it works.
I no longer check what fork we are on based on number, but rather based on variant of SignedBeaconBlock.
If you think it would be good, I can also add a check to make sure that variant is at the expected fork.

morph-dev:
@KolbyML, ready for a review.

bin/e2hs-writer/src/reader.rs
KolbyML: Yeah the changes look good.

> If you think it would be good, I can also add a check to make sure that variant is at the expected fork.
I think this would be overkill to be honest, but either way it isn't like we are using hardcoded values anymore which is nice

KolbyML (approved):
:shipit: looks good

PR #1801, opened 2025-05-03
feat: update beacon content types to support Pectra

morph-dev:
### What was wrong?

The Beacon content types should be updated in order to support Pectra. 
See ethereum/portal-network-specs#397 for details.

### How was it fixed?

Modified existing type to support Pectra.

I had to update and remove all usage of the old type. That involved rewriting tests (I also added more test cases), which was tricky at times we now have different test data (I updated `portal-spec-tests` submodule)

### Followup work

1. Update beacon bridge to be able to create and gossip new type as well
    - This requires some refactoring in how we fetch BeaconState and I want to do it in a separate PR
2. Investigate if something else should be updated (either in `light-client` or `trin-beacon` crates)

### To-Do


KolbyML:
> * I want to match it with ForkDigest (and ForkName), so Deneb makes it consistent

I am not sure I understand this as the HistoricalSummariesWithProof format isn't likely to change in the future past Electra. So `HistoricalSummariesWithProofElectra` would encompass all future forks, hence I don't get the need to call `HistoricalSummariesWithProof` `HistoricalSummariesWithProofDeneb` when it isn't specific to Deneb and there isn't a consistent trend in future forks

KolbyML:
Isn't Electra hitting mainnet on May 7th so in 3 days? Shouldn't we just remove support for `HistoricalSummariesProofDeneb` right now, as we will need to remove it regardless on Wednesday? 

I don't see a reason to add support for both types, when we will just be removing the old type on Wednesday, we might as only only support `HistoricalSummariesProofElectra` in this PR, as unlike other types `HistoricalSummariesProof` is ephemeral

> Worth mentioning that the old type will most likely go away from the spec after the fork as we won't use it anymore.
^ which was also stated in the PR's description

KolbyML:
I think we should just replace our current `HistoricalSummariesWithProofDeneb` with the new format `HistoricalSummariesWithProofElectra` and call it a day, I don't see much value in us supporting both types for 2-3 days.

KolbyML:
image

shouldn't we enable pectra for all the other RPC endpoints? or do you still plan to make a separate PR for this?

morph-dev:
> I think we should just replace our current `HistoricalSummariesWithProofDeneb` with the new format `HistoricalSummariesWithProofElectra` and call it a day, I don't see much value in us supporting both types for 2-3 days. Adding complexity to handle both cases just to remove it in 2-3 days, doesn't make much sense to me.

I did as suggested.

It did cause complexity in refactoring tests. I removed BeaconState that we used for tests before and now I'm using test vector that I added in https://github.com/ethereum/portal-spec-tests/pull/51.

morph-dev:
> shouldn't we enable pectra for all the other RPC endpoints? or do you still plan to make a separate PR for this?

Separate PR

crates/ethportal-api/src/types/consensus/historical_summaries.rs
```diff
@@ -16,12 +16,15 @@ pub struct HistoricalSummary {
 }
 
 pub type HistoricalSummaries = VariableList<HistoricalSummary, typenum::U16777216>;
-pub type HistoricalSummariesStateProof = FixedVector<B256, typenum::U5>;
+
+pub const HISTORICAL_SUMMARIES_GINDEX: usize = 91;
+
+pub type HistoricalSummariesProof = FixedVector<B256, typenum::U6>;
```
KolbyML, line 22: Shouldn't the types be grouped together, and the constant should be defined at the top of the file, below the imports. no?

crates/ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -482,28 +482,23 @@ pub struct ForkVersionedHistoricalSummariesWithProof {
...
+    fn is_ssz_fixed_len() -> bool {
+        false
+    }
 
-        let mut data = fork_digest.to_vec();
-        data.extend(self.historical_summaries_with_proof.as_ssz_bytes());
-        data
+    fn ssz_append(&self, buf: &mut Vec<u8>) {
+        self.fork_name.as_fork_digest().ssz_append(buf);
+        self.historical_summaries_with_proof.ssz_append(buf);
```
KolbyML, line 492:
```suggestion
    fn ssz_append(&self, buf: &mut Vec<u8>) {
        buf.extend(&self.fork_name.as_fork_digest());
        self.historical_summaries_with_proof.ssz_append(buf);
```

the fork digest is just supposed to be appended not ssz encoded right? or do this encode to the same thing or something?

crates/ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -482,28 +482,23 @@ pub struct ForkVersionedHistoricalSummariesWithProof {
...
+        ssz::BYTES_PER_LENGTH_OFFSET
+    }
 
-        Ok(Self {
-            fork_name,
-            historical_summaries_with_proof: summaries_with_proof,
-        })
+    fn ssz_bytes_len(&self) -> usize {
+        self.fork_name.as_fork_digest().ssz_bytes_len()
+            + self.historical_summaries_with_proof.ssz_bytes_len()
```
KolbyML, line 501:
```suggestion
    fn ssz_bytes_len(&self) -> usize {
        4 + self.historical_summaries_with_proof.ssz_bytes_len()
```

crates/ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -512,22 +507,33 @@ impl Decode for ForkVersionedHistoricalSummariesWithProof {
...
+            return Err(DecodeError::InvalidByteLength {
+                len: bytes.len(),
+                expected: fork_digest_len,
+            });
+        };
 
-    fn ssz_append(&self, buf: &mut Vec<u8>) {
-        buf.extend_from_slice(&self.encode());
-    }
+        let fork_digest = ForkDigest::from_ssz_bytes(fork_digest_bytes)?;
```
KolbyML, line 525:
```suggestion
        let fork_digest = ForkDigest::try_from(fork_digest_bytes)?;
```

this shouldn't be ssz no?

crates/ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -512,22 +507,33 @@ impl Decode for ForkVersionedHistoricalSummariesWithProof {
...
-}
 
-impl Encode for ForkVersionedHistoricalSummariesWithProof {
-    fn is_ssz_fixed_len() -> bool {
-        false
-    }
+    fn from_ssz_bytes(bytes: &[u8]) -> Result<Self, DecodeError> {
+        let fork_digest_len = <ForkDigest as Decode>::ssz_fixed_len();
+        let Some((fork_digest_bytes, historical_summaries_with_proof_bytes)) =
+            bytes.split_at_checked(fork_digest_len)
```
KolbyML, line 517:
```suggestion
        let Some((fork_digest_bytes, historical_summaries_with_proof_bytes)) =
            bytes.split_at_checked(4)
```

KolbyML (commented):
Most of the PR looks good. I was am a little confused why we changed the code to encode the fork_digest using ssz, I thought we were supposed to should set it as the pre-fix. I didn't see anything in the PR description about there being a bug with how we did it before.

So I am a little confused? Maybe it encodes to the same output? Maybe I am missing something?

crates/ethportal-api/src/types/consensus/historical_summaries.rs
morph-dev: I wanted `HISTORICAL_SUMMARIES_GINDEX` and `HistoricalSummariesProof` together as they are related (that's the index that we use when we verify content).

The other two types are also related (one is VariableList of the other).

So for me, it makes sense to group them like this. Unless we have some rule about constants going first after imports (which I didn't know about).

Let me know if you think I should change it.

KolbyML:
```
pub const HISTORICAL_SUMMARIES_GINDEX: usize = 91;

pub type HistoricalSummariesProof = FixedVector<B256, typenum::U6>;
pub type HistoricalSummaries = VariableList<HistoricalSummary, typenum::U16777216>;
```

Could we rearrange it like this? I am fine if the const isn't at the top of the file. Ideally the types are connected though I would prefer if `HistoricalSummaries`, was over `HistoricalSummariesProof`

```
pub type HistoricalSummaries = VariableList<HistoricalSummary, typenum::U16777216>;
pub type HistoricalSummariesProof = FixedVector<B256, typenum::U6>;

pub const HISTORICAL_SUMMARIES_GINDEX: usize = 91;
```

Ideally it is formated like this then ^, the type definition are grouped and the const is below.

Idk let me know what you think. I just find it weird the const is standwiched by type definitions.

morph-dev:
> Most of the PR looks good. I was am a little confused why we changed the code to encode the fork_digest using ssz, I thought we were supposed to should set it as the pre-fix. I didn't see anything in the PR description about there being a bug with how we did it before.
> 
> So I am a little confused? Maybe it encodes to the same output? Maybe I am missing something?

The ssz encoding and actual ForkDigest are the same, so it doesn't make a difference.
But you are right, it shouldn't be ssz encoded/decoded (even if that doesn't do anything).

I can see how it can easily cause confusion, so I removed it.

crates/ethportal-api/src/types/content_value/beacon.rs
```diff
@@ -512,22 +507,33 @@ impl Decode for ForkVersionedHistoricalSummariesWithProof {
...
 
-    fn ssz_append(&self, buf: &mut Vec<u8>) {
-        buf.extend_from_slice(&self.encode());
-    }
+        let fork_digest = ForkDigest::try_from(fork_digest_bytes)
+            .map_err(|err| DecodeError::BytesInvalid(err.to_string()))?;
+        let fork_name = match ForkName::try_from(fork_digest) {
+            Ok(fork_name) => fork_name,
+            Err(err) => return Err(DecodeError::BytesInvalid(err.to_string())),
+        };
```
KolbyML, line 529: It might be nice to be consistent with `match`/`map_err` usage here

KolbyML (approved):
:shipit: looks good

PR #1806, opened 2025-05-05
fix: update bootnodes

morph-dev:
### What was wrong?

Our bootnodes are outdated and they contain v0 protocol version.

This would usually not be a problem, but they store beacon content and they are the first one that we contact during light-client sync (because we start light-client syncing before everything else, so they are the only one we know about at that point), and we assume protocol v0 version and fail to decode the content.

Technically, this shouldn't happen even with this setup, as we should get new Enr when discv5 session is established.
I didn't investigate but I suspect that "Request" object still has old Enr and that's what it uses to decode the response.

### How was it fixed?

Updated bootnode Enrs according to glados.

### To-Do

### DIFF
--- a/bin/trin/src/cli.rs
+++ b/bin/trin/src/cli.rs
@@ -1097,7 +1097,7 @@ mod tests {
             let config = TrinConfig::new_from([APP_NAME]).unwrap();
             assert_eq!(config.bootnodes, Bootnodes::Default);
             let bootnodes: Vec<Enr> = config.bootnodes.to_enrs(Network::Mainnet);
-            assert_eq!(bootnodes.len(), 11);
+            assert_eq!(bootnodes.len(), 10);
         }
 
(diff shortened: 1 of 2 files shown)


KolbyML:
> ### What was wrong?
> Our bootnodes are outdated and they contain v0 protocol version.
> 
> This would usually not be a problem, but they store beacon content and receiving LightClientBootstrap from them fail (because we assume protocol v0).

I am confused, in version 1 we didn't change anything with LightClientBootstraps? Also we support version 0 and version 1.

KolbyML (approved):
:shipit: looks good

morph-dev:
> I am confused, in version 1 we didn't change anything with LightClientBootstraps right? How would bootnodes reporting v0 cause a failure? Also we support version 0 and version 1.

We discussed it and I updated the PR description to explain it better.

PR #1807, opened 2025-05-05
feat: make beacon bridge use Electra variants

morph-dev:
This is a followup of https://github.com/ethereum/trin/pull/1801

### What was wrong?

Beacon Bridge will stop working after Pectra fork.

### How was it fixed?

Update our types to be Pectra instead of Deneb.

NOTE: We don't have a way to test these endpoints until the fork, as they would fail (except in some cases when format didn't change).

Ideally, we should be able to decode/deserialize correct version and it's something I will work on. But I don't have time to do it until the fork.

This code can/should be deployed as soon as the fork happens on mainnet.

### To-Do


KolbyML (approved):
:shipit: looks good

PR #1808, opened 2025-05-06
feat: validate blocks while building E2HS file

KolbyML:
### What was wrong?

Building E2HS files is time consuming. There isn't a good way for the bridge to validate post-capella headers due to the E2HS bridge not having access to a provider. We want to know sooner then later if a E2HS file is valid or not.

### How was it fixed?

Validate the block, use the latest HistoricalSummaries if avaliable, if HistoricalSummaries aren't avaliable there are no post capella headers to prove so there is no issue.


morph-dev (commented):
:ship: 

It would be good if you would locally verify that this works for fork boundary e2hs files.

morph-dev (approved):

KolbyML:
I tested the fork boundry for capellia and it worked

PR #1809, opened 2025-05-06
feat: update light-client to use Electra types

morph-dev:
### What was wrong?

The light-client doesn't support Electra.

### How was it fixed?

Added support for Electra types.

Note: There are couple of issues with this PR, that I plan to address in the followup work:

- No tests vectors - I will create and add this after Electra fork
- Not able to test this - working on testing this on Sepolia, but requires a lot of changes that I don't want to commit
- We are still not able to cross forks (we will only work for Pectra types) - This requires much bigger refactoring and will be done in the future


KolbyML (approved):
:shipit: looks good

PR #1810, opened 2025-05-06
feat: add sepolia network specification

KolbyML:
### What was wrong?

We didn't have a network specification for Sepolia

### How was it fixed?

By adding one

### Followup PRS I will make

- add sepolia accumulator-hashes
- calculate fork digest properly, instead of using hard coded value


morph-dev (approved):
There are several things on beacon side that should be updated as well. On top of my head:
- We should use sepolia BaseConfig (which I'm adding in #1811)
- We should not used hard-coded fork-digest (or we should use correct hard-coded fork-digests)

It's fine to do these in followup PRs.

PR #1811, opened 2025-05-06
feat: light-client should use its config for slot_timestamp

morph-dev:
### What was wrong?

The light-client only uses `network_spec()` for calculating slot's timestamp.

Since it already has its own config and which uses for most calculations, it should use it for this as well.

### How was it fixed?

Added `slot_timestamp` function in light-client.

Also added Sepolia testnet configuration, which I used to test this (I left it as others might fine it useful as well).

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

PR #1812, opened 2025-05-06
refactor(light-client): use FixedBytes not Vec<u8>

KolbyML:
### What was wrong?

For some reason the light-client crate is using Vec<u8> instead of fixed bytes. This isn't consistent with the rest of our codebase, these types have defined sizes so we should use types which define these constraints

- https://github.com/ethereum/trin/pull/1811#issuecomment-2856020128
- https://github.com/ethereum/trin/pull/1811#issuecomment-2856023345

### How was it fixed?

Updated the types


crates/light-client/src/database.rs
```diff
@@ -48,41 +49,38 @@ impl Database for FileDB {
...
 
-        if buf.len() == 32 && res.is_ok() {
-            Ok(buf)
+        if buf.len() == 32 && result.is_ok() {
+            Ok(B256::from_slice(&buf))
         } else {
-            Ok(self.default_checkpoint.clone())
+            Ok(self.default_checkpoint)
         }
     }
```
morph-dev, line 65: We can replace this entire function with something like:
```rust
let Ok(bytes) = fs::read(self.data_dir.join("checkpoint")) else {
    return Ok(self.default_checkpoint);
};
Ok(B256::try_from(bytes.as_slice()).unwrap_or(self.default_checkpoint))
```

Up to you. If you accept suggestion, you can also use `fs::write(..)` and replace almost entire `save_checkpoint` with one line (we still need `fs::create_dir_all(&self.data_dir)?;`).

crates/light-client/src/utils.rs
```diff
@@ -9,14 +8,11 @@ pub fn u64_to_hex_string(val: u64) -> String {
 
 #[derive(Default, Debug, TreeHash)]
 struct ForkData {
-    current_version: FixedVector<u8, U4>,
+    current_version: B32,
```
morph-dev, line 11: and here

crates/light-client/tests/sync.rs
```diff
@@ -16,9 +17,9 @@ async fn setup() -> ConsensusLightClient<MockRpc> {
     };
 
     let checkpoint =
-        hex::decode("c62aa0de55e6f21230fa63713715e1a6c13e73005e89f6389da271955d819bde").unwrap();
+        B256::from_hex("c62aa0de55e6f21230fa63713715e1a6c13e73005e89f6389da271955d819bde").unwrap();
```
morph-dev, line 20: nit: should we use `b256!`?

crates/light-client/src/node.rs
```diff
@@ -24,7 +25,7 @@ impl<R: ConsensusRpc> Node<R> {
     pub fn new(config: Arc<Config>) -> Result<Self, NodeError> {
         let consensus_rpc = &config.consensus_rpc;
         let checkpoint_hash =
-            &config
+            *config
                 .checkpoint
                 .as_ref()
```
morph-dev, line 30: nit: I could be wrong, but I think you can now remove `.as_ref()`.

morph-dev (approved):
Yes, so much nicer!

But I'm really against `B32`... I think we should just use `FixedBytes<4>` as it's cleaner (B32 is just an alias for it).

I marked all occurrences as it would be easy to miss (since they are alias).

KolbyML:
> But I'm really against `B32`... I think we should just use `FixedBytes<4>` as it's cleaner (B32 is just an alias for it).

I think we should use B32 to be consistent with our B256 usage, but sure I can switch it to FixedBytes<4>

PR #1821, opened 2025-05-13
feat: add docker image for e2hs-writer

KolbyML:
### What was wrong?

I am going to deploy e2hs-writer so that we can maintain and always keep the list up to date https://e2hs.ethportal.net/index.html

### How was it fixed?

By adding a docker image for e2hs-writer


morph-dev (approved):
Circleci config looks good. Docker file is more of a rubber stamp approval

:shipit:

PR #1822, opened 2025-05-13
feat: add latest end of block range option for e2hs bridge

KolbyML:
### What was wrong?
I am setting up E2HS bridges but a problem that occurs is there are always new e2hs files being added. We want to deploy bridges which cover ranges of the block space so for example if I have 5 bridges

And the total block range is roughly 0 to 21 million

- 0 to 5_000_000
- 5_000_001 to 10_000_000
- 10_000_001 to 15_000_00
- 15_000_001 to 20_000_00
- 20_000_001 to latest

^ I would probably want to deploy the bridges something like this where. Everytime the 5th bridge does a cycle it checks what is the latest E2HS file I can gossip, then includes that moving window in its range.

### How was it fixed?

adding a latest option which can be specified as the end of the range.

At the initialization of the bridge check what is the latest E2HS file we can bridge

### DIFF
--- a/bin/portal-bridge/src/bridge/e2hs.rs
+++ b/bin/portal-bridge/src/bridge/e2hs.rs
@@ -1,5 +1,6 @@
 use std::{
     collections::HashMap,
+    fmt::{Display, Formatter},
     str::FromStr,
     sync::{Arc, Mutex},
     time::Instant,
@@ -8,9 +9,8 @@ use std::{
 use alloy::primitives::B256;
 use anyhow::ensure;
 use bytes::Bytes;
-use clap::Parser;
 use e2store::{
-    e2hs::{BlockTuple, E2HSMemory},
+    e2hs::{BlockTuple, E2HSMemory, BLOCKS_PER_E2HS},
     utils::get_e2hs_files,
 };
 use ethportal_api::{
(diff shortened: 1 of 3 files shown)


bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -497,11 +508,28 @@ impl Gossiper {
...
+    }
+}
+
 /// BlockRange used specifically for the E2HS bridge
-#[derive(Parser, Debug, Clone)]
+#[derive(Debug, Clone)]
 pub struct BlockRange {
     pub start: u64,
-    pub end: u64,
+    pub end: BlockRangeEnd,
```
morph-dev, line 532: In my opinion, this should have been: `Option<u64>` (with short explanation). That way you don't need `BlockRangeEnd` structure at all.

I would also consider specifying the flag in the following way: `100-` (instead of `100-latest`).

bin/portal-bridge/src/bridge/e2hs.rs
```diff
@@ -111,7 +111,18 @@ impl E2HSBridge {
...
 
-        let epochs = block_range_to_epochs(self.block_range.start, self.block_range.end);
+        let block_range_end = match self.block_range.end {
+            BlockRangeEnd::Latest => {
+                let max_epoch = self
+                    .e2hs_files
+                    .keys()
+                    .max()
+                    .expect("to be able to get max epoch");
+                max_epoch * BLOCKS_PER_E2HS as u64 + BLOCKS_PER_E2HS as u64
```
morph-dev, line 121:
```suggestion
                max_epoch * BLOCKS_PER_E2HS as u64 + BLOCKS_PER_E2HS as u64 - 1
```

My understanding is that end of the range is inclusive, so it should be like I suggested, right?

You can also do it this way:
```suggestion
                (max_epoch + 1) * BLOCKS_PER_E2HS as u64 - 1
```

morph-dev (approved):
I strongly suggest about using `Option<u64>`, but I'm not going to block this PR if you are strongly against it. So it's up to you at the end.

Calculating `block_range_end` correctly is mandatory, but I'm pre-approving it as I think it shouldn't be controversial.

:rocket:

bin/portal-bridge/src/bridge/e2hs.rs
KolbyML: sure I think 100- is good, I was originally going to do it like that, but I wasn't sure if it would be ok

PR #1827, opened 2025-05-14
ci: fix workflow triggering on tag + matrix parameters

morph-dev:
### What was wrong?

The recently updated circleci config to build and publish docker images is not triggering on tag.

This is not something that I was able to test, but I'm pretty confident that this will solve the problem.

### How was it fixed?

Now I only use `<< pipeline.git.tag >>` as a `when` config for a `docker-master-tag` workflow.

I also removed the duplication of invoking all jobs manually, and I'm using [matrix jobs](https://circleci.com/docs/configuration-reference/#matrix).

### To-Do

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -317,114 +317,40 @@ workflows:
   docker-master-commit:
     jobs:
       - docker-build:
-          name: docker_build_trin
-          target: trin
+          name: docker_build__<< matrix.target >>
           tags: latest
+          matrix:
+            parameters:
+              target: ["trin", "bridge", "e2hs-writer", "trin-execution"]
       - docker-publish:
-          name: docker_publish_trin
-          target: trin
+          name: docker_publish__<< matrix.target >>
           requires:
-            - docker_build_trin
+            - docker_build__<< matrix.target >>
           filters:
             branches:
               only: master
-      - docker-build:
-          name: docker_build_bridge
-          target: bridge
-          tags: latest
-      - docker-publish:
-          name: docker_publish_bridge
-          target: bridge
-          requires:
-            - docker_build_bridge
-          filters:
-            branches:
-              only: master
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

.circleci/config.yml
```diff
@@ -317,114 +317,40 @@ workflows:
   docker-master-commit:
     jobs:
       - docker-build:
-          name: docker_build_trin
-          target: trin
+          name: docker_build__<< matrix.target >>
```
KolbyML, line 320:
```suggestion
          name: docker_build_<< matrix.target >>
```

were the double _ intentional?

morph-dev: Yes. We already have one in "docker_build" and some targets might contain "_" in the future. I figured that using two would make things clearer.

KolbyML: Oh ok :+1:

carver:
> Now I only use `<< pipeline.git.tag >>` as a `when` config for a `docker-master-tag` workflow.

Just to make sure I understand the theory, the reason that the tag build was not triggering is that:

```
[ << pipeline.git.tag >>, << pipeline.git.branch.is_default >> ]
```

will never be true, because it won't simultaneously treat the event as a tag and as a branch, right?

One option is to go the route of planning other tags to look like `bridge-v1.2.3` if we want to trigger tag events for different binaries.

morph-dev:
> will never be true, because it won't simultaneously treat the event as a tag and as a branch, right?

Yes, that's correct.

> One option is to go the route of planning other tags to look like `bridge-v1.2.3` if we want to trigger tag events for different binaries.

Yes, that would be an option. If we start doing that frequently, it would make sense to adjust circleci config to support that.
If it is one-time thing, then we can just do it manually, like we did before.

PR #1828, opened 2025-05-14
refactor: move block/header builder for ephemeral bridge

KolbyML:
### What was wrong?

We defined `ExecutionBlockBuilder` and `ExecutionHeaderBuilder` in `e2hs-writer`, but I need it in `portal-bridge` to write the ephemeral header bridge, since I need need bridge anchored headers after the period is confirmed

### How was it fixed?

Moving the structs to `ethportal-api`


crates/ethportal-api/src/types/execution/builders/execution_block_builder.rs
morph-dev, line 1: nit: considering that these files are now in `ethportal-api/src/types/execution/builders`, I think we can name this file as `block.rs` or `block_builder.rs`. I would leave the name of the struct as it is (i.e. `ExecutionBlockBuilder`). Same for the other file.

---

nit: Also, since all these functions have similar signature, I think we should create function like this:
```rust
pub fn build(
    block: &BeaconBlock,
    historical_batch: &HistoricalBatch,
) -> anyhow::Result<(HeaderWithProof, BlockBody)> {
    match block {
        BeaconBlock::Bellatrix(block) => Self::bellatrix(block, historical_batch),
        ...
    }
}
```

Maybe `block` should be `SignedBeaconBlock`, or maybe we should have for both, depending on how it's used.

morph-dev (approved):
Looks ok.

Main suggestion is that most of the logic in `bin/trin-execution/src/era/beacon.rs` can be simplified/removed.

I think it would be nice, but none of the comments are blocking. So up to you.

:ship:

PR #1829, opened 2025-05-14
ci: fix job names across workflows

morph-dev:
### What was wrong?

The naming of the jobs across workflows wasn't very consistent and nice.

### How was it fixed?

Renamed jobs so it's clear what is their trigger.


KolbyML (approved):
:shipit: looks good

PR #1830, opened 2025-05-14
feat: add trace offer for a series of content keys

KolbyML:
### What was wrong?

`send_offer_trace` only accepts 1 key-value pair. In most cases that is fine, but for we want to send multiple key-value pairs in the Ephemeral History Bridge, so this doesn't work.

### How was it fixed?

I am open to suggestions. I think it is easier/cleaner to expose an api which allows for the 1 key-value pair case and the multiple key-value case as they are pretty distinct.

### DIFF
--- a/crates/portalnet/src/overlay/protocol.rs
+++ b/crates/portalnet/src/overlay/protocol.rs
@@ -609,48 +609,21 @@ impl<
         content_key: RawContentKey,
         content_value: RawContentValue,
     ) -> Result<OfferTrace, OverlayRequestError> {
-        // Construct the request.
-        let (result_tx, mut rx) = tokio::sync::mpsc::unbounded_channel();
-        let request = Request::PopulatedOfferWithResult(PopulatedOfferWithResult {
-            content_items: vec![(content_key, content_value)],
-            result_tx,
-        });
+        match self
+            .send_offer_trace_with_multiple_items(enr, vec![(content_key, content_value)])
+            .await?
+        {
+            OfferTraceMultipleItems::Success(accept_code_list) => {
+                if accept_code_list.len() != 1 {
+                    return Err(OverlayRequestError::Failure(format!(
+                        "Expected exactly one AcceptCode in the response. Got {}",
+                        accept_code_list.len()
+                    )));
+                }
 
-        let direction = RequestDirection::Outgoing {
-            destination: enr.clone(),
-        };
-
-        // Send the offer request and wait on the response.
-        // Ignore the accept message, since we only care about the trace.
-        self.send_overlay_request(request, direction).await?;
-
-        // Wait for the trace response.
-        match rx.recv().await {
-            Some(accept) => {
-                let offer_trace = match accept {
-                    OfferTraceMultipleItems::Success(accept_code_list) => {
-                        if accept_code_list.len() != 1 {
-                            return Err(OverlayRequestError::Failure(format!(
-                                "Expected exactly one AcceptCode in the response. Got {}",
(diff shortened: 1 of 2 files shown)


crates/portalnet/src/overlay/protocol.rs
```diff
@@ -612,7 +612,58 @@ impl<
...
+                );
+                Err(OverlayRequestError::ChannelFailure(
+                    "Error receiving TraceOffer query response".to_string(),
+                ))
+            }
+        }
+    }
+
+    /// Send Offer request with trace, without storing the content into db, with multiple items
+    pub async fn send_offer_trace_with_multiple_items(
```
morph-dev, line 658: There is some duplicated logic between this and function above it (send_offer_trace).
Can't we make `send_offer_trace` simpler, something like this:
```rust
pub async fn send_offer_trace(...) -> Result<OfferTrace, OverlayRequestError> {
    let content_items = vec![content_key, content_value];
    match self.send_offer_trace_with_multiple_items(enr, content_items).await? {
        OfferTraceMultipleItems::Success(accept_code_list) => {
            ...
        }
        OfferTraceMultipleItems::Failed => Ok(OfferTrace::Failed),
    }
}
```

crates/portalnet/src/overlay/service/offer.rs
```diff
@@ -358,13 +360,9 @@ impl<
         // keys
         if content_keys.all_declined() {
             if let Some(tx) = gossip_result_tx {
-                if content_keys.len() != 1 {
-                    error!(
-                        "OfferTrace only supports one content key at a time, got {}",
-                        content_keys.len()
-                    );
+                if let Err(err) = tx.send(OfferTraceMultipleItems::Success(content_keys)) {
+                    warn!(%err, "Unable to send OfferTrace Success result");
```
morph-dev, line 364: If I'm not mistaken, `tx.send` would fail only if channel is closed, but we wouldn't know why it's closed.

And since all error messages are the same (except Success / Failed part), it would be hard to know in which case this happens.

I think it's probably not so important to even log these messages (how it was before), but if we already want to do it, I would describe them a bit better (e.g. they would include something that indicates which line actually printed the message). If you want, I can make suggestions for that.

Either way, no big deal. Up to you.

crates/portalnet/src/overlay/service/offer.rs
```diff
@@ -432,15 +434,11 @@ impl<
...
-                            content_keys.len()
-                        );
+                    if let Err(err) = tx.send(OfferTraceMultipleItems::Success(content_keys)) {
+                        warn!(%err, "Unable to send OfferTrace Success result");
                     }
-                    let _ = tx.send(OfferTrace::Success(content_keys[0]));
-                } else {
-                    let _ = tx.send(OfferTrace::Failed);
+                } else if let Err(err) = tx.send(OfferTraceMultipleItems::Failed) {
+                    warn!(%err, "Unable to send OfferTrace Failed result");
```
morph-dev, line 441: nit: I would do this slightly differently:

```rust
let result = if result {
    OfferTraceMultipleItems::Success(content_keys)
} else {
    OfferTraceMultipleItems::Failed
};
if let Err(err) = tx.send(result) {
    warn!(%err, "Unable to send OfferTrace result");
}
```

morph-dev (approved):
Looks mostly ok.

crates/portalnet/src/overlay/protocol.rs
KolbyML: Good idea!

PR #1831, opened 2025-05-15
refactor: break e2hs-writer and portal-bridge dependency

### LINKED ISSUE #1824: Break e2hs-writer and portal-bridge dependency
The `e2hs-writer` binary depends only on small part of `portal-bridge`, primarily structures in [bin/portal-bridge/src/api](https://github.com/ethereum/trin/blob/master/bin/portal-bridge/src/api).
I think that they are isolated enough and could be extracted into separate crate.

Benefit (other than better structured code) would be that `e2hs-writer` binary would be smaller as it wouldn't need to pull all transitive dependencies from `portal-bridge`.

Arya-A-Nair:
### What was wrong?
The e2hs-writer binary depends only on small part of portal-bridge, primarily structures in [bin/portal-bridge/src/api](https://github.com/ethereum/trin/blob/master/bin/portal-bridge/src/api).
they are isolated enough and could be extracted into separate crate.
Closes #1824 

### How was it fixed?
Code migrated to a different crate and updated references

### To-Do


Cargo.toml
```diff
@@ -119,6 +120,7 @@ trin-state = { path = "crates/subnetworks/state" }
 trin-storage = { path = "crates/storage" }
 trin-utils = { path = "crates/utils" }
 trin-validation = { path = "crates/validation" }
+ethereum-rpc-client = { path = "crates/ethereum-rpc-client" }
```
KolbyML, line 123: This should be sorted aphabetically in the list

KolbyML (commented):
I will let @morph-dev review this as he made the issue for it

morph-dev:
I didn't check in details, but files in [bin/portal-bridge/src/api](https://github.com/ethereum/trin/blob/master/bin/portal-bridge/src/api) should no longer be there.

Instead, the `portal-bridge` crate should depend and use newly created crate.

Arya-A-Nair:
> I didn't check in details, but files in [bin/portal-bridge/src/api](https://github.com/ethereum/trin/blob/master/bin/portal-bridge/src/api) should no longer be there.
> 
> Instead, the `portal-bridge` crate should depend and use newly created crate.

@morph-dev I have removed the code inside [bin/portal-bridge/src/api](https://github.com/ethereum/trin/blob/master/bin/portal-bridge/src/api) and updated the references everywhere

crates/ethereum-rpc-client/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "ethereum-rpc-client"
+version = "0.1.0"
+edition = "2021"
+description = "Ethereum RPC client for consensus and execution APIs. Extracted from portal-bridge."
+
+[dependencies]
```
morph-dev, line 7: Couple of comments:
- keep this list sorted
- use `<dep>.workspace = true` whenever possible, probably for all of them

bin/portal-bridge/Cargo.toml
morph-dev, line 1: Now that some logic is moved out of this crate, remove dependencies that are no longer needed (e.g. I think `reqwest-middleware` is one of them).

crates/ethereum-rpc-client/src/lib.rs
```diff
@@ -1,3 +1,4 @@
 pub mod consensus;
```
morph-dev, line 1: Add this at the begining of file:

```rust
#![warn(clippy::unwrap_used)]
```

crates/ethereum-rpc-client/src/constants.rs
```diff
@@ -0,0 +1,10 @@
+use tokio::time::Duration;
+
+/// The timeout in seconds is applied from when the request starts connecting until the response
+/// body has finished. Also considered a total deadline.
+pub const DEFAULT_TOTAL_REQUEST_TIMEOUT: u64 = 20;
+
+// Number of seconds to wait before retrying a provider request
+pub const FALLBACK_RETRY_AFTER: Duration = Duration::from_secs(5);
```
morph-dev, line 8: These 2 constants should now be removed from `bin/portal-bridge/src/constants.rs`

crates/ethereum-rpc-client/src/constants.rs
```diff
@@ -0,0 +1,10 @@
...
+
+pub const DEFAULT_BASE_EL_ENDPOINT: &str = "https://geth-lighthouse.mainnet.eu1.ethpandaops.io/";
```
morph-dev, line 10: All 4 constants from `bin/portal-bridge/src/lib.rs` should be moved here (together with comments).

testing/ethportal-peertest/Cargo.toml
```diff
@@ -17,6 +17,7 @@ alloy-hardforks.workspace = true
 anyhow.workspace = true
 discv5.workspace = true
 e2store.workspace = true
+ethereum-rpc-client.workspace = true
```
morph-dev, line 20: No need for `ethportal-peertest` to depend on either `portal-bridge` or `ethereum-rpc-client`.

It seems that all it uses is `construct_proof` function, which isn't used from anywhere else. So remove that function from `ethereum-rpc-client` and move it here (it's very small, so you can even inline it).

morph-dev (changes requested):

testing/ethportal-peertest/src/scenarios/offer_accept.rs
```diff
@@ -357,11 +360,16 @@ pub async fn test_offer_concurrent_utp_transfer_limit(peertest: &Peertest, targe
...
-        );
+
+        let proof =
+            PreMergeAccumulator::construct_proof(&tuple.header.header.clone(), &epoch_acc).unwrap();
+        let proof = BlockHeaderProof::HistoricalHashes(proof);
+
+        let header_value = HistoryContentValue::BlockHeaderWithProof(HeaderWithProof {
+            header: tuple.header.header.clone(),
+            proof,
+        });
```
morph-dev, line 371:
```suggestion
        let header = tuple.header.header.clone();
        let proof =
            PreMergeAccumulator::construct_proof(&header, &epoch_acc).unwrap();
        let proof = BlockHeaderProof::HistoricalHashes(proof);

        let header_value = HistoryContentValue::BlockHeaderWithProof(HeaderWithProof {
            header,
            proof,
        });
```

morph-dev: We could extract header at the start, which both requires only one clone, and simplifies following logic.

crates/ethereum-rpc-client/src/execution.rs
morph-dev, line 1: We can remove `construct_proof` function from this file, as it is not longer used.
Without it, I believe that this crate doesn't need to depend on `trin-validation` crate anymore.

crates/ethereum-rpc-client/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+ethereum_ssz_derive.workspace = true
+ethportal-api.workspace = true
+reqwest = { version = "0.11", features = ["json", "stream"] }
+reqwest-middleware = "0.2"
+reqwest-retry = "0.2"
+serde.workspace = true
+serde_json.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+trin-validation.workspace = true
```
morph-dev, line 28: After removing `construct_proof` function, I think we can remove `trin-validation` dependency.

crates/ethereum-rpc-client/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+[dependencies]
+alloy.workspace = true
+alloy-hardforks.workspace = true
+anyhow.workspace = true
+ethereum_ssz.workspace = true
+ethereum_ssz_derive.workspace = true
+ethportal-api.workspace = true
+reqwest = { version = "0.11", features = ["json", "stream"] }
+reqwest-middleware = "0.2"
+reqwest-retry = "0.2"
```
morph-dev, line 23: Any reason why these are not the same as the ones in portal-bridge before this PR, meaning, these:

```rust
reqwest.workspace = true
reqwest-middleware = { version = "0.3.3", features = ["json"] }
reqwest-retry = "0.6.1"
```

I would be fine with changing them as long as:
- reqwest uses workspace version
    - could use additional features if it needs to (but it didn't before so I don't know why it would need it now)
- other two are newer, not older versions (as long as they are compatible with `reqwest.workspace`)

morph-dev (commented):
Almost there

testing/ethportal-peertest/src/scenarios/offer_accept.rs
morph-dev, line 1: I think that `ethportal-peertest` crate no longer needs to depend on `portal-bridge` crate.

morph-dev (approved):
I think that there is just one more dependency that should be removed and it's good to be merged. Can you do it?

Thank you for your contribution

morph-dev:
Thank you for contribution. Merging now

PR #1832, opened 2025-05-16
feat: create ChainHead

morph-dev:
### What was wrong?

We need a structure that will handle head of the chain data, including:
- latest block
- latest finalized block
- latest historical summaries
- ephemeral history content (to be added in the future PRs)

### How was it fixed?

Created structure and added some simple ways for handling data.

Structure is currently not initialized or updated. I started working on this but it's getting more complicated than expected and I decided to split it into separate PR.


KolbyML (approved):
:shipit: looks fine I wish it was used though

PR #1834, opened 2025-05-17
chore: bump dependencies

KolbyML:
### What was wrong?

We haven't updated most of our dependencies in a long long time

### How was it fixed?

bump most of them

~~I left 2~~
I left 1
- jsonrpcee since we depend on reth-ipc for it and currently there is an awkward dependency issue they are working out
~~- vergen as updating it is a little involved as they almost redid the libraries interface, so probably best for a future PR~~
vergen wasn't that bad I will include it in this PR


bin/trin-execution/src/trie_walker/mod.rs
```diff
@@ -172,7 +172,7 @@ impl<DB: TrieWalkerDb> Iterator for TrieWalker<DB> {
 
 #[cfg(test)]
 mod tests {
-    use std::{str::FromStr, sync::Arc};
+    use std::{result::Result::Ok, str::FromStr, sync::Arc};
```
morph-dev, line 175: Why do we have to import `Ok`?

crates/ethportal-api/build.rs
```diff
@@ -1,15 +1,25 @@
...
-        .cargo_target_triple()
+    let build = BuildBuilder::default().build_timestamp(true).build()?;
+    let cargo = CargoBuilder::default()
+        .features(true)
+        .target_triple(true)
+        .build()?;
+    let git2 = Git2Builder::default()
+        .describe(false, true, None)
+        .dirty(true)
+        .sha(false)
```
morph-dev, line 15: This used to be `true`. Was change intentional?

crates/ethportal-api/build.rs
```diff
@@ -1,15 +1,25 @@
...
-        .cargo_features()
-        .cargo_target_triple()
+    let build = BuildBuilder::default().build_timestamp(true).build()?;
+    let cargo = CargoBuilder::default()
+        .features(true)
+        .target_triple(true)
+        .build()?;
+    let git2 = Git2Builder::default()
+        .describe(false, true, None)
+        .dirty(true)
```
morph-dev, line 14: We didn't use this before. Why do we need it now?

Cargo.toml
```diff
@@ -59,47 +59,47 @@ ethereum_ssz_derive = "0.9"
...
 tokio-test = "0.4.2"
 tracing = "0.1.36"
 tracing-subscriber = "0.3.15"
-tracing-test = "0.1"
+tracing-test = "0.2"
 tree_hash = "0.10"
 tree_hash_derive = "0.10"
 uds_windows = "1.0.1"
-ureq = { version = "2.5.0", features = ["json"] }
+ureq = { version = "3.0", features = ["json"] }
```
morph-dev, line 102: I think you removed all usage of `ureq`. Should we just remove it?

morph-dev (approved):

bin/trin-execution/src/trie_walker/mod.rs
KolbyML: image

traced_test throws this error unless we import it

KolbyML: I removed the import, by just adding an empty anyhow result as the return type

crates/ethportal-api/build.rs
KolbyML: The true condition doesn't enable anything we use from my testing

bin/trin-execution/src/trie_walker/mod.rs
morph-dev: Do we even need `traced_test` and why?

Reading at documentation,my understanding is that it's used so you can assert that certain logs were/weren't logged, and I don't think we are doing that. What am I missing?

morph-dev: This is the documentation that I was looking at: https://docs.rs/tracing-test/latest/tracing_test/index.html

KolbyML: I don't know why it is there to be honest

crates/ethportal-api/build.rs
KolbyML: If I don't have this line the build script fails

KolbyML: I am assuming it enables a few environment variables we need

bin/trin-execution/src/trie_walker/mod.rs
KolbyML: Yeah I was able to remove it without issues

PR #1835, opened 2025-05-17
chore: bump rust version to 1.87

KolbyML:
### What was wrong?

There is a new stable release

### How was it fixed?

bump it. I noticed the newest reth-ipc requires a newer stable version, I didn't end up bumping that crate to to other issues with reth-ipc/jsonrpcee as jsonrpcee privated members, reth-ipc needed to work or something


morph-dev (approved):
Mostly rubber stamped

:ship:

PR #1838, opened 2025-05-18
feat: remove DistanceFunction

morph-dev:
### What was wrong?

We have an object called DistanceFunction that doesn't seem to be used a lot.
Instead, we use `Metric` trait (as generic type) much more.

### How was it fixed?

Removed DistanceFunction and replaced with with generic type where needed.

### To-Do


KolbyML (approved):
:shipit: PR looks good.

morph-dev:
> Aren't there still plans to use a non-XOR distance function for the tip of the state network?
> 
> I think that was the point of the generic distance function.

We can use Metric for that, right? The same way we use XorMetric at the moment.

PR #1840, opened 2025-05-19
feat: subnetwork stores should use affected_by_radius

morph-dev:
Followup to #1839

### What was wrong?

When implementing `is_key_within_radius_and_unavailable` function, content store should take into consideration whether content key is affected by radius or not.

See #1839 for more details.

### How was it fixed?

Implemented missing functionality.

More work will follow regarding this as ephemeral content will not be stored in the current storage object.

### To-Do


KolbyML (approved):
I am surprised this is it's own pr

morph-dev:
> I am surprised this is it's own pr

It kind of makes sense as it's not really related to other work. Also, if there is concern with other prs, I don't want this to be blocked on it (or the other way around).

PR #1842, opened 2025-05-19
refactor: move EthereumAPI functions into Execution/Consensus API

KolbyML:
### What was wrong?

I need the same API calls I added for E2HS-Writer for the Ephemeral History Bridge

### How was it fixed?

Remove `EthereumAPI` and split the functions I needed out into `ExecutionAPI` and `ConsensusAPI` respectively


crates/ethereum-rpc-client/src/consensus/mod.rs
```diff
@@ -154,6 +158,87 @@ impl ConsensusApi {
...
+        Ok(state)
+    }
+
+    /// Fetches the Beacon Block for a given slot.
+    ///
+    /// Slots can be skipped, so we walk until we find a non-skipped slot.
+    ///
+    /// The retry limit is 5, because I don't think there is a gap of more than 5 slots in the
+    /// beacon chain
+    pub async fn fetch_beacon_block_retry(
```
morph-dev, line 208: nit: I don't like the `retry` part of the name, because it can be confusing (while it was used by e2hs_writer, it was ok as it was used only there, and now it's more public).

I would say that it's more common to use `retry` when we retry the same query multiple times (using same or fallback endpoint). But here, we try different slots until we find the one that matches, so it's not really "retry".

Sadly, I don't have a good suggestion to a name, so I will give you a few and you can pick some (or came up with better one):
- find_first_beacon_block (I like this one the most)
- fetch_beacon_block_with_advancing
- fetch_first_available_beacon_block
- fetch_beacon_block_until_available

crates/ethereum-rpc-client/src/consensus/mod.rs
```diff
@@ -154,6 +158,87 @@ impl ConsensusApi {
...
+                        slot += 1;
+                        tries += 1;
+                        continue;
+                    }
+                    ensure!(
+                        state.slot == slot,
+                        "The slot {slot} is not equal to the state slot {}",
+                        state.slot
+                    );
+                    break state;
```
morph-dev, line 188: nit: Can we just return here? I think it will make code a bit cleaner

crates/ethereum-rpc-client/src/consensus/mod.rs
```diff
@@ -154,6 +158,87 @@ impl ConsensusApi {
...
+    pub async fn fetch_beacon_block_retry(
+        &self,
+        slot: u64,
+    ) -> anyhow::Result<SignedBeaconBlockElectra> {
+        let mut tries = 0;
+        let mut slot = slot;
+
+        let block = loop {
+            match self.get_beacon_block(slot.to_string()).await {
+                Ok(block) => break block,
```
morph-dev, line 217: nit: Same as above, I think it will be cleaner if we just return here.

morph-dev (approved):
:shipit:

PR #1843, opened 2025-05-20
fix: failed to read s3 bucket in dockerfile install ca-certificates

KolbyML:
### What was wrong?

e2hs-writer is failing to read the s3 bucket due to not having access to `ca-certificates`

### How was it fixed?

Install them, I tested this change and it works


morph-dev (approved):
:ship:

PR #1846, opened 2025-05-22
feat: implement ephemeral bridge

KolbyML:
### What was wrong?

Fixes https://github.com/ethereum/trin/issues/1789

### How was it fixed?

By implementing the ephemeral bridge

PR #1847, opened 2025-05-22
feat: add retry to get_receipts

KolbyML:
This PR is a requirement for
- https://github.com/ethereum/trin/pull/1846

### What was wrong?

There is a delay for when receipts become available, as the EL has to generate them. I add something similar to this code because we needed it to make the latest bridge reliable. Due to the aforementioned issue above. I removed this logic in this PR https://github.com/ethereum/trin/pull/1757 well optimizing the Single-E2HS-Writer, because we no longer used this functionality.

### How was it fixed?

Added the retry mechanism back, but only for `get_receipts()`, because it isn't needed in `get_receipts_range` as that function doesn't really have a reason to be used at the head of the chain.

I choose 5, because as I have seen delays of 3 seconds

### DIFF
--- a/crates/ethereum-rpc-client/src/execution.rs
+++ b/crates/ethereum-rpc-client/src/execution.rs
@@ -98,11 +98,12 @@ impl ExecutionApi {
         Ok((content_key, content_value))
     }
 
-    /// This function should only be used if we know the receipts should exist.
+    /// Fetches the receipts for a given block number.
     ///
-    /// The function will retry 5 times if the receipts are not ready yet.
-    /// Especially when querying for receipts at the end of the chain their is a delay until they
-    /// become available.
+    /// The function will retry 5 times if the result is null. Execution providers may return null
+    /// for any reason,
+    /// - if receipts are not available for the block yet
+    /// - too many requests are sent to the provider to rate limit
     pub async fn get_receipts(&self, block_number: u64) -> anyhow::Result<Receipts> {
         let block_param = format!("0x{block_number:01X}");
         let params = Params::Array(vec![json!(block_param)]);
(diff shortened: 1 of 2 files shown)


crates/ethereum-rpc-client/src/execution.rs
```diff
@@ -98,13 +98,33 @@ impl ExecutionApi {
         Ok((content_key, content_value))
     }
 
+    /// This function should only be used if we know the receipts should exist.
```
morph-dev, line 101: Can you clarify what you mean by: "if we know the receipts should exist"?
It implies that somebody might use it when either:
- they know it doesn't exist (in which case, why would they call it in the first place?)
- they are not sure if it exists. How can this happen? What should they do instead?

nit: somewhat less important, the first line of the function documentation usually stats on high level what function does, and the other lines describe more detailed behaviour

crates/ethereum-rpc-client/src/execution.rs
```diff
@@ -98,13 +98,33 @@ impl ExecutionApi {
...
+    ///
+    /// The function will retry 5 times if the receipts are not ready yet.
+    /// Especially when querying for receipts at the end of the chain their is a delay until they
```
morph-dev, line 104:
```suggestion
    /// Especially when querying for receipts at the head of the chain, there is a delay until they
```

crates/ethereum-rpc-client/src/execution.rs
```diff
@@ -98,13 +98,33 @@ impl ExecutionApi {
...
+    /// become available.
     pub async fn get_receipts(&self, block_number: u64) -> anyhow::Result<Receipts> {
```
morph-dev, line 106: nit: would it make sense to have two functions:
- `get_receipts` - without retry logic
- `get_receipts_retry` - with retry logic, that would call `get_receipts`

Up to you

morph-dev (approved):

crates/ethereum-rpc-client/src/execution.rs
KolbyML: I think this method is inherently more reliable so there is only upside's in all our usages. For example in the e2hs-writer head bridge I seen it fail to get receipts once, so it is fairly rare, but an EL can return null for any reason

PR #1850, opened 2025-05-22
feat: add select_random_peers to census

KolbyML:
### What was wrong?

This is a chunk from https://github.com/ethereum/trin/pull/1846

`select_peers()` take a content_key this doesn't make when gossiping a series of EphemeralHeadersByOffer as we just want working random Enr's

### How was it fixed?

At the top level of `census` add a function `select_random_peers()`, then just pass an option from there down.

### DIFF
--- a/bin/portal-bridge/src/census/scoring.rs
+++ b/bin/portal-bridge/src/census/scoring.rs
@@ -13,17 +13,20 @@ use super::peer::{Peer, PeerInfo};
 pub trait Weight: Send + Sync {
     fn weight(
         &self,
-        content_key: &impl OverlayContentKey,
-        content_id: &[u8; 32],
+        content_key: Option<&impl OverlayContentKey>,
+        content_id: &Option<[u8; 32]>,
         peer: &Peer,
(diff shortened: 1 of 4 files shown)


bin/portal-bridge/src/census/scoring.rs
```diff
@@ -13,17 +13,20 @@ use super::peer::{Peer, PeerInfo};
 pub trait Weight: Send + Sync {
     fn weight(
         &self,
-        content_key: &impl OverlayContentKey,
-        content_id: &[u8; 32],
+        content_key: Option<&impl OverlayContentKey>,
+        content_id: &Option<[u8; 32]>,
```
morph-dev, line 17: I think this should now be something like:
`content_key_and_id: &Option<(&impl OverlayContentKey, [u8; 32])>`

and line 26 should be:
`let content_key_and_id = content_key.map(|key| (key, key.content_id()));`

morph-dev (approved):
Good work. Few suggestions.

:rocket:

PR #1851, opened 2025-05-22
Add ephemeral history gossiper

KolbyML:
This PR depends on 
- https://github.com/ethereum/trin/pull/1850

### What was wrong?
This is a chunk from https://github.com/ethereum/trin/pull/1846

We need a Gossiper which handles the unique cases fund in the Ephemeral History Bridge we do different functions
- flushing proven headers once a period is provable
- gossiping an ephemeral bundle which includes a chain of Header's, but it can include multiple bodies/receipts if a re-org happens

### How was it fixed?

Add the code

### Work for followup PR's
- Add support for metrics for header series
- Think about how the AcceptCodeList from HeaderSeries would make sense in terms of the census OfferResult algo

### DIFF
--- a/bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
+++ b/bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
@@ -57,14 +57,14 @@ impl Gossiper {
         history_network: Arc<HistoryNetwork>,
         metrics: BridgeMetricsReporter,
         head_offer_limit: usize,
-        proven_headers_offer_limit: usize,
+        non_ephemeral_offer_limit: usize,
     ) -> Self {
         Self {
             census,
             history_network,
             metrics,
             head_semaphore: Arc::new(Semaphore::new(head_offer_limit)),
-            non_ephemeral_headers_semaphore: Arc::new(Semaphore::new(proven_headers_offer_limit)),
+            non_ephemeral_headers_semaphore: Arc::new(Semaphore::new(non_ephemeral_offer_limit)),
         }
     }
 
@@ -192,19 +192,15 @@ impl Gossiper {
         let encoded_content_value = content_value.encode();
         let mut tasks = vec![];
         for peer in peers.clone() {
-            let offer_permit = if is_head_offer {
-                self.head_semaphore
-                    .clone()
-                    .acquire_owned()
-                    .await
-                    .expect("to be able to acquire semaphore")
+            let semaphore = if is_head_offer {
+                self.head_semaphore.clone()
             } else {
-                self.non_ephemeral_headers_semaphore
-                    .clone()
-                    .acquire_owned()
-                    .await
-                    .expect("to be able to acquire semaphore")
+                self.non_ephemeral_headers_semaphore.clone()
             };
+            let offer_permit = semaphore
+                .acquire_owned()
+                .await
+                .expect("to be able to acquire semaphore");
             let content_key = content_key.clone();
             let raw_content_value = encoded_content_value.clone();
 


KolbyML:
@morph-dev this PR is ready to review now

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
```diff
@@ -0,0 +1,430 @@
...
+        receipts: Receipts,
+    ) -> JoinHandle<Vec<OfferTrace>> {
+        let content_key = HistoryContentKey::new_block_receipts(block_hash);
+        let content_value = HistoryContentValue::Receipts(receipts);
+        self.start_gossip_task(content_key, content_value)
+    }
+
+    /// The finalized state root should be for a finalized period. The beacon blocks passed in must
+    /// be contained with that respective period's block roots to be provable.
+    pub async fn gossiped_proven_headers(
```
morph-dev, line 168: If I'm not mistaken, this would start close to `8192*2` gossiping tasks.
I'm worried that all these tasks might starve the offer permit semaphore, especially because semaphore is fair. Meaning, we might not be able to gossip ephemeral content for some time.

Instead, we could do one of the following:
- send this content sequentially (it will take longer)
- use different semaphore

What do you think?

morph-dev (commented):
There is one/two high level comments that I would like to discuss.

But no major concerns and PR looks mostly ok.

bin/portal-bridge/src/bridge/ephemeral_history/ephemeral_bundle.rs
```diff
@@ -0,0 +1,36 @@
...
+use ethportal_api::{consensus::beacon_block::BeaconBlockElectra, BlockBody, Receipts};
+use revm_primitives::B256;
+
+pub struct EphemeralBundle {
+    pub head_block_root: B256,
+    /// Used to generate header series
+    pub beacon_blocks: Vec<BeaconBlockElectra>,
+    pub bodies: Vec<(B256, BlockBody)>,
+    pub receipts: Vec<(B256, Receipts)>,
+}
```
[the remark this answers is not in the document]
KolbyML: > I checked #1852 and exactly how this is initialized and used. The `beacon_blocks` has 8 more blocks, compared to `bodies` and `receipts`, right? That means that `bodies` and `receipts` are gossiped only once (at the very tip of the chain). I think this is very likely not going to be reliable enough.

I don't want to gossip 9 bodies and receipts every slot, if there is a problem with reliability I don't think that would be with the bridge.

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
```diff
@@ -0,0 +1,430 @@
...
+                Ok(offer_trace) => offer_trace,
+                Err(err) => {
+                    error!("Sending offer task failed: {err}");
+                    OfferTraceMultipleItems::Failed
+                }
+            })
+            .collect()
+    }
+
+    async fn send_offer_with_header_series(
```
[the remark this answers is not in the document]
KolbyML: This function assumes that the content is a header series. Multi offer would imply you could multi offer any set of data which isn't the case here

bin/portal-bridge/src/bridge/ephemeral_history/ephemeral_bundle.rs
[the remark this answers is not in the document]
KolbyML: Second if I was to do this we would do `pub blocks: Vec<(BeaconBlockElectra, Receipts)>` it wouldn't make sense for the Header/Body to be here when they can be derived

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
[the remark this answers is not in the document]
KolbyML: I will change it's input parameter's then to restrict it more

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
KolbyML: I did the third non-listed option of using a future stream then buffering the futures, which will basically rate limit it so only X futures can run at the same time.

- sequentially
this one isn't good
- semaphore
There are many ways this could be implemented using semaphore, so the suggestion is too broad for me to make any direct statements.

Anyways let me know what you think

KolbyML:
@morph-dev ready for another look, let me know if I missed anything

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
morph-dev: I think that you implementation doesn't work as you want it. I tried it in simplified form here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f9efd12f88caac5f19a20c9e2b8a1310

The reason is that because we have `tokio::spawn`, background task will start as soon as it's called (not when it's buffered).

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
```diff
@@ -0,0 +1,372 @@
...
+            }
+            Err(_) => {
+                error!("trace_offer timed out on history  : indicating a bug is present");
+                OfferTraceMultipleItems::Failed
+            }
+        };
+
+        // todo: add census record offer result for multiple items
+
+        // todo: add metrics report offer for multiple items
```
morph-dev, line 356: Do you plan to do this soon-ish (e.g. after ongoing PRs are merged)? If not, please open an issue and add link to it here

morph-dev (commented):
I think that buffered futures doesn't work as you are expecting. Happy to have a call and discuss it.

Other than that, no big concerns.

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
KolbyML: Your simplifed code isn't the same as our code hence it gives an invalid result. I updated it so that `tokio::spawn` is awaited in the function called and it works as expected https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f9efd12f88caac5f19a20c9e2b8a1310

KolbyML:

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
KolbyML: We talked on the call, and I did have an oversight in my implementation. After discussion using a semaphore makes the most sense, so I implemented it.

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
```diff
@@ -45,23 +45,26 @@ pub struct Gossiper {
...
 }
 
 impl Gossiper {
     pub fn new(
         census: Census,
         history_network: Arc<HistoryNetwork>,
         metrics: BridgeMetricsReporter,
-        offer_limit: usize,
+        head_offer_limit: usize,
+        proven_headers_offer_limit: usize,
```
morph-dev, line 60: nit: maybe name the variable (`proven_headers_offer_limit`) and semaphore (`non_ephemeral_headers_semaphore`) in the same way? Meaning, `non_ephemeral_offer_limit`

bin/portal-bridge/src/bridge/ephemeral_history/gossiper.rs
```diff
@@ -179,7 +192,19 @@ impl Gossiper {
...
+                    .acquire_owned()
+                    .await
+                    .expect("to be able to acquire semaphore")
+            } else {
+                self.non_ephemeral_headers_semaphore
+                    .clone()
+                    .acquire_owned()
+                    .await
+                    .expect("to be able to acquire semaphore")
+            };
```
morph-dev, line 207:
```suggestion
            let semaphore = if is_head_offer {
                self.head_semaphore.clone()
            } else {
                self.non_ephemeral_headers_semaphore.clone()
            };
            let offer_permit = semaphore
                    .acquire_owned()
                    .await
                    .expect("to be able to acquire semaphore")
```

morph-dev (approved):
Looks good! 🚀

PR #1857, opened 2025-05-25
chore: fix confusing comment in E2HSManager

KolbyML:
### What was wrong?

There was some confusion on a confusing comment https://github.com/ethereum/trin/pull/1855#discussion_r2106156634

### How was it fixed?

Updating the comment to be more accuate

### DIFF
--- a/bin/trin-execution/src/e2hs/manager.rs
+++ b/bin/trin-execution/src/e2hs/manager.rs
@@ -60,7 +60,7 @@ impl E2HSManager {
 
     pub async fn last_fetched_block(&self) -> anyhow::Result<&ProcessedBlock> {
         let Some(current_e2hs) = &self.current_e2hs else {
-            bail!("No blocks were fetched yet, current_e2hs is initialized in get_next_block");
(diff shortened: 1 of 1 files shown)


bin/trin-execution/src/e2hs/manager.rs
```diff
@@ -60,7 +60,7 @@ impl E2HSManager {
 
     pub async fn last_fetched_block(&self) -> anyhow::Result<&ProcessedBlock> {
         let Some(current_e2hs) = &self.current_e2hs else {
-            panic!("current_e2hs should always be present, perhaps it wasn't initialized in E2HSManager::new()?");
+            bail!("No blocks were fetched yet, current_e2hs is initialized in get_next_block");
```
morph-dev, line 63:
```suggestion
            bail!("No blocks were fetched yet, last_fetched_block is available only after successful get_next_block");
```
nit: I would phrase it like this. Up to you.

morph-dev (approved):
Thanks for the followup :shipit:

PR #1863, opened 2025-05-27
feat: add chain head metrics

morph-dev:
### What was wrong?

We don't have any metrics that indicate how close we are to the head of the chain.

### How was it fixed?

Added ChainHead metrics


crates/metrics/src/chain_head.rs
```diff
@@ -0,0 +1,115 @@
...
+        }
+        self.metrics.finalized_slot.set(finalized_slot as i64);
+    }
+}
+
+impl Default for ChainHeadMetricsReporter {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
KolbyML, line 115: couldn't we just use the Default derive?

KolbyML (approved):
:shipit: looks good

PR #1866, opened 2025-05-28
docs: mention the --mb alias in the CLI

carver:
Also, regenerate the help in the trin book, using the longer format.

### What was wrong?

The flag `--mb` is used pretty liberally in the trin book. It is not mentioned anywhere what it is.

### How was it fixed?

Having the flag mentioned in the long CLI help seems like the least we can do.
### To-Do

### DIFF
--- a/bin/trin/src/cli.rs
+++ b/bin/trin/src/cli.rs
@@ -136,7 +136,7 @@ pub struct TrinConfig {
         long = "storage.total",
         alias = "mb",
         help = "Maximum storage capacity (in megabytes), shared between enabled subnetworks",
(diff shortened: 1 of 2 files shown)


KolbyML:
> Also, regenerate the help in the trin book, using the longer format.
> 
> ### What was wrong?
> The flag `--mb` is used pretty liberally in the trin book. It is not mentioned anywhere what it is.
> 
> ### How was it fixed?
> Having the flag mentioned in the long CLI help seems like the least we can do.
> 
> ### To-Do
> * [ ]  Clean up commit history and use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/).

I will make a PR in a bit which will auto build these cli doc's pages for us

PR #1867, opened 2025-05-29
chore: add script to update Trin Book's cli page

KolbyML:
### What was wrong?

We were manually updating our Trin Book's cli configs this has 2 issues
- the book is often out of date for long periods of time
- We manually need to do this tedious work

### How was it fixed?

Reth has a script for this https://github.com/paradigmxyz/reth/blob/a505f4914780c9ec6d1ca4ae1182244b851b89a7/book/cli/help.rs so I took it and left a comment on the top of the file attributing them. We have done this in a lot of our CI code, I can probably name like 5 CI jobs which were taken from Reth/Lighthouse

I also add make pr, which contributors should run before making a PR, which runs all the recommended commands and updates the book


carver (commented):
I'm not doing a *deep* review of `book/src/cli/help.rs`, there's a lot there, but I guess you ran it already and the only difference was that temporary directory remapping, so that seems promising. I guess hopefully we would notice a bug pretty easily when the CLI output in the book changes in some weird way.

I don't love having a lot of functionality around that is unused or buggy, so I'm in favor of dropping a lot of stuff we don't need/use. More below

PR #1868, opened 2025-05-30
feat: add support for validating multiple content items

morph-dev:
### What was wrong?

When we are offered multiple content, the overlay service does all of the following on per-content basis:
- checks whether they should be accepted 
- validates them
- stores them

This doesn't work for EphemeralHeaders, which we have to process together (at least checking which content to accept and validating). This PR adds support for that.

### How was it fixed?

Added `should_we_store_batch` function that checks whether multiple content items should be stored.

Refactored Validator trait in the following ways:
- removed `additional_content_to_propagate` from ValidationResult (it's not used)
- made ValidationResult an enum
    - we can consider changing `Invalid(String)` variant into more descriptive type, e.g. `Invalid(ValidationFailureReason)` where:
        ```rust
        enum ValidationFailureReason {
            ContentValueNotDecodable,
            Invalid(<reason?>),  // e.g. merkle proof not valid
            FetchingAnchorContentFailuer(<reason?>), // e.g. failed to fetch Header when verifying BlockBody
            ...
        }
        ``` 
- added support for validating multiple content items

Refactored `OverlayService::handle_offer` to use these functions instead

### Remaining work

We would still do a fallback fetching in request fails or content is not validated.
I would have to think more whether this is desired for EphemeralContent, and if not, how exactly to distinguish/handle this case. This is something that I plan to do in a followup PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8518,6 +8518,7 @@ dependencies = [
  "ethereum_ssz",
  "ethereum_ssz_derive",
  "ethportal-api",
+ "futures",
  "lazy_static",
  "parking_lot 0.12.3",
  "quickcheck",
--- a/crates/portalnet/src/overlay/protocol.rs
+++ b/crates/portalnet/src/overlay/protocol.rs
@@ -227,11 +227,7 @@ impl<
         content_key: TContentKey,
         content_value: RawContentValue,
     ) -> PutContentInfo {
-        let should_we_store = match self
-            .store
-            .lock()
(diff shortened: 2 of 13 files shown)


morph-dev:
If desired, I could probably split this into 2 smaller PRs:
- refactor Validator and ValidationResult
    - most of these changes are simple where instead of returning `anyhow::Error`, we return `ValidationResult::Invalid(...)`, but they affect a lot of lines of code
    - changes in portalnet (i.e. `offer.rs`) to actually use functions for accepting and validating multiple content together

PR #1869, opened 2025-05-30
refactor: add batch content validation

morph-dev:
This was extracted from #1868

### What was wrong?

`Validator` is currently able to validate only one content at the time. This is not enough for ephemeral headers.
Similar for `Store` when answering whether content is/should be stored.

### How was it fixed?

Rename `ContentStore::is_key_within_radius_and_unavailable` to `ContentStore::should_we_store` and add `Store::should_we_store_batch` (for multiple items.

Add `Validator::validate_content_batch` for validating multiple content items.

Refactor `ValidationResult`:
- removed `additional_content_to_propagate` from ValidationResult (it's not used)
- made ValidationResult an enum
    - enum type works better as a result of `validate_content_batch`, because it is `Vec<ValidationResult>` instead of `Vec<Result<ValidationResult>>`
    - we can consider in the future changing `Invalid(String)` variant into more descriptive type, e.g. `Invalid(ValidationFailureReason)` where:
        ```rust
        enum ValidationFailureReason {
            ContentValueNotDecodable,
            Invalid(<reason?>),  // e.g. merkle proof not valid
            FetchingAnchorContentFailuer(<reason?>), // e.g. failed to fetch Header when verifying BlockBody
            ...
        }
        ``` 

### Future work

Start using _batch_ functions from `offer.rs` (#1870) and add custom implementation of the batch functionality for history network (future PR).

### DIFF
--- a/crates/subnetworks/beacon/src/validation.rs
+++ b/crates/subnetworks/beacon/src/validation.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use alloy::primitives::B256;
 use anyhow::anyhow;
 use chrono::Duration;
 use ethportal_api::{
@@ -59,17 +58,17 @@ impl Validator<BeaconContentKey> for BeaconValidator {
     async fn validate_content(
         &self,
         content_key: &BeaconContentKey,
-        content: &[u8],
-    ) -> anyhow::Result<ValidationResult<BeaconContentKey>> {
+        content_value: &[u8],
+    ) -> ValidationResult {
         match content_key {
             BeaconContentKey::LightClientBootstrap(_) => {
-                let bootstrap = ForkVersionedLightClientBootstrap::from_ssz_bytes(content)
-                    .map_err(|err| {
-                        anyhow!(
-                            "Fork versioned light client bootstrap has invalid SSZ bytes: {:?}",
-                            err
-                        )
-                    })?;
+                let Ok(bootstrap) =
+                    ForkVersionedLightClientBootstrap::from_ssz_bytes(content_value)
+                else {
+                    return ValidationResult::Invalid(
+                        "Error decoding light client bootstrap content value".to_string(),
+                    );
+                };
 
                 // Check if the light client bootstrap slot is ole than 4 months
                 let four_months = Duration::days(30 * 4);
@@ -79,9 +78,8 @@ impl Validator<BeaconContentKey> for BeaconValidator {
                 let bootstrap_slot = bootstrap.get_slot();
 
                 if bootstrap_slot < four_months_ago_slot {
-                    return Err(anyhow!(
-                        "Light client bootstrap slot is too old: {}",
-                        bootstrap_slot
+                    return ValidationResult::Invalid(format!(
+                        "Light client bootstrap slot is too old: {bootstrap_slot}",
                     ));
                 }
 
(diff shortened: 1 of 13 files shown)


crates/subnetworks/beacon/src/validation.rs
```diff
@@ -90,24 +88,23 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
-                            "Light client updates by range has invalid SSZ bytes: {:?}",
-                            err
-                        )
-                    })?;
+                let Ok(lc_updates) = LightClientUpdatesByRange::from_ssz_bytes(content_value)
+                else {
+                    return ValidationResult::Invalid(
+                        "Error decoding Light client updates content value".to_string(),
+                    );
+                };
```
KolbyML, line 103: same comment as above?

crates/subnetworks/beacon/src/validation.rs
```diff
@@ -126,35 +123,39 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
-                        err
-                    )
-                })?;
+                let Ok(lc_finality_update) =
+                    ForkVersionedLightClientFinalityUpdate::from_ssz_bytes(content_value)
+                else {
+                    return ValidationResult::Invalid(
+                        "Error decoding Light client finality update content value".to_string(),
+                    );
+                };
```
KolbyML, line 154: same comment as above

crates/subnetworks/beacon/src/validation.rs
```diff
@@ -182,36 +183,38 @@ impl Validator<BeaconContentKey> for BeaconValidator {
...
-                    )
-                        },
-                    )?;
+                let Ok(lc_optimistic_update) =
+                    ForkVersionedLightClientOptimisticUpdate::from_ssz_bytes(content_value)
+                else {
+                    return ValidationResult::Invalid(
+                        "Error decoding Light client optimistic update content value".to_string(),
+                    );
+                };
```
KolbyML, line 213: same comment as above

crates/subnetworks/history/src/validation.rs
```diff
@@ -36,98 +34,127 @@ impl Validator<HistoryContentKey> for ChainHistoryValidator {
...
             HistoryContentKey::BlockReceipts(key) => {
-                let receipts = Receipts::from_ssz_bytes(content).map_err(|msg| {
-                    anyhow!("Block Receipts content has invalid encoding: {:?}", msg)
-                })?;
-                let trusted_header: Header = self
+                let Ok(receipts) = Receipts::from_ssz_bytes(content_value) else {
+                    return ValidationResult::Invalid(
+                        "Block Receipts content has invalid encoding".to_string(),
+                    );
+                };
```
KolbyML, line 124: Why not pass the error up?

morph-dev:
Since all the comments are on the same topic, I will just reply here.

I don't think that logging or returning `ssz::DecodeError` is useful. I mean, it's useful to know that it's decoding that failed but the debug message of the failure is in my opinion not useful at all because:
- It will just say something like: `InvalidByteLength { len: 1000, expected: 104 }` or `OutOfBoundsByte { i: 1000 }` which isn't useful at all
- And even if you would try to understand the actual message and what specific byte is wrong (which I don't see a reason for doing anyway), the error message can come from decoding any of the inner fields (or inner-inner fields) and offset would be very off so you wouldn't be able to do it anyway.

So I decided to just return error that indicates that error happened because of the decoding (and not because of e.g. validation).

I now added `debug!()` log with the actual content value that decoding failed for, which would actually be useful if somebody wants to debug the issue.

---

I'm not too set on this. If you think we should propagate error I can do so, but I would like to hear the argument for it.

KolbyML:
Would we be able to embeded the debugging context in the ContentValidation::Invalid() instead of defining a debug! Message? 

I am not too picky I think logging it is definitely an improvement regardless. Up to you. It seems wasteful to have both though

KolbyML:
I am fine not propagating the error in this case, my question is would anybody ever have the debug logs enabled

morph-dev:
> Would we be able to embeded the debugging context in the ContentValidation::Invalid() instead of defining a debug! Message?

I don't want to embed debug message inside `Invalid(..)` variant as `content_value` can be very big. That's why I think it should only be used as part of the `debug!` message.

> I am fine not propagating the error in this case, I agree what you did it higher value, my question is would anybody ever have the debug logs enabled.

Our prod nodes have `debug` enabled.

If somebody is debugging this bug, they would enable it. Otherwise, it's probably not useful anyway.

---

I will leave it as is, and we can change it in the future if needed.

PR #1870, opened 2025-05-30
feat: use batch validation

morph-dev:
This is followup for #1869, which should be reviewed and merged first.

### What was wrong?

When we are offered multiple content, the overlay service does all of the following on per-content basis:

- checks whether they should be accepted
- validates them
- stores them

This doesn't work for EphemeralHeaders, which we have to process together (at least checking which content to accept and validating). This PR uses batch functions added in #1869.

### How was it fixed?

The `OverlayService::handle_offer` now uses these `batch` functions to process all accepted content together.

### Remaining work (future PRs)

We would still do a fallback fetching if request fails or received content is not valid.
I would have to think more whether this is desired for EphemeralContent, and if not, how exactly to distinguish/handle this case.

Modify history Store and Validator to actually implement custom logic for handling multiple content items

### DIFF
--- a/crates/portalnet/src/overlay/service/offer.rs
+++ b/crates/portalnet/src/overlay/service/offer.rs
@@ -12,14 +12,12 @@ use ethportal_api::{
         portal_wire::{
             Accept, Content, FindContent, Offer, OfferTraceMultipleItems, Request, Response,
         },
-        protocol_versions::ProtocolVersion,
     },
     OverlayContentKey, RawContentKey, RawContentValue,
 };
-use futures::{channel::oneshot, future::join_all};
+use futures::{channel::oneshot, future::JoinAll};
 use itertools::Itertools;
 use parking_lot::Mutex;
-use tokio::task::JoinHandle;
 use tracing::{debug, enabled, error, trace, warn, Level};
 use trin_storage::{ContentStore, ShouldWeStoreContent};
 use trin_validation::validator::Validator;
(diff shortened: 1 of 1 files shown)


carver:
> We would still do a fallback fetching if request fails or received content is not valid.
> I would have to think more whether this is desired for EphemeralContent, and if not, how exactly to distinguish/handle this case.

I don't think the same logic applies for content that is supposed to end up in every node. We can either just wait for the next offer to come in, or ask some random node for it. The reasoning was that just a couple bridges inject content into just a couple locations. So it seems unnecessary for this pattern than gossips everywhere. (Good background reading in the issue and its related PR https://github.com/ethereum/trin/issues/856 )

carver (commented):
I didn't do a broad and deep review, but it looks good

carver (approved):

crates/portalnet/src/overlay/service/offer.rs
```diff
@@ -245,63 +262,39 @@ impl<
...
+                        // validation
+                        let utp_processing = utp_processing.clone();
+                        tokio::spawn(async move {
+                            if let Err(err) = Self::fallback_find_content(
+                                key.clone(),
+                                utp_processing,
+                                protocol_version
+                            )
+                            .await {
+                                debug!(%err, ?key, "Fallback FINDCONTENT task failed, after validating and storing content failed");
```
KolbyML, line 289: I am surprised you switched this spawn to not be awaited, because in that one bridge PR you said we should await all spawned tasks, what do you think is different about this situation? compared to the e2hs bridge?

KolbyML (approved):
:shipit: overall the PR looks good

crates/portalnet/src/overlay/service/offer.rs
morph-dev: You comment made me look at this PR in details again and you are right, we should await all these. Here is why initially I thought we don't have to.

We actually don't use the result of `Self::fallback_find_content` anywhere in this function. We treat this content item as failed (regardless whether fallback was successful).

However, I realized that we are actually holding `inbound permit` while in this function, and that we should hold it until all these fallback tasks finish as well.
If we weren't holding the `inbound permit`, I would say there would be no need to await these tasks.

---

The logic was similar with e2hs bridge. There, if I remember correctly, we don't want to download, decode, process and gossip multiple e2hs files in parallel, as that can be quite resource intensive. So it's good to have control of each task from each file and await them together.

Using semaphore on each individual gossip doesn't actually make these guarantees.

morph-dev:
While looking at the exact logic and code, I found few more issues with the existing implementation of fallback logic, so I refactored a bit:
- the `protocol_version` that we used corresponds to the original peer, not to the fallback peer
- we used to fetch all offered content keys, not only accepted ones (I fixed this in one place in the initial commit, but I missed one more)
- extracted fallback logic for multiple content keys into separate function (and added documentation with more details)
- don't use fallback logic for content that is not affected by radius
    - I agree with Jason that in this case we could maybe use any random peer (not just the ones that offered content to use)
    - since ephemeral headers use different content key for `offer` and `find_content`, I decided to just disable this functionality (we can revist this in the future)

